Issue #5018, opened 2021-07-13
Support statuses from multiple controllers


MitchAman:
Add support for having statuses from more than one controller

kdorosh:
This is useful and intended for testing control plane canaries

PR #5020, opened 2021-07-14
Add support for statuses from multiple controllers

MitchAman:
# Description

Add support for having statuses from more than one controller.  This PR depends on https://github.com/solo-io/solo-kit/pull/429, which creates the `NamespacedStatuses` proto message and necessary generated code.  Resources which previously contained a `Status` now have a new `oneof` containing the existing `Status` and a new `NamespacedStatuses`. The `oneof` allows for backwards compatibility with respect to to read operations.  All write operations going forward will use the `NamespacedStatuses`.

# Context

Multiple gloo instances in different namespaces would previously all write to the same `Status`.  This is problematic for canary architectures where the instances may disagree on the Status of a particular resource.  This PR provides a way for multiple gloo instances to each maintain their own status.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5018


projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -75,6 +75,8 @@ func (s *translatorSyncer) Sync(ctx context.Context, snap *v1.ApiSnapshot) error
 		logger.Debug(syncutil.StringifySnapshot(snap))
 	}
 
+	logger.Debugw("snap", "stringified", syncutil.StringifySnapshot(snap))
```
kdorosh, line 78: todo: remove

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -505,14 +519,18 @@ func checkRouteOptions(ctx context.Context, namespaces []string) ([]string, erro
...
+				switch status.GetState() {
+				case core.Status_Rejected:
+					errMessage := fmt.Sprintf("Found rejected RouteOption by '%s': %s ", reporter, renderMetadata(routeOpt.GetMetadata()))
+					errMessage += fmt.Sprintf("(Reason: %s)", status.Reason)
+					multiErr = multierror.Append(multiErr, errors.New(errMessage))
+				case core.Status_Warning:
+					errMessage := fmt.Sprintf("Found RouteOption with warnings by '%s': %s ", reporter, renderMetadata(routeOpt.GetMetadata()))
+					errMessage += fmt.Sprintf("(Reason: %s)", status.Reason)
+					multiErr = multierror.Append(multiErr, errors.New(errMessage))
+				}
```
kdorosh, line 533: does the new `glooctl check` crash or fail to report errors on older gloo versions? (i.e., the old status format?)

projects/gloo/cli/pkg/printers/helpers.go
```diff
@@ -40,3 +43,20 @@ func PrintKubeCrdList(in resources.InputResourceList, resourceCrd crd.Crd) error
...
+	var sb strings.Builder
+	var firstController = true
+	for controller, status := range reporterStatus.GetStatuses() {
+		if !firstController {
+			sb.WriteString("\n")
+		}
+		sb.WriteString(controller)
+		sb.WriteString(": ")
+		sb.WriteString(statusProcessor(status))
+		firstController = false
```
kdorosh, line 59: instead of having `firstController` variable, we could just write the newline last

kdorosh: and skip the newline if we're on the last item in the loop

projects/gloo/cli/pkg/cmd/check/root.go
MitchAman: Ugh, totally overlooked that, thanks!

At the moment it will not crash, but won't report errors for `core.Status` objects.

projects/gloo/pkg/plugins/kubernetes/uds_convert.go
```diff
@@ -133,7 +133,7 @@ func upstreamsEqual(original, desired *v1.Upstream) bool {
 	copyDesired := *desired
 
 	copyOriginal.Metadata = copyDesired.Metadata
-	copyOriginal.Status = copyDesired.Status
+	copyOriginal.SetReporterStatus(copyDesired.GetReporterStatus())
```
kdorosh, line 136: let's copy both over? and if we have the reporter status, let's only copy that one

projects/knative/pkg/translator/translator_syncer.go
```diff
@@ -170,12 +170,13 @@ func (s *translatorSyncer) propagateProxyStatus(ctx context.Context, proxy *gloo
 			if err != nil {
 				return err
 			}
-			switch updatedProxy.Status.State {
+			// TODO(mitchaman): Should this be gloo?
+			switch updatedProxy.GetStatusForReporter("gloo").GetState() {
```
kdorosh, line 174: i think we should remove the `reportedBy` part, but i believe this would say `knative`

projects/gloo/pkg/plugins/kubernetes/uds_convert.go
MitchAman: We can't copy both because the oneof will clear whichever was copied first.  Updated to check which form of status the copyDesired has.

go.mod
```diff
@@ -73,7 +73,8 @@ require (
 	github.com/solo-io/skv2 v0.17.2
 	// Pinned to the `rate-limiter-v0.1.8` tag of solo-apis
 	github.com/solo-io/solo-apis v0.0.0-20210122162349-0e170e74af10
-	github.com/solo-io/solo-kit v0.21.1
+	// TODO(mitchaman): Update once https://github.com/solo-io/solo-kit/pull/429 is merged/released
```
MitchAman, line 76: Address this

projects/gloo/api/v1/proxy.proto
```diff
@@ -43,12 +43,16 @@ message Proxy {
...
     repeated Listener listeners = 2;
 
-    // Status indicates the validation status of this resource.
-    // Status is read-only by clients, and set by gloo during validation
-    core.solo.io.Status status = 6 [(extproto.skip_hashing) = true];
+    oneof status_oneof {
+        // Status indicates the validation status of this resource.
+        // Status is read-only by clients, and set by gloo during validation
+        core.solo.io.Status status = 6 [(extproto.skip_hashing) = true];
+        core.solo.io.ReporterStatus reporter_status = 8 [(extproto.skip_hashing) = true];
```
saiskee, line 50: Can we copy the comments to everywhere with `reporter_status`?
```suggestion
        // ReporterStatus indicates the validation status of this resource.
        // ReporterStatus is read-only by clients, and set by gloo during validation
        core.solo.io.ReporterStatus reporter_status = 8 [(extproto.skip_hashing) = true];
```

projects/gloo/api/v1/proxy.proto
```diff
@@ -315,9 +319,12 @@ message UpstreamGroup {
...
     repeated WeightedDestination destinations = 1;
 
-    // Status indicates the validation status of this resource.
-    // Status is read-only by clients, and set by gloo during validation
-    core.solo.io.Status status = 6 [(extproto.skip_hashing) = true];
+    oneof status_oneof {
+        // Status indicates the validation status of this resource.
+        // Status is read-only by clients, and set by gloo during validation
+        core.solo.io.Status status = 6 [(extproto.skip_hashing) = true];
+        core.solo.io.ReporterStatus reporter_status = 8 [(extproto.skip_hashing) = true];
```
saiskee, line 326: Copy reporter_status comment here.

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -328,17 +328,34 @@ func checkUpstreams(ctx context.Context, namespaces []string) ([]string, error)
...
-				multiErr = multierror.Append(multiErr, errors.New(errMessage))
-			}
-			if upstream.Status.GetState() == core.Status_Warning {
-				errMessage := fmt.Sprintf("Found upstream with warnings: %s ", renderMetadata(upstream.GetMetadata()))
-				errMessage += fmt.Sprintf("(Reason: %s)", upstream.Status.Reason)
-				multiErr = multierror.Append(multiErr, errors.New(errMessage))
+			if upstream.HasReporterStatus() {
+				reporterStatus := upstream.GetReporterStatus()
+				for reporter, status := range reporterStatus.GetStatuses() {
+					switch status.GetState() {
```
saiskee, line 334: What is an example of what a `reporter` would be here? Would it just be the namespace, like `gloo-system`?

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -328,17 +328,34 @@ func checkUpstreams(ctx context.Context, namespaces []string) ([]string, error)
...
+					case core.Status_Rejected:
+						errMessage := fmt.Sprintf("Found rejected upstream by '%s': %s ", reporter, renderMetadata(upstream.GetMetadata()))
```
saiskee, line 336: [nit] Found rejected upstream reported by '%s'? I'm not sure if it will make sense though, depending on what the `reporter` is..

saiskee: If the `reporter` is just the namespace, can we specify that in the error message, just to make it clear for users what the first %s is.

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -388,16 +421,33 @@ func checkAuthConfigs(ctx context.Context, namespaces []string) ([]string, error
...
 		for _, authConfig := range authConfigs {
-			if authConfig.Status.GetState() == core.Status_Rejected {
-				errMessage := fmt.Sprintf("Found rejected auth config: %s ", renderMetadata(authConfig.GetMetadata()))
-				errMessage += fmt.Sprintf("(Reason: %s)", authConfig.Status.Reason)
-				multiErr = multierror.Append(multiErr, errors.New(errMessage))
-			} else if authConfig.Status.GetState() == core.Status_Warning {
-				errMessage := fmt.Sprintf("Found auth config with warnings: %s ", renderMetadata(authConfig.GetMetadata()))
-				errMessage += fmt.Sprintf("(Reason: %s)", authConfig.Status.Reason)
-				multiErr = multierror.Append(multiErr, errors.New(errMessage))
+			if authConfig.HasReporterStatus() {
```
saiskee, line 424: Do we not need to check if it `HasStatus()` as well? Or does `HasReporterStatus()` get the `Status` if `ReporterStatus` doesn't exist?

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -388,16 +421,33 @@ func checkAuthConfigs(ctx context.Context, namespaces []string) ([]string, error
...
+				reporterStatus := authConfig.GetReporterStatus()
+				for reporter, status := range reporterStatus.GetStatuses() {
+					switch status.GetState() {
+					case core.Status_Rejected:
+						errMessage := fmt.Sprintf("Found rejected auth config by '%s': %s ", reporter, renderMetadata(authConfig.GetMetadata()))
```
saiskee, line 429: This logic seems to be used a lot, maybe we can extract this shared logic into `func checkReporterStatus() error` and `func checkStatus() error`
and `multierror.Append` the error they return.

projects/gloo/cli/pkg/cmd/create/output_format_test.go
```diff
@@ -38,14 +38,14 @@ static:
...
-+--------------------+--------+---------+---------------------------------+
-| jsonplaceholder-80 | Static | Pending | hosts:                          |
-|                    |        |         | -                               |
-|                    |        |         | jsonplaceholder.typicode.com:80 |
-|                    |        |         |                                 |
-+--------------------+--------+---------+---------------------------------+`
+		tableOutput := `+--------------------+--------+--------+---------------------------------+
+|      UPSTREAM      |  TYPE  | STATUS |             DETAILS             |
++--------------------+--------+--------+---------------------------------+
+| jsonplaceholder-80 | Static |        | hosts:                          |
```
saiskee, line 44: Why did we clear this out? What should be in place here?

projects/gloo/cli/pkg/cmd/check/root.go
MitchAman: There is an else which covers the Status case. Because they're in a oneof, if there's no ReporterStatus then we assume there's a Status.

projects/gloo/cli/pkg/cmd/check/root.go
MitchAman: That's correct.  I'm going to re-work a lot of this to address your other comments

install/helm/gloo/crds/gateway.solo.io_v1_VirtualService.yaml
```diff
@@ -23,6 +23,14 @@ spec:
           properties:
             displayName:
               type: string
+            namespacedStatuses:
```
saiskee, line 26: https://github.com/solo-io/solo-kit/pull/429#pullrequestreview-719057649
> It seems like we're now writing the namespacesStatuses to the spec as well as the status, on this InputResource, whereas previously we were not.

I think the `status` field did exist under the `spec` previously ([here](https://github.com/solo-io/gloo/blob/c585ecab6dc2fd2c8a3748b0ef80f9dd0253772a/install/helm/gloo/crds/gateway.solo.io_v1_VirtualService.yaml#L116)). It just wasn't being populated by us. The `namespacedStatus` should live at [this level](https://github.com/solo-io/gloo/blob/c585ecab6dc2fd2c8a3748b0ef80f9dd0253772a/install/helm/gloo/crds/gateway.solo.io_v1_VirtualService.yaml#L116) instead.

kdorosh:
superseded by https://github.com/solo-io/gloo/pull/5191

PR #5029, opened 2021-07-15
make route names more unique

jenshu:
# Description

In the gateway translator, make the generated route names more unique in an attempt to avoid duplicate name errors. We now take into account the gateway and proxy name, virtual service namespace, and the route index.

# Context

Users ran into this bug when configuring basic rate limits on Virtual Services.

# Checklist:


MitchAman (approved):
LGTM

projects/gateway/pkg/translator/converter.go
```diff
@@ -36,6 +37,9 @@ var (
 	InvalidQueryParamErr = errors.New("invalid route: route table matchers must have all query params that were specified on their parent route's matcher")
 	InvalidMethodErr     = errors.New("invalid route: route table matchers must have all methods that were specified on their parent route's matcher")
 
+	UnnamedRoute = func(index int) string {
```
sam-heilbron, line 40: Why do unnamed routes now include an index?

sam-heilbron (commented):
There were a number of cases you had identified during testing where this was broken previously (multiple gateways referencing same proxy, multiple proxies relying on the same vs, routes across namespaces with the same name). How have we verified that those cases are resolved with this fix?

projects/gateway/pkg/translator/converter.go
jenshu: this is to handle the case where a VS delegates to a RouteTable with multiple unnamed routes. Currently you'd get a duplicate name error because each unnamed route from the same RouteTable will end up with the same generated name

sam-heilbron: Isn't one of the requirements of the BasicRateLimit config that routes are named uniquely? I figured, if someone provides unnamed routes, we should be failing, since we want to enforce a unique name on each.

jenshu: (spoke offline and agreed to keep this in)
currently we require the top level route name (i.e. in the VirtualService) to have an explicit name, but do not have any such requirement for routes in delegated RouteTables. this change will ensure that if the RouteTable routes are unnamed, we auto-generate names for uniqueness

sam-heilbron (approved):

Issue #4900, opened 2021-06-18
CI Test Flake: Kube2e: gateway tests for the validation server disable_transformation_validation is set [It] will not reject invalid transformation 


sam-heilbron:
**Describe the bug**
Gateway regression test flake:
Kube2e: gateway tests for the validation server disable_transformation_validation is set [It] will not reject invalid transformation 

**Additional context**
https://github.com/solo-io/gloo/runs/2850002066

sam-heilbron:
This has been fixed on master and 1.8.x. We opted to not backport the fix to other LTS branches since there were merge conflicts, however if the flake appears more frequently on those branches, we can extract some of this work and apply it to those branches.

PR #5033, opened 2021-07-15
Ensure invalid virtual hosts are not duplicated on proxy

sam-heilbron:
# Description
- Fixed logic in our proxy reconciler, which allowed invalid virtual hosts to be added to a proxy twice. 
- Add some utility methods for our Kube2e tests, to make it easier for developers to add logic around waiting for resources to be accepted and cleaned up
- Fix test pollution that did lead to flakes
- Fix robustness_test, which was written in a way that relied on the exact bug that this PR fixes, to be present
- Fix disable_transformation_test (https://github.com/solo-io/gloo/issues/4900) which would flake occasionally, and did during the development of this PR 

# Context

### Mix of valid and invalid virtual services
Example logs: https://console.cloud.google.com/cloud-build/builds/a13c32e3-fddf-4263-b28a-e931933d9f87;step=14?project=solo-public

This test flake was due to a bug in our proxy_reconciler. We incorrectly were duplicating virtual hosts on the proxy object. The proxy only transitioned in certain situations (depending on metadata of the object) which is why this didn't occur more frequently. However, when it did happen, it caused this flaking test.

I added a unit test to verify the change that I made fixes the problem. 
I also added a comment to the code so that future developers would understand the reason for the change.

### Does not equal length of map
Sometimes, I ran into a Gateway test (I don't have the logs handy) that produced the following error:
```
[
gloo-system-gateway-443_gloo-system-16567209401833135566:true 
gloo-system-gateway-proxy-443_gloo-system-7590740161429284937:true 
gloo-system-gateway-proxy-80_gloo-system-7992917100079133684:true 
gloo-system-gloo-9976_gloo-system-12285151928708083525:true 
gloo-system-gloo-9977_gloo-system-7484856654061583909:true 
gloo-system-gloo-9979_gloo-system-5451814594336066521:true 
gloo-system-gloo-9988_gloo-system-1559281721111094449:true 
gloo-system-testrunner-1234_gloo-system-10782942849661275980:true 
kube-svc:gloo-system-gateway-443_gloo-system-10160320302259603858:true 
kube-svc:gloo-system-gateway-proxy-443_gloo-system-13361889343733600589:true 
kube-svc:gloo-system-gateway-proxy-80_gloo-system-15324810366434219216:true 
kube-svc:gloo-system-gloo-9976_gloo-system-6098176559075073065:true 
kube-svc:gloo-system-gloo-9977_gloo-system-8667782289304402933:true
 kube-svc:gloo-system-gloo-9979_gloo-system-1598973103534690965:true
 kube-svc:gloo-system-gloo-9988_gloo-system-6721171892144873845:true 
kube-svc:gloo-system-testrunner-1234_gloo-system-1741026987556571732:true
] 

does not equal length of map

[
gloo-system-gateway-443_gloo-system-16567209401833135566:0xc0008e6ce0 
gloo-system-gateway-proxy-443_gloo-system-7590740161429284937:0xc0008e6da0 
gloo-system-gateway-proxy-80_gloo-system-7992917100079133684:0xc0008e6e50 
gloo-system-gloo-9976_gloo-system-12285151928708083525:0xc0008e6ec0 
gloo-system-gloo-9977_gloo-system-7484856654061583909:0xc0008e6f30 
gloo-system-gloo-9979_gloo-system-5451814594336066521:0xc0008e6fa0
gloo-system-gloo-9988_gloo-system-1559281721111094449:0xc0008e7010 
gloo-system-redbluezqw6b-5678_gloo-system-6470157035427515329:0xc0008e74b0
gloo-system-testrunner-1234_gloo-system-10782942849661275980:0xc0008e7060 
kube-svc:gloo-system-gateway-443_gloo-system-10160320302259603858:0xc0008e70d0 
kube-svc:gloo-system-gateway-proxy-443_gloo-system-13361889343733600589:0xc0008e7190 
kube-svc:gloo-system-gateway-proxy-80_gloo-system-15324810366434219216:0xc0008e7270 
kube-svc:gloo-system-gloo-9976_gloo-system-6098176559075073065:0xc0008e72f0 
kube-svc:gloo-system-gloo-9977_gloo-system-8667782289304402933:0xc0008e7360 
kube-svc:gloo-system-gloo-9979_gloo-system-1598973103534690965:0xc0008e73d0 
kube-svc:gloo-system-gloo-9988_gloo-system-6721171892144873845:0xc0008e7440 
kube-svc:gloo-system-redbluezqw6b-5678_gloo-system-5865765942139567028:0xc0008e74d0 
kube-svc:gloo-system-testrunner-1234_gloo-system-1741026987556571732:0xc0008e7490
]
```
I believe this is just due to test pollution, since the mismatched service (ie the service in one map but not the other) is `gloo-system-redbluezqw6b-5678_gloo-system-6470157035427515329:0xc0008e74b0`. There is only one test that uses this service, so I added an Eventually to ensure it is cleaned up after the test 

### Gateway robustness test changes
This test intentionally wrote a virtual service with duplicate domains, which triggers the exact bug that this PR fixes. After introducing the fix, and not changing the test at all, the test consistently fails, because it expects the proxy to enter a Rejected state. This would happen before, but with the fix, invalid virtual hosts are stripped properly, and not duplicated, so the proxy no longer enters a rejected state.

I updated the test, and simplified some of the logic to make it consistent with another test that is validating similar behavior.

### Disable transformation validation kube2e test flake

This flake was due to behavior of the gateway validation server. When the gateway validation settings change (as we do in this test), the server restarts and we need to afford it enough time to reboot. I re-used a pattern we have relied on in other tests, to ensure that the settings are given enough time to propagate before moving forward with the test execution.

# Checklist:

# Testing
I have another branch with the tests re-written in a way to make running a single test over and over easier/faster. As of now, we can supply the `untilItFails` flag to ginkgo, but since Kube2e tests install gloo and uninstall it before/after Suites, I re-wrote the tests so that I could repeat tests without having to re-install gloo each time. This was just useful for debugging, and I didn't add the code to the PR. If we're interested in a true debugging utility for kube2e tests, there are likely better ways to set that up. The branch is: https://github.com/solo-io/gloo/tree/regression-test-flakes

Here are the steps you can run to run these tests quickly:
1. Checkout code from testing branch: `git checkout regression-test-flakes`
1. Checkout code from this branch: `git checkout fix-regression-test-flakes`
1.  Build images with version, and push to kind cluster. `VERSION=v1.8.2 CLUSTER_NAME=regression-flakes CLUSTER_NODE_VERSION=v1.21.1 ./ci/deploy-to-kind-cluster.sh`
1. Checkout files from test branch: `git checkout regression-test-flakes test/kube2e/util.go`,  `git checkout regression-test-flakes test/kube2e/gateway/gateway_test.go`
1. Run focused regression tests: ` KUBE2E_TESTS=gateway TEAR_DOWN=true WAIT_ON_FAIL=1 make run-ci-regression-tests
`

This will run the previously flaking test 100 times in a row. Before this change (you can verify for yourself), if I ran this, most times a flake would occur within the 100 times. However, with this change, I consistently complete all 100 without an error.

### DIFF
--- a/projects/gateway/pkg/reconciler/proxy_reconciler.go
+++ b/projects/gateway/pkg/reconciler/proxy_reconciler.go
@@ -192,16 +192,18 @@ func stripInvalidListenersAndVirtualHosts(ctx context.Context, proxiesToWrite Ge
 // which is invalid and will be rejected by Envoy
 func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc {
 	return func(original, desired *gloov1.Proxy) (b bool, e error) {
-		// if any listeners from the old proxy were rejected in the new reports, preserve those
-		if err := forEachListener(original, proxiesToWrite[desired], func(listener *gloov1.Listener, accepted bool) {
-			// old listener was rejected, preserve it on the desired proxy
-			if !accepted {
-				desired.Listeners = append(desired.Listeners, listener)
-			}
-		}); err != nil {
-			// should never happen
-			return false, err
-		}
+
+		// We intentionally process desired.Listeners first, and then original.Listeners second
+		// We modify the desired proxy object and have to perform 2 steps:
+		//	- Apply invalid listeners to the desired proxy
+		// 	- Apply invalid vhosts to the desired listener
+		// Since we are modifying the proxy directly, if we process invalid listeners first,
+		// we will append those, and then try to process invalid virtual hosts on those same
+		// listeners again, causing us to write the virtual host twice: first when we
+		// copied the stripped listener, second when we copied the stripped virtual host.
+		// To avoid this, we first process desired.Listeners to reconcile invalid virtual
+		// hosts on those listeners, and then process the original.Listeners to reconcile
+		// invalid listeners.
 
 		// preserve previous vhosts if new vservice was errored
 		for _, desiredListener := range desired.Listeners {
@@ -239,6 +241,17 @@ func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc 
 				return desiredHttpListener.VirtualHosts[i].Name < desiredHttpListener.VirtualHosts[j].Name
 			})
 
+		}
+
+		// if any listeners from the old proxy were rejected in the new reports, preserve those
+		if err := forEachListener(original, proxiesToWrite[desired], func(listener *gloov1.Listener, accepted bool) {
+			// old listener was rejected, preserve it on the desired proxy
+			if !accepted {
+				desired.Listeners = append(desired.Listeners, listener)
+			}
+		}); err != nil {
+			// should never happen
+			return false, err
 		}
 
 		sort.SliceStable(desired.Listeners, func(i, j int) bool {
--- a/projects/gateway/pkg/reconciler/proxy_reconciler_test.go
+++ b/projects/gateway/pkg/reconciler/proxy_reconciler_test.go
@@ -30,6 +30,17 @@ import (
 
 var _ = Describe("ReconcileGatewayProxies", func() {
 
+	// DEVELOPER NOTE: Listeners and VirtualHosts are sorted by name.
+	// Therefore, there are test cases that compare
+	//	Gateway[0] -> Listener[1]
+	//	or
+	//	VirtualService[1] -> VirtualHost[0]
+	//
+	// This is because ordering of the Gateway and VirtualServices
+	// are defined by the user.
+	// But the ordering of the Listener and VirtualHosts are
+	// defined by the name of those resources.
+
 	var (
 		ctx = context.TODO()
 
@@ -198,25 +209,53 @@ var _ = Describe("ReconcileGatewayProxies", func() {
 		})
 
 		Context("a virtual service has a reported error", func() {
+
 			It("only updates the valid virtual hosts", func() {
 				samples.AddVsToSnap(snap, us, ns)
 				genProxy()
 				reconcile()
 
 				snap.VirtualServices[0].Metadata.Generation = 100
-				snap.VirtualServices[1].Metadata.Generation = 100
+				snap.VirtualServices[1].Metadata.Generation = 101
 				genProxy()
 				addErr(snap.VirtualServices[1])
 
 				reconcile()
 
 				px := getProxy()
-
 				vhosts := px.Listeners[1].GetHttpListener().GetVirtualHosts()
 
 				Expect(vhosts).To(HaveLen(2))
-				Expect(vhosts[1]).To(HaveGeneration(100))
+				Expect(vhosts[1]).To(HaveGeneration(100)) // vhosts[1] maps to VirtualServices[0]
 				Expect(vhosts[0]).To(HaveGeneration(0))
+			})
+
+			It("only updates the valid virtual hosts, without duplicating any", func() {
+				samples.AddVsToSnap(snap, us, ns)
+				genProxy()
+				reconcile()
+
+				// Update the Generation value, to ensure that proxy reconciliation decides
+				// to transition and persist a new proxy
+				snap.Gateways[0].Metadata.Generation = 100
+				snap.Gateways[1].Metadata.Generation = 100
+				snap.VirtualServices[0].Metadata.Generation = 100
+				snap.VirtualServices[1].Metadata.Generation = 101
+
+				genProxy()
+
+				// Add an error on the Gateway, ensuring the Listener is not accepted
+				addErr(snap.Gateways[0])
+				// Add an error on the VirtualService, ensuring the VirtualHost is not accepted
+				addErr(snap.VirtualServices[0])
+
+				reconcile()
+
+				px := getProxy()
+				vhosts := px.Listeners[1].GetHttpListener().GetVirtualHosts()
+
+				// We still only have 2 VirtualServices, which should always translate to 2 VirtualHosts
+				Expect(vhosts).To(HaveLen(2))
 			})
 		})
 
--- a/test/kube2e/gateway/gateway_test.go
+++ b/test/kube2e/gateway/gateway_test.go
@@ -1806,43 +1806,14 @@ spec:
 				"header template ':status': [inja.exception.parser_error] expected statement close, got '%'")))
 		})
 
-		// Using a seperate Context here in order to take advantage of Before/After Each.
-		// They are safer for cleaning up state as they will run regardless of whether a test fails
 		Context("disable_transformation_validation is set", func() {
 
-			var (
-				settingsClient gloov1.SettingsClient
-			)
-
 			BeforeEach(func() {
-				settingsClient = clienthelpers.MustSettingsClient(ctx)
-
-				settingsList, err := settingsClient.List(testHelper.InstallNamespace, clients.ListOpts{})
-				Expect(err).NotTo(HaveOccurred())
-				Expect(settingsList).To(HaveLen(1))
-				settings := settingsList[0]
-
-				settings.Gateway.Validation.DisableTransformationValidation = &wrappers.BoolValue{
-					Value: true,
-				}
-
-				_, err = settingsClient.Write(settings, clients.WriteOpts{
-					OverwriteExisting: true,
-				})
-				Expect(err).NotTo(HaveOccurred())
-
+				kube2e.UpdateDisableTransformationValidationSetting(ctx, true, testHelper.InstallNamespace)
 			})
 
 			AfterEach(func() {
-				settingsList, err := settingsClient.List(testHelper.InstallNamespace, clients.ListOpts{})
-				Expect(err).NotTo(HaveOccurred())
-				Expect(settingsList).To(HaveLen(1))
-				settings := settingsList[0]
-				settings.Gateway.Validation.DisableTransformationValidation = nil
-				_, err = settingsClient.Write(settings, clients.WriteOpts{
-					OverwriteExisting: true,
-				})
-				Expect(err).NotTo(HaveOccurred())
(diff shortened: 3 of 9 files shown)


test/kube2e/gateway/gateway_test.go
```diff
@@ -590,93 +603,111 @@ var _ = Describe("Kube2e: gateway", func() {
...
 			var vs *gatewayv1.VirtualService
-			BeforeEach(func() {
 
-				kube2e.UpdateSettings(func(settings *gloov1.Settings) {
-					Expect(settings.Gloo).NotTo(BeNil())
-					Expect(settings.Gloo.InvalidConfigPolicy).NotTo(BeNil())
-					settings.Gloo.InvalidConfigPolicy.ReplaceInvalidRoutes = true
-				}, ctx, testHelper.InstallNamespace)
+			BeforeEach(func() {
+				kube2e.UpdateReplaceInvalidRoutes(ctx, true, testHelper.InstallNamespace)
```
kdorosh, line 710: Why are we setting this now? it might help suppress the flake, but ideally we can solve the issue here with your virtual host deduplication above

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -228,7 +228,12 @@ func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc
 			if err := forEachVhost(originalListener, proxiesToWrite[desired], func(vhost *gloov1.VirtualHost, accepted bool) {
 				// old vhost was rejected, preserve it on the desired proxy
 				if !accepted {
-					desiredHttpListener.VirtualHosts = append(desiredHttpListener.VirtualHosts, vhost)
+					if isVhostInList(vhost, desiredHttpListener.VirtualHosts) {
+						// vhost already contained in desiredHttpListener, do nothing
+					} else {
+						desiredHttpListener.VirtualHosts = append(desiredHttpListener.VirtualHosts, vhost)
+					}
+
```
kdorosh, line 236: looks like this is the only real functional change here, correct? everything else appears to be test refactor to DRY, cleanup test pollution, and updating tests to use route replacement

test/kube2e/gateway/gateway_test.go
sam-heilbron: Oh this is functionally equivalent. I wanted all validation settings updates to live next to eachother in case we ever remove the time.Sleep that we rely on to wait for validation chages to propagate.

projects/gateway/pkg/reconciler/proxy_reconciler.go
sam-heilbron: Correct, this is the only functional change

sam-heilbron: Happy to split this out into its own PR to make clearer

test/kube2e/gateway/gateway_test.go
```diff
@@ -491,37 +495,46 @@ var _ = Describe("Kube2e: gateway", func() {
...
+					return virtualServiceClient.Read(testHelper.InstallNamespace, invalidVsName, clients.ReadOpts{})
+				}, "15s", "0.5s")
+
 				_ = virtualServiceClient.Delete(testHelper.InstallNamespace, validVsName, clients.DeleteOpts{})
+				helpers.EventuallyResourceDeleted(func() (resources.InputResource, error) {
+					return virtualServiceClient.Read(testHelper.InstallNamespace, validVsName, clients.ReadOpts{})
+				}, "15s", "0.5s")
+
 				_ = virtualServiceClient.Delete(testHelper.InstallNamespace, petstoreName, clients.DeleteOpts{})
-				_ = kubeClient.CoreV1().Services(testHelper.InstallNamespace).Delete(ctx, petstoreName, metav1.DeleteOptions{})
```
sam-heilbron, line 518: I moved these into a lower context, since they were only being used for a single test. I wanted the creation and deletion of resources to be as closely defined to where they were being used as possible

test/kube2e/gateway/gateway_test.go
```diff
@@ -491,37 +495,46 @@ var _ = Describe("Kube2e: gateway", func() {
 						},
 					}, nil)))
 
-				Eventually(func() error {
```
sam-heilbron, line 494: We don't need to wrap writes in eventually

test/kube2e/gateway/gateway_test.go
```diff
@@ -1476,6 +1528,11 @@ var _ = Describe("Kube2e: gateway", func() {
 				Expect(err).NotTo(HaveOccurred())
 			}
 
+			// Ensure the redblue service is deleted
```
sam-heilbron, line 1531: This fixed the other test flake that I alluded to in the PR description

test/kube2e/gateway/robustness_test.go
```diff
@@ -41,12 +42,6 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
-const (
```
sam-heilbron, line 44: I moved these down to the one test that was using them

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -228,7 +228,15 @@ func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc
 			if err := forEachVhost(originalListener, proxiesToWrite[desired], func(vhost *gloov1.VirtualHost, accepted bool) {
 				// old vhost was rejected, preserve it on the desired proxy
 				if !accepted {
-					desiredHttpListener.VirtualHosts = append(desiredHttpListener.VirtualHosts, vhost)
+					if isVhostInList(vhost, desiredHttpListener.VirtualHosts) {
+						// vhost already contained in desiredHttpListener, do nothing
+						// this can happen since we're saving entire Listeners earlier in the transitionFunc
+						// and then attempting to save the same VirtualHost if we generated a new one
+						// that is not accepted.
```
kdorosh, line 235: we want the newly generated valid vhost to replace the one that was saved from the listener that's already deployed in the cluster, rather than reusing old config (in the event there is a diff between the two vhosts)

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -192,16 +192,18 @@ func stripInvalidListenersAndVirtualHosts(ctx context.Context, proxiesToWrite Ge
 // which is invalid and will be rejected by Envoy
 func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc {
 	return func(original, desired *gloov1.Proxy) (b bool, e error) {
-		// if any listeners from the old proxy were rejected in the new reports, preserve those
```
sam-heilbron, line 195: This was just moved to lower in the function. No changes were made to the logic.

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
```diff
@@ -30,6 +30,17 @@ import (
 
 var _ = Describe("ReconcileGatewayProxies", func() {
 
+	// DEVELOPER NOTE: Listeners and VirtualHosts are sorted by name.
```
sam-heilbron, line 33: This confused me when reading tests. I opted to limit the scope of my change and just flag the behavior, instead of changing the name of the virtual hosts to make vs[0] map to vhost[0]

kdorosh (dismissed):
🔥 love the unit test and fix by changing code execution order

sam-heilbron:
Test flake: https://github.com/solo-io/gloo/pull/5033/checks?check_run_id=3127277163. This has existed and I had hoped would be resolved by this work, but sadly isn't. If intentionally kicking it once to see if it passes. If so, ill monitor how frequently this occurs.

test/kube2e/gateway/robustness_test.go
```diff
@@ -112,79 +96,57 @@ var _ = Describe("Robustness tests", func() {
 		err = proxyClient.Register()
 		Expect(err).NotTo(HaveOccurred())
 
-		upstreamClient, err = gloov1.NewUpstreamClient(ctx, upstreamClientFactory)
```
sam-heilbron, line 115: upstreamClient only existed because we were writing an upstream. But that upstream wasn't used anywhere so I deleted it

test/kube2e/gateway/robustness_test.go
```diff
@@ -112,79 +96,57 @@ var _ = Describe("Robustness tests", func() {
...
-		err = upstreamClient.Register()
-		Expect(err).NotTo(HaveOccurred())
-
-		appDeployment, appService, err = createDeploymentAndService(kubeClient, namespace, appName)
+		appDeployment, appService, err = createEchoDeploymentAndService(kubeClient, testHelper.InstallNamespace, appName)
 		Expect(err).NotTo(HaveOccurred())
 	})
 
 	AfterEach(func() {
-		if virtualService != nil {
```
sam-heilbron, line 125: I moved the virtual service cleanup into the context that used it

test/kube2e/gateway/robustness_test.go
```diff
@@ -112,79 +96,57 @@ var _ = Describe("Robustness tests", func() {
...
-				return len(services.Items) == 0
-			}, "15s", "0.5s").Should(BeTrue())
-		}
+			return len(services.Items) == 0
+		}, "15s", "0.5s").Should(BeTrue())
 
 		cancel()
 	})
 
-	It("updates Envoy endpoints even if proxy is rejected", func() {
```
sam-heilbron, line 160: These two tests shared lots of setup steps. I moved them into a context with shared setup and cleanup

kdorosh (approved):

kdorosh (approved):

Issue #5043, opened 2021-07-19
cannot use Gateway's socketOptions in v1.8.0


bcollard:
**Describe the bug**
When using `socketOptions` on a Gateway CR, the Gateway controller reports error like this:
```json
{"level":"error","ts":1626686105.8066683,"logger":"gateway.v1.event_loop.gateway","caller":"syncer/setup_syncer.go:296","msg":"error: event_loop: 1 error occurred:\n\t* reconciling resource gateway-proxy: creating kube resource gateway-proxy: Proxy.gloo.solo.io \"gateway-proxy\" is invalid: [spec.listeners.options.socketOptions.level: Invalid value: \"string\": spec.listeners.options.socketOptions.level in body must be of type integer: \"string\", spec.listeners.options.socketOptions.name: Invalid value: \"string\": spec.listeners.options.socketOptions.name in body must be of type integer: \"string\", spec.listeners.options.socketOptions.intValue: Invalid value: \"string\": spec.listeners.options.socketOptions.intValue in body must be of type integer: \"string\"]\n\n","version":"1.8.0","stacktrace":"github.com/solo-io/gloo/projects/gateway/pkg/syncer.RunGateway.func1\n\t/workspace/gloo/projects/gateway/pkg/syncer/setup_syncer.go:296"}
```

Helm values example:
```yaml
gloo:
  gatewayProxies:
    gatewayProxy:
      gatewaySettings:
        options:
          socketOptions:
            - description: "enable keep-alive"
              level: 1
              name: 9
              intValue: 1
              state: STATE_LISTENING
            - description: "idle time before first keep-alive probe is sent"
              level: 6
              name: 4
              intValue: 60
              state: STATE_LISTENING
            - description: "keep-alive interval"
              level: 6
              name: 5
              intValue: 20
              state: STATE_LISTENING
            - description: "keep-alive probes count"
              level: 6
              name: 6
              intValue: 2
              state: STATE_LISTENING
```

The helm-generated `gateway-proxy` CR:
```yaml
...
spec:
  bindAddress: '::'
  bindPort: 8080
  httpGateway: {}
  options:
    socketOptions:
    - description: enable keep-alive
      intValue: 1
      level: 1
      name: 9
      state: STATE_LISTENING
    - description: idle time before first keep-alive probe is sent
      intValue: 60
      level: 6
      name: 4
      state: STATE_LISTENING
    - description: keep-alive interval
      intValue: 20
      level: 6
      name: 5
[...]

Issue #4789, opened 2021-05-27
Current virtualService and Proxy CRD seem invalid due an annotation of too much bytes


wilmardo:
**Describe the bug**
Annotations have too much bytes and they can't be applied. Seems to be done in #4757 

Also nice moment to upgrade to the new `apiextensions.k8s.io/v1 CustomResourceDefinition` instead of the deprecated `apiextensions.k8s.io/v1beta1` see the first warning in the output.

**To Reproduce**
```console
$ kubectl apply -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/enterprise.gloo.solo.io_v1_AuthConfig.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gateway.solo.io_v1_Gateway.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gateway.solo.io_v1_RouteTable.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gateway.solo.io_v1_VirtualService.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gloo.solo.io_v1_Proxy.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gloo.solo.io_v1_Settings.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gloo.solo.io_v1_Upstream.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/gloo.solo.io_v1_UpstreamGroup.yaml -f https://raw.githubusercontent.com/solo-io/gloo/master/install/helm/gloo/crds/ratelimit_config.yaml
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
customresourcedefinition.apiextensions.k8s.io/authconfigs.enterprise.gloo.solo.io created
customresourcedefinition.apiextensions.k8s.io/gateways.gateway.solo.io created
customresourcedefinition.apiextensions.k8s.io/routetables.gateway.solo.io created
customresourcedefinition.apiextensions.k8s.io/settings.gloo.solo.io created
[...]

PR #5050, opened 2021-07-20
Validation Schemas: remove descriptions and support int64 primitive type

sam-heilbron:
Relies on: https://github.com/solo-io/solo-kit/pull/431

# Description
- Fix issue with serialization of int64's and validation schemas: https://github.com/solo-io/gloo/issues/5043
- Remove descriptions from validation schemas: https://github.com/solo-io/gloo/issues/4789

# Context
## int64's in validation schemas (https://github.com/solo-io/gloo/issues/5043)

Int64's are serialized as strings. Since we introduced validation schemas, the schema expects an integer, but the value provided is a stringified version of an integer (ie "2"). A recent commit to the library we use for schema generation (https://github.com/sam-heilbron/protoc-gen-openapi/commit/e703b93862db646a58a3c5d0a83777e2c59b692d) fixed this, by including the `x-kubernetes-int-or-string` property (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#intorstring) on the schema.

### Testing
- [x] Ran an end-to-end test to verify that this new field would behave as expected:
1. I pointed to the solo-kit PR (https://github.com/solo-io/solo-kit/pull/431) that introduced this feature
1. I built the chart, and created a test cluster: `VERSION=v0.0.0-helm CLUSTER_NAME=helm-test CLUSTER_NODE_VERSION=v1.21.1 ./ci/deploy-to-kind-cluster.sh`
1. I installed gloo using that new chart: `helm install gloo _test/gloo-v0.0.0-helm.tgz --namespace gloo-system -f /path/to/yaml/file --create-namespace`, where path/to/yaml/file is the path to the values override file presented in the issue
1. I added a route so that a proxy would be generated, and verified that there was no error
1. I manually changed the validation schema on the proxy to remove the new `x-kubernetes-int-or-string` property, updated the route so the proxy would re-generate, and confirmed that an error now occurred.

## descriptions in validation schemas (https://github.com/solo-io/gloo/issues/4789)
Descriptions in validation schemas aren't functional, shouldn't be used as a source of truth around the API, and don't affect the error message presented when a resource errors due to validation schema constraints. Also, since descriptions are generated from the comments above the source proto field, and those comments also generate docs in Gloo, descriptions are quite verbose. This causes the CRD to be massive, containing lots of information that isn't relevant to the CRD itself. If the CRD get's too large, it can no longer be applied, which is what happened in this issue.

I opted to remove descriptions altogether. We could play the game to find the exact number of characters that each descriptions should have (this is a supported feature of solo-kit), but the first 15-20 characters alone are not informative, and are not used by the customer.
 
### Testing
- [x] Add tests to verify that all CRDs can be applied successfully. This failed when I ran it on master (with excessively large CRDs) and passed on this branch

# Checklist:


install/helm/gloo/crds/enterprise.gloo.solo.io_v1_AuthConfig.yaml
```diff
@@ -20,33 +20,17 @@ spec:
     openAPIV3Schema:
       properties:
         spec:
-          description: This is the user-facing auth configuration. When processed
```
saiskee, line 23: image

saiskee (approved):
LGTM!

PR #5074, opened 2021-07-27
Remove extra `oneWayTls` from UpstreamOptions in Settings proto

saiskee:


projects/gloo/api/v1/settings.proto
```diff
@@ -389,9 +389,6 @@ message Settings {
 message UpstreamOptions {
     // Default ssl parameter configuration to use for upstreams
     SslParameters ssl_parameters = 1;
-
-    // Default one way TLS configuration to use for upstreams
-    google.protobuf.BoolValue one_way_tls = 2;
```
sam-heilbron, line 394: Flagging that we're intentionally not marking this field as reserved, since we never released the code changes that introduced this field in the first place

elcasteel (approved):

PR #5092, opened 2021-07-30
Fix ci

kdorosh:


kdorosh:
/kick data race flake? https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4d60d2a2-042d-47c8-854f-b023b15f5283

jenshu (approved):

PR #5095, opened 2021-07-30
[wip] Dont run regression tests on docs changes

saiskee:
Don't run regression tests on branches where docs only changes / changelog only changes. This will help increase the velocity of docs-only change PRs.

Note: Currently github actions will report `Expected -- Waiting for status to be reported` when using `on.pull_request.paths` or `on.pull_request.paths-ignore`. Expected behaviour (see below) is that the workflow will not run, and be skipped and not block CI.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths
> The filter determines if a workflow should run by evaluating the changed files and running them against the paths-ignore or paths list. If there are no files changed, the workflow will not run.


saiskee:
closing; regression tests are required under branch protection rules, so they will continue to report as "Expected" if skipped.

PR #5105, opened 2021-08-02
Http passthrough api

saiskee:
## Summary
Implementing a passthrough HTTP ext-auth system, similar to our currently offered [GRPC passthrough auth](3da073419ea12fd8/pkg/config/passthrough/grpc/grpc_auth_service.go) system. 

## Context
Users want more flexibility from our auth-service. They can either roll their own auth service, use an ext-auth plugin (which is REALLY inconvenient to maintain), or use grpc passthrough auth. The main pros and cons of these systems are highlighted [here](https://docs.solo.io/gloo-edge/latest/guides/security/auth/extauth/passthrough_auth/). Http passthrough auth will offer everything that gRPC auth offers, with the added flexibility of using a regular REST server rather than a gRPC server.


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,54 @@ message PassThroughGrpc {
     google.protobuf.Duration connection_timeout = 2;
 }
 
+message PassThroughHttp {
```
jenshu, line 759: can you add a brief description of PassThroughHttp here?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
...
+
+
+```yaml
+"uri": string
+"path": string
+"request": .enterprise.gloo.solo.io.PassThroughHttp.Request
+"response": .enterprise.gloo.solo.io.PassThroughHttp.Response
+"passThroughState": bool
+"passThroughFilterMetadata": bool
+"passThroughBody": bool
```
sam-heilbron, line 1056: What's the relationship between these 3? Is there a reason I'd set one and not the others? What's the effect on the performance if I do not set these?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
 
 
 
+---
+### PassThroughHttp
+
+
+
+```yaml
+"uri": string
```
sam-heilbron, line 1050: Can we add an example, and specify that this is a fully qualified domain name (is it?)

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
...
+"path": string
```
sam-heilbron, line 1051: What's the purpose of this field? I know it's a setting on the http ext_authz envoy filter, but for what purpose would someone set this header? Also, how does this value mix with the request.HeaderToAdd? Does one overwrite the other?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
...
+
+```yaml
+"allowedHeaders": []string
+"headersToAdd": map<string, string>
+
+```
+
+| Field | Type | Description |
+| ----- | ---- | ----------- | 
+| `allowedHeaders` | `[]string` | These headers will be copied from the incoming request to the request going to the auth server. Note that in addition to the user's supplied matchers: 1. *Host*, *Method*, *Path* and *Content-Length* are automatically included to the list. 2. *Content-Length* will be set to 0 and the request to the authorization service will not have a message body. |
```
sam-heilbron, line 1088: will the request to the authorization service not have a request body?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
...
+
+```yaml
+"allowedUpstreamHeaders": []string
+"allowedClientHeaders": []string
+
+```
+
+| Field | Type | Description |
+| ----- | ---- | ----------- | 
+| `allowedUpstreamHeaders` | `[]string` | When this is set, authorization response headers that have a header in this list will be added to the original client request and sent to the upstream. Note that coexistent headers will be overridden. |
```
sam-heilbron, line 1107: What's the use case for these settings?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -1036,6 +1041,75 @@ https://github.com/envoyproxy/envoy/blob/ae1ed1fa74f096dabe8dd5b19fc70333621b030
 
 
 
+---
+### PassThroughHttp
```
sam-heilbron, line 1045: I think we should clarify that a POST request will be executed from the auth service to the configured URI. Can you think of a situation where a user would want to use a GET request instead? 

How are retries handled?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
saiskee: I'm not sure what you mean by relationship here. The passthrough request body to the custom auth server will look like this when all three are `true`:
```json
{ 
  "body" : "some body",
  "filterMetadata": {
    "io.solo.http.passthrough": [
       "some metadata"
    ],
  "state": {
    "key1": "value1", 
    "key2": "value2"
}
```
I have provided the option to not pass through any one of these, because json unmarshalling is a costly operation, and this is on the request path. The body will be completely empty and no unmarshalling will be performed if all three of these are `false`.

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
saiskee: I'm removing this field in favor of setting the path in the `url` field. In addition, after further testing, pseudo-headers can not be set in the `allowed_headers` or `headers_to_add` field because they are not real headers according to http/1. I added a comment to those fields about this.

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
saiskee: These are similar settings to what we offer with our http_service extauth: https://docs.solo.io/gloo-edge/master/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk/#httpservice

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
saiskee: As of right now, I think leaving it as POST request is sufficient, because allowing the user to specify the method might unnecessarily complicate things. e.g. If they set `pass_through_body` to `true` but they set the method as `GET`, then do we override the method to `POST` to accomodate for the request body, or do we ignore the body completely?

Retry configuration is not handled by the client in this case. I think if they were to want retry behaviour, they could request their auth requests through envoy and use retries as specified on the virtual service.

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
sam-heilbron: Is there a situation where a client would want to set 1 and not the others? I figured the only decision a client cares about is whether to parse the body or not. Would it simplify things if we just exposed a single flag, similar to how we expose parseBodyBehavior in transformations, and if the client sets it to parse (or some setting) we include the body in the request, with the body, filterMetadata and state, and if not, we don't pass any.

elcasteel (dismissed):

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/kube/client/informers/externalversions/enterprise.gloo.solo.io/v1/authconfig.go
```diff
@@ -61,13 +62,13 @@ func NewFilteredAuthConfigInformer(client versioned.Interface, namespace string,
 				if tweakListOptions != nil {
 					tweakListOptions(&options)
 				}
-				return client.EnterpriseV1().AuthConfigs(namespace).List(options)
+				return client.EnterpriseV1().AuthConfigs(namespace).List(context.TODO(), options)
```
saiskee, line 65: Regenerated the authconfig client by running this script. https://github.com/solo-io/gloo/blob/dbc80cd5b787937a6b9647f16fde458bf9ca36be/projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/kube/hack/update-codegen.sh#L2

Otherwise the authclient.go is currently broken on master:
https://github.com/solo-io/gloo/blob/3682a443cb04ab58d4ef881c029cc2fd1a349b12/projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/kube/client/clientset/versioned/typed/enterprise.gloo.solo.io/v1/authconfig.go#L33

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,59 @@ message PassThroughGrpc {
...
 }
 
+// Authorizes requests by making a POST request to a custom HTTP auth server
+// Assumes the request is authorized if the server returns a OK (200) status code,
+// else the request is unauthorized.
+message PassThroughHttp {
+    // Required: URL of the passthrough http service, is a fully qualified domain name.
+    // Example: http://ext-auth-service.svc.local:9001.
+    // Path provided in the URL will be respected.
+    string url = 1;
```
kdorosh, line 766: is https supported?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,59 @@ message PassThroughGrpc {
...
+
+    message Request {
+        // These headers will be copied from the incoming request to the request going
+        // to the auth server. By default, no headers are copied from the incoming request.
+        // Pseudo-headers such as `:Path`, and `:Method` can not be specified here.
```
kdorosh, line 771: so we only support http2?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,59 @@ message PassThroughGrpc {
...
+
+        // Whether or not to include the filter metadata in the passthrough request body.
+        // If pass_through_body, pass_through_filter_metadata and pass_through_state are false,
+        // the authorization request body will be empty.
+        bool pass_through_filter_metadata = 4;
+
+        // Whether or not to include the body in the passthrough request body.
+        // If pass_through_body, pass_through_filter_metadata and pass_through_state are false,
+        // the authorization request body will be empty.
+        bool pass_through_body = 5;
```
kdorosh, line 792: we would we ever want these to be false? feel like i might be missing something obvious here

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
sam-heilbron: Bump on this^

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
sam-heilbron: Makes sense!

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/kube/client/informers/externalversions/enterprise.gloo.solo.io/v1/authconfig.go
sam-heilbron: Why do we need to regenerate the authconfig client for this change and not previous API changes?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sam-heilbron: My understanding was that setting these to true (one or all) adds latency to the request, and that if a client does not need to read/write state to be shared by other filters, they would set these to false. Is that accurate @saiskee ?

kdorosh: if so, we should include that tradeoff in the docs here for users

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,59 @@ message PassThroughGrpc {
...
+
+    message Response {
+        // When this is set, authorization response headers that have a header in this list will be added to the original client request and sent to the upstream
+        // when the auth request is successful.
+        // If this is empty, by default, no authorization response headers will be added to the upstream request.
+        repeated string allowed_upstream_headers = 1;
+
+        // When this is set, authorization response headers in this list will be added to the client's response when the auth request is denied.
+        // If this is empty, by default, no authorization response headers will be added to the client response.
+        repeated string allowed_client_headers_on_denied = 2;
```
kdorosh, line 804: how does it get added if the header already existed? comma delimited values, or replace the value?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,59 @@ message PassThroughGrpc {
...
+    }
+    Response response = 4;
+
+    // Timeout for the auth server to respond. Defaults to 5s
+    google.protobuf.Duration connection_timeout = 8;
```
kdorosh, line 809: default to 2s? 5s seems aggressive

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/kube/client/informers/externalversions/enterprise.gloo.solo.io/v1/authconfig.go
saiskee: This doesn't have anything to do with the API changes. I just noticed that we had not run clientgen in a while for the authconfig client and it was currently broken. I'm not sure we ever use this client, but figured I'd rather leave it in a working state.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
saiskee: Only http1. The golang client we are using only supports http1 and can not set pseudo headers.

kdorosh: ok, let's make that clear in the docs here

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
saiskee: replace. added comments.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
saiskee: https://github.com/solo-io/ext-auth-service/pull/241#discussion_r684416316

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,66 @@ message PassThroughGrpc {
     google.protobuf.Duration connection_timeout = 2;
 }
 
+// Authorizes requests by making a POST HTTP/1 request to a custom HTTP auth server
+// Assumes the request is authorized if the server returns a OK (200) status code,
+// else the request is unauthorized.
+message PassThroughHttp {
+    // Required: URL of the passthrough http service, is a fully qualified domain name.
+    // Example: http://ext-auth-service.svc.local:9001. Path provided in the URL will be respected.
+    // To use https, provide the cert in the HTTPS_PASSTHROUGH_CA_CERT environment variable as a base64-encoded string
```
sam-heilbron, line 765: Which deployment/pod requires this env variable?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sam-heilbron: With grpc passthrough, we had opted to expect requests to be insecure, and delegate security to a sidecar (https://github.com/solo-io/ext-auth-service/blob/master/pkg/config/passthrough/grpc_client_manager.go#L41). Any particular reason we don't do something similar for this auth service as well?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -755,6 +756,66 @@ message PassThroughGrpc {
...
+        // Whether or not to include the body in the passthrough request body.
+        // If pass_through_body, pass_through_filter_metadata and pass_through_state are false,
+        // the authorization request body will be empty. A non-empty body will increase latency times
+        // slightly, so this is set to false by default, and should only be set to to true if the
+        // request body is needed in the auth request.
+        bool pass_through_body = 5;
+    }
+    Request request = 3;
+
+    message Response {
```
sam-heilbron, line 802: We have a couple flags that collectively determine whether a body is included in the request. Do those same flags affect parsing the response as well?

saiskee: Only `pass_through_state` does, included a comment about this.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
saiskee: I think the cost of securing with https was only a few lines of code, so it was okay to include that as part of this work.

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
saiskee: I can imagine a situation where a client only wants to read the state but doesn't ever care about the request body. In a situation like that, where request bodies can become huge, it would really help latency to be able to flag that, which is why I've allowed each one to be configured separately.

sam-heilbron (approved):

PR #5129, opened 2021-08-09
Fix Potential NPEs

saiskee:
There are a couple instances in our code where we don't use the Getters provided by protobuf library. This has caused us NPE panics in customer's production instances. The linter used to find and fix these issues automatically is still under development, found under https://github.com/saiskee/goprotogettercheck.

### DIFF
--- a/pkg/utils/selectionutils/virtual_service.go
+++ b/pkg/utils/selectionutils/virtual_service.go
@@ -85,11 +85,11 @@ func (s *virtualServiceSelector) build(ctx context.Context, ref *core.ResourceRe
 			Domains: []string{"*"},
 		},
 	}
-	if virtualService.GetMetadata().Namespace == "" {
-		virtualService.Metadata.Namespace = s.podNamespace
+	if virtualService.GetMetadata().GetNamespace() == "" {
+		virtualService.GetMetadata().Namespace = s.podNamespace
 	}
-	if virtualService.GetMetadata().Name == "" {
-		virtualService.Metadata.Name = "default"
+	if virtualService.GetMetadata().GetName() == "" {
+		virtualService.GetMetadata().Name = "default"
 	}
 
 	return virtualService, nil
--- a/projects/gateway/pkg/syncer/translator_syncer.go
+++ b/projects/gateway/pkg/syncer/translator_syncer.go
@@ -94,8 +94,8 @@ func (s *translatorSyncer) generatedDesiredProxies(ctx context.Context, snap *v1
 				compress.SetShouldCompressed(proxy)
 			}
 
-			logger.Infof("desired proxy %v", proxy.Metadata.Ref())
-			proxy.Metadata.Labels = s.managedProxyLabels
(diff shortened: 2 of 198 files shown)


pkg/utils/selectionutils/virtual_service.go
```diff
@@ -85,11 +85,11 @@ func (s *virtualServiceSelector) build(ctx context.Context, ref *core.ResourceRe
 			Domains: []string{"*"},
 		},
 	}
-	if virtualService.GetMetadata().Namespace == "" {
-		virtualService.Metadata.Namespace = s.podNamespace
+	if virtualService.GetMetadata().GetNamespace() == "" {
+		virtualService.GetMetadata().Namespace = s.podNamespace
```
sam-heilbron, line 89: If Metadata is nil, won't we still NPE here? I know we're safe with Metadata, since Metadata is always available, but wanted to call out that when setting a field, using an getter on an intermediate field won't protect us.

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -94,8 +94,8 @@ func (s *translatorSyncer) generatedDesiredProxies(ctx context.Context, snap *v1
 				compress.SetShouldCompressed(proxy)
 			}
 
-			logger.Infof("desired proxy %v", proxy.Metadata.Ref())
-			proxy.Metadata.Labels = s.managedProxyLabels
+			logger.Infof("desired proxy %v", proxy.GetMetadata().Ref())
+			proxy.GetMetadata().Labels = s.managedProxyLabels
```
jenshu, line 98: hmm should we be using the getters on the left side of assignment operations?

EItanya:
These 2 previous comments bring up the limitation of this kind of checking, e.g. `Setters`. I'm not sure there's an obviously good answer for them, as they are so reliant on the code that precedes them. This is something we could build another static checker for, but I think using the getters is a good practice anyway, and doesn't change the behavior of setting lines in any meaningful way where it's worth filtering them out.

sam-heilbron:
> These 2 previous comments bring up the limitation of this kind of checking, e.g. `Setters`. I'm not sure there's an obviously good answer for them, as they are so reliant on the code that precedes them. This is something we could build another static checker for, but I think using the getters is a good practice anyway, and doesn't change the behavior of setting lines in any meaningful way where it's worth filtering them out.

Definitely agreed and I probably should have clarified this in my comment: I'm not proposing we change the setters to not use the getters, I just wanted to point out that this change will not protect us against NPEs in setters.

sam-heilbron:
This is awesome! At the moment this will be run by devs manually to be notified of potential NPE's and then it's up to the dev to resolve it, is that correct? Can we include the command you ran to determine these, so others can leverage it as well? Are there plans to have this linter run during CI at some point, once the tool is complete?

saiskee:
The tool is currently a WIP, but should be completed in the next few days. This PR was a passthrough of the entire repo, and I will have a follow up PR to include the linter tool as part of codegen / CI.

Issue #5085, opened 2021-07-28
helm chart invalid field validation_server_grpc_max_size


asayah:
gatewayProxies.NAME.gatewaySettings.options.validation.validation_server_grpc_max_size.value
is invalid, this setting belong in the setting crd.

PR #5137, opened 2021-08-10
Replace validationServerGrpcMaxSize with validationServerGrpcMaxSizeBytes

MitchAman:
# Description

Remove `validationServerGrpcMaxSize`. Instead use `validationServerGrpcMaxSizeBytes`, which is a replacement for `validationServerGrpcMaxSize`. `validationServerGrpcMaxSizeBytes` is now an int32, which fixes the int64 version's broken functionality.

This fixes https://github.com/solo-io/gloo/issues/5085

# Context

gRPC server `maxReceiveMessageSize` is an int32 (https://github.com/grpc/grpc-go/blob/a42567fe92f005c47e60146bdbb0d5f7fc232219/server.go#L155), but our `validationServerGrpcMaxSize` value is an int64.  We are attempting to cast from int64 to int32 when initializing the validation server: https://github.com/solo-io/gloo/blob/2aeb6563a36692b95916a88fb4b6cc76f35750ca/projects/gloo/pkg/syncer/setup/setup_syncer.go#L243.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5085
resolves https://github.com/solo-io/gloo/issues/5160


projects/gloo/api/v1/settings.proto
```diff
@@ -566,7 +566,7 @@ message GatewayOptions {
         // Setting this value sets the gRPC max message size in bytes for the gloo validation server. This should
         // only be changed if necessary.
         // If not included, the gRPC max message size will be the default of 4 MB.
-        google.protobuf.Int64Value validation_server_grpc_max_size = 10;
+        int32 validation_server_grpc_max_size = 10;
```
sam-heilbron, line 569: Are we safe to convert from a message to a primitive here?

MitchAman: Per our discussion earlier, I've changed this to `google.protobuf.Int32Value`

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=45008eda-a17e-4e47-9a8d-4329599eacdb data race

MitchAman:
/kick https://console.cloud.google.com/cloud-build/builds;region=global/f0884d67-c7d6-43aa-97a8-982e346efc39;step=10?project=solo-public data race

install/helm/gloo/crds/gloo.solo.io_v1_Settings.yaml
```diff
@@ -210,13 +210,11 @@ spec:
...
-                    validationServerGrpcMaxSize:
-                      properties:
-                        value:
-                          format: int64
-                          type: integer
-                          x-kubernetes-int-or-string: true
-                      type: object
+                    validationServerGrpcMaxSizeBytes:
+                      maximum: 2147483647
+                      minimum: -2147483648
```
bewebi, line 215: Seems unlikely we'll have negative numbers of bytes
Is there any advantage (or disadvantage) to using `unit32` in order to double the maximum and remove the negative case? Even if we do want to use a signed int, it seems like we could make the minimum on the CRD 0 (or 1) to better reflect the use case.

MitchAman: I had considered this, but the gRPC `serverOptions` requires an int (https://github.com/grpc/grpc-go/blob/master/server.go#L155), and we'd run into issues when passing this value through here: https://github.com/solo-io/gloo/blob/bb110e837508742e6de4de069ac375e7d031854e/projects/gloo/pkg/syncer/setup/setup_syncer.go#L243

Do you know how to make codegen modify the min/max values in the CRD?

bewebi: That tracks
I didn't realize the CRD was from codegen, but that makes sense...I don't know how to modify that but I'm not sure I'd want to
Did you/can we try providing a <=0 value and see what happens to make sure it's not too catastrophic and/or provides a clear error message

MitchAman: Good idea, I'll give that a try

MitchAman: When `validationServerGrpcMaxSizeBytes` is set to -5MB, the gateway pod crash loops with this fatal log message:
```
{"level":"fatal","ts":1629131521.7243476,"logger":"gateway","caller":"setuputils/main_setup.go:73","msg":"error in setup: failed to read notifications from stream: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (0 vs. -5000000)","version":"1.9.0-beta8-46-gbb110e837","stacktrace":"... (omitted) ..."}
```
Do you think that's sufficient?

bewebi: I don't love it, especially since it doesn't name the setting per se. If you had just tweaked the one setting and then all of a sudden saw that it would be pretty clear, but if you had changed a handful of things and/or were setting up from scratch, it might not be clear what to search for in order to troubleshoot. 

It would be preferable to detect that and crash before getting to the point of a failed grpc request. 
That said, this was a size XS and L impact, it seems relatively unlikely that a user would actually set such a value, and if they did and sent us that error statement we could easily get it straightened out. 
If it's fairly quick to achieve a better error out put and/or earlier in the flow that would be great, but otherwise sufficient IMO to document in some form on our side the behavior that results from that type of bad input.

MitchAman: That makes sense. I went ahead and added a check for this in `setup_syncer.go` and verified that I did indeed see the new error message in logs with a negative `validationServerGrpcMaxSizeBytes`

bewebi (dismissed):
🚀

bewebi (approved):

PR #5145, opened 2021-08-11
Add Getter check linter

saiskee:
Adds a tool to CI which lints for unused getters.

This will help us prevent a lot of future NPEs by using generated getters, which are NPE-safe when accessing.


Makefile
```diff
@@ -138,6 +138,7 @@ install-go-tools: mod-download
 	GOBIN=$(DEPSGOBIN) go install github.com/golang/mock/gomock
 	GOBIN=$(DEPSGOBIN) go install github.com/golang/mock/mockgen
 	GOBIN=$(DEPSGOBIN) go install github.com/onsi/ginkgo/ginkgo
+	GOBIN=$(DEPSGOBIN) go get github.com/saiskee/gettercheck
```
sam-heilbron, line 141: I know I'm no one to talk given `go install github.com/sam-heilbron/protoc-gen-openapi` but should we either put this in a solo repo now, or at least create a github issue to track migrating both of these tools to the solo org?

sam-heilbron: Also why `go get`?

Makefile
```diff
@@ -190,6 +191,7 @@ $(OUTPUT_DIR)/.generated-code:
 	PATH=$(DEPSGOBIN):$$PATH rm docs/content/reference/cli/glooctl*; GO111MODULE=on go run projects/gloo/cli/cmd/docs/main.go
 	PATH=$(DEPSGOBIN):$$PATH gofmt -w $(SUBDIRS)
 	PATH=$(DEPSGOBIN):$$PATH goimports -w $(SUBDIRS)
+	PATH=$(DEPSGOBIN):$$PATH gettercheck -ignoretests -ignoregenerated -write ./...
```
sam-heilbron, line 194: What's the behavior if a developer fails to use a getter? Will this fail the generated-code GHA?

saiskee: If the developer fails to use a getter and doesn't run codegen, there will be a diff. This replaces all unused getter calls with their respective Getters. so

```go
a := &envoy.VirtualHost{}
_ = a.Options
```
will be replaced with 

```go
a := &envoy.VirtualHost{}
_ = a.GetOptions()
```
by running `gettercheck -write`. It's similar to if you forget to run `goimports` locally, CI will complain because there will be a diff.

saiskee: I've created an issue to migrate over the tools here: https://github.com/solo-io/gloo/issues/5155

Makefile
EItanya: You can put it in `build/tools.go` to fix this I believe

sam-heilbron (approved):

PR #5162, opened 2021-08-13
Backport disable tls session resumption to 1.6.x

ben-taussig-solo:
### Description
- adding an option to disable tls session resumption (Ticket only)
- Backport of #5136


sam-heilbron (dismissed):

go.mod
```diff
@@ -74,7 +74,7 @@ require (
 	github.com/solo-io/protoc-gen-ext v0.0.15
 	github.com/solo-io/reporting-client v0.2.0
 	github.com/solo-io/skv2 v0.15.2
-	github.com/solo-io/solo-apis v0.0.0-20201208172621-1ef725df2bff
+	github.com/solo-io/solo-apis v1.6.31-0.20210803135442-d12f957649db
```
sam-heilbron, line 77: Why do we need to update solo-apis?

ben-taussig-solo: I don't have a great answer to this question. The reason I've done so is because I've been introducing changes made in other recent dependency bump/CI fix PRs to try and resolve CI failures with this PR. This change is what finally resolved the CI failures. I'm going to temporarily revert this change this change, record the CI failures that result, and post them under this comment to give more clarity into why this is necessary.

sam-heilbron (approved):

go.mod
ben-taussig-solo: huh, it passed without that change. I don't know why CI was failing without that change earlier, but I'm not even going to worry about it.

PR #5169, opened 2021-08-13
Fips installation option

elcasteel:
# Description
Add an entry in values.go for fips option for images.

### DIFF
--- a/install/helm/gloo/generate/values.go
+++ b/install/helm/gloo/generate/values.go
@@ -54,7 +54,7 @@ type Image struct {
 	PullPolicy *string `json:"pullPolicy,omitempty"  desc:"image pull policy for the container"`
 	PullSecret *string `json:"pullSecret,omitempty" desc:"image pull policy for the container "`
 	Extended   *bool   `json:"extended,omitempty" desc:"if true, deploy an extended version of the container with additional debug tools"`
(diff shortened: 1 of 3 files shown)


install/helm/gloo/generate/values.go
```diff
@@ -54,6 +54,7 @@ type Image struct {
 	PullPolicy *string `json:"pullPolicy,omitempty"  desc:"image pull policy for the container"`
 	PullSecret *string `json:"pullSecret,omitempty" desc:"image pull policy for the container "`
 	Extended   *bool   `json:"extended,omitempty" desc:"if true, deploy an extended version of the container with additional debug tools"`
+	Fips       *bool   `json:"fips,omitempty" desc:"if true, deploy a version of the container built with FIPS-compliant crypto libraries"`
```
sam-heilbron, line 57: Should we clarify that this will only apply to both enterprise and certain deployments within enterprise?

kdorosh (approved):

install/helm/gloo/generate/values.go
kdorosh: seems like something we could put in a more complete guide that also details to pros / cons of fips (e.g., explains why we wouldn't always just install fips-compliant images)

PR #5285, opened 2021-09-01
Add RegexAction to user facing settings

elcasteel:
# Description
My previous change updated our internal envoy API, but I missed updating the user config.


projects/gloo/api/v1/enterprise/options/dlp/dlp.proto
```diff
@@ -194,4 +196,8 @@ message CustomAction {
     // default value: 75%
     // rounds ratio (percent/100) by std::round http://www.cplusplus.com/reference/cmath/round/
     .solo.io.envoy.type.Percent percent = 4;
+    // List of regexes to apply to the response body to match data which should be
+    // masked. They will be applied iteratively in the order which they are
+    // specified.
+    repeated .envoy.config.filter.http.transformation_ee.v2.RegexAction regex_actions = 5;
```
kdorosh, line 202: what happens if i provide regexes and regex_actions? we should tell the user, and deprecate regexes

kdorosh (approved):

Issue #5173, opened 2021-08-16
Webhook Validation for all Gloo resources (Upstreams)


sam-heilbron:
**Subtask of larger feature to support webhook validation for all gloo resources: https://github.com/solo-io/gloo/issues/2797**

## What is a Validation Webhook
Validation Webhooks are one type of Admission webhook, which are part of Kubernetes [dynamic admission control](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/). They are invoked by Kubernetes prior to persisting an object in ETCD. Below is an illustration of the order of steps taken before persisting an object in storage:

![1_gv_Da-8VDFo5PPPjmCFHjQ](https://user-images.githubusercontent.com/5279627/129572084-3d9e9666-ed54-4cf6-ac09-74e97c844477.jpeg)

## Does Gloo Have Webhook Support
Gloo already supports webhook validation for Gateway resources. 

It is defined here: https://github.com/solo-io/gloo/tree/master/projects/gateway/pkg/services/k8sadmisssion

And most of the validation logic lives here: https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/validation/validator.go

## Why is Webhook Validation necessary for Gloo resources?
It is possible for an Upstream to have valid syntax (fields have the proper types) but invalid semantics (the collective definition for that upstream is invalid). For example, if an Upstream defines an [ssl config](https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/upstream.proto#L52) with a [secretRef](https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/ssl.proto#L60) but the secret does not exist, the Upstream should be considered invalid. 

Introducing webhook validation on Gloo resources will both:
- Improve the user experience of applying Gloo configuration, since semantically incorrect resources can be denied admission (rejected at write time) rather than be accepted and then reported on. 
- Reduce likelihood of control plane crashes. Invalid resources lead to edge cases that the control plane needs to protect against. 

## Parent Issues
This work to introduce webhook validation for Upstreams is part of a larger feature to support webhook validation for all Gloo resources: https://github.com/solo-io/gloo/issues/2797

## Subtasks

- [x] validate single upstream
- [ ] validate list of upstreams
- [ ] validate upstream deletion

jenshu:
reopening to handle deletion and lists

PR #5301, opened 2021-09-07
Add upstream to gRPC validation service

MitchAman:
# Description

Add upstreams to gRPC validation service

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5173


artberger (commented):
A couple small doc suggestions, thanks!

projects/gloo/pkg/utils/validation/upstream_validation_test.go
```diff
@@ -0,0 +1,38 @@
...
+	. "github.com/onsi/gomega"
+	. "github.com/solo-io/gloo/projects/gloo/pkg/api/grpc/validation"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/utils/validation"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+)
+
+var _ = Describe("Upstream validation utils", func() {
+
+	Context("SSL config", func() {
```
kdorosh, line 14: can we have a regression test that validates that attempts to write an invalid upstream get rejected by the validation webhook?

kdorosh: e.g. https://github.com/solo-io/gloo/blob/3c4d38014ef65a7ea7b54562bb5a749f367c4232/test/kube2e/gateway/gateway_test.go#L1712

projects/gloo/pkg/validation/server.go
```diff
@@ -176,19 +177,29 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
 
-	logger.Infof("received proxy validation request")
-	xdsSnapshot, resourceReports, report, err := s.translator.Translate(params, req.GetProxy())
-	if err != nil {
-		logger.Errorw("failed to validate proxy", zap.Error(err))
-		return nil, err
+	response := validation.GlooValidationServiceResponse{}
+
+	if req.GetProxy() != nil {
+		logger.Infof("received proxy validation request")
```
kdorosh, line 183: instead of treating one or the other request, i think we should update the webhook to take an entire gloo api snapshot and validate that at once. e.g. the proxy might point to upstreams that were also added (think about the case where we are validating a list of resources in a manifest at once)

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -16,10 +17,12 @@ service GlooValidationService {
 
 message GlooValidationServiceRequest {
     gloo.solo.io.Proxy proxy = 1;
+    gloo.solo.io.Upstream upstream = 2;
```
kdorosh, line 20: we want to update this to match the gloo api snapshot, e.g. https://github.com/solo-io/gloo/blob/f48051cfd78fb7e2eb664e2401c7e1df4e14dd94/projects/gloo/pkg/api/v1/api_snapshot.sk.go#L19-L28

(i.e. make those `repeated`)

kdorosh (dismissed):
needs some changes to handle lists of resources in one validation request

projects/gloo/pkg/validation/server.go
kdorosh: this is what i mean by validating a list of resources https://github.com/solo-io/gloo/blob/517c44d2b960c46b9c5376b4a6fc130a5a764b77/projects/gateway/pkg/validation/validator.go#L63

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -23,11 +23,16 @@ webhooks:
       namespace: {{ .Release.Namespace }}
       path: "/validation"
     caBundle: "" # update manually or use certgen job or cert-manager's ca-injector
+  matchPolicy: Exact
```
MitchAman, line 26: Non-functional change: `Exact` is the default in [v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#validatingwebhook-v1beta1-admissionregistration-k8s-io), but in [v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#validatingwebhook-v1-admissionregistration-k8s-io) the default is `Equivalent`, just guarding against unintended behavior change should we bump the version

projects/gateway/pkg/services/k8sadmisssion/gateway-webhook-validation.drawio.xml
```diff
@@ -0,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jenshu, line 1: should this be here? is it a diagram?

jenshu: and the jpg file too. should they go in docs?

MitchAman: Yes, that's the "source code" of the diagram so it can be edited

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -16,10 +17,12 @@ service GlooValidationService {
 
 message GlooValidationServiceRequest {
     gloo.solo.io.Proxy proxy = 1;
```
jenshu, line 19: should this be a list of proxies as well?

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -16,10 +17,12 @@ service GlooValidationService {
...
+    repeated gloo.solo.io.Upstream upstreams = 2;
 }
 
 message GlooValidationServiceResponse {
     ProxyReport proxy_report = 1;
+    UpstreamReport upstream_report = 2;
```
jenshu, line 25: should these be lists of reports?

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -15,11 +21,28 @@ service GlooValidationService {
...
 }
 
 message GlooValidationServiceResponse {
+    repeated GlooValidationReport gloo_validation_reports = 1;
+}
+
+// A Gloo validation report represents the warnings/errors produced during a single translation loop of a proxy
+message GlooValidationReport {
     ProxyReport proxy_report = 1;
+    repeated ResourceReport upstream_reports = 2;
```
sam-heilbron, line 39: Does this need to be upstream specific? Will the gateway logic handling these ResourceReports be different for upstreams than for other gloo resources? Can we just make this `resource_reports` or something generic

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -38,7 +61,8 @@ message NotifyOnResyncResponse {
 * If the report contains no errors, the (sub-)resource is valid.
 */
 message ProxyReport {
-    repeated ListenerReport listener_reports = 1;
+    core.solo.io.ResourceRef proxy_ref = 1;
```
sam-heilbron, line 64: How is the proxy_ref used?

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,109 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	// update the snapshot with the given proxy
+	if req.GetProxy() != nil {
+		proxyRef := req.GetProxy().GetMetadata().Ref()
+		for i, proxy := range snap.Proxies {
+			if proxy.GetMetadata().Ref().Equal(proxyRef) {
+				// replace the existing proxy in the snapshot
+				snap.Proxies[i] = req.GetProxy()
+			}
+		}
+	}
```
sam-heilbron, line 221: Is this doing anything? Don't we pass in the proxy to the translate function, so if the req.GetProxy() is not nil, we pass that value? I don't think we do anythning with the snap.Proxies, so I assume we can remove this block altogether.

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,109 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	if req.GetUpstreams() != nil {
```
sam-heilbron, line 222: nit: I think we can remove the `if` clause, since ranging over the upstreams will be skipped if there are no upstreams

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,109 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+		}
+	}
+}
+
+func convertToGlooValidationReport(proxyReport *validation.ProxyReport, resourceReports reporter.ResourceReports) *validation.GlooValidationReport {
+	var upstreamReports []*validation.ResourceReport
+
+	for resource, report := range resourceReports {
+		switch resources.Kind(resource) {
+		case "*v1.Upstream":
```
sam-heilbron, line 248: Is there a reason we only return Upstream reports and not all warnings/errors?

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -15,11 +21,28 @@ service GlooValidationService {
...
+    repeated gloo.solo.io.Upstream upstreams = 2;
+    repeated core.solo.io.ResourceRef deleted_upstreams = 3;
 }
 
 message GlooValidationServiceResponse {
+    repeated GlooValidationReport gloo_validation_reports = 1;
+}
+
+// A Gloo validation report represents the warnings/errors produced during a single translation loop of a proxy
+message GlooValidationReport {
```
sam-heilbron, line 37: nit: Perhaps this is a ProxyValidationReport, since it contains details for a specific proxy

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,109 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	// update the snapshot copy with the resources from the request
+	applyRequestToSnapshot(&snapCopy, req)
+
 	ctx = contextutils.WithLogger(ctx, "proxy-validator")
 
 	params := plugins.Params{Ctx: ctx, Snapshot: &snapCopy}
 
 	logger := contextutils.LoggerFrom(ctx)
 
-	logger.Infof("received proxy validation request")
```
sam-heilbron, line 179: Can we keep the logging for received request, and validation report result? I think clients might be confused/concerned if those disappeared

projects/gateway/pkg/validation/validator.go
```diff
@@ -561,10 +592,94 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+	response, err := v.sendGlooValidationServiceRequest(ctx, &validation.GlooValidationServiceRequest{
+		// Sending a nil proxy causes the upstream to be translated with all proxies in gloo's snapshot
+		Proxy:     nil,
+		Upstreams: []*gloov1.Upstream{us},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(response.GetGlooValidationReports()) == 0 {
```
sam-heilbron, line 611: What case would cause this to happen?

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,109 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
 	snapCopy := s.latestSnapshot.Clone()
 	s.lock.RUnlock()
 
+	// update the snapshot copy with the resources from the request
+	applyRequestToSnapshot(&snapCopy, req)
```
sam-heilbron, line 175: 🎉

sam-heilbron (commented):
Looking good! A couple of questions

projects/gloo/api/grpc/validation/gloo_validation.proto
jenshu: it's not currently used; i'd be ok with removing this until we need it

projects/gloo/api/grpc/validation/gloo_validation.proto
jenshu: the issue is that each ResourceReport only has a ResourceRef and doesn't indicate what type of resource it is. If we have multiple resources with same name/namespace but different types, we wouldn't be able to differentiate them in the list of ResourceReports.  i don't know if we *need* to differentiate them, but seems like it would be useful to know if an error occurred on an upstream vs. some other type of resource

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -15,11 +21,28 @@ service GlooValidationService {
 }
 
 message GlooValidationServiceRequest {
+    // If a proxy is provided in the request, the response will contain only the report for that proxy.
+    // If no proxy is provided, the response will contain a report for each proxy in the Gloo API snapshot.
     gloo.solo.io.Proxy proxy = 1;
+
+    repeated gloo.solo.io.Upstream upstreams = 2;
+    repeated core.solo.io.ResourceRef deleted_upstreams = 3;
```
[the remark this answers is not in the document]
jenshu: that sounds good to me

projects/gloo/pkg/validation/server.go
jenshu: yeah, i wasn't sure if we need this section.  i know the proxy is passed separately, but also the whole snapshot is passed into Translate via the Params, so i wanted to make the snapshot up to date. looks like Translate is currently using the Upstreams and Endpoints from the snapshot, but i'm not certain whether the Proxies are being used somewhere

projects/gateway/pkg/validation/validator.go
jenshu: it shouldn't happen unless we make a coding error, but since we need to access the `[0]` element i wanted to be safe and do a check beforehand

projects/gloo/pkg/validation/server.go
jenshu: currently we only handle upstreams which go into the UpstreamReports field, but pending the discussion above about making this field more generic, we may be able to remove this check

projects/gloo/api/grpc/validation/gloo_validation.proto
jenshu: changed it to just ValidationReport

projects/gloo/api/grpc/validation/gloo_validation.proto
jenshu: removed

projects/gloo/api/grpc/validation/gloo_validation.proto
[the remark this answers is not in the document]
jenshu: removed

projects/gloo/pkg/validation/server.go
jenshu: removed

projects/gloo/api/grpc/validation/gloo_validation.proto
jenshu: spoke offline, will keep as is for now

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -350,7 +349,6 @@ func (wh *gatewayValidationWebhook) makeAdmissionResponse(ctx context.Context, r
 			AdmissionResponse: &v1beta1.AdmissionResponse{
 				Allowed: true,
 			},
-			Proxies: proxies,
```
jenshu, line 353: did we purposely remove this?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -372,7 +370,6 @@ func (wh *gatewayValidationWebhook) makeAdmissionResponse(ctx context.Context, r
 				Details: details,
 			},
 		},
-		Proxies: proxies,
```
jenshu, line 375: and this

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -120,8 +120,12 @@ func (s *proxyReconciler) addProxyValidationResults(ctx context.Context, proxies
 			logger.Warnw("Proxy had invalid config", zap.Any("proxy", proxy.GetMetadata().Ref()), zap.Error(validateErr))
 		}
 
+		if len(glooValidationResponse.GetValidationReports()) != 1 {
```
sam-heilbron, line 123: Can we add a comment outlining why there should/will only be a single ValidationReport in this instance?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -327,10 +327,9 @@ func (wh *gatewayValidationWebhook) makeAdmissionResponse(ctx context.Context, r
 		dryRun = *req.DryRun
 	}
 
-	proxyReports, validationErrs := wh.validate(ctx, gvk, ref, req.Object.Raw, isDelete, dryRun)
-	var proxies []*gloov1.Proxy
-	for proxy, _ := range proxyReports {
-		proxies = append(proxies, proxy)
+	reports, validationErrs := wh.validate(ctx, gvk, ref, req.Object.Raw, isDelete, dryRun)
+	if reports == nil {
```
sam-heilbron, line 331: What causes reports to be nil?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: +1

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -405,87 +401,112 @@ func (wh *gatewayValidationWebhook) validate(
...
+}
+
+func (wh *gatewayValidationWebhook) validateUpstream(ctx context.Context, rawJson []byte, dryRun bool, isDelete bool) (*validation.Reports, *multierror.Error) {
+	var (
+		us      gloov1.Upstream
+		reports *validation.Reports
+		err     error
+	)
+	if err := protoutils.UnmarshalResource(rawJson, &us); err != nil {
+		return nil, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
```
sam-heilbron, line 497: Do we treat unmarshal errors differently for upstreams?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -405,87 +401,112 @@ func (wh *gatewayValidationWebhook) validate(
...
+	}
+	if skipValidationCheck(us.GetMetadata().GetAnnotations()) {
+		return nil, nil
+	}
+	if isDelete {
```
sam-heilbron, line 502: Is this case possible? I thought we we removed all 'delete' functionality

projects/gateway/pkg/validation/validator.go
```diff
@@ -567,6 +596,102 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+			errs = multierr.Append(errs, errors.Wrapf(err, "failed to validate Proxy with Gloo validation server"))
+		}
+		if warnings := validationutils.GetProxyWarning(report.GetProxyReport()); !v.allowWarnings && len(warnings) > 0 {
+			for _, warning := range warnings {
+				errs = multierr.Append(errs, errors.New(warning))
+			}
+		}
+
+		// Append upstream errors
+		for _, usRpt := range report.GetUpstreamReports() {
```
sam-heilbron, line 629: Does this fail if any upstream reports an error? What if the upstream we're modifying is not related to the other upstream? Does the webhook reject the request?

projects/gateway/pkg/validation/validator.go
```diff
@@ -567,6 +596,102 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+		}
+		return &Reports{
+			ProxyReports:    &proxyReports,
+			UpstreamReports: &upstreamReports,
+		}, errors.Wrapf(errs, "validating %T %v", us, us.GetMetadata().Ref())
+	}
+
+	logger.Debugf("Accepted Upstream %v", us.GetMetadata().Ref())
+	if !dryRun {
+		utils2.MeasureOne(ctx, mValidConfig)
```
sam-heilbron, line 655: mValidConfig is specificically for gateway configuration. I don't think we need to add metrics for this just yet. I like the idea and think it would be useful, but I think we can do in a follow up

projects/gateway/pkg/validation/validator.go
```diff
@@ -253,12 +268,23 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
 			continue
 		}
 
-		proxyReports[proxy] = proxyReport.GetProxyReport()
-		if err := validationutils.GetProxyError(proxyReport.GetProxyReport()); err != nil {
+		if len(glooValidationResponse.GetValidationReports()) != 1 {
```
sam-heilbron, line 271: What can cause this situation?

MitchAman:
> LGTM for docs, thank you!

@artberger This PR has changed substantially since you gave your stamp of approval. Happy to address any changes if you want to take another once-over

projects/gateway/pkg/validation/validator.go
MitchAman: `len(glooValidationResponse.GetValidationReports())` is > 1 when no proxy is provided in the `GlooValidationServiceRequest` and there's multiple proxies in the Gloo API snapshot. This is just a safety net for calling `glooValidationResponse.GetValidationReports()[0]` below

projects/gateway/pkg/validation/validator.go
MitchAman: Whoops, overlooked that when copy/pasting. I thought it was for config in general, not just for gateway. Removed.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -28,6 +28,10 @@ webhooks:
     apiGroups: ["gateway.solo.io"]
     apiVersions: ["v1"]
     resources: ["*"]
+  - operations: [ "CREATE", "UPDATE" ]{{/* TODO(mitchaman): Implement deletes */}}
+    apiGroups: ["gloo.solo.io"]
+    apiVersions: ["v1"]
+    resources: ["upstreams"]{{/* TODO(https://github.com/solo-io/gloo/issues/2797): Extend to all gloo resources */}}
```
artberger, line 34: I just haven't seen this before, is it ok that the Helm yamls have `TODO`s left in them?

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/grpc/validation/gloo_validation.proto.sk.md
```diff
@@ -68,14 +72,55 @@ weight: 5
...
+"warnings": []string
+"errors": []string
+
+```
+
+| Field | Type | Description |
+| ----- | ---- | ----------- | 
+| `resourceRef` | [.core.solo.io.ResourceRef](../../../../../../../solo-kit/api/v1/ref.proto.sk/#resourceref) |  |
+| `warnings` | `[]string` |  |
+| `errors` | `[]string` |  |
```
artberger, line 123: Should these have descriptions? If so, maybe something that explains why these matter, like:

```suggestion
| `resourceRef` | [.core.solo.io.ResourceRef](../../../../../../../solo-kit/api/v1/ref.proto.sk/#resourceref) |  A way to refer to resources in different namespaces by including the name and namespace of the resource in this `resourceRef`.
- `|
| `warnings` | `[]string` |  The warnings that are returned for the resource. Warnings do not necessarily prevent an operation from happening, but might require action.  |
| `errors` | `[]string` | The errors that are returned for the resource. You cannot modify the resource until the errors are resolved. For help troubleshooting, see the [Debug guide]({{< versioned_link_path fromRoot="/operations/debugging_gloo/" >}}). |
```

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/grpc/validation/gloo_validation.proto.sk.md
```diff
@@ -68,14 +72,55 @@ weight: 5
...
+| `validationReports` | [[]gloo.solo.io.ValidationReport](../gloo_validation.proto.sk/#validationreport) |  |
+
+
+
+
+---
+### ValidationReport
+
+ 
+A validation report represents the warnings/errors produced during a single translation loop of a proxy
```
artberger, line 91: If it succeeds, is anything returned? or does this only work if there _are_ warnings and errors? Maybe add that.
```suggestion
A validation report represents the warnings and errors that produced during a single translation loop of a proxy. If no warnings or errors are produced, the report is empty.
```

artberger (commented):
@MitchAman left a couple questions, thanks!

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
MitchAman: When a helm chart is generated from this template the comments are removed. This section renders to: 

```yaml
    rules:
    - apiGroups:
      - gateway.solo.io
      apiVersions:
      - v1
      operations:
      - CREATE
      - UPDATE
      - DELETE
      resources:
      - '*'
      scope: '*'
    - apiGroups:
      - gloo.solo.io
      apiVersions:
      - v1
      operations:
      - CREATE
      - UPDATE
      resources:
      - upstreams
      scope: '*'
```

artberger: ok cool, thank you!

projects/gateway/pkg/validation/validator.go
sam-heilbron: validateSnapshot is only called for proxy validation (ie gateway resources), correct? So we should never hit the case where there are more than 1 validation report returned. I like the defensive programming, but I'm trying to understand how the glooValidationResponse can return a non-error and an empty proxy, which is what this is protecting against, right?

test/kube2e/gateway/gateway_test.go
```diff
@@ -1782,6 +1783,19 @@ spec:
...
+apiVersion: gloo.solo.io/v1
+kind: Upstream
+metadata:
+  name: invalid-upstream
+  namespace: gloo-system
+spec:
+  static:
+    hosts:
+      - addr: ~
+`, expectedErr: "The Upstream \"invalid-upstream\" is invalid: spec.static.hosts.addr: Invalid value: \"null\": spec.static.hosts.addr in body must be of type string: \"null\"",
```
sam-heilbron, line 1797: Is this error caught by gloo translation or schema validation?

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,99 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	case *multierror.Error:
+		var errorStrings []string
+		for _, e := range err.(*multierror.Error).Errors {
+			errorStrings = append(errorStrings, e.Error())
+		}
+		return errorStrings
+	}
+	if err == nil {
+		return []string{}
+	}
```
jenshu, line 265: can we move this before the `switch`?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
MitchAman: Purposely removed it, accidentally didn't add it back. All set now.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
MitchAman: We do not

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
MitchAman: It's not possible right now, because the webhook config does not match against delete operations for upstreams: https://github.com/solo-io/gloo/blob/a1a58a0d4ddcf611e5f84cf6e56fe27124baf9cf/install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml#L31

projects/gloo/pkg/validation/server.go
MitchAman: Good call, done.

test/kube2e/gateway/gateway_test.go
MitchAman: This is caught by translation: https://github.com/solo-io/gloo/blob/61f15014f66266dbfd4b01d3bfaa9ea2b19402ec/projects/gloo/pkg/plugins/static/plugin.go#L74

`~` in yaml is the empty string

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -54,9 +54,9 @@ var _ = Describe("Validator", func() {
 				snap := samples.GatewaySnapshotWithDelegates(us.Metadata.Ref(), ns)
 				err := v.Sync(context.TODO(), snap)
 				Expect(err).NotTo(HaveOccurred())
-				proxyReports, err := v.ValidateRouteTable(context.TODO(), snap.RouteTables[0], false)
+				reports, err := v.ValidateRouteTable(context.TODO(), snap.RouteTables[0], false)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(proxyReports).To(HaveLen(1))
+				Expect(*(reports.ProxyReports)).To(HaveLen(1))
```
MitchAman, line 59: Add these assertions back in

projects/gateway/pkg/validation/validator.go
MitchAman: Yes, that's the behavior in this PR.

Are there any ways in which multiple upstreams can cause eachother to error? Like duplicate names or something?  I'm not sure if we should ignore translation errors from non-modified upstreams or not. If something is failing translation, it's indicative that whatever changed in the snapshot caused that error.

projects/gateway/pkg/validation/validator.go
MitchAman: > validateSnapshot is only called for proxy validation (ie gateway resources), correct?

Correct.

> So we should never hit the case where there are more than 1 validation report returned.

That's correct, it *shouldn't* ever happen. I'm happy to remove this if you have strong feelings against it. It just feels wrong to me to access an array without confirming its size

projects/gateway/pkg/validation/validator.go
```diff
@@ -567,6 +611,105 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+		}
+
+		// Append proxies
+		if report.GetProxy() != nil {
+			proxies = append(proxies, report.GetProxy())
+		}
+	}
+
+	if errs != nil {
+		logger.Debugf("Rejected Upstream %v: %v", us.GetMetadata().Ref(), errs)
```
sam-heilbron, line 664: remove since we return an error?

projects/gateway/pkg/validation/validator.go
```diff
@@ -302,10 +340,13 @@ func (v *validator) ValidateList(ctx context.Context, ul *unstructured.Unstructu
 		var itemProxyReports, err = v.processItem(ctx, item)
 
 		errs = multierror.Append(errs, err)
-		for proxy, report := range itemProxyReports {
+		for _, report := range *itemProxyReports.ProxyReports {
```
sam-heilbron, line 343: if processItem returns an error, will `*itemProxyReports` be a NPE?

projects/gateway/pkg/validation/validator.go
```diff
@@ -233,11 +254,16 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
 			continue
 		}
 
+		proxies = append(proxies, proxy)
+
 		// validate the proxy with gloo
-		var proxyReport *validation.GlooValidationServiceResponse
+		var glooValidationResponse *validation.GlooValidationServiceResponse
 		err := retry.Do(func() error {
```
sam-heilbron, line 261: you add a nice utility method that does the same thing, can we re-use that here?

projects/gateway/pkg/validation/validator.go
```diff
@@ -207,7 +227,8 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
 
 	var (
 		errs         error
-		proxyReports ProxyReports = map[*gloov1.Proxy]*validation.ProxyReport{}
+		proxyReports ProxyReports
```
sam-heilbron, line 230: This used to be a mapping of proxy -> proxy Report, which I assume made it easier to track report errors back to a particular proxy. Is there a concern that using lists removes this functionality?

projects/gateway/pkg/validation/validator.go
```diff
@@ -32,7 +32,21 @@ import (
 	"go.uber.org/zap"
 )
 
-type ProxyReports map[*gloov1.Proxy]*validation.ProxyReport
+type Reports struct {
+	Proxies         []*gloov1.Proxy
+	ProxyReports    *ProxyReports
+	UpstreamReports *UpstreamReports
```
sam-heilbron, line 38: An UpstreamReport and ProxyReport is for a particular proxy. If we have a list of proxies in this message, are the UpstreamReports merged for all proxies in that list?

projects/gateway/pkg/validation/validator.go
```diff
@@ -567,6 +611,105 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+	logger.Debugf("Got response from GlooValidationService: %s", response.String())
+
+	var (
+		errs            error
+		proxyReports    ProxyReports
+		upstreamReports UpstreamReports
+		proxies         []*gloov1.Proxy
+	)
+	for _, report := range response.GetValidationReports() {
+		// Append proxy errors
```
sam-heilbron, line 633: For this logic, we are validating that when provided with an upstream to change, that gloo can accept that upstream. Why are we unable to check if the provided upstream is in the response validation reports?

projects/gateway/pkg/validation/validator.go
```diff
@@ -60,12 +78,14 @@ var _ Validator = &validator{}
...
-	ValidateGateway(ctx context.Context, gw *v1.Gateway, dryRun bool) (ProxyReports, error)
-	ValidateVirtualService(ctx context.Context, vs *v1.VirtualService, dryRun bool) (ProxyReports, error)
+	ValidateList(ctx context.Context, ul *unstructured.UnstructuredList, dryRun bool) (*Reports, *multierror.Error)
+	ValidateGateway(ctx context.Context, gw *v1.Gateway, dryRun bool) (*Reports, error)
+	ValidateVirtualService(ctx context.Context, vs *v1.VirtualService, dryRun bool) (*Reports, error)
 	ValidateDeleteVirtualService(ctx context.Context, vs *core.ResourceRef, dryRun bool) error
-	ValidateRouteTable(ctx context.Context, rt *v1.RouteTable, dryRun bool) (ProxyReports, error)
+	ValidateRouteTable(ctx context.Context, rt *v1.RouteTable, dryRun bool) (*Reports, error)
 	ValidateDeleteRouteTable(ctx context.Context, rt *core.ResourceRef, dryRun bool) error
+	ValidateUpstream(ctx context.Context, us *gloov1.Upstream, dryRun bool) (*Reports, error)
```
sam-heilbron, line 87: how is dryRun used for upstreams?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -254,6 +253,7 @@ func (wh *gatewayValidationWebhook) ServeHTTP(w http.ResponseWriter, r *http.Req
 			admissionReview.Response.UID = review.Request.UID
 		}
 	}
+	logger.Debugf("Returning AdmissionResponse: %s", admissionReview.Response)
```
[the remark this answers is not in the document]
MitchAman: The admission review contains the Admission Response, so I'd say there's not a whole lot of value in this extra log line. I'll remove

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,100 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	return &validation.GlooValidationServiceResponse{
+		ValidationReports: validationReports,
+	}, nil
+}
 
-	logger.Infof("proxy validation report result: %v", report.String())
-	return &validation.GlooValidationServiceResponse{ProxyReport: report}, nil
+// updates the given snapshot with the resources from the request
+func applyRequestToSnapshot(snap *v1.ApiSnapshot, req *validation.GlooValidationServiceRequest) {
+	for _, us := range req.GetUpstreams() {
```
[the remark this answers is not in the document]
MitchAman: Added this in a `// TODO` for when we do lists

projects/gloo/pkg/validation/server.go
```diff
@@ -170,25 +171,100 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+			if existingUs.GetMetadata().Ref().Equal(usRef) {
+				// replace the existing upstream in the snapshot
+				snap.Upstreams[i] = us
+				isUpdate = true
+				break
+			}
+		}
+		if !isUpdate {
+			snap.Upstreams = append(snap.Upstreams, us)
+			snap.Upstreams.Sort()
```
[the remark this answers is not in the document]
MitchAman: Yup, good call

projects/gateway/pkg/validation/validator.go
MitchAman: From what I can tell, we never actually cared about this mapping.  Ultimately the `ProxyReports` would get returned back to `makeAdmissionResponse()`, which only cares about the `Proxy` keys, ignoring the `ProxyReport` values: https://github.com/solo-io/gloo/blob/b25062e48273db70f7ddbcb018db3f3f07377b6e/projects/gateway/pkg/services/k8sadmisssion/validating_admission_webhook.go#L330-L334

This is because the errors from within the `ProxyReport` objects have already been merged into the `validationErrs` multierr

projects/gateway/pkg/validation/validator.go
jenshu: processItem eventually calls through to validateSnapshot, which can return any combination of `nil, err`, `nil, nil`, and `reports, nil`.  so we should probably be safe and check that itemProxyReports is not nil

projects/gateway/pkg/validation/validator.go
```diff
@@ -567,6 +600,103 @@ func (v *validator) validateGatewayInternal(ctx context.Context, gw *v1.Gateway,
...
+		if warnings := validationutils.GetProxyWarning(report.GetProxyReport()); !v.allowWarnings && len(warnings) > 0 {
+			for _, warning := range warnings {
+				errs = multierr.Append(errs, errors.New(warning))
+			}
+		}
+
+		// Append upstream errors
+		for _, usRpt := range report.GetUpstreamReports() {
+			upstreamReports = append(upstreamReports, usRpt)
+			if err := resourceReportToMultiErr(usRpt); err != nil {
```
jenshu, line 636: hmm, we are converting the upstream errors from a multierr into a string array here: https://github.com/solo-io/gloo/pull/5301/files#diff-2ab04dbc4a0dcf23297b16ea7ec39f37d76eff5eaa33fe90db0ead0b88c7b022R245
only to convert them back into a multierr here. should we just change the field to a multierror? (assuming we can use multierrors in the .proto file)

test/kube2e/gateway/gateway_test.go
```diff
@@ -1707,6 +1707,7 @@ var _ = Describe("Kube2e: gateway", func() {
 	Context("tests for the validation server", func() {
 		testValidation := func(yaml, expectedErr string) {
 			out, err := install.KubectlApplyOut([]byte(yaml))
+			fmt.Printf("out: %v, err: %v", out, err)
```
jenshu, line 1710: do we want to remove this?

projects/gateway/pkg/validation/validator.go
jenshu: can you clarify "unable to check if the provided upstream is in the response validation reports"?

MitchAman: Are you implying we should only care about errors for the upstream that's being validated (the one in the request)?

MitchAman: > Are you implying we should only care about errors for the upstream that's being validated (the one in the request)?

Yes

projects/gateway/pkg/validation/validator.go
MitchAman: I don't think we're able to use multierrors in a proto

projects/gateway/pkg/validation/validator.go
MitchAman: Per discussion we will address dryRun race condition in a separate PR

projects/gateway/pkg/validation/validator.go
```diff
@@ -233,32 +254,34 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
...
-		)
+		glooValidationResponse, err := v.sendGlooValidationServiceRequest(ctx, &validation.GlooValidationServiceRequest{
+			Proxy: proxy,
+		})
 		if err != nil {
 			err = errors.Wrapf(err, "failed to communicate with Gloo validation server")
-			if v.ignoreProxyValidationFailure {
-				contextutils.LoggerFrom(ctx).Error(err)
-			} else {
+			if !v.ignoreProxyValidationFailure {
```
jenshu, line 265: should we keep this here, since user still might want the logs?

projects/gateway/pkg/validation/validator.go
MitchAman: Done.

projects/gateway/pkg/validation/validator.go
MitchAman: Yes, I had thought the calling function logged on error, but doesn't look like that's the case

jenshu (approved):

jenshu (approved):

test/kube2e/gateway/gateway_test.go
MitchAman: @jenshu

Issue #3160, opened 2020-06-15
Enrich event object for AWS Lambda integration


lgadban:
**Is your feature request related to a problem? Please describe.**
The `event` object that an AWS Lambda receives when executing is the main way for the function to get contextual data from the incoming request. Currently, the Gloo integration only provides the request body as the received `event`.

**Describe the solution you'd like**
The `event` should additional information from the request (e.g. headers, query string, etc.).
For example:
[`event` format when Lambda called via ALB](https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html)
[`event` format when Lambda called via AWS API Gateway](https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html)

**Describe alternatives you've considered**
Transformations can get you most of the way, but become extremely cumbersome/verbose as the you try to reach parity with e.g. AWS API Gateway

via-jordan-sokolic:
+1 for this. It would make transitioning from AWS API Gateway to Gloo super straightforward.

Something similar to how Kong handles AWS Lambda integration would be ideal IMO with feature flag `config.is_proxy_integration`: https://docs.konghq.com/hub/kong-inc/aws-lambda/ 
The AWS docs specify the [format](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format) for integrating with their API gateway. 

From our testing it seems this is partially implemented in [`aws.responseTransformation`](https://docs.solo.io/gloo/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/aws/aws.proto.sk/), but that only handles the body and ignores the rest of the response (`statusCode`, `headers`, etc.).

For now, we're using the following transformation for `statusCode`:
```yaml
      transformations:
        responseTransformation:
          transformationTemplate:
            headers:
              ":status": 
                text: '{% if default(statusCode, "") != "" %}{{ statusCode }}{% else %}{{ header(":status") }}{% endif %}'
```

RichiCoder1:
We're also looking for something pretty close to the AWS API Gateway event payload for invoked lambdas.

ben-taussig-solo:
[...]

Issue #5206, opened 2021-08-20
aws lambda: golang contains no event payload


lilley2412:
**Describe the bug**

Golang lambdas are invoked with no event payload, I cannot find a way to access headers, querystring, or request body.

**To Reproduce**

* Create a basic lambda:
```go
package main

import (
	"context"
	"fmt"
	"os"
	"strings"

	"github.com/aws/aws-lambda-go/lambda"
	"github.com/aws/aws-lambda-go/lambdacontext"
	"github.com/davecgh/go-spew/spew"
)

func main() {
	lambda.Start(func(ctx context.Context, event interface{}) (map[string]string, error) {
		spew.Dump(os.Environ())
		lc, _ := lambdacontext.FromContext(ctx)
		ctxDump := strings.ReplaceAll(spew.Sdump(lc), "\n", "\\n")
		eventDump := strings.ReplaceAll(spew.Sdump(event), "\n", "\\n")
		fmt.Println(ctxDump)
		fmt.Println(eventDump)
		return map[string]string{
			"event": eventDump,
			"ctx":   ctxDump,
		}, nil
	})
}
```
* Create a vs / upstream pointing to the lambda
* Invoke the lambda

**Expected behavior**

The api response and cloudwatch logs should show the event type and values from`spew.Sdump(event)`.  The event type is `map[string]interface{}`, which seems correct, but the value is empty.  

I also checked to see if anything is being stuffed into context, but the spew.Dump() of the lambdacontext shows that clientContext.Env and clientConext.Custom are empty map[string]string.  

**Additional context**
- Gloo edge 1.8.1
- Kubernetes 1.21
- I don't think this matters - but I'm using IAM roles for service accounts and not static aws keys for permissions (the lambda does get invoked and the response is written correctly)

jameshbarton:
Internal Reference: 
https://solo-io.slack.com/archives/G0133U4GUP5/p1629470023063200

chrisgaun:
@ben-taussig-solo moving this to close. LMK if that is mistake

ben-taussig-solo:
@chrisgaun sounds good! Detailed update below:

[...]

PR #5308, opened 2021-09-09
Aws Lambda Request Transform

ben-taussig-solo:
# Description
  - Adds the `headerBodyRequestTransform` property to AWS destinationSpec, allowing users to easily forward incoming request headers to an AWS Lambda upstream.
  - By default, the payload received by an AWS Lambda upstream just contains the incoming request body
  - With the `headerBodyRequestTransform` property set to `true`, the payload received by the upstream will contain the incoming request body as well as the headers on the incoming request
  - Diagram of the described behavior: diagram
  - This transform plays nicely with the `ResponseTransform` property on the AWS destinationSpec, which parses the lambda response as JSON and sets the response body to the value of the top-level field `"body"` 
  - New tests interact with the `dumpContext` Lambda in the solo-io AWS organization. The source of the lambda is as follows (python)
  ```python
import json

def main(event, context):
    print(f'Logging event: {event}')
    print(f'Logging context: {context}')
    return {
        'statusCode': 200,
        'eventDump': json.dumps(event),
        'body': json.dumps(event.get('body')),
        'headers': json.dumps(event.get('headers'))
    }
  ```

# Context
- This PR addresses the following 2 issues related to the event payload received by AWS lambdas:
  -  [aws lambda: golang contains no event payload #5206](https://github.com/solo-io/gloo/issues/5206)
  -  [Enrich event object for AWS Lambda integration #3160](https://github.com/solo-io/gloo/issues/3160)
  - In short, users are frustrated by the limited amount of data that is sent to an AWS Lambda Upstream
    - Incoming request bodies are forwarded to Lambda Upstreams as the event payload
    - Users want additional information in the Lambda event payload
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3160


projects/gloo/api/v1/options/aws/aws.proto
```diff
@@ -62,6 +62,9 @@ message DestinationSpec {
         SYNC = 0;
         ASYNC = 1;
     }
+    // Include headers in the event sent to aws lambda
+    bool header_body_request_transformation = 6;
```
kdorosh, line 66: can we improve the transformations here to additionally grab the status code, request path, and any query params too? akin to https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html

ben-taussig-solo: Update on this:
  - Request path and query params are available in the current implementation as part of the `:path` header.
    - Unfortunately, in this implementation it is up to the upstream to parse these values
  - If we use a transformation template instead of a header-body transformation, we can send separate request path and querystring values to the upstream
    - We do not have the ability to isolate unspecified query-params (as in the payload at the link you posted) without making changes to the translation plugin
    - If we use a transformation template, request headers which are not explicitly extracted in the transformation will not be included in the payload received by the lambda.
  - Status code is not available at the relevant stage in the request path
  - I think the best path forward is to continue with the current implementation. In the future, I think it would be valuable to add the ability to extract all headers to transformation templates, as we had discussed previously.

ben-taussig-solo: Update: 
The lambda payload created by the current implementation looks like the following:
```json
{
  "body": "\"solo.io\"",
  "headers": {
    ":authority": "localhost:11082",
    ":method": "POST",
    ":path": "/1?param_a=value_a&param_b=value_b",
    ":scheme": "http",
    "accept-encoding": "gzip",
    "content-length": "9",
    "content-type": "application/octet-stream",
    "user-agent": "Go-http-client/1.1",
    "x-forwarded-proto": "http",
    "x-request-id": "4973ce46-9918-4dcf-8ba0-685a5b8899ad"
  },
  "httpMethod": "POST",
  "path": "/1",
  "querystring": "param_a=value_a&param_b=value_b"
}
```

Whereas the lambda payload created by AWS API gateway is as follows:
```json
{
    "requestContext": {
        "elb": {
            "targetGroupArn": "arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/lambda-279XGJDqGZ5rsrHC2Fjr/49e9d65c45c6791a"
        }
    },
    "httpMethod": "GET",
    "path": "/lambda",
    "queryStringParameters": {
        "query": "1234ABCD"
    },
    "headers": {
        "accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8",
        "accept-encoding": "gzip",
        "accept-language": "en-US,en;q=0.9",
        "connection": "keep-alive",
        "host": "lambda-alb-123578498.us-east-2.elb.amazonaws.com",
        "upgrade-insecure-requests": "1",
        "user-agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36",
        "x-amzn-trace-id": "Root=1-5c536348-3d683b8b04734faae651f476",
        "x-forwarded-for": "72.12.164.125",
        "x-forwarded-port": "80",
        "x-forwarded-proto": "http",
        "x-imforwards": "20"
    },
    "body": "",
    "isBase64Encoded": false
}
```

ben-taussig-solo: The differences are as follows:
 - Our payload lacks the `isBase64Encoded` field
 - Our payload lacks the `RequestContext` field, which seems to contain information about which load balancer the request originally passed through
 - Our payload does not separate querystring parameters. Instead, the querystring is left to the user to parse

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -264,7 +325,24 @@ func (p *plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
...
 		return nil, err
 	}
-	return []plugins.StagedHttpFilter{
+
+	filters := []plugins.StagedHttpFilter{
 		f,
-	}, nil
+	}
+
+	if *p.transformsAdded {
```
ben-taussig-solo, line 333: I added the following block to resolve the issue Kevin helped me debug this morning
Early transformations defined in the aws plugin were not being executed because we were not applying a transformation filter at the early stage

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -264,7 +325,24 @@ func (p *plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
...
+		earlyPluginStage := plugins.AfterStage(plugins.FaultStage)
+		earlyStageConfig := &envoy_transform.FilterTransformations{
+			Stage: transformation.EarlyStageNumber,
+		}
+
+		tf, err := plugins.NewStagedFilterWithConfig("io.solo.transformation", earlyStageConfig, earlyPluginStage)
```
sam-heilbron, line 339: nit: let's use the constant that defines this value: `transformation.FilterName`

projects/gloo/pkg/plugins/aws/plugin.go
sam-heilbron: What is the case that we need to add this? Is it just when we are using the headerBody transformation? If that's the case, I think we need the logic to be slightly more specific. `transformsAdded` becomes true when either a request or response transformation is applied.

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -264,7 +325,24 @@ func (p *plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
...
 	}
-	return []plugins.StagedHttpFilter{
+
+	filters := []plugins.StagedHttpFilter{
 		f,
-	}, nil
+	}
+
+	if *p.transformsAdded {
+		earlyPluginStage := plugins.AfterStage(plugins.FaultStage)
```
sam-heilbron, line 334: I see this code elsewhere in our transformation plugin. How did we decide on 'after fault stage' as an early plugin stage? 
If this setting needs to be consistent for all filters that use early staged transformations, we could have a utility function earlyTransformationFilter() that returns a transformation filter for the early stage

test/e2e/README.md
```diff
@@ -25,13 +25,25 @@ Example:
 ENVOY_IMAGE_TAG=solo-test-image TEST_PKG=./test/e2e/... make run-tests
 ```
 
+### Example Debug Workflow: Use `WAIT_ON_FAIL` to Dump Envoy Config
```
sam-heilbron, line 28: Thanks for writing this up!

projects/gloo/pkg/plugins/aws/plugin.go
ben-taussig-solo: Replaced this with `if if p.requestTransformsAdded {` in order to be more specific.

projects/gloo/pkg/plugins/aws/plugin.go
ben-taussig-solo: This is a transformation-specific concept -- for now I've exported EarlyPluginStage in plugins/transformation/plugin.go

test/e2e/README.md
```diff
@@ -25,13 +25,25 @@ Example:
...
+The `WAIT_ON_FAIL` environment variable can be used to inspect Gloo/Envoy state during an e2e test. Instructions to do so are as follows:
+- Run your e2e test with `WAIT_ON_FAIL=1`, in order to prevent the Gloo installation from being torn down on failure
+  - It is important that your test fails -- teardown will occur as normal if the test suite runs successfully
+- When failure occurs, inspect running Docker containers using `docker ps`
+  - You should see a container which matches the following criteria:
+  - 
+    |IMAGE|PORTS|NAMES|
+    |-----|-----|-----|
+    |quay.io/solo-io/gloo-envoy-wrapper:<ENVOY_IMAGE_TAG>|0.0.0.0:11082-11083->11082-11083/tcp, :::11082-11083->11082-11083/tcp, 0.0.0.0:21001->21001/tcp, :::21001->21001/tcp|e2e_envoy|
```
sam-heilbron, line 37: This is great. I think it could be valuable to point our where this 21001 port comes from. 

we default the adminPort to 20000 (https://github.com/solo-io/gloo/blob/master/test/services/envoy.go#L36)
and when we create a new instance we add some digits to ensure it can be run in parallel (https://github.com/solo-io/gloo/blob/master/test/services/envoy.go#L401)

I believe our current way of running tests only runs a single envoy at a time, so this will always be 1 instance more than the default case, which is why it is port 21001.

ben-taussig-solo: Added this to the e2e readme :)

projects/gloo/api/v1/options/aws/aws.proto
kdorosh, line 66: we should still update the name of the user-facing API and comment here

ben-taussig-solo: Done here: https://github.com/solo-io/gloo/pull/5308/commits/00fd94b94b2bb4d52d429b53da7a0cbd6fae5fd2

projects/gloo/pkg/plugins/aws/plugin.go
ben-taussig-solo: Update: Resolved this issue by exporting the transformation plugin's `RequireEarlyTransformation` property, passing it into the AWS plugin during registry, and setting the property to `true` when the early transformations are applied.

```go
aws.NewPlugin(&transformationPlugin.RequireTransformationFilter, &transformationPlugin.RequireEarlyTransformation),
```

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -215,28 +217,89 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
...
+											MergeExtractorsToBody: &envoy_transform.MergeExtractorsToBody{},
+										},
+									},
 								},
 							},
 						},
 					},
-				},
+				})
+				*p.earlyTransformsAdded = true
```
sam-heilbron, line 271: I think this needs to set both earlyTransforms and regularTransforms.

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -215,28 +217,89 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
...
+							RequestTransformation: &envoy_transform.Transformation{
+								TransformationType: &envoy_transform.Transformation_TransformationTemplate{
+									TransformationTemplate: &envoy_transform.TransformationTemplate{
+										Extractors: map[string]*envoy_transform.Extraction{
+											"path": {
+												Source:   &envoy_transform.Extraction_Header{Header: ":path"},
+												Regex:    `([^\?]+)(\?.*)?`,
+												Subgroup: uint32(1),
+											},
+											"querystring": {
```
kdorosh, line 251: the aws lamda proxy integration (and httpMethod below) use camel case, i think we should do the same here

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -215,28 +217,89 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
...
+												Text: "text/html",
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				})
+				*p.transformsAdded = true
```
kdorosh, line 298: is this even used by anything? we should go through the git history and figure out it's original intended purpose, see if we currently have a bug on master.

if not, we should delete this if it has always been unused

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -44,18 +44,20 @@ func getLambdaHostname(s *aws.UpstreamSpec) string {
 	return fmt.Sprintf("lambda.%s.amazonaws.com", s.GetRegion())
 }
 
-func NewPlugin(transformsAdded *bool) plugins.Plugin {
+func NewPlugin(transformsAdded, earlyTransformsAdded *bool) plugins.Plugin {
 	return &plugin{
-		transformsAdded: transformsAdded,
+		transformsAdded:      transformsAdded,
+		earlyTransformsAdded: earlyTransformsAdded,
```
kdorosh, line 50: what is this read by? if it's not used for anything we should remove it

projects/gloo/pkg/plugins/aws/plugin.go
kdorosh: i think we either have a bug or should remove it

test/e2e/README.md
```diff
@@ -45,3 +59,21 @@ chmod +x echoapp2
 sudo ./echoapp2 --port 80 &
 ```
 
+### Notes on AWS Lambda Tests (`test/e2e/aws_test.go`)
```
kdorosh, line 62: 👏

test/e2e/aws_test.go
```diff
@@ -237,6 +237,111 @@ var _ = Describe("AWS Lambda", func() {
 		validateLambda(1, defaults.HttpPort, `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>`)
 	}
 
+	testProxyWithRequestTransform := func() {
+		err := envoyInstance.RunWithRoleAndRestXds(services.DefaultProxyName, testClients.GlooPort, testClients.RestXdsPort)
```
kdorosh, line 241: do we need to run with rest xds? we default to grpc eds now, probably should update this to use grpc as well as it's more indicative of default installs. also the envoy logs will be cleaner and easier to grok on failure

projects/gloo/pkg/plugins/aws/plugin.go
ben-taussig-solo: These are set to `&transformationPlugin.RequireTransformationFilter` and `&transformationPlugin.RequireEarlyTransformation` during plugin registry:
https://github.com/solo-io/gloo/blob/5ffc060f4af715911bee8b311208785c457d45a2/projects/gloo/pkg/plugins/registry/registry.go#L61

`transformationPlugin.RequireEarlyTransformation` determines whether early-stage transformations are applied:
https://github.com/solo-io/gloo/blob/5f2dce76a76eb8fb1b38449de0fec4364884a256/projects/gloo/pkg/plugins/transformation/plugin.go#L161-L166

I think at one point `&transformationPlugin.RequireTransformationFilter` had a similar functionality, but I am not confident that it does anything currently. It is referenced in a number of other plugins, so I'm not sure if I feel comfortable removing it as part of this PR

test/e2e/aws_test.go
```diff
@@ -237,6 +237,111 @@ var _ = Describe("AWS Lambda", func() {
 		validateLambda(1, defaults.HttpPort, `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>`)
 	}
 
+	testProxyWithRequestTransform := func() {
```
kdorosh, line 240: can we improve the testing assertions to include some expected values for headers and queryString to prove that the regex is grabbing the fields as expected?

projects/gloo/pkg/plugins/aws/plugin.go
ben-taussig-solo: I think at one point it controlled whether the regular-stage transformation got applied. A quick review of the transformation plugin seems to suggest that this is no longer the case, but I'd need to do further testing to be confident of that. The property in the transformation plugin that `transformsAdded` corresponds to is controlled by a number of different plugins, so I don't feel comfortable removing it at the current moment. Additional discussion in this comment: https://github.com/solo-io/gloo/pull/5308#discussion_r707387876

projects/gloo/pkg/plugins/aws/plugin.go
kdorosh: the plugin registry is read at startup of the event loop for gloo, i would expect the plugin to determine the need for using the early transformation (or not) while configuring the routes rather than once up front for all aws upstreams

kdorosh (approved):
very nice!! 🚀

Issue #4790, opened 2021-05-27
Gloo discovery creates invalid ownerReferences in upstream


andylibrian:
**Describe the bug**
When Gloo discovery creates upstreams from services, it also copies the `metadata.ownerReferences` of the service to the upstream. This becomes a problem when it discovers services from different namespaces, resulting in the upstream having ownerReference from different namespace.

Kubernetes Garbage Collector would delete the upstreams due to having ownerReferences in different namespace.

ref:
- https://github.com/kubernetes/kubernetes/issues/65200
- https://github.com/kubernetes/kubernetes/pull/92743

**To Reproduce**
Steps to reproduce the behavior (tested on k8s 1.19):

1. Create a service with ownerReferences

   ```
   kubectl create deployment nginx --image=nginx --port=80 -n default
   kubectl expose deployment/nginx --port=80
   
   kubectl edit svc nginx ; add ownerReferences
   metadata:
     ownerReferences:
     - apiVersion: apps/v1
       blockOwnerDeletion: true
       controller: true
       kind: Deployment
       name: nginx
       uid: uid_of_deployment/nginx
   ```

2. Install gloo through helm chart with additional watch namespaces:

   ```
   kubectl create ns gloo-system
   helm upgrade -i gloo gloo/gloo --timeout 15m --set "settings.watchNamespaces={gloo-system,default}" -n gloo-system
   ```

2. See the upstream created with ownerReferences copied from the service, which is invalid / cross-namespace

**Expected behavior**
ownerReferences shouldn't be copied to different namespaces.

**Additional context**
- This tool https://github.com/kubernetes-sigs/kubectl-check-ownerreferences also reports this as an error (child namespace does not match owner namespace)
   ```
   GROUP          RESOURCE    NAMESPACE     NAME               OWNER_UID                              LEVEL   MESSAGE
   gloo.solo.io   upstreams   gloo-system   default-nginx-80   df502980-70a1-47a8-946e-b6fabe6580e5   Error   child namespace does not match owner namespace (default)

   1 error, 0 warnings
   ```

- "Cross-namespace owner references are disallowed by design." source: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

- In my real world use-case, I use flagger with gloo for canary / progressive deployment. Flagger creates 2 services: canary and primary, with ownerReferences to the canary object (CRD). Gloo discovers these services and creates upstreams with cross-namespace ownerReferences.

saiskee:
[...]

PR #5340, opened 2021-09-16
remove ownerReference mapping for UDS created upstreams

kevin-shelaga:
# Description

Remove ownerReference mapping for UDS created upstreams

# Context

A growing number of customers have run into this issue to where K8s 1.20+ GC removes discovery created upstream because we copy over the services ownerReferences.

This is the 2nd half of the fix. 1st part: https://github.com/solo-io/solo-kit/pull/446

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/4790


sam-heilbron (changes requested):
Thanks for doing this!

I see we're keeping ownerReferences for all resources, other than discovered services. Is there a reason why we need ownerReferences for most of these other resources? I had assumed this was most useful for objects that are dependent on other objects (like a ReplicaSet). If they are valuable, should we be even more specific and only exclude the owner ref of discovered upstream if the service namespace doesn't match the write namespace.

kevin-shelaga:
@sam-heilbron I think there is a strong argument to almost never copy over ownerReferences. In most circumstances they will probably not be needed by gloo. That being said, the change was made in such a way that we can easily stop the copying of the ownerReferences for resources as required. 

Right now we know that UDS created upstreams are an issue when we copy over the ownerReferences. Arguably you could check every resources to see if its cross-namespaced, and if it is then just dont copy over the ownerReference.

Given how late in the release it is i opted for the smallest impact of "dont fix what isnt broken" and only make the change to UDS.

This was just my reasoning for the current change but, I will make whatever change you think is best.

sam-heilbron (approved):
Spoke offline:

The work required to only ignore ownerRef's for services discovered outside the writeNamespace is a little more involved, and since the ownerRefs aren't useful for any Upstreams, we opted to keep the change simple and just ignore ownerRefs for all discovered upstreams.

LGTM

PR #5347, opened 2021-09-17
Fail loudly when docker push fails

elcasteel:
# Description

Use `${PIPESTATUS[0]}` to get the result of `make docker-push` so that we can fail builds when docker push fails.

# Context

Solo-projects releases were silently failing to push images to quay. Preemptively porting the same fix here.


elcasteel:
I don't remember why I created 2 PRs but closing the duplicate.

PR #5348, opened 2021-09-17
Fail release loudly

elcasteel:
# Description

Use `PIPESTATUS` to get the results of `make docker-push`.

# Context

Releases were silently failing to push builds to quay on solo-projects. This change preemptively ports the same fix to gloo.


Makefile
```diff
@@ -613,7 +613,7 @@ docker-build: discovery-docker gateway-docker gloo-docker \
 # docker-push is intended to be run by CI
 .PHONY: docker-push
 docker-push: $(DOCKER_IMAGES)
-ifeq ($(CREATE_ASSETS), "true")
+ifeq ($(RELEASE), "true")
```
sam-heilbron, line 616: How does CREATE_ASSETS differ from RELEASE? Can we remove CREATE_ASSETS entirely?

sam-heilbron (commented):
LGTM, one question about duplicate env variables

Makefile
elcasteel: CREATE_ASSETS is also used for the cloudbuilds we run as part of CI. It is true when TEST_ASSET_ID is set: 
https://github.com/solo-io/gloo/blob/master/Makefile#L22
https://github.com/solo-io/gloo/blob/master/cloudbuild.yaml#L122

elcasteel:
The gloo change also needs this change: https://github.com/solo-io/solo-projects/pull/2742/files however it winds up being implemented.

Makefile
elcasteel: I switched this back to `if $(CREATE_ASSETS)` and the cloudbuild worked so I probably misinterpreted the error message when this failed before.

bewebi (approved):

PR #5352, opened 2021-09-20
Support Namespaced Statuses

sam-heilbron:
Blocked by:
- [x] Solo-kit supports namespaced statuses (https://github.com/solo-io/solo-kit/pull/447)
- [x] Solo-apis supports namespaced statuses for rate limit config (https://github.com/solo-io/solo-apis/pull/229)
- [x] Solo-projects WIP PR maintly passing CI (https://github.com/solo-io/solo-projects/pull/2712)

# Description
Add support for having statuses from more than one controller.

# Context
This PR adds a new proto message, NamespacedStatuses, that acts as a replacement for Status. NamespacedStatuses is a mapping from namespace to Status. This allows multiple gloo instances in different namespaces to record their own statuses without overwriting one another.

*There are a lot of files changed in this PR, below are some notable types of changes *

# Notable Changes

### Proto Files and Generated Files
The proto message for each gloo resource was modified to remove (set as reserved) the deprecated status field, and introduce a new namespaced_statuses field.

This change causes the generated markdown files to change, as well as the CRDs

### Setup Syncer 
We determine the namespace of the controller at setup time. All of the setup_syncers have been modified to create a `StatusClient` a simple wrapper responsible for getting and setting a status on a resource, so the controller doesn't need to track which namespace to write to (or what the shape of the data on the resource is). 

To decide on the namespace, we first try to read the `POD_NAMESPACE` env variable. This will work for k8s installs, since this value is exposed via the downward API to the container running the controlller. However, non-k8s installs will not have this value, so we fallback to the `writeNamespace` value provided in settings. If that is not provided, we fallback to `gloo-system`. 

### Gloo CLI
The CLI was modified to print statuses in a useful way. Now that a resource has a map of statuses, we introduced a simple aggregator, so that we consistently aggregate resource statuses.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5018


docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/external/envoy/api/v2/cluster/outlier_detection.proto.sk.md
```diff
@@ -30,11 +30,11 @@ See the `architecture overview (arch_overview_outlier_detection)` for
 more information on outlier detection.
 
 ```yaml
-"consecutive5xx": .google.protobuf.UInt32Value
+"consecutive5Xx": .google.protobuf.UInt32Value
 "interval": .google.protobuf.Duration
 "baseEjectionTime": .google.protobuf.Duration
 "maxEjectionPercent": .google.protobuf.UInt32Value
-"enforcingConsecutive5xx": .google.protobuf.UInt32Value
+"enforcingConsecutive5Xx": .google.protobuf.UInt32Value
```
kdorosh, line 37: hmm this is the second time i've seen this strange diff, still not sure why we're getting this. you?

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -190,7 +197,7 @@ func stripInvalidListenersAndVirtualHosts(ctx context.Context, proxiesToWrite Ge
 // stripping invalid virtual hosts / listeners from the desired proxy,
 // else we will wind up with both an invalid and valid version of the same listener/vhost on our proxy
 // which is invalid and will be rejected by Envoy
-func transitionFunc(proxiesToWrite GeneratedProxies) gloov1.TransitionProxyFunc {
+func transitionFunc(proxiesToWrite GeneratedProxies, defaultTransitionFunc gloov1.TransitionProxyFunc) gloov1.TransitionProxyFunc {
```
kdorosh, line 200: does this ever get called with another func? `defaultTransitionFunc` seems like an odd name if not

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
```diff
@@ -152,7 +157,15 @@ var _ = Describe("ReconcileGatewayProxies", func() {
...
 				Expect(err).NotTo(HaveOccurred())
-				liveProxy.Status.State = core.Status_Accepted
+
+				liveProxyStatus := statusClient.GetStatus(liveProxy)
+				if liveProxyStatus == nil {
+					liveProxyStatus = &core.Status{State: core.Status_Accepted, ReportedBy: "gateway"}
+					statusClient.SetStatus(liveProxy, liveProxyStatus)
+				} else {
+					liveProxyStatus.State = core.Status_Accepted
+				}
```
kdorosh, line 167: not sure i understand this change. is this working around a race?

projects/gateway/pkg/syncer/translator_syncer_integration_test.go
```diff
@@ -133,37 +145,41 @@ var _ = Describe("TranslatorSyncer integration test", func() {
...
 			}
-			proxyState := subresouce["*v1.Proxy.gloo-system.gateway-proxy"]
+			proxyState := subresource["*v1.Proxy.gloo-system.gateway-proxy"]
 			if proxyState == nil {
 				return core.Status_Pending, fmt.Errorf("no state")
 			}
-			return proxyState.State, nil
-		})
+			return proxyState.GetState(), nil
+		}, "60s")
```
kdorosh, line 158: this was probably good for debugging but we should lower it for ci

projects/gateway/pkg/syncer/translator_syncer_integration_test.go
```diff
@@ -133,37 +145,41 @@ var _ = Describe("TranslatorSyncer integration test", func() {
...
+		statusClient.SetStatus(proxy, &core.Status{State: core.Status_Accepted})
+
 		_, err = proxyClient.Write(proxy, clients.WriteOpts{OverwriteExisting: true})
-		Expect(err).NotTo(HaveOccurred())
+		ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	}
 
-	It("should set status correctly even when the status from the snapshot was not updated", func() {
-
+	FIt("should set status correctly even when the status from the snapshot was not updated", func() {
```
kdorosh, line 182: unfocus

projects/gloo/cli/pkg/cmd/root.go
```diff
@@ -52,6 +52,14 @@ func App(opts *options.Options, preRunFuncs []PreRunFunc, optionsFunc ...cliutil
 			}
 			return nil
 		},
+		PersistentPostRunE: func(cmd *cobra.Command, args []string) error {
```
kdorosh, line 55: what is this change?

projects/gloo/cli/pkg/prerun/prerun_output.go
```diff
@@ -26,3 +29,12 @@ func SetKubeConfigEnv(opts *options.Options, cmd *cobra.Command) error {
...
 }
+
+func SetPodNamsepaceEnv(opts *options.Options, cmd *cobra.Command) error {
+	podNamepace := defaults.GlooSystem
+	if opts.Metadata.GetNamespace() != "" {
+		podNamepace = opts.Metadata.GetNamespace()
+	}
+
+	return os.Setenv(statusutils.PodNamespaceEnvName, podNamepace)
+}
```
kdorosh, line 40: should this be its own option? sounds like we default to the namespace of the resource we want to get, rather than allowing users to get the status of a resource from another controller

e.g. `glooctl get vs -n my-ns` will only make the status written by the `my-ns` controller readable. perhaps it's easier to just list them all? or add another flag?

projects/gloo/cli/pkg/cmd/root.go
```diff
@@ -118,14 +126,17 @@ func GlooCli() *cobra.Command {
 		)
 	}
 
-	preRunFuncs := []PreRunFunc{
+	preRunFuncs := []RunnableCommand{
 		// should make sure to read the config file first
 		ReadConfigFile,
 		prerun.SetKubeConfigEnv,
+		prerun.SetPodNamsepaceEnv,
```
kdorosh, line 133: we should add comments why this is necessary

projects/gloo/cli/pkg/printers/upstream.go
```diff
@@ -55,6 +56,25 @@ func UpstreamTable(xdsDump *xdsinspection.XdsDump, upstreams []*v1.Upstream, w i
...
 }
 
+func upstreamStatus(us *v1.Upstream) string {
+	namespacedStatuses := us.GetNamespacedStatuses()
+
+	// If there are no statuses defined for this resource, default to a pending status for the resource namespace
+	if namespacedStatuses == nil {
+		namespacedStatuses = &core.NamespacedStatuses{
+			Statuses: map[string]*core.Status{
+				us.GetMetadata().GetNamespace(): {
```
kdorosh, line 66: why are we choosing this namespace? why are we putting a fake pending status on the upstream at all while printing out?

projects/gloo/cli/pkg/printers/virtualservice.go
```diff
@@ -135,9 +135,28 @@ func getRouteTableStatus(vs *v1.RouteTable) string {
 }
 
 func getStatus(ctx context.Context, res resources.InputResource, namespace string) string {
+	namespacedStatuses := res.GetNamespacedStatuses()
+
+	// If there are no statuses defined for this resource, default to a pending status for the resource namespace
+	if namespacedStatuses == nil {
+		namespacedStatuses = &core.NamespacedStatuses{
+			Statuses: map[string]*core.Status{
+				namespace: {
```
kdorosh, line 144: same thought here as https://github.com/solo-io/gloo/pull/5352/files#r713218425

kdorosh (commented):
lgtm overall! just a couple questions / nits

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
sam-heilbron: no good call. I inherited this change without thinking much about it. Updated to mirror previous behavior

projects/gateway/pkg/syncer/translator_syncer_integration_test.go
sam-heilbron: whoops good call!

projects/gloo/cli/pkg/prerun/prerun_output.go
sam-heilbron: yeah good call. i don't actually think this is necessary. at the moment, all statuses are checked (or printed). I agree there could be use in exposing a field to say, "just get me the status from this ns", but I'm not sure it's necessary now. I've removed this set, and going to verify that the cli still behaves appropriately

projects/gloo/cli/pkg/cmd/root.go
sam-heilbron: yeah good call, this may be unnecessary altogether. looking into that now

projects/gloo/cli/pkg/printers/upstream.go
sam-heilbron: I updated this to make it cleare. The namespace wasn't actually being used. This was just to ensure that the printer output a `pending` status for any resource missing a status. This was the previous default behavior.

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/external/envoy/api/v2/cluster/outlier_detection.proto.sk.md
sam-heilbron: Yeah I want to investigate this. Looking now

projects/gateway/pkg/reconciler/proxy_reconciler.go
sam-heilbron: good call, I can change

projects/gloo/cli/pkg/prerun/prerun_output.go
sam-heilbron: So I updated the code to make it clearer (I think). We just need to explicitly set the nil status to pending, since previously (for single statuses) that was the behavior. If we don't explicitly print 'pending' for nil statuses, and empty string will be returned

projects/gloo/cli/pkg/cmd/root.go
sam-heilbron: Updated with a comment for why it's necessary

kdorosh (approved):

PR #5356, opened 2021-09-20
Add CI check for DO_NOT_SUBMIT comments in go files.

MitchAman:
# Description

Add a CI check which fails when `// DO_NOT_SUBMIT` is found in *.go files that were modified in a PR.

Example failure: https://github.com/solo-io/gloo/runs/3654089442?check_suite_focus=true

# Checklist:

### DIFF
--- a/changelog/v1.9.0-beta19/do-not-submit.yaml
+++ b/changelog/v1.9.0-beta19/do-not-submit.yaml
@@ -1,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Add a CI check which fails when `// DO_NOT_SUBMIT` is found in *.go files that were modified in a PR.
--- a/.github/workflows/do-not-submit.yaml
+++ b/.github/workflows/do-not-submit.yaml
@@ -1,0 +1,19 @@
+name: CI
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    name: Test changed-files
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v1.1.2
+      - name: Check for DO_NOT_SUBMIT
+        run: ./ci/do-not-submit.sh "${{ steps.changed-files.outputs.all_modified_files }}"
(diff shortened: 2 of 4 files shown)


changelog/v1.9.0-beta19/do-not-submit.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Add a CI check which fails when `// DO_NOT_SUBMIT` is found in *.go files that were modified in a PR.
```
kdorosh, line 3: what files do we generate have this header today?

MitchAman: What do you mean by that? No files contain "DO_NOT_SUBMIT" at the moment (besides those introduced in this PR, but they aren't go files

sam-heilbron (commented):
I believe the goal of this is to avoid merging PRs until they are 'ready'. How does this GHA differ from the behavior of the 'work in progress' label or creating a PR as a draft?

MitchAman:
> I believe the goal of this is to avoid merging PRs until they are 'ready'. How does this GHA differ from the behavior of the 'work in progress' label or creating a PR as a draft?

The behavior is largely the same as the 'work in progress' label, though `// DO_NOT_SUBMIT` can be added before a PR is opened while it's still fresh in your mind, rather than having to remember to apply the 'work in progress' label.

One scenario (although not actually supported yet because the file extension isn't *.go) where this would be useful is within a gloo PR which depends on a solo-kit PR that's still in review.  Imagine that within gloo's `go.mod` the solo-kit version is set to the latest commit of the solo-kit PR. We don't want the gloo PR to get merged until the solo-kit PR is merged and released, so that we can update the specific commit reference to the newly released version.  Dropping the following comment in `go.mod`:

1. makes it clear what still needs to be done in the PR
2. can be applied immediately when changing the code (rather than after PR is opened)
3. provides a layer of protection if one forgets to apply 'work in progress' and/or a reviewer doesn't catch it
```
// DO_NOT_SUBMIT: Update to new solo-kit version once it's released
github.com/solo-io/solo-kit v0.22.1-0.20210920203028-060e333b9a2d
```
In this scenario, we want the gloo PR to be reviewed - everything is ready to go, it's just blocked by the solo-kit change, so a draft PR feels inappropriate. The 'work in progress' label could be used to protect against merging in this scenario, but one needs to remember to apply it. The 'work in progress' label also does not provide any context around what part of the PR is still being worked on, while the comment makes it very clear.

sam-heilbron (approved):

PR #5367, opened 2021-09-22
Fix the tags for -extended gcr images

elcasteel:
# Description

The -extended part of the tag was missing when we retagged images from quay to gcr.


changelog/v1.9.0-beta20/gcr-push-fix.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+ - type: FIX
```
kdorosh, line 2: non user facing, since this is a ci change

PR #5404, opened 2021-09-29
Use generalized scan CLI

mlholland:
Changes the `run-security-scans` makefile command to use the generalized CLI that's in the works.


mlholland:
This is just a demo pr. If we actually want to replace existing functionality with the go-utils cli, we should open an issue.

PR #5529, opened 2021-11-02
Update graphql docs

Rachael-Graham:
Fix docs from https://www.apollographql.com/docs/intro/benefits/#graphql-provides-declarative-efficient-data-fetching

### DIFF
--- a/docs/content/guides/traffic_management/request_processing/graphql/_index.md
+++ b/docs/content/guides/traffic_management/request_processing/graphql/_index.md
@@ -5,111 +5,104 @@ description: Enables graphql resolution
 ---
 
 {{% notice note %}}
-Available in Gloo Edge Enterprise as of v1.10.0-beta1.
+This feature is available only in Gloo Edge Enterprise version 1.10.0-beta1 and later.
 {{% /notice %}}
 
 ## Why GraphQL?
-GraphQL is a server-side query language and runtime you can use to expose your APIs. As an alternative to REST APIs.
-GraphQL is powerful because it allows you to request only the data you want, and handle any following requests on
-the server-side, saving potentially numerous expensive origin to client requests and instead handling them in your
+GraphQL is a server-side query language and runtime you can use to expose your APIs as an alternative to REST APIs.
+GraphQL allows you to request only the data you want and handle any subsequent requests on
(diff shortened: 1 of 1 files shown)


docs/content/guides/traffic_management/request_processing/graphql/_index.md
```diff
@@ -5,111 +5,104 @@ description: Enables graphql resolution
...
-And get the following response:
-```json
-{"data":{"pets":[{"name":"Dog"},{"name":"Cat"}]}}
-```
-
-Remember that the REST request returned to our graphql server was the following json:
+1. Create a virtual service that defines a `Route` with a `graphqlSchemaRef` as the
+destination. In this example, all traffic to `/graphql` is handled by the GraphQL server in the Envoy proxy.
+   {{< highlight yaml "hl_lines=12-16" >}}
+   cat << EOF | kubectl apply -f -
```
kdorosh, line 44: is this the style we've committed to across all our docs? I feel like `cat << EOF` on all of our yaml is a bit overkill since we can simply copy/paste then apply with `pbpaste | k apply -f -` without `kubectl` on all of our yaml

kdorosh (commented):
just one question, the proposed changes lgtm

kdorosh (approved):
thanks @Rachael-Graham !

docs/content/guides/traffic_management/request_processing/graphql/_index.md
```diff
@@ -5,111 +5,104 @@ description: Enables graphql resolution
...
+               typedProvider:
+                 value: 'GET'
+             ':path':
+               typedProvider:
+                 value: '/api/pets'
+         upstreamRef:
+           name: default-petstore-8080
+           namespace: gloo-system
+     schema: "schema { query: Query } type Query { pets: [Pet] } type Pet { name: String }"
+   EOF
```
kdorosh, line 90: we need to be careful when writing and updating docs that have copy / paste code paths. The indentation here is wrong, thus the command no longer is accepted. To see the fix, look at https://github.com/solo-io/gloo/pull/5535 . Happy to walk through how we can validate and test these docs changes if we're unsure if this will be accepted.

Issue #4367, opened 2021-03-01
Tech Debt: Consolidate Upgradeable Plugins and Plugin interface


kdorosh:
# Problem

Easy to forget to make plugin upgradeable when adding new plugins.

# Context

we should consolidate https://github.com/solo-io/gloo/blob/f5ea636e8d72e7270b307dd3c19d2b47672020e9/projects/gloo/pkg/plugins/plugin_interface.go#L22 and https://github.com/solo-io/gloo/blob/f5ea636e8d72e7270b307dd3c19d2b47672020e9/projects/gloo/pkg/plugins/plugin_interface.go#L29 if possible

upgradeable plugins were added to improve UX when using enterprise features in open source (admittedly a bit leaky abstraction as well, would like to improve)

PR #5536, opened 2021-11-03
Add Graphql Client to Discovery and Extension point for enterprise discovery.

saiskee:
## Description

Adding an extension point to Gloo Edge discovery so that we can add enterprise discoveries, such as graphql schema discovery from scraped openapi specs.


projects/discovery/pkg/fds/discoveries/registry/registry_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func TestPlugins(t *testing.T) {
+	opts := bootstrap.Opts{}
+	plugins := Plugins(opts)
+	pluginTypes := make(map[reflect.Type]int)
+	for index, plugin := range plugins {
+		pluginType := reflect.TypeOf(plugin)
+		pluginTypes[pluginType] = index
+	}
+	if len(plugins) > len(pluginTypes) {
```
kdorosh, line 18: shouldn't this be inverted?

projects/discovery/pkg/fds/interface.go
```diff
@@ -11,6 +12,18 @@ import (
 
 type UpstreamMutator func(*v1.Upstream) error
 
+// Upgradable function discovery factories are those which can be
```
kdorosh, line 15: i was never a huge fan of the upgradeable pattern of plugins. we're leaking enterprise implementation details into open source. we should allow users to call extensions points (what we're calling `plugins` here) and construct the discovery event loop with extra logic just by fulfilling the interface

e.g. https://github.com/solo-io/gloo/blob/1ac616e99cdcf8f99c313d4703bfd5813edf89be/projects/gloo/pkg/syncer/setup/setup_syncer.go#L70-L81

kdorosh: i see we followed this pattern later in the PR, but here we don't have upgradeable implementations (the new functionality is enterprise-only) so we should be able to remove the upgradeable interfaces

projects/discovery/pkg/fds/discoveries/registry/registry_test.go
saiskee: we are checking for `multiple plugins with same type`. This is true if there are more plugins than types. Taken from 
https://github.com/solo-io/gloo/blob/68f091e58125110221cbfe8a2009a6a4c5f83b23/projects/gloo/pkg/plugins/registry/registry_test.go#L12

projects/discovery/pkg/fds/syncer/setup_syncer.go
```diff
@@ -107,6 +121,44 @@ func RunFDS(opts bootstrap.Opts) error {
...
+		pe := discoveryFactoryExtension()
+		if uPlug, ok := pe.(fds.Upgradable); ok && uPlug.IsUpgrade() {
+			upgradedDiscoveryFactories[uPlug.FunctionDiscoveryFactoryName()] = true
+		}
+		discFactories = append(discFactories, pe)
+	}
+	return reconcileUpgradedDiscoveryFactories(discFactories, upgradedDiscoveryFactories)
+}
+
+func reconcileUpgradedDiscoveryFactories(factoryList []fds.FunctionDiscoveryFactory, upgradedFactories map[string]bool) []fds.FunctionDiscoveryFactory {
```
kdorosh, line 142: do we really need this? i think we should. not a great model to follow, see https://github.com/solo-io/gloo/issues/4367

projects/discovery/pkg/fds/discoveries/registry/registry_test.go
kdorosh: ah i had it inverted yes. perhaps easier to read and no less correct to use `!=`?

kdorosh (commented):
you probably meant to delete `envoy_.gstmp`?

install/helm/gloo/templates/9-gateway-proxy-configmap.yaml
```diff
@@ -140,6 +140,21 @@ data:
                       - name: envoy.filters.http.router
 {{- end}} {{- /* if $spec.readConfig */}}
       clusters:
+# cluster referenced by graphql plugin but is unused as an actual cluster
```
kdorosh, line 143: does this mean we can close https://github.com/solo-io/gloo/pull/5568 now?

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
```diff
@@ -35,6 +36,9 @@ message ValueProvider {
     // Optional: fetches the value in the argument selected at this key.
     // If the key is invalid, returns the zero-value primitive or null.
     repeated PathSegment path = 2;
+    // If this is set to true, then a schema error will be returned to user when the graphql arg is not found.
+    // Defaults to false, so schema error will not be returned to user when the graphql arg is not found.
```
kdorosh, line 40: this api also applies to graphql parents and other complex objects with nesting, not just graphql arg. can we clarify that here?

saiskee: I've only implemented `required` for graphql arg. Do you mean that I should implement this with graphql parent?

projects/discovery/pkg/fds/interface.go
saiskee: Removed Upgradable pattern from this PR.

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
kdorosh: let's call out that this isn't supported (yet) with graphql parent then

projects/gloo/pkg/translator/route_config.go
```diff
@@ -293,11 +293,12 @@ func (h *httpRouteConfigurationTranslator) setAction(
 		}
 
 	case *v1.Route_GraphqlSchemaRef:
-		out.Action = &envoy_config_route_v3.Route_DirectResponse{
-			DirectResponse: &envoy_config_route_v3.DirectResponseAction{
-				Status: 200, // unused
-				Body: &envoy_config_core_v3.DataSource{
-					Specifier: &envoy_config_core_v3.DataSource_InlineString{InlineString: "unused, graphql overtakes router filter"},
+		// Envoy needs the route to have an action, so we use a dummy cluster here
+		// But this cluster doesn't really exist.
```
kdorosh, line 297: add note about CORS filter needing a route hence no direct response action?

kdorosh (commented):
lgtm, pending ci

kdorosh (approved):

PR #5568, opened 2021-11-15
Use Dummy cluster for graphql routing

saiskee:
# Description
 Due to an upstream envoy issue where the cors filter is disabled on direct response action routes, we are switching the way in which   we configure graphql routes. Previously they used a direct response action, but now they will use a graphql "dummy" cluster. This cluster is never used, similar to the direct response action, but is simply a destination which is required for the route configuration. The graphql filter will handle routing, so this is not an important change but merely one needed to make CORS work.

## Notes
- We are adding a dummy cluster in the envoy config bootstrap by default in this change. Not sure if this is ideal or if we should feature flag this dummy cluster and add it as part of an initial CDS update.


install/helm/gloo/templates/9-gateway-proxy-configmap.yaml
```diff
@@ -140,6 +140,21 @@ data:
                       - name: envoy.filters.http.router
 {{- end}} {{- /* if $spec.readConfig */}}
       clusters:
+# cluster referenced by graphql plugin but is unused as an actual cluster
+      - name: graphql.dummy.cluster
+        load_assignment:
+          cluster_name: graphql.dummy.cluster
+          endpoints:
+          - lb_endpoints:
```
yuval-k, line 148: do we need endpoints here? what happens if we have 0 endpoints?

saiskee: No endpoints are needed, tested.

PR #5571, opened 2021-11-15
Hg hybridlistenerreport list

bewebi:
use a list instead of a map for the `matched_listener_reports` field of the `HybridListenerReport` proto


sam-heilbron (approved):
LGTM

PR #5574, opened 2021-11-15
Introduce ListenerTranslator and RouteConfigurationTranslator

sam-heilbron:
# Description

Refactor some code to make it easier to implement a new HybridGateway type, where multiple FilterChains can be explicitly configured by users.

# Context

We are in the process of extending the API to enable a single Gloo Listener to configure multiple filter chains. To enable that, our Gateway resource (that is translated into a Gloo Listener) will support a TcpGateway (it already does), HttpGateway (it alreayd does) and a new HybridGateway (not yet added), which will be a list of Tcp or Http gateways. 

Our translation loop currently makes the assumption that a Gloo Listener will have a single instance of either a TcpGateway or an HttpGateway. To support the new API, we will need to change all the code that currently make that assumption.

This is Part 1 of this change. The goal was to make a non-functional change, keeping our existing tests, and introduce an abstraction (https://martinfowler.com/bliki/BranchByAbstraction.html) that we can implement for our new Gateway type. 

### What's the approach?
This attempts to follow the Dependency Inversion Principle:
```
High-level modules should not depend on low-level modules. Both should depend on abstractions.
Abstractions should not depend on details. Details should depend on abstractions.
```

In our code, we handle TcpLIsteners and HttpListeners in most of the same code, and each function handles the branching logic itself. This makes it challenging to track which code is actually responsible for generating TcpListeners and which is used for HttpListeners. To enable this, we make our translation loop depend on abstractions and instantiate separate code paths for http and tcp support.

An HttpListener needs:
- A way to create its RouteConfiguration
- A way to create an HttpFilterChain

A TcpListener needs:
- A way to create a TcpFilterChain

### This is a big PR, how do we know it is safe?
All changes are contained within the translatorInstance (https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/translator/translator.go#L86) and we have unit tests for this instance. We've changed the implementation of that translator (introducing sub-translators which have more focused responsibilities). If all the existing tests pass, I'm comfortable we didn't break anything.

### What code changed?
Largely this PR copied/pasted code. By creating actual instances of translators, the main differences are just that some instances that were previously passed as arguments in functions are now struct variables. 

There we a couple of small changes: 
1. We used to call `routeConfigName()` deep down in the translation loop. Now I do it at constructor time and pass in the value, ensuring that the value is the same for the RouteConfig and Listener.
2. We used to determine if the VirutalHosts required Tls by checking the SslConfiguration lenght. Similar to above, we inject the dependencies instead of handling that logic in the translation loop. We rely on the same function. The purpose of this change is that when we support a HybridGateway, the determination for whether or not to require Tls on VirtualHosts will be different.

### What comes next?
This first pass tries to create a clear abstraction that the new gateway type could implement. Next, I'd like to write tests around the sub-translators. Then, I'd like to change the interface of plugins so that instead of receiving a top-level listener, they receive only the parameters they really need (either a TcpListener or HttpListener). This way, we can run the translation loop N times, for N instances of either a TcpListener or HttpListener on our new hybrid gateway type.

# Checklist:


sam-heilbron:
/kick data race: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=c9b34878-ba1c-4879-be47-8d6ba86ce4e4

projects/gloo/pkg/translator/listener_subsystem.go
```diff
@@ -0,0 +1,91 @@
...
+	sslConfigTranslator utils.SslConfigTranslator,
+) *ListenerSubsystemTranslatorFactory {
+	return &ListenerSubsystemTranslatorFactory{
+		plugins:             plugins,
+		proxy:               proxy,
+		sslConfigTranslator: sslConfigTranslator,
+	}
+}
+
+func (l *ListenerSubsystemTranslatorFactory) GetListenerSubsystemTranslators(listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
```
nfuden, line 36: Question: How would we feel about GetTranslators as this is already an instance of ListenerSubststemTranslatorFactory?
```suggestion
func (l *ListenerSubsystemTranslatorFactory) GetTranslators(listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
```

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -0,0 +1,247 @@
...
+			mergedSslSecrets[key] = ptrToCopy
+			result = append(result, ptrToCopy)
+		}
+	}
+
+	return result
+}
+
+func merge(values []string, newvalues ...string) []string {
+	existing := map[string]bool{}
```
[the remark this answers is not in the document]
sam-heilbron: I like it. I had just copied this code and avoided making changes to reduce the number of changes in this PR. Happy to make this change in a follow up if that's ok.

projects/gloo/pkg/translator/listener.go
```diff
@@ -3,68 +3,55 @@ package translator
...
-				validation.AppendListenerError(listenerReport,
-					validationapi.ListenerReport_Error_ProcessingError,
-					err.Error())
-				continue
-			}
-			filterChains = append(filterChains, result...)
-		}
-	}
+type ListenerTranslator interface {
+	// A single Gloo Listener produces a single Envoy listener
```
kdorosh, line 17: this will no longer be true with the hybrid listener

kdorosh (commented):
lgtm overall

projects/gloo/pkg/translator/route_config.go
```diff
@@ -33,71 +33,76 @@ var (
...
-	routeCfgName string,
-	listenerReport *validationapi.ListenerReport,
-) *envoy_config_route_v3.RouteConfiguration {
-	if listener.GetHttpListener() == nil {
-		return nil
-	}
+type RouteConfigurationTranslator interface {
+	// A Gloo listener may produce multiple filter chains. Each one may contain its own route configuration
+	// https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_routing#arch-overview-http-routing
+	ComputeRouteConfiguration(params plugins.Params) []*envoy_config_route_v3.RouteConfiguration
```
bewebi, line 39: Why do we return a slice here _and not_ in `ComputeListener()`?
Both had returned single pointers previously and both will need to return slices once HybridListeners are in play, so they should remain consistent now one way or the other

sam-heilbron: I tried to call this out in the comment, but perhaps I wasn't clear enough. A single Gloo Listener only produces a single Envoy listener. Within that listner, there will be multiple filter chains. Therefore, a HybridGateway still only translates to a single envoy listener. However, within that listener, there will be multiple filter chains, each potentially configuring an HCM network filter that points to a RouteConfiguration

bewebi (approved):

Issue #3665, opened 2020-09-25
Support disabling UDS but allowing FDS and other discovery functionality


lgadban:
**Is your feature request related to a problem? Please describe.**
Need to be able to disable `Upstream` creation via discovery while still being able to use FDS for e.g. Lambdas

**Describe the solution you'd like**
A flag to disable the UDS portion of the `discovery` component

**Describe alternatives you've considered**
N/A, need to fully shut down discovery

bcollard:
There are chances a few customers need that backported to v1.9.

MitchAman:
> There are chances a few customers need that backported to v1.9. 

Ack, I can knock that out fairly quickly

bcollard:
Thank you Mitch!

PR #5546, opened 2021-11-08
install/helm: Update description for customstaticlayer to be more ver…

nfuden:
# Description
Update the document for static layer in helm to be more descriptive and link to what it interacts with along with its limitations.

# Context
The customstaticlayer fields documentation was sparse to a degree that it did not have any actionable state. 

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5481

PR #5586, opened 2021-11-17
Support disabling UDS

MitchAman:
# Description

Support disabling upstream discovery independent of the function discovery service.

This functionality is exposed via a new `discovery.udsOptions.enabled` value, which defaults to true. Setting `discovery.udsOptions.enabled=true` will have no affect if `discovery.enabled` is false (in other words, you cannot use upstream discovery if the discovery feature as a whole has been turned off. This is consistent with our current behavior, just calling out that `discovery.enabled` takes precedence).

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3665


projects/discovery/pkg/uds/syncer/setup_syncer.go
```diff
@@ -16,6 +16,13 @@ import (
 )
 
 func RunUDS(opts bootstrap.Opts) error {
+	udsEnabled := GetUdsEnabled(opts.Settings)
+	if !udsEnabled {
+		contextutils.LoggerFrom(opts.WatchOpts.Ctx).Infof("Upstream discovery "+
+			"(settings.discovery.udsEnabled) disabled. To enable, modify "+
+			"gloo.solo.io/Settings - %v", opts.Settings.GetMetadata().Ref())
+		return nil
```
MitchAman, line 24: Q: Should we return `(bool, error)` instead, where the bool indicates whether or not UDS has actually been started? This would allow us to differentiate between (UDS started with no errors) and (UDS did not start because it's disabled) in tests. Not sure there's much value for the `bool` in production?

install/helm/gloo/generate/values.go
```diff
@@ -224,6 +224,7 @@ type GlooDeployment struct {
 type Discovery struct {
 	Deployment     *DiscoveryDeployment `json:"deployment,omitempty"`
 	FdsMode        *string              `json:"fdsMode,omitempty" desc:"mode for function discovery (blacklist or whitelist). See more info in the settings docs"`
+	UdsEnabled     *bool                `json:"udsEnabled,omitempty" desc:"enable upstream discovery service (default true)."`
```
nfuden, line 227: Per https://github.com/solo-io/gloo/pull/5546#discussion_r745082615 docs would like this in `sentence-case`

```suggestion
	UdsEnabled     *bool                `json:"udsEnabled,omitempty" desc:"Enable upstream discovery service (default true)."`
```

projects/discovery/pkg/uds/syncer/setup_syncer.go
nfuden: This actually plays into a larger question I have about our change here.

The format you have here does correspond with the historical strategy from FDS disabled but we now have this odd case where someone could have the following:
Discovery is enabled but both FDS and UDS are disabled. This means that we can spin up an image that will forever sit at https://github.com/solo-io/gloo/blob/bd336085243b44b348dc96d3f27a199a245943d5/projects/discovery/cmd/main.go#L25 with nothing coming in.

While this probably shouldnt happen perhaps we should consider some sensible reporting at that level?

install/helm/gloo/generate/values.go
nfuden: Also question do we foresee wanting to turn this into something like we have for fds? Namely for future proofing do we think there may be a next request to set this to something like allowlist or blocklist?

nfuden (commented):
Steps seem sensible although there are a few lingering questions from a future looking perspective that while not blocking should at least be considered at this point.

install/helm/gloo/generate/values.go
MitchAman: I like your thought process. It's certainly possible we may want more than just an on/off switch for UDS in the future.  Which sounds more reasonable to you?

1. Replace `UdsEnabled` here with a reference to a `type UdsOptions struct` which in turn holds an `Enabled` field. This struct can be added to in the future if needed - more or less just for organization.
2. Replace `UdsEnabled` here with a `UdsMode` enum (analogous to `FdsMode`).  The issue I see here is that we need to have a default for the enum. The sane approach would be to have `UdsMode` default to `ENABLED` (maintaining the current behavior). But then that means we need a `UdsMode.DISABLED` enum value as well to support this feature.  Now if we want to elaborate this in the future, we would have `UdsMode = { ENABLED, DISABLED, WHITELIST, BLACKLIST }`, which is pretty ugly.

I'm in favor of option 1. If we have a primary on/off switch for UDS, we can still introduce new UDS behavior by adding fields to the `UdsOptions` struct. If `UdsOptions.Enabled = false`, all other fields in that struct would be ignored.

nfuden: I really like your point about approach two its nice to call out that the default behavior would have to be different from Fds for legacy default enablements.

Given this I think you are right that approach 1 is the better approach for future proofing so I think that as long as we forsee it being likely that we add power to options then we take this approach and if we are very confident that we wont add anything in the future then we go with the bool as is.

projects/discovery/pkg/uds/syncer/setup_syncer.go
MitchAman: Per offline discussion, logging has been added for when this state is detected.

projects/discovery/pkg/fds/syncer/setup_syncer.go
```diff
@@ -19,10 +20,14 @@ import (
...
+	fdsMode := syncerutils.GetFdsMode(opts.Settings)
 	if fdsMode == v1.Settings_DiscoveryOptions_DISABLED {
-		contextutils.LoggerFrom(opts.WatchOpts.Ctx).Info("function discovery disabled. to enable, modify "+
-			"gloo.solo.io/Settings %v", opts.Settings.GetMetadata().Ref())
+		contextutils.LoggerFrom(opts.WatchOpts.Ctx).Infof("Function discovery "+
+			"(settings.discovery.fdsMode) disabled. To enable, modify "+
+			"gloo.solo.io/Settings - %v", opts.Settings.GetMetadata().Ref())
+		// TODO: This should probably be called before RunFDS() is invoked, but the current discovery
+		//  architecture doesn't provide a better location
+		syncerutils.LogIfDiscoveryServiceUnused(&opts)
```
nfuden, line 30: Per agreement  this is the line that needs review from another team member to make sure we are all on board.

nfuden (commented):
Other than the comment on the log line that an another set of eyes is needed for this pull seems otherwise reasonable and good!

projects/discovery/pkg/syncer/utils.go
```diff
@@ -0,0 +1,41 @@
+package syncer
+
+import (
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+// If discovery is enabled, but both UDS & FDS are disabled, the discovery pod will not return
```
jenshu, line 9: is there a way for us to return early from the discovery `run()` if neither UDS or FDS is enabled?

projects/gloo/api/v1/settings.proto
```diff
@@ -174,6 +174,13 @@ message Settings {
         }
 
         FdsMode fds_mode = 1;
+
+        message UdsOptions {
+            // Enable upstream discovery service (default true).
```
Rachael-Graham, line 179: just to follow general convention ive seen in other protos
```suggestion
            // Enable upstream discovery service. Defaults to true.
```

projects/discovery/pkg/syncer/utils.go
MitchAman: @nfuden and I spent ~10-15 mins this morning poking around to see if we could do something like that, but there didn't appear to be an easy way without significant refactoring.

projects/gloo/api/v1/settings.proto
```diff
@@ -174,6 +174,13 @@ message Settings {
         }
 
         FdsMode fds_mode = 1;
+
+        message UdsOptions {
+            // Enable upstream discovery service. Defaults to true.
+            bool enabled = 1;
```
kdorosh, line 180: should we make this a google boolvalue so its nilable? then we can make nil default to true, and the settings object doesn't have to bloat with a setting most users won't use

kdorosh: bit of a nit, but it's nice to leave non-nullable fields in our API to ones that most users care about

projects/discovery/pkg/syncer/utils.go
MitchAman: Per stand up discussion, I've changed this to now return an error when this combination of configuration is encountered.

projects/discovery/pkg/fds/syncer/setup_syncer.go
MitchAman: Decision was to change this to error loudly rather than log and hang.

projects/discovery/pkg/fds/syncer/setup_syncer.go
```diff
@@ -19,11 +20,14 @@ import (
...
-		contextutils.LoggerFrom(opts.WatchOpts.Ctx).Info("function discovery disabled. to enable, modify "+
-			"gloo.solo.io/Settings %v", opts.Settings.GetMetadata().Ref())
-		return nil
+		contextutils.LoggerFrom(opts.WatchOpts.Ctx).Infof("Function discovery "+
+			"(settings.discovery.fdsMode) disabled. To enable, modify "+
+			"gloo.solo.io/Settings - %v", opts.Settings.GetMetadata().Ref())
+		if err := syncerutils.ErrorIfDiscoveryServiceUnused(&opts); err != nil {
+			return err
+		}
 	}
```
jenshu, line 31: should we otherwise return nil here, similar to RunUDS?

projects/discovery/pkg/syncer/utils.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+)
+
+// If discovery is enabled, but both UDS & FDS are disabled, we should error loudly as the
+// discovery pod is being deployed for no reason.
+func ErrorIfDiscoveryServiceUnused(opts *bootstrap.Opts) error {
+	settings := opts.Settings
+	udsEnabled := GetUdsEnabled(settings)
+	fdsEnabled := GetFdsEnabled(settings)
+	if !udsEnabled && !fdsEnabled {
```
jenshu, line 16: do we also have to add a check here that discovery is enabled? or does this function only get called if discovery is enabled?

projects/discovery/pkg/fds/syncer/setup_syncer.go
MitchAman: Yes, thanks for catching that!

projects/discovery/pkg/syncer/utils.go
MitchAman: If discovery isn't enabled, then the pod won't be deployed, so we're good.

https://github.com/solo-io/gloo/blob/05cf3b40529c2c54489ac157e8d89f36a7508d92/install/helm/gloo/templates/3-discovery-deployment.yaml#L2

jenshu (approved):

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -93,10 +93,14 @@ spec:
       validationServerGrpcMaxSizeBytes: {{ .Values.gateway.validation.validationServerGrpcMaxSizeBytes }}
 {{- end }}
 
-{{- if ne .Values.discovery.fdsMode "" }}
+{{- if .Values.discovery }}
   discovery:
+{{- if ne .Values.discovery.fdsMode "" }}
     fdsMode: {{.Values.discovery.fdsMode}}
 {{- end }}
+    udsOptions:
```
kdorosh, line 101: now that this is nillable, we should be able to omit it if the helm value is null. then our tests don't need to have this value populated either

install/helm/gloo/values-template.yaml
```diff
@@ -52,6 +52,8 @@ gloo:
 discovery:
   enabled: true
   fdsMode: WHITELIST
+  udsOptions:
+    enabled: true
```
kdorosh, line 56: we should be able to omit this now that it's nillable

install/helm/gloo/templates/18-settings.yaml
kdorosh: [for example](https://github.com/solo-io/gloo/blob/05cf3b40529c2c54489ac157e8d89f36a7508d92/install/helm/gloo/templates/2-gloo-service.yaml#L11-L13)

projects/discovery/pkg/syncer/utils.go
```diff
@@ -0,0 +1,41 @@
...
+		return errors.New("discovery (discovery.enabled) is enabled, but both UDS " +
+			"(discovery.udsOptions.enabled) and FDS (discovery.fdsMode) are disabled. " +
+			"While in this state, the discovery pod will be blocked. Consider disabling " +
```
kdorosh, line 19: not blocked, but perhaps just a no-op?

kdorosh (commented):
lgtm overall, just that last nit. will leave final review to another team member. thanks @MitchAman !

sam-heilbron (approved):
LGTM!

PR #5592, opened 2021-11-18
Feature/add shortname to crds

antonioberben:
# Description

Add a another shortname to VirtualService so I can use `kubectl get gvs -A` to list all VS resources.

When Istio is installed, the shortname and the resource name collides.

# Context

Having Istio and GE isntalled, a user listing all Istio VS resources has to:

```
kubectl get virtualservices.networking.istio.io -A
```

# Checklist:


antonioberben:
Agreed with @sam-heilbron , change will apply to all CRDs to be consistent

kdorosh (changes requested):
the crd files you have edited here are generated by solokit. the source of truth is [here](https://github.com/solo-io/gloo/blob/e06fa4697f39256aa031afc4695c9e3bc8f96ff2/projects/gateway/api/v1/gateway.proto#L25), for example

not sure off the top of my head, but we might need to update solokit support to add multiple short names

can point you towards tests once the implementation is working

kdorosh:
per @sam-heilbron 

> This is passing codgen because our CRD generation isn't entirely handled by solo-kit yet (though it would be good to). Right now we just take the CRD, re-gen the schemas and then modify the schemas. That is why we could manually change the shortname without codegen failures

sam-heilbron:
@antonioberben is this ready for review?

antonioberben:
@sam-heilbron , if we are ok for the auto-generated CRDs topic and I do not need to add tests out of this changes, I think yes. I will wait for your thumbsup regarding this two points. Thank you

kdorosh (approved):

PR #5602, opened 2021-11-19
(hybrid gateway) unique matcher name

bewebi:
Add a function to return a constant sized unique identifier for matchers instead of the `.String()` function

Unclear, though seemingly unlikely, that we require the standardization of equivalent matchers, but it's unlikely to be harmful either

### DIFF
--- a/projects/gloo/pkg/translator/names.go
+++ b/projects/gloo/pkg/translator/names.go
@@ -1,9 +1,8 @@
 package translator
 
 import (
-	"crypto/md5"
 	"fmt"
-	"io"
+	"hash/fnv"
 	"sort"
 
 	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
@@ -29,8 +28,8 @@ func matcherID(matcher *v1.Matcher) string {
 	sort.Strings(matcher.GetSslConfig().GetSniDomains())
 	sort.Strings(matcher.GetSslConfig().GetVerifySubjectAltName())
 
-	h := md5.New()
-	io.WriteString(h, matcher.String())
+	h := fnv.New64()
+	h.Write([]byte(matcher.String()))
 
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
--- a/projects/gloo/pkg/translator/names.go
+++ b/projects/gloo/pkg/translator/names.go
@@ -1,7 +1,9 @@
 package translator
 
 import (
+	"crypto/md5"
 	"fmt"
+	"io"
 	"sort"
 
 	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
@@ -27,5 +29,8 @@ func matcherID(matcher *v1.Matcher) string {
 	sort.Strings(matcher.SslConfig.SniDomains)
 	sort.Strings(matcher.SslConfig.VerifySubjectAltName)
 
-	return matcher.String()
+	h := md5.New()
+	io.WriteString(h, matcher.String())
+
+	return fmt.Sprintf("%x", h.Sum(nil))
 }
--- a/projects/gloo/pkg/translator/names.go
+++ b/projects/gloo/pkg/translator/names.go
@@ -1,7 +1,31 @@
 package translator
 
-import v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+import (
+	"fmt"
+	"sort"
+
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+)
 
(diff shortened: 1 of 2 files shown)


projects/gloo/pkg/translator/names.go
```diff
@@ -1,7 +1,36 @@
...
+			return matcher.GetSourcePrefixRanges()[i].GetAddressPrefix() < matcher.GetSourcePrefixRanges()[j].GetAddressPrefix()
+		}
+		return matcher.GetSourcePrefixRanges()[i].GetPrefixLen().GetValue() < matcher.GetSourcePrefixRanges()[j].GetPrefixLen().GetValue()
+	})
+
+	sort.Strings(matcher.GetSslConfig().GetAlpnProtocols())
+	sort.Strings(matcher.GetSslConfig().GetSniDomains())
+	sort.Strings(matcher.GetSslConfig().GetVerifySubjectAltName())
+
+	h := md5.New()
```
kdorosh, line 32: super nit: prefer fnv since it's faster and we aren't hashing here for security

projects/gloo/pkg/translator/names.go
```diff
@@ -1,7 +1,36 @@
...
+	"sort"
+
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+)
 
 func routeConfigName(listener *v1.Listener) string {
 	return listener.GetName() + "-routes"
 }
+
+func matchedRouteConfigName(listener *v1.Listener, matcher *v1.Matcher) string {
```
kdorosh, line 16: what will this be used for? any reason we don't just use the listener name appended with the index of the matcher? (like we have in other places IIRC)

projects/gloo/pkg/translator/names.go
```diff
@@ -1,7 +1,36 @@
...
+	sort.Slice(matcher.GetSourcePrefixRanges(), func(i, j int) bool {
+		if matcher.GetSourcePrefixRanges()[i].GetAddressPrefix() != matcher.GetSourcePrefixRanges()[j].GetAddressPrefix() {
+			return matcher.GetSourcePrefixRanges()[i].GetAddressPrefix() < matcher.GetSourcePrefixRanges()[j].GetAddressPrefix()
+		}
+		return matcher.GetSourcePrefixRanges()[i].GetPrefixLen().GetValue() < matcher.GetSourcePrefixRanges()[j].GetPrefixLen().GetValue()
+	})
+
+	sort.Strings(matcher.GetSslConfig().GetAlpnProtocols())
+	sort.Strings(matcher.GetSslConfig().GetSniDomains())
+	sort.Strings(matcher.GetSslConfig().GetVerifySubjectAltName())
```
kdorosh, line 30: if we keep this implementation, we should add some kind of reflection test that fails when we add new fields so we remember to update this code

in fact, doesn't solokit generate a hashing function for us we can use here? it might not have the sorting logic which could lead to different hashes for similar / equivalent protos, but if so, a more robust solution here might be to add a deterministic/stable hasher

projects/gloo/pkg/translator/names.go
bewebi: Will be used [here](https://github.com/solo-io/gloo/compare/feature/hybrid-gateway...hg-listener-translation-take-2#diff-3fc9c8b1139f1242e49ff7e5ccca3e48791180b5f90134688c610f3e21813f7dR216) and anywhere else we need a route config name 
I'm not sure if the index will always be in scope or consistent, though I think that may be the case
Regardless, I think it may be good not to rely on those

projects/gloo/pkg/translator/names.go
bewebi: solo-kit does generate hashing functions, but the Matcher isn't a solo-kit type
(if solo-kit generates functions for gloo types then I could use a refresher on how that works)

the reflection test is a good one to have, though I expect that the test for the exact hash will/may also fail if the type structure changes

sam-heilbron (commented):
The RouteCOnfigName is just a unique identifier used to match the ID of the RouteCOnfiguration (sent to envoy via xDS), with the HttpConnectionManager NetworkFilter. I think as long as we ensure that on each translation loop the RouteConfig and the HCM RouteConfig name match, then we are good. This means we just need this function to return a unique name. I think because of this, we should be safe to just rely on the index of the matcher since it's always unique for a single iteration.

bewebi:
Currently I think the parent listener is in context everywhere we need a matcher ID, so I'll try to proceed with that solution on the existing listener translation branch and come back to this PR if needed

PR #5615, opened 2021-11-22
Translate HybridGateway

bewebi:
Add HybridGateway translator which translates a HybridGateway to a HybridListener
Logic for MatchedGateway translation closely mirrors existing logic for translating TcpGateways and HttpGateways

Single unit test so far is mainly a sanity check, since no meaningful logic is meant to be added/changed for the sub-gateway translation. That said, additional tests may be necessary to validate logic that has been copied/rewritten in hybrid-specific functions.


projects/gateway/pkg/translator/hybrid.go
```diff
@@ -0,0 +1,213 @@
...
+				if len(snap.VirtualServices) == 0 {
+					snapHash := hashutils.MustHash(snap)
+					contextutils.LoggerFrom(ctx).Debugf("%v had no virtual services", snapHash)
+					continue // TODO: do we want the entire listener to fail or just this matched listener?
+				}
+
+				virtualServices := getVirtualServicesForMatchedHttpGateway(matchedGateway, gateway, snap.VirtualServices, reports)
+				applyGlobalVirtualServiceSettings(ctx, virtualServices)
+				validateVirtualServiceDomains(gateway, virtualServices, reports)
+				httpListener := t.desiredHttpListenerForHybrid(gateway, proxyName, virtualServices, snap, reports)
```
sam-heilbron, line 45: Is it possible to delegate responsibility to the HttpTranslator since it does the same thing here? 
ie a HybridTranslator has an HttpTranslator member, and when it encounters a MatchedGateway_HttpGateway, it relies on the HttpTranslator to run generate the config, and then modifies that output wit hthe matcher logic? That way if/when the HttpTranslator changes, the HybridTranslator just receives those changes for free

bewebi: Refactored to make this paradigm possible

projects/gateway/pkg/translator/http.go
```diff
@@ -269,10 +266,32 @@ func hasSsl(vs *v1.VirtualService) bool {
 
 func (t *HttpTranslator) desiredListenerForHttp(gateway *v1.Gateway, proxyName string, virtualServicesForGateway v1.VirtualServiceList, snapshot *v1.ApiSnapshot, reports reporter.ResourceReports) *gloov1.Listener {
 	var (
-		virtualHosts []*gloov1.VirtualHost
-		sslConfigs   []*gloov1.SslConfig
+		sslConfigs []*gloov1.SslConfig
 	)
```
MitchAman, line 270: nit: `var sslConfigs []*gloov1.SslConfig`

projects/gateway/pkg/translator/hybrid.go
```diff
@@ -0,0 +1,80 @@
...
+				SourcePrefixRanges: matchedGateway.GetMatcher().GetSourcePrefixRanges(),
+			}
+
+			switch gt := matchedGateway.GetGatewayType().(type) {
+			case *v1.MatchedGateway_HttpGateway:
+				// logic mirrors HttpTranslator.GenerateListeners
+				if len(snap.VirtualServices) == 0 {
+					snapHash := hashutils.MustHash(snap)
+					contextutils.LoggerFrom(ctx).Debugf("%v had no virtual services", snapHash)
+					continue // TODO: do we want the entire listener to fail or just this matched listener?
```
MitchAman, line 40: I would assume just this matched listener. We should probably still add an error to the reports here though

bewebi: That's what I thought initially (when I left this comment) but my thinking has changed since then
The thing is all http listeners will fail this check, so we could reasonably expect that there will be multiple/persistent failures, all with the same error message, and with only TcpListeners actually added to the HybridListener
I'm not sure how useful that is, but in the case where only HttpGateways are present in the first place the behavior will be the same regardless apart from the error being logged repeatedly
We could at least limit the logging to at most once

projects/gateway/pkg/translator/http.go
```diff
@@ -157,13 +157,13 @@ func validateVirtualServiceDomains(gateway *v1.Gateway, virtualServices v1.Virtu
 	}
 }
 
-func getVirtualServicesForGateway(gateway *v1.Gateway, virtualServices v1.VirtualServiceList, reports reporter.ResourceReports) v1.VirtualServiceList {
+func getVirtualServicesForHttpGateway(httpGateway *v1.HttpGateway, parentGateway *v1.Gateway, virtualServices v1.VirtualServiceList, reports reporter.ResourceReports, gatewaySsl bool) v1.VirtualServiceList {
```
sam-heilbron, line 160: why do we pass both httpGateway and gatewaySsl to this function? Can't we derive one from the other?

sam-heilbron (commented):
Looking good! Can we add some e2e tests as well?

projects/gateway/pkg/translator/http.go
bewebi: This is in order to not have to do switching logic based on the type of the parent gateway within the function
It's much simpler to have the caller do that since the type is known at call-time [here](https://github.com/solo-io/gloo/pull/5615/files#diff-2d2f05b899c26dca4d0a1a9549c903fcde9c1b61cef6470de185af0bb9f1cac2R95) and [here](https://github.com/solo-io/gloo/pull/5615/files#diff-9b34716355b8c5a1533d9b1ab57ec8779a52408616494e5f469d73757f5378b1R50)

bewebi:
TODO: e2e tests for hybrid gateway translation

Issue #4447, opened 2021-03-17
Refresh grant to extend the session, given a minimum TTL threshold


bcollard:
Gloo refreshes the `id_token` when it is expired. 
If I want to prevent edge cases where the `id_token` is about to expire (say, in the next 10ms), and the backend actually receives an expired token. 
Can I prevent that with a threshold, like a minimum acceptable TTL, that will trigger the refresh grant?

sam-heilbron:
During OIDC, if the jwt is expired, we execute a refresh flow: https://github.com/solo-io/ext-auth-service/blob/6b82e1ddfb89107e48c365255f81e74cedf9d90c/pkg/config/oidc/oidc_request.go#L450

However, if the jwt is not yet expired, it is accepted and the request is sent upstream. If that upstream validates the token separately, it will validate that token, which is now expired. The effect is that the request is denied, even though there is a valid token. 

The ask is to provide a buffer to the jwt expiry, so instead of refreshing after it is expired, refresh after expiry_time-buffer. 

The oidc refresh flow: https://github.com/solo-io/ext-auth-service/blob/6b82e1ddfb89107e48c365255f81e74cedf9d90c/pkg/config/oidc/oidc_request.go#L450

which is triggered after the token claims are parsed: https://github.com/solo-io/ext-auth-service/blob/6b82e1ddfb89107e48c365255f81e74cedf9d90c/pkg/config/oidc/oidc_request.go#L390

we could probably rely on the `VerifyExpiresAt` exposed on the MapClaims here: https://github.com/dgrijalva/jwt-go/blob/9742bd7fca1c67ba2eb793750f56ee3094d1b04f/map_claims.go#L36

sam-heilbron:
Should the buffer be hardcoded or configurable? Perhaps configurable but with a sane default (2s)

@chrisgaun why do we need this feature in the first place?

marcelsigner:
There are several reasons for the request:

- first of all every backend must repeat the access control because backends also call each other because we have applied a zero-trust policy inside the cluster. Backends do a much more fine-grained access control. In Gloo Edge we only do a coarse grained access control.
- even if the access token is still valid when backend A receives the request, the access_token may expire when backend A calls backend B.
- in some cases backend A has to exchange the token to call backend B. The OAuth Server rejects the token_exchange if the token is already expired. Reasons for the token_exchange:
       - in some cases backend A sends an asynchronous message to backend B passing the access_token in the message header.
[...]

PR #5616, opened 2021-11-22
Add pre_expiry_buffer to extauth API

MitchAman:
# Description

Add `pre_expiry_buffer` to extauth API to allow for refreshing id-tokens moments before they're going to expire.

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -310,6 +310,9 @@ message UserSession {
         // When set, refresh expired id-tokens using the refresh-token. Defaults to true.
         // Explicitly set to false to disable refreshing.
         google.protobuf.BoolValue allow_refreshing = 4;
+        // Specifies an offset in which an id-token will be refreshed prior to its actual expiration.
+        // Defaults to 2 seconds. A duration of 0 will wait to refresh a token until it has expired.
+        google.protobuf.Duration pre_expiry_refresh_offset = 5;
```
MitchAman, line 315: Please somebody help me name this better

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -310,6 +310,10 @@ message UserSession {
         // When set, refresh expired id-tokens using the refresh-token. Defaults to true.
         // Explicitly set to false to disable refreshing.
         google.protobuf.BoolValue allow_refreshing = 4;
+        // Specifies a time buffer in which an id-token will be refreshed prior to its
+        // actual expiration. Defaults to 2 seconds. A duration of 0 will only refresh
+        // tokens after they have already expired.
+        google.protobuf.Duration pre_expiry_buffer = 5;
```
EItanya, line 316: What do you think about making this a grace period ratio, rather than a duration, as I mentioned in standup. For example: if the token lifetime is 15 minutes, and you specify a `0.9`, then the token would be refreshed at 13.5 minutes. This way the timing is flexible, and based on the token lifetime.

MitchAman: @EItanya Given the use case: https://github.com/solo-io/gloo/issues/4447#issuecomment-918873027

It seems that users want to a way to say "refresh this token if it's going to expire within _x amount of time_ after gloo extauth validates it." While a grace period ratio addresses the issue of a token expiring between the times that extauth and a backend service validate it, it's less obvious what the ratio should be set to.  Because the token TTL is not configured by gloo, in order to set an appropriate grace period ratio, they would first have to check what their IDP has configured the TTL to.

That being said, I'm not strictly opposed to the idea, but not sure I see the advantage to having this buffer scale with the TTL of the token. If the token TTL is lowered, the buffer could be lowered short enough that tokens begin expiring before backends validate them again. If the token TTL is increased, then the buffer window is increased, and tokens may be getting refreshed more frequently than necessary (which I don't necessarily see as a bad thing).

LMK if you feel strongly about this

MitchAman: A potential concern would be that since extauth is not configured with the TTL, we would have to calculate that via `exp - iat`. The OIDC spec [1] indicates that while `exp` is REQUIRED, `iat` is only OPTIONAL. In the event that `iat` is missing, we would have no way of determining the TTL

[1] https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

EItanya: This all sounds good to me 👍

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -520,6 +521,7 @@ Deprecated: Prefer OAuth2
 | `keyPrefix` | `string` | Key prefix inside redis. |
 | `cookieName` | `string` | Cookie name to set and store the session id. If empty the default "__session" is used. |
 | `allowRefreshing` | [.google.protobuf.BoolValue](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/bool-value) | When set, refresh expired id-tokens using the refresh-token. Defaults to true. Explicitly set to false to disable refreshing. |
+| `preExpiryBuffer` | [.google.protobuf.Duration](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/duration) | Specifies a time buffer in which an id-token will be refreshed prior to its actual expiration. Defaults to 2 seconds. A duration of 0 will only refresh tokens after they have already expired. |
```
artberger, line 524: Does this require them to set the `allowRefreshing` setting to `true`? If so, might be worth mentioning. Otherwise, lgtm for docs.

> Specifies a time buffer in which an id-token will be refreshed prior to its actual expiration. Defaults to 2 seconds. A duration of 0 will only refresh tokens after they have already expired. To refresh tokens, you must also set 'allowRefreshing' to 'true'; otherwise, this field is ignored.

MitchAman: Yes, that's correct, updating now. Thanks!

ben-taussig-solo (approved):

PR #5623, opened 2021-11-23
Support New GatewayType: HybridGateway

### LINKED ISSUE #5324: Client IP matching in Virtual Services
**Describe the solution you'd like**
Virtual Services should also support Client IP matching. Some use cases include:
- Different TLS and routing configurations for the same domain, depending on Client IP
- Exposing Virtual Services to specific Client IPs

bewebi:
# Description

- Add HybridGateway and HybridListener types which contain a list of MatchedGateway and MatchedListener respectively
  - MatchedGateway consists of a Matcher and a oneof HttpGateway and TcpGateway, MatchedListener consists of a Matcher and a oneof HttpListener and TcpGateway
- Implement and test translation of HybridGateway to HybridListener and HybridListener to Envoy Proxy
- Logic is copied/reused as much as possible from existing Http/Tcp functionality

# Context

- Resolves #5324 and goes beyond the specific scope of that issue, allowing SNI domain matching and Http/Tcp gateway mixing in addition to the original ask
- Components were developed in separate PRs and merged into this branch until part of the way through e2e testing; it may help to reference those since the feature branch has such a large diff
- An additional SNI matching test and/or successful TCP Dial test may be added but may not be strictly necessary

# Checklist:

- [x] I have made corresponding changes to the documentation (see #5676)

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5324


changelog/v1.10.0-beta11/hybrid-gateway.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/5324
+    description: >
+      Add HybridGateway CRD to allow for exposing multiple HTTP and/or TCP Gateway on a single port, matching based on
```
sam-heilbron, line 5: Let's clarify that we're not adding a new CRD, but intsead a new type of Gateway, one that wraps multiple existing gateway types, and allows more fine-grained control over FilterChainMatching.

projects/gateway/api/v1/gateway.proto
```diff
@@ -126,6 +130,26 @@ message TcpGateway {
...
+    Matcher matcher = 1;
+
+    oneof GatewayType {
+        HttpGateway http_gateway = 2;
+        TcpGateway tcp_gateway = 3;
+    }
+}
+
+message Matcher {
+    gloo.solo.io.SslConfig ssl_config = 1;
```
sam-heilbron, line 148: Perhaps a comment outlining how this config will affect the envoy config. For example, only the server_names will be used to generate the filterchain match, but the other ssl config will affect the transport socket on the FilterChain.

projects/gateway/api/v1/gateway.proto
```diff
@@ -126,6 +130,26 @@ message TcpGateway {
     gloo.solo.io.TcpListenerOptions options = 8;
 
 }
+
+message HybridGateway {
+    repeated MatchedGateway matched_gateways = 1;
+}
+
+message MatchedGateway {
+    Matcher matcher = 1;
```
sam-heilbron, line 139: Let's add a comment outlining what matchers are, and what happens if a user configures overlapping/identical matchers. What if the matcher is empty? Can you have multiple empty matchers?

projects/gateway/pkg/translator/hybrid.go
```diff
@@ -0,0 +1,100 @@
...
+						contextutils.LoggerFrom(ctx).Debugf("%v had no virtual services", snapHash)
+						loggedError = true // only log no virtual service error once
+					}
+					continue
+				}
+
+				virtualServices := getVirtualServicesForHttpGateway(matchedGateway.GetHttpGateway(), gateway, snap.VirtualServices, reports, matchedGateway.GetMatcher().GetSslConfig() != nil)
+				applyGlobalVirtualServiceSettings(ctx, virtualServices)
+				validateVirtualServiceDomains(gateway, virtualServices, reports)
+				httpListener := t.HttpTranslator.desiredHttpListener(gateway, proxyName, virtualServices, snap, reports)
```
sam-heilbron, line 61: Is there a way to delegate more of this to a single function call? I'm a little confused why we need to rely on a private method of the HttpTranslator, and can't instead just call `t.HttpTranslator` to handle all the generation.

projects/gateway/pkg/translator/hybrid.go
```diff
@@ -0,0 +1,100 @@
...
+
+				hybridListener.MatchedListeners = append(hybridListener.GetMatchedListeners(), &gloov1.MatchedListener{
+					Matcher: matcher,
+					ListenerType: &gloov1.MatchedListener_HttpListener{
+						HttpListener: httpListener,
+					},
+				})
+			case *v1.MatchedGateway_TcpGateway:
```
sam-heilbron, line 69: Similar to the http case, can we just delegate to a TcpTranslator member on this type? That way this translator is just responsible for translating the matcher, but everything else is done by the existing code.

projects/gloo/pkg/plugins/tls_inspector/plugin.go
```diff
@@ -37,21 +37,52 @@ func (p *plugin) ProcessListener(params plugins.Params, in *v1.Listener, out *en
 		},
 	}
 
-	// automatically add tls inspector when ssl is enabled
-	if in.GetSslConfigurations() != nil {
+	if shouldIncludeTlsInspectorListenerFilter(in) {
```
sam-heilbron, line 40: I know I had written this, so apologies for the laziness. Lets add a comment to this plugin outlining why we're handling the hybrid gateway directly in the plugin.

projects/gloo/pkg/translator/route_config.go
```diff
@@ -469,6 +470,56 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+	report       *validationapi.HybridListenerReport
+
+	requireTlsOnVirtualHosts bool
+}
+
+func (h *hybridRouteConfigurationTranslator) ComputeRouteConfiguration(params plugins.Params) []*envoy_config_route_v3.RouteConfiguration {
+	var outRouteConfigs []*envoy_config_route_v3.RouteConfiguration
+	for _, matchedListener := range h.listener.GetMatchedListeners() {
+		httpListener := matchedListener.GetHttpListener()
+		if httpListener == nil {
```
sam-heilbron, line 490: Can we add a comment outlining why it is safe to continue?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -469,6 +470,56 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+
+			routeConfigName:          rcName,
+			requireTlsOnVirtualHosts: matcher.GetSslConfig() != nil,
+		}
+		virtualHosts := matchedListenerRouteConfigurationTranslator.computeVirtualHosts(params)
+
+		outRouteConfigs = append(outRouteConfigs, &envoy_config_route_v3.RouteConfiguration{
+			Name:                           rcName,
+			VirtualHosts:                   virtualHosts,
+			MaxDirectResponseBodySizeBytes: h.parentListener.GetRouteOptions().GetMaxDirectResponseBodySizeBytes(),
```
sam-heilbron, line 516: Any reason we can't delegate to the public interface instead of using `computeVirtualHosts` qand then duplicating the RouteConfguration constructor?

projects/gloo/pkg/utils/names.go
```diff
@@ -0,0 +1,26 @@
+package utils
+
+import (
+	"fmt"
+
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+)
+
+func RouteConfigName(listener *v1.Listener) string {
```
sam-heilbron, line 9: Since this is only used by the RouteConfigurationTranslators (I could be wrong), could we have these utilities live closer to the place they're used?

projects/gloo/pkg/utils/names.go
```diff
@@ -0,0 +1,26 @@
...
+		return RouteConfigName(listener)
+	}
+
+	for i, mg := range hybridListener.GetMatchedListeners() {
+		if mg.GetMatcher().Equal(matcher) {
+			return fmt.Sprintf("%s-%d", RouteConfigName(listener), i)
+		}
+	}
+
+	return fmt.Sprintf("%s-%s", RouteConfigName(listener), matcher.String())
```
sam-heilbron, line 25: I don't quite follow why we iterate the matchers of the gateways and then if we don't find one return the stringified matcher. Can't we just return the routeConfigName + stringified matcher without having to iterate the list?

projects/gloo/pkg/utils/names.go
```diff
@@ -0,0 +1,26 @@
...
+
+func RouteConfigName(listener *v1.Listener) string {
+	return listener.GetName() + "-routes"
+}
+
+func MatchedRouteConfigName(listener *v1.Listener, matcher *v1.Matcher) string {
+	hybridListener := listener.GetHybridListener()
+	if hybridListener == nil {
+		return RouteConfigName(listener)
+	}
```
sam-heilbron, line 17: Definiteily open to your thoughts here, but I'd almost rather this function just handle a hybridListener and a matcher and return the string. This way it does one thing, provide the RouteConfig name for a matched gateway, instead of trying to handle the case where it's also not a matched gateway. Then the calling function can decide which utility to rely on based on the listener type

projects/gloo/pkg/utils/validation/proxy_validation_test.go
```diff
@@ -13,7 +18,9 @@ var _ = Describe("validation utils", func() {
 	numListeners := 3
 	numVhosts := 5
 	numRoutes := 8
-	makeProxy := func(http bool) *v1.Proxy {
+	numTcpListeners := 4
+	numHttpListeners := 2
+	makeProxy := func(http, tcp bool) *v1.Proxy {
```
sam-heilbron, line 23: Not a huge deal since this is a test, but I find it that if we're passing booleans into a function, it likely means we could just have different functions, each with a single responsibility: https://twitter.com/dmokafa/status/1399735140234973186

projects/gateway/pkg/translator/hybrid.go
bewebi: Are you suggesting moving the first three calls in that sequence under the umbrella of the `desiredHttpListener()` function?

bewebi: See [here](https://github.com/solo-io/gloo/pull/5615#discussion_r754587361) for previous conversation on this section

projects/gateway/pkg/translator/hybrid.go
bewebi: The signature for `GenerateListeners()` doesn't really allow for this, since it accepts `[]*v1.Gateway`
It's conceivable that some shared logic could be factored out in order to get a helper function function that accepts a `*v1.TcpGateway` which both Hybrid and Tcp `GenerateListeners()` could invoke

projects/gateway/pkg/translator/hybrid.go
bewebi: Also relevant to the HTTP case: https://github.com/solo-io/gloo/pull/5623#discussion_r765080309

projects/gateway/api/v1/gateway.proto
```diff
@@ -126,6 +130,34 @@ message TcpGateway {
...
+    // Empty Matchers are effectively catch-alls, and there can be no more than one empty Matcher per HybridGateway
+    Matcher matcher = 1;
+
+    oneof GatewayType {
+        HttpGateway http_gateway = 2;
+        TcpGateway tcp_gateway = 3;
+    }
+}
+
+message Matcher {
```
gunnar-solo, line 151: maybe I'm missing some understanding or context, here.  Why are we bothering to define a new `Matcher` type?  It feels like unnecessary nesting for just the 2 fields of `ssl_config` and `source_prefix_ranges`

projects/gloo/pkg/translator/route_config.go
bewebi: This seems like a decent idea, and I don't recall whether I'd considered it
The main reason not to would be to avoid validating the parent listener's ssl config, which is deprecated for HybridListeners and should be ignored if specified

projects/gloo/pkg/utils/names.go
bewebi: `MatchedRouteConfigName()`, which calls `RouteConfigName()`, is used by proxy validation [here](https://github.com/solo-io/gloo/blob/5ff3657459874a08647ac6aab45b686e001e09a6/projects/gloo/pkg/utils/validation/proxy_validation.go#L60)
I don't remember the exact details, but leaving `names.go` in the translator package leads to an import cycle

projects/gloo/pkg/utils/names.go
bewebi: The goal here is to have a deterministic, unique, and ideally concise route config name
`matcher.String()` does not satisfy this since it does not account for slice/map ordering and is extremely verbose
Using the index of the matcher's MatchedListener within the HybridListener satisfies all three

We really should never exit the `for` loop. The catchall case is mainly there as an "emergency" measure, to output something unique so we can be sure that nothing is overwritten.

projects/gloo/pkg/utils/names.go
bewebi: Since this function does not return an error, I wanted to have a meaningful return value regardless of if the input is what we expect
In practice we do expect the caller to use the appropriate function and not to call `MatchedRouteConfigName()` when `RouteConfigName()` would suffice

projects/gateway/api/v1/gateway.proto
bewebi: Two main reasons:
- The matcher mirrors Envoy’s [FilterChainMatch](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchainmatch) and we want to leave it nice and extensible should we want to expose more of those Envoy fields in the future
- The matcher is a useful object for us to pass around that isn’t always tied to the Listener/Gateway, for example [here](https://github.com/solo-io/gloo/pull/5623/files#diff-3fc9c8b1139f1242e49ff7e5ccca3e48791180b5f90134688c610f3e21813f7dR253)

projects/gateway/api/v1/gateway.proto
Rachael-Graham: could you add periods to the end of each comment? theyre rendered all together [in the generated doc](https://github.com/solo-io/gloo/pull/5623/files#diff-a2bd4c126696137192fe7f8dc9845ae974bc45b0b286dd3d75c04dc371175b41R149)

periods needed [here](https://github.com/solo-io/gloo/pull/5623/files#diff-09228a26f085b5fd0cb14643f37fa6c2fd386042a48d2c8363abf82bf2c7964cR153) too

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -188,3 +189,102 @@ func merge(values []string, newValues ...string) []string {
...
+		Filters: listenerFilters,
+	}
+	return m.applyMatcherToFilterChain(snap, matcher, fc)
+}
+
+func (m *matcherFilterChainTranslator) applyMatcherToFilterChain(snap *v1.ApiSnapshot, matcher *v1.Matcher, fc *envoy_config_listener_v3.FilterChain) *envoy_config_listener_v3.FilterChain {
+	fcm := fc.GetFilterChainMatch()
+	if fcm == nil {
+		fcm = &envoy_config_listener_v3.FilterChainMatch{}
+		fc.FilterChainMatch = fcm
```
sam-heilbron, line 257: I'm a little confused by what these lines are doing/why. Can we add a comment clarifying? 
Side note: a good quote from Clean Code "Functions should do something, or answer something, but not both.". Is there a way to make it so this fucntion either returns the FilterChain or modifies the FilterChain passed in, but not both?

projects/gloo/pkg/translator/route_config.go
sam-heilbron: One way to get around this is to make that behavior configurable. So when you create a RouteConfigTranslator, you tell it whether or not to validate the parent listener's ssl config, or you just provide it the ssl config to validate (similate to how you can inject `requireTlsOnVirtualHosts`). I'm nervous that if we have seprate code paths which are intended to do the same thing, over time they will diverge

projects/gloo/pkg/utils/names.go
sam-heilbron: That makes sense. Since matcher.String does not satisfy the requirements however, should we instead return an error? I'm concerned that we'll lose the context on why we have this return statement.

Also, how does the slice ordering affect matcher.String? I'm confused how adding the index differs from the string value. The index will be unique but will it be deterministic?

test/e2e/hybrid_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+		// Create static upstream for tcp server
+		_, err = testClients.UpstreamClient.Write(tcpServerUs, clients.WriteOpts{Ctx: ctx})
+		Expect(err).NotTo(HaveOccurred())
+
+		// Run envoy
+		err = envoyInstance.RunWithRoleAndRestXds(services.DefaultProxyName, testClients.GlooPort, testClients.RestXdsPort)
+		Expect(err).NotTo(HaveOccurred())
+
+		// Create a hybrid proxy routing to the upstream and wait for it to be accepted
```
sam-heilbron, line 92: nit: this comment is misleading, sice we do not in fact wait for the proxy to be accepted

test/e2e/hybrid_test.go
```diff
@@ -0,0 +1,346 @@
...
+				resp, err := client.Do(req)
+				if err != nil {
+					return 0, err
+				}
+				return resp.StatusCode, nil
+			}, "5s", "0.5s").Should(Equal(http.StatusOK))
+
+		})
+
+		//It("tcp connection fails", func() {
```
sam-heilbron, line 200: nit: remove

test/e2e/gateway_test.go
```diff
@@ -713,6 +716,573 @@ var _ = Describe("Gateway", func() {
 			})
 
 		})
+
+		Context("hybrid gateway", func() {
+
```
sam-heilbron, line 721: nit: Let's add a comment here outlining the goal/intention of these tests. I think we should be clear that they are intended as smoke tests for the HybridGateway. Most of our testing is around the Http and Tcp Gateway's since the core logic is shared with those Gateways

test/e2e/hybrid_test.go
```diff
@@ -0,0 +1,346 @@
...
+				return nil
+			}, "3s", "0.5s").Should(HaveOccurred())
+
+		})
+
+	})
+
+})
+
+func getProxyHybridNoMatcher(namespace, name string, envoyPort uint32, tcpUpsteam *core.ResourceRef) *gloov1.Proxy {
```
sam-heilbron, line 259: why do we have an httpGateway and a tcpGateway on this proxy?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -469,6 +470,57 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+
+		matchedListenerRouteConfigurationTranslator := &httpRouteConfigurationTranslator{
+			plugins: h.plugins,
+			proxy:   h.proxy,
+
+			parentListener: h.parentListener,
+			listener:       httpListener,
+
+			parentReport: h.parentReport,
+			report:       h.report.GetMatchedListenerReports()[utils.MatchedRouteConfigName(h.parentListener, matcher)].GetHttpListenerReport(),
```
sam-heilbron, line 507: Are we sure this won't throw an NPE?

projects/gateway/pkg/translator/hybrid.go
sam-heilbron: Yeah I was just suggesting that if we delegate to an instance of an HttpTranslator, that we use the public methods instead of a private one.

projects/gateway/api/v1/gateway.proto
```diff
@@ -126,6 +130,26 @@ message TcpGateway {
...
+    gloo.solo.io.SslConfig ssl_config = 1;
+
+    repeated .solo.io.envoy.config.core.v3.CidrRange source_prefix_ranges = 2;
```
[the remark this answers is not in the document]
bewebi: I'm not sure, and I don't really recall a conversation around this
If I understand correctly, and I'm not sure I do, the original request seems to describe `source_prefix_range` since it's the _client_ IP which would be the first in the request chain rather than `direct_source_prefix_range` which I interpret as the last hop's IP
I don't have an issue with adding both, but it might be worthwhile to ask directly, or even release this feature as is and see if it works for them, before going to the trouble
I would expect this addition to take about a day give or take depending on how exhaustively it's tested

projects/gateway/pkg/translator/hybrid.go
bewebi: The public [`GenerateListeners()`](https://github.com/solo-io/gloo/blob/a58d3415ce6ec8d8053bea91760947df0dacef28/projects/gateway/pkg/translator/http.go#L83) function simply doesn't match signature-wise with what we're working with in the HybridGateway scope
It wants a `[]*v1.Gateway` but we're dealing with an `*v1.HttpGateway`

I'm still not clear if this is what you meant though, might be worth sorting out offline

projects/gloo/pkg/translator/route_config.go
bewebi: Pushed a commit where we skip validation in the http implementation of the public interface if the parent listener is a hybrid listener and call that from the hybrid implementation
I _think_ this accomplishes what we're looking for here, feel free to clarify if not

projects/gloo/pkg/utils/names.go
bewebi: The `.String()` function returns a full representation of the struct, which has no bound on length...given that the name is propagated to Envoy, used in testing, and potentially visible elsewhere (I haven't followed that thread all the way to the end) it's definitely desirable, if not strictly necessary from a technical standpoint, to have a concise, ideally constant sized, name

I'm not sure if the slice ordering piece would affect determinism in practice, since references should always stem from the same source of truth (ie the Gateway CR), but it is the case that structs that has the same values in a slice field but in a different order will have different `.String()` outputs.

Likewise, the index should be deterministic since all references stem from the Gateway CR

(If you've got more questions about how/why to use index here perhaps reach out to the [developer who suggested it](https://github.com/solo-io/gloo/pull/5602#pullrequestreview-813061899) 😉)

bewebi: As far as returning an error, it's mostly just clunky, as we use the result directly to index into maps within reports and set fields directly in struct construction, so if we can prove/be super highly confident that it will never happen in practice I'd prefer not to
That said, no technical reason it can't be done

projects/gateway/api/v1/gateway.proto
```diff
@@ -127,6 +131,34 @@ message TcpGateway {
...
+
+    oneof GatewayType {
+        HttpGateway http_gateway = 2;
+        TcpGateway tcp_gateway = 3;
+    }
+}
+
+message Matcher {
+    // Gloo use SNI domains as matching criteria for Gateway selection
+    // The other ssl_config properties will be applied to the outputFilterChain's transport socket
```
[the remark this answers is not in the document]
bewebi: It's effectively ignored afaict
I can clarify behavior with a comment, but will that be ok?

projects/gateway/pkg/translator/hybrid.go
```diff
@@ -0,0 +1,100 @@
...
+			switch gt := matchedGateway.GetGatewayType().(type) {
+			case *v1.MatchedGateway_HttpGateway:
+				// logic mirrors HttpTranslator.GenerateListeners
+				if len(snap.VirtualServices) == 0 {
+					if !loggedError {
+						snapHash := hashutils.MustHash(snap)
+						contextutils.LoggerFrom(ctx).Debugf("%v had no virtual services", snapHash)
+						loggedError = true // only log no virtual service error once
+					}
+					continue
```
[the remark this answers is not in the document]
bewebi: Because it's not actually an error unless/until we have a MatchedGateway with an HttpGateway member
There might be a cleverer way to determine this, but this way we don't iterate over the MatchedGateways more than we need to

projects/gateway/pkg/translator/translator.go
```diff
@@ -49,6 +49,7 @@ func NewDefaultTranslator(opts Opts) *translator {
 	return NewTranslator([]ListenerFactory{
 		&HttpTranslator{WarnOnRouteShortCircuiting: warnOnRouteShortCircuiting},
 		&TcpTranslator{},
+		&HybridTranslator{HttpTranslator: &HttpTranslator{WarnOnRouteShortCircuiting: warnOnRouteShortCircuiting}},
```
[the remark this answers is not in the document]
bewebi: What would the tcp one accomplish? If the idea is to use the public interface (in the tcp case that's the only member function) the same issue applies as with [http](https://github.com/solo-io/gloo/pull/5623#discussion_r764401499)

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -188,3 +189,102 @@ func merge(values []string, newValues ...string) []string {
...
+				m.parentListener,
+				listenerType.HttpListener,
+				m.report.GetMatchedListenerReports()[utils.MatchedRouteConfigName(m.parentListener, matchedListener.GetMatcher())].GetHttpListenerReport(),
+				m.httpPlugins,
+				m.hcmPlugins,
+				utils.MatchedRouteConfigName(m.parentListener, matchedListener.GetMatcher()))
+			networkFilters := nft.ComputeNetworkFilters(params)
+			if len(networkFilters) == 0 {
+				return nil
+			}
```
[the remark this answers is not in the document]
bewebi: I think there's a meaningful different between returning `nil` compared to a filter chain with no network filters
That said, I think this is a good catch in that we'd rather `continue` here and collect filter chains from other gateways than return nothing because of the configuration of one gateway

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -188,3 +189,102 @@ func merge(values []string, newValues ...string) []string {
...
+			sublistenerFilterChainTranslator := tcpFilterChainTranslator{
+				plugins: m.tcpPlugins,
+
+				parentListener: m.parentListener,
+				listener:       matchedListener.GetTcpListener(),
+
+				report: m.report.GetMatchedListenerReports()[utils.MatchedRouteConfigName(m.parentListener, matchedListener.GetMatcher())].GetTcpListenerReport(),
+			}
+
+			unmatchedFilterChains := sublistenerFilterChainTranslator.ComputeFilterChains(params)
```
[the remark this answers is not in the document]
bewebi: I'm not sure I see where that's happening, if you're implying that there's an exponential behavior her, or what the alternative would be

projects/gloo/pkg/translator/filter_chain.go
bewebi: The filter chain may or may not have the `FilterChainMatch` field set
Iff it isn't set yet, then we need to create an empty one and assign it

I'll refactor so that it doesn't return anything

test/e2e/hybrid_test.go
bewebi: We use both for testing purposes

projects/gloo/pkg/translator/route_config.go
bewebi: For a given iteration we can be confident of this, since we create the reports objects using the same function [here](https://github.com/solo-io/gloo/blob/5ff3657459874a08647ac6aab45b686e001e09a6/projects/gloo/pkg/utils/validation/proxy_validation.go#L60)
We could make this a little more bulletproof, since it's possible for an upstream code change to break the assumption that the keys don't change after creation, but it should be fine as-is

projects/gateway/api/v1/gateway.proto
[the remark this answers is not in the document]
sam-heilbron: I'm good with holding off and adding it later. I don't think this change should be that time consuming though, since we don't need to test the envoy functionality per se, just that we configure it properly. So I could see the new field and a unit test being enough.

sam-heilbron (commented):
LGTM!

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -179,7 +177,32 @@ func stripInvalidListenersAndVirtualHosts(ctx context.Context, proxiesToWrite Ge
...
+						}
+					}); err != nil {
+						return nil, err
+					}
+
+					sort.SliceStable(validVhosts, func(i, j int) bool {
+						return validVhosts[i].GetName() < validVhosts[j].GetName()
+					})
+
+					httpListener.VirtualHosts = validVhosts
```
sam-heilbron, line 204: Can we DRY this, it's identical to the http case? I wonder if we just make it a function and the httplistener calls it once and hybrid calls it n times. Or we say extract all http listeners from the listener (both http and hybrid.http) and then for each one of those, execute this code.

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -238,21 +256,33 @@ func transitionFunc(proxiesToWrite GeneratedProxies, statusClient resources.Stat
...
+				originalMatchedListeners := originalListener.GetHybridListener().GetMatchedListeners()
+				for _, matchedListener := range subListener.HybridListener.GetMatchedListeners() {
+					desiredHttpListener := matchedListener.GetHttpListener()
+					if desiredHttpListener == nil {
+						continue
+					}
+					var originalHttpListener *gloov1.HttpListener
+					for _, oml := range originalMatchedListeners {
+						if oml.GetMatcher().Equal(matchedListener.GetMatcher()) {
+							originalHttpListener = oml.GetHttpListener()
```
sam-heilbron, line 277: Why do we need the matcher?

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
```diff
@@ -188,18 +194,20 @@ var _ = Describe("ReconcileGatewayProxies", func() {
...
 
 				px := getProxy()
 
-				Expect(px.Listeners).To(HaveLen(3))
+				Expect(px.Listeners).To(HaveLen(4))
 				Expect(px.Listeners[0]).To(HaveGeneration(100))
 				Expect(px.Listeners[1].Name).To(Equal(translator.ListenerName(snap.Gateways[0]))) // maps to gateway[0]
 				Expect(px.Listeners[1]).To(HaveGeneration(0))                                     // maps to gateway[0]
 				Expect(px.Listeners[2]).To(HaveGeneration(100))
+				Expect(px.Listeners[3]).To(HaveGeneration(100))
```
sam-heilbron, line 210: What is this testing? Does this validate that an invalid vhost is stripped from a hybrid gateway?

sam-heilbron (commented):
Couple small nits/questions

projects/gateway/pkg/reconciler/proxy_reconciler.go
bewebi: Good spot
The assumption that I'm making is that the "original" httpListener that precedes a "desired" httpListener within a hybridListener is that with the same matcher
I'm not sure whether this holds up, but I don't think there are better/more intuitive criteria

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
bewebi: This particular test demonstrates that valid listeners are updated
Since we add a HybridListener to the testing Proxy, we have an additional listener to update

test/e2e/gateway_test.go
```diff
@@ -53,6 +53,7 @@ var _ = Describe("Gateway", func() {
 		defaults.HttpPort = services.NextBindPort()
 		defaults.HttpsPort = services.NextBindPort()
 		defaults.TcpPort = services.NextBindPort()
+		defaults.HybridPort = defaults.HttpPort // TODO: why is this necessary?
```
MitchAman, line 56: Hmm nothing is coming to my mind. What happens if this isn't set?

bewebi: Good question, and good to document here
If `defaults.HybridPort` is set to `services.NextBindPort()` as the others are, we are unable to connect on that port

At the moment I'm getting a different timeout trying to run the tests locally, regardless of whether the workaround is used, so I don't have the relevant error message to post

MitchAman:
This LGTM, though I'll hold off on hitting approve to give docs team a chance to review

bewebi:
@solo-io/solo-docs weighed in on the Guide already via #5676 
I'm not sure if this PR also needs their sign off

MitchAman (approved):

PR #5632, opened 2021-11-24
HybridListener translation

bewebi:
# Description

Translate HybridListeners to Envoy proxy
- Add `matcherFilterChainTranslator` implementation of `FilterChainTranslator`
- Add `hybridRouteConfigurationTranslator` implementation of `RouteConfigurationTranslator`
- Add `ListenerSubsystemTranslatorFactory.GetHybridListenerTranslators` to return instances of the above
- Move `names.go` to `utils`, add `MatchedRouteConfigName()` and use it to identify matchers throughout
- Add unit tests and modify existing tests to account for HybridListeners

# Context

If merged before #5627, that PR will need to uncomment the TlsInspector test

Logic follows the example of existing translation flows as much as possible

Note that HybridListeners use Matchers as the source of truth for SSL config rather than the top level listener field, which is/will be deprecated


projects/gloo/pkg/translator/route_config.go
```diff
@@ -468,6 +469,57 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+
+		matchedListenerRouteConfigurationTranslator := &httpRouteConfigurationTranslator{
+			plugins: h.plugins,
+			proxy:   h.proxy,
+
+			parentListener: h.parentListener,
+			listener:       httpListener,
+
+			parentReport: h.parentReport,
+			report:       h.report.GetMatchedListenerReports()[utils.MatchedRouteConfigName(h.parentListener, matcher)].GetHttpListenerReport(),
```
sam-heilbron, line 506: The instance has a routeConfigName variable, but then we use a different function to create the route config name, what is the routeConfigName instance variable used for?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -468,6 +469,57 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+
+			routeConfigName:          rcName,
+			requireTlsOnVirtualHosts: matcher.GetSslConfig() != nil,
+		}
+		virtualHosts := matchedListenerRouteConfigurationTranslator.computeVirtualHosts(params)
```
sam-heilbron, line 511: Why do we only call compute virtualHosts and not use the public ComputeRouteConfiguration?

bewebi: We don't want to [validate the parent listener's SSL config](https://github.com/solo-io/gloo/blob/b2a1097c6f31d4df760e6c2d81b37f59727ab92e/projects/gloo/pkg/translator/route_config.go#L70-L75), which should be ignored if it exists, or use the [parent's route config name](https://github.com/solo-io/gloo/blob/b2a1097c6f31d4df760e6c2d81b37f59727ab92e/projects/gloo/pkg/translator/route_config.go#L78), since we need a different rcName per matchedListener

Issue #5620, opened 2021-11-23
 Use generated Clone() functions for Snapshopt clone


EItanya:
Currently when cloning snapshots, Gloo uses `proto.Clone()`. However, generated Clone() functions were recently added to [`protoc-gen-ext`](https://github.com/solo-io/protoc-gen-ext/blob/0d0ed8040ffdf7902f7f7a13a3420274e04a0cd9/extproto/ext.proto#L24) so we should test that those are faster, and if so use them instead.

kdorosh:
some helpful steps for profiling control plane components can be found in the pr description [here](https://github.com/solo-io/gloo/pull/2658)

sam-heilbron:
re-open for backports

chrisgaun:
Putting In Progress until we backport

PR #5658, opened 2021-12-01
feat(proto): added `clone_all` interfaces to all proto objects

gunnar-solo:
# Description
feat(proto): added `clone_all` interfaces to all proto objects
* excluded `health_check.proto` from generation process
* upgraded to solo-kit v0.24.2
* added test case to indicate difference in code generation (nil --> empty slice)

closes https://github.com/solo-io/gloo/issues/5620
# Context

Not a _bug fit_ per se, but a speedup/efficiency gain.  Tests run as a part of this [solo kit PR](https://github.com/solo-io/solo-kit/pull/455)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5620


sam-heilbron (commented):
Looks good! Small nit about a comment. Can we also add a description for what this PR is, why we're adding hte Clone methods, and why that test changed? (For the test change we can link to the https://github.com/solo-io/protoc-gen-ext issue that tracks the difference in clone functionality)

projects/gloo/cli/pkg/cmd/create/upstream_test.go
```diff
@@ -163,13 +163,13 @@ var _ = Describe("Upstream", func() {
 		It("should create kube upstream with default namespace and port", func() {
 			err := testutils.Glooctl("create upstream kube --name kube-upstream --kube-service kube-service")
 			Expect(err).NotTo(HaveOccurred())
-			expectKubeUpstream("kube-service", "default", uint32(80), nil)
+			expectKubeUpstream("kube-service", "default", uint32(80), map[string]string{})
```
sam-heilbron, line 166: Cross posting so other have context on this:

This test changed because the Clone() functionality for this resource changed. Our InMemory resource clients (used by this test) rely on Clone() to avoid data races: https://github.com/solo-io/solo-kit/blob/master/pkg/api/v1/clients/memory/resource_client.go#L140

Since the behavior of the Clone methods changed, we needed to update the test as well.

The following code can be copy/pasted to run and witness the differing behavior.

The historic clone functionality (https://github.com/protocolbuffers/protobuf-go/blob/master/proto/merge.go#L55) returns the Zero type for empty resources, meaning that a type with an empty slice becomes a type with a nil map when it is cloned.

The new clone functionality (https://github.com/solo-io/protoc-gen-ext/pull/14/files#diff-c79b8f443100751834a73bc767d8a720adfb1b79dba5ffe794885c1e962fcef8R59) takes a non-nil slice and clones it, even if it is empty.

The result is that the old clone treats an empty slice and nil slice identically, but the new clone treats them differently: 

```go
// This is the type we manually create (a kube upstream spec with an empty map of selectors)
emptySliceType := &kubernetes.UpstreamSpec{
  Selector: make(map[string]string),
}

// this is the type we read after we create the above type (a kube upstream spec with a nil map)
nilSliceType := &kubernetes.UpstreamSpec{
  Selector: nil,
}

// Our old implementation uses protobuf.Clone to clone a value
// This demonstrates that the old method treated empty and nil as interchangeable (and nillable)
// compare protobuf.Clone(type)
Expect(proto.Clone(emptySliceType).(*kubernetes.UpstreamSpec).Selector).To(BeNil())
Expect(proto.Clone(nilSliceType).(*kubernetes.UpstreamSpec).Selector).To(BeNil())

// Proto.Clone uses the Zero type, and the Zero type for both of these are identical
emptyZeroType := emptySliceType.ProtoReflect().Type().Zero().Interface()
nilZeroType := nilSliceType.ProtoReflect().Type().Zero().Interface()
Expect(emptyZeroType).To(Equal(nilZeroType))
Expect(emptyZeroType).To(BeNil())


// Our new implementation uses type.Clone() to clone a value
// This demonstrates that the new method no longer treats empty and nil as interchangeable
// compare type.Clone()
Expect(emptySliceType.Clone().(*kubernetes.UpstreamSpec).Selector).To(BeEquivalentTo(make(map[string]string)))
Expect(nilSliceType.Clone().(*kubernetes.UpstreamSpec).Selector).To(BeNil())
```

sam-heilbron (approved):

Issue #5502, opened 2021-10-25
Rejected RouteTable prevents updates for other RouteTables on same tree


lgadban:
**Describe the bug**
When delegating control for a given prefix to route tables via selection, if a selected RT contains a prefix mismatch, any other routes defined in selected RTs will not be processed. In other words, if a single route anywhere in the tree for a given prefix is `Rejected`, users can not update existing routes or add new routes to that node. This may prevent all updates to the VS (?) -- I haven't tested this piece, which could be even more impactful

**To Reproduce**
1. Create VS
```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: petstore
  namespace: gloo-system
spec:
  virtualHost:
    domains:
    - '*'
    routes:
    - matchers:
       - prefix: '/foo'
      delegateAction:
        selector:
          labels:
            routes: petstore
          namespaces:
          - gloo-system
```

2. Create RT for /foo
```
apiVersion: gateway.solo.io/v1
kind: RouteTable
metadata:
  name: petstore-routes-foo
  namespace: gloo-system
  labels:
    routes: petstore
spec:
  routes:
  - matchers:
    - prefix: /foo
    routeAction:
      single:
        upstream:
          name: default-petstore-8080
          namespace: gloo-system
```

3. Create RT with incorrect prefix
```
apiVersion: gateway.solo.io/v1
kind: RouteTable
metadata:
  name: petstore-routes
  namespace: gloo-system
  labels:
    routes: petstore
spec:
  routes:
  - matchers:
    - prefix: /
    routeAction:
      single:
        upstream:
          name: default-petstore-8080
          namespace: gloo-system
```

4. Modify existing good RT to use a directResponse
```
apiVersion: gateway.solo.io/v1
kind: RouteTable
metadata:
  name: petstore-routes-foo
  namespace: gloo-system
  labels:
    routes: petstore
spec:
  routes:
  - matchers:
    - prefix: /foo
    directResponseAction:
      status: 200
      body: "Hello, world!"
```

curl against /foo:
```
% curl localhost:8080/foo/
{"code":404,"message":"path /foo/ was not found"}
```

Should be direct response, but get routed to petstore still.

**Expected behavior**
The rejected RT should be sanitized/removed and other updates from other RTs should work fine.

**Additional context**
GlooE 1.8.15

sam-heilbron:
Duplicate of: https://github.com/solo-io/gloo/issues/5154

[...]

PR #5668, opened 2021-12-03
Fix for rejected RT preventing updates for other RTs

jenshu:
Treat RouteTable parent-child matcher errors as warnings during gateway translation, so that updates to other RTs can still  occur even when a bad RT is applied.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5502


projects/gateway/pkg/syncer/setup_syncer.go
```diff
@@ -143,6 +143,12 @@ func Setup(ctx context.Context, kubeCache kube.SharedCache, inMemoryCache memory
 		}
 	}
 
+	warnOnDelegateMatcherErrors := false
```
sam-heilbron, line 146: nit: we could probably just set this to ettings.GetGloo().GetInvalidConfigPolicy().GetReplaceInvalidRoutes()

projects/gateway/pkg/syncer/setup_syncer.go
```diff
@@ -160,6 +166,7 @@ func Setup(ctx context.Context, kubeCache kube.SharedCache, inMemoryCache memory
 		},
 		DevMode:                       true,
 		ReadGatewaysFromAllNamespaces: settings.GetGateway().GetReadGatewaysFromAllNamespaces(),
+		WarnOnDelegateMatcherErrors:   warnOnDelegateMatcherErrors,
```
sam-heilbron, line 169: What is the effect of always treating this as a warning, whether or not invalid route repalcement is enabled?

jenshu: from a quick test, it seems to behave the same way whether route replacement is enabled or not. but i'm also noticing that hitting the bad endpoint is not returning the expected error message (when route replacement is enabled). going to dig into that further

projects/gateway/pkg/translator/converter.go
```diff
@@ -515,13 +517,7 @@ func validateAndMergeParentRoute(child *gatewayv1.Route, parent *routeInfo) (*ga
...
 	// Merge options from parent routes
 	// If an option is defined on a parent route, it will override the child route's option
-	merged, err := mergeRouteOptions(child.GetOptions(), parent.options)
-	if err != nil {
-		// Should never happen
-		return nil, errors.Wrapf(err, "internal error: merging route options from parent to delegated route")
-	}
-
-	child.Options = merged
+	child.Options = mergeRouteOptions(child.GetOptions(), parent.options)
```
sam-heilbron, line 520: I've wanted to change this for so long, thanks for doing it!

projects/gateway/pkg/translator/converter.go
```diff
@@ -43,23 +43,23 @@ var (
 	DelegationCycleErr = func(cycleInfo string) error {
 		return errors.Errorf("invalid route: delegation cycle detected: %s", cycleInfo)
 	}
-	InvalidRouteTableForDelegatePrefixErr = func(delegatePrefix, prefixString string) error {
+	InvalidRouteTableForDelegatePrefixWarning = func(delegatePrefix, prefixString string) error {
```
sam-heilbron, line 46: nit: I'm torn between naming this with the warning suffix vs error suffix. I had assumed this was suffixed with Error because of it's type, and not necessarily because Gloo Edge treats this type of Error as a Reports Error. Open to thoughts either way.

projects/gateway/pkg/translator/merge_options.go
```diff
@@ -9,13 +9,13 @@ import (
...
-func mergeRouteOptions(dst, src *v1.RouteOptions) (*v1.RouteOptions, error) {
+func mergeRouteOptions(dst, src *v1.RouteOptions) *v1.RouteOptions {
 	if src == nil {
-		return dst, nil
+		return dst
 	}
 
 	if dst == nil {
-		return proto.Clone(src).(*v1.RouteOptions), nil
+		return proto.Clone(src).(*v1.RouteOptions)
```
sam-heilbron, line 18: We don't need to do this now, but given https://github.com/solo-io/gloo/pull/5658, we may also be able to convert these Clone methods to be use the faster implementation. Just commenting here for transparency, I don't think there's anything to be done for now

sam-heilbron (commented):
This looks great! Were you able to get to the bottom of where these warnings are processed during the translation and why we never reach the invalid route replacement step?

projects/gateway/pkg/translator/converter.go
jenshu: yeah, i didn't make this change at first, but then saw that [route_table_selector](https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/translator/route_table_selector.go) uses the `Warning` naming convention. (that's what produces the warnings where the [other addWarning](https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/translator/converter.go#L242) is used in converter.go)

jenshu:
> This looks great! Were you able to get to the bottom of where these warnings are processed during the translation and why we never reach the invalid route replacement step?

yep, so the place in converter.go where we changed it from addError to addWarning, still skips processing the invalid route as before, so we end up only adding the valid routes to the virtualhost (so same behavior as before). but in the [proxy reconciler](https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/reconciler/proxy_reconciler.go#L170) since there are now only warnings rather than errors on the virtualhost, it does not strip the virtualhost from the proxy, and we are left with the virtualhost containing only the valid routes

projects/gateway/pkg/translator/converter.go
```diff
@@ -202,19 +202,21 @@ func (rv *routeVisitor) visit(
...
 			var err error
 			routeClone, err = validateAndMergeParentRoute(routeClone, parentRoute)
 			if err != nil {
-				reporterHelper.addError(resource.InputResource(), err)
+				// An error occurs here when a delegated route's matcher is incompatible with its parent route's
+				// matcher. If we were to add this as an error on the resource, it would cause the entire VirtualHost
+				// to get stripped from the Proxy during gateway translation sync, thereby preventing the Proxy from
+				// receiving updates to other valid routes within the same VirtualHost.
+				// Therefore, we treat these as warnings here, allowing other valid routes on the same VirtualHost to
+				// still pass through and get onto the Proxy.
```
kdorosh, line 218: 👏

kdorosh (approved):

Issue #5670, opened 2021-12-06
cliutil: GetIngressHost support more types of services.


Cluas:
**Is your feature request related to a problem? Please describe.**
The current `glooctl proxy url` only supports two ServiceTypes, `LoadBalancer` and `NodePort`, and it returns the wrong url when the ServiceType is `ClusterIP`

**Describe the solution you'd like**
I will submit a PR to add support for `ClusterIP`

PR #5671, opened 2021-12-06
cliutil: GetIngressHost support more types of services.

Cluas:
# Description
When I tested the hello world demo, I found that when I changed the serviceType of the gateway-proxy to `ClusterIP`, the `glooctl proxy url` returned an error url. See https://github.com/solo-io/gloo/issues/5670

# Context

pkg/cliutil/uri.go:GetIngressHost

# Checklist:


nfuden:
While I like the idea my current question is if we should emit an extra warning if the gateway proxy is of type ClusterIP.

According to k8s pkg it seems that ClusterIP is not a compatible idea for an IngressHost 

`	
// ServiceTypeClusterIP means a service will only be accessible inside the
// cluster, via the cluster IP.
`
https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4132

Cluas:
> 虽然我喜欢这个想法，但我目前的问题是，如果网关代理的类型为 ClusterIP，我们是否应该发出额外警告。
> 
> 根据 k8s pkg 似乎 ClusterIP 不是 IngressHost 的兼容想法
> 
> ` // ServiceTypeClusterIP means a service will only be accessible inside the // cluster, via the cluster IP. ` https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4132



> 虽然我喜欢这个想法，但我目前的问题是，如果网关代理的类型为 ClusterIP，我们是否应该发出额外警告。
> 
> 根据 k8s pkg 似乎 ClusterIP 不是 IngressHost 的兼容想法
> 
> ` // ServiceTypeClusterIP means a service will only be accessible inside the // cluster, via the cluster IP. ` https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4132



> While I like the idea my current question is if we should emit an extra warning if the gateway proxy is of type ClusterIP.
> 
> According to k8s pkg it seems that ClusterIP is not a compatible idea for an IngressHost
> 
> ` // ServiceTypeClusterIP means a service will only be accessible inside the // cluster, via the cluster IP. ` https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4132

I agree with this view, adding a hint would be better.

Cluas:
In fact, using ClusterIP defeats the purpose of being Ingress, so should we just report an error and tell the user not to do that?

Cluas:
Or allow the user to do so, but give a friendly reminder.  I don't know which is better. does anyone have any suggestions?

sam-heilbron:
/kick flake https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e590e863-2d71-4b85-ae14-2f416389a2c7?

nfuden:
So after chatting with Sam I think we have come down with the following statement.

Print the friendly error perhaps something like
`Warning: Potentially invalid proxy configuration service type of ClusterIP`
but still return the ip like you have here.

As a side idea at some point in the future we can also enhance glooctl check to fire a similar warning on clusterIP but that is out of scope for this given issue.

In other words if you add the error message to what you have then this looks like a great update!

pkg/cliutil/uri.go
```diff
@@ -91,16 +91,26 @@ func GetIngressHost(ctx context.Context, proxyName, proxyNamespace, proxyPort st
...
-	// gateway-proxy is an externally load-balanced service
-	if len(svc.Status.LoadBalancer.Ingress) == 0 || localCluster {
-		// assume nodeport on kubernetes
+	serviceType := svc.Spec.Type
+	if localCluster {
+		serviceType = v1.ServiceTypeNodePort
+	}
+	switch serviceType {
+	case v1.ServiceTypeClusterIP:
+		fmt.Println("Warning: Potentially invalid proxy configuration service type of ClusterIP")
```
sam-heilbron, line 100: Perhaps this is unnecessary since we have the PR threads as context, but I wonder if a comment in the code here would be valuable outlining why this is invalid configuration, and why we are still setting the host/port and not erroring. @nfuden thoughts?

nfuden: Im always a fan of more comments but in this case since its a cli warning I tend to think we should make sure that the log line we are emitting has at least 90% of the context as otherwise we will have users who will wonder the same question.

In that case I propose  the following if it works for you Sam
```suggestion
		// There are a few edge cases where glooctl could be run in an environment where this is not a fatal error
		// However the service type ClusterIP does not accept incoming traffic which doesnt work as a ingress
		logger := GetLogger()
		logger.Write([]byte(""Warning: Potentially invalid proxy configuration, proxy may not accepting incoming connections"))
```

I am thinking this because then we can pull out that warning in the future to be a const so that we eventually support multi language and reuse of the invalid proxy configuration line.

sam-heilbron (approved):

sam-heilbron (approved):

nfuden (approved):

PR #5676, opened 2021-12-07
Hybrid Gateway guide

bewebi:
# Description

Add a guide that demonstrates use of Hybrid Gateways to match to a particular Http Gateways based on Client IPs
This is the explicit request from #5324

# Context

I run Gloo on `kind`, so the `glooctl proxy url` trick doesn't work for me
In order to `curl` the proxy I use the CI `deploy-to-kind-cluster.yaml` script with additional values to expose the proxy ports in Docker, and then port-forward


docs/content/introduction/traffic_management/_index.md
```diff
@@ -71,7 +71,7 @@ The listeners on a gateway typically listen for HTTP requests coming in on a spe
...
 
-Gloo Edge is meant to serve as an abstraction layer, simplifying the configuration of the underlying Envoy proxy and adding new functionality. The advanced options on Envoy are not exposed by default, but they can be accessed by adding an `httpGateway` section to your listener configuration. 
+Gloo Edge is meant to serve as an abstraction layer, simplifying the configuration of the underlying Envoy proxy and adding new functionality. The advanced options on Envoy are not exposed by default, but they can be accessed by adding an `httpGateway` or `hybridGateway` section to your listener configuration.
```
sam-heilbron, line 74: I'm torn on referencing the hybridGateway here. I think the paragraph (and subsequent config) is outlining how an HttpGateway can be configured for some advanced envoy functionality. The hybridGateway doesn't provide anything new/relevant to this config, it just exposes a way to configure this (and more).

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,200 @@
+---
+title: Hybrid Gateway
+weight: 10
+description: Define multiple HTTP or TCP Gateways within a single Gateway CRD
+---
+
+Hybrid Gateways allow you to define multiple HTTP or TCP Gateways for a single Gateway CRD with distinct matching criteria.
```
sam-heilbron, line 7: I think it would be valuable to add more background on what problem hybrid gateway's solve. Why would I benefit from using one? I think outlining that a HybridGateway gives you all the functionality of the existing HttpGateway and TcpGateway plus the ability to dynamically opt into each of them based on criteria (that we can explicitly list out) will help users understand the situations they might want to leverage one.

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,200 @@
...
+
+---
+
+## Only accept requests from a particular CIDR range
+
+Hybrid Gateways allow us to treat traffic from particular IPs differently.
```
sam-heilbron, line 13: Perhaps this is unnecessary, but I think we could provide even more context around why we would want to treat different IPs differently. ie we could outline a situation where 
```
A certain service wishes to enforce security best practices for most connections. For example, only allow TLS >= 1.2, and a curated list of ciphers. Unfortunately, some legacy clients have not yet migrated, and they must still use TLS1.1. We want to allow TLS1.1 connections only from these clients. Crucially these originate from the same domain, so the most trivial client discriminator that we can have is the source IP.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,200 @@
...
+  useProxyProto: false
+status: # collapsed for brevity
+{{< /highlight >}}
+
+Note: We use a range of 0.0.0.0/1 in order to have a high chance of matching the client's IP without knowing it specifically. A different and/or narrower range may be used if we know more about the client's IP.
+
+This results in a proxy that looks like:
+
+```yaml
+apiVersion: gloo.solo.io/v1
```
sam-heilbron, line 80: Given that this is an intermediate resource, do we need to expose this to the user? If so, is there specific config that is relevant to them that we can highlight or direct them to?

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
sam-heilbron: I also think we could speak more about how this HybridGateway allows users more control over the FilterChainMatching.

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Only accept requests from a particular CIDR range
+
+Hybrid Gateways allow us to treat traffic from particular IPs differently.
+
+In this example we will demonstrate how to only allow requests from one IP to reach an upstream while short-circuiting all other IPs with a direct response action.
+
+We will pick up where the [Hello World guide]({{< versioned_link_path fromRoot="/guides/traffic_management/hello_world" >}}) leaves off.
+
+To start we will add a second VirtualService that also matches all requests and has a directResponseAction:
```
sam-heilbron, line 19: Can we either update the comment 'matches all requests' to clarify that it's the `all-pets` endpoint, or just update the VS to match all paths?

sam-heilbron (commented):
Looks good! Left a couple comments, happy to discuss more

docs/content/introduction/traffic_management/_index.md
bewebi: better to have something like
```
but they can be accessed by adding an `httpGateway` section to your listener configuration, including as part of a `hybridGateway`.
```
or just remove the reference entirely?

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
bewebi: Modeled on [the tcp proxy guide](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/listener_configuration/tcp_proxy/), but I suppose isn't really relevant here
If you think it can go, I'm on board

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
bewebi: I updated it everywhere but the VS definition 🤦‍♂️

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+title: Hybrid Gateway
+weight: 10
+description: Define multiple HTTP or TCP Gateways within a single Gateway CRD
+---
+
+Hybrid Gateways allow users to define multiple HTTP or TCP Gateways for a single Gateway CRD with distinct matching criteria. 
+
+---
+
+Hybrid Gateways provide all of the functionality of HTTP and TCP Gateways with the added ability to dynamically select which Gateway a given request is routed to based on request properties.
```
Rachael-Graham, line 11:
```suggestion
Hybrid gateways expand the functionality of HTTP and TCP gateways by letting you use request properties to choose which gateway the request routes to.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
Rachael-Graham: +1 i think some more context might be handy

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+Selection is done based on `Matcher` fields, which map to a subset of Envoy [`FilterChainMatch`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchainmatch) fields.
+
+## Only accept requests from a particular CIDR range
+
+Hybrid Gateways allow us to treat traffic from particular IPs differently.
+One case where this might come in handy is if a set of clients are at different stages of migrating to TLS >=1.2 support, and therefore we want to enforce different TLS requirements depending on the client.
+If the clients originate from the same domain, it may be necessary to dynamically route traffic to the appropriate Gateway based on source IP.
+
+In this example we will demonstrate how to only allow requests from one IP to reach an upstream while short-circuiting all other IPs with a direct response action.
```
Rachael-Graham, line 20:
```suggestion
In this example, we will allow requests only from one IP to reach an upstream, while short-circuiting requests from all other IPs by using a direct response action.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+
+We will pick up where the [Hello World guide]({{< versioned_link_path fromRoot="/guides/traffic_management/hello_world" >}}) leaves off.
```
Rachael-Graham, line 22: just for some visual variety to signal where the actual task starts
```suggestion
**Before you begin**: Complete the [Hello World guide]({{< versioned_link_path fromRoot="/guides/traffic_management/hello_world" >}}) demo setup.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+            - name: client-ip-reject
+              namespace: gloo-system
+        matcher: {}
+  proxyNames:
+  - gateway-proxy
+  useProxyProto: false
+status: # collapsed for brevity
+{{< /highlight >}}
+
+Note: We use a range of 0.0.0.0/1 in order to have a high chance of matching the client's IP without knowing it specifically. A different and/or narrower range may be used if we know more about the client's IP.
```
Rachael-Graham, line 80:
```suggestion
Note: The range of 0.0.0.0/1 provides a high chance of matching the client's IP without knowing the specific IP. If you know more about the client's IP, you can specify a different, narrower range.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+    name: listener-::-8080
+    useProxyProto: false
+  - bindAddress: '::'
+    bindPort: 8443
+    httpListener: {}
+    metadata: # collapsed for brevity
+status: # collapsed for brevity
+```
+
+We can make a request to the proxy and will find that we get the `200` response:
```
Rachael-Graham, line 140:
```suggestion
Make a request to the proxy, which returns a `200` response because the client IP address matches to the 0.0.0.0/1 range:
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+
+```bash
+$ curl "$(glooctl proxy url)/all-pets"
+[{"id":1,"name":"Dog","status":"available"},{"id":2,"name":"Cat","status":"pending"}]
+```
+
+Also observe that if we make a request to an endpoint not matched by the `default` VirtualService we get a `404` response and _do not_ hit the `client-ip-reject` VirtualService:
```
Rachael-Graham, line 147:
```suggestion
Note that a request to an endpoint that is not matched by the `default` VirtualService returns a `404` response, and the request _does not_ hit the `client-ip-reject` VirtualService:
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+```bash
+$ curl -i "$(glooctl proxy url)/foo"
+HTTP/1.1 404 Not Found
+date: Tue, 07 Dec 2021 17:48:49 GMT
+server: envoy
+content-length: 0
+```
+This is because the `Matcher`s in the `HybridGateway` determine which `MatchedGateway` a request will be routed to, regardless of what routes that gateway has.
+
+### Observe that request from unmatched IP hits catchall gateway
```
Rachael-Graham, line 157:
```suggestion
### Route requests from non-matching IPs to a catchall gateway 
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+$ curl -i "$(glooctl proxy url)/foo"
+HTTP/1.1 404 Not Found
+date: Tue, 07 Dec 2021 17:48:49 GMT
+server: envoy
+content-length: 0
+```
+This is because the `Matcher`s in the `HybridGateway` determine which `MatchedGateway` a request will be routed to, regardless of what routes that gateway has.
+
+### Observe that request from unmatched IP hits catchall gateway 
+If we update the matcher to have a specific IP range that our client's IP is not a member of, we will expect our request to miss the matcher and fall through to the catchall gateway which is configured to respond `403`.
```
Rachael-Graham, line 158:
```suggestion
Next, update the matcher to use a specific IP range that our client's IP is not a member of. Requests from this client IP will now skip this matcher, and will instead match to a catchall gateway that is configured to respond with `403`.
```

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+        matcher: {}
+  proxyNames:
+  - gateway-proxy
+  useProxyProto: false
+status: # collapsed for brevity
+{{< /highlight >}}
+
+The Proxy will update accordingly.
+
+We can now make a request to the proxy and will find that we get the `403` response for any endpoint:
```
Rachael-Graham, line 193:
```suggestion
Make a request to the proxy, which now returns a `403` response for any endpoint:
```

Rachael-Graham (commented):
docs team review

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
Rachael-Graham: I think something should be added in this top section too about how this allows you to expose multiple HTTP and/or TCP gateways on a single port, seems important to highlght

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
bewebi: The two lines below were meant to add that context
Is there more that can/should be added?

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
+---
+title: Hybrid Gateway
+weight: 10
+description: Define multiple HTTP or TCP Gateways within a single Gateway CRD
```
sam-heilbron, line 4: Let's either refer to this as a Gateway CR or just Gateway. The CRD is the definition of what a Gateway can be (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) whereas the CR (https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) is what a specific gateway is

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+---
+
+Hybrid Gateways allow users to define multiple HTTP or TCP Gateways for a single Gateway CRD with distinct matching criteria.
```
sam-heilbron, line 7: same as above: let's use Gateway CR or just Gateway

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+      - matchers:
+          - prefix: /
+        directResponseAction:
+          status: 403
+          body: "client ip forbidden\n"
+EOF
+```
+
+
+Next let's update the existing `gateway-proxy` Gateway CRD, replacing the default `httpGateway` with a [`hybridGateway`]({{< versioned_link_path fromRoot="/reference/api/github.com/solo-io/gloo/projects/gateway/api/v1/gateway.proto.sk/#hybridgateway" >}}) as follows:
```
sam-heilbron, line 47: crd -> cr

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
sam-heilbron: Yeah I think yaml should provide guidance for an action to take, and since there is no action taken as a result of this Proxy yaml, I think we should just remove it.

docs/content/introduction/traffic_management/_index.md
sam-heilbron: I'm leaning towards just removing, since I feel like the hybridGateway refernece might confuse people who are just looking for the basic functionality of an http gateway

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
```diff
@@ -0,0 +1,205 @@
...
+
+In this example, we will allow requests only from a particular CIDR range to reach an upstream, while short-circuiting requests from all other IPs by using a direct response action.
+
+**Before you begin**: Complete the [Hello World guide]({{< versioned_link_path fromRoot="/guides/traffic_management/hello_world" >}}) demo setup.
+
+To start we will add a second VirtualService that also matches all requests and has a directResponseAction:
+
+```yaml
+kubectl apply -n gloo-system -f - <<EOF
+apiVersion: gateway.solo.io/v1
```
sam-heilbron, line 28: This VS is only used in the second example, should we move the resource definition closer to the steps that rely on it?

sam-heilbron (commented):
last few nits, LGTM

docs/content/guides/traffic_management/listener_configuration/hybrid_gateway.md
bewebi: Even though we don't request it til later, we refer to it in the next step when we edit/create the Hybrid Gateway
While we could make a Hybrid Gateway with just the one matched gateway consisting of the petstore VS, I think it's meaningful to show that the routing is making a deterministic selection between multiple Gateways

sam-heilbron:
LGTM

PR #5693, opened 2021-12-10
use newer kind node image (1.21.1)

jenshu:
Use kindest/node v1.21.1 in tests and scripts


.github/workflows/regression-tests.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
         KUBE2E_TESTS: ${{ matrix.kube-e2e-test-type }}
         USE_XDS_RELAY: ${{ matrix.xds-relay }}
         CLUSTER_NAME: 'kind'
-        CLUSTER_NODE_VERSION: 'v1.17.17@sha256:66f1d0d91a88b8a001811e2f1054af60eef3b669a9a74f9b6db871f2f1eeed00'
+        CLUSTER_NODE_VERSION: 'v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6'
```
sam-heilbron, line 72: Can we also upgrade the kubectl version we use a few lines above?

sam-heilbron: Also, while we're here, can we just change the version we inject into valid semver?

sam-heilbron:
Closing in favor of: https://github.com/solo-io/gloo/pull/5725

Issue #3964, opened 2020-12-08
Publish metrics for invalid resources


asayah:
**Is your feature request related to a problem? Please describe.**
When Gloo isn't able to process an object (a Virtual Service, ...), it updates its status to provide information about the error, but there's no way to be alerted

**Describe the solution you'd like**
Publish metrics in Prometheus to allow users to create alerts when it occurs

Metrics that show if a configuration is ok/missing/wrong that can be consumed using Prometheus for example. 
use case: Alerting based on these metrics 

**Additional Context**
Duplicate of: https://github.com/solo-io/gloo/issues/3866

asayah:
Metric per CRD type, for example 
virtualservice.foo = 1 
upstream.bar=0
...

sam-heilbron:
I think the ask here is straightforward (publish metrics for resources so users can alert on those metrics) and useful (automate alerting on out of sync resources instead of relying on glooctl). I think we should start by Spiking out a solution, gaining feedback from field engineers and clients before implementing.

elcasteel:
Write up of options for formatting metrics: https://soloio.slab.com/posts/metrics-for-invalid-resources-spike-o9b4u7xg
@djannot @asayah do either of you have thoughts on whether a count of the invalid resources of each type would work for our expected use cases?

bcollard:
quick notice on this other contribution https://github.com/solo-io/solo-projects/blob/master/install/helm/gloo-ee/dashboards/gloo-overview.json (internal ref) which exposes many control-plane metrics in a new Grafana dashboard. It's a good start to set up alerting on invalid resources. It doesn't answer the request made here.

PR #5710, opened 2021-12-15
Add metrics around resources with invalid configuration

MitchAman:
# Description

Provides the ability to record metrics tracking the configuration status of various resource types. Metrics are not recorded by default. Metrics can be enabled for individual resource types, with additional configuration for applying labels to each individual resource.

For example, the following configuration enables status metric reporting for upstreams and virtual services, with the resource names and namespaces used as labels:
```yaml
observabilityOptions:
  configStatusMetricLabels:
    Upstream.v1.gloo.solo.io:
      labelToPath:
        name: '{.metadata.name}'
        namespace: '{.metadata.namespace}'
    VirtualService.v1.gateway.solo.io:
      labelToPath:
        name: '{.metadata.name}'
        namespace: '{.metadata.namespace}'
```
And produces the following data on the gateway and gloo metrics endpoints:
```
validation_gateway_solo_io_upstream_config_status{name="default-petstore-8080",namespace="gloo-system"} 0
```
```
validation_gateway_solo_io_virtual_service_config_status{name="default",namespace="gloo-system"} 0
```

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3964


projects/gloo/api/v1/settings.proto
```diff
@@ -385,6 +385,26 @@ message Settings {
...
         // Options to configure Gloo's integration with [Kubernetes](https://www.kubernetes.io/).
         GrafanaIntegration grafanaIntegration = 1;
+
+        // Provides the ability to specify which label(s) to apply to the
+        // `*_config_status` metric for each resource type.
+        message ConfigStatusMetricsOptions {
+            // Defines which labels to apply for a metric.
+            message MetricLabels {
+                // Each (key, value) pair in the map defines a label to be applied.
+                //   key:   Specifies the name of the label (i.e. "namespace")
```
kdorosh, line 395: super nit,
```suggestion
                //   key:   Specifies the name of the label (e.g. "namespace")
```
instead of i.e. in both places

projects/gloo/api/v1/settings.proto
```diff
@@ -385,6 +385,26 @@ message Settings {
...
+                //   value: Specifies the jsonpath string corresponding to the field on the resource (i.e. "metadata.namespace")
+                map<string, string> label_to_path = 1;
+            }
+
+            // Labels for the virtual_service_config_status metric. If unspecified
+            // or the map is empty, then the virtual_service_config_status metric
+            // will not be recorded.
+            MetricLabels virtual_service_labels = 1;
```
kdorosh, line 403: any reason for the nested message instead of just using a `map<string, string>` here? what kinds of things are we anticipating adding here?

MitchAman: Yes, we'll be adding `MetricLabels gateway_service_labels = 2`, `MetricLabels route_table_labels = 3`, and so on

sam-heilbron: Could we instead have a type that defines the GVK for a resource (https://github.com/solo-io/gloo/blob/master/install/helm/gloo/crds/gateway.solo.io_v1_VirtualService.yaml#L9) instead of having the explcit types mentioend in the API?

MitchAman: We could. Is there a way to represent that as a single string? I feel like this yaml is going to real messy real quick

This would also open the door for users to specify a resource type that we don't validate, and therefore can't record these metrics on. That's easy enough to detect though, so we could fail loudly in that case

projects/gateway/pkg/validation/metricutils/resource_metrics.go
```diff
@@ -0,0 +1,204 @@
+package metricutils
```
nfuden, line 1: how would you feel about calling this package metrics then we can simplify many of these names by removing the metrics prefix.

projects/gateway/pkg/validation/metricutils/resource_metrics.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	output := strings.Join(valueStrings, ",")
+	return output, nil
+}
+
+// Returns a resourceMetric, or nil if labelToPath is nil or empty. An error is returned if the
+// labelToPath configuration is invalid (for example, specifies an invalid label key).
+func newResourceMetric(gvk schema.GroupVersionKind, labelToPath map[string]string) (*resourceMetric, error) {
+	numLabels := len(labelToPath)
+	if numLabels > 0 {
```
nfuden, line 187: nit: Reduce nesting by if numLabels == 0{ return nil, nil}

projects/gateway/pkg/validation/metricutils/resource_metrics.go
```diff
@@ -0,0 +1,204 @@
...
+// resourceMetric is a wrapper around a gauge. In addition to a gauge itself, it stores information
+// regarding which labels should get applied to it, and how to obtain the values for those labels.
+type resourceMetric struct {
+	gauge       *stats.Int64Measure
+	labelToPath map[string]string
+}
+
+// NewConfigStatusMetrics creates and returns a ConfigStatusMetrics from the specified options.
+// If the options are invalid, an error is returned.
+func NewConfigStatusMetrics(opts map[string]*MetricLabels) (*ConfigStatusMetrics, error) {
```
nfuden, line 53: Why are we using a map of pointers here?

nfuden (commented):
The general approach seems reasonable. I think my biggest question becomes whether this is the right place for this particular package.

projects/gateway/pkg/validation/metricutils/resource_metrics.go
MitchAman: Because that's the way they come out of the generated code from the proto

projects/gloo/api/v1/settings.proto
```diff
@@ -386,6 +386,21 @@ message Settings {
 
         // Options to configure Gloo's integration with [Kubernetes](https://www.kubernetes.io/).
         GrafanaIntegration grafanaIntegration = 1;
+
+
+        message MetricLabels {
+            // Each (key, value) pair in the map defines a label to be applied. Keys specify the name of the label
```
elcasteel, line 392: I don't understand this comment. Can you clarify what it means for "a label to be applied" and possibly give a more concrete example of a key-value pair and the resulting metrics?

projects/gateway/pkg/utils/metrics.go
```diff
@@ -0,0 +1,204 @@
...
+var MetricNames = map[schema.GroupVersionKind]string{
+	gwv1.VirtualServiceGVK: "validation.gateway.solo.io/virtual_service_config_status",
+	gwv1.GatewayGVK:        "validation.gateway.solo.io/gateway_config_status",
+	gwv1.RouteTableGVK:     "validation.gateway.solo.io/route_table_config_status",
+	gloov1.UpstreamGVK:     "validation.gateway.solo.io/upstream_config_status",
+	gloov1.SecretGVK:       "validation.gateway.solo.io/secret_config_status",
+}
+
+var metricDescriptions = map[schema.GroupVersionKind]string{
+	gwv1.VirtualServiceGVK: "TODO", // TODO(mitchaman)
```
elcasteel, line 31: Are you going to fill these in as part of this PR or in the future?

elcasteel:
The code looks good, I had some documentation related comments.

projects/gateway/pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+		return schema.GroupVersionKind{}, errors.Errorf("config status metric reporting is not supported for resource type: %T", resource)
+	}
+}
+
+func (m *ConfigStatusMetrics) SetResourceStatus(ctx context.Context, resource resources.Resource, status *core.Status) {
+	if status.GetState() == core.Status_Warning || status.GetState() == core.Status_Rejected {
+		m.SetResourceInvalid(ctx, resource)
+		return
+	}
+	m.SetResourceValid(ctx, resource)
```
sam-heilbron, line 115: This willl affect both 'pending' and 'accepted' (https://github.com/solo-io/solo-kit/blob/master/api/v1/status.proto#L28) statuses. Is that what we want?

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
 
 More information on Gloo Edge's admin ports can be found [here]({{% versioned_link_path fromRoot="/introduction/observability/#grafana-and-prometheus" %}}).
 
+### Monitoring the Status of Gloo Resources
+
+When Gloo Edge fails to process a resource, we see that reflected in the resource's {{< protobuf name="core.solo.io.Status" display="Status">}}.
+Gloo Edge can also be configured to publish metrics which record the health status of resources.
```
Rachael-Graham, line 117:
```suggestion
You can configure Gloo Edge to publish metrics that record the configuration status of resources.
```

changed health to configuration bc that's what we call it in the reference doc

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
...
+
+The status metrics can be enabled by specifying a resource type, along with the label(s) that should be applied to the metric, within the
+`observabilityOptions` in the Settings CRD. Consider the following, which will add metrics for virtual services and
```
Rachael-Graham, line 120:
```suggestion
In the `observabilityOptions` of the Settings CRD, you can enable status metrics by specifying the resource type and any labels to apply
to the metric. Consider the following, which will add metrics for virtual services and
```

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
...
+      labelToPath:
+        name: '{.metadata.name}'
+        namespace: '{.metadata.namespace}'
+    VirtualService.v1.gateway.solo.io:
+      labelToPath:
+        name: '{.metadata.name}'
+        namespace: '{.metadata.namespace}'
+```
+
+After completing the [Hello World]({{% versioned_link_path fromRoot="/guides/traffic_management/hello_world/" %}})
```
Rachael-Graham, line 136:
```suggestion
After completing the [Hello World guide]({{% versioned_link_path fromRoot="/guides/traffic_management/hello_world/" %}}) 
```

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
...
+
+After completing the [Hello World]({{% versioned_link_path fromRoot="/guides/traffic_management/hello_world/" %}}) 
+to generate some resources, the metrics defined above can be found at: <http://localhost:9091/metrics>. If the port
+forwarding is directed towards the gloo pod, the `default-petstore-8080` upstream can be seen in a healthy state:
+```
+validation_gateway_solo_io_upstream_config_status{name="default-petstore-8080",namespace="gloo-system"} 0
+```
+
+If the port forwarding is then switched to the gateway pod, the metric defined for virtual services can be found
+on the same metrics endpoint (after a refresh):
```
Rachael-Graham, line 144: a refresh of what?

projects/gloo/api/v1/settings.proto
```diff
@@ -386,6 +386,21 @@ message Settings {
...
+            // (e.g. "namespace"). Values specify the jsonpath (https://kubernetes.io/docs/reference/kubectl/jsonpath/)
+            // string corresponding to the field of a resource to use as the label value (e.g. "{.metadata.namespace}").
+            map<string, string> label_to_path = 1;
+        }
+
+        // Provides the ability to record metrics tracking the configuration status of various resource types.
```
Rachael-Graham, line 398:
```suggestion
        // Enable metrics that track the configuration status of various resource types.
```

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
 
 More information on Gloo Edge's admin ports can be found [here]({{% versioned_link_path fromRoot="/introduction/observability/#grafana-and-prometheus" %}}).
 
+### Monitoring the Status of Gloo Resources
```
Rachael-Graham, line 114:
```suggestion
### Monitoring the configuration status of Gloo resources
```

projects/gloo/api/v1/settings.proto
```diff
@@ -386,6 +386,21 @@ message Settings {
...
+        // Each (key, value) pair in the map defines a metric for a particular resource type. Configuration status
+        // metrics are not recorded by default; metrics will only be recorded for the resources specified in this map.
```
Rachael-Graham, line 400:
```suggestion
        // metrics are not recorded by default; metrics are recorded only for the resources specified in this map.
```

projects/gloo/api/v1/settings.proto
```diff
@@ -386,6 +386,21 @@ message Settings {
...
+        // Keys specify the resource type (GroupVersionKind) whose status is to be tracked
```
Rachael-Graham, line 401:
```suggestion
        // Keys specify the resource type (GroupVersionKind) to track for status changes
```

Rachael-Graham (commented):
Docs review

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
...
+
+When Gloo Edge fails to process a resource, we see that reflected in the resource's {{< protobuf name="core.solo.io.Status" display="Status">}}.
+Gloo Edge can also be configured to publish metrics which record the health status of resources.
+
+The status metrics can be enabled by specifying a resource type, along with the label(s) that should be applied to the metric, within the
+`observabilityOptions` in the Settings CRD. Consider the following, which will add metrics for virtual services and
+upstreams, both having labels that include the namespace and name of each individual resource:
+
+```yaml
+observabilityOptions:
```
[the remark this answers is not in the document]
MitchAman: We don't do any handling or even concern ourselves with the semantics of the labels selected, we simply grab the field out of the resource and stick it onto the metric.

If `{.metadata.namespace}` is used as a label value in the "two installations, separate namespaces" scenario, and both installations are writing a status metric for the same underlying resource, then both installations would use the same values for the labels. The installations could, however, have different opinions on the status of the resource. The label values relate to fields on the resource, and don't have anything to do with the installation. This code would simply look at the resource, grab whatever value is in `resource.metadata.namespace` and stick that onto the metric.

In this scenario, would each installation have their own prometheus server? IIUC, when setting metrics, they're just set in memory until something comes by, scrapes them, and persists them somewhere long-term. In order to avoid two installations writing metrics over top of one another, whoever is collecting the metrics is the person who is responsible for keeping the two separated.  This applies for all metrics, not just statuses.

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -343,11 +346,12 @@ func (s *statusSyncer) syncStatus(ctx context.Context) error {
 		}
 		if err := s.reporter.WriteReports(ctx, reports, currentStatuses); err != nil {
 			errs = multierror.Append(errs, err)
-		} else {
-			// cache the status written by the reporter
-			status := s.reporter.StatusFromReport(subresourceStatuses, currentStatuses)
-			localInputResourceLastStatus[inputResource] = status
+			continue
```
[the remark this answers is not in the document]
MitchAman: It was intentional, but I'm not married to the idea of it. I'm conflicted between the two schools of thought:
1. If we didn't actually write the new status, then it's simply not the status. The metric reflects the status of a resource, so if the resource's status doesn't change, the metric doesn't change. We'll catch it the next time around. Versus,
2. We know what the status _should_ be, and that's what's important. Just because a solar flare prevented us from writing the status doesn't mean we should pretend the status never changes. It's not the _true_ state of the system

WDYT?

projects/gateway/pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+		return gloov1.SecretGVK, nil
+	case *gloov1.Upstream:
+		return gloov1.UpstreamGVK, nil
+	default:
+		return schema.GroupVersionKind{}, errors.Errorf("config status metric reporting is not supported for resource type: %T", resource)
+	}
+}
+
+func (m *ConfigStatusMetrics) SetResourceStatus(ctx context.Context, resource resources.Resource, status *core.Status) {
+	if status.GetState() == core.Status_Warning || status.GetState() == core.Status_Rejected {
```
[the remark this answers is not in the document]
MitchAman: IMO we're better off not bringing `allowWarnings` into the equation as it already has its own responsibilities over in validation. If we start tying separate isolated pieces of functionality together by a single config value, we're going to cause a lot of headaches and unintended behavior changes.

If we get an ask to allow for excluding `WARNING` from the status metrics we can create a new field within `ObservabilityOptions` to handle that. I think that by default `WARNING` should flip the status metric to invalid - the point of warnings are to catch attention :)

docs/content/guides/observability/prometheus/metrics.md
```diff
@@ -111,6 +111,42 @@ And then open <http://localhost:9091> for the admin page, including the Promethe
...
+  configStatusMetricLabels:
+    Upstream.v1.gloo.solo.io:
+      labelToPath:
```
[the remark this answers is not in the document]
MitchAman: The criteria for what we can set the label value on the metric to isn't all that restrictive: https://opencensus.io/tag/value/

I don't believe there's any restrictions on what's "valid" to pull out of the CR, as long as it meets the tag value criteria. Grabbing fields from the spec shouldn't be an issue. It really just comes down to what customers would find valuable to have. For the most part I would think namespace and name are enough to identify the resource experiencing issues

docs/content/guides/observability/prometheus/metrics.md
MitchAman: The page in a browser

MitchAman: I went ahead and rephrased it, let me know if that works for you.

Thanks for the review!

projects/gateway/pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+		log.Errorf("Unable to set config status valid for resource '%s': %s", resource.GetMetadata().Ref(), err.Error())
+		return
+	}
+	if m.metrics[gvk] != nil {
+		log.Debugf("Setting '%s' config metric valid", resource.GetMetadata().Ref())
+		mutators, err := getMutators(m.metrics[gvk], resource)
+		if err != nil {
+			log.Errorf("Error setting labels on %s: %s", Names[gvk], err.Error())
+		}
+		utils.MeasureZero(ctx, m.metrics[gvk].gauge, mutators...)
```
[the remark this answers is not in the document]
MitchAman: It does challenge general practices for things of this nature, but it is intentional. By setting valid = 0 and invalid = 1, all of the time series data can be summed to get a total of how many resources are invalid across the whole system.

projects/gateway/pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		utils.MeasureOne(ctx, m.metrics[gvk].gauge, mutators...)
+	}
+}
+
+func (m *ConfigStatusMetrics) insertMetric(gvk schema.GroupVersionKind, metric *resourceMetric) {
+	m.metrics[gvk] = metric
+}
+
+func getMutators(metric *resourceMetric, resource resources.Resource) ([]tag.Mutator, error) {
```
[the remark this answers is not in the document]
MitchAman: Good catch! I think this is definitely worthwhile to do, but there may be more than one approach. For instance, some JSONPath libraries may have caching features built in.

Seeing as this feature is disabled by default, mind if I file a separate issue and tackle that as a follow-up?

projects/gateway/pkg/utils/metrics/metrics.go
MitchAman: I believe all resources are valid until proven otherwise in the court of validation!

It's actually a small optimization to just not set the metric for `Pending` resources, since it will transition to a new state very quickly. I went ahead and made that change

projects/gateway/pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+	case *gloov1.Secret:
+		return gloov1.SecretGVK, nil
+	case *gloov1.Upstream:
+		return gloov1.UpstreamGVK, nil
+	default:
+		return schema.GroupVersionKind{}, errors.Errorf("config status metric reporting is not supported for resource type: %T", resource)
+	}
+}
+
+func (m *ConfigStatusMetrics) SetResourceStatus(ctx context.Context, resource resources.Resource, status *core.Status) {
```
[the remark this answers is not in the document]
MitchAman: If this feature hasn't been configured, then `SetResourceValid()` and `SetResourceInvalid()` will see that `m.metrics[gvk]` is nil, and then immediately return.

Because we support enabling this feature for individual resource types, we need to call `resourceToGVK()` to figure out what type the resource is.  We could probably have a condition outside of this file, which checks if the `settings.observabilityOptions.configStatusMetricLabels` map is empty, and never call `SetResourceStatus()` to begin with, but then we would need to do that check in many places. Seems cleaner and negligible to just keep that logic contained here - we would only save ourselves at most one switch statement

projects/gateway/pkg/utils/proxy_transition_function.go
```diff
@@ -32,6 +35,8 @@ func updateDesiredStatus(original, desired *v1.Proxy, statusClient resources.Sta
 	// until we make the full move.
 	equal, ok := hashutils.HashableEqual(original, desired)
 	if ok && equal {
-		statusClient.SetStatus(desired, statusClient.GetStatus(original))
+		status := statusClient.GetStatus(original)
+		statusClient.SetStatus(desired, status)
+		statusMetrics.SetResourceStatus(ctx, desired, status)
```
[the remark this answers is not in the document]
MitchAman: Yeah, doesn't look it's necessary here. Thanks

projects/gateway/pkg/syncer/translator_syncer.go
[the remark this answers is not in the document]
sam-heilbron: Yeah I think both approaches are reasonable. I'm more leaning towards the second option. In the end, the metric (and status) is intended as a vehicle to provide transparency around invalid config. Regardless of whether the status was persisted (transient network error or something), the intended status is known and will exist so we should notify the user asap.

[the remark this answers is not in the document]
MitchAman: Sounds reasonable to me - Done.

projects/gateway/pkg/utils/metrics/metrics.go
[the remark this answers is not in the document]
MitchAman: https://github.com/solo-io/gloo/issues/5794

projects/ingress/pkg/setup/setup_syncer.go
```diff
@@ -247,6 +253,9 @@ func RunIngress(opts Opts) error {
 			ingressClient := clusteringressv1alpha1.NewClusterIngressClientWithBase(baseClient)
 			clusterIngTranslatorEmitter := clusteringressv1.NewTranslatorEmitter(ingressClient)
 			statusClient := statusutils.GetStatusClientForNamespace(opts.StatusReporterNamespace)
+			if err != nil {
```
sam-heilbron, line 256: What is this conditional protecting?

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -24,13 +25,15 @@ type ProxyReconciler interface {
 
 type proxyReconciler struct {
 	statusClient   resources.StatusClient
+	statusMetrics  metrics.ConfigStatusMetrics
 	proxyValidator validation.GlooValidationServiceClient
 	baseReconciler gloov1.ProxyReconciler
 }
 
-func NewProxyReconciler(proxyValidator validation.GlooValidationServiceClient, proxyClient gloov1.ProxyClient, statusClient resources.StatusClient) *proxyReconciler {
+func NewProxyReconciler(proxyValidator validation.GlooValidationServiceClient, proxyClient gloov1.ProxyClient, statusClient resources.StatusClient, statusMetrics metrics.ConfigStatusMetrics) *proxyReconciler {
```
sam-heilbron, line 33: Is this update used anymore?

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -344,10 +347,12 @@ func (s *statusSyncer) syncStatus(ctx context.Context) error {
 		if err := s.reporter.WriteReports(ctx, reports, currentStatuses); err != nil {
 			errs = multierror.Append(errs, err)
 		} else {
-			// cache the status written by the reporter
+			// The inputResource's status was successfully written, update the cache and metric with that status
 			status := s.reporter.StatusFromReport(subresourceStatuses, currentStatuses)
 			localInputResourceLastStatus[inputResource] = status
 		}
+		status := s.reporter.StatusFromReport(subresourceStatuses, currentStatuses)
```
sam-heilbron, line 354: What does this line do? I see we have a line very similar to it in the else a few lines up

sam-heilbron (commented):
Small cleanup, looks good to go

projects/gateway/pkg/syncer/translator_syncer.go
MitchAman: It looks through the report to see if there is an error or warnings, and creates/returns a Status to reflect that

projects/ingress/pkg/setup/setup_syncer.go
MitchAman: no longer necessary and I removed it, thanks

projects/gateway/pkg/reconciler/proxy_reconciler.go
MitchAman: nope, removed.

gunnar-solo (approved):
I've...
* read through the changes
* observed metrics being hosted from a local build
* checked for other reviewer comments I consider blocking
With all of that said, I'm happy to say "do it!"

or, the more classical "LGTM"

PR #5724, opened 2021-12-17
Glooctl check eventually healthy gives more descriptive error

saiskee:
If the `glooctl check` fails in our e2e tests, we don't know why. This will provide a more descriptive error with the glooctl check error message.


jenshu (approved):

nfuden (approved):
Seems reasonable

Issue #5523, opened 2021-11-01
Kubernetes 1.22 has dropped support for extensions/v1beta1 APIs


pseudonator:
**Describe the bug**
Kubernetes 1.22 has [dropped support](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/) for both `extensions/v1beta1` and `networking.k8s.io/v1beta1`. APIs have been migrated to `networking.k8s.io/v1`.

When Gloo Edge is deployed in Ingress mode it fails due to the error,
```
{"level":"fatal","ts":"2021-10-30T08:41:08.653Z","logger":"ingress","caller":"setuputils/main_setup.go:73","msg":"error in setup: starting snapshot watch: initial Ingress list: listing ingressObjs in : the server could not find the requested resource (get ingresses.extensions)","version":"1.9.1","stacktrace":"github.com/solo-io/gloo/pkg/utils/setuputils.Main\n\t/workspace/gloo/pkg/utils/setuputils/main_setup.go:73\ngithub.com/solo-io/gloo/projects/ingress/pkg/setup.Main\n\t/workspace/gloo/projects/ingress/pkg/setup/setup.go:12\nmain.main\n\t/workspace/gloo/projects/ingress/cmd/main.go:9\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:225"}
```

**To Reproduce**
1. Run kube2e tests for ingress with `CLUSTER_NAME=solo-test-cluster CLUSTER_NODE_VERSION=v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047 VERSION=v1.0.0-solo-test ci/deploy-to-kind-cluster.sh && KUBE2E_TESTS=ingress make run-ci-regression-tests`
This should produce the error,
```
Fail handler msg Unexpected error:
    <*errors.StatusError | 0xc000596000>: {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "the server could not find the requested resource (post ingresses.extensions)",
            Reason: "NotFound",
            Details: {
                Name: "",
                Group: "extensions",
                Kind: "ingresses",
                UID: "",
                Causes: [
                    {
                        Type: "UnexpectedServerResponse",
[...]

Issue #5707, opened 2021-12-15
deprecate knative


jenshu:
**Is your feature request related to a problem? Please describe.**
Maintaining Knative support is becoming a burden, especially as not many users seem to be using it. We should add deprecation notices in the appropriate places in the docs.

**Describe the solution you'd like**
Deprecate knative in gloo v1.10 and remove it in v1.11

**Describe alternatives you've considered**
We tried [upgrading to a more recent knative version](https://github.com/solo-io/gloo/pull/5695) but ran into issues.

PR #5695, opened 2021-12-11
[DNM] Validate Knative changes

sam-heilbron:
# Do Not Merge
Just opened this to validate proposed changes, https://github.com/solo-io/gloo/pull/5693 is the PR that will add this functionality

# Description

Update default knative version that is installed

# Context

https://github.com/solo-io/gloo/pull/4796: We ran into challenges upgrading k8s versions in CI.

To upgrade the k8s version now (to validate kube 1.22 breaking changes) we need to upgrade the knative version we install.

We previously installed v0.10.0 but the latest version is v0.26.1. Between these versions, the manifests have been renamed. Therefore, we needed to update our install helper to find the proper manifests.

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #5708, opened 2021-12-15
Delete knative code


jenshu:
We are [deprecating knative support](https://github.com/solo-io/gloo/issues/5707) in the upcoming v1.10 release. In the release after that (v1.11), we should delete/clean up all the knative related code and docs.

PR #5725, opened 2021-12-18
Update all apis to v1

sam-heilbron:
Blocked by:
- [x] k8s-utils ephemeral containers supports kube 1.22 (https://github.com/solo-io/k8s-utils/pull/16)
- [x] Investigation into why validation server does not start up properly

# Description

Update CI to rely on Kubernetes 1.22 (client and server)

# Context

There are a number of changes in this PR. I opted to do them all at once since most are intertwined, but I try to spell out what each change is below and why we include it in the PR.

### Kind Upgrade
Our k8s client libraries have been upgrade to v0.22.x. We should have our server version used in CI match. This is particularly important both for general maintenance/upkeep but also because the version [removes a number of deprecated APIs](https://kubernetes.io/blog/2021/08/04/kubernetes-1-22-release-announcement/#removal-of-several-deprecated-beta-apis)  We had missed a couple of places where we still relied on these deprecated APIs, causing the bugs to occur for users when using kube 1.22.

### Kubectl Upgrade
I think its best to keep our client/server versions in sync. Upgrading kubectl caused some of our tests to fail. To resolve this, we needed to update our [k8s-utils library](https://github.com/solo-io/k8s-utils/pull/16) to handle creating [ephemeral containers in a pod](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/).

### Helm Upgrade
Our helm library has updated, so we should try to keep the CI version up to date as well. I did not notice any breaking changes when I updated this version.

### Istio Upgrade
In our `glooctl` tests we require istio to be installed on the cluster. Our [script for preparing a cluster](https://github.com/solo-io/gloo/blob/fbdf9e2fd10e8b19175b75ac43749943249a4482/ci/deploy-to-kind-cluster.sh#L91) is responsible for choosing the istio version to install. Per [istio support matrix](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases), we need to upgrade istio to 1.11 to work on a kube 1.22 cluster. 

### Knative Tests
In order to upgrade our kube version used in CI, we also need to update the knative version we support. We [attempted](https://github.com/solo-io/gloo/pull/5695) to do this briefly, however when it proved to be more than a day/two of work, we opted to [deprecate the feature](https://github.com/solo-io/gloo/issues/5707) since it is not actively used and has no active development in the last 2 years. 

The feature is deprecated, but we still have tests that will fail on kube 1.22. This PR comments out those tests so they are not executed as part of CI. It's not ideal, but it allows us to not EOL the feature too quickly, while also upgrading CI.

*Our branch protection rules expect the knative tests to succeed, yet this PR disables those tests from running. Therefore, it will be listed as pending. After this PR is approved, I will update the branch protection rules to not expect this test anymore, and that will allow bulldozer to merge this PR*

### Ingress API Changes
https://github.com/solo-io/gloo/issues/5523

As part of getting to 1.22, we need to upgrade some packages that are removed in 1.22. This work was [begun here](https://github.com/solo-io/gloo/pull/5697) and I opted to move it into this larger PR since they are closely related.

*Note: Ingress tests rely on a kube cluster, yet are run with our other e2e tests on cloudbuild. Therefore, we opted to move these tests to our regression test suite, which run against our kind clusters.*

### Validation Server Test Changes
Upgrading kube versions caused validation server tests to fail: specifically the test we had written to validate that an invalid upstream (Gloo resource) would be rejected. 

For Gloo resources to be validated, a Proxy must exist. Otherwise, there is no translation loop to run and the default behavior is to return a nil error. We noticed that for some reason, a Proxy was no longer available, so the Upstream in question was never being validated, and it was accepting the invalid upstream instead of rejecting it. It turned out the test was looking for the wrong error message (an invalid type error), so we updated the test to check for an error produced explicitly by the Gloo translation loop.

### Gateway Test Failures
When running on updated versions of kube client and server, our validation server was not starting up correctly. For one, the certgen job was terminating and it was unclear why. Without the job, we do not update the validation webhook config with the client CA Bundle, and no connection can be made between gloo and gateway. I noticed events in the cluster:

`Warning   FailedMount         pod/gateway-certgen--1-kmthl          MountVolume.SetUp failed for volume "kube-api-access-r2xw5" : object "gloo-system"/"kube-root-ca.crt" not registered`

This led me down a bit of a rabbit hole, I came across [this thread here](https://issueexplorer.com/issue/kubernetes/kubernetes/105204). Though ultimately since it's a warning it does not seem to be the root issue. This ConfigMap became a feature in 1.21 :

```
A ConfigMap containing a CA bundle used for verifying connections to the kube-apiserver. This feature depends on the RootCAConfigMap feature gate, which publishes a "kube-root-ca.crt" ConfigMap to every namespace. RootCAConfigMap feature gate is graduated to GA in 1.21 and default to true. (This flag will be removed from --feature-gate arg in 1.22)
```

and I confirmed that the ConfigMap did exist.

I got the certgen job to complete (and produce the CABundle) though the validation server still does not startup properly. 

*Outcome: This research was a rabbit hole. I did not find a resolution, and was able to get tests passing locally without fixing this. I'm unclear what led to these errors, since I don't think they were caused by test pollution, but since I have been unable to reproduce locally, I am happy to proceed without understanding this behavior.*

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5523


.github/workflows/regression-tests.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        kube-e2e-test-type: ['gateway', 'ingress', 'knative', 'helm', 'gloomtls', 'glooctl', 'eds']
+        kube-e2e-test-type: ['gateway', 'ingress', 'helm', 'gloomtls', 'glooctl', 'eds']
```
kdorosh, line 13: comment out and add reason for knative deprecation?

ci/deploy-to-kind-cluster.sh
```diff
@@ -4,7 +4,7 @@
 # The name of the kind cluster to deploy to
 CLUSTER_NAME="${CLUSTER_NAME:-kind}"
 # The version of the Node Docker image to use for booting the cluster
-CLUSTER_NODE_VERSION="${CLUSTER_NODE_VERSION:-v1.17.17@sha256:66f1d0d91a88b8a001811e2f1054af60eef3b669a9a74f9b6db871f2f1eeed00}"
+CLUSTER_NODE_VERSION="${CLUSTER_NODE_VERSION:-v1.22.4}"
```
kdorosh, line 7: shouldn't this match the kind version we use in `.github/workflows/regression-tests.yaml`?

cloudbuild.yaml
```diff
@@ -87,9 +87,11 @@ steps:
   dir: *dir
   id: 'install-go-tools'
 
-- name: 'gcr.io/$PROJECT_ID/e2e-go-mod-ginkgo:0.4.15'
+- name: 'gcr.io/solo-public/e2e-go-mod-ginkgo:0.0.0-dev'
```
kdorosh, line 90: why not leave `PROJECT_ID` since that is solo-public?

kdorosh: oh i see this is a WIP PR, probably updated for local testing.. will stop commenting for now

.github/workflows/regression-tests.yaml
sam-heilbron: https://github.com/solo-io/gloo/pull/5725/commits/f023719dc9e06df8d38a56308e111ad28d771f11

ci/deploy-to-kind-cluster.sh
sam-heilbron: Yeah for consistency that's probably best. This value is just the default, and the value in regression-tests is the value we use in CI (if empty, use default). So they don't need to match, but no reason not to

cloudbuild.yaml
sam-heilbron: Yeah I was jsut testing a dev build but it was unnecessary in the end. Reverted.

.github/workflows/regression-tests.yaml
```diff
@@ -10,7 +10,9 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        kube-e2e-test-type: ['gateway', 'ingress', 'knative', 'helm', 'gloomtls', 'glooctl', 'eds']
```
gunnar-solo, line 13: I see that later on in these file changes there are further preventative measures from running `knative` tests.

At what point would it be appropriate to entirely scrape the knative functionality from `gloo`?  Re:  dead code tends to cause headaches

gunnar-solo (commented):
a few scattered questions and nits

.github/workflows/regression-tests.yaml
nfuden: There is an issue linked later in the knative test

.github/workflows/regression-tests.yaml
```diff
@@ -60,22 +62,24 @@ jobs:
...
       with:
-        version: v3.2.0
+        version: v3.6.3
     - name: Setup test env
       env:
         KUBE2E_TESTS: ${{ matrix.kube-e2e-test-type }}
         USE_XDS_RELAY: ${{ matrix.xds-relay }}
         CLUSTER_NAME: 'kind'
-        CLUSTER_NODE_VERSION: 'v1.17.17@sha256:66f1d0d91a88b8a001811e2f1054af60eef3b669a9a74f9b6db871f2f1eeed00'
-        VERSION: 'v0.0.0-kind'
```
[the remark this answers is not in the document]
sam-heilbron: I think both the version with/without the prefixed v, will work. I had opted to remove the 'v', only because I found it slightly misleading. Our references to TAGGED_VERSION include the v prefix since they refer to the github tag (https://github.com/solo-io/gloo/releases/tag/v1.11.0-beta1) whereas our references to VERSION have tended to be that tag, minus the v. All in all, I don't think it changes the behavior.

.github/workflows/regression-tests.yaml
sam-heilbron: We have an issue https://github.com/solo-io/gloo/issues/5708 to track. I included the issue that deprecated knative, and that issue includes a reference to the "delete knative code" issue which is why I felt safe not including the delete issue directly. Do you think it would be clearer if I included that issue link as well? Happy to do so if you think it's valuable

cloudbuild.yaml
```diff
@@ -90,6 +90,8 @@ steps:
 - name: 'gcr.io/$PROJECT_ID/e2e-go-mod-ginkgo:0.4.15'
   entrypoint: make
   env:
+  - 'ACK_GINKGO_RC=true'
+  - 'ACK_GINKGO_DEPRECATIONS=1.16.5'
```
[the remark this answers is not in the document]
sam-heilbron: Good call! I can open one

projects/ingress/pkg/translator/translate.go
```diff
@@ -129,50 +128,55 @@ func upstreamForBackend(upstreams gloov1.UpstreamList, services []*kubev1.Servic
...
-		return nil, errors.Errorf("discovery failure: upstream not found for kube service %v with port %v", backend.ServiceName, backend.ServicePort)
+		return nil, errors.Errorf("discovery failure: upstream not found for kube service %v with port %v", serviceName, servicePort)
 	}
 	return matchingUpstream, nil
 }
 
-func getServicePort(services []*kubev1.Service, name, namespace string, servicePort intstr.IntOrString) (int32, error) {
-	if servicePort.Type == intstr.Int {
-		return servicePort.IntVal, nil
+func getServiceNameAndPort(services []*kubev1.Service, namespace string, ingressService *networkingv1.IngressServiceBackend) (string, int32, error) {
```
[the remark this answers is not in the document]
sam-heilbron: Good call, will update

.github/workflows/regression-tests.yaml
gunnar-solo: My bad!  Missed the issue!  Happy with what we already have

cloudbuild.yaml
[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/issues/5743

.github/workflows/regression-tests.yaml
sam-heilbron: All good! I appreciate the idea, and think it's important was stay vigilant about tracking deprecated features / dead code

ci/deploy-to-kind-cluster.sh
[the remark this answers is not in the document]
sam-heilbron: Yeah to be honest I wasn't sure about this change. I had noticed that in previous CI runs (https://github.com/solo-io/gloo/runs/4613620640?check_suite_focus=true), even supplying the digest still results in the particular tag being used to install:

```
+ activeClusters=
+ [[ '' =~ .*kind.* ]]
+ kind create cluster --name kind --image kindest/node:v1.17.17@sha256:66f1d0d91a88b8a001811e2f1054af60eef3b669a9a74f9b6db871f2f1eeed00 --config=-
+ cat
Creating cluster "kind" ...
 • Ensuring node image (kindest/node:v1.17.17) 🖼  ...
 ✓ Ensuring node image (kindest/node:v1.17.17) 🖼
```

Since this is the default value we use, I could ensure the env variable used in CI specifies the digest so we maintain that behavior. What do you think?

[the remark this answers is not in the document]
sam-heilbron: I opted to leave the default value as the tag, and specify the digest for CI

gunnar-solo (approved):

PR #5729, opened 2021-12-20
api: Http only cookie setting

yuval-k:


ci/tools.go
```diff
@@ -21,6 +21,7 @@ import (
 	_ "github.com/cratonica/2goarray"
 	_ "github.com/envoyproxy/protoc-gen-validate"
 	_ "github.com/golang/mock/mockgen"
+	_ "github.com/onsi/ginkgo/ginkgo"
```
nfuden, line 24: This seems unrelated. Why do we now need to import ginkgo for side effects?

yuval-k: good question; make codegen didn't work for me without it. it could be because i'm using go 1.17

nfuden (approved):

PR #5735, opened 2021-12-21
Graphql - Using directives to attach resolver defs to fields

saiskee:
Previously, resolvers were defined like
```yaml
resolutions:
- matcher:
    fieldMatcher:
      type: Query
      field: author
    restResolver: 
      ... # resolver definition here
```

to match a resolver to a field in graphql schema.

Now we can just name the resolver:
```yaml
resolutions:
   authorResolver:
      restResolver:
         .... # resolver definition here
```

and in the SDL we can use a directive to attach the resolver to a field instead of using a field matcher on the resolver:

```gql
type Query {
  author: String @resolve(name: "authorResolver")
}

type Book {
  author: String @resolve(name: "authorResolver") # yes, they're reusable too!
}
```


changelog/v1.10.0-beta12/new-resolver-api.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/5718
+    resolvesIssue: false
+    description: |
+      (Enterprise-only) Refactor GraphQL API to allow for less verbosity when configuring resolvers.
```
saiskee, line 6: fix chagnelog dir and make this more descriptive.

saiskee: what changelog dir should this go to? I'm leaning towards making this a 1.11-beta but we are still currently releasing 1.10-rc

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -235,8 +220,26 @@ message ExecutableSchema {
 message Executor {
   // Execute schema using resolvers.
   message Local {
-    // The resolver map to use to resolve the schema.
-    repeated Resolution resolutions = 1;
+    /* Mapping of resolver name to resolver definition.
+     The names are used to reference the resolver in the graphql schema.
+     For example, a resolver with name "authorResolver" can be defined as
```
kdorosh, line 225: do we want to include anything about how to reference the name/namespace (ie ref) of the cr?

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -235,8 +220,26 @@ message ExecutableSchema {
...
+         upstreamRef: ...
+         request:
+           ...
+         response:
+           ...
+      ```
+      and referenced in the graphql schema as
+      ```gql
+      type Query {
+        author: String @resolve(name: "authorResolver")
```
kdorosh, line 238: if updated to include optional ref field, then we don't need to worry about global uniqueness of resolver names

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
saiskee: I was planning on this step introducing the directive only (aka getting rid of field matcher) and then doing the new CR in a follow up to minimize PR size blowup.

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
saiskee: see https://github.com/solo-io/gloo/pull/5735#discussion_r773485597

kdorosh (approved):

kdorosh:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=47666326-b2d2-426f-a67b-0ff84be503bc

PR #5748, opened 2021-12-30
add license key to gloo and discovery deployments

saiskee:
# Description
Some enterprise plugins require the gloo license key. This PR allows the license key to be available to gloo and discovery deployments with the default gloo edge enterprise installation.


kdorosh (dismissed):

changelog/v1.11.0-beta2/add-license-env-var.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
```
kdorosh, line 1: this should be beta3

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -167,6 +167,13 @@ spec:
           mountPath: /etc/gloo
           readOnly: true
         env:
+{{- if .Values.license_secret_name }}
```
kdorosh, line 170: we'd need some docs changes if we did it this way too, since the subchart would have a different helm value that needs to be set.

kdorosh (changes requested):
i think we should change the implementation a bit so we only have to configure this once

install/helm/gloo/templates/1-gloo-deployment.yaml
kdorosh: to improve the UX and only require configuring once, I think we should leverage this to inject the yaml here https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L1-L2

docs/content/operations/upgrading/1.10.0.md
```diff
@@ -0,0 +1,73 @@
+---
```
saiskee, line 1: Putting this under 1.10 because this license work is going to be backported to 1.10.

changelog/v1.11.0-beta3/add-license-env-var.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Allows for GLOO_LICENSE_KEY env var to be automatically populated from generated license secret in the gloo and discovery pods.
+  - type: HELM
+    issueLink: https://github.com/solo-io/gloo/issues/5751
+    resolvesIssue: false
+    description: `license_secret_name` should now be under `gloo.license_secret_name` in the helm `values.yaml` file if it is used.
```
kdorosh, line 7: let's call this non user facing and only mark it in the enterprise changelog?

install/helm/gloo/templates/3-discovery-deployment.yaml
```diff
@@ -1,3 +1,6 @@
+{{- if .Values.gateway.updateValues -}}
+{{- include "gloo.updatevalues" . -}}
+{{- end -}}
```
kdorosh, line 3: we don't need these anymore

install/helm/gloo/templates/1-gloo-deployment.yaml
kdorosh: after discussing offline, this wouldnt work unless `license_secret_name` was in a global helm value.

`kubeResourceOverrides` don't help

if https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors don't help, then i think the current approach of updating the helm value used for enterprise installs (`.Values.license_secret_name` to `.Values.gloo.license_secret_name`) is the appropriate solution

kdorosh: discussed further, yaml anchors look like they will work, so we should be able to remove the extra upgrade steps

kdorosh (changes requested):
edit: ignore this; i was conflating `license_secret_name` and `license_key`

~we need to update `glooctl` too~ https://github.com/solo-io/gloo/blob/38ee93876c12273e15b1b8e0cb49ae05435a2ff8/projects/gloo/cli/pkg/cmd/install/enterprise.go#L24

projects/gloo/cli/pkg/cmd/install/enterprise.go
```diff
@@ -1,49 +1,49 @@
...
-	return cmd
+  cmd := &cobra.Command{
+    Use:    "enterprise",
+    Short:  "install the Gloo Enterprise Gateway on Kubernetes",
+    Long:   "requires kubectl to be installed",
+    PreRun: setVerboseMode(opts),
+    RunE: func(cmd *cobra.Command, args []string) error {
+
+      extraValues := map[string]interface{}{
+        "gloo.license_key": opts.Install.LicenseKey,
```
kdorosh, line 24: this should be
```golang
        "gloo": {"license_key": opts.Install.LicenseKey},
```

kdorosh (approved):

PR #5786, opened 2022-01-12
Resolve HybridListener marshal panic

sam-heilbron:
# Description

Ensure that invalid configuration defined on a HybridListener does not cause the translation loop to panic

# Context

## How did we notice this?
Our CI pipeline failed: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=2ae522e9-cca1-4d87-9257-36deb08a1258

## What is the bug?
During the generation of the FilterChain, we convert a user defined SslConfig, into the definition for a FilterChainMatch. If the SslConfig is invalid (for example, an invalid secret reference) we should catch the error, write it to a report, skip the config and continue processing.

We do this for HttpListeners: https://github.com/solo-io/gloo/blob/fccf272cb170ba4f1531013bacaa915d98e2e99d/projects/gloo/pkg/translator/filter_chain.go#L99

However, we did not do this for HybridListeners: https://github.com/solo-io/gloo/blob/fccf272cb170ba4f1531013bacaa915d98e2e99d/projects/gloo/pkg/translator/filter_chain.go#L265. Instead we were processing the error, writing it to a report and then processing that config. It would then panic when trying to marshal a nil value: https://github.com/solo-io/gloo/blob/fccf272cb170ba4f1531013bacaa915d98e2e99d/projects/gloo/pkg/translator/filter_chain.go#L274

## What is the fix?
We identify the invalid config, and do not create a FilterChain if the config is invalid. This mirrors the behavior of our HttpListener.

## PluginRegistry Changes
In earlier PRs, we introduce the concept of a PluginRegistry, as a way to process plugins ahead of time, and provide specific types to translators. Previously each translator would have to process all plugins and decide for itself which implemented the required interface. In this PR, we extend the usage of the registry to other translators that were still iterating over all plugins.

# Checklist:

### DIFF
--- a/projects/gloo/pkg/translator/listener_subsystem.go
+++ b/projects/gloo/pkg/translator/listener_subsystem.go
@@ -21,42 +21,39 @@ import (
 // The ListenerSubsystemTranslatorFactory returns a ListenerTranslator and RouteConfigurationTranslator for a given Gloo Listener
 type ListenerSubsystemTranslatorFactory struct {
 	pluginRegistry      plugins.PluginRegistry
-	proxy               *v1.Proxy
 	sslConfigTranslator utils.SslConfigTranslator
 }
 
 func NewListenerSubsystemTranslatorFactory(
 	pluginRegistry plugins.PluginRegistry,
-	proxy *v1.Proxy,
 	sslConfigTranslator utils.SslConfigTranslator,
 ) *ListenerSubsystemTranslatorFactory {
 	return &ListenerSubsystemTranslatorFactory{
 		pluginRegistry:      pluginRegistry,
-		proxy:               proxy,
 		sslConfigTranslator: sslConfigTranslator,
 	}
 }
 
-func (l *ListenerSubsystemTranslatorFactory) GetTranslators(ctx context.Context, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
+func (l *ListenerSubsystemTranslatorFactory) GetTranslators(ctx context.Context, proxy *v1.Proxy, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
 	ListenerTranslator,
 	RouteConfigurationTranslator,
 ) {
 	switch listener.GetListenerType().(type) {
 	case *v1.Listener_HttpListener:
-		return l.GetHttpListenerTranslators(ctx, listener, listenerReport)
+		return l.GetHttpListenerTranslators(ctx, proxy, listener, listenerReport)
 
 	case *v1.Listener_TcpListener:
 		return l.GetTcpListenerTranslators(ctx, listener, listenerReport)
 
 	case *v1.Listener_HybridListener:
-		return l.GetHybridListenerTranslators(ctx, listener, listenerReport)
+		return l.GetHybridListenerTranslators(ctx, proxy, listener, listenerReport)
 	default:
 		// This case should never occur
 		return &emptyListenerTranslator{}, &emptyRouteConfigurationTranslator{}
 	}
 }
 
-func (l *ListenerSubsystemTranslatorFactory) GetHttpListenerTranslators(ctx context.Context, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
+func (l *ListenerSubsystemTranslatorFactory) GetHttpListenerTranslators(ctx context.Context, proxy *v1.Proxy, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
 	ListenerTranslator,
 	RouteConfigurationTranslator,
 ) {
@@ -106,8 +103,8 @@ func (l *ListenerSubsystemTranslatorFactory) GetHttpListenerTranslators(ctx cont
 	// of configuration for the HttpConnectionManager, we only produce
 	// a single RouteConfiguration
 	routeConfigurationTranslator := &httpRouteConfigurationTranslator{
-		plugins:                  l.pluginRegistry.GetPlugins(),
-		proxy:                    l.proxy,
+		pluginRegistry:           l.pluginRegistry,
+		proxy:                    proxy,
 		parentListener:           listener,
 		listener:                 listener.GetHttpListener(),
 		parentReport:             listenerReport,
@@ -152,7 +149,7 @@ func (l *ListenerSubsystemTranslatorFactory) GetTcpListenerTranslators(ctx conte
 	return listenerTranslator, routeConfigurationTranslator
 }
 
-func (l *ListenerSubsystemTranslatorFactory) GetHybridListenerTranslators(ctx context.Context, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
+func (l *ListenerSubsystemTranslatorFactory) GetHybridListenerTranslators(ctx context.Context, proxy *v1.Proxy, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
 	ListenerTranslator,
 	RouteConfigurationTranslator,
 ) {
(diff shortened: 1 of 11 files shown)


projects/gloo/pkg/translator/listener_subsystem.go
```diff
@@ -21,42 +21,39 @@ import (
...
 ) *ListenerSubsystemTranslatorFactory {
 	return &ListenerSubsystemTranslatorFactory{
 		pluginRegistry:      pluginRegistry,
-		proxy:               proxy,
 		sslConfigTranslator: sslConfigTranslator,
 	}
 }
 
-func (l *ListenerSubsystemTranslatorFactory) GetTranslators(ctx context.Context, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
+func (l *ListenerSubsystemTranslatorFactory) GetTranslators(ctx context.Context, proxy *v1.Proxy, listener *v1.Listener, listenerReport *validationapi.ListenerReport) (
```
sam-heilbron, line 37: I modified this factory. It can now live for the duration of a Translator instance, as opposed to regenerating one for each process we process. To do this, I removed the proxy from the constructor and pass it in as an argument

projects/gloo/pkg/translator/listener_subsystem.go
```diff
@@ -106,8 +103,8 @@ func (l *ListenerSubsystemTranslatorFactory) GetHttpListenerTranslators(ctx cont
 	// of configuration for the HttpConnectionManager, we only produce
 	// a single RouteConfiguration
 	routeConfigurationTranslator := &httpRouteConfigurationTranslator{
-		plugins:                  l.pluginRegistry.GetPlugins(),
-		proxy:                    l.proxy,
+		pluginRegistry:           l.pluginRegistry,
```
sam-heilbron, line 106: We previously iterated over all plugins, processing some. Passing the plugin registry allows the translator to extract just the relevant plugins.

projects/gloo/pkg/translator/listener_subsystem.go
```diff
@@ -177,13 +174,33 @@ func (l *ListenerSubsystemTranslatorFactory) GetHybridListenerTranslators(ctx co
...
 	}
 
-	routeConfigurationTranslator := &hybridRouteConfigurationTranslator{
-		plugins:        l.pluginRegistry.GetPlugins(),
-		proxy:          l.proxy,
-		parentListener: listener,
-		listener:       listener.GetHybridListener(),
-		parentReport:   listenerReport,
-		report:         hybridListenerReport,
+	// For each HttpListener defined on the HybridListener,
```
sam-heilbron, line 177: I just copy/pasted this logic from the implementation of the translator, to the constructor.

projects/gloo/pkg/translator/route_config.go
```diff
@@ -64,21 +64,9 @@ type httpRouteConfigurationTranslator struct {
 func (h *httpRouteConfigurationTranslator) ComputeRouteConfiguration(params plugins.Params) []*envoy_config_route_v3.RouteConfiguration {
 	params.Ctx = contextutils.WithLogger(params.Ctx, "compute_route_config."+h.routeConfigName)
 
-	virtualHosts := h.computeVirtualHosts(params)
-
-	// validate ssl config if the parent listener specifies any and is not a hybrid listener
-	if h.parentListener.GetHybridListener() == nil {
```
sam-heilbron, line 70: This logic is unnecessary. The purpose is to process ssl config, and validate it. However, the outcome is that we append an error on the report and still generate the route configuration. We perform this same check later on, append the same error, so I removed this duplicate reference.

https://github.com/solo-io/gloo/blob/fccf272cb170ba4f1531013bacaa915d98e2e99d/projects/gloo/pkg/translator/filter_chain.go#L97

projects/gloo/pkg/translator/route_config.go
```diff
@@ -278,7 +262,7 @@ func (h *httpRouteConfigurationTranslator) setAction(
 
 		// DirectResponseAction supports header manipulation, so we want to process the corresponding plugin.
 		// See here: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#route-directresponseaction
-		for _, plug := range h.plugins {
+		for _, plug := range h.pluginRegistry.GetRoutePlugins() {
```
sam-heilbron, line 265: This one is a bit weird. We only process the HeaderPlugin, not all RoutePlugins. I wasn't sure if it was safe to process all RoutePlugins, so I kept the behavior.The small optimization is to iterate over RoutePlugins (Header plugin is one of them) instead of iterating over all plugins

projects/gloo/pkg/translator/route_config.go
```diff
@@ -350,23 +330,23 @@ func (h *httpRouteConfigurationTranslator) runRoutePlugins(
...
 			validation.AppendRouteError(routeReport,
 				validationapi.RouteReport_Error_ProcessingError,
-				fmt.Sprintf("%T: %v", routePlugin, err.Error()),
+				fmt.Sprintf("%T: %v", plugin, err.Error()),
 				out.GetName(),
 			)
 		}
 	}
 
+	if in.GetRouteAction() == nil || out.GetRoute() == nil {
```
sam-heilbron, line 339: Just pulled this conditional out of the for loop

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a4d2b718-6464-4127-9079-071496e084fb (it should run with cors, potential test flake)

projects/gloo/pkg/translator/translator.go
```diff
@@ -66,33 +66,42 @@ func (t *translatorFactory) Translate(
...
 ) (envoycache.Snapshot, reporter.ResourceReports, *validationapi.ProxyReport, error) {
-
+	// setup tracing, logging
 	ctx, span := trace.StartSpan(params.Ctx, "gloo.translator.Translate")
-	params.Ctx = ctx
 	defer span.End()
+	params.Ctx = contextutils.WithLogger(ctx, "translator")
 
-	params.Ctx = contextutils.WithLogger(params.Ctx, "translator")
+	// re-initialize plugins on each loop, this is done for 2 reasons:
```
elcasteel, line 100: :tada:

elcasteel (dismissed):

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=1d128d4b-b75f-4fca-9f3b-3e6f62a83f84

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=9c244dc8-a06c-4427-8da6-1068855758d0

another flake. I don't think the test is flakey, as much as there is something else at play here. Too many tests that have never flaked before are starting to

nfuden:
It looks like its mostly the slow tests? Does this perhaps point to some old magic numbers that are no longer sufficient due to some extra overhead slowing gloo down?

sam-heilbron:
> It looks like its mostly the slow tests? Does this perhaps point to some old magic numbers that are no longer sufficient due to some extra overhead slowing gloo down?

In any particular tests? The regression test failure shouldn't be time based since it's trying to perform and operation and expecting it to fail (but it succeeds). The e2e test I'm going to look further into, becuase the CORS test has failed multiple times, perhaps I introduced a regression

sam-heilbron:
https://github.com/solo-io/gloo/pull/5786/commits/961f416e9df4ce7dd70e8d5cdc531468c6f0af63 🤦 helps to process the VirutualhostPlugins, if I want the config they generate...

elcasteel (approved):

Issue #4020, opened 2020-12-23
Improve UX around Enterprise-only features


Sodman:
**Is your feature request related to a problem? Please describe.**
Currently, when users try to use enterprise-only Gloo Edge features when running the Open Source edition of Gloo Edge, there are no warnings or errors. The configuration of enterprise features is just silently ignored when running the OS version of Gloo Edge. This is largely because the protos defined in OS Gloo Edge are a superset of the Gloo APIs, including features from both OS and EE versions of Gloo Edge.

**Describe the solution you'd like**
Ideally we'd have a way to at least put warnings on resources that are configured to use Enterprise-only features but are running in an environment without Gloo Edge Enterprise installed.

**Describe alternatives you've considered**
We want to avoid splitting the protos into two sets, one for EE and one for OS, as it's a much bigger maintenance burden and a big potential source of bugs.

Sodman:
Re-opening, only partially fixed for wasm, not rate-limit or enterprise for example.

PR #5803, opened 2022-01-20
Update Plugin Processing UX

sam-heilbron:
Blocked by:
- [x] https://github.com/solo-io/solo-projects/pull/2980 PR that introduces corresponding changes to solo-projects

# Description

Update the plugin interface, and the way that enterprise plugins are injected. 

# Context

### Upgradable Plugin
I removed this interface. The purpose of this plugin was to flag a plugin as having an enterprise equivalent that would override it. This made the plugin a little leaky, since it was aware of the enterprise application.

I exposed a Name on each Plugin, ensuring that in Solo-Projects we can still perform the same Upgrade reconciliation, relying on the Plugin.Name() to identify which plugins to upgrade.

### Consistent Plugin Layout (Purely Cosmetic)
Plugins did not have a consistent format. I chose one, and ensured that all of them follow this format. This makes it easier to reason through the plugins. I was changing aspects of each plugin to remove the Upgradable plugin interface, so this felt like an appropriate time to make this cosmetic change as well.

### Enterprise Plugin Errors (UX)
We had performed some work a while back to ensure that if an open source user configured Gloo with enterprise config that they could see errors on those resources easily. The previous solution to propagate enterprise errors was based on the Upgradable Plugin solution. 

We had an existing bug, [as pointed out here](https://github.com/solo-io/gloo/issues/4020#issuecomment-953957183), that allowed many enterprise features to go unused, but without a helpful error message. This was because we never loaded the open source plugin shims into the registry, and thus they were never processed.

To avoid creating a placeholder shim for every enterprise plugin, I opted to create a single enterprise_warning plugin. It performs the same validation, but does so in a single place.

### Load Enterprise Plugins from Solo-Projects via Extensions
Gloo has always used `Extensions` as a mechanism for providing extension points that the enterprise application can leverage. For plugins, we previously exposed a list of plugins, and then reconciled those plugins inside of Gloo. This meant that open source Gloo owned the logic for reconciling plugins defined by an outside source. This felt somewhat backwards to me. 

I changed this to allow for the extension point to be a PluginRegistryFactory. This is the function that returns a plugin registry for each translation run. Open source Gloo can load the relevant plugins into the registry. Enterprise Gloo can make decisions around licenses, and availabe open source gloo plugins, and load the plugins it cares about, in the order it decides into the registry. Now Solo-Projects defines the logic to register plugins that it is potentially overriding in open source gloo.

### Only Load Filters If Used
This will be completed in a follow up PR. 

I want to write a test that validates that the open source plugin registry factory produces plugins which, when processed, don't add filters to the filter chain unnecessarily. This will be used as a template for the solo-projects plugin registry factory.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/4020
resolves https://github.com/solo-io/gloo/issues/4367


sam-heilbron:
Added work in progress just to make sure I get CI in solo-projects passing/confirmed before it merges. Good to be reviewed / approved since the lable will block merges

bewebi (dismissed):

bewebi (approved):

projects/gloo/pkg/plugins/enterprise_warning/plugin.go
```diff
@@ -0,0 +1,242 @@
...
+	AdvancedHttpExtensionName          = "advanced_http"
+	DlpExtensionName                   = "dlp"
+	FailoverExtensionName              = "failover"
+	JwtExtensionName                   = "jwt"
+	LeftmostXffAddressExtensionName    = "leftmost_xff_address"
+	ProxyLatencyExtensionName          = "proxy_latency"
+	RbacExtensionName                  = "rbac"
+	SanitizeClusterHeaderExtensionName = "sanitize_cluster_header"
+	WafExtensionName                   = "waf"
+	WasmExtensionName                  = "wasm"
```
kdorosh, line 30: not a huge fan that this list is leaking enterprise implementation details, and needs to be updated as we add new enterprise features (easily forgotten)

perhaps we can instead determine this more dynamically using the following assumptions:
- if the config is valid schematically (lean on openapi schema) then it is supported api
- if at the end of translation, none of our plugins modified any of the `out` resources for a given `in` resource, we can assume this is an enterprise feature that is not supported

thoughts?

kdorosh (commented):
really excited about these changes, just want to be a little more clever with any leaking implementation details

projects/gloo/pkg/plugins/enterprise_warning/plugin.go
nfuden: If we were to do that then we would likely want a more generic warning right? Not that its a bad thing but if I am correct then this approach would detect other unused options which may be intended to be consumed by a customer made extension or for a plugin that is implemented in a later gloo version. In fact whether my above statement is true or not I like the idea of validating that we arent dropping any of the data being supplied in the our gloo custom resources.

kdorosh: we spoke offline, our proposal with @sam-heilbron is to:

- add a new proto annotation for enterprise fields
- add a new open source plugin that runs last, looks for any non-nil enterprise fields
- add a new closed source plugin interface requires a call to base plugin that ensures all fields configured/looked at have the annotation

thoughts @nfuden ?

nfuden: I like it. It will be faster than scraping for format although we wont get it to check for things like future plugins.
Given that this is in a hot path and the extra cleanliness of using the annotations I think the proposed path makes a lot of sense. 
If we are updating doc gen to respect the extra annotations perhaps we can also take the chance to make it respect a few other annotations that we have left over in some of our envoy protos.

sam-heilbron: Confirmed offline with @nfuden that the annotation proposal is a good path forward, and can be done in a follow-up

PR #5826, opened 2022-01-26
feat(docs): split vulnerabillity scans to render in smaller chunks

gunnar-solo:
# Description

When running CI, we've found that our vulnerability reports are too large to process.  This PR splits the reports into smaller sections, as well as renders said sections individually


sam-heilbron:
/skip-ci https://github.com/unchartedsoftware/witch/issues/4 this is a safe change to merge without CI run. The failure seems to be unrelated. We will need to tackle this dependency upgrade in CI at some point, but right now it is not necessary.

sam-heilbron (approved):

PR #5843, opened 2022-01-28
Slack Notification when CI Fails

sam-heilbron:
# Description

For critical Github actions, notify the team of failure

# Context

Github actions are used on PRs and merges to master to run tests, generate security scans, and build our docs. If these actions fail when producing production assets, we want to be notified immediately. Alternatively, we need to monitor the failed GHA's.

### Security Scan
We already had a notification. I modified it to use the same step and format as other jobs

### Docs
I added a notification if this job fails on merges to master. 

### Docs Generation for Community User PRs
Docs generation was failing for community users. This was due to us skipping the generation of scan reports since it relies on tokens for accessing private resources that community users do not have access to. However, in skipping the generation of those reports, our docs wouldn't build because templates couldn't find the files for those reports.

As a solution, we disable docs generation for community PRs. By enabling notifications if the build fails for merges to master, we can catch a potential break soon after it is introduced.

# Checklist:

### DIFF
--- a/.github/workflows/docs-gen.yaml
+++ b/.github/workflows/docs-gen.yaml
@@ -80,17 +80,17 @@ jobs:
         entryPoint: ./docs/ci
     - name: Notify On Failure
       id: notify-on-failure
-      # if: github.ref == 'refs/heads/master' && failure()
-      # Use preconfigured webhook to post to #edge-team-bots
-      run : |
-        curl -X POST\
-             -H 'Content-type: application/json'\
-             --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch.  FYI <@U02GW11FVJP>"}'\
-             ${{ secrets.SLACK_INTEGRATION_TESTING_WEBHOOK }}
-
-      # For testing out bot edits, we instead have a webhook that posts to the #edge-team-bots channel
+      if: github.ref == 'refs/heads/master' && failure()
+      # Use preconfigured webhook to post to #slack-integration-testing
       # run : |
       #   curl -X POST\
       #        -H 'Content-type: application/json'\
       #        --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch.  FYI <@U02GW11FVJP>"}'\
-      #        ${{ secrets.EDGE_TEAM_BOTS_WEBHOOK }}
+      #        ${{ secrets.SLACK_INTEGRATION_TESTING_WEBHOOK }}
+
+      # For testing out bot edits, we instead have a webhook that posts to the #edge-team-bots channel
+      run : |
+        curl -X POST\
+             -H 'Content-type: application/json'\
+             --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch.  FYI <@U02GW11FVJP>"}'\
+             ${{ secrets.EDGE_TEAM_BOTS_WEBHOOK }}
--- a/.github/workflows/docs-gen.yaml
+++ b/.github/workflows/docs-gen.yaml
@@ -52,13 +52,11 @@ jobs:
         echo "Pull Request is from a fork. Setting IS_COMMUNITY_PR to true"
         echo "::set-output name=IS_COMMUNITY_PR::true"
     - name: Generate versioned docs site
+      if: ${{ !steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
       run: make -C docs build-site
       env:
         GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
-        SKIP_CHANGELOG_GENERATION: ${{ steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
-        SKIP_SECURITY_SCAN: ${{ steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
-        SKIP_ENTERPRISE_DOCS_GENERATION: ${{ steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
-        USE_PR_SHA_AS_MASTER: ${{ github.event_name == 'pull_request' && !steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
+        USE_PR_SHA_AS_MASTER: ${{ github.event_name == 'pull_request' }}
         PULL_REQUEST_SHA: ${{ github.event.pull_request.head.sha }}
     - name: Deploy to Firebase (live docs)
       # Do not publish docs if this workflow was triggered by a pull request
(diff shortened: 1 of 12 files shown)


.github/workflows/docs-gen.yaml
```diff
@@ -52,13 +52,11 @@ jobs:
         echo "Pull Request is from a fork. Setting IS_COMMUNITY_PR to true"
         echo "::set-output name=IS_COMMUNITY_PR::true"
     - name: Generate versioned docs site
+      if: ${{ !steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
       run: make -C docs build-site
       env:
         GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
-        SKIP_CHANGELOG_GENERATION: ${{ steps.community-pr-check.outputs.IS_COMMUNITY_PR }}
```
gunnar-solo, line 58: calling out:

It's "possibly" worth it to remove these environment variables from the `make -C docs build-site` command logic.  I figure extracting it is more trouble than it's worth, but I'm willing to be overruled

:shrug:

.github/workflows/docs-gen.yaml
```diff
@@ -80,3 +78,19 @@ jobs:
         firebaseServiceAccount: ${{ secrets.GC_FIREBASE_GHA_SA_JSON_KEY }}
         projectId: ${{ secrets.GC_FIREBASE_GHA_PROJECT_ID }}
         entryPoint: ./docs/ci
+    - name: Notify On Failure
+      id: notify-on-failure
+      if: github.ref == 'refs/heads/master' && failure()
+      # Use preconfigured webhook to post to #slack-integration-testing
+      # run : |
```
gunnar-solo, line 85: Intended dev use it to hotswap the different `run: |` blocks, so that they can do a rapid series of tests without spamming other developers who are subscribed to the #edge-team-bots channel

.github/workflows/docs-gen.yaml
```diff
@@ -80,3 +78,19 @@ jobs:
         firebaseServiceAccount: ${{ secrets.GC_FIREBASE_GHA_SA_JSON_KEY }}
         projectId: ${{ secrets.GC_FIREBASE_GHA_PROJECT_ID }}
         entryPoint: ./docs/ci
+    - name: Notify On Failure
+      id: notify-on-failure
+      if: github.ref == 'refs/heads/master' && failure()
```
gunnar-solo, line 83: I am "pretty confident" that this is the correct restriction.  HOWEVER running it in actuality is difficult, since it requires a merge to master, first.

Time will tell?

gunnar-solo (commented):
some comments for whomever comes next

.github/workflows/docs-gen.yaml
```diff
@@ -80,3 +78,19 @@ jobs:
...
+      #   curl -X POST\
+      #        -H 'Content-type: application/json'\
+      #        --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch.  FYI <@U02GW11FVJP>"}'\
+      #        ${{ secrets.SLACK_INTEGRATION_TESTING_WEBHOOK }}
+
+      # For testing out bot edits, we instead have a webhook that posts to the #edge-team-bots channel
+      run : |
+        curl -X POST\
+             -H 'Content-type: application/json'\
+             --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch.  FYI <@U02GW11FVJP>"}'\
```
gunnar-solo, line 95: missed this in my initial pass/info comments:

`<@U02GW11FVJP>` is my personal tag, @GunnarHorve.  Stuff like this is _tricky_ to test perfectly, since we need to merge to a different branch, first.  This should reduce cycles/let me hop on stuff more quickly

.github/workflows/docs-gen.yaml
sam-heilbron: I like having this setup directly in the workflow so its easier for devs! One suggestion, we have a top level env variable `DEBUG_WORFKLOW=false` and then both steps are defined, but one step uses `if is true` and the other `if is false`. That way when testing, you just toggle a boolean and the rest is handled.

.github/workflows/docs-gen.yaml
sam-heilbron: Do we want this additional `if ref = master`? At the top of a workflow we get to define triggers (the `on` section) which initiate the worfkflow. I figure if the workflow is initiated and it fails, we want to alert.

.github/workflows/docs-gen.yaml
sam-heilbron: I'm happy to not tag people and jsut set up alerts/notifications on the slack channel we post to. That way its the teams responsiblity to respond to. Alternatively, we could tag edge-on-call

.github/workflows/docs-gen.yaml
gunnar-solo: that will have us being notified _every_ time a doc failure occurs.  Including when someone happens to be working on a docs feature in their PR branch

done this way, we are able to
a) render docs on builds associated with a `PR to master`
b) render docs on builds associated with `master branch`
b) get notified on doc build failures associated with `master branch`
d) _not_ get notified on doc build failures assoicated with a `PR to master`

.github/workflows/docs-gen.yaml
gunnar-solo: Neat!  I didn't know that we could [define environment variables for an entire action](https://docs.github.com/en/actions/learn-github-actions/environment-variables) (the `yaml` threw me off).

Any opposition to me doing something like
```yaml
env:
    SLACK_DEBUG_TESTING: false
...
         curl -X POST\
           -H 'Content-type: application/json'\
           --data '{"text":"Gloo Edge has <https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}|failed a docs build> on `master` branch"}'\
           ${{ $SLACK_DEBUG_TESTING && $secrets.SLACK_INTEGRATION_TESTING_WEBHOOK || $secrets.EDGE_TEAM_BOTS_WEBHOOK }}
           # if SLACK_DEBUG_TESTING=true, post to `#slack-integration-testing`.    Otherwise, post to #edge-team-bots
```

source:  [this reasonable comment](https://github.com/actions/runner/issues/409#issuecomment-752775072)

.github/workflows/docs-gen.yaml
sam-heilbron: You're totally right! I forgot abou the logic we have for distinguishing between PRs and merges. We have the 'live deployment step' (https://github.com/solo-io/gloo/pull/5843/files#diff-e2fba0153ff45c9e1769a63e168fa3d8e17158d83038de4ab71d894f342dc522R66) to handle this logic. Can we just mae sure that both steps use the same logic since they should both occur in tandem.

.github/workflows/docs-gen.yaml
```diff
@@ -80,3 +80,11 @@ jobs:
         firebaseServiceAccount: ${{ secrets.GC_FIREBASE_GHA_SA_JSON_KEY }}
         projectId: ${{ secrets.GC_FIREBASE_GHA_PROJECT_ID }}
         entryPoint: ./docs/ci
+    - name: Notify On Failure
+      id: notify-on-failure
+      if: ${{ github.event_name != 'pull_request' }}
```
sam-heilbron, line 85: I think we need an `& failure()`

gunnar-solo: good catch!

.github/workflows/docs-gen.yaml
nfuden: i like the idea as potential future work

changelog/v1.11.0-beta8/notify-on-gha-failure.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Disbled docs gen for community PRs.  Notify dev team on docs build failure.
```
kdorosh, line 3: do we want to expand this to notify dev team on any failed PR run? (i.e., not just docs. message the dev rather than the whole team)

.github/workflows/docs-gen.yaml
nfuden: Ya this needs to be addressed so we dont get too many extra calls

nfuden (approved):

sam-heilbron:
`Gateway in memory hybrid gateway tcp ssl [It] should work with ssl` is a consistently flaking test, i will investigate as part of my work in that area

Issue #5767, opened 2022-01-07
Hybrid Gateway resourceRef option


bewebi:
### Version

1.10.x (beta)

### Is your feature request related to a problem? Please describe.

We've introduced Hybrid Gateways, which consist of any number of MatchedGateway members, each of which must contain a complete HttpGateway or TcpGateway definition
Therefore Gateway CRs are liable to grow to unwieldy sizes

### Describe the solution you'd like

We should consider adding a new CRD for "single" gateway members of Hybrid Gateways
These would serve the exclusive purpose of being referred to by reference in Hybrid Gateways, adding a third option for MatchedGateways that bypasses the need for complete in-line Gateway definitions within the Gateway CR

### Describe alternatives you've considered

The current solution is functional, but may get out of hand for large numbers of MatchedGateways

### Additional Context

[Original issue](https://github.com/solo-io/gloo/issues/5324)
[Original feature PR](https://github.com/solo-io/gloo/pull/5623)

guydc:
Also relevant here: https://github.com/solo-io/gloo/issues/5549. With this improvement, it would be easier to scope HCM settings to specific virtual services without increasing the size of the gateway resource.

sam-heilbron:
Plan of action:

- [x] Introduce API, go types, CRD, simplify rbac, add basic helm tests (https://github.com/solo-io/gloo/pull/5844)
- [ ] Reporting
  -  [ ] Make it easier to aggregate resourceReports on a ProxyReport, regardless of Gateway type. This will allow our translators to not have to distinguish reports for each of the gateway types.
  - [ ] Introduce suport for new http gateway.
  - [ ] Proxy reconciliation support.
  - [ ] Tests
- [ ] Gateway translation:
  - [ ] make http translator more generic so it can handle HttpGateway on a Gateawy, HttpGateway on a HybridGateway and HttpGateway on a DelegatedGatweay.
  - [ ] Make our logic around merging sslConfig (or which ssl config to aggregate) more explicit. We should easily be able to understand the behavior of translators
  - [ ] HybridTranslator needs to support new type
  - [ ] Tests
- [ ] Gateway validation:
  - [ ] validate new hybrid gateway type
  - [ ] Tests, test, test
- [ ] Internal API changes:
  - [ ] Define Proxy API. This is the shape of data that Gloo will consume
  - [ ] Define LIstenerReport for new type. This is the type that is consumed by the validation webhook(and gloo plugins write to)
- [ ] Gloo translation
[...]

PR #5790, opened 2022-01-13
fix(gateway): allow error clearing for gateways without children

gunnar-solo:
# Description
These changes solve a small bug, where users could generate an error state on a `VirtualService`, which was then (correctly) propagated to its parent `Gateway`.  When deleting said `VirtualService`, said error state _was not removed_ from the `Gateway`.

# Context
Pretty well established with [this "test steps" comment](https://github.com/solo-io/gloo/pull/5790#issuecomment-1012440689)

related https://github.com/solo-io/gloo/issues/5578

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5578

PR #5844, opened 2022-01-30
Introduce MatchableHttpGateway

sam-heilbron:
# Description

Introduce MatchableHttpGateway, a new custom resource that makes it easier to configure multiple FilterChains on a single Listener

# Context

The work to support this feature is being done in stages. The plan/progress can be [tracked here](https://github.com/solo-io/gloo/issues/5767#issuecomment-1026475592)

### New CRD

I added the new protos, as well as moved some shared logic into an `http_gateway.proto` file.

By default codegen will create all appropriate resources, but it will fail with `some methods are missing DeepCopyObject`. These methods are built using a part of codegen that is off by default (because it is time intensive, and doesn't change that often). The output can be found [here](https://github.com/solo-io/gloo/blob/5b1cff19750b27286ea1ef9833f3696178231f98/projects/gloo/pkg/api/v1/kube/apis/gloo.solo.io/v1/zz_generated.deepcopy.go)

To get past this error, you must follow the directions outlined [here](https://github.com/solo-io/gloo/blob/5b1cff19750b27286ea1ef9833f3696178231f98/projects/gloo/pkg/api/v1/kube/apis/gloo.solo.io/v1/doc.go#L7)

### RBAC Changes
https://kubernetes.io/docs/reference/access-authn-authz/rbac/

I made a small modification to the gateway role to include mutating the new CRD.

However, it also cleans up the existing RBAC in a number of ways:
1. I converted the definitions into stricter yaml
2. I broke up some of the rules to draw more clarity to the behavior. For example, I extracted the Gateway create rule to make it abundantly clear why we need to support that
3. I used RoleBindings to apply the Gloo and Gateway Roles to the Gloo ServiceAccount. Instead of modifying the Role for Gateway, we should instead use Bindings to give it the necessary permissions.
4. I removed some of the uncessary permissons on graphql schemas (the ratelimit/status support is only needed since rate limiter is backed by skv2 and the status is written as a subresource in skv2)

### Translator Syncer Sorts Proxies (again)

In this PR, we re-add support to sort proxies during the gateway translator syncer. I had initially said it was ok to remove in a [previous PR](https://github.com/solo-io/gloo/pull/5790/files). However, while working on this PR, I encountered a [test failure](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e8b61d11-ad3a-4101-af36-8b7c2c389868). The failure is that reports were expected to have errors in a particular order. I noticed that when I ran this test with `untilItFails` it failed more frequently, since the ordering is not ensured any more.

I figured we could either:
(A) Change the tests to not expect the order, and just expect the values in any order
(B) Change the code to sort proxies

Given that statuses might be updated unnecessarily if the reports are identical but in a different order, I opted to re-add the sorting behavior, with an explanatory comment so we know for the future.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5767


nfuden (approved):

changelog/v1.11.0-beta8/new-http-gateway-crd.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NON_USER_FACING
```
nrjpoddar, line 2: Why is this NON_USER_FACING?

nrjpoddar:
@sam-heilbron can I see an example of what the new Gateway resource looks like after the new CR has been referenced? I'm assuming all of this is backwards compatible but wanted to confirm nonetheless.

changelog/v1.11.0-beta8/new-http-gateway-crd.yaml
sam-heilbron: Just because the funcrionality was not yet completed. In the follow up PRs to add pieces of functionality I kept each unit of work as non-user facing, but the last PR will describe the NEW_FEATURE

guydc:
Hi @sam-heilbron , @nfuden . Are these new APIs also available for federation: Federated CRDs/go types and gloo-fed implementation?

sam-heilbron:
Hi @guydc these new APIs are not yet supported for federation. We could certainly introduce those if that was of interest. Could you write up an issue outlining the request and urgency and we can prioritize it appropriately?

Issue #5662, opened 2021-12-02
Improve error message for Secrets in logs and glooctl check


antonioberben:
**Describe the bug**
When configuring SSL, if I create a cert with a key which does not match this error appears:
```
[2021-12-01 07:51:00.643][7][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:127] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) listener-::-8443: Failed to load private key from <inline>, Cause: error:0b000074:X.509 certificate routines:OPENSSL_internal:KEY_VALUES_MISMATCH
```

It would be interesting to show which VS and which Secret is causing the problem in the logs and mark it as an ERROR rather than WARNING which is how i is now.

**To Reproduce**
1. run following set of commands:
```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 \  
   -keyout tls.key1 -out tls.crt1 -subj "/CN=*"

kubectl create secret tls upstream-tls --key tls.key1 \
   --cert tls.crt1 --namespace gloo-system --dry-run=client -oyaml > secret1.yaml

openssl req -x509 -nodes -days 365 -newkey rsa:2048 \  
   -keyout tls.key2 -out tls.crt2 -subj "/CN=*"

kubectl create secret tls upstream-tls --key tls.key2 \
   --cert tls.crt2 --namespace gloo-system --dry-run=client -oyaml > secret2.yaml
```

2. override the value of `tls.key` from `secret2.yaml` to `secret1.yaml`
**The intention is to have one secret with one cert which does not match its key.**

3. Apply the secret.
4. Create a VS to use that secret. Something like:
```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: httpbin
  namespace: gloo-system
spec:
  sslConfig:
    secretRef:
      name: <secret name>
      namespace: gloo-system
[...]
```
4. Check the logs from the proxy (i.e. `gateway-proxy` pod)
You should see this kind of log trace level warning
```
[...]

PR #5872, opened 2022-02-03
Validate Integrity of TLS Secrets

ben-taussig-solo:
# Description
Gloo now confirms that TLS secrets contain a valid key/value pair before resolving SSL config. If invalid TLS configuration is identified, a report will be generated at the cluster-level, allowing the user to highlight the problematic resources using `glooctl check`

I am opening this as a draft because I still have a couple of concerns about the UX of this feature. Outstanding questions are as follows:
 - Currently bad TLS config will not cause any resources to be rejected. I.E., if a user creates a VS that references a bad TLS secret in its SSL config, the VS will still be applied. Should this change? If so, what resources should we reject?
 - Issues with TLS config are reported at the gateway level -- they still point to the invalid secret resource but I think it would be nicer for the end user if these errors would show up when "Checking secrets" -- If someone can point me towards how to accomplish this, I would appreciate it. Example `glooctl check` output is as follows:
```
Checking deployments... OK
Checking pods... OK
Checking upstreams... OK
Checking auth configs... OK
Checking RouteOptions... OK
Checking secrets... OK
Checking gateways... 1 Errors!
Checking proxies... OK
Error: 1 error occurred:
        * Found rejected gateway by 'gloo-system': gloo-system gateway-proxy-ssl
Reason: 1 error occurred:
        * Listener Error: SSLConfigError. Reason: name:"upstream-tls" namespace:"gloo-system" is not a valid TLS secret -- invalid certificate key pair
```

# Context

Prior to these changes, Gloo would not reject TLS configuration if it appeared that a key/value pair was present. Error handling around the case in which an invalid key/value pair was present was nonexistent, making it difficult for users to debug their issues in this case

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5662


changelog/v1.11.0-beta9/tls-secret-error-handling.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/5662
+    resolvesIssue: true
+    description: Validate the integrity of TLS secrets, allowing users to call glooctl check to identify bad TLS configuration
```
kdorosh, line 5: 👏

projects/gloo/pkg/utils/ssl.go
```diff
@@ -32,6 +35,15 @@ var (
 		return eris.Errorf("%v is not a TLS secret", ref)
 	}
 
+	InvalidTlsSecretError = func(ref *core.ResourceRef, err error) error {
+		if ref == nil {
+			return eris.Wrapf(err, "Invalid TLS secret")
+		} else {
+			errorString := fmt.Sprintf("%v is not a valid TLS secret", ref)
+			return eris.Wrapf(err, errorString)
```
kdorosh, line 43: are we ever concerned about `err` having sensitive info here? do we have some examples of what that might look like?

kdorosh (commented):
looks like a great UX improvement to me!

MitchAman (commented):
The approach LGTM

> Currently bad TLS config will not cause any resources to be rejected. I.E., if a user creates a VS that references a bad TLS secret in its SSL config, the VS will still be applied. Should this change? If so, what resources should we reject?

I want to yes, this should change. a VS referencing a bad TLS secret should probably be rejected. On the other hand, I worry of the possibility that users may be creating resources in bulk without considering that the secrets need to be created before the virtual services that reference them. That being said, the error reporting is pretty clear, so it would be pretty straightforward in that case to switch the ordering. I don't have a strong opinion here.

> Issues with TLS config are reported at the gateway level -- they still point to the invalid secret resource but I think it would be nicer for the end user if these errors would show up when "Checking secrets" -- If someone can point me towards how to accomplish this, I would appreciate it.

Agreed, this would be nice. I will dig through the code a bit and see if any ideas come to mind

projects/gloo/pkg/utils/ssl.go
ben-taussig-solo: From my testing, `err` seems to be purely descriptive I.E., I don't think this will cause us to dump sensitive private-key information

Here's an example from a failing CI build (need to update an e2e test that uses invalid keys as dummy TLS config)
```
invalid secrets for host : Invalid TLS secret: tls: failed to find any PEM data in certificate input
```

jenshu: I looked through the errors thrown from [tls.X509KeyPair](https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L245) and the functions it calls. There seem to be some places it could print a [uri](https://github.com/golang/go/blob/2aafd9ea179ec308607c131dbda438abac2b9184/src/crypto/x509/parser.go#L408) or [domain](https://github.com/golang/go/blob/2aafd9ea179ec308607c131dbda438abac2b9184/src/crypto/x509/parser.go#L563) - is that sensitive?
(i might have missed some spots - the top level function calls out to a bunch of other functions, maybe someone can double check?)

jenshu (approved):

projects/gloo/pkg/utils/ssl.go
jenshu: (spoke offline - official [docs](https://pkg.go.dev/crypto/tls#X509KeyPair) are logging the error in the example, so we should be good)

PR #5882, opened 2022-02-04
Translate MatchableHttpGateway into Proxy

sam-heilbron:
# Description

Translate MatchableHttpGateways into a Proxy.

# Context

### ListenerFactory Changes
We used to support a ListenerFactory, which takes a list of Gateways and computes a list of Listeners. We would iterate over all ListenerFactories, and process each Gateway in the list, to generate appropriate Listeners.
- I opted to make the ListenerFacotry take a single Gateway and produce a single Listener. This made it easier to re-use the code and write simpler tests.
- I split each of the factories into their own file, and created a co-located test file as well. Most of the contents of the test file were just taken from the aggregate `translator_test` file.
- I introduced the implementation to support finding MatchableHttpGateways and producing Listeners for this new HybridGateway with external refs

### HttpGatewaySelector
This matches a lot of the logic that we used for VirtualServiceSelection and RouteTableSelection. It would be great if we could simplify some of this logic and perhaps create some re-usable selector logic. However, I couldn't see a good way to do that, and figured the scope of this PR was large enough as is.

### Validation
I mirror the original HybridGateway PR and added support for this new config in a HybridGateway.

### HttpListenerOptions Bug
https://github.com/solo-io/gloo/blob/062fb693f3fb297b7ed5ab6cbf7833b6696f7e4e/projects/gateway/pkg/translator/http.go#L306 we previously were not translating the HttpListenerOptions for HybridGateways. That has been resolved as part of this PR.

### What's next
We were able to translate the existing inline HybridGateway definitions into a Listener. This new Gateway configuration could be translated into the same Listener format. However, we also aggregate the SslConfigurations for all selected virtual services. We now need to create duplicate filter chains for those SslConfigs, mirroring the behavior today that we support for HttpGateways.

# Checklist:


projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -0,0 +1,188 @@
...
+		if len(hybridListener.GetMatchedListeners()) == 0 {
+			// matched gateways are define inline, and therefore if they don't produce
+			// any matched listeners, there is an error on the gateway resource
+			params.reports.AddError(gateway, errors.New(EmptyHybridGatewayMessage))
+			return nil
+		}
+	}
+
+	// DelegatedHttpGateways is only processed if there are no MatchedGateways defined
+	if hybridListener == nil {
```
nfuden, line 55: micro-nit: I like the comment on this but couldnt this just become an else statement at line 52?

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -0,0 +1,188 @@
...
+			}
+		}
+
+		hybridListener.MatchedListeners = append(hybridListener.GetMatchedListeners(), matchedListener)
+	}
+
+	return hybridListener
+}
+
+func (t *HybridTranslator) ComputeHybridListenerFromDelegatedGateway(
```
nfuden, line 133: Is there a reason that this is exported?

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -0,0 +1,188 @@
...
+		matchedListener := &gloov1.MatchedListener{
+			Matcher: &gloov1.Matcher{
+				SslConfig:          matchedGateway.GetMatcher().GetSslConfig(),
+				SourcePrefixRanges: matchedGateway.GetMatcher().GetSourcePrefixRanges(),
+			},
+		}
+
+		switch gt := matchedGateway.GetGatewayType().(type) {
+		case *v1.MatchedGateway_HttpGateway:
+			if len(snap.VirtualServices) == 0 {
```
nfuden, line 98: Why do we only care about this in the http case? I know that this is just mirroring the current setup for http vs tcp but do we need this check in the tcp case as well?

projects/gateway/pkg/translator/hybrid_translator.go
sam-heilbron: Yeah good call

projects/gateway/pkg/translator/hybrid_translator.go
sam-heilbron: nope! I can change

projects/gateway/pkg/translator/hybrid_translator.go
sam-heilbron: VirtualServices are aggregated (selected) on an HttpGateway. A TcpGateway does not have a concept of VirtualServices, hence why the conditional is not needed. I'm not even sure if this is the right place for that logic, it feels like the translator should just be able to handle this case appropriately and if we want to add some logging for UX, we handle that earlier on. But for simplicity I had mirrored the HttpGateway use case

projects/gateway/pkg/validation/validator.go
```diff
@@ -891,40 +917,61 @@ func getDelegateRef(delegate *v1.DelegateAction) *core.ResourceRef {
 	return nil
 }
 
-func gatewayListContainsVirtualService(gwList v1.GatewayList, vs *v1.VirtualService) bool {
+func gatewayListContainsVirtualService(ctx context.Context, gwList v1.GatewayList, httpGwList v1.MatchableHttpGatewayList, vs *v1.VirtualService) bool {
 	for _, gw := range gwList {
-		if gw.GetTcpGateway() != nil {
```
sam-heilbron, line 896: I believe this was a bug before. It would cause non-tcp gateways to always claim that a set of gateways do not reference a virtual service. This would cause the validator to be not validate the Gateway for the deletion of a VirtualService

projects/gloo/api/v1/proxy.proto
```diff
@@ -187,6 +187,10 @@ message MatchedListener {
         HttpListener http_listener = 2;
         TcpListener tcp_listener = 3;
     }
+
+    // SSL Config is optional for the MatchedListener. If provided, the listener will serve TLS for connections.
+    // Multiple SslConfigs are supported for the purpose of SNI. Be aware that the SNI domain provided in the SSL Config
+    repeated SslConfig ssl_configurations = 4;
```
bewebi, line 193: Where is/will this be consumed? Is this new behavior and/or fixing a bug?

bewebi (commented):
Big PR, would have been good to break down if possible
A couple comments/questions, but overall looking good!

projects/gloo/api/v1/proxy.proto
sam-heilbron: This is new behavior, I can call it out in the description. It introduces the same duplication behavior we have for existing httpGateways. This enables teams to define their own VirtualServices with ssl config and aggregate them. otherwise we are kind of limiting 1 vs per 1 httpgateway

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -0,0 +1,185 @@
...
+	if matchedGateways != nil {
+		hybridListener = t.computeHybridListenerFromMatchedGateways(params, proxyName, gateway, matchedGateways)
+		if len(hybridListener.GetMatchedListeners()) == 0 {
+			// matched gateways are define inline, and therefore if they don't produce
+			// any matched listeners, there is an error on the gateway resource
+			params.reports.AddError(gateway, errors.New(EmptyHybridGatewayMessage))
+			return nil
+		}
+	} else {
+		// DelegatedHttpGateways is only processed if there are no MatchedGateways defined
```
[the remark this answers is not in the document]
sam-heilbron: It's a good idea. I would have preferred that too. Unfortunately, you can't place `repeated` in a oneof so we would need to introduce a breaking change to the proxy to support it. I figured it was easier to just have both. In fact, there's no reason a user couldn't just configure both types of definitions but I opted to try to make it more restrictive

projects/gateway/pkg/translator/http_translator_test.go
```diff
@@ -0,0 +1,1500 @@
...
+
+				listener := translator.ComputeListener(params, defaults.GatewayProxyName, snap.Gateways[0])
+				Expect(listener).NotTo(BeNil())
+				Expect(reports.ValidateStrict()).NotTo(HaveOccurred())
+
+				httpListener := listener.ListenerType.(*gloov1.Listener_HttpListener).HttpListener
+				Expect(httpListener.VirtualHosts).To(HaveLen(1))
+			})
+
+			It("should not select using both labels and expressions", func() {
```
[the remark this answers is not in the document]
sam-heilbron: I didn't actually change this test from what existed. This test produces 3 virtual services, 2 with the labelSet and 1 without.It then defines an empty expression config, which is treated as a catch-all. I think the test is aiming to validate that we only process the expression config (and select all 3 virtual services) instead of processing both the expression and labels (which would have produced 2 virtual services)

projects/gloo/api/v1/proxy.proto
bewebi: We discussed offline, and, while the actual functionality that uses this field is coming in the next PR, the documentation can be made clearer. Specifically it should add a note as to how we expect the field to be populated in practice (which is to say from constituent VirtualServices of the HybridGateway).

bewebi (approved):

PR #5886, opened 2022-02-04
Fix TLS secrets

jenshu:
# Description

This PR fixes 2 issues:
- TLS secrets created via `glooctl create secret tls` are now created with type `kubernetes.io/tls` instead of `Opaque`
- Fixes a regression in gloo `v1.10.5` and `v1.11.0-beta7` where previously generated Opaque TLS secrets could not be converted back into Gloo TLS secrets

### Background:
When creating a secret via `glooctl create secret...`, these steps occur:
- We create a [Gloo Secret](https://github.com/solo-io/gloo/blob/ee2e1115631e36a6b6136101f5d330dde810be74/projects/gloo/api/v1/secret.proto#L27) of the appropriate type
- We then convert the Gloo secret to a k8s secret via the solo-kit [kubesecret ResourceClient](https://github.com/solo-io/solo-kit/blob/ddf6f96e8a26856642e094fa5bc8c7df0b4406f0/pkg/api/v1/clients/kubesecret/resource_client.go#L149)
    - This in turn goes through each registered converter in our [SecretConverterChain](https://github.com/solo-io/gloo/blob/ee2e1115631e36a6b6136101f5d330dde810be74/projects/gloo/pkg/api/converters/kube/secretconverter.go#L21) and calls `ToKubeSecret`. The first one that returns a non-nil secret will be returned. If none return a non-nil secret, then the default [solo-kit ToKubeSecret](https://github.com/solo-io/solo-kit/blob/ddf6f96e8a26856642e094fa5bc8c7df0b4406f0/pkg/api/v1/clients/kubesecret/resource_client.go#L76) will be called.
- The k8s secret is written to storage
- The k8s secret is read back and [converted back into a Gloo secret](https://github.com/solo-io/solo-kit/blob/ddf6f96e8a26856642e094fa5bc8c7df0b4406f0/pkg/api/v1/clients/kubesecret/resource_client.go#L241)
    - Similar to above, this will go through each registered converter in the SecretConverterChain and call `FromKubeSecret`. The first one that returns non-nil will be used.

Previous to this fix, in the initial conversion from Gloo TLS secret to k8s secret, we were [checking for an annotation](https://github.com/solo-io/gloo/blob/ee2e1115631e36a6b6136101f5d330dde810be74/projects/gloo/pkg/api/converters/kube/secretconverter.go#L97) that was never actually being set anywhere. So TLS secrets created via glooctl would not run through the TLSSecretConverter, and they would end up using the fall-back solo-kit converter and end up as an Opaque k8s secret.

Then, on the conversion back to a Gloo secret, our recently-added HeaderSecretConverter [would see the Opaque type](https://github.com/solo-io/gloo/blob/ee2e1115631e36a6b6136101f5d330dde810be74/projects/gloo/pkg/api/converters/kube/header_secret.go#L30) and incorrectly convert it into a Gloo Header secret. Prior to the HeaderSecretConverter opaque secret support, we would fall back to the [solo-kit FromKubeSecret](https://github.com/solo-io/solo-kit/blob/ddf6f96e8a26856642e094fa5bc8c7df0b4406f0/pkg/api/v1/clients/kubesecret/resource_client.go#L40) which would successfully unmarshal the k8s secret into a gloo secret of TLS type, as long as the expected TLS fields were on the secret.

To fix the above regression but still allow the Header converter to support Opaque secrets with arbitrary key-value pairs, this PR adds a new OpaqueSecretConverter which runs before the HeaderSecretConverter, and calls the solo-kit `FromKubeSecret`. solo-kit will only try to convert secrets with a `resource_kind: *v1.Secret` annotation, that are able to be unmarshaled into a valid Gloo secret type. This will allow TLS secrets that were previously created as Opaque to be read back correctly as Gloo TLS secrets. Newly created TLS secrets will be of type `kubernetes.io/tls`.

# Checklist:


projects/gloo/pkg/api/converters/kube/secretconverter.go
```diff
@@ -93,29 +87,24 @@ func (t *TLSSecretConverter) FromKubeSecret(_ context.Context, _ *kubesecret.Res
 func (t *TLSSecretConverter) ToKubeSecret(_ context.Context, _ *kubesecret.ResourceClient, resource resources.Resource) (*kubev1.Secret, error) {
 	if glooSecret, ok := resource.(*v1.Secret); ok {
 		if tlsGlooSecret, ok := glooSecret.GetKind().(*v1.Secret_Tls); ok {
-			if glooSecret.GetMetadata().GetAnnotations() != nil {
-				if glooSecret.GetMetadata().GetAnnotations()[annotationKey] == annotationValue {
```
jenshu, line 97: this annotation was never set by `glooctl create secret...` and was causing the rest of this function to get skipped, so removed the checks for the annotation

nfuden:
Only 2 small questions here, mainly for my own edification, as it looks great and the write up for the issue was super useful for me to get a base understanding of the context here.
1. Isn't there a chance that header secret will be captured in the opaque secret converter and fail to get extra meta data added? (perhaps a small test to show that it will work fine?)
2. I am confused by this old test https://github.com/solo-io/gloo/blob/secret-test/projects/gloo/pkg/api/converters/kube/secretconverter_test.go#L70-L71 as it seems to really want to enforce that we don't find a valid secret but the phrasing of the descriptor details otherwise. Everything points to the new path being correct but it still gives me pause.

jenshu:
@nfuden good questions! I will add some more details to the PR description and some tests to verify the changes, but in short, the OpaqueSecretConverter calls out to the solo-kit secret converter (which is normally the fallback converter) which only processes secrets with a special `resource_kind` annotation and that it can unmarshal into a gloo resource. So the solo-kit converter will ignore Opaque secrets with arbitrary key/value pairs such as

jenshu:
@nfuden good questions! I will add more details to the PR description and tests to verify the changes, but in short, the OpaqueSecretConverter will only convert a secret that has both a special `resource_kind` annotation and that it can successfully unmarshal into a gloo secret. It will not be able to convert secrets with arbitrary key/value pairs such as
```
apiVersion: v1
data:
  key1: value1
  key2: value2
kind: Secret
metadata:
  name: test
  namespace: gloo-system
type: Opaque
```
which would be handled by the HeaderSecretConverter.

I will go through the old tests and make sure they're up to date.

projects/gloo/cli/pkg/cmd/create/secret/extauth_oauth.go
```diff
@@ -25,9 +25,7 @@ func ExtAuthOathCmd(opts *options.Options) *cobra.Command {
 		Use:   "oauth",
 		Short: `Create an OAuth secret with the given name (Enterprise)`,
 		Long: "Create an OAuth secret with the given name. The OAuth secret contains the client_secret as defined in [RFC 6749](https://tools.ietf.org/html/rfc6749). " +
-			"This is an enterprise-only feature. The format of the secret data is: `{\"oauth\" : [client-secret string]}`. " +
-			"Note that the annotation `resource_kind: '*v1.Secret'` is added in order for Gloo to find this secret. " +
```
jenshu, line 29: annotation no longer used since we have a custom converter

projects/gloo/api/v1/secret.proto
```diff
@@ -121,10 +121,6 @@ message AzureSecret {
     map<string,string> api_keys = 1;
 }
 
-/*
-Note that the annotation `resource_kind: '*v1.Secret'` is needed for Gloo to find this secret.
```
jenshu, line 125: annotation no longer used since we have a custom converter

bewebi (approved):
LGTM

Issue #5651, opened 2021-12-01
Don't add unused filters to the filter chain


Spoonbender:
**Is your feature request related to a problem? Please describe.**
Currently, the filter chains constructed by Gloo Edge include various filters which we don't actually use, such as:
-`io.solo.filters.http.solo_jwt_authn_staged`
-`envoy.filters.http.ext_authz`
-`envoy.filters.http.grpc_web`
-`envoy.filters.http.ratelimit`
-`io.solo.filters.http.graphql`

This is poses two issues:
1. **Security**: having additional, unneeded filters in the chain increases the attack surface. 
   - Even if the filters supposedly don't do anything, it is easier to argue about the security of the system if the filters are not included in the chain at all
2. **Performance**: there is _some_ performance penalty for executing these filters for each HTTP request
   - Even if they do nothing and return almost immediately, this is still a function call with prologue and epilogue etc.

Notice that in a "vanilla" envoy - no filters are added to the chain unless explicitly configured to.

**Describe the solution you'd like**
Gloo should only add filters to the chain if they are actually in use.
For example:
- Don't add the JWT authentication feature if we haven't configured the system to use it
- Don't add the GraphQL filter if we haven't configured Sqoop

**Describe alternatives you've considered**
1. Enabling users to modify the global list of filters Gloo adds to every filter chain
2. Enabling users to configure the list of filters Gloo applies to each VirtualService
3. Enabling users to disable specific features, such as rate-limit or GraphQL, which would in turn exclude these filters from the chain by implication

chrisgaun:
related to https://github.com/solo-io/gloo/issues/5650

nrjpoddar:
@nfuden can you share what user config are you planning to expose to make this configurable?

nfuden:
@nrjpoddar its a good question and one that is not entirely set. 

The current statements are under consideration. 
1. We don't want to break backwards compatibility by changing our plugin interface so it is likely to be advantageous  to have  the configuration setting be usable in the future
2. We want likely want this toggleable at a per plugin level (this may not be true but is likely true)

Current thought is to have an option in Settings for PluginOptions which would have information for allow-listing/deny-listing and an array of plugins to allow/disallow. 

So far I've been digging in on a per filter to see what it means for a filter to not be needed in a given configuration and need to get some feedback on the settings naming/structuring.

nfuden:
Is there a specific version requested for backports? Or are we fine with it being in 1.11?

PR #5887, opened 2022-02-04
 Feature/unused filters

nfuden:
# Description

Oss section to track not adding filters to the filter chain if they are not currently in use.
In general this happens for a plugin in the following instances:
1. The plugin does not require a base configuration (if it does then it is valuable to see if your configuration was valid)
2. Current resource set does not have any routes / virtual host level configurations in effect.
3. The plugin does not have basic gloo enabled settings

# Context
Unused filters were polluting envoy instances adding latency and potential concerns.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5651


yuval-k:
the reason that we've done it in the past, is that if you add a feature in RDS, you need to wait for LDS update before it works. it's probably not a big problem, we can discuss.

nfuden:
Yuval considering that how would you feel about having it be toggleable per filter. That way if we experience issues then for a specific filter we can avoid it.
Or if we think that it makes sense we can have this be purely an opt in feature.

Either way new listeners with entirely new plugin needs should hopefully be a fairly rare update in most mature installations.

yuval-k:
@nfuden i think we can make it opt-in globally in the settings (to simplify it vs per-filter). That way there's no behaviour change for existing installs.

nfuden:
Sure thing Ill change the global value to false by default.

changelog/v1.11.0-beta9/dont-add-unused-filters.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/5651
+    description: >
+      No longer add filter chains if they are not used.
```
sam-heilbron, line 5: nit: It's not that the FilterChains are unused / not added, but the HttpFilters within the HCM that are unused

projects/gloo/api/v1/settings.proto
```diff
@@ -533,6 +534,10 @@ message GlooOptions {
     // hostname with the configured frequency to update endpoints with any changes to DNS resolution.
     // Defaults to 10s.
     google.protobuf.Duration failover_upstream_dns_polling_interval = 13;
+
+    // Most plugins will not be addded to envoy filter chains unless this setting is true
```
sam-heilbron, line 538: nit: `plugin` is kind of an internal word (describing our translation loop) that doesn't mean much to users. Could we clarify what behavior this setting controls. What is the impact to users, why would they want to use this

projects/gloo/pkg/plugins/registry/registry.go
```diff
@@ -129,6 +130,9 @@ type pluginRegistry struct {
 	weightedDestinationPlugins   []plugins.WeightedDestinationPlugin
 }
 
+// NewPluginRegistry creates a plugin registry and places all registered plugins
```
sam-heilbron, line 133: 👏

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -65,6 +66,7 @@ func (p *Plugin) Name() string {
 
 func (p *Plugin) Init(params plugins.InitParams) error {
 	p.RequireEarlyTransformation = false
+	p.filterNeeded = !params.Settings.GetGloo().GetRemoveUnusedFilters().GetValue()
```
sam-heilbron, line 69: https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/plugins/registry/registry.go#L62 the Aws plugin consumes the 'RequireEarlyTransformation' state of the transformation plugin. Does the filterNeeded state affect that logic at all?

sam-heilbron (commented):
This looks great! 

Out of curiosity, why do we only need to add this state to a subset of the plugins? 

Also, can we open the Solo-Projects PR before merging this just to confirm that there are no side effects on the enterprise side?

projects/gloo/pkg/plugins/transformation/plugin.go
nfuden: Nope filterneeded only short circuits the HTTPFilter functions on plugins and require early transformation is computed based on virtualhost processing. 
In general the approach taken across the plugins is to not stop anything other than httpfilter computation and to avoid changing any other plugin sideeffects.

changelog/v1.11.0-beta9/dont-add-unused-filters.yaml
nfuden: Good point I have updated the changelog accordingly. ( I was also a bit lazy earlier while seeing if there was any other structural changes that were going to be requested)

projects/gloo/api/v1/settings.proto
nfuden: Thanks for the callout! Updated accordingly

sam-heilbron (approved):
LGTM!

PR #5888, opened 2022-02-04
Feature/merge gloo gateway

elcasteel:
Merge the gloo and gateway pods by moving the gateway functionality into the gloo pod and combining previously separate translation loops. 

### Translation 
This change unifies previously separate translation loops to generate Proxy CRs and generate Envoy config. 
Details and diagrams about the steps in translation are here: https://soloio.slab.com/posts/gloo-gateway-translation-overview-2k05nawb

### Validation
The validation webhook used to live in the gateway pod and has been moved to the gloo pod. The old workflow was roughly:

1. Gateway pod gets a validation request
2. Perform gateway translation (to check for validity)
3. Send grpc validation request to gloo pod
4. Gloo pod runs translation/validation and responds to gateway pod with result
5. Gateway pod responds to client. 

With the pods merged, the grpc calls are replaced with function calls and the steps to translate/validate are the same.
 
The GRPC server is still present in the gloo pod for customer integrations. 

### Helm
The number of replicas for the gateway pod is hard-coded to 0. In future changes we can remove more of the gateway pod-related infrastructure.
There are rbac changes to give the gloo pod permissions to update gateway resources.
The validating webhook configuration and certificates are moved to the gloo pod. 

### New Settings
Two new settings have been added to the Settings CRD: `persistProxySpec` and `enableGatewayController`.

`enableGatewayController` flags all the functionality that used to be contained in the gateway pod. It's expected that this is only set at install time (when the gateway pod would have been started)

`persistProxySpec` controls whether proxies are stored in memory or in etcd. It should be `true` if users are generating proxies with custom controllers. It is also used for tests and debugging. 

### User facing documentation
I updated the documentation that shows users what pods to look for when they install Gloo (removed gateway pod). 
Currently inspecting Proxies only works when `persistProxySpec` is true, although one of the early pieces of follow up work will be to fix that. We can document the behavior changes as part of that.


projects/gateway/pkg/reconciler/memory_client.go
```diff
@@ -0,0 +1,115 @@
+package reconciler
```
elcasteel, line 1: The relationship between memoryResourceClient and memoryProxyClient is kind of backwards, I plan to fix it.

projects/gateway/pkg/setup/setup.go
```diff
@@ -2,19 +2,15 @@ package setup
...
-
-	"github.com/solo-io/gloo/pkg/version"
-
-	"github.com/solo-io/gloo/pkg/utils/setuputils"
-	"github.com/solo-io/gloo/projects/gateway/pkg/syncer"
 )
 
 func Main(customCtx context.Context) error {
-	return setuputils.Main(setuputils.SetupOpts{
+	/*return setuputils.Main(setuputils.SetupOpts{
```
elcasteel, line 8: Gateway install/start up will be removed properly as part of a future task.

projects/gateway/pkg/reconciler/memory_client.go
sam-heilbron: We can use solo-kit to create resourceClients for various persistence layers. Is there a reason we can't leverage what we do in this test: https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/syncer/translator_syncer_integration_test.go#L45 to create an InMemoryResourceClient for the Proxy?

elcasteel: That's probably better- thanks for the heads up

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -542,14 +559,111 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions, apiEmitte
...
+	var (
+		// this constructor should be called within a lock
+		validationClient             validationclients.GlooValidationServiceClient
+		ignoreProxyValidationFailure bool
+		allowWarnings                bool
+	)
+	if gwOpts.Validation != nil {
+		//TODO: this can be in memory
+		logger.Infof("[ELC]creating validation client %s", gwOpts.Validation.ProxyValidationServerAddress)
+		validationClient, err = gwvalidation.NewConnectionRefreshingValidationClient(
```
elcasteel, line 644: One of the next steps is to remove GRPC calls and replace with in-memory operations

projects/gateway/pkg/reconciler/memory_client.go
sam-heilbron: Is this file still needed or can we just instantiate the solo-kit proxy memory client that i linked in the above comment?

projects/gateway/pkg/setup/setup.go
sam-heilbron: What happens if this setup remains? I had imagined we just wouldn't scale the gateway deployment and keep all gateway code the same. But perhaps there is an effect of haivng it run?

projects/gateway/pkg/syncer/setup_syncer.go
```diff
@@ -284,7 +285,6 @@ func RunGateway(opts translator.Opts) error {
 	translatorSyncer := NewTranslatorSyncer(
 		ctx,
 		opts.WriteNamespace,
-		proxyClient,
```
sam-heilbron, line 287: Is there a reason we remove the proxyClient entirely instead of just keeping it, but injecting an in memory implementation that satisfies the same interface?

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -24,44 +24,39 @@ import (
...
 	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
 	"github.com/solo-io/solo-kit/pkg/api/v2/reporter"
-	"github.com/solo-io/solo-kit/pkg/errors"
 )
 
-type translatorSyncer struct {
+type TranslatorSyncer struct {
 	writeNamespace     string
 	reporter           reporter.Reporter
-	proxyWatcher       gloov1.ProxyWatcher
```
sam-heilbron, line 37: similar to the proxyClient, i wonder if we can/should leverage a similar approach. In that gateway insatllation case we don't need any proxy watching, but in other installations we'll still likely need that behavior, right? Should we just implement a no-op watcher that gloo uses for the gateway install an then keep the real proxy watcher for ingress?

projects/gloo/pkg/validation/server.go
```diff
@@ -110,7 +140,18 @@ func (s *validator) pushNotifications() {
...
+		gwSnap := &gatewayv1.ApiSnapshot{
+			VirtualServices:    snap.VirtualServices,
+			RouteTables:        snap.RouteTables,
+			Gateways:           snap.Gateways,
+			VirtualHostOptions: snap.VirtualHostOptions,
+			RouteOptions:       snap.RouteOptions,
+		}
+		s.gwValidator.Sync(ctx, gwSnap)
+	}
+	if s.shouldNotify(snap) || gatewayChange {
```
sam-heilbron, line 154: I think we want to keep shouldNotify with all of this logic so that we have a consistent pattern of causing syncs: clon snapshot -> compare hash -> if different, push notification to channel which causes a sync .
 
 In the current pattern (I may be wrong), won't we call the gwValidator.sync and then push a notifcation to the channel which calls gloo sync to run, which now includes the gw sync as well? So we would be syncing gateway resources twice?

sam-heilbron (commented):
Overall looks good! 
I think it will be helpful to diagram the order of events of translation in this new in memory proxy world, compared to the old translation loop. It may help clarify the expected order of operations and potentially allow us to simplify the translation loop even further

projects/gateway/pkg/reconciler/memory_client.go
elcasteel: This can be deleted/is deleted in my environment but I missed it in my git commits.

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
             - path: "labels"
               fieldRef:
                 fieldPath: metadata.labels
+      - name: validation-certs
+        secret:
+          defaultMode: 420
```
MitchAman, line 65: Just curious what this mode does, and the significance of 420?

projects/gloo/pkg/syncer/translator_syncer.go
```diff
@@ -115,3 +130,28 @@ func (s *translatorSyncer) Sync(ctx context.Context, snap *v1snap.ApiSnapshot) e
...
+func (s *translatorSyncer) translateProxies(ctx context.Context, snap *v1snap.ApiSnapshot) {
+	gwSnap := &gatewayv1.ApiSnapshot{
+		VirtualServices:    snap.VirtualServices,
+		Gateways:           snap.Gateways,
+		RouteTables:        snap.RouteTables,
+		RouteOptions:       snap.RouteOptions,
+		VirtualHostOptions: snap.VirtualHostOptions,
+	}
+	logger := contextutils.LoggerFrom(ctx)
+	s.gatewaySyncer.Sync(ctx, gwSnap)
```
MitchAman, line 143: Should handle the case where `Sync` returns an error?

MitchAman (commented):
Looking great so far - no objections with the general approach

projects/gateway/pkg/syncer/translator_syncer.go
elcasteel: this is code that currently only runs in gateway mode

install/helm/gloo/templates/1-gloo-deployment.yaml
kdorosh: how the secret gets mounted, i.e. the `chmod` permissions for the file. for more, see https://kb.iu.edu/d/abdb (search "absolute form")

kdorosh: but also 🍁 🔥 ☁️ ??

projects/gloo/pkg/syncer/translator_syncer.go
```diff
@@ -115,3 +133,28 @@ func (s *translatorSyncer) Sync(ctx context.Context, snap *v1snap.ApiSnapshot) e
...
+	proxyList, _ := s.proxyClient.List("gloo-system", clients.ListOpts{})
```
kdorosh, line 147: just calling out that we won't want to hardcode the namespace here (this is likely test code, but still might be easy to forget about)

oneroute.txt
```diff
@@ -0,0 +1,87 @@
+{"level":"info","ts":"2022-03-03T21:06:25.293Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:195","msg":"received validation request on webhook","version":"1.11.0-logs"}
```
kdorosh, line 1: todo: remove

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -24,44 +27,39 @@ import (
...
 		proxyReconciler: proxyReconciler,
 		translator:      translator,
 		statusSyncer:    newStatusSyncer(writeNamespace, proxyWatcher, reporter, statusClient, statusMetrics),
 		managedProxyLabels: map[string]string{
 			"created_by": "gateway",
 		},
 	}
-
-	go t.statusSyncer.watchProxies(ctx)
+	// go t.statusSyncer.watchProxies(ctx)
```
kdorosh, line 55: we probably need this to be gateway by the configuration option (if we are using / writing proxy CRs or not)

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -209,41 +212,55 @@ func (s *statusSyncer) watchProxiesFromChannel(ctx context.Context, proxies <-ch
...
 
+//Separate this functionality out so it can be called outside of the watch (hopefully remove watch too)
+func (s *statusSyncer) handleUpdatedProxies(ctx context.Context, proxyList gloov1.ProxyList) {
+	logger := contextutils.LoggerFrom(ctx)
+	if proxyList == nil {
+		proxyList, _ = s.proxyClient.List(s.writeNamespace, clients.ListOpts{
+			Ctx: ctx,
+		})
+	}
+	var previousHash uint64
```
kdorosh, line 228: this looks like a bug; we used to initialize this before the loop and it would be persisted / reused across invocations of what is now `handleUpdatedProxies`, but this is no longer true

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -209,41 +212,55 @@ func (s *statusSyncer) watchProxiesFromChannel(ctx context.Context, proxies <-ch
...
-			}
+			s.handleUpdatedProxies(ctx, proxyList)
 		}
 	}
 }
 
+//Separate this functionality out so it can be called outside of the watch (hopefully remove watch too)
+func (s *statusSyncer) handleUpdatedProxies(ctx context.Context, proxyList gloov1.ProxyList) {
+	logger := contextutils.LoggerFrom(ctx)
+	if proxyList == nil {
```
kdorosh, line 223: why this addition?

projects/gloo/pkg/validation/server.go
```diff
@@ -105,17 +136,36 @@ func (s *validator) pushNotifications() {
...
+		}
+		err = s.gwValidator.Sync(ctx, gwSnap)
+		if err != nil {
+			//TODO: better log, do something with error
+			contextutils.LoggerFrom(ctx).Errorf("Error running gateway validation, %v", err)
+		//	return err
+		}
+	}
+	//TODO: if s.shouldNotify(snap) || (gatewayChange && gwMode)
+	if s.shouldNotify(snap) || gatewayChange {
```
kdorosh, line 163: not sure we need / want `gatewayChange`.. why do we need this? We only validate on gateway api resource events so we only need to notify on gloo event changes (since the gateway validator should already be run automatically with a change to gateway api event group)

projects/gateway/pkg/syncer/translator_syncer.go
elcasteel: this is an artifact of switching back and forth between calling this from a watch and calling it immediately after the gloo translation finishes that I need to clean up

projects/gloo/pkg/validation/server.go
```diff
@@ -137,34 +105,17 @@ func (s *validator) pushNotifications() {
...
-	contextutils.LoggerFrom(ctx).Infof("[ELC] gloo validation sync ready? %v", s.gwValidator.Ready())
-	if !s.gwValidator.Ready() || gatewayChange {
-		gwSnap := &gatewayv1.ApiSnapshot{
-			VirtualServices:    snap.VirtualServices,
-			RouteTables:        snap.RouteTables,
-			Gateways:           snap.Gateways,
-			VirtualHostOptions: snap.VirtualHostOptions,
-			RouteOptions:       snap.RouteOptions,
-		}
-		err = s.gwValidator.Sync(ctx, gwSnap)
```
elcasteel: I think this call to Sync() still needs to happen

install/helm/gloo/generate/values.go
```diff
@@ -265,6 +265,7 @@ type Gateway struct {
 	ReadGatewaysFromAllNamespaces *bool              `json:"readGatewaysFromAllNamespaces,omitempty" desc:"if true, read Gateway custom resources from all watched namespaces rather than just the namespace of the Gateway controller"`
 	CompressedProxySpec           *bool              `json:"compressedProxySpec,omitempty" desc:"if true, enables compression for the Proxy CRD spec"`
 	LogLevel                      *string            `json:"logLevel,omitempty" desc:"Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
+	PersistProxySpec              *bool              `json:"persistProxySpec,omitempty" desc:"Enable writing Proxy CRD to etc. Disabled by default for performance."`
```
kdorosh, line 268:
```suggestion
	PersistProxySpec              *bool              `json:"persistProxySpec,omitempty" desc:"Enable writing Proxy CRD to etcd Disabled by default for performance."`
```

install/test/fixtures/settings/disabled_gateway.yaml
```diff
@@ -0,0 +1,37 @@
...
+    app: gloo
+    gloo: settings
+  name: default
+  namespace: {{ . }}
+spec:
+  discovery:
+    fdsMode: WHITELIST
+  gateway:
+    readGatewaysFromAllNamespaces: false
+    gatewayMode: false
```
kdorosh, line 14: can we rename this to something a bit more accurate / specific, such as `enableGatewayController`? both with and without the controller we are still running in "gateway" mode

projects/gloo/api/v1/settings.proto
```diff
@@ -639,4 +639,12 @@ message GatewayOptions {
...
     // When these properties are defined on a specific VirtualService, this configuration will be ignored
     VirtualServiceOptions virtual_service_options = 7;
+
+    // Set this to persist the Proxy CRD to etcd. By default, proxies are kept in memory to improve performance.
+    // Proxies can be persisted to etcd to allow external tools and other pods to read the contents the Proxy CRD.
+    bool persist_proxy_spec = 8;
+
+    // This is set based on the install mode. It indicates to gloo whether or not it should run the gateway
+    // translations and validation.
+    bool gateway_mode = 9;
```
kdorosh, line 649: ah, i see. we probably don't need this at all. users could previously install with both ingress and gateway, we should be able to support that as well. rather than rename this we should be able to remove it

elcasteel: The name may not be the best, but we need a flag that is accessible from the gloo setup to indicate whether or not the gloo pod should generate proxies from from gateway types and in the future whether to start the validation grpc server.

kdorosh: We could also address those concerns by:
- always starting gloo validation server
- relying on `persist_proxy_spec` as signal for whether or not to be in "`gateway_mode`" (i.e. have both gloo and gateway controllers read this single global setting value)

elcasteel: Currently regression tests rely on setting `persist_proxy_spec` while also running in `gateway_mode` (and generally I think it's a useful setting to have for debugging)

kdorosh: if it's for developers/debugging, can we put it in an undocumented env var rather than our public facing API? trying to keep the user experience as simple as reasonably possible (fewer toggles, less to worry about)

.github/workflows/regression-tests.yaml
```diff
@@ -12,7 +12,8 @@ jobs:
       matrix:
         # knative support has been deprecated: https://github.com/solo-io/gloo/issues/5707
         # We have removed it from our CI regression tests
-        kube-e2e-test-type: ['gateway', 'ingress', 'helm', 'gloomtls', 'glooctl']
+        #TODO: not sure why helm tests aren't working in ci, trying to run the other suites
```
sam-heilbron, line 15: What failures were you seeing with helm tests?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook_test.go
```diff
@@ -353,6 +354,15 @@ func parseReviewResponse(resp *http.Response) (*AdmissionReviewWithProxies, erro
 	}
 	return &review, nil
 }
+func convertSnapshot(glooSnap *gloov1snap.ApiSnapshot) *v1.ApiSnapshot {
+	return &v1.ApiSnapshot{
+		VirtualServices:    glooSnap.VirtualServices,
+		RouteTables:        glooSnap.RouteTables,
+		Gateways:           glooSnap.Gateways,
```
sam-heilbron, line 361: Do we need httpGateways here as well? Can we add a test that will start failing when the number of properties in gatewaySnapshot changes?

projects/gateway/pkg/setup/setup.go
sam-heilbron: Is this still intentionally commented out? I see that we have replicas=0 and some other helm flags to toggle the gateway behavior

projects/gateway/pkg/syncer/translator_syncer.go
sam-heilbron: bump on this. I think kevin meant "gated by the configuration option" a

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -139,18 +137,21 @@ type statusSyncer struct {
 	reporter                reporter.StatusReporter
 
 	proxyWatcher   gloov1.ProxyWatcher
```
sam-heilbron, line 139: is the proxyWatcher still needed?

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -139,18 +137,21 @@ type statusSyncer struct {
...
+	proxyClient    gloov1.ProxyClient
 	writeNamespace string
 	statusClient   resources.StatusClient
 	statusMetrics  metrics.ConfigStatusMetrics
 	syncNeeded     chan struct{}
+	previousHash   uint64
```
sam-heilbron, line 145: what is this a hash of?

projects/gloo/api/v1/settings.proto
```diff
@@ -639,4 +639,12 @@ message GatewayOptions {
     // Default configuration to use for VirtualServices, when not provided by a specific virtual service
     // When these properties are defined on a specific VirtualService, this configuration will be ignored
     VirtualServiceOptions virtual_service_options = 7;
+
+    // Set this to persist the Proxy CRD to etcd. By default, proxies are kept in memory to improve performance.
+    // Proxies can be persisted to etcd to allow external tools and other pods to read the contents the Proxy CRD.
+    bool persist_proxy_spec = 8;
```
sam-heilbron, line 645: If I set this to true, are proxies still kept in memory as well?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -368,13 +381,14 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions, apiEmitte
 		return err
 	}
 
-	proxyClient, err := v1.NewProxyClient(watchOpts.Ctx, opts.Proxies)
+	memoryProxyClient, err := v1.NewProxyClient(watchOpts.Ctx, opts.Proxies)
```
sam-heilbron, line 384: Should this still be a proxyClient? Can't this be either the proxy client backed in memory or by etcd?

test/kube2e/util.go
```diff
@@ -71,10 +71,21 @@ settings:
   singleNamespace: true
   create: true
   replaceInvalidRoutes: true
+gateway:
+  persistProxySpec: true
 gatewayProxies:
   gatewayProxy:
     healthyPanicThreshold: 0
 `))
+	/*
```
sam-heilbron, line 80: nit: remove?

test/samples/sample_gloo_api_snapshot.go
```diff
@@ -391,3 +443,13 @@ func GatewaySnapshotWithDelegateSelector(us *core.ResourceRef, namespace string)
 	snap.RouteTables = []*gwv1.RouteTable{rt}
 	return snap
 }
+
+func GatewayToGlooSnapshot(snap *gwv1.ApiSnapshot) *v1snap.ApiSnapshot {
+	return &v1snap.ApiSnapshot{
+		VirtualServices:    snap.VirtualServices,
+		RouteTables:        snap.RouteTables,
+		Gateways:           snap.Gateways,
```
sam-heilbron, line 451: missing httpGateways?

elcasteel: httpGateways aren't used in any of these unit tests although maybe we should also add them to tests.

projects/gloo/pkg/syncer/setup/setup_syncer.go
elcasteel: I changed the name back. The client will either be backed by memory or etcd depending on the factory created in `opts.Proxies`

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -589,17 +700,15 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions, apiEmitte
...
-	if err != nil {
-		return err
-	}
-	translationSync := syncer.NewTranslatorSyncer(t, opts.ControlPlane.SnapshotCache, xdsHasher, xdsSanitizer, rpt, opts.DevMode, syncerExtensions, opts.Settings, statusMetrics)
+	translationSync := syncer.NewTranslatorSyncer(t, opts.ControlPlane.SnapshotCache, xdsHasher, xdsSanitizer, rpt, opts.DevMode, syncerExtensions, opts.Settings, statusMetrics, gwTranslatorSyncer, proxyClient, opts.WriteNamespace)
 
 	syncers := v1snap.ApiSyncers{
-		translationSync,
 		validator,
+		translationSync,
```
elcasteel, line 707: Switching the order that these syncs run in made some e2e tests work and I don't think it will impact normal operation but I want to call this out in case I'm wrong

sam-heilbron (commented):
Changes generally look good. There are still a number of todos and commented out lines in the PR. I think it would be helpful too to outline the different behaviors between with and without the gateway controller. There are a number of components that have custom branching logic to sync proxies or statuses and I think some comments oultining specifically what they do and why would be helpful

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -87,9 +87,13 @@ spec:
 
   gateway:
     readGatewaysFromAllNamespaces: {{ .Values.gateway.readGatewaysFromAllNamespaces }}
+{{- if or .Values.gateway.persistProxySpec (not .Values.gateway.enabled) }}
```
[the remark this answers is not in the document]
elcasteel: Yes that's what's expected. `persistProxySpec` means get proxies from etcd. 
If gateway is disabled then it needs to be true (proxies are coming from either the ingress pod or a customer controller). If gateway is enabled then by default the proxies are entirely in memory but we have the option of also persisting them to etcd for debugging.

projects/gloo/pkg/syncer/translator_syncer.go
```diff
@@ -64,30 +70,45 @@ func NewTranslatorSyncer(
...
 }
 
 func (s *translatorSyncer) Sync(ctx context.Context, snap *v1snap.ApiSnapshot) error {
 	logger := contextutils.LoggerFrom(ctx)
-	var multiErr *multierror.Error
 	reports := make(reporter.ResourceReports)
+
+	logger.Infof("translation for snapshot %v", snap.Proxies)
+	//generate proxies
+	// TODO: only run if there was an update to a gw type
```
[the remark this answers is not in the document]
elcasteel: This takes a gateway snapshot and creates Proxy CRs. After talking with Kevin I'm not going to attempt to decide whether the gateway snapshot has changed and instead run it on every translation loop (when in gateway mode)

test/e2e/aws_ec2_test.go
```diff
@@ -241,7 +241,7 @@ var _ = Describe("AWS EC2 Plugin utils test", func() {
 		defaults.HttpPort = services.NextBindPort()
 		defaults.HttpsPort = services.NextBindPort()
 
-		testClients = services.RunGateway(ctx, false)
+		testClients = services.RunGateway(ctx, true)
```
elcasteel, line 244: This test does not use the gateway translations and instead writes proxies using proxyClient. Before nothing in the gateway snapshot was changing so it didn't matter whether `justGloo` was set or not. With the merged gloo and gateway pod we regenerate proxies on every translation loop so tests will break if this is not set correctly everywhere.

projects/gateway/pkg/syncer/translator_syncer.go
elcasteel: which is to say that the watcher was looking for status updates from gloo to proxies generated by the gateway logic. I replaced the watcher and there is more documentation for that in other parts of this file.

projects/gateway/pkg/setup/setup.go
elcasteel: This part is no longer commented although there isn't a configuration only way to switch between merged gloo and gateway pods and separated pods. Among other things, we would need to keep the GRPC calls to validate resources which is a big part of the performance gains so I don't think we should try to do it.

projects/gloo/pkg/syncer/setup/proxy_cleanup.go
```diff
@@ -0,0 +1,59 @@
+package setup
```
elcasteel, line 1: I'm working on unit tests for this

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -76,12 +72,16 @@ func (s *translatorSyncer) Sync(ctx context.Context, snap *v1.ApiSnapshot) error
...
 	}
 
-	desiredProxies, invalidProxies := s.generatedDesiredProxies(ctx, snap)
+	desiredProxies, invalidProxies := s.GeneratedDesiredProxies(ctx, snap)
 
 	return s.reconcile(ctx, desiredProxies, invalidProxies)
 }
 
-func (s *translatorSyncer) generatedDesiredProxies(ctx context.Context, snap *v1.ApiSnapshot) (reconciler.GeneratedProxies, reconciler.InvalidProxies) {
+//This replaced a watch on the proxy CR from when the gloo and gateway pods were separate
```
sam-heilbron, line 80: This is a helpful comment. Can we also add some context around who is responsible for calling this function and why?

projects/gateway/pkg/syncer/translator_syncer_integration_test.go
```diff
@@ -189,14 +189,14 @@ var _ = Describe("TranslatorSyncer integration test", func() {
 
 		// write the proxy status.
 		AcceptProxy()
-
+		ts, _ := ts.(*TranslatorSyncer)
+		ts.UpdateProxies(ctx)
```
sam-heilbron, line 193: These explicit calls are required because of the removed proxy watch, correct?

projects/gateway/pkg/validation/validator.go
```diff
@@ -85,29 +91,43 @@ type Validator interface {
...
-	latestSnapshot               *v1.ApiSnapshot
-	latestSnapshotErr            error
-	translator                   translator.Translator
-	validationClient             validation.GlooValidationServiceClient
+	lock              sync.RWMutex
+	latestSnapshot    *v1.ApiSnapshot
+	latestSnapshotErr error
+	translator        translator.Translator
+	//This function replaces a grpc client from when gloo and gateway pods were separate.
+	validationFunc func(
```
sam-heilbron, line 105: nit: can we use the ValidatorFunc type?

projects/gateway/pkg/validation/validator.go
```diff
@@ -170,6 +194,39 @@ func (v *validator) Sync(ctx context.Context, snap *v1.ApiSnapshot) error {
...
+	if v.latestSnapshot == nil {
+		return true
+	}
+	//look at the hash of resources that affect the gateway snapshot
+	hashFunc := func(snap *v1.ApiSnapshot) uint64 {
+		toHash := append([]interface{}{}, snap.VirtualHostOptions.AsInterfaces()...)
+		toHash = append(toHash, snap.VirtualServices.AsInterfaces()...)
+		toHash = append(toHash, snap.Gateways.AsInterfaces()...)
+		toHash = append(toHash, snap.RouteOptions.AsInterfaces()...)
+		toHash = append(toHash, snap.RouteTables.AsInterfaces()...)
```
sam-heilbron, line 218: Do we need to add HttpGateways as well?

projects/gateway/pkg/validation/validator.go
```diff
@@ -170,6 +194,39 @@ func (v *validator) Sync(ctx context.Context, snap *v1.ApiSnapshot) error {
 
 type applyResource func(snap *v1.ApiSnapshot) (proxyNames []string, resource resources.Resource, ref *core.ResourceRef)
 
+func convertSnapshot(glooSnap *gloov1snap.ApiSnapshot) *v1.ApiSnapshot {
+	return &v1.ApiSnapshot{
```
sam-heilbron, line 198: In the future if we add a new type to our snapshot, is there a way we can be sure to update this function as well?

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -87,9 +87,13 @@ spec:
...
+    persistProxySpec: true
+{{- end}}
 {{- if .Values.gateway.compressedProxySpec}}
     compressedProxySpec: {{ .Values.gateway.compressedProxySpec}}
 {{- end}}
+    enableGatewayController: {{ .Values.gateway.enabled }}
```
sam-heilbron, line 96: Right now we expose an 'enable_gateway_controller' field in settings. Should that be the field that is consumed here?

projects/gateway/pkg/validation/validator.go
```diff
@@ -251,7 +312,7 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
 			continue
 		}
 
-		if v.validationClient == nil {
+		if v.validationFunc == nil {
```
sam-heilbron, line 315: Is this still possible?

test/kube2e/gateway/gateway_test.go
```diff
@@ -495,9 +495,11 @@ var _ = Describe("Kube2e: gateway", func() {
 						Upstream: ref,
 					},
 				}
-
-				_, err := virtualServiceClient.Write(getVirtualService(dest, nil), clients.WriteOpts{})
-				Expect(err).NotTo(HaveOccurred())
+				//TODO: This used to work immediately
```
sam-heilbron, line 498: TODO

test/kube2e/gateway/gateway_test.go
```diff
@@ -556,6 +559,9 @@ var _ = Describe("Kube2e: gateway", func() {
 					return err
 				}, time.Second*10).ShouldNot(HaveOccurred())
 
+				helpers.EventuallyResourceRejected(func() (resources.InputResource, error) {
```
sam-heilbron, line 562: 👏

sam-heilbron (commented):
Given the size of this change, I think it would be helpful to include an overview about how the relevant components differ before and after this change, and what fields affect those settings. I could see this as either being a diagram or just text in the PR description.

projects/gateway/pkg/syncer/translator_syncer_integration_test.go
elcasteel: Correct. I'm not sure if this test is super meaningful in the current state though, I'm open to suggestions to restructure it. I think basically we just need to show that statuses on gateway types are updated when proxy statuses change. Merging the pods and no longer relying on a watch have simplified this process a bit and some of these tests seemed to be essentially testing the watch.

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -60,6 +60,12 @@ spec:
             - path: "labels"
               fieldRef:
                 fieldPath: metadata.labels
+      {{- if .Values.gateway.enabled }}
```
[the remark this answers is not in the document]
elcasteel: In the current version, it controls whether the gateway pod is started or not (because all the functionality is contained in that pod).

test/kube2e/gateway/gateway_suite_test.go
```diff
@@ -132,6 +131,8 @@ settings:
   singleNamespace: true
   create: true
   replaceInvalidRoutes: true
+gateway:
+  persistProxySpec: true
```
elcasteel, line 135: We have tests that create or read proxies so I added this, but the result is that we're not testing operations with proxies in memory. I'm not sure what edge cases we might want to handle to ensure there is coverage for the proxies in memory case (restart the gloo pod?). Possibly we can just run some of the happy path cases. Alternatively we can set `persistProxySpec` explicitly during the tests that require it.

install/helm/gloo/templates/23-namespace-clusterrolebinding-gateway.yaml
```diff
@@ -108,6 +108,9 @@ subjects:
 - kind: ServiceAccount
   name: gateway
   namespace: {{ .Release.Namespace }}
+- kind: ServiceAccount
```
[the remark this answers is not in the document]
elcasteel: this whole chart is already gated by `.Values.gateway.enabled` (line 2)

install/helm/gloo/generate/values.go
sam-heilbron: etcd?

install/helm/gloo/templates/1-gloo-deployment.yaml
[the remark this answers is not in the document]
sam-heilbron: Later on in this template we guard the validation-certs by `{{- if and .Values.gateway.validation.enabled .Values.gateway.enabled }}`. Does it make sense to only add this secret if both gateway and gateway.validation are enabled?

install/helm/gloo/templates/18-settings.yaml
sam-heilbron: Oh my mistake I thought I had seen it in helm values previously, but I double checked and there is not one.

install/helm/gloo/templates/2-gloo-service.yaml
```diff
@@ -25,6 +25,11 @@ spec:
   - name: wasm-cache
     port: 9979
     protocol: TCP
+  - name: https
```
sam-heilbron, line 28: Does this need to be guarded by a conditional?

projects/gateway/pkg/validation/validator.go
```diff
@@ -170,6 +191,39 @@ func (v *validator) Sync(ctx context.Context, snap *v1.ApiSnapshot) error {
...
+
+	if v.latestSnapshot == nil {
+		return true
+	}
+	//look at the hash of resources that affect the gateway snapshot
+	hashFunc := func(snap *v1.ApiSnapshot) uint64 {
+		toHash := append([]interface{}{}, snap.VirtualHostOptions.AsInterfaces()...)
+		toHash = append(toHash, snap.VirtualServices.AsInterfaces()...)
+		toHash = append(toHash, snap.Gateways.AsInterfaces()...)
+		toHash = append(toHash, snap.RouteOptions.AsInterfaces()...)
```
sam-heilbron, line 214: I think we need virtualHostOptions as well?

sam-heilbron (commented):
Looking good! Small questions/nits that aren't blocking. I think we can remove the debug lines from the helm test.

What sort of documentation do you think we need? Perhaps that is accumulated in a follow up, but I imagine customers who are upgrading to this version may have questions

projects/gateway/pkg/validation/validator.go
sam-heilbron: Agreed, when will we be able to do that?

projects/gateway/pkg/syncer/translator_syncer.go
elcasteel: I addressed this and the other comment on this file, not sure why the updated version isn't showing up here :/

gunnar-solo (approved):
It is time.

elcasteel:
/kick
I think this test has passed before, failing logs in case I broke something today: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=44811f77-cf1b-42d9-82e4-7c46e82b6cf7

elcasteel:
/kick hopefully this is another flake https://console.cloud.google.com/cloud-build/builds;region=global/18a0b6bc-2beb-445a-9b83-a53ce1760e86?project=solo-public

nfuden (approved):

gunnar-solo (approved):

nfuden (approved):

elcasteel:
/kick 
For some reason two cloudbuilds started at once and then both were cancelled

PR #5893, opened 2022-02-07
Fix TLS secrets [1.10]

jenshu:
backport #5886


nfuden:
LGTM the changelog error is weird I will take a look to see why its happening as everything looks correct in the file structure.

nfuden (approved):

PR #5898, opened 2022-02-08
Docs: version bumps for latest

Rachael-Graham:
Backport of https://github.com/solo-io/gloo/pull/5897/ for 1.10 branch


docs/active_versions.json
```diff
@@ -1,12 +1,12 @@
 {
-  "latest": "1.10.0",
+  "latest": "1.10.6",
```
artberger, line 2: So I don't think the older versions build based off this file (so we typically don't update it), but I also think it's ok to update it, so will approve.

artberger (approved):

docs/active_versions.json
artberger: FYI, latest is now 1.10.7

jenshu (approved):

Issue #5905, opened 2022-02-09
Envoy extension information in Edge docs


nfuden:
### Version

master

### Describe the requested changes

Gloo Edge currently does not display (leak?) information about Solo's envoy extensions in our docs.

This issue MUST decide a final location for the new security posture docs which are generated and validated in our repositories.

This issue MAY decide if there is anything else that should be brought in (underlying envoy version being relied on?)

Envoyproxy itself exposes their security posture yaml in a docs view like the following: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/threat_model 
which is based off a similar yaml
https://github.com/envoyproxy/envoy/blob/main/source/extensions/extensions_metadata.yaml

### Link to any relevant existing docs

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/threat_model 

### Additional Context

This may not be purely a docs issue and there has been some discussion started in the gloo-edge group internally. 

Yuval has a PDF.

Rachael-Graham:
For:
> This issue MAY decide if there is anything else that should be brought in (underlying envoy version being relied on?)

see discussion in https://github.com/solo-io/gloo/pull/5906/files#r803052638 about whether to add the underlying version of Envoy to the supported versions reference page in the docs

chrisgaun:
@sam-heilbron is there a reason we couldn't mirror this Envoy security posture to a public repo?

nfuden:
We could mirror it over it just means that we are segmenting the posture from the code which means it is more easily able to become unsynced. 
Either way this issue is not just for linking and publicizing our private extensions but also the extensions that we have in envoy-gloo (OSS extensions)

kcbabo:
@Nadine2016 this came up again in a customer call today to get the security posture documentation at a public (or private if needed) URL. Do we plan on publishing this in our docs and updating it with each release?

artberger:
OSS: https://github.com/solo-io/envoy-gloo/blob/master/security_posture.yaml
EE: https://github.com/solo-io/envoy-gloo-ee/blob/master/security_posture.yaml

kcbabo:
Confirmed with Yuval that we are good to publish this info in our docs.

artberger:
Right now, we manually combined the OSS and Enterprise files together to publish in the doc. We might want a follow-up issue to look into automating the update of these for future releases. https://docs.solo.io/gloo-edge/master/reference/security-posture/

PR #5906, opened 2022-02-09
Docs: Clarify version support statement

Rachael-Graham:
# Description

This fixes the version support statement in the reference documentation.

# Context

The previous support statement did not clarify n-3 or n-2. See [internal slack thread](https://solo-io-corp.slack.com/archives/CEDCS8TAP/p1642097915100800) for context.

### DIFF
--- a/docs/content/reference/support.md
+++ b/docs/content/reference/support.md
@@ -4,56 +4,51 @@ description: Expected release cadence and support of Gloo Edge
 weight: 52
 ---
 
+Review the following information about supported release versions for Gloo Edge Enterprise and Gloo Edge OSS (open source).
 
-The following documents the expected release cadence and support of both Gloo Edge Enterprise and Gloo Edge OSS (opensource). In general, Gloo Edge Enterprise releases will follow from the Gloo Edge OSS releases (ie, Gloo Edge Enterprise 1.3 would come from Gloo Edge OSS 1.3 code base). In other words, the community will always see the "latest releases" while enterprise would always see the first stable release of that version. 
+## Supported versions
 
-
-## Stable releases every three months
-
-For both Gloo Edge Enterprise and open-source Gloo Edge, we release stable builds approximately every three months. These builds will be tagged from `master` and will bump the minor version (ie, `1.3` to `1.4`). Support for both Gloo Edge Enterprise and Gloo Edge OSS will be provided on stable branches of these releases. 
-
-## Support will be release N through N-2
-
-When stable branches are created for a particular release, i.e. a Gloo Edge Enterprise release, we'll support both EE and OSS from that branch. We will support the current release as well as the two previous releases. For example, if Gloo Edge 1.5 is the latest Enterprise release, we will be releasing 1.6.x from master (OSS) but supporting Enterprise/OSS 1.5.x, 1.4.x, and 1.3.x. Support in this context means bug fixes and critical security fixes. Gloo Edge Enterprise customers will always have priority-fix support from Solo.io, but the OSS project will eventually get those fixes as well as whatever else the community contributes.
-
-Gloo Edge Enterprise customers can purchase additional N-x support, _but the max support that can be purchased is N-3_.
-
-The recommended version for helm clients is v3.0 or greater, but helm v2 will also work.
+Gloo Edge Enterprise offers `n-3` patching support for bug and critical security fixes. In other words, the current release and the two previous releases are supported. For example, if the latest stable Gloo Edge Enterprise release is 1.10, then Gloo Edge Enterprise 1.9.x and 1.8.x are also supported.
 
 | Gloo Edge | Kubernetes | Envoy | Helm | Istio |
 |------|----------|---------|--------|------|
+| 1.10.x | 1.19 - 1.22 | v3 xDS API | >= 3.0 | 1.7 - 1.11 |
 | 1.9.x | 1.19 - 1.22 | v3 xDS API | >= 3.0 | 1.7 - 1.11 |
 | 1.8.x | 1.19 - 1.21 | v3 xDS API | >= 3.0 | 1.7 - 1.8 |
 | 1.7.x | 1.19 - 1.21 | v3 xDS API | >= 3.0 | 1.7 - 1.8 |
 
+## Release cadence
 
+Gloo Edge Enterprise releases typically follow the equivalent Gloo Edge OSS release. The OSS version is always released as the latest build, while Enterprise version is always released as the first stable build of that version. For example, the latest build of Gloo Edge OSS is {{< readfile file="static/content/version_geoss_latest.md" markdown="true">}}, while the latest stable build of Gloo Edge Enterprise is {{< readfile file="static/content/version_gee_latest.md" markdown="true">}}. The Enterprise 1.10 version is built on the OSS 1.10 version code base.
 
-## New features developed on master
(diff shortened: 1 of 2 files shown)


docs/content/reference/support.md
```diff
@@ -4,56 +4,51 @@ description: Expected release cadence and support of Gloo Edge
...
-
-Gloo Edge Enterprise customers can purchase additional N-x support, _but the max support that can be purchased is N-3_.
-
-The recommended version for helm clients is v3.0 or greater, but helm v2 will also work.
+Gloo Edge Enterprise offers `n-3` patching support for bug and critical security fixes. In other words, the current release and the two previous releases are supported. For example, if the latest stable Gloo Edge Enterprise release is 1.10, then Gloo Edge Enterprise 1.9.x and 1.8.x are also supported.
 
 | Gloo Edge | Kubernetes | Envoy | Helm | Istio |
 |------|----------|---------|--------|------|
+| 1.10.x | 1.19 - 1.22 | v3 xDS API | >= 3.0 | 1.7 - 1.11 |
```
Rachael-Graham, line 15: Is this info correct for 1.10? Is Istio 1.12 fully tested/supported yet?

docs/content/reference/support.md
```diff
@@ -4,56 +4,51 @@ description: Expected release cadence and support of Gloo Edge
...
 
+### Stable release process
 
+Development of a quality stable release on `master` typically follows this process:
+1. New feature development is suspended on `master`.
+2. Release candidates are created, such as `1.10.0-rc1`, `1.10.0-rc2`, and so on.
+3. A full suite fo tests is performed for each release candidate. Testing includes all documented workflows, a test matrix of all supported platforms, and more.
+4. Documentation for that release is prepared, vetted, and staged.
+5. The stable minor version is released.
+6. Feature development on `main` is resumed.
```
artberger, line 40:
```suggestion
6. Feature development on `master` is resumed.
```

docs/content/reference/support.md
```diff
@@ -4,56 +4,51 @@ description: Expected release cadence and support of Gloo Edge
...
+Gloo Edge Enterprise is supported and tested for the latest Kubernetes version and all Kubernetes versions released up to 1 year before the latest version. However, because you might need to run Gloo Edge on older versions of Kubernetes, Gloo Edge development attempts to maintain backwards compatibility for earlier releases up to Kubernetes 1.11.
 
 ### Envoy
-Officially, Envoy community releases are supported only N-1 (one prior release). We have the ability to back port to N-2 or more, etc in support situations without bumping the Envoy version (ie, we will fix based on the code you have deployed within the N-2 window). 
+Officially, Gloo Edge Enterprise offers support for `n-1` of Envoy community releases. In specific support situations, fixes can be backported to `n-2` or more without bumping the Envoy minor version. In other words, a fix can be developed based on the code that you deployed within the `n-2` release timeframe.
```
artberger, line 48: I am a bit confused on what the Envoy versions are, because in the table at the top, we have it formatted as `v3 xDS API`...so are we saying basically we support v3 and v2? or does this refer to n-1 patch versions? For reference, this is the list of Envoy versions I found, which seems different than what's in the table: https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history

I guess the table was just referring to the API version 3 and xDS support, not specific patches. 

And so, does our latest GEE version support Envoy 1.21 and 1.20 currently then based on this statement?

docs/content/reference/support.md
```diff
@@ -4,56 +4,51 @@ description: Expected release cadence and support of Gloo Edge
...
+3. A full suite fo tests is performed for each release candidate. Testing includes all documented workflows, a test matrix of all supported platforms, and more.
+4. Documentation for that release is prepared, vetted, and staged.
+5. The stable minor version is released.
+6. Feature development on `main` is resumed.
 
 ## Additional support information
 
 ### Kubernetes 
-We officially support and test 1 year of previous versions of Kubernetes, however, we expect customers to run older versions and for Gloo Edge Enterprise would provide ability to do so. As we add new features to Gloo Edge, we are mindful of what Kubernetes features we leverage so as not to raise the minimum required version. Currently our backward compatibility target is Kubernetes 1.11.
+Gloo Edge Enterprise is supported and tested for the latest Kubernetes version and all Kubernetes versions released up to 1 year before the latest version. However, because you might need to run Gloo Edge on older versions of Kubernetes, Gloo Edge development attempts to maintain backwards compatibility for earlier releases up to Kubernetes 1.11.
```
artberger, line 45: > attempts to maintain backwards compatibility for earlier releases up to Kubernetes 1.11.

This seems like an odd statement, since in the table that latest version we say is 1.19.

docs/content/reference/support.md
Rachael-Graham: So i actually saw a convo about this this morning in the team channel: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1644426496977369

Wondering if we should change the table at the top to have the specific version of envoy instead? so like for 1.10 we'd list Envoy 1.20 (and maybe 1.19 for n-1) in the table row

nfuden: Our bundled extended version of envoy is currently based off 1.20 in GEE 
If there is a desire to use a different distro of envoy we communicate via the v3 XDS api so one could slot in another envoy distro and still use gloo to configure it. See Envoy's xds api [versioning scheme](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/versioning)[](https://github.com/envoyproxy/go-control-plane#deprecated).
That being said it is likely the commitment being made here is that we by default conform to Envoy's model (by nature of just using the nth version and it having support) while also noting that we have the resources to potentially write something / backport something on a custom envoy wrapper. (disclaimer not entirely sure for this last statement, if need be we can ask others on the team).

docs/content/reference/support.md
nfuden: 1.11 has been tested but I dont see any evidence of 1.12 being tested 
That being said the breaking changes in 1.11 to 1.12 dont look like they should be an issue.

docs/content/reference/support.md
nfuden: Oh I also opened https://github.com/solo-io/gloo/issues/5905 with regards to general envoy doc status. 
That is somewhat more focused on publishing the security models for our extra envoy filters but could also be a nice chance to pull in more details about the bundled versions of envoy.

Rachael-Graham: Okay, I think for now I will leave the table as-is, and we will look at this in more detail in that issue.

docs/content/reference/support.md
Rachael-Graham: Okay, I am leaning more towards leaving it as 1.11 since 1.12 has not been fully tested yet.

nrjpoddar: Why do we specify Istio version for Edge? cc @sam-heilbron @kdorosh

nfuden: We do support some cli interactions and have code specifically for plumbing with 1.16 versus newer istio
https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/cmd/istio/sidecars/matcher.go#L21

artberger (dismissed):

artberger (approved):

docs/content/reference/support.md
nrjpoddar: We should document what is tested. @sam-heilbron how do we qualify new Istio releases? Should we work on qualifying v1.12?

PR #5933, opened 2022-02-16
Respect aws proto configs

nfuden:
# Description
Make sure that we can set the 2 newer lambda configuration options via xds and not just helm


projects/gloo/cli/pkg/printers/upstream.go
```diff
@@ -100,9 +101,10 @@ func upstreamDetails(up *v1.Upstream, xdsDump *xdsinspection.XdsDump) []string {
 		for _, fn := range usType.Aws.GetLambdaFunctions() {
 			functions = append(functions, fn.GetLambdaFunctionName())
 		}
+
 		add(
 			fmt.Sprintf("region: %v", usType.Aws.GetRegion()),
-			fmt.Sprintf("secret: %v", usType.Aws.GetSecretRef().Key()),
+			fmt.Sprintf("secret: %s", stringifyKey(usType.Aws.GetSecretRef())),
```
sam-heilbron, line 107: perhaps a long term solution would be to modify: https://github.com/solo-io/solo-kit/blob/master/pkg/api/v1/resources/core/ref_extensions.go#L11, but this works for now

projects/gloo/api/v1/settings.proto
```diff
@@ -470,6 +470,17 @@ message GlooOptions {
             // and therefore must be explicitly specified via the uri
             envoy.config.filter.http.aws_lambda.v2.AWSLambdaConfig.ServiceAccountCredentials service_account_credentials = 2;
         }
+
+        // Send downstream path and method as `x-envoy-original-path` and
+        // `x-envoy-original-method` headers on the request to AWS lambda. 
+        // Defaults to false.
+        bool propagate_original_routing = 3;
```
sam-heilbron, line 477: nit: should we expose this as a nullable value to minimize configuration that other users don't care about?

projects/gloo/cli/pkg/printers/upstream.go
```diff
@@ -219,6 +221,17 @@ func linesForServiceSpec(serviceSpec *plugins.ServiceSpec) []string {
 	return spec
 }
 
+// stringifyKey for a resource likely could be done more nicely with spew
```
sam-heilbron, line 224: 👏

projects/gloo/cli/pkg/printers/upstream.go
nfuden: Unfortunately the call on the nil would still break it so it would be an extra call to secretref or something

sam-heilbron (approved):

projects/gloo/cli/pkg/printers/upstream.go
```diff
@@ -219,6 +221,17 @@ func linesForServiceSpec(serviceSpec *plugins.ServiceSpec) []string {
...
+// or a better accessor but minimall this avoids panicing on nested references to nils
+func stringifyKey(plausiblyNilRef *core.ResourceRef) string {
+
+	if plausiblyNilRef == nil {
+		return "<None>"
+	}
+	return plausiblyNilRef.Key()
```
MitchAman, line 231: Huh, I don't know why it took me this long to realize this `Key()` function exists, but I like it.

MitchAman (commented):
I don't have the full context on the issue you're trying to resolve in order to review this for correctness. That being said, I have no concerns with the implementation/changes being made.

If you need the approval and want to get me up to speed on the problem being solved, I'd be happy to put a stamp on this for you

PR #5957, opened 2022-02-22
Helm improvements for docs

sam-heilbron:
# Description

Improve accuracy of Helm documentation

# Context

Closed source follow-up: https://github.com/solo-io/solo-projects/pull/3171

### How are helm docs built?
Helm values are published to our external docs here: https://docs.solo.io/gloo-edge/latest/reference/helm_chart_values/

These are generated in the following order:
1. Define a values-template file which is yaml containing default values
2. Unmarshal into a specific Go type (https://github.com/solo-io/gloo/blob/master/install/helm/gloo/generate/values.go#L9)
3. Modify the Go type with common version details (https://github.com/solo-io/gloo/blob/00b465d3965a3da3d13570e927fa1bfdac5873b3/install/helm/gloo/generate.go#L157)

We rely on the HelmConfig Go struct as a way of ensuring that the default values of the chart conform to a certain structure.

### Why are the values and templates not in sync?
In the helm templates, (https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/8-gateway-proxy-service.yaml#L57) paths to values are used. 

The Go struct that we described earlier only impacts the default values. Therefore, if you write a test and use the helm `--set` option, we never verify that the path defined in the set conforms to the Go struct.

The effect of this was that our helm templates were behaving properly, but the Go struct used to generate our docs was not in sync.

### What is the fix?
In our test suite we now attempt to take the helm values for a test (unstructured) and unmarshal them into the Go struct. We perform strict unmarshalling which fails if an unstructured field is not available in the Go struct. This identified fields in our tests that were not the proper type (or were not exposed in our docs in the first place).

**There are also Kube tests as well to validate that the generated yaml can be applied to a running cluster. This ensures that our fixture data (the source of truth for our unit tests) is valid (some of which was not)**

### What is the impact?
In the future if a developer adds a field to our helm template, and writes a test to validate it, but forgets to add that field to the Go struct used to generate our docs, the test will fail (previously it would not).

# Checklist:


sam-heilbron:
/kick old logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=7f7a2f0d-43e5-407d-a60e-7c06d904901e

sam-heilbron:
/kick bad logs https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=53dc1972-ee38-47ee-8e90-40aaa2b2ac8f

test/kube2e/helm/helm_test.go
```diff
@@ -197,3 +247,20 @@ func GetGlooServerVersion(ctx context.Context, namespace string) (v string) {
...
+
+func makeUnstructured(yam string) *unstructured.Unstructured {
+	jsn, err := yaml.YAMLToJSON([]byte(yam))
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	runtimeObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, jsn)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	return runtimeObj.(*unstructured.Unstructured)
+}
+
+func makeUnstructureFromTemplateFile(fixtureName string, values interface{}) *unstructured.Unstructured {
```
bewebi, line 259:
```suggestion
func makeUnstructuredFromTemplateFile(fixtureName string, values interface{}) *unstructured.Unstructured {
```

bewebi (commented):
One nit/typo/gotcha on a function name
Otherwise looks good!

test/kube2e/helm/helm_test.go
```diff
@@ -182,6 +188,50 @@ var _ = Describe("Kube2e: helm", func() {
...
+		It("works using kubectl apply", func() {
+			err := filepath.Walk(settingsFixturesFolder, func(settingsFixtureFile string, info os.FileInfo, err error) error {
+				if err != nil {
+					return err
+				}
+				if info.IsDir() {
+					return nil
+				}
+
+				templatedSettings := makeUnstructureFromTemplateFile(settingsFixtureFile, namespace)
```
bewebi, line 220:
```suggestion
				templatedSettings := makeUnstructuredFromTemplateFile(settingsFixtureFile, namespace)
```

bewebi (approved):

PR #5963, opened 2022-02-22
fix regex route short circuiting check [1.10]

jenshu:
Backport of #5959

### DIFF
--- a/projects/gateway/pkg/translator/http_translator.go
+++ b/projects/gateway/pkg/translator/http_translator.go
@@ -471,7 +471,7 @@ func regexShortCircuits(laterMatcher, earlierMatcher *matchers.Matcher) bool {
 	foundIndex := re.FindStringIndex(laterPath)
 	// later matcher is always non-regex. to validate against the regex, we need to ensure that it's either
 	// unset or set to false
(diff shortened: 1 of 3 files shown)


projects/gateway/pkg/translator/http_translator.go
```diff
@@ -471,7 +471,7 @@ func regexShortCircuits(laterMatcher, earlierMatcher *matchers.Matcher) bool {
 	foundIndex := re.FindStringIndex(laterPath)
 	// later matcher is always non-regex. to validate against the regex, we need to ensure that it's either
 	// unset or set to false
-	return foundIndex != nil && laterMatcher.GetCaseSensitive() == nil || laterMatcher.GetCaseSensitive().GetValue() == false
+	return foundIndex != nil && (laterMatcher.GetCaseSensitive() == nil || laterMatcher.GetCaseSensitive().GetValue() == false)
```
kdorosh, line 474: super nit, not blocking, but i believe this is equivalent to
```suggestion
	return foundIndex != nil && laterMatcher.GetCaseSensitive().GetValue() == false
```

sam-heilbron (approved):

Issue #5633, opened 2021-11-24
Support Trailer preservation


guydc:
**Is your feature request related to a problem? Please describe.**
Envoy drops proxied trailers by default. It should be possible to preserve trailers. There is an option to enable trailers in the [Http1ProtocolOptions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-http1protocoloptions). Please note that this setting must be configured both in the HCM and Cluster Http1ProtocolOptions, for the trailers to be proxied successfully, as mentioned [here](https://github.com/envoyproxy/envoy/issues/18120). 

**Describe the solution you'd like**
Expose this option as a Setting or as Gateway and Upstream attributes.

PR #5971, opened 2022-02-23
Expose enable_trailers setting in Envoy HCM

ben-taussig-solo:
Adds support for the go-control-plane HTTP Connection Manager setting [enable_trailers](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-http1protocoloptions)

The parent issue specifies:
> Please note that this setting must be configured both in the HCM and Cluster Http1ProtocolOptions, for the trailers to be proxied successfully

As such, upstream-level and gateway-level settings have been added to control each of these configurations.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5633


projects/gloo/api/v1/options/hcm/hcm.proto
```diff
@@ -60,8 +60,10 @@ message HttpConnectionManagerSettings {
     bool accept_http_10 = 15;
     string default_host_for_http_10 = 16;
 
-    // For an explanation of this setting, see: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-http1protocoloptions
+
+    // For an explanation of these settings, see: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#config-core-v3-http1protocoloptions
     bool allow_chunked_length = 34;
+    bool enable_trailers = 35;
```
nfuden, line 66: Given that this is a whole sub object (Http1ProtocolOptions) does it make sense to pull in any of the other fields? 
If so then would we want it to a be a nested option to keep our protos smaller?

ben-taussig-solo: I think in an ideal world `Http1ProtocolOptions` and `CommonHttpProtocolOptions` would each be nested options. In my mind, the organization of Listener `HttpConnectionManagerSettings` from most preferable to least preferable looks like the following:
1. `Http1ProtocolOptions` and `CommonHttpProtocolOptions` are each nested options
2. A selection of `Http1ProtocolOptions` settings and `CommonHttpProtocolOptions` settings are available at the top level of Listener `HttpConnectionManagerSettings` (this is the way things are currently)
3. Some `Http1ProtocolOptions`/`CommonHttpProtocolOptions` settings are available at the top level of Listener `HttpConnectionManagerSettings`, while some are available in nested options

Because of this, I would want to take all of the existing `Http1ProtocolOptions` and `CommonHttpProtocolOptions` settings and place them in nested options if I were to place /any/ of them in nested options. This would entail moving the following `HttpConnectionManagerSettings` options into nested objects:
 - `Http1ProtocolOptions`: `AllowChunkedLength`, `EnableTrailers`
   - Both of these have been added recently so I don't think this change would be particularly disruptive
 - `CommonHttpProtocolOptions`: `IdleTimeout`, `MaxConnectionDuration`, `MaxStreamDuration`, `MaxHeadersCount`, `MaxRequestsPerConnection`, `HeadersWithUnderscoresAction`
   - There are enough settings here that I am inclined to hold off on making this change. I think too many users will be potentially disrupted by this sort of change that I'm not comfortable making it at the moment. But I think it makes sense to keep our eyes on this proto, because I can easily see this situation getting out of hand.

ben-taussig-solo: TL;DR I would only want to make this change if I were to put ALL of the HttpProtocolOptions settings into nested objects, and I don't feel comfortable doing that as it would potentially disrupt a lot of users

kdorosh (approved):

PR #5974, opened 2022-02-23
Fix route replacement when service is missing

jenshu:
# Description

During gloo translation, a [snapshot sanitizer](https://github.com/solo-io/gloo/blob/f4255209a99293f386322664490ffdba194e4444/projects/gloo/pkg/syncer/setup/setup_syncer.go#L558) runs, which first runs an [upstream removing sanitizer](https://github.com/solo-io/gloo/blob/f4255209a99293f386322664490ffdba194e4444/projects/gloo/pkg/syncer/sanitizer/upstream_removing_sanitizer.go) and then a [route replacing sanitizer](https://github.com/solo-io/gloo/blob/f4255209a99293f386322664490ffdba194e4444/projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go).

When we have an upstream that points to a missing service, the upstream removing sanitizer removes the upstream/cluster from the xds snapshot. Then the route replacing sanitizer is run, which creates a fallback cluster for any invalid routes. The bug was that the route replacing sanitizer was assuming all upstreams in the gloo snapshot were valid clusters. The fix is to *also* check that the cluster still exists in the xds snapshot (i.e. hadn't already been removed by another sanitizer). When the route replacing sanitizer sees that the cluster doesn't exist in the xds snapshot, it will create a fallback route.

# Context

# Checklist:


projects/gateway/pkg/translator/http_translator.go
```diff
@@ -471,7 +471,7 @@ func regexShortCircuits(laterMatcher, earlierMatcher *matchers.Matcher) bool {
 	foundIndex := re.FindStringIndex(laterPath)
 	// later matcher is always non-regex. to validate against the regex, we need to ensure that it's either
 	// unset or set to false
-	return foundIndex != nil && (laterMatcher.GetCaseSensitive() == nil || laterMatcher.GetCaseSensitive().GetValue() == false)
+	return foundIndex != nil && laterMatcher.GetCaseSensitive().GetValue() == false
```
jenshu, line 474: this is just addressing a nit from a previous (unrelated) PR

projects/gloo/pkg/syncer/sanitizer/upstream_removing_sanitizer.go
```diff
@@ -54,6 +54,10 @@ func (s *UpstreamRemovingSanitizer) SanitizeSnapshot(
 		if reports[up].Errors != nil {
 
 			clusterName := translator.UpstreamToClusterName(up.GetMetadata().Ref())
+			if clusters.Items[clusterName] == nil {
```
jenshu, line 57: this change is unrelated to the route replacement bug. not sure exactly how to reproduce, but we have seen cases where the `clusters.Items[clusterName].ResourceProto()...` line below panics. just adding a check here to avoid that situation

nfuden (approved):
Thanks for the PR description! It not only made the rationale super clear it also helped answer some questions I almost put on the PR!
🥇

Issue #5604, opened 2021-11-19
[Helm] multiple containers break block 


asayah:
**Describe the bug**

Anything in this block breaks if you also have multiple containers enabled: https://github.com/solo-io/gloo/blob/88a16c5218a6c8a03ba3271a6c0f9480e791fade/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L302-L323
```
7-gateway-proxy-deployment.yaml
      {{- if $spec.kind.daemonSet }}
      {{- if $spec.kind.daemonSet.hostPort }}
      hostNetwork: {{ ne "false" ($spec.kind.daemonSet.hostNetwork | toString) }} {{/* defaults to true if undefined */}}
      dnsPolicy: ClusterFirstWithHostNet
      {{- end}}
```

inFocus7:
Hi Adam! 👋 I have a few quick questions!

- What exactly is meant by having multiple containers enabled?
- Do you happen to remember how you pinpointed the errors to those block of codes?

PR #5987, opened 2022-02-28
[Helm] Move pod-level options out of container config

inFocus7:
# Description

Bug fix moves pod-level options outside of the container configuration.

# Context

Issue was reported by user stating the code block I moved broke configuration, since it was inside the the `containers` block (in between glooMtls & istioSDS).

After own investigation, the generated k8s pod specification was still valid even with this bug/issue present in our Helm template file. It looks like k8s (or something in our code-base) still correctly parses + fixes this issue for us.

It's still worth moving this block outside from in-between containers, to better closely match the generated/final k8s spec and to not cause future confusion!

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works (current tests are okay, this doesn't affect anything build-wise)

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5604


sam-heilbron (commented):
Do we have any tests to validate the proxy deployment when multiple containers are included? If not, do you think it's worth it to add one?

Changes look good, and I appreciate the extra context in the PR!

inFocus7:
> Do we have any tests to validate the proxy deployment when multiple containers are included? If not, do you think it's worth it to add one?
> 
> Changes look good, and I appreciate the extra context in the PR!

Added unit test to make sure multiple containers are able to be set up 👍 May need to streamline the test

gunnar-solo (dismissed):
LGTM!

gunnar-solo (approved):

PR #5998, opened 2022-03-01
Fix discovery when multiple discoveries apply to a single upstream

saiskee:
We recently introduced the fact that multiple discoveries could apply to a single upstream, for example grpc function discovery as well as grpc-graphql discovery. This exposed a bug where only the first function discovery will run for an upstream indefinitely.  

Solve this by creating a goroutine for each discovery that applies to an upstream.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/pull/5998


projects/discovery/pkg/fds/updater.go
```diff
@@ -309,12 +307,16 @@ func (u *updaterUpdater) Run() error {
...
-		err := discoveryForUpstream.DetectFunctions(context.Background(), resolvedUrl, u.dependencies, upstreamSave)
-		if err != nil {
-			return eris.Wrapf(err, "Error doing discovery %T", discoveryForUpstream)
-		}
+		go func(d UpstreamFunctionDiscovery) {
+			err := d.DetectFunctions(context.Background(), resolvedUrl, u.dependencies, upstreamSave)
+			if err != nil {
+				logger.Errorf("Error doing discovery %T: %s", d, err.Error())
+			}
+		}(discoveryForUpstream)
```
kdorosh, line 319: i think this will leak go-routines

see that this gets called by https://github.com/solo-io/gloo/blob/4043b34c5d780bba92860d0b3af61ac0ea556602/projects/discovery/pkg/fds/updater.go#L133-L138 , which runs when upstreams get added. we need to make sure we can cancel the old infinite loops and terminate those go-routines when new upstreams are added

kdorosh (changes requested):
let's address potential go routine leak

projects/discovery/pkg/fds/updater.go
saiskee: How is that goroutine you linked above being cleaned? Shouldn't it also leak given that `updater.Run()` runs indefinitely?

kdorosh: I think it also leaks, but this makes it exponentially worse 😓

projects/discovery/pkg/fds/updater.go
```diff
@@ -132,7 +130,6 @@ func (u *Updater) UpstreamAdded(upstream *v1.Upstream) {
 	u.activeUpstreams[key] = updater
 	go func() {
 		updater.Run()
-		cancel()
```
saiskee, line 135: Removed this cancel as now the go funcs / endless loops return when the context is cancelled.

projects/discovery/pkg/fds/updater.go
```diff
@@ -300,21 +311,36 @@ func (u *updaterUpdater) Run() error {
...
 	for _, discoveryForUpstream := range discoveriesForUpstream {
-		err := discoveryForUpstream.DetectFunctions(context.Background(), resolvedUrl, u.dependencies, upstreamSave)
-		if err != nil {
-			return eris.Wrapf(err, "Error doing discovery %T", discoveryForUpstream)
-		}
+		go func(d UpstreamFunctionDiscovery) {
+			for {
+				select {
+				case <-u.ctx.Done():
+					return
```
kdorosh, line 332: let's debug/trace log when this happens

projects/discovery/pkg/fds/updater.go
```diff
@@ -300,21 +311,36 @@ func (u *updaterUpdater) Run() error {
...
+		}
 	}
+	logger := contextutils.LoggerFrom(u.ctx)
 	for _, discoveryForUpstream := range discoveriesForUpstream {
-		err := discoveryForUpstream.DetectFunctions(context.Background(), resolvedUrl, u.dependencies, upstreamSave)
-		if err != nil {
-			return eris.Wrapf(err, "Error doing discovery %T", discoveryForUpstream)
-		}
+		go func(d UpstreamFunctionDiscovery) {
+			for {
```
kdorosh, line 329: let's add a comment to make it clearer that each discovery is responsible for sleeping itself so this isn't spammy (ideally in a configurable amount). a comment on the interface would be nice too

kdorosh: also if none of the discoveries are valid then this is spammy and will run CPU hot (the original intent of 
```golang
go func() {
		updater.Run()
		cancel()
}
```
in `updater.go`.

let's make sure we sleep in the top-level loop at least a little if none of the discoveries were invoked

projects/discovery/pkg/fds/discoveries/swagger/swagger.go
```diff
@@ -187,31 +187,35 @@ func (f *SwaggerFunctionDiscovery) DetectFunctions(ctx context.Context, _ *url.U
...
-		if err := contextutils.Sleep(ctx, f.functionPollTime); err != nil {
+		err = f.detectFunctionsFromSpec(ctx, spec, in, updatecb)
+		if err != nil {
 			return err
 		}
+		return nil
+	})
+	if err != nil {
+		if ctx.Err() != nil {
+			return ctx.Err()
```
kdorosh, line 206: multierror here?

projects/discovery/pkg/fds/discoveries/grpc/grpc.go
```diff
@@ -99,25 +99,29 @@ func (f *UpstreamFunctionDiscovery) DetectType(ctx context.Context, url *url.URL
...
+	})
 
-		if err != nil {
-			if ctx.Err() != nil {
-				return ctx.Err()
-			}
-			// ignore other errors as we would like to continue forever.
+	if err != nil {
+		if ctx.Err() != nil {
+			return ctx.Err()
```
kdorosh, line 109: multierror?

changelog/v1.11.0-beta15/fix-multiple-function-discoveries.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/pull/5998
+    description: Fix bug where when multiple function discoveries apply to a single upstream,
+      only one would run indefinitely.
```
kdorosh, line 5: can we add a unit test of some sort to ensure that this is now fixed?

projects/discovery/pkg/fds/updater.go
saiskee: I'm settling for the design that top level loop sleeps, rather than delegating that to individual discoveries.

kdorosh: @saiskee the old version still had configurable polling intervals per discovery. do we plan on removing that?

saiskee: those configurable polls are used in each discoveries exponential backoff, in the case where the discovery doesn't work due to network failure / pod blip. Those still exist.

saiskee:
added [test](https://github.com/solo-io/gloo/pull/5998/files?diff=split&w=0#diff-18bf42ff87b152088e791a3cfc56cd7cce902153a4e8c5b40289051272f829e1R174) to make sure that multiple discoveries run simultaneously and one doesn't block the other.

projects/discovery/pkg/fds/updater.go
```diff
@@ -300,21 +320,36 @@ func (u *updaterUpdater) Run() error {
...
+					return
+				default:
+					// continue to detect functions, as you were
+				}
+				err := d.DetectFunctions(context.Background(), resolvedUrl, u.dependencies, upstreamSave)
+				if err != nil {
+					logger.Errorf("Error doing discovery %T: %s", d, err.Error())
+					return
+				}
+				time.Sleep(30 * time.Second)
```
kdorosh, line 350: looking over the code again, I was wrong, every implementer of `DetectFunctions` has a sleep so are fine to remove this (i think we should for responsiveness)

kdorosh (dismissed):

kdorosh:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a63c7bd8-ac38-4d2f-8756-ed34e08d360b

kdorosh (dismissed):

kdorosh (dismissed):

saiskee:
Looks like this is failing a legitamate test in build-bot

kdorosh (approved):

Issue #5774, opened 2022-01-11
Doc update for the Prometheus Operator using ServiceMonitor 


artberger:
### Version

master

### Describe the requested changes

See background discussion in Slack: https://solo-io-corp.slack.com/archives/C02KUC45C13/p1641919631067400

Another thread: https://solo-io-corp.slack.com/archives/CHJV572TG/p1648736394403229

Discussion:
* Service monitor approach changed in 1.10, causing issues for some users
* Did this change the annotations, or just an alternative way of configuring it?
* Why do we need to add another endpoint called metrics? Should be able to work with anything?

Related issues:
* https://github.com/solo-io/gloo/issues/1773

### Link to any relevant existing docs

* https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/1-gloo-deployment.yaml#L161-L164
* 1.9 method: https://docs.solo.io/gloo-edge/latest/guides/observability/prometheus/metrics/

PR #6001, opened 2022-03-02
Helm improvements for monitoring

sam-heilbron:
# Description

Improve helm configuration of monitoring resources

# Context

- Expose helm values used to configure monitoring in our external docs
- Configure the monitoring components using more granular configuration (see Breaking Changes below)

### Breaking Changes
These helm changes are breaking because we are relying on new fields to expose ports used for monitoring.

#### Before this Change
- `.Values.global.glooStats.serviceMonitor` was used to expose ports on the underlying services and deployments as well as configure a ServiceMonitor and PodMonitor (Gloo Edge Enterprise only).

#### Why are changes necessary?
1. Security. The `gateway-proxy` service exposes a port on the default service, exposing data plane metrics outside of the cluster.
2. Granularity. All resources (services and deployment ports as well as PodMonitor and ServiceMonitor config) are based on the same boolean. 

#### After this Change
- `Values.global.glooStats.enabled` will continue to define whether a component supports scraping statistics
- `.Values.global.glooStats.serviceMonitorEnabled` is used to expose the monitoring port on the control plane services (to be used with a ServiceMonitor CR)
- `.Values.global.glooStats.podMonitorEnabled` is used to expose the monitoring port on the control plane deployments (to be used with a PodMonitor CR)

### Documentation Follow-Up
There are some further improvements that we can make to our external docs around configuring Gloo Edge to support Service Monitors via Helm. That worked is tracked in [this issue](https://github.com/solo-io/gloo/issues/5774) and will be picked up by the docs team. I feel comfortable allowing that work to be completed in a follow-up since we outline above the breaking change.

# Checklist:


install/helm/gloo/generate/values.go
```diff
@@ -371,6 +371,7 @@ type GatewayProxy struct {
 	EnvoyStaticClusters            []map[string]interface{}     `json:"envoyStaticClusters,omitempty" desc:"List of extra static clusters to be added to envoy bootstrap config. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-msg-config-cluster-v3-cluster"`
 	HorizontalPodAutoscaler        *HorizontalPodAutoscaler     `json:"horizontalPodAutoscaler,omitempty" desc:"HorizontalPodAutoscaler for the GatewayProxy. Used only when Kind is set to Deployment. Resources must be set on the gateway-proxy deployment for HorizontalPodAutoscalers to function correctly"`
 	PodDisruptionBudget            *PodDisruptionBudget         `json:"podDisruptionBudget,omitempty" desc:"PodDisruptionBudget is an object to define the max disruption that can be caused to the gate-proxy pods"`
+	PodMonitorEnabled              *bool                        `json:"podMonitorEnabled,omitempty"`
```
jenshu, line 374: can you add a description for this?

install/helm/gloo/generate/values.go
```diff
@@ -565,8 +567,10 @@ type K8s struct {
 }
 
 type Stats struct {
-	Enabled            *bool   `json:"enabled,omitempty" desc:"Controls whether or not envoy stats are enabled"`
-	RoutePrefixRewrite *string `json:"routePrefixRewrite,omitempty" desc:"The envoy stats endpoint to which the metrics are written"`
+	Enabled               *bool   `json:"enabled,omitempty" desc:"Controls whether or not envoy stats are enabled"`
```
jenshu, line 570: is stats is disabled, can the podmonitor/servicemonitor be enabled independently?

install/helm/gloo/templates/8-gateway-proxy-service.yaml
```diff
@@ -147,7 +164,10 @@ spec:
 ---
 {{- include "gloo.util.merge" (list $ctx $configDumpServiceYamlOverride "gatewayProxy.configDumpServiceSpec") -}}
 {{- end }} {{/* if and ($spec.readConfig $spec.readConfigMulticluster) */}}
+{{- if $spec.service.serviceMonitorEnabled }}
```
jenshu, line 167: does this inherit the global default if it's not set on the gateway proxy?
(can we add a test for that if we don't already have one?)

install/helm/gloo/templates/8-gateway-proxy-service.yaml
```diff
@@ -147,7 +164,10 @@ spec:
...
+---
+{{- include "gatewayProxy.monitoringServiceSpec" $ctx -}}
```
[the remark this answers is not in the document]
sam-heilbron: I dont think we need yaml overrides here, since the service shouldn't need to be modified at all

install/helm/gloo/generate/values.go
sam-heilbron: Not currently, no. The stats.enabled flag controls whether we add the START_STATS_SERVER env variable and allow the stats to be scraped using prometheus annotations. The new serviceMonitorEnabled and podMonitorEnabled flags expose ports on the service and deployment template respectively, but require the stats to be enabled as well (so there is something to scrape). I chose to keep them as separate flags to keep our existing behavior, and allow for granular adoption of the new monitoring features. Open to other suggestions though!

sam-heilbron: I added comments to clarify this, and ensuring the templates respect that logic directly too

install/helm/gloo/templates/8-gateway-proxy-service.yaml
sam-heilbron: RE testing: great call, I missed this! https://github.com/solo-io/gloo/pull/6001/commits/82c837dcee32886c3b4afefac2e267cb7ef7e4d1 Added a test to validate the the podMonitorEnabled value handles overrides correctly. Any other test cases worth adding?

sam-heilbron:
/kick flake https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e82c445e-9265-4a69-bd9c-f0560c8ae2fa

install/helm/gloo/generate/values.go
```diff
@@ -565,8 +565,10 @@ type K8s struct {
...
+	RoutePrefixRewrite    *string `json:"routePrefixRewrite,omitempty" desc:"The envoy stats endpoint to which the metrics are written"`
+	ServiceMonitorEnabled *bool   `json:"serviceMonitorEnabled,omitempty" desc:"Whether or not to expose an http-monitoring port that can be scraped by a Prometheus Service Monitor. Requires that 'stats.enabled' is also true"`
+	PodMonitorEnabled     *bool   `json:"podMonitorEnabled,omitempty" desc:"Whether or not to expose an http-monitoring port that can be scraped by a Prometheus Pod Monitor. Requires that 'stats.enabled' is also true"`
```
jenshu, line 571:
```suggestion
	ServiceMonitorEnabled *bool   `json:"serviceMonitorEnabled,omitempty" desc:"Whether or not to expose an http-monitoring port that can be scraped by a Prometheus Service Monitor. Requires that 'enabled' is also true"`
	PodMonitorEnabled     *bool   `json:"podMonitorEnabled,omitempty" desc:"Whether or not to expose an http-monitoring port that can be scraped by a Prometheus Pod Monitor. Requires that 'enabled' is also true"`
```
nit: the prefix will probably actually be `glooStats` but since this message can be reused in different places maybe we just say `enabled` without the prefix? i think it will be clear that it's referring to the field within the same Stats struct

install/helm/gloo/templates/3-discovery-service.yaml
```diff
@@ -0,0 +1,18 @@
+{{- $statsConfig := coalesce .Values.discovery.deployment.stats .Values.global.glooStats -}}
```
jenshu, line 1: this applies to all the helm templates - do we want to coalesce or merge?
if we do merge, then user could say set `global.glooStats.enabled=true` to enable for everything, then selectively set podMonitorEnabled/serviceMonitorEnabled=true/false on individual services/deployments, without *also* having to set enabed=true. If we use coalesce, since we're using the first nonempty object in the list, this means that user needs to set all fields (both enabled and podMonitorEnabled/serviceMonitorEnabled) on the service/deployment-level one

install/helm/gloo/generate/values.go
sam-heilbron: yeah good call

install/helm/gloo/templates/3-discovery-service.yaml
sam-heilbron: It's a good question, and one that is worth reconsidering in the future. Given that existing deployments (https://github.com/solo-io/gloo/blame/master/install/helm/gloo/templates/5-gateway-deployment.yaml#L3) use coalesce, I opted to keep that behavior in the new features. My concern with switching is that it could be a breaking change for some users.

gunnar-solo (approved):
LGTM

PR #6015, opened 2022-03-03
revert(settings): revert field, (ratelimit --> rateLimit)

gunnar-solo:
# Description


jenshu (approved):

bewebi:
The use of `ratelimit` rather than `rateLimit` is necessary since it needs to match the [CRD](https://github.com/solo-io/gloo/blob/ef66036d1e44b765df00254d74ead11a719f6cdf/install/helm/gloo/crds/gloo.solo.io_v1_Settings.yaml#L464)
It is likely that the existing helm template has been quietly broken for some time

PR #6089, opened 2022-03-15
new page compiling all the timeout settings

bcollard:
# Description

Single documentation page compiling (all?) the timeout settings. This is going to be really handy for FEs!

Doc only.


sam-heilbron (commented):
Can you help me understand what situation this cheatsheet will be more helpful than going to the API definition directly? My only concern is that there will be easy drift (new fields, new defaults) between the protos and this doc.

If our APIs are missing comments with the default values, I'd be in favor of adding those defaults directly to the proto comments (which are used to generate the docs). If there are groupings of timeout settings that should be confiugred together, I'm in favor of a doc outlining how to sovle a problem using those settings in tandem. What do you think?

bcollard:
This comes from my personal notes (on a notepad I used to refer to as needed). So this comes from my field experience, where at some point I had to compile all these settings somewhere to answer questions faster. Some of the defaults are hidden in the Envoy doc. So, having this collection of timeout settings on a single page is really handy on a daily basis, as a Field Engineer. For example: when you want to configure the `idleTimeout`, good luck to find under which API this exists. Of course, this becomes easier over time as a FE, but as an end-user, this is another story. 

AFAICR, these default values never changed over 18 months. 

I was also thinking about referencing this page in the production deployment guidelines. As a platform operator, this helps to check that you've tuned everything you should for better performance and security. 

These are my thoughts and I understand we need to find a consensus.

sam-heilbron:
```
I was also thinking about referencing this page in the production deployment guidelines. As a platform operator, this helps to check that you've tuned everything you should for better performance and security.
```

I like the proposal. This looks good to me but I'll let @solo-io/solo-docs give a final review

bcollard:
ok cool! @solo-io/solo-docs I'll finish the PR on Monday for final review. Thank you!

docs/content/reference/cheatsheets/timeouts/_index.md
```diff
@@ -0,0 +1,118 @@
+---
+title: Timeout defaults
+weight: 60
+description: A global reference to all the timeout settings
+---
+
+This page compiles a non-exhaustive list of timeout settings available in Gloo Edge.
```
artberger, line 7: Is there somewhere we can point them for more information or how to find these themselves? Maybe just the API reference doc?

```suggestion
Review this page for a list of commonly used timeout settings in Gloo Edge, organized by custom resource. For more information, see the [API reference]({{< versioned_link_path fromRoot="/reference/api/" >}}) for each resource.
```

docs/content/reference/cheatsheets/timeouts/_index.md
```diff
@@ -0,0 +1,118 @@
+---
+title: Timeout defaults
+weight: 60
+description: A global reference to all the timeout settings
```
artberger, line 4:
```suggestion
description: Quick reference for timeout settings in Gloo Edge resources.
```

docs/content/reference/cheatsheets/_index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Cheatsheets
+weight: 60
+description: A collection of quick references.
+---
+
+A collection of quick references.
```
artberger, line 7:
```suggestion
You can use cheatsheets to quickly review common settings for Gloo Edge resources, such as setting timeouts. These examples are provided based on real-world usage, but are not meant to be exhaustive of everything that you can set in Gloo Edge.
```

bcollard:
now ready for final review

artberger (approved):

PR #6110, opened 2022-03-17
Stitching Control Plane API

saiskee:
For additional context around [these proto messages](https://github.com/solo-io/gloo/pull/6110/files#diff-99cc29e1ee70cb2e6e5e6c5d5fcd9447deafef6d0e7f8c5cc446b6f801f8dfb8R16-R48):
We are using a javascript library called `graphql-tools` to generate our stitching info. We're implementing it in such a way that graphql-tools will marshal those proto messages from a base64-encoded proto message that the control plane passes to the graphql-tools js script via args. 
By sharing the same definitions between our golang control-plane code and javascript stitching info script, the two are able to communicate with eachother on a shared protobuf contract.


projects/gloo/api/external/envoy/extensions/graphql/stitching.proto
```diff
@@ -0,0 +1,116 @@
...
+option java_outer_classname = "GraphQLStitchingProto";
+option java_multiple_files = true;
+
+import "google/protobuf/any.proto";
+import "validate/validate.proto";
+import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/extensions/graphql/graphql.proto";
+
+
+// ------------- Graphql Tools JS Input -------------
+// This is not user-facing and is only used to pass to the graphql-tools js script
```
saiskee, line 17: Should I put this in a separate proto file?

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -211,7 +226,10 @@ message GraphQLApi {
   // Metadata contains the object metadata for this resource
   core.solo.io.Metadata metadata = 2;
 
-  ExecutableSchema executable_schema = 6;
+  oneof schema {
+    ExecutableSchema executable_schema = 6;
+    GatewaySchema gateway_schema = 7;
```
kdorosh, line 231: why gateway_schema? this might not always be the top level (i.e. gateway) schema; what if we have a stitched schema that is itself composed of other stitched schemas? (e.g., our potential workspace-level schema in mesh)

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -175,6 +175,21 @@ message GrpcResolver {
...
+  message SubschemaConfig {
+    message TypeMergeConfig {
+      string selection_set = 1;
+      string query_name = 2;
+      map<string, string> args = 3;
+    }
+    string name = 1;
+    string namespace = 2;
+    map<string,TypeMergeConfig> type_merge = 3;
+  }
```
kdorosh, line 189: let's add better comments / examples as this is the user-facing yaml API

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/stitching_info.proto
```diff
@@ -0,0 +1,40 @@
...
+
+package graphql.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/enterprise/options/graphql/v1alpha1";
+
+import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/extensions/graphql/stitching.proto";
+
+// ------------- Graphql Tools JS Input -------------
+// This is not user-facing and is only used to pass to the graphql-tools js script
+// This is the message which the graphql-tools js script will consume
```
kdorosh, line 11: should we move this file to a new external directory? (rather than the external envoy one)

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
kdorosh: i imagine you got / copied the name from https://www.graphql-tools.com/docs/schema-stitching/stitch-combining-schemas

saiskee: so let's say we had a stitched schema A that was made of B and C:
```yaml
kind: GraphqlApi
name: A
spec:
  gatewaySchema:
     subschemas:
         - B
         - C
```
and we wanted another gateway schema made up of A and gateway schema D,
```yaml
kind: GraphqlApi
name: super-super-schema
spec:
  gatewaySchema:
     subschemas:
         - A
         - D
```

kdorosh: spoke offline; my concern was pedantic and related to naming, we will update the name

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -175,6 +175,21 @@ message GrpcResolver {
   string span_name = 4;
 }
 
+message GatewaySchema {
+
+  message SubschemaConfig {
+    message TypeMergeConfig {
+      string selection_set = 1;
```
kdorosh, line 182: let's add a comment akin to

> selectionSet specifies one or more key fields required from other services to perform this query. Query planning will automatically resolve these fields from other subschemas in dependency order.

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -175,6 +175,21 @@ message GrpcResolver {
...
+      string query_name = 2;
+      map<string, string> args = 3;
```
kdorosh, line 184: let's add comment akin to

> args formats the initial object representation into query arguments.

projects/gloo/pkg/syncer/translator_syncer.go
```diff
@@ -18,20 +18,22 @@ import (
...
+		reporter   reporter.StatusReporter
+		// used for debugging purposes only
+		latestSnap *v1snap.ApiSnapshot
+		extensions []TranslatorSyncerExtension
+		// used to track which envoy node IDs exist without belonging to a proxy
+		extensionKeys map[string]struct{}
+		settings      *v1.Settings
+		statusMetrics metrics.ConfigStatusMetrics
+	}
+)
```
kdorosh, line 36: why this change?

saiskee: recently changed my goformatter, seems to have done this on accident. will revert.

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
saiskee: updated the name from gateway_schema to `stitched_schema`

kdorosh (dismissed):

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -175,6 +175,93 @@ message GrpcResolver {
...
+          selection_set: '{ username }'
+          query_name: 'GetUser'
+          args:
+            username: username
+      ```
+      the stitched schema will now be able to provide the full user type to all types that require it. In this case,
+      we can now get the first name of an author from the Review.author field even though the Reviews schema doesn't
+      provide the full User type.
+     */
+    map<string,TypeMergeConfig> type_merge = 3;
```
kdorosh, line 258: in the future we may try to simplify / remove unneeded config here to improve the user-experience; but this is fine for a first pass as this feature is currently in alpha

kdorosh: this looks like a pretty clean mapping of https://www.graphql-tools.com/docs/schema-stitching/stitch-type-merging

> Type merging requires a query from each subschema to provide its version of a merged type

There may not be simplifications to make here with respect to type merging, but at least this field can be omitted if types don't need to be merged

kdorosh (approved):

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
saiskee: Let's continue this discussion in the future in https://github.com/solo-io/gloo/issues/6118

kdorosh: spoke offline; starting with what we have here so the responsibility for type merging config falls on the owner of the parent resource rather than child

PR #6123, opened 2022-03-18
ui settings

jenshu:
# Description

Add settings to configure GlooEE UI/Console options.

# Checklist:


projects/gloo/api/v1/settings.proto
```diff
@@ -640,3 +642,15 @@ message GatewayOptions {
     // When these properties are defined on a specific VirtualService, this configuration will be ignored
     VirtualServiceOptions virtual_service_options = 7;
 }
+
+// Settings used by the Enterprise Console (UI)
+message ConsoleOptions {
+    // If true, then custom resources can be created, updated, and deleted via the UI.
+    // If false, then resources can only be viewed in read-only mode.
+    // Defaults to false.
+    google.protobuf.BoolValue write_enabled = 1;
```
MitchAman, line 651: What do you think about `read_only` instead of `write_enabled`? It's the inverse of what you've described in the comment, but (IMO) is a bit more intuitive that [create, update, delete] are all allowed when set to false

jenshu: i actually had it as read_only originally, but wanted to change it to mirror the api_explorer_enabled field, so that you would turn each flag to true to enable the feature. don't have a super strong preference though

install/helm/gloo/generate/values.go
```diff
@@ -32,6 +32,7 @@ type Global struct {
 	IstioIntegration  IstioIntegration `json:"istioIntegration,omitempty" desc:"Configs user to manage Gloo pod visibility for Istio's' automatic discovery and sidecar injection."`
 	ExtraSpecs        *bool            `json:"extraSpecs,omitempty" desc:"Add additional specs to include in the settings manifest, as defined by a helm partial. Defaults to false in open source, and true in enterprise."`
 	ExtauthCustomYaml *bool            `json:"extauthCustomYaml,omitempty" desc:"Inject whatever yaml exists in .Values.global.extensions.extAuth into settings.spec.extauth, instead of structured yaml (which is enterprise only). Defaults to true in open source, and false in enterprise"`
+	Console           *ConsoleOptions  `json:"console,omitempty" desc:"Configuration options for the Enterprise Console (UI)."`
```
sam-heilbron, line 35: For extauth, we use a trick of defining an interface{} for extensions, and then actually typing that in enterprise. We inject the settings here: https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/18-settings.yaml#L125. We could potentially follow that paradigm here and allow the ConsoleOptions to be defined in the same place as other UI settings: https://github.com/solo-io/solo-projects/blob/master/install/helm/gloo-fed/generate/values.go#L25

projects/gloo/api/v1/settings.proto
jenshu: changed to read_only

install/helm/gloo/generate/values.go
jenshu: changed to interface and defined options in SP

MitchAman (approved):

Issue #6112, opened 2022-03-17
glooctl check should print warning when client and server versions don't match


jenshu:
### Is your feature request related to a problem? Please describe.

Users often run into issues when their glooctl version doesn't match the gloo version they have installed (see #6111 for details). 

### Describe the solution you'd like

We should have `glooctl check` print a warning at the bottom if the client and server versions don't match (`glooctl version` does something similar already)

Issue #6108, opened 2022-03-17
[ci test flake] Gateway in memory hybrid gateway tcp ssl [It] should work with ssl 


jenshu:
### Gloo Edge Version

1.11.x (beta)

### Describe the bug

example logs:

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=f7ae4c3f-cdcb-40ef-8fbf-3ffef13923d8

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=9f775840-517a-460f-aa92-1484e6904a25

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4dd7e49e-7782-4d7f-a6e8-2d60ce747b52

failure on different ssl test but possibly related?

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=89162e79-f7b5-40d5-88e7-cfaf3b4ad3ae

### Steps to reproduce the bug

This often occurs during test runs in ci

### Expected Behavior

Tests should pass consistently

sam-heilbron:
[Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=b8084cee-0ab4-4813-b3bc-ede82a332f8a

sam-heilbron:
[Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=fbb1c719-930b-43b4-883b-475b13b4199c

MitchAman:
[Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=cc98c058-a286-486d-b0a1-ef900bdc76e6

inFocus7:
[Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d1cbc61a-edad-42de-9987-31b4113e9433

inFocus7:
[Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher

[...]

PR #6125, opened 2022-03-19
Client server glooctl check

kevin-shelaga:
# Description

Add extra check to verify that the client and server versions match. If they do not match, display a warn and suggest updating the client to match the server versions. Right now I am checking the gateway version as its the only container tag that is open source and thus should be the oss equivalent for Gloo ee.

```sh
Checking deployments... OK
Checking pods... OK
Checking upstreams... OK
Checking auth configs... OK
Checking RouteOptions... OK
Checking secrets... OK
Checking gateways... OK
Checking proxies... OK
Checking rate limit server... OK
No problems detected.
Skipping Gloo Instance check -- Gloo Federation not detected

WARN: Client and Server versions do not match. Consider running `glooctl upgrade --release=v1.9.6`
```

# Context

Fixes: https://github.com/solo-io/gloo/issues/6112

# Checklist:

### DIFF
--- a/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
+++ b/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/options"
+	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/version"
 	"github.com/solo-io/gloo/projects/gloo/cli/pkg/helpers"
 	glooinstancev1 "github.com/solo-io/solo-apis/pkg/api/fed.solo.io/v1"
 	"github.com/solo-io/solo-apis/pkg/api/fed.solo.io/v1/types"
@@ -21,6 +22,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
+
+func CheckVersionsMatch(opts *options.Options) {
+	vrs, _ := version.GetClientServerVersions(opts.Top.Ctx, version.NewKube(opts.Metadata.GetNamespace()))
+
+	clientVersionStr := vrs.GetClient().GetVersion()
+
+	serverVersionStr := ""
+	for _, v := range vrs.GetServer() {
+		for _, cvr := range v.GetKubernetes().Containers {
+			if cvr.GetName() == "gateway" {
+				serverVersionStr = cvr.GetTag()
+				break
+			}
+		}
+		if serverVersionStr != "" {
+			break
+		}
+	}
+
+	if clientVersionStr != serverVersionStr {
+		printer.AppendMessage(fmt.Sprintf("\nWARN: %s\n", "Client and Server versions do not match. Consider running `glooctl upgrade --release=v"+serverVersionStr+"`"))
+	}
+}
 
 func CheckMulticlusterResources(opts *options.Options) {
 	// check if the gloo fed deployment exists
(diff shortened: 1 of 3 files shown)


projects/gloo/cli/pkg/cmd/check/gloo_instances.go
```diff
@@ -22,6 +23,29 @@ import (
...
+				serverVersionStr = cvr.GetTag()
+				break
+			}
+		}
+		if serverVersionStr != "" {
+			break
+		}
+	}
+
+	if clientVersionStr != serverVersionStr {
```
jenshu, line 44: if they're not running in gateway mode then the serverVersionStr may be empty. should we just not warn in that case?

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
```diff
@@ -22,6 +23,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
+func CheckVersionsMatch(opts *options.Options) {
```
jenshu, line 26: nit: this file seems to be Gloo Fed related. can we move this func to the main root.go or a new file?

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
```diff
@@ -22,6 +23,29 @@ import (
...
+				break
+			}
+		}
+		if serverVersionStr != "" {
+			break
+		}
+	}
+
+	if clientVersionStr != serverVersionStr {
+		printer.AppendMessage(fmt.Sprintf("\nWARN: %s\n", "Client and Server versions do not match. Consider running `glooctl upgrade --release=v"+serverVersionStr+"`"))
```
jenshu, line 45: should we also print what the current client version is?

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
kevin-shelaga: good catch

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
kevin-shelaga: I updated the message now

```
WARN: Client (v1.10.16) and Server (v1.9.6) versions do not match.
Consider running `glooctl upgrade --release=v1.9.6` to upgrade your client to match the deployed server version.
```

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
kevin-shelaga: Good call, moved to root

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -927,3 +931,30 @@ func isCrdNotFoundErr(crd crd.Crd, err error) bool {
 		return false
 	}
 }
+
+func CheckVersionsMatch(opts *options.Options) {
+	vrs, _ := version.GetClientServerVersions(opts.Top.Ctx, version.NewKube(opts.Metadata.GetNamespace()))
```
jenshu, line 936: we should handle errors here and just return early (looks like errors are only thrown when we can't get the server version)

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -927,3 +931,30 @@ func isCrdNotFoundErr(crd crd.Crd, err error) bool {
...
+
+	clientVersionStr := vrs.GetClient().GetVersion()
+
+	glooTypeGateway := false
+	serverVersionStr := ""
+	for _, v := range vrs.GetServer() {
```
jenshu, line 942: hmm we should probably handle multiple server versions (i.e. multiple gloo installations). do we want to print a warning for every mismatch or just the first mismatch we find?

jenshu (dismissed):
LGTM!

jenshu:
you may need to run gofmt

jenshu (approved):

sam-heilbron:
/kick [Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher](https://github.com/solo-io/gloo/issues/6108#issuecomment-1075050891)

jenshu (approved):

Issue #5101, opened 2021-08-02
Validating webhook broken when reinstalling Gloo Edge


guydc:
**Describe the bug**
When Gloo is uninstalled and reinstalled with helm, the validating webhook client config CA bundle is empty. As a result, the API server does not trust the webhook server cert. Admission is either skipped or all client operations are rejected (depending on Webhook Failure Policy). 

The issue did not reproduce with helm upgrade of Gloo, and can be worked around by manually removing Gloo resources that are not cleaned-up during uninstall.  

**To Reproduce**
1. install Gloo and verify CA Bundle is configured:
```
❯ helm install gloo gloo/gloo --version=1.8.0 --namespace test-ns --set gateway.validation.alwaysAcceptResources=false
NAME: gloo
NAMESPACE: test-ns
STATUS: deployed
REVISION: 1
❯ kubectl get validatingwebhookconfigurations gloo-gateway-validation-webhook-test-ns -o yaml | yq eval '.webhooks[]' -
admissionReviewVersions:
  - v1beta1
clientConfig:
  caBundle: [...]
  service:
    name: gateway
    namespace: test-ns
    path: /validation
    port: 443
failurePolicy: Ignore
matchPolicy: Exact
name: gateway.test-ns.svc
namespaceSelector: {}
objectSelector: {}
rules:
  - apiGroups:
      - gateway.solo.io
    apiVersions:
      - v1
    operations:
      - CREATE
      - UPDATE
      - DELETE
    resources:
      - '*'
    scope: '*'
sideEffects: None
timeoutSeconds: 30
```
2. uninstall gloo, note that the validating webhook and validation certs still exist:

```
❯ helm uninstall gloo --namespace test-ns
release "gloo" uninstalled
❯ kubectl  get validatingwebhookconfigurations gloo-gateway-validation-webhook-test-ns
NAME                                      WEBHOOKS  
gloo-gateway-validation-webhook-test-ns 
❯ kubectl  get secret gateway-validation-certs -n test-ns
NAME                       TYPE                DATA   
gateway-validation-certs   kubernetes.io/tls   3
```

3. Reinstall Gloo, see that the Validating webhook resource exists, but the CA Bundle is removed
```
❯ helm install gloo gloo/gloo --version=1.8.0 --namespace test-ns --set gateway.validation.alwaysAcceptResources=false
NAME: gloo
NAMESPACE: test-ns
STATUS: deployed
REVISION: 1
[...]

PR #6126, opened 2022-03-21
ensure update CA bundle in validating webhook configuration

mordabi:
# Description

This fix should update the CA bundle field in the validating webhook resource in case the secret is exist but the validating webhook does not contain the secret reference.

# Context

Users ran into this bug after delete gloo using helm and then reinstall it ,the CA certs secret already exist in the cluster but the webhook does not contain the CaBundle secret reference field  

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5101


TomerJLevy:
Hi, @sam-heilbron and @elcasteel, can you please help push this item forward?

changelog/v1.11.0-beta22/fix-certgen-CaBundle.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+- type: NON_USER_FACING
+  resolvesIssue: false
+  description: this fix should update the CA bundle field in the validating webhook resource in case the secret is exist but the validating webhook does not contain the secret reference.
```
elcasteel, line 4: Move this file to v1.11.0-beta24 and also start the description with a capital letter.

jobs/pkg/run/run.go
```diff
@@ -78,12 +78,12 @@ func Run(ctx context.Context, opts Options) error {
 			contextutils.LoggerFrom(ctx).Infow("existing TLS secret found, skipping update to TLS secret and ValidatingWebhookConfiguration since the old TLS secret is still existAndValid",
 				zap.String("secretName", secretConfig.SecretName),
 				zap.String("secretNamespace", secretConfig.SecretNamespace))
-			return nil
 		}
 	}
-
-	if err := kube.CreateTlsSecret(ctx, kubeClient, secretConfig); err != nil {
-		return eris.Wrapf(err, "failed creating secret")
+	if !existAndValid{
```
elcasteel, line 83:
```suggestion
	if !existAndValid {
```

elcasteel: From codegen check

changelog/v1.11.0-beta24/fix-certgen-Cabundle.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+- type: NON_USER_FACING
+  resolvesIssue: false
+  description: This fix should update the CA bundle field in the validating webhook resource in case the secret is exist but the validating webhook does not contain the secret reference.
```
sam-heilbron, line 4: Given that this resolves a bug for a certain scenario, would this be a FIX? Could we create a GitHub issue to describe the behavior so that other users can track what version this fix was added to?

jobs/pkg/run/run.go
```diff
@@ -67,9 +67,9 @@ func Run(ctx context.Context, opts Options) error {
 		Cert:               caCert,
 		CaBundle:           certs.CaCertificate,
 	}
-
+	existAndValid := false
```
sam-heilbron, line 70: Could we add a test to validate this change? I could be misunderstanding the logic but what does this change?

From my understanding the following behavior takes place before and after your change:

| forceRotation      | secretExistsAndValid | createNewSecret? |
| ----------- | ----------- |  ----------- | 
| false      | true       | FALSE |
| false   | false        | TRUE |
| true      | true       | TRUE |
| false   | false        | TRUE |

Could you help me understand which of these test cases (or a new case that I didn't include) changes?

sam-heilbron (commented):
thanks for the contribution!

jobs/pkg/run/run.go
TomerJLevy: Thanks for pushing this fix.
The table is almost correct, I guess you meant the last value of `forceRotation` is also true and then it will be correct.
But this PR doesn't affect the creation of a new secret, it should ensure the CA Bundle is set in the ValidatingWebhookConfiguration.

Currently, if the existing secret is valid, it skips the remaining of this function and therefore in case the CA Bundle wasn't set before it won't set it.

And as @mordabi mentioned, this is the case when you run `helm uninstall` and then install again, since the secret remains in the K8S cluster and the secret is still valid, but no one assign this CA Bundle to the ValidatingWebhookConfiguration

sam-heilbron (commented):
Thanks so much for the clarification! Can you move the changelog to the 1.11.0-rc2 directory, and then this LGTM

mordabi:
> Thanks so much for the clarification! Can you move the changelog to the 1.11.0-rc2 directory, and then this LGTM

done

sam-heilbron (approved):

elcasteel:
Looks good, this will merge automatically after codegen runs.

Issue #5949, opened 2022-02-18
validation_gateway_solo_io_valid_config missing from metrics


vladimir-cmd:
### Gloo Edge Version

1.10.x (latest stable)

### Kubernetes Version

1.19.x

### Describe the bug

When I run curl http://10.15.38.26:9091/metrics, I'm getting most of the metrics except the following metrics which are missing and one of them being validation_gateway_solo_io_valid_config:
```
...snip...
runtime_goroutines 120
# HELP setup_gloo_solo_io_emitter_resources_in The number of resource lists received on open watch channels
# TYPE setup_gloo_solo_io_emitter_resources_in counter
setup_gloo_solo_io_emitter_resources_in{namespace="gloo-system",resource="settings"} 4
# HELP setup_gloo_solo_io_emitter_snap_in Deprecated. Use [setup.gloo.solo.io/emitter/resources_in](http://setup.gloo.solo.io/emitter/resources_in). The number of snapshots updates coming in.
# TYPE setup_gloo_solo_io_emitter_snap_in counter
setup_gloo_solo_io_emitter_snap_in 4
# HELP validation_gateway_solo_io_certificate_reload_success Number of successful certificate reloads
# TYPE validation_gateway_solo_io_certificate_reload_success counter
validation_gateway_solo_io_certificate_reload_success 1
```
I'm configuring this metric on Gateway Pod.

### Steps to reproduce the bug

Gloo: 1.10.1
Gloo Enterprise: 1.10.3

When I run curl http://10.15.38.26:9091/metrics, I'm getting most of the metrics bar last two, one of them being validation_gateway_solo_io_valid_config:
```
...snip...
runtime_goroutines 120
# HELP setup_gloo_solo_io_emitter_resources_in The number of resource lists received on open watch channels
# TYPE setup_gloo_solo_io_emitter_resources_in counter
setup_gloo_solo_io_emitter_resources_in{namespace="gloo-system",resource="settings"} 4
# HELP setup_gloo_solo_io_emitter_snap_in Deprecated. Use [setup.gloo.solo.io/emitter/resources_in](http://setup.gloo.solo.io/emitter/resources_in). The number of snapshots updates coming in.
# TYPE setup_gloo_solo_io_emitter_snap_in counter
setup_gloo_solo_io_emitter_snap_in 4
# HELP validation_gateway_solo_io_certificate_reload_success Number of successful certificate reloads
# TYPE validation_gateway_solo_io_certificate_reload_success counter
[...]

PR #6132, opened 2022-03-21
Initialize the validation_gateway_solo_io_valid_config metric to 1

MitchAman:
# Description

Initialize the `validation_gateway_solo_io_valid_config` metric when the gateway pod starts up.

# Context

`validation_gateway_solo_io_valid_config` will not appear on the gateway pod's `/metrics` endpoint until a value is assigned to it. This PR initializes the metric based on the first snapshot that the gateway pod receives after starting. Previously, the metric would not have been initialized until a resource was applied and validated by the webhook.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5949


saiskee (approved):

projects/gateway/pkg/validation/validator.go
```diff
@@ -115,13 +117,16 @@ func NewValidatorConfig(translator translator.Translator, validationClient valid
...
-	return &validator{
+	v := &validator{
 		translator:                   cfg.translator,
 		validationClient:             cfg.validationClient,
 		writeNamespace:               cfg.writeNamespace,
 		ignoreProxyValidationFailure: cfg.ignoreProxyValidationFailure,
 		allowWarnings:                cfg.allowWarnings,
 	}
+	// Initialize validation_gateway_solo_io_valid_config to 1 (valid)
+	utils2.Measure(cfg.ctx, mValidConfig, 1)
```
jenshu, line 128: i don't think we should be setting a value if nothing has been validated yet

MitchAman:
> The issue seems to indicate that the valid_config metric never gets a value assigned to it, even if the configuration is valid. Have we been able to how this behavior changed from previous versions that worked to 1.10?
> 
> Also, what steps did you take to validate this change and confirm that configuration can go from good -> bad -> good and the metric will change as well? It could be helpful to add a test to validate this functionality given that it's a useful metric for users to report on

AFAICT there is no behavior change. It seems this is intermittently reproducible. I did manage to see that the valid_config metric was not present after install at least once. After applying a resource to be validated, the metric then showed up. If the webhook is not enabled, it's possible the metric never gets a value assigned to it.

I just ran the through the following steps to see that the metric is indeed working:

```
$ kind create cluster
```
```
$ glooctl version
Client: {"version":"1.10.17"}
Server: version undefined, could not find any version of gloo running
```
```
$ glooctl install gateway
```
```
$ kubectl port-forward -n gloo-system deploy/gateway 9091:9091
```
```
$ curl "http://localhost:9091/metrics?match[]=\{gloo!=''\}" | grep validation_gateway_solo_io_valid_config
# HELP validation_gateway_solo_io_valid_config A boolean indicating whether gloo config is valid
# TYPE validation_gateway_solo_io_valid_config gauge
validation_gateway_solo_io_valid_config 1
```

If I then kill the gateway pod (and restarted the port forwarding), the metric disappears:
```
$ curl "http://localhost:9091/metrics?match[]=\{gloo!=''\}" | grep validation_gateway_solo_io_valid_config
```
(no response - I checked in browser, and the metric was not present)

I can then proceed to create valid resources, and the metric is still missing. However, if I set `allowWarnings=false` and apply an invalid resource (warning, not error), then it reappears with its value changed to 0.
```
$ kubectl apply -f https://raw.githubusercontent.com/solo-io/solo-blog/main/prometheus-auth-intg/httpbin-vs-original.yaml
``` 
```
$ curl "http://localhost:9091/metrics?match[]=\{gloo!=''\}" | grep validation_gateway_solo_io_valid_config
# HELP validation_gateway_solo_io_valid_config A boolean indicating whether gloo config is valid
# TYPE validation_gateway_solo_io_valid_config gauge
validation_gateway_solo_io_valid_config 0
```

projects/gateway/pkg/validation/validator.go
MitchAman: How would we feel about setting it to -1?

In my comment [here](https://github.com/solo-io/gloo/pull/6132#issuecomment-1075139423), I'm noticing the metric only seems to appear once its value is **changed**. If it were initialized to -1, then the first validation run would change it to 0 or 1

elcasteel:
Currently the `mValidConfig` is set from `validateSnapshot` which is only called from the validating webhook which means that it's not going to be set until a new resource is applied. We might want to set it during `Sync` (at the very least on the first call to `Sync` when `latestSnapshot` is nil

MitchAman:
> Did you confirm that the metric doesn't disappear with the new approach?

Yes. I re-ran through the steps in [this comment](https://github.com/solo-io/gloo/pull/6132#issuecomment-1075139423). This time around when I killed the gateway pod, the metric was present when it came back up :+1:

> Also, do we want the metric to work when the webhook is disabled? If so, it might make sense to set it from `Sync()` instead of from `validateSnapshot` although I'm not sure if you considered that when this was originally implemented.

I did give that some consideration, but I decided to leave the semantics of the metric alone. This PR addresses the core issue of `validation_gateway_solo_io_valid_config` not always being present on the `/metrics` endpoint.

In the mean time, we do have `configStatusMetricLabels` within the [ObservabilityOptions](https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/settings.proto.sk/#observabilityoptions) which, if configured, provide configuration status metrics that don't rely on the webhook being enabled.

elcasteel (dismissed):

projects/gateway/pkg/validation/validator.go
```diff
@@ -146,6 +146,18 @@ func (v *validator) Sync(ctx context.Context, snap *v1.ApiSnapshot) error {
 	v.lock.Lock()
 	defer v.lock.Unlock()
 
+	// When the pod is first starting (aka the first snapshot is received),
+	// set the value of mValidConfig with respect to the translation loop above.
+	// Without this, mValidConfig will not be exported on /metrics until a new
+	// resource is applied (https://github.com/solo-io/gloo/issues/5949).
+	if v.latestSnapshot == nil {
```
sam-heilbron, line 153: Perhaps I was misunderstanding the bug, but I thought the issue was that the mValidConfig metric wasn't incrementing even after a resource was applied. Was that not the case?

elcasteel (approved):

sam-heilbron:
/kick hybrid gateway flake https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=0da02f9d-0d78-4205-8df2-1e86a15987e6

PR #6139, opened 2022-03-22
fix(docs): removed patch versions from doc URLs

gunnar-solo:
# Description
Changed a high-level Hugo routing file such that website paths no longer include patch version information

related https://github.com/solo-io/doctopus/issues/9

# Context
With the way we manage our docs site, we occasionally get very specific google results pointing to docs pages which are out of date (because we've change where they are hosted based on minor versions)

# Checklist:


docs/build-docs.sh
```diff
@@ -99,6 +99,13 @@ function generateHugoVersionsYaml() {
   do
     echo "  - $hugoVersion" >> $yamlFile
   done
+
+  # copy temporary file line by line, pruning path versions from lines that have them
```
sam-heilbron, line 103: Can you share an example of what the lines look like? Is there a way to modify what is being written to the original file instead of relying on this post-processing?

PR #6151, opened 2022-03-24
Graphql max query depth

saiskee:


projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
```diff
@@ -462,6 +462,19 @@ message Executor {
...
     bool enable_introspection = 2;
+
+    // The max amount of nesting a query can be executed against this schema.
+    // e.g. the following query has these depths:
+    // query {         # Depth: 0
+    // me {            # Depth: 1
+    //     friends {   # Depth: 2
+    //        friends  # Depth: 3
+    //     }
+    //   }
```
jenshu, line 473: should this be indented another level?

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -389,6 +389,49 @@ message ExecutableSchema {
...
+      //       "manager" : {
+      //         "name" : "Manager 1",
+      //         "manager"  : {
+      //           "name" : null
+      //   }}}},
+      //   "errors": [
+      //      {"message": "field 'name' exceeds the max operation depth of 3 for this schema"}
+      //    ]
+      // }
+      uint32 max_depth = 1;
```
kdorosh, line 432: thoughts on making this a google wrapper uint32 value? https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto#L96

mostly so that when we serialize / deserialize we can keep noisy config to a minimum

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
```diff
@@ -462,6 +462,19 @@ message Executor {
...
+    // query {         # Depth: 0
+    //   me {          # Depth: 1
+    //     friends {   # Depth: 2
+    //        friends  # Depth: 3
+    //     }
+    //   }
+    // }
+    // If the max_depth is set to 2, then the query at depth 3 will receive an error as a response.
+    // An unset max_depth value (default) or 0 for the max_depth will be unbounded query depth.
+    google.protobuf.UInt32Value max_depth = 3;
```
kdorosh, line 477: i think you changed the wrong one here (meant to change CP API not dataplane)

kdorosh (changes requested):
i think you made a tiny mixup

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
jenshu: should we change both?

kdorosh: i don't think we should; the dataplane api change would require changes to our already released envoy and it's not user-facing, which was the reason for proposing the simplification to the api

kdorosh: this proto must match the one here in gloo https://github.com/solo-io/envoy-gloo-ee/blob/8e4d1c90737efa15c327b93e72e6fc2099da98a0/api/envoy/config/filter/http/graphql/v2/graphql.proto#L486 we are effectively manually vendoring it

jenshu: oops realized i added my original comment to the wrong file - meant to do it for the control plane one

ok this one can be reverted

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
```diff
@@ -379,7 +379,7 @@ message CacheControl {
   // - root fields (i.e. Query, Mutation, Subscription) default to 0s
   // - non-root, non-scalar fields (i.e. object, interface, or union; or a list of those types) default to 0s
   // - all other fields inherit the max_age from their parent.
-  google.protobuf.UInt32Value max_age = 1;
+  uint32 max_age = 1;
```
jenshu, line 382: unless i'm reading this incorrectly, this one should be reverted to match the data plane api, and the max_depth below should be the one to be changed to uint32

saiskee: I changed the wrong one, my bad.

projects/gloo/api/v1/enterprise/options/graphql/v1alpha1/graphql.proto
```diff
@@ -389,6 +389,50 @@ message ExecutableSchema {
...
+      // ```json
+      // { "data" : {
+      //     "employee" : {
+      //       "manager" : {
+      //         "name" : "Manager 1",
+      //         "manager"  : {
+      //           "name" : null
+      //   }}}},
+      //   "errors": [
+      //      {"message": "field 'name' exceeds the max operation depth of 3 for this schema"}
```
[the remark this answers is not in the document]
saiskee: ah this is a good idea. The impl in the data plane has already been done, so this would require a data plane change.

projects/gloo/api/external/envoy/extensions/graphql/graphql.proto
```diff
@@ -462,6 +462,19 @@ message Executor {
...
+    // e.g. the following query has these depths:
+    // query {         # Depth: 0
+    //   me {          # Depth: 1
+    //     friends {   # Depth: 2
+    //        friends  # Depth: 3
+    //     }
+    //   }
+    // }
+    // If the max_depth is set to 2, then the query at depth 3 will receive an error as a response.
+    // An unset max_depth value (default) or 0 for the max_depth will be unbounded query depth.
```
jenshu, line 476: nit: update comment, can't be unset anymore since we switched it back to uint32

Issue #6163, opened 2022-03-25
Transformation caching logic crashes gloo pod


saiskee:
### Gloo Edge Version

1.11.x (beta)

### Kubernetes Version

1.22.x

### Describe the bug

https://github.com/solo-io/gloo/blob/c2e025728df3c66c67275ac718e251a275d32bd3/projects/gloo/pkg/plugins/transformation/plugin.go#L309

`err` may be nil here so this is causing a nil interface conversion panic.

### Steps to reproduce the bug

na

### Expected Behavior

it does not crash

sam-heilbron:
This was introduced in Gloo Edge OSS 1.11.0-beta18. It will affect anyone using transformations. This will be resolved in Gloo Edge 1.11.0-rc3

PR #6164, opened 2022-03-25
plugins/transformations: Update error handling logic on successful transforms

nfuden:
# Description

Nillity check safety

# Context

Repeated calls with the same valid transform ran into a type conversion between nil interfaces.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6163


projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -307,6 +307,8 @@ func (p *Plugin) validateTransformation(
 	if err, ok := p.validationLruCache.Get(transformHash); ok {
 		// Error may be nil here since it's just the cached result
 		return err.(error)
```
elcasteel, line 309: do you need to delete this return?

nfuden: LOLOL i was testing and forgot to remove good catch

nfuden:
Updating for better testing per conversation

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -306,7 +306,9 @@ func (p *Plugin) validateTransformation(
 	// This transformation has already been validated, return the result
 	if err, ok := p.validationLruCache.Get(transformHash); ok {
 		// Error may be nil here since it's just the cached result
-		return err.(error)
+		// so return it as a nil err after cast worst case.
+		errCasted, _ := err.(error)
+		return errCasted
```
EItanya, line 311: The other thing we can do is type cast error when putting it into the cache. Could be one extra line of protection.

EItanya (commented):
Can we mock the validate function so we can also test errors?

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -61,7 +61,8 @@ type Plugin struct {
 	TranslateTransformation    TranslateTransformationFn
 	settings                   *v1.Settings
 
-	// validationLruCache is a map of: (transformation hash) -> (error state)
+	// validationLruCache is a map of: (transformation hash) -> error state
+	// where error state is of type interface{} on return
```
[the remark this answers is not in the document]
nfuden: Fair point on the semantics there. How do you feel about the updated description?

sam-heilbron (commented):
LGTM. I'd be in favor of casting the error before placing it in the cache, and then just returning the cached value, but this works as well

Implausiblyfun:
It may not be crazy to cast prior to adding to the cache as well but from an ongoing safety perspective it probably doesn't hurt to keep this on access to future proof against other entries to the cache.
While making the input clean is good the conversion is quick and prevents panics.

sam-heilbron:
Yeah that's fair. Given that, I think this is good to go as is

sam-heilbron (approved):

PR #6172, opened 2022-03-28
Feature/ginkgo 2

nfuden:
# Description

Naive attempt to update to ginkgo v2. Opened to run ci to make sure there arent any gotchas.

Note this is not currently a prioritized issue so it may take a while to complete.

Steps that will have to be taken in concert here: 
- update cloudbuild to have ginkgo 2
- make branch of solo-kit to have ginkgo v2
- maybe some other things (tbd)
- rebase as we are taking off keep pr updated to make sure that we dont have too much CI overhead


EItanya (commented):
:tada

yuval-k:
> update cloudbuild to have ginkgo 2

this may not be strickly needed, we can vendor ginkgo like we do in GME or run it with `go run github.com/onsi/ginkgo/v2/ginkgo`

sam-heilbron:
@nfuden given that this is out of date and requires some other repos to be updated as well, do we want to continue or pause this and remove the 'keep pr updated' label?

nfuden:
https://github.com/solo-io/gloo/pull/6172#issuecomment-1147961013

We already removed keep pr updated a while ago but it looks like changelog bot is still being annoying. 
Already have the branch so I will close this branch as its a bit out of date. 
It does outline most of the approach so we can revisit it whenever we actually have time

PR #6188, opened 2022-03-30
Added Default Value to SameSite

jackstine:
# Description

Added optional DefaultMode for SameSite enum options. Will act as same value for Lax. More information can be found [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite).

# Context

Before Lax had to be the default option. This would either force or ignore LaxMode as an option entirely. Now Users can use Lax as an option, and still have the option to have SameSite set as Lax optionally, when DefaultMode is selected.

# Checklist:

- [X] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.


gunnar-solo:
/kick [Step #10 - "test": [Fail] Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=2b30f017-7039-4a9c-8c74-858d8b2b9f42)

gunnar-solo (approved):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -351,14 +351,16 @@ message UserSession {
         google.protobuf.StringValue path = 3;
         // The SameSite options. The default value is LaxMode.
         enum SameSite {
+            // Default Mode is the same as LaxMode but will not show up in the Cookie Header. This value is ignored.
+            DefaultMode = 0;
```
nfuden, line 355: I like the comment but in code writing checking for samesite.defaultmode wont show the comment nicely so I wonder if there isnt a different name here that would be more descriptive.
Perhaps something along the lines of DontAddMode

jackstine: Check out the go.http package [here](https://pkg.go.dev/net/http#SameSite)
`const (
	SameSiteDefaultMode SameSite = iota + 1
	SameSiteLaxMode
	SameSiteStrictMode
	SameSiteNoneMode
)`

jackstine:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=0f74fe04-125c-4107-931c-bfb0463b642a

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
nfuden: Works for me

tilgovi:
Sorry to revive an old issue, but the comment around this code still says that LaxMode is the default. It what sense is it the default if the zero value for this enum changed to DefaultMode?

Issue #5202, opened 2021-08-19
CI flakes in regression gateway test on RouteOption


artberger:
**Describe the bug**

From the [logs](https://github.com/solo-io/gloo/runs/3375224676):
```
...
{"level":"info","ts":1629401253.8677585,"logger":"proxy-validator","caller":"validation/server.go:190","msg":"proxy validation report result: listener_reports:{http_listener_report:{virtual_host_reports:{route_reports:{}}}} listener_reports:{http_listener_report:{}}"} {"level":"info","ts":1629401254.845505,"logger":"proxy-validator","caller":"validation/server.go:179","msg":"received proxy validation request"} {"level":"info","ts":1629401254.8462656,"logger":"proxy-validator.translator","caller":"translator/translator.go:177","msg":"computing envoy resources for listener: listener-::-8080"} {"level":"info","ts":1629401254.8465867,"logger":"proxy-validator.translator","caller":"translator/translator.go:177","msg":"computing envoy resources for listener: listener-::-8443"} {"level":"info","ts":1629401254.8477402,"logger":"proxy-validator","caller":"validation/server.go:190","msg":"proxy validation report result: listener_reports:{http_listener_report:{virtual_host_reports:{route_reports:{}}}} listener_reports:{http_listener_report:{}}"}
LOGS FROM gloo-system.testrunner: STARTING SLEEP! Access me.
** End Kubernetes Dump ** 
[AfterEach] tests with RouteOptions
  /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:960
[AfterEach] Kube2e: gateway
  /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:186
make: *** [run-tests] Error 1
SSSSSSSS
JUnit report was created: /home/runner/work/gloo/gloo/test/kube2e/gateway/junit.xml


Summarizing 1 Failure:

[Fail] Kube2e: gateway tests with RouteOptions [It] correctly delegates options from RouteOption 
/home/runner/go/pkg/mod/github.com/solo-io/go-utils@v0.21.16/testutils/assertions.go:27

Ran 13 of 21 Specs in 183.525 seconds
FAIL! -- 12 Passed | 1 Failed | 0 Pending | 8 Skipped
--- FAIL: TestGateway (183.81s)
FAIL

Ginkgo ran 2 suites in 14m46.640050239s
Test Suite Failed
Makefile:148: recipe for target 'run-tests' failed
[...]

PR #6211, opened 2022-03-31
Reduce test flakes

sam-heilbron:
# Description

Cleanup some tests to reduce flakes as well as make them more readable

# Context

## Envoy Instance Cleanup
The e2e tests rely on an Envoy Service, which exposes a Clean method. The method [returned an error](https://github.com/solo-io/gloo/blob/4aa5b68e54a77eac8b856f2fcc55cbfd58da9957/test/services/envoy.go#L318) even though it was always nil. I removed the returned error and cleanup up the usage.

## Hybrid Gateway Flakes (https://github.com/solo-io/gloo/issues/6108)

I believe these issues were due to our e2e tests creating a default hybrid gateway that was invalid. A HybridGateway is configured with an inline definition of an HttpGateway. If there are no VirtualServices, then the HybridGateway is considered invalid and is rejected. Since tests would first write the Gateway (invalid originally) and then write the VirtualService, this test would frequently fail.

What I'm unclear on, is why this test passes every now and then. I spent a little time investigating and then decided to clean up the order of operations and confirm that resource are accepted consistently.

## Route/VirtualHost Option Flakes (https://github.com/solo-io/gloo/issues/5202)

These tests work by applying some resources, waiting for them to be accepted and then validating the state of the Proxy. Since VirtualHostOptions and RouteOptions do not have their statuses modified when they have been processed, we do not have an easy way of identifying when the Gloo controller has processed them (normally we would wait for the INputResource to have an accepted status). Therefore, we must validate the Proxy state in an Eventually, since the first time the Proxy is valid, it might not have processed the RouteOptions or VirtualHostOptions

## Gateway E2E Test Cleanup

Some of these tests wrote resources inside tests and did not clean them up. I tried to move resource creation and cleanup to BeforeEach and AfterEach, to follow more of an AAA model (https://java-design-patterns.com/patterns/arrange-act-assert/)

## Potential Follow-Up
I tried to limit the scope of these changes to just the area of tests that have been flaking. However, if other tests flake as well, it may be due to test pollution, and we can follow a similar pattern of being explicit about the resources that are being created and deleted.

# Checklist:


test/v1helpers/test_upstream.go
```diff
@@ -112,7 +112,7 @@ func newTestUpstream(addr string, ports []uint32, responses <-chan *ReceivedRequ
 	}
 	u := &gloov1.Upstream{
 		Metadata: &core.Metadata{
-			Name:      fmt.Sprintf("local-%d", id),
+			Name:      fmt.Sprintf("local-test-upstream-%d", id),
```
sam-heilbron, line 115: This is a non-functional change. I saw in some failures references to 'local-x' and couldn't easily determine where those were created. When it was clean that each of those are references to a TestUpstream instance, I decided to rename it so that it's easier to associate.

sam-heilbron:
Build bot results:

1. [IN PROGRESS](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d6eb4130-f45a-475e-b18e-1e19621fc294)

sam-heilbron:
/skip-changelog this will fail until we release 1.12.0-beta1

test/consulvaulte2e/consul_vault_test.go
```diff
@@ -178,10 +178,8 @@ var _ = Describe("Consul + Vault Configuration Happy Path e2e", func() {
 			err = vaultInstance.Clean()
 			Expect(err).NotTo(HaveOccurred())
 		}
-		if envoyInstance != nil {
```
jenshu, line 181: is it safe to remove the nil checks in all these files, i.e. is envoyInstance always initialized?

jenshu: ooh i see, the function checks for nil

test/services/envoy.go
```diff
@@ -330,15 +330,22 @@ func (ef *EnvoyFactory) Clean() error {
 	return nil
 }
 
-func (ei *EnvoyInstance) EnvoyConfig() (*http.Response, error) {
+func (ei *EnvoyInstance) EnvoyConfigDump() (string, error) {
```
ben-taussig-solo, line 333: 🙌

test/e2e/gateway_test.go
```diff
@@ -596,67 +611,58 @@ var _ = Describe("Gateway", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+			})
 
-				TestUpstreamSslReachableTcp := func() {
-					cert := gloohelpers.Certificate()
-					v1helpers.TestUpstreamReachable(defaults.HttpsPort, tu, &cert)
+			JustAfterEach(func() {
+				for _, gw := range defaultGateways {
+					err := testClients.GatewayClient.Delete(gw.GetMetadata().GetNamespace(), gw.GetMetadata().GetName(), clients.DeleteOpts{})
+					Expect(err).NotTo(HaveOccurred(), "Should be able to write default gateways")
```
bewebi, line 652:
```suggestion
					Expect(err).NotTo(HaveOccurred(), "Should be able to delete default gateways")
```

bewebi: A couple more instances of this on 788, 791

test/e2e/gateway_test.go
```diff
@@ -207,6 +232,10 @@ var _ = Describe("Gateway", func() {
 				Expect(service.Ref.Namespace).To(Equal(svc.Namespace))
 				Expect(service.Ref.Name).To(Equal(svc.Name))
 				Expect(service.Port).To(BeEquivalentTo(svc.Spec.Ports[0].Port))
+
+				// clean up the virtual service that we created
+				err = testClients.VirtualServiceClient.Delete(vs.GetMetadata().GetNamespace(), vs.GetMetadata().GetName(), clients.DeleteOpts{})
```
ben-taussig-solo, line 237: I think we should consider writing/deleting the VS in a beforeEach/afterEach to better match the AAA pattern we're following

sam-heilbron: Yeah it's a good idea. the reason I didn't for this test, is because the test takes a set of resources, and then verifies that it can add/delete resources as part of the test. Unfortuatnely the Act and Assert portions require writing and deleting resources so I couldn't find a good way to pull them out

test/e2e/gateway_test.go
sam-heilbron: Good catch, I think I got these all in https://github.com/solo-io/gloo/pull/6211/commits/9d824c48bfe7b50cd55a405de2f858f2328a605a

test/e2e/gateway_test.go
```diff
@@ -1210,55 +1156,57 @@ var _ = Describe("Gateway", func() {
...
-												PrefixLen: &wrappers.UInt32Value{
-													Value: 32,
-												},
-											},
-										},
-									},
-									GatewayType: &gatewayv1.MatchedGateway_HttpGateway{
-										HttpGateway: &gatewayv1.HttpGateway{},
-									},
-								},
```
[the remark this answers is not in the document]
bewebi: You're right that the presence of an HTTP gateway shouldn't shouldn't be relevant to TCP gateways
The inclusion of an HTTP gateway in the input is meant to demonstrate that this is the case
If there were a future code change that somehow caused the presence of an HTTP gateway to affect the intended behavior for TCP we would only expect to catch it if we tested that case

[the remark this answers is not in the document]
sam-heilbron: Sounds good! I can add back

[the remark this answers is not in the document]
sam-heilbron: Thanks for the clarification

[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/pull/6211/commits/219e20395b86b5d8017c9cd015524156bff1085c

test/e2e/gateway_test.go
ben-taussig-solo: Ah, fair enough. Resolving this comment.

bewebi (commented):
LGTM

test/kube2e/util.go
```diff
@@ -117,6 +119,15 @@ func EventuallyReachesConsistentState(installNamespace string) {
 	}, "30s", eventuallyConsistentPollingInterval).Should(Equal(lastSnapOut))
 }
 
+// Copied from: https://github.com/solo-io/go-utils/blob/176c4c008b4d7cde836269c7a817f657b6981236/testutils/assertions.go#L20
+func ExpectEqualProtoMessages(g Gomega, a, b proto.Message, optionalDescription ...interface{}) {
```
jenshu, line 123: can't we just call the testutils func directly? i believe we're doing that elsewhere in the gloo tests already

jenshu: oh is this different?

jenshu (approved):

Issue #5917, opened 2022-02-11
The JWT option on VirtualService cannot anymore be disabled at the route level


bcollard:
### Gloo Edge Version

1.10.x (latest stable)

### Describe the bug

Given a JWT verification configured at the VirtualService level
and a route with JWT option disabled
then the request matching this route is still subject to JWT validation
Also, the JWT option at the VS level is using `claimsToHeader`, so might be related to the filter flow when clear route cache is used.

Works as expected until Gloo EE version `1.10.0-beta8`
It doesn't work anymore since Gloo EE version `1.10.0-rc1`

### Steps to reproduce the bug

see the full test in this gist: https://gist.github.com/bcollard/75be986596bc9c6c9b066e14ae31d5bd

### Expected Behavior

no JWT verification on the route where JWT is disabled

ben-taussig-solo:
Update: 
 - This issue appears to be the result of changes to the behavior of the `extensions.filters.http.jwt_authn.v3.JwtRequirement` filter in Envoy v1.20.0. Unfortunately, these changes are not, as far as we can tell, listed in the v1.20.0 changelog and we have yet to identify them.
   - Essentially, as of envoy v1.20.0/solo-projects v1.10.0-rc1, the envoy [`JwtRequirement`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto.html?highlight=allow_missing#extensions-filters-http-jwt-authn-v3-jwtrequirement) filter's [`allow_missing_or_failed`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto.html?highlight=allow_missing#envoy-v3-api-field-extensions-filters-http-jwt-authn-v3-jwtrequirement-allow-missing-or-failed) setting now must be enabled in order to allow JWT verification to be disabled on a specific route.
 -  For now, you may be able to work around the issue by setting `virtualHost.options.jwtStaged.beforeExtAuth.allowMissingOrFailedJwt` to true.

kevin-shelaga:
The proposed workaround would open all routes to allow a missing or failed jwt, which is not ideal for my use case.

ben-taussig-solo:
Some clarification -- the JWT filter can be disabled at route level when the route specifies a `routeAction`. Take, for example, the following virtual service configured with the other resources present in the gist linked in the original issue:
```
apiVersion: gateway.solo.io/v1
[...]

PR #1914, opened 2019-12-12
Direct response header manipulation

marcogschmidt:
Currently Gloo skips the processing of route plugins on DirectResponseAction route actions. Envoy supports header manipulation on direct responses. This PR exposes that feature through Gloo.

This includes the ability to copy request headers to the direct response via the `%REQ(header-name)%` format specifier.

PR #6220, opened 2022-04-01
Create DirectResponseRoutePlugin and RedirectActionRoutePlugin plugin types

ben-taussig-solo:
# Description

 - Prior to this PR, route-level plugin configuration for routes which specify `directResponseAction`s and `redirectAction`s was only supported in the control-plane for the header manipulation plugin.
 - This PR introduces `DirectResponseRoutePlugin` and `RedirectActionRoutePlugin` plugin types. Similarly to how a `RoutePlugin` can set route-level config for a route which defines a `routeAction`, these new plugin types can set route-level config for routes which define `directResponseAction`s and `redirectAction`s, respectively.

# Context

 - This functionality can be leveraged to resolve the issue described here: https://github.com/solo-io/gloo/issues/5917
   - In short, the issue is caused by a failure to process route-level config for routes which specify a `redirectAction`. By converting the enterprise JWT authn plugin to become a `redirectAction`, we will be able to resolve the issue.


projects/gloo/pkg/translator/route_config.go
```diff
@@ -265,15 +264,15 @@ func (h *httpRouteConfigurationTranslator) setAction(
 
 		// DirectResponseAction supports header manipulation, so we want to process the corresponding plugin.
 		// See here: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#route-directresponseaction
-		for _, plug := range h.pluginRegistry.GetRoutePlugins() {
-			if plug.Name() != headers.ExtensionName {
-				continue
-			}
+		for _, plug := range h.pluginRegistry.GetDirectResponseRoutePlugins() {
+			fmt.Printf("\nDirect Response Action: Processing Plugin with name: %s\n", plug.Name())
```
kdorosh, line 268: remove printlns

ben-taussig-solo: oof thanks

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -107,6 +109,14 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 	return nil
 }
 
+func (p *plugin) ProcessDirectResponseRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	return p.ProcessRoute(params, in, out)
+}
+
+func (p *plugin) ProcessRedirectActionRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	return p.ProcessRoute(params, in, out)
```
kdorosh, line 117: why not just make sure we call `ProcessRoute` on direct and redirect actions?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -265,15 +264,15 @@ func (h *httpRouteConfigurationTranslator) setAction(
...
 
-			if err := plug.ProcessRoute(params, in, out); err != nil {
```
kdorosh, line 273: we probably still want process route for non direct and non redirect routes

projects/gloo/pkg/plugins/headers/plugin.go
ben-taussig-solo: I considered that, but I ended up deciding against that for a couple of reasons. Currently, we support configuration for direct and redirect actions for the Header Manipulation filter only. We accomplish that by explicitly calling out that filter as one that shouldn't be skipped in the `projects/gloo/pkg/translator/route_config.go` `setAction` method (excerpt below):
```go
	case *v1.Route_DirectResponseAction:
                ...
		for _, plug := range h.pluginRegistry.GetRoutePlugins() {
			if plug.Name() != headers.ExtensionName {
				continue
			}
                        if err := plug.ProcessRoute(params, in, out); err != nil {
                        ...
                 ...
```

We could amend this to do something like the following, which would have the same effect as the propsed PR:
```go
                        ...
			if plug.Name() != headers.ExtensionName && plug.Name() != "jwt" {
				continue
			}
                        ...
```

I prefer the approach currently taken by the PR for two reasons:
 - Because the jwt plugin is defined in solo-projects, we cannot compare `plug.Name()` against `jwt.ExtensionName` like we do for the header manipulation plugin. For that reason, we have to hardcode `"jwt"` (unless there is another approach to this), which I do not like. 
   - Additionally, if we decide that we want more plugins to be configurable with direct and redirect actions, this conditional will continue to grow, potentially containing more hardcoded plugin names if we choose to opt more enterprise plugins into this functionality.
 - I think opting into route-level direct/redirect action configuration at the plugin level (as opposed to during translation) is easier to reason about for maintainers. As it currently stands, you need to be familiar with the inner workings of translation in order to know that route-level config is only applied to the header manipulation plugin. I think that if this information is supplied in the plugin definition, it will be easier to understand what we expect a given plugin to do. I.E., if you see `_ plugins.DirectResponseRoutePlugin = new(plugin)` at the top of some plugin's `plugin.go` file, you will be able to conclude that it is capable of applying config to a route with a direct response action.

projects/gloo/pkg/translator/route_config.go
ben-taussig-solo: We still do -- the highlighted block only applies when the incoming routeAction is a `Route_DirectResponseAction` (see line 257)
```go
	case *v1.Route_DirectResponseAction:
```

sam-heilbron (commented):
I like the idea, though I'd love to understand why we only call the headers plugin for direct response action, before me make this change.




I wonder if we could modify the RoutePlugin implementations to return nil if they dont operate on a direct response action or redirect action route? To me this will make it harder for devs to forget to implement this logic than having seprate interfaces to implement. Though I need to look more closely at the code to understand the implications / reasons why we may not have done that.

ben-taussig-solo:
> I'd love to understand why we only call the headers plugin for direct response action

[Here's](https://github.com/solo-io/gloo/pull/1914/files) the PR that introduced support for configuring the header manipulation plugin for `directResponseAction`s and `redirectAction`s. Since then, no other plugin has opted in to configuration for these action types.

ben-taussig-solo:
> I wonder if we could modify the RoutePlugin implementations to return nil if they don't operate on a direct response action or redirect action route?

We could do this, and it would have the same effect as the current set of changes. Personally, I prefer the current changeset -- the proposed change would expand the responsibilities of each RoutePlugin, when only two (header manipulation, jwt authn) need the updated behavior. As a consequence, each RoutePlugin definition becomes more verbose and (realistically this is the greater concern) more significant changes to structure of `httpRouteConfigurationTranslator.setAction` will be required.

projects/gloo/pkg/plugins/plugin_interface.go
```diff
@@ -75,6 +75,16 @@ type RoutePlugin interface {
 	ProcessRoute(params RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error
 }
 
+type DirectResponseRoutePlugin interface {
+	Plugin
+	ProcessDirectResponseRoute(params RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error
```
sam-heilbron, line 80: If we are moving in the direction of having these new interfaces/plugins, I think we should try to scope the interface as much as possible. Currently we accept a Route (gloo and envoy). I think we could lean on the RouteAction plugin which takes a RouteAction and modifies one. That way plugins only expose the configuration that they are able to modify.

ben-taussig-solo: I like that!

ben-taussig-solo: One complication -- RouteActions do not have access to RouteOptions, by which a user can specify route-level config

Issue #5448, opened 2021-10-08
Expose regex_rewrite in RedirectAction


djannot:
**Is your feature request related to a problem? Please describe.**
Currently, there's no way in Gloo Edge to perform similar than what the `regex_rewrite` option provide.

**Describe the solution you'd like**
Expose the `regex_rewrite` (https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-redirectaction) option in Gloo Edge `RedirectAction` (https://docs.solo.io/gloo-edge/master/reference/api/envoy/api/v2/route/route.proto.sk/#redirectaction)

nfuden:
Whats the difference of this and  functionality seen here https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/regex_rewrite/

It may be in a better location for setting but it sure seems like this covers it?

mkazin:
/kick
Timeout in test:
`Step #10 - "test": Gateway in memory hybrid gateway http traffic ssl [It] should work with ssl if ssl config is present in matcher
Step #10 - "test": /workspace/gloo/test/e2e/gateway_test.go:1059`

PR #6226, opened 2022-04-04
utils/regexutils: Re-allow nil engine types in regex.

nfuden:
We have documents showing how to setup regex rewrites without requiring engine.
We stopped allowing this behavior but will once more respect this.

It makes sense to not require the engine as envoy has moved away from interacting with the engine (only field on it is now deprecated). It does raise questions as to whether there needs to be more regex validation on the control plane prior to translation but this minimally fixes the issue seen in 1.11.0-rc5/rc6 of gloo
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5448


mkazin:
LGTM!

sam-heilbron (approved):

PR #6231, opened 2022-04-04
Add README.md to .github/workflows 

ben-taussig-solo:
# Description

 - Adds a `README.md` to `.github/workflows` with guides for manually administering the push-solo-apis workflow.


.github/workflows/README.md
```diff
@@ -0,0 +1,10 @@
+# GH Workflows
+
+## Push API Changes to Solo-APIs
+ - This action is used to open a PR in Solo-APIs which corresponds to a set of changes in Gloo OSS
+ - The action is run when a Gloo OSS release is published
+ - The action can be run manually from the "Actions" tab in Github while viewing the Gloo OSS repo
```
inFocus7, line 6: nit: Should the wording for "Push API Changes to solo-apis" be 'workflow' instead of 'action'?

Seen this article defining actions, workflows, etc. and I think this is technically a workflow we're manually running (since it's defined in a `*.yaml` config and it runs jobs/actions)

https://dev.to/github/whats-the-difference-between-a-github-action-and-a-workflow-2gba

ben-taussig-solo:
> nit: Should the wording for these lines be 'workflow' instead of 'action'?
Just addressed this here: [8fb4183](https://github.com/solo-io/gloo/pull/6231/commits/8fb41838715f08ad4d0a9550b49f9ae3fb5c9192)

PR #6233, opened 2022-04-04
Process all route plugin config in direct and redirect actions

ben-taussig-solo:
# Description
 - Prior to this PR, route-level plugin configuration for routes which specify directResponseActions and redirectActions was only supported in the control-plane for the header manipulation plugin.
 - Amends translation for routes which specify direct response and redirect actions, so that provided config is always processed for all route plugins

# Context
 - This functionality will resolve the issue described here: https://github.com/solo-io/gloo/issues/5917
   - In short, the issue is caused by a failure to process route-level config for routes which specify a `redirectAction`. By converting the enterprise JWT authn plugin config for `redirectAction`s, we will resolve the issue.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5917


projects/gloo/pkg/translator/route_config.go
```diff
@@ -327,6 +322,19 @@ func (h *httpRouteConfigurationTranslator) setAction(
 				}
 			}
 		}
+
+		for _, plug := range h.pluginRegistry.GetRoutePlugins() {
```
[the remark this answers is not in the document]
ben-taussig-solo: Agreed, I was just looking at that -- do you think it makes sense to just call runRoutePlugins for all action types?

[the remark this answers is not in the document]
ben-taussig-solo: Implemented my suggestion here: 16b7a35f55ae98b56e8ae1761c378c7383264be9 -- Let me know what you think

projects/gloo/pkg/plugins/basicroute/plugin.go
```diff
@@ -48,9 +48,15 @@ func (p *plugin) ProcessVirtualHost(
 }
 
 func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	// This plugin is only available for routeActions. return early if a different action is specified.
```
sam-heilbron, line 51: Why did this guard not exist before? Have we changed where we check this?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -328,6 +306,7 @@ func (h *httpRouteConfigurationTranslator) setAction(
 			}
 		}
 	}
+	h.runRoutePlugins(params, routeReport, in, out)
```
sam-heilbron, line 309: RunRoutePlugins runs both RoutePlugins and RouteActionPlugins..is this intentional? Perhaps we could rename this method for clarity.

projects/gloo/pkg/plugins/basicroute/plugin.go
ben-taussig-solo: This guard is not strictly necessary -- each method in the basicroute plugin used to apply specific settings (i.e., `applyPrefixRewrite`, `applyRegexRewrite`, `applyTimeout`) does this check:
```
	routeAction, ok := out.GetAction().(*envoy_config_route_v3.Route_Route)
	if !ok {
		return errors.Errorf("prefix rewrite is only available for Route Actions")
	}
```

for now, I am going to remove the check highlighted in your comment for the sake of brevity

ben-taussig-solo: removed this check here: 2e32ff9441570f59f43d35629d66dbe1adc92149

projects/gloo/pkg/translator/route_config.go
ben-taussig-solo: Separated `runRoutePlugins` into two methods here: 5d44291116d1e3db4f0f8d754271d78ebd8e0f0e

changelog/v1.12.0-beta2/process-route-plugins-all-actions.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NON_USER_FACING
```
sam-heilbron, line 2: I think this should be a FIX. We are fixing a bug (jwt disabling) and I think it's important to call out the difference in behavior (we used to not process config and now we are). Though it's not breaking, the changelog will be the easiest way to call out that potential change.

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -791,26 +802,137 @@ var _ = Describe("Translator", func() {
...
 					},
 				},
+				Options: options,
 			}
 			routes = []*v1.Route{redirectRoute, directResponseRoute}
 		})
 
-		It("reports no errors with a redirect route or direct response route", func() {
+		JustBeforeEach(func() {
+			// we need to set this in order for tracing settings to be processedRegexMatcher_GoogleRe2
```
sam-heilbron, line 854: I don't quite follow why this is included, can you help me understand?

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -791,26 +802,137 @@ var _ = Describe("Translator", func() {
...
+				Pattern: &v32.RegexMatcher{
+					Regex: "Test",
+				},
+			}
+
+			for _, route := range proxy.GetListeners()[0].GetHttpListener().GetVirtualHosts()[0].GetRoutes() {
+				route.Options = invalidOptions
+			}
+
+			// re-run translation and confirm that the new settings do not affect envoy config
```
sam-heilbron, line 932: Perhaps I need to look more closely at the code, but why would these new settings not affect envoy config? Is it because these settings are configured by plugins that are not RoutePlugins and thus are never processed?

sam-heilbron (commented):
Looking good! Small questions about tests but LGTM

projects/gloo/pkg/translator/translator_test.go
ben-taussig-solo: This block was unnecessary -- I thought that the tracing filter would not process incoming route config unless the corresponding HCM setting was enabled, but some testing seems to suggest that that is not the case.

ben-taussig-solo: Removed here: dd50d19e6e84f83bdea13fb6835c6818809c750f

changelog/v1.12.0-beta2/process-route-plugins-all-actions.yaml
ben-taussig-solo: I think that's fair, especially given how many RoutePlugins implicitly assume that it is possible that they may be processing a route which specifies a direct response/redirect action

PR #6235, opened 2022-04-04
Adding else clause in case discovery is disabled, because it fails if…

vladimir-cmd:
… not empty list.

# Description

- added else clause for Discovery feature if disabled to add an empty list
- this bug fixes the issue with an empty value under subjects, in case .Values.discovery.enabled = false

# Context

Users needed this feature when doing an upgrade of gloo to 1.10.18


kdorosh (commented):
can we add a changelog per https://github.com/solo-io/go-utils/tree/master/changelogutils#changelog

mkazin:
@vladimir-cmd Thank you for raising this issue and for your code contribution.
I have included this commit in #6249 and closing this PR.
Please let me know if you have any other concerns.

vladimir-cmd:
Thanks @mkazin for fast response. 
Vlad

Issue #6237, opened 2022-04-04
gateway_test: invalid memory address or nil pointer dereference


jenshu:
### Gloo Edge Version

1.11.x (beta)

### Describe the bug

```
Step #10 - "test": • [SLOW TEST:8.036 seconds]
Step #10 - "test": Gateway in memory http gateway won't allow a bad authconfig in a virtualservice to block updates to a gateway 
Step #10 - "test": /workspace/gloo/test/e2e/gateway_test.go:212
Step #10 - "test": ------------------------------
Step #10 - "test": • [SLOW TEST:13.049 seconds]
Step #10 - "test": Gateway in memory http gateway traffic works when rapid virtual service creation and deletion causes no race conditions 
Step #10 - "test": /workspace/gloo/test/e2e/gateway_test.go:367
Step #10 - "test": ------------------------------
Step #10 - "test": • [SLOW TEST:5.163 seconds]
Step #10 - "test": Gateway in memory http gateway traffic should work with no ssl and clean up the envoy config when the virtual service is deleted 
Step #10 - "test": /workspace/gloo/test/e2e/gateway_test.go:413
Step #10 - "test": ------------------------------
Step #10 - "test": ••••panic: runtime error: invalid memory address or nil pointer dereference
Step #10 - "test": [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x5124816]
Step #10 - "test": 
Step #10 - "test": goroutine 22392 [running]:
Step #10 - "test": github.com/solo-io/gloo/projects/gateway/pkg/validation.(*connectionRefreshingValidationClient).Validate.func1(0x0, 0x0, 0xc0095af170, 0x0)
Step #10 - "test": 	/workspace/gloo/projects/gateway/pkg/validation/robust_client.go:66 +0x56
Step #10 - "test": github.com/solo-io/gloo/projects/gateway/pkg/validation.(*connectionRefreshingValidationClient).retryWithNewClient.func1(0x0, 0x0)
Step #10 - "test": 	/workspace/gloo/projects/gateway/pkg/validation/robust_client.go:88 +0x155
Step #10 - "test": github.com/avast/retry-go.Do(0xc0095af398, 0xc0095af388, 0x2, 0x2, 0x1, 0xc007b8fb30)
Step #10 - "test": 	/go/pkg/mod/github.com/avast/retry-go@v2.4.3+incompatible/retry.go:103 +0x247
[...]

PR #6260, opened 2022-04-07
Resolve panic in gateway validation client

sam-heilbron:
# Description

Ensure that the gateway validation client is never set to nil

# Context

We encountered this bug in tests that would occasionally panic, however the bug itself could impact customers

## Verification

I wrote a test first, and add that as the first commit. If you checkout just that test (no other changes) it will panic in the way that we noticed in some e2e tests. If you then checkout my fix and re-run the same test, it will pass.

## General Thoughts

I believe this issue arises because we have a local variable for clientReinstantiateError and a class variable for constructorClient. This means that we can ensure that on a clientReinstantiateError we do not retry, but we do set the client to nil which could be picked up on the subsequent request.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6237

### DIFF
--- a/changelog/v1.12.0-beta2/robust-client-panic.yaml
+++ b/changelog/v1.12.0-beta2/robust-client-panic.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6237
+    resolvesIssue: true
+    description: Ensure that the validation client is nevel nil, even after the connection has been refreshed.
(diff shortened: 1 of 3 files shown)


changelog/v1.12.0-beta2/robust-client-panic.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6237
+    resolvesIssue: true
+    description: Ensure that the validation client is nevel nil, even after the connection has been refreshed.
```
kdorosh, line 5:
```suggestion
    description: Ensure that the validation client is never nil, even after the connection has been refreshed. 
```

kdorosh (commented):
seems reasonable to me

gunnar-solo (approved):
Nice fix!  The null guard makes sense to me, per the [original issue](https://github.com/solo-io/gloo/issues/6237) hitting a `nil` on `robust_client.go:66`.

The patched line is the only place (I can see) where `c.validationClient` is set

PR #4563, opened 2021-04-07
Fix consul DNS resolution error handling

jahvon:
# Description

Use an address' previous DNS resolutions if consul DNS resolution returns an error.

# Context

Currently the endpoint will be removed from EDS if DNS resolution fails. This resolves that issue.

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/2576

PR #6271, opened 2022-04-07
Resolve consul plugin data race

sam-heilbron:
# Description

Resolve a data race when resolving IP addresses

# Context

## Why fallback on errors?

In the consul plugin, we attempt to resolve IP addresses and then fallback to some previous state. This was introduced in https://github.com/solo-io/gloo/pull/4563 and that PR/issue have the relevant context for why we want to support that.

## What is the bug?

The data race occurs because we are accessing a map without synchronization

## What is the solution?

We have a DnsResolver in the Consul plugin. It is responsible for taking an address and resolving it to return a list of IP Addresses or an error. Our decision to cache previous results can be wrapped into this resolver and let the plugin be unaware of its existence. Therefore, I introduced a new DnsResolver implementation, one that wraps the previous one but incorporates memoization. We inject the plugin with this new resolver so the plugin doesn't have to be aware of the change.

## How did I test it?

We have an [existing eds test](https://github.com/solo-io/gloo/blob/c687552a730ad13bf7e47795628f2ed0b4d3526a/projects/gloo/pkg/plugins/consul/eds_test.go#L705) which I updated to use our new dns resolver and confirmed the test still works as expected.

# Checklist:


projects/gloo/pkg/plugins/consul/eds.go
```diff
@@ -285,13 +283,7 @@ func buildEndpoints(
 
 	ipAddresses, err := getIpAddresses(ctx, address, resolver)
 	if err != nil {
-		addresses, resolvedPreviously := previousResolutions[address]
-		if !resolvedPreviously {
-			return nil, err
-		}
-		ipAddresses = addresses
-	} else {
-		previousResolutions[address] = ipAddresses
```
sam-heilbron, line 294: The data race occurred because we were writing to the map from multiple places. One solution would have been to pass around the map and its lock (or some synchronized map). I opted instead to pull this caching behavior out of the plugin altogether.

changelog/v1.12.0-beta2/consul-plugin-data-race.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/4782
+    resolvesIssue: false
```
sam-heilbron, line 4: leaving open for backports

projects/gloo/pkg/plugins/consul/dns_resolver.go
```diff
@@ -38,3 +55,31 @@ func (c *ConsulDnsResolver) Resolve(ctx context.Context, address string) ([]net.
...
+	mutex               sync.RWMutex
+	previousResolutions map[string][]net.IPAddr
+}
+
+func (d *DnsResolverWithFallback) Resolve(ctx context.Context, address string) ([]net.IPAddr, error) {
+	ipAddrs, err := d.resolver.Resolve(ctx, address)
+
+	// If we successfully resolved the addresses, update our last known state and return
+	if err == nil {
+		d.mutex.Lock()
```
nfuden, line 71: nit: we do this in two places and we use a defer. Might as well pull this above the err check.

projects/gloo/pkg/plugins/consul/dns_resolver.go
```diff
@@ -38,3 +55,31 @@ func (c *ConsulDnsResolver) Resolve(ctx context.Context, address string) ([]net.
 	}
 	return ipAddrs, nil
 }
+
+type DnsResolverWithFallback struct {
+	resolver DnsResolver
+
+	mutex               sync.RWMutex
```
nfuden, line 62: Question/nit: Do you think we could embed the mutex? 
If so I personally prefer the signatures of d.Lock() over d.mutex.Lock()

projects/gloo/pkg/plugins/consul/dns_resolver.go
sam-heilbron: Yeah I like it. It's actually how I initially wrote it, and figured it might be a slight optimization that reads aren't blocked. However, the potential for that use case is minimal compared to the readability, so I'll move the locking out of the conditional

projects/gloo/pkg/plugins/consul/dns_resolver.go
sam-heilbron: Love it! I hadn't considered that

kdorosh (commented):
lgtm!

projects/gloo/pkg/plugins/consul/dns_resolver.go
sam-heilbron: Should be handled by https://github.com/solo-io/gloo/pull/6271/commits/465e5d97ecf7f8c1357216d667fb70b5768c8723

projects/gloo/pkg/plugins/consul/dns_resolver.go
sam-heilbron: Should be handled by https://github.com/solo-io/gloo/pull/6271/commits/465e5d97ecf7f8c1357216d667fb70b5768c8723

nfuden (approved):
Nice fix! This one's been a pain for a while.

PR #6282, opened 2022-04-08
Resolve data race in setup syncer test

sam-heilbron:
# Description

Resolve a data race that occurs in our setup syncer test

# Context

I think this data race is due to our tests and not due to our implementation. Per https://github.com/kubernetes/kubernetes/pull/89019#issuecomment-600278461, we do setup schemes at construction time, as we should. Moreover, this flake has only occurred in this specific test. If this were a bug in our Gloo code, I would have expected to see it fail in other tests.

# Checklist:

### DIFF
--- a/test/kube2e/gloo/setup_syncer_test.go
+++ b/test/kube2e/gloo/setup_syncer_test.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/solo-io/gloo/pkg/utils/setuputils"
+
 	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/registry"
 
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/enterprise/options/graphql/v1beta1"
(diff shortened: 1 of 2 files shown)


test/kube2e/gloo/setup_syncer_test.go
```diff
@@ -55,6 +58,25 @@ var _ = Describe("SetupSyncer", func() {
 		ctx = settingsutil.WithSettings(ctx, settings)
 	}
 
+	// SetupFunc is used to configure Gloo with appropriate configuration
+	// It is assumed to run once at construction time, and therefore it executes directives that
+	// are also assumed to only run at construction time.
+	// One of those, is the construction of schemes: https://github.com/kubernetes/kubernetes/pull/89019#issuecomment-600278461
```
nfuden, line 64: Ooh I remember seeing that and thinking it was just something we were doing in gloo. 
Its great that you were able to identify where we actually call the scheme setup!

nfuden (approved):

PR #6289, opened 2022-04-09
Update Sanitizers to never error

kdorosh:
# Description

testing ci for now

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #6115, opened 2022-03-18
Write last xds snapshot to persisted storage


kdorosh:
### Is your feature request related to a problem? Please describe.

we need to ensure [Ensure Gloo Edge is reliable across all pod restarts and invalid configuration](https://github.com/solo-io/gloo/issues/5721#)

the shortest path to ensure reliable xds configuration is always being served is to write last acked xds snapshot to persistent storage, and load that if gloo translation is unable to complete (related https://github.com/solo-io/gloo/issues/6114)

### Describe the solution you'd like

alternative solution: write last xds cache to persistent storage. Size M, Risk M

### Additional Context

downside: breaks multitenancy, which may or may not be a product requirement in all gloo settings deployments

chrisgaun:
@kdorosh we need to consider the implications of adding private keys - certificates when not using SDS - to a PV.

kdorosh:
related blocker i ran into while doing the work https://github.com/solo-io/solo-kit/issues/461

kdorosh:
related: https://github.com/solo-io/gloo/issues/5022

steps to reproduce:
- `kind create cluster --name kind --image kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6`
- `glooctl install gateway enterprise --version 1.11.0-beta8 --license-key $LICENSE_KEY`
- `kubectl scale -n gloo-system deploy/discovery --replicas 0`
- `kubectl apply -f https://raw.githubusercontent.com/solo-io/gloo/v1.2.9/example/petstore/petstore.yaml`
- apply the following yaml
```yaml
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: default
  namespace: gloo-system
spec:
  virtualHost:
    domains:
    - '*'
    routes:
    - matchers:
      - exact: /all-pets
      options:
        prefixRewrite: /api/pets
      routeAction:
        single:
          upstream:
            name: default-petstore-8080
            namespace: gloo-system
    - matchers:
      - exact: /all-pets2
      options:
        prefixRewrite: /api/pets
      routeAction:
        single:
          upstream:
            name: default-petstore2-8080
            namespace: gloo-system
---
apiVersion: v1
kind: Service
metadata:
  name: petstore
  namespace: default
  labels:
    service: petstore
spec:
  ports:
  - port: 8080
    protocol: TCP
  selector:
    app: petstore
---
apiVersion: v1
[...]

PR #6303, opened 2022-04-12
Add e2e test for sanitizers never error

kdorosh:
# Description

Add k8s e2e test for https://github.com/solo-io/gloo/pull/6289#pullrequestreview-938372272 , basically equivalent to these steps: https://github.com/solo-io/gloo/issues/6115#issuecomment-1091856972

also ran the new test locally against gloo edge 1.12.0-beta2 and confirmed it fails before the changes aforementioned, and passes on version 1.12.0-beta3

# Checklist:

### DIFF
--- a/test/kube2e/gateway/robustness_test.go
+++ b/test/kube2e/gateway/robustness_test.go
@@ -410,56 +410,49 @@ var _ = Describe("Robustness tests", func() {
 			}
 
 			It("works, even with deleted services", func() {
-				// Ensure we can route to the service
+				// Ensure we can route to the first service
 				testHelper.CurlEventuallyShouldRespond(firstRouteCurlOpts(), expectedResponse(appName), 1, 90*time.Second, 1*time.Second)
-
-				// Ensure we can route to the service
+				// Ensure we can route to the second service
 				testHelper.CurlEventuallyShouldRespond(secondRouteCurlOpts(), expectedResponse(appName2), 1, 90*time.Second, 1*time.Second)
 
+				// Delete the k8s service behind the second echo app
 				err = kubeClient.CoreV1().Services(appService2.Namespace).Delete(ctx, appService2.Name, metav1.DeleteOptions{})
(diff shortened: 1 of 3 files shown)


test/kube2e/gateway/robustness_test.go
```diff
@@ -292,6 +329,134 @@ var _ = Describe("Robustness tests", func() {
...
+
+				// can no longer route to appName2 since its k8s service has been removed
+				Eventually(func() (string, error) {
+					return testHelper.Curl(secondRouteCurlOpts())
+				}, "30s", "1s").Should(BeEmpty())
+				Consistently(func() (string, error) {
+					return testHelper.Curl(secondRouteCurlOpts())
+				}, "5s", "1s").Should(BeEmpty())
+
+				// roll pods to ensure we are resilient to pod restarts
```
bewebi, line 436: What makes pod restarts an interesting case to test?
If it's conceivable that the "works with deleted services" behavior will be good but "works with deleted services after pod restart(s)" won't, we should consider having a separate `It()` for the latter

kdorosh: if we have invalid config we may not update our xds cache that gloo is serving (this was the old bug)

without rolling the pods envoy will just keep the old config; but we need to validate that even after pods roll that it is actively getting good config still

the test before this one checks for endpoint updates without a pod roll which seems to cover the other desired case

bewebi: Discussed offline, it's worth the quick add of a test that does everything up to this point as a simpler case that does not test resilience to pod restarts since that behavior could be correct even if were not resilient to pod restarts
This was effectively the case prior to the change we're testing

kdorosh: addressed in https://github.com/solo-io/gloo/pull/6303/commits/6ea3b4744c1cc90a0548d702854bbce5fd45e04d

jenshu (approved):

PR #6304, opened 2022-04-12
Run gzip compression after transformation

jenshu:
# Description

Currently, the gzip compression filter runs before the transformation filter on the response path (gzip does not run on the request path). This can result in corrupt data when the transformation filter modifies the gzip blob. To fix this, we now run gzip after transformation on the response path. This means that we need to configure gzip to run *before* transformation on the request path (even though gzip isn't actually used on the request path). Since the earliest stage that the transformation filter can run is the [stage after the FaultStage](https://github.com/solo-io/gloo/blob/081423f6837f7c6ad4be4ac8f4eaf5e98b074beb/projects/gloo/pkg/plugins/transformation/plugin.go#L44), gzip needs to run during the FaultStage.

# Checklist:

### DIFF
--- a/projects/gloo/pkg/plugins/gzip/plugin.go
+++ b/projects/gloo/pkg/plugins/gzip/plugin.go
@@ -26,7 +26,7 @@ const (
 )
 
 // filter should be called after routing decision has been made
-var pluginStage = plugins.DuringStage(plugins.RouteStage)
+var pluginStage = plugins.DuringStage(plugins.FaultStage)
 
 type plugin struct{}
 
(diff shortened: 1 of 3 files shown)


sam-heilbron (commented):
LGTM!

projects/gloo/pkg/plugins/gzip/plugin.go
```diff
@@ -26,7 +26,7 @@ const (
 )
 
 // filter should be called after routing decision has been made
```
sam-heilbron, line 28: let's update this comment as it is no longer accurate

kdorosh (approved):

nrjpoddar:
Even though the behavior might have been incorrect, does this change count as backward incompatible change? Should we have given users an option to preserve old behavior?

Issue #6272, opened 2022-04-07
Add option on the helm chart to use Grafana + Prometheus and Datadog


DiegoLelis:
### Version

1.10.x (latest stable)

### Is your feature request related to a problem? Please describe.

Currently it is quite challenging to manage an installation using Grafana + Prometheus and Datadog at the same time because it needs a manual step as outlined here: https://docs.solo.io/gloo-edge/master/guides/integrations/datadog/#configure-the-gloo-edge-deployment 

Adding an option on the helm chart that it would just enable/disable the Datadog integration, while keeping the Prometheus integration would make much easier to manage that.

### Describe the solution you'd like

Add an option on the chart that it will just enable the datadog endpoint for metrics without disabling the prometheus one, something like: `.Values.global.glooStats.enableAddDatadogRoute`

PR #6273, opened 2022-04-07
Adding enableAddDatadogRoute parameter

DiegoLelis:
# Description

Adds a new parameter on the helm chart that it will allow to enable/disable the datadog metrics endpoint, without disabling the Prometheus endpoint, this way will be possible to manage exposing both endpoints using the helm chart.
#6273 

# Context

Users needed this feature to be able to use Datadog and Grafana + Prometheus at the same time without any manual step.

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6272

PR #6306, opened 2022-04-12
Additional helm configurable paths to envoy stats

nfuden:
Potential more generalized approach to solve  https://github.com/solo-io/gloo/issues/6272

Based on: https://github.com/solo-io/gloo/pull/6273

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6272


install/helm/gloo/templates/30-knative-internal-proxy-configmap.yaml
```diff
@@ -103,6 +103,16 @@ data:
...
+                            {{ if .Values.global.glooStats.enableAdditionalStatsRoute}}
+                            - match:
+                                prefix: "/stats"
+                                headers:
+                                - name: ":method"
+                                  exact_match: GET
+                              route:
+                                prefix_rewrite: {{ .Values.global.glooStats.additionalStatsPrefixRewrite }}
+                                cluster: admin_port_cluster
+                            {{ end }}
```
kdorosh, line 115: this route ends up being very similar to the `"/metrics"` route; do we really need both? can we reuse the other one and make the prefix configurable?

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -58,12 +58,18 @@ spec:
...
 {{ $annotationExist = true}}
       annotations:
       {{- range $key, $value := $spec.podTemplate.extraAnnotations }}
         {{ $key }}: {{ $value | quote }}
       {{- end }}
+      {{- if .Values.global.glooStats.setDatadogAnnotations }}
+        ad.datadoghq.com/gateway-proxy.check_names: '["envoy"]'
+        ad.datadoghq.com/gateway-proxy.init_configs: '[{}]'
+        ad.datadoghq.com/gateway-proxy.instances: '[{"stats_url": "http://%%host%%:8081/stats"}]'
+        ad.datadoghq.com/gateway-proxy.logs: '[{"source": "envoy", "service": "gloo","log_processing_rules":[{"type": "multi_line", "name": "log_start_with_date","pattern" : "^\\[[0-9]{4}-[0-9]{2}-[0-9]{2}"}]}]'
```
kdorosh, line 71: are these values standard for all datadog installs?

nfuden: Per https://docs.solo.io/gloo-edge/latest/guides/integrations/datadog/ these are standard although if one wanted to these can still be set via the extra annotations.

install/helm/gloo/templates/30-knative-internal-proxy-configmap.yaml
nfuden: our current guidance is to add another route to the envoy stats if you want both but thats not configurable at helm level.
Hypothetically we can use the new prefix rewrite in case there is a certain use case where an additional stats route is desired that starts with the format defined.

install/helm/gloo/templates/15-clusteringress-proxy-configmap.yaml
```diff
@@ -103,6 +103,16 @@ data:
...
                                 prefix_rewrite: {{ .Values.global.glooStats.routePrefixRewrite }}
                                 cluster: admin_port_cluster
+                            {{- if .Values.global.glooStats.enableAdditionalStatsRoute}}
+                            - match:
+                                prefix: "/stats"
+                                headers:
+                                - name: ":method"
+                                  exact_match: GET
+                              route:
+                                prefix_rewrite: {{ .Values.global.glooStats.additionalStatsPrefixRewrite }}
```
kdorosh, line 113: nit, let's remove the word additional from all these names to make them shorter

kdorosh (commented):
lgtm beyond a naming nit

install/helm/gloo/templates/15-clusteringress-proxy-configmap.yaml
nfuden: It will make them a bit more readable for sure but my concern is that we have an existing corpus of documentation where we tell users to use the metrics route for all stats related scraping (namely via its prefixrewrite).

Do we think this could cause undue confusion or do we think that the description of the helm variables will be sufficient to avoid issue?

ben-taussig-solo (commented):
@nfuden Looks like you may need to re-run codegen, but otherwise this looks good to me. DM me when CI passes and I will approve this :)

ben-taussig-solo (approved):
LGTM

install/helm/gloo/generate/values.go
```diff
@@ -572,6 +572,9 @@ type K8s struct {
 type Stats struct {
 	Enabled               *bool   `json:"enabled,omitempty" desc:"Controls whether or not envoy stats are enabled"`
 	RoutePrefixRewrite    *string `json:"routePrefixRewrite,omitempty" desc:"The envoy stats endpoint to which the metrics are written"`
+	SetDatadogAnnotations *bool   `json:"setDatadogAnnotations,omitempty" desc:"Sets the default datadog annotations"`
+	EnableStatsRoute      *bool   `json:"enableStatsRoute,omitempty" desc:"Enables an additional route to the stats cluster defaulting to /stats"`
```
Rachael-Graham, line 576:
```suggestion
	SetDatadogAnnotations *bool   `json:"setDatadogAnnotations,omitempty" desc:"Sets the default Datadog annotations"`
	EnableStatsRoute      *bool   `json:"enableStatsRoute,omitempty" desc:"Enables an additional route to the stats cluster, which defaults to /stats"`
```

Rachael-Graham (commented):
doc

Issue #6384, opened 2022-04-28
Validation webhook cert is not regenerated on upgrade


jenshu:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

The normal order of events during a helm install is that we create the `ValidatingWebhookConfiguration` with an [empty caBundle](https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml#L25) field, then the [gateway-certgen job](https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/install/helm/gloo/templates/6.5-gateway-certgen-job.yaml) runs, which generates a cert, and patches the ValidatingWebhookConfiguration caBundle. Now that we reinstall the validation webhook [on upgrade](https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml#L12), we also need to call the gateway certgen job on upgrade, otherwise the webhook will be left with an empty caBundle. 

Additionally, there's a bug in the cert generation code where if it determines the existing cert is still valid, it erroneously [updates the webhook with the newly generated cert](https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/jobs/pkg/run/run.go#L101) instead of using the old existing one.

### Steps to reproduce the bug

helm install gloo, then helm upgrade gloo

After the upgrade, the validation webhook will no longer work:
- If failurePolicy=Fail then the helm upgrade will error.
- If failurePolicy=Ignore, then calls to the validation webhook will silently fail.

### Expected Behavior

Helm upgrade should succeed and validation webhook should still work

PR #6349, opened 2022-04-21
Fix Gloo Edge install when failurePolicy is Fail

jenshu:
# Description

The gloo helm charts include gloo custom resources (specifically Gateways) that pass through the validation webhook when they are applied. When webhook failurePolicy is set to Fail, then any issues reaching the validation service would result in an error. This means that it's possible for the install to fail due to attempting to apply custom resources before the validation service is running. 

To allow installs/upgrades to succeed when failurePolicy=Fail, we need to wait until the validation service is ready before applying custom Gateway resources. This is achieved by creating a job on install/upgrade that waits for the gloo deployment to be rolled out, and then applying the Gateway resources in a post-install/post-upgrade hook.

After this change, users will no longer need to do the workaround of [disabling default gateways](https://github.com/solo-io/gloo/pull/6017) when using failurePolicy=Fail.

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6403

PR #6385, opened 2022-04-28
Fix certgen upgrade

jenshu:
# Description

- Make sure the gateway certgen job runs on upgrade, since we now reinstall the validation webhook on upgrade and it needs the caBundle field to be populated by the certgen job.
- Fix a bug in the certgen job where if we don't create a new Secret (because the existing one is still valid), we were still setting a newly generated caBundle on the validation webhook (which was not attached to any persisted secret)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6384

### DIFF
--- a/changelog/v1.12.0-beta8/fix-certgen-upgrade.yaml
+++ b/changelog/v1.12.0-beta8/fix-certgen-upgrade.yaml
@@ -1,0 +1,4 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6384
+    description: Run gateway certgen job on upgrade and make sure the validation webhook gets an updated caBundle on upgrade.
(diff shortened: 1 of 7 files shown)


jenshu:
opted not to add e2e tests in this PR because the problem manifests itself when failurePolicy=Fail and i'll be adding e2e tests for that in [this PR](https://github.com/solo-io/gloo/pull/6349)

jenshu:
> Approving as draft pr wont automerge: Question do we think we should make a helm test for the upgrade case that is fixed in this? Looks good functionally!

Yep, was planning to add an e2e helm test as alluded to above. We have existing [helm e2e tests](https://github.com/solo-io/gloo/blob/master/test/kube2e/helm/helm_test.go) that do upgrades, but currently they don't fail because the failurePolicy is Ignore and there isn't an easy way to tell if things are broken with the validation webhook (since failurePolicy=Ignore means that any errors calling the webhook would be ignored). In the [next PR](https://github.com/solo-io/gloo/pull/6349) I'll add tests that make sure the webhook doesn't fail on upgrades when failurePolicy=Fail.

bewebi (approved):
LGTM

Issue #6356, opened 2022-04-21
ci test flake: Staged Transformation with auth [It] should transform response code details


jenshu:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

```
Step #10 - "test": [Fail] Staged Transformation with auth [It] should transform response code details 
Step #10 - "test": /workspace/gloo/test/e2e/staged_transformation_test.go:345
```

example logs:
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=15c11eec-37ec-4ebf-b00a-47ccf7fa17b7
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=addc9a7c-01cb-4df9-a0b6-dfa88b119503

### Steps to reproduce the bug

run gloo ci

### Expected Behavior

test should pass

sam-heilbron:
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=11731a4e-f341-4226-82f6-c2ce3eac25fb

sam-heilbron:
The error I see earlier in the logs:

````
Step #10 - "test": Proxies:[listeners:{name:"listener"  bind_address:"::"  bind_port:11102  http_listener:{virtual_hosts:{name:"virt1"  domains:"*"  routes:{route_action:{single:{upstream:{name:"local-test-upstream-14"  namespace:"default"}}}}  options:{extauth:{custom_auth:{}}  staged_transformations:{early:{response_transforms:{response_code_details:"ext_authz_error"  response_transformation:{transformation_template:{body:{text:"early-transformed"}  parse_body_behavior:DontParse}}}}}}}}}  namespaced_statuses:{statuses:{key:"gloo-system"  value:{state:Rejected  reason:"1 error occurred:\n\t* HttpListener Error: ProcessingError. Reason: extauth server upstream not found name:\"extauth-server\"  namespace:\"default\"\n\n\nwarning: \n  Route Warning: InvalidDestinationWarning. Reason: *v1.Upstream { default.local-test-upstream-14 } not found"  reported_by:"gloo"}}}  
```

To me this means we're not applying resources in the proper order or waiting long enough for resources to be available.

jenshu:
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=b6409ec5-e5b9-4d90-89df-001d3698b12b

elcasteel:
[...]

PR #6393, opened 2022-04-29
Test flake fixes

elcasteel:
# Description
The test sets up an upstream and some additional configuration and expects to get a 403 when all the config is propagated. The previous readiness check also succeeded on 503 which would sometimes happen before the config was finished propagating (I think as a result of config from a previous test). In this change the test looks for a 403. 
Running with `untilItFails` locally, it would fail after <7 runs before the fix and did not fail after 15 runs after the fix.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6356


sam-heilbron (approved):
🎉

PR #6394, opened 2022-04-29
feat(DelegatedHttpGateway): allow top-level config setting/sharing

gunnar-solo:
# Description
This PR expands on the functionality of `DelegatedHttpGateway` and `MatchableHttpGateway`.  It allows for anscestry-level setting of TLS and HCM configuration.  Phrased differently:  configuration can now be set for a single `DelegatedHttpGateway` that will be propagated to all `MatchableHttpGateway`s.  Setting the `preventChildOverrides` flag specifies that parent config is to be prioritized over child.  For example:

* `preventChildOverrides=false | {"parent":{"foo":"bar"}, "child":{"foo":"baz"}} --> {"foo":"baz"}`
* `preventChildOverrides=true  | {"parent":{"foo":"bar"}, "child":{"foo":"baz"}} --> {"foo":"bar"}`

# Context
This feature is intended to be useful to a multi-team gloo management environment.  Picture a top-level, well-informed team managing a single `DelegatedHttpGateway` with a bunch of useful defaults.  Then, individual teams/microservices manage their own `MatchableHttpGateway`(s)

# Checklist:


projects/gateway/api/v1/gateway.proto
```diff
@@ -128,6 +129,21 @@ message DelegatedHttpGateway {
         // Delegate to the MatchableHttpGateway that match the given selector.
         selectors.core.gloo.solo.io.Selector selector = 4;
     }
+
+    // Used as a meta modifier to the `options` and `ssl_config` fields in a DelegatedHttpGateway.  When set, provided ancestor config cannot be overriden by matched HttpGateways.
```
bewebi, line 133: By `options` do we mean `http_connection_manager_settings`?

projects/gateway/pkg/translator/hybrid_translator_test.go
```diff
@@ -288,6 +290,288 @@ var _ = Describe("Hybrid Translator", func() {
...
+						// evaluate results
+						hcm_after := child.GetHttpGateway().GetOptions().GetHttpConnectionManagerSettings()
+						ssl_after := child.GetMatcher().GetSslConfig()
+
+						Expect(hcm_after.GetSkipXffAppend()).To(Equal(true))
+						Expect(ssl_after.GetOneWayTls().GetValue()).To(Equal(true))
+					})
+				})
+
+				Context("preventChildOverrides == true", func() {
```
bewebi, line 474: `true` should be `false` here

projects/gateway/pkg/translator/hybrid_translator_test.go
```diff
@@ -288,6 +290,288 @@ var _ = Describe("Hybrid Translator", func() {
...
+						SkipXffAppend: true,
+					}
+					hcm_false = &hcm.HttpConnectionManagerSettings{
+						SkipXffAppend: false,
+					}
+					hcm_empty = &hcm.HttpConnectionManagerSettings{}
+
+				})
+
+				Context("preventChildOverrides == true", func() {
```
bewebi, line 374: All of the tests in this context and the next one seem like they could be condensed to a single test table

bewebi (commented):
Typo/style suggestion in tests
Additionally some style/cleanliness suggestions were made in Slack which don't need to be taken, but if not please clarify that there is no intent to

gunnar-solo:
slack discussions focused around style changes WRT the inheritance logic written in to `hybrid_translator.go`.  Transcribing asks/rationales, here:

## Ask 1:  we should set `parent` and `child` as "empty" structs, rather than `nil`
This would have simplified some complex(ish) `nil` handling, and allow for simpler code flow.  UNFORTUNATELY, I don't think this is viable, given that an `empty` struct is not the same as a struct defaulting a boolean to `false`

## Ask 2:  we should condense the logic, such that `PreventChildOverrides` is the top level logic gate
This is _ok_, and would work.  I think it's not very clean, given that the above change isn't going in.  If(when) more fields get added to this logic, it'd be cleaner to have the gates situated how they are, IMO

gunnar-solo:
/kick [Step #10 - "test": [Fail] Access Log in memory Access Logs Grpc [It] can stream access logs](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=63c0449e-e40a-44c9-816a-120dc09f8c94)

nfuden (commented):
Can we add a test around overwriting of defaults. Namely I would be interested in seeing the merge of onewaytls if you have not set the option explicitly.

Ie merge behaviour on our *wrappers types when nested and how we currently dont force users to set all options

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -160,6 +162,64 @@ func (t *HybridTranslator) computeMatchedListener(
...
+		if preventChildOverrides {
+			// merge, preferring parentHcmOptions
+			mergo.Merge(childHcmOptions, parentHcmOptions, mergo.WithOverride)
+		} else {
+			// merge, preferring childHcmOptions
+			mergo.Merge(childHcmOptions, parentHcmOptions)
+		}
+	}
+
+	if matchableHttpGateway.GetHttpGateway().GetOptions() != nil {
```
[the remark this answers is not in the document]
gunnar-solo: good catch!  That's a legitimate logic bug.  Calling out the parallel logic of
```go
	if matchableHttpGateway.GetMatcher() != nil {
		matchableHttpGateway.GetMatcher().SslConfig = childSslConfig
	}
``` 

is _not_ a bug, though, since no matcher means a worthless matchableHttpGateway, anyway

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -160,6 +162,68 @@ func (t *HybridTranslator) computeMatchedListener(
...
+	if matchableHttpGateway.GetMatcher() != nil {
+		childSslConfig = matchableHttpGateway.GetMatcher().GetSslConfig()
+	}
+
+	if childSslConfig == nil {
+		// use parentSslConfig exactly as-is
+		childSslConfig = parentSslConfig
+	} else if parentSslConfig == nil {
+		// use childSslConfig exactly as-is
+	} else if childSslConfig != nil && parentSslConfig != nil {
```
nfuden, line 180:
```suggestion
	if childSslConfig == nil {
	  childSslConfig = parentSslConfig
	} else if parentSslConfig != nil{```

gunnar-solo: I agree that this logic is _correct_ and can be factored as such.  I'd prefer to have the larger cases explicitly written out, though, as I find it easier to parse/not wonder about the other cases when reading

nfuden: good enough for me

nfuden (approved):

PR #6396, opened 2022-04-29
Doc fixes and backports for 1.11.x

Rachael-Graham:
Doc fixes for 1.11.x branch week of 04/25, including version bumps and backports.


docs/content/guides/graphql/stitching.md
```diff
@@ -0,0 +1,219 @@
...
+
+{{% notice warning %}}
+Stitching is a beta feature. Do not use this feature in a production environment.
+{{% /notice %}}
+
+When you use GraphQL in Gloo Edge, you can stitch multiple schemas together to expose one unified GraphQL server to your clients.
+
+Consider a cluster that has two different gRPC services, `user-svc` and `product-svc`. These services are either native GraphQL servers, or have been converted to GraphQL via automatic schema discovery. Each service has similar information that you might want to provide as part of a unified data model. 
+
+Typically, clients must stitch together the services in the frontend. With Gloo Edge, you can instead stitch the GraphQL schemas for these services together in the backend, and expose a unified GraphQL server to your clients. This frees your clients to consider only what data that they want to fetch, not how to fetch the data.
```
Nadine2016, line 15:
```suggestion
Typically, clients must stitch together the services in the frontend. With Gloo Edge, you can instead stitch the GraphQL schemas for these services together in the backend, and expose a unified GraphQL server to your clients. This enables your clients to consider only what data that they want to fetch, not how to fetch the data.
```

docs/content/guides/graphql/stitching.md
```diff
@@ -0,0 +1,219 @@
...
+
+Follow along with the user and product service example.
+
+## Reviewing each service's configuration
+
+To understand how stitching occurs, consider the data model and example queries for both services, starting with the user service. 
+
+**User service**: The `GraphQLApi` resource for the user service defines a partial type definition for the `User` type, and a query for how to get the full name of a user given the username.
```
Nadine2016, line 23: why is this only a partial user definitition?

Nadine2016: ok I got it reading further :-) 

However, I was wondering how Gloo Edge knows what user ID belongs to which user? If I am reading correctly, the user service has username and full name, and the product service has the user ID. To stitch them together, I would think that there must be something that's overlapping, such as the ID to know which ID belongs to which name. Or maybe I missed a detail in the stitching.

docs/content/guides/graphql/stitching.md
```diff
@@ -0,0 +1,219 @@
...
+        User:
+          selectionSet: '{ username }'
+          queryName: GetUser
+          args:
+            username: username
+    - name: product-svc
+      namespace: product-app
+```
+
+As a result, Gloo Edge generates a **stitched service**. From this one stitched service, a client can provide the product ID, and recieve the product name, the username of the seller, the user ID of the seller, _and_ the full name of the seller.
```
Nadine2016, line 164:
```suggestion
As a result, Gloo Edge generates a **stitched service**. From this one stitched service, a client can provide the product ID, and receive the product name, the username of the seller, the user ID of the seller, _and_ the full name of the seller.
```

docs/content/guides/graphql/stitching.md
Rachael-Graham: AFAIK that would happen in the details you provided when you created the services themselves. For example in [this file](https://github.com/solo-io/graphql-bookinfo/blob/main/openapi/detail.yaml) (this is for a different demo that uses bookinfo), all of the pieces are listed together. and this [json file](https://github.com/solo-io/graphql-bookinfo/blob/main/etc/data.json) is an example of the values your services would use for each of those fields

Nadine2016: I think if that is the case, we should maybe just show how an example service configuration would look like so that users know what to do. could also be a prerequisites section or something similar.

Nadine2016 (approved):

Nadine2016 (approved):

docs/content/guides/graphql/stitching.md
Rachael-Graham: FYI, we chatted in Slack and I added the userId to the User type for the user service, just to ensure that there is overlap for the connection between username and userId. Will be looking into this more later this week.

PR #6412, opened 2022-05-04
Feature/caching service

nfuden:
# Description

Add basic caching setup for enterprise caching service
BOT NOTES: 
resolves https://github.com/solo-io/caching-service/issues/1


projects/gloo/pkg/syncer/cachingservice/cachingservice_translator_syncer.go
```diff
@@ -0,0 +1,60 @@
+package cachingservice
```
sam-heilbron, line 1: I'm not sure this is necessary. This is an enterprise only feature, and we will only load the trnaslator_syncer_extension in enterprise. I think we can even remove the rate-limit and ext-auth ones but that can be done separately. They are around becuase there was a time when Gloo defined placeholders and enterprise would override those (UPgradableTranslatorSyncerExtension). We removed the Upgradable behavior and now Enterprise just defines the extensions that it wants directly.

projects/gloo/api/v1/options.proto
```diff
@@ -113,6 +114,8 @@ message HttpListenerOptions {
     enterprise.gloo.solo.io.Settings extauth = 10;
     // Enterprise-only: Settings for the rate limiting server itself
     ratelimit.options.gloo.solo.io.Settings ratelimit_server = 11;
+    // Enterprise-only: Settings for the cache server itself
+    caching.options.gloo.solo.io.Settings caching = 17;
```
sam-heilbron, line 118: Our currently far from perfect plugin that warns users about enterprise config (https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/plugins/enterprise_warning/plugin.go#L113) could be extended to warn on configuration. The plugin is only executed for open source.

sam-heilbron (commented):
In the EE PR we expose a helm value but we don't seem to consume that in the OSS chart. Am I missing something?

projects/gloo/pkg/syncer/cachingservice/cachingservice_translator_syncer.go
nfuden: Sounds good it took me a while to understand why we did this on the other plugins. If we are fine with it then I am happy to excise this.

projects/gloo/api/v1/options.proto
nfuden: Good point have done so

projects/gloo/api/v1/settings.proto
```diff
@@ -372,6 +373,10 @@ message Settings {
     // If an auth server name is not supplied on a route, the default auth server will be applied.
     map<string, enterprise.gloo.solo.io.Settings> named_extauth = 33;
 
+    // Enterprise-only: Settings for the rate limiting server itself
```
sam-heilbron, line 376: nit: caching server, not rate limiting

projects/gloo/pkg/syncer/cachingservice/cachingservice_translator_syncer.go
sam-heilbron: Yeah this was leftover from an earlier state. In https://github.com/solo-io/gloo/pull/5803 we made some changes to our Upgradable interface since it was leaky and required having open source equivalent placeholders for enterprise features. I didn't make changes to the syncer extensions as well, but probably should have

projects/gloo/api/v1/settings.proto
```diff
@@ -372,6 +373,10 @@ message Settings {
...
+    caching.options.gloo.solo.io.Settings caching_server = 36;
```
sam-heilbron, line 377: Is this configurable through Helm? If not, should it be?

sam-heilbron (commented):
API change looks good. One comment needs to be updated, and then a question about Helm.

projects/gloo/api/v1/settings.proto
nfuden: It is https://github.com/solo-io/gloo/blob/26bd7b6f0dcbec299bcd6260921ea3914f6def87/install/helm/gloo/templates/18-settings.yaml#L129 

Sorry for the slow response I thought I hit comment here a while ago

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -126,6 +126,11 @@ spec:
 {{- toYaml .Values.settings.rateLimit | nindent 4 }}
 {{- end }}
 
+{{- if .Values.settings.cachingService }}
+  caching_server:
```
sam-heilbron, line 130: Can we add a helm test to confirm that this can be applied? Since our CRD validation schemas require camelCase, I believe that this snake_case will prevent the resource from being applied to a cluster.

nfuden: I found a more elegant way to handle this. 
Namely we already perform injections into settings for ratelimit. So I am following that so as to not have to worry about it here and to avoid leaking implementation details.

projects/gloo/api/external/envoy/extensions/cache/grpc/config.proto
```diff
@@ -0,0 +1,30 @@
...
+import "validate/validate.proto";
+
+option (solo.io.udpa.annotations.file_status).package_version_status = ACTIVE;
+
+// [#protodoc-title: Cache]
+
+// [#proto-status: experimental]
+message GrpcCacheConfig {
+
+  .solo.io.envoy.config.core.v3.GrpcService service = 2 [(validate.rules).message = {required: true}];
```
jenshu, line 27: why does this start at 2?

projects/gloo/api/v1/enterprise/options/caching/caching.proto
```diff
@@ -0,0 +1,23 @@
...
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
+option (extproto.clone_all) = true;
+
+
+
+
+message Settings {
+  core.solo.io.ResourceRef caching_service_ref = 1;
+  repeated .solo.io.envoy.type.matcher.v3.StringMatcher allowed_vary_headers = 2;
```
jenshu, line 20: can you add some comments for these fields?

projects/gloo/api/v1/settings.proto
```diff
@@ -372,6 +373,10 @@ message Settings {
     // If an auth server name is not supplied on a route, the default auth server will be applied.
     map<string, enterprise.gloo.solo.io.Settings> named_extauth = 33;
 
+    // Enterprise-only: Settings for the caching server itself
```
jenshu, line 376: is this meant to be a global option, that can be overridden by the listener-specific options from options.proto? (if that's the case, can you clarify that in the comments?)

projects/gloo/api/v1/enterprise/options/caching/caching.proto
jenshu: also super-nit: there's a lot of extra whitespace in this file :)

nfuden: done and done

jenshu (dismissed):

jenshu (approved):

PR #6422, opened 2022-05-06
Truncate long proxy statuses

elcasteel:
# Description

In large deployments, when every warning is added to the proxy status, the field can grow too large to persist to etcd. This change introduces an environment variable `PROXY_STATUS_MAX_SIZE`. When the status is longer than the provided size (in bytes), the status will be truncated. This way, users can still see some of the warnings associated with their proxies but the size of the status field can be limited. 

For example install gloo with the following values override: 
```
cat <<EOF | glooctl install gateway -f  --values -
gateway:
  persistProxySpec: true
gloo:
  deployment:
    customEnv:
      - name: "PROXY_STATUS_MAX_SIZE"
        value: "100"
EOF
```

And create virtual services pointing to upstreams that do not exist. The proxy will have a status with 100 characters and a message that the status was truncated:
```
namespacedStatuses:
  statuses:
    gloo-system:
      reason: "warning: \n  Route Warning: InvalidDestinationWarning. Reason: *v1.Upstream
        { gloo-system.default-dis\nStatus truncated, see logs for details."
      reportedBy: gloo
      state: Warning
```
# Context

# Checklist:

### DIFF
--- a/projects/gateway/pkg/syncer/translator_syncer.go
+++ b/projects/gateway/pkg/syncer/translator_syncer.go
@@ -3,6 +3,7 @@ package syncer
 import (
 	"context"
 	"fmt"
+	"os"
 	"sort"
 	"sync"
 	"time"
@@ -37,6 +38,7 @@ type TranslatorSyncer struct {
 	translator         translator.Translator
 	statusSyncer       statusSyncer
 	managedProxyLabels map[string]string
+	proxyStatusMaxSize string
 }
 
 func NewTranslatorSyncer(ctx context.Context, writeNamespace string, proxyWatcher gloov1.ProxyClient, proxyReconciler reconciler.ProxyReconciler, reporter reporter.StatusReporter, translator translator.Translator, statusClient resources.StatusClient, statusMetrics metrics.ConfigStatusMetrics) *TranslatorSyncer {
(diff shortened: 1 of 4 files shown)


projects/gloo/pkg/api/compress/compress_test.go
```diff
@@ -135,6 +135,54 @@ var _ = Describe("Compress", func() {
 
 			Expect(status1).To(BeEquivalentTo(status2))
 		})
+		It("should truncate the status reason when annotated with max length", func() {
```
jenshu, line 138: if not too much trouble, could you also add an e2e test?

EItanya:
Just a quick question on this approach, does the user really care about the size in bytes? Would it be potentially more useful to expose this as `MAX_DISPLAYED_WARNINGS` or something to that effect.

EItanya:
Is the size of the proxy and the status computed separately? If not then I'm not sure how this solves the fundamental issue.

projects/gloo/pkg/api/compress/compress_test.go
elcasteel: Since this doesn't really affect how we configure envoy, I decided it made more sense to add some more checks to our other translation unit tests.

elcasteel:
They have the proxy spec compressed and they can set the maximum size of of the status to be something less than the maximum size in etcd minus the normal proxy spec size.

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -50,6 +52,9 @@ func NewTranslatorSyncer(ctx context.Context, writeNamespace string, proxyWatche
 			"created_by": "gloo-gateway-translator",
 		},
 	}
+	if pxStatusSizeEnv := os.Getenv("PROXY_STATUS_MAX_SIZE_BYTES"); pxStatusSizeEnv != "" {
```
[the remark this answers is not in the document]
jenshu: can you also update the PR description with the correct name?

sam-heilbron (commented):
LGTM! Not approving since there are still open questions, but happy to do so once they're resolved

jenshu (approved):

Issue #6195, opened 2022-03-31
Istio Compat: Route Kubernetes Upstreams based on the Kube Service rather than the Pod VIP


murphye:
### Version

1.10.x (latest stable)

### Is your feature request related to a problem? Please describe.

Right now, there is only basic mTLS integration with Gloo Edge and Istio. SDS is hijacked from an Istio sidecar, but otherwise Gloo Edge Envoy proxy remains outside the service mesh.

To fully engage Gloo Edge into the Istio service mesh, an Istio sidecar is needed for all traffic routing, therefore requiring double proxying of requests inside of the `gateway-proxy` pod. This is a reasonable trade-off at the moment to perform this extra proxying in order to leverage the benefits of routing requests from Gloo Edge into the Istio service mesh and have the ability to perform failover routing to an east-west gateway, as an example.

There is a problem in having Gloo Edge be compatible with Istio in that Kubernetes destinations (Upstreams) route to the Pod rather than the Service VIP. Without routing through the Service, Strict mTLS in Istio will fail to work.

### Describe the solution you'd like

Here is an example to show how this can work with a Static Upstream (which is not a good solution for this problem, only to show it can work):

Here is an existing Upstream that doesn't work correctly with Istio (i.e. fails with Strict mTLS):

**Example 1**:
```
          single:
            upstream:
              name: petstore-petstore-8080
              namespace: gloo-system
---
kind: Upstream
metadata:
  name: petstore-petstore-8080
  namespace: gloo-system
spec:
  kube:
    selector:
      app: petstore
    serviceName: petstore
    serviceNamespace: petstore
    servicePort: 8080
```

However, when this Upstream is defined as static and referencing the `petstore.petstore`, it WILL work in Istio even with Strict mTLS. This is because traffic is routed via the Service rather than directly to the Pod.

**Example 2**:
```
          single:
            upstream:
              name: petstore-static-upstream
              namespace: gloo-system
---
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: petstore-static-upstream
  namespace: gloo-system
spec:
  static:
    hosts:
      - addr: petstore.petstore
        port: 8080
````

**Desired Solution**

Offer a new Helm chart configuration that will change how Gloo Edge routes to Upstream Kubernetes destinations based on Service rather than Pod VIP. It might be something like:

```
gloo.settings.kubernetesDestinationsAsServices
```

[...]

PR #6312, opened 2022-04-13
POC: Full istio Sidecar Support

ben-taussig-solo:
# Description

 - This PR introduces support for the `"global.istioIntegration.enableIstioSidecarOnGateway` helm value
 - When enabled, the `sidecar.istio.io/inject` annotation will be set to `true` on any gateway-proxy deployment
   - Additionally, the `traffic.sidecar.istio.io/excludeInboundPorts` annotation will be set to contain the values of any ports reserved by the deployment for http/https traffic. E.G., it will take the value `8080,8443` for the default gateway-proxy deployment, assuming that the default http/https ports have not been overwritten
 - This PR is fairly bare-bones. In the future we intend to introduce testing around Istio sidecar functionality.

# Context

 - From the issue:
    > To fully engage Gloo Edge into the Istio service mesh, an Istio sidecar is needed for all traffic routing, therefore requiring double proxying of requests inside of the gateway-proxy pod. 

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

Issue #6196, opened 2022-03-31
Istio Compat: Full Istio Sidecar on Gloo Edge Gateway


murphye:
### Version

1.10.x (latest stable)

### Is your feature request related to a problem? Please describe.

Right now, there is only basic mTLS integration with Gloo Edge and Istio. SDS is hijacked from an Istio sidecar, but otherwise Gloo Edge Envoy proxy remains outside the service mesh.

To fully engage Gloo Edge into the Istio service mesh, an Istio sidecar is needed for all traffic routing, therefore requiring double proxying of requests inside of the `gateway-proxy` pod. This is a reasonable trade-off at the moment to perform this extra proxying in order to leverage the benefits of routing requests from Gloo Edge into the Istio service mesh and have the ability to perform failover routing to an east-west gateway, as an example.

Given the benefits, a higher Istio integration is needed when Gloo Edge is intalled.

### Describe the solution you'd like

In order to make Gloo Edge's `gateway-proxy` work with the `istio-proxy` sidecar, 2 things are needed:

**1. Add the Istio Sidecar to the `gateway-proxy` Deployment **

```
kubectl patch deployment gateway-proxy -n gloo-system --patch '{"spec": {"template": {"metadata": {"labels": {"sidecar.istio.io/inject": "true"}}}}}'
```

**2. Add Exclude Inbound Ports from Sidecar to the `gateway-proxy` Deployment **

This will prevent the Istio sidecar from capturing incoming requests. They will be routed instead directly to Gloo's gateway proxy.

```
kubectl patch deployment gateway-proxy -n gloo-system --patch '{"spec": {"template": {"metadata": {"annotations": {"traffic.sidecar.istio.io/excludeInboundPorts": "8080,8443"}}}}}'
```

### Describe alternatives you've considered

I have tried the existing Gloo Edge Istio mTLS integration, and it doesn't provide the needed functionality to leverage Istio at the gateway.

### Additional Context

1. This would also enable Gloo Portal, via Gloo Edge, to integrate with Istio and Gloo Mesh.
2.  This provides a better avenue for existing Gloo Edge users to integrate with Istio
3.  Here is a demo project to show how to integrate Gloo Edge gateway proxy into the Istio service mesh:  https://github.com/murphye/demo-gloo-edge-istio

murphye:
See also: https://github.com/solo-io/gloo/issues/6195

PR #6441, opened 2022-05-10
Istio compat - use kub service ip for upstreams

mkazin:
# Description

This change includes support for setting endpoints based on service ip addresses rather than dns names, which is required for integration with Istio.

This required breaking up a large and complicated function which was making code changes difficult. 

# Context
This is phase two of the work required to prepare Edge for full Istio integration (see s-p 3461)
It follows the work done in #6196 , utilizing the helm value established there to enable this code change.

# Testing
This was tested manually by repeating the method described in the earlier work to enable injection ([see here](https://github.com/solo-io/gloo/pull/6312#issuecomment-1098467998)), along with steps [described here](https://github.com/murphye/demo-gloo-edge-istio) to generate an error from the petstore sevice when strict mTLS is enabled .

A config dump shows the desired endpoint configuration:
```

 "endpoint_config": {
  "@type": "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment",
  "cluster_name": "kube-svc:petstore-petstore-8080_petstore-16672974176135880674",
  "endpoints": [
   {
    "locality": {},
    "lb_endpoints": [
     {
      "endpoint": {
       "address": {
        "socket_address": {
         "address": "10.40.0.174",
         "port_value": 8080
        }
       },
       "health_check_config": {}
      },
      "health_status": "HEALTHY",
      "metadata": {
       "filter_metadata": {
        "envoy.lb": {
         "service": "petstore"
        }
       }
      },
      "load_balancing_weight": 1
     }
    ]
   }
```

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6195

### DIFF
--- a/changelog/v1.12.0-beta13/istio-ip-routing.yaml
+++ b/changelog/v1.12.0-beta13/istio-ip-routing.yaml
@@ -2,4 +2,4 @@ changelog:
 - type: NEW_FEATURE
   issueLink: https://github.com/solo-io/gloo/issues/6195
   resolvesIssue: true
-  description: Add support for enabling ip routing when injecting istio sidecar.
+  description: Add support for istio integration in endpoint discovery, routing to service IP rather than pod IP.
--- a/changelog/v1.12.0-beta13/istio-ip-routing.yaml
+++ b/changelog/v1.12.0-beta13/istio-ip-routing.yaml
@@ -1,0 +1,5 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/6195
+  resolvesIssue: true
+  description: Add support for enabling ip routing when injecting istio sidecar.
--- a/install/helm/gloo/templates/1-gloo-deployment.yaml
+++ b/install/helm/gloo/templates/1-gloo-deployment.yaml
@@ -222,7 +222,7 @@ spec:
             value: {{ .Values.gloo.logLevel }}
         {{- end}}
         {{- if .Values.global.istioIntegration.enableIstioSidecarOnGateway }}
-          - name: ENABLE_ISTIO_SIDECAR
+          - name: ENABLE_ISTIO_INTEGRATION
             value: "true"
         {{- end}}
 {{- if not .Values.global.glooMtls.enabled }}
--- a/install/helm/gloo/templates/1-gloo-deployment.yaml
+++ b/install/helm/gloo/templates/1-gloo-deployment.yaml
@@ -221,6 +221,10 @@ spec:
           - name: LOG_LEVEL
             value: {{ .Values.gloo.logLevel }}
         {{- end}}
+        {{- if .Values.global.istioIntegration.enableIstioSidecarOnGateway }}
+          - name: ENABLE_ISTIO_SIDECAR
+            value: "true"
+        {{- end}}
 {{- if not .Values.global.glooMtls.enabled }}
         readinessProbe:
           tcpSocket:
--- a/projects/gloo/pkg/plugins/kubernetes/eds.go
+++ b/projects/gloo/pkg/plugins/kubernetes/eds.go
@@ -5,8 +5,10 @@ import (
 	"encoding/binary"
 	"fmt"
 	"hash/fnv"
+	"os"
 	"sort"
 	"strings"
+	"time"
 
 	errors "github.com/rotisserie/eris"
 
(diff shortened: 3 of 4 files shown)


projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -257,55 +265,87 @@ func filterEndpoints(
...
-							continue
-						}
-						// pod hasn't been assigned address yet
-						if addr.IP == "" {
-							continue
-						}
-					}
-					key := Epkey{addr.IP, port, podName, podNamespace, usRef}
+
+				if filterToIpAddresses {
```
nfuden, line 274: Filtertoipaddresses doesnt mean much to me. Do you think we can either change the name to imply that its due to hitting services or just putting a comment here?

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -257,55 +265,87 @@ func filterEndpoints(
...
+			}
+		}
+		key := Epkey{addr.IP, port, podName, podNamespace, usRef}
+		copyRef := *usRef
+		endpointsMap[key] = append(endpointsMap[key], &copyRef)
+	}
+	return warnings
+}
+
+func findPortForEndpointSubset(subset kubev1.EndpointSubset, singlePortService bool, kubeServicePort *kubev1.ServicePort) uint32 {
```
nfuden, line 326: findFirstPortInEndpointSubsets?

nfuden (commented):
Do we think we can have a test with this os env or is that a large lift?

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -203,50 +207,54 @@ func (c *edsWatcher) watch(writeNamespace string, opts clients.WatchOpts) (<-cha
...
+		if len(svc.Spec.Ports) == 1 {
+			if spec.GetServicePort() == uint32(svc.Spec.Ports[0].Port) {
+				return &svc.Spec.Ports[0], true
+			}
+		}
+		for _, port := range svc.Spec.Ports {
+			if spec.GetServicePort() == uint32(port.Port) {
+				return &port, false
+			}
+		}
```
nfuden, line 232: Since we are already reframing this code we could consider calling out the fact that we return whether this is a single port service.
Either as a method comment and/or doing something like the following

```suggestion
		for _, port := range svc.Spec.Ports {
			if spec.GetServicePort() == uint32(port.Port) {
				return &port,  (len(svc.Spec.Ports) == 1)
			}
		}
```

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -221,6 +221,10 @@ spec:
           - name: LOG_LEVEL
             value: {{ .Values.gloo.logLevel }}
         {{- end}}
+        {{- if .Values.global.istioIntegration.enableIstioSidecarOnGateway }}
+          - name: ENABLE_ISTIO_SIDECAR
```
nfuden, line 225: nit: For future reusability of this feature (say if this is not only triggerable on istio) I would love a name for the env var to be descriptive of this change. Say KUBE_IP_FOR_SERVICES (preferably better than that but cant think of a good one off the top of my head)

projects/gloo/pkg/plugins/kubernetes/eds.go
mkazin: Much better!

install/helm/gloo/templates/1-gloo-deployment.yaml
mkazin: I don't know if getting more specific about the details of implementation is the right direction. Should more code changes become necessary to support istio, that name would become obsolete.
For the same reason, I'm also removing the "sidecar" mention (which isn't part of the code implementation in eds).
Rather, I'm suggesting ENABLE_ISTIO_INTEGRATION.

projects/gloo/pkg/plugins/kubernetes/eds.go
mkazin: As I mentioned elsewhere, I'm switching this name to indicate the purpose of istio integration rather the implementation details, adding a comment to explain the required implementation.

mkazin:
> Do we think we can have a test with this os env or is that a large lift?

Not a comprehensive test to verify the ip/hostname translation. Given no real testing on this module that'd take more time than I want to spend now.

I've extracted  the environment variable logic into its own method and added the fairly trivial testing it could use.

changelog/v1.12.0-beta13/istio-ip-routing.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/6195
+  resolvesIssue: true
+  description: Add support for enabling ip routing when injecting istio sidecar.
```
murphye, line 5: I would be more descriptive here just for reference. Routing to service IP rather than pod IP

ben-taussig-solo (commented):
Looks good to me, waiting for all comments to be resolved and then I'm happy to sign off on this :)

changelog/v1.12.0-beta13/istio-ip-routing.yaml
mkazin: Thanks- how's this new one I changed it to?
> Add support for istio integration in endpoint discovery, routing to service IP rather than pod IP.

mkazin:
/kick build-bot
different known flake- [Access Log in memory Access Logs Grpc](https://storage.googleapis.com/solo-public-build-logs/log-21f6849a-ce2d-423b-9d66-ee38890ada8b.txt)

nfuden (approved):

PR #6465, opened 2022-05-17
Reconcile Proxies using both new and old labels

sam-heilbron:
# Description

Reconcile proxies using labels that follow the current naming convention, and the previous naming convention.

# Context

Gloo maintains an internal CR, known as a [Proxy](https://github.com/solo-io/gloo/blob/e5df9c889c00a4756fc468432160732706554257/projects/gloo/api/v1/proxy.proto#L30). We identify these Proxies with [unique labels](https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/projects/gateway/pkg/syncer/translator_syncer.go#L49). We rely on these labels to both read Proxies (to know which to reconcile) and write Proxies (to know which to reconcile on the next iteration). Recently we changed these labels on Proxies, which means that after an upgrade from a version with the old label, to a version with the new label, the Proxy controller will attempt to reconcile the existing persisted Proxies. However, since the Proxy uses the old label, but the new controller looks for the new label, it will find no proxies, and expect to create one. Our toolkit for modifying resources will then error, since it is attempting to create a resource that is meant for deletion.

# Checklist:

### DIFF
--- a/test/kube2e/gateway/gateway_test.go
+++ b/test/kube2e/gateway/gateway_test.go
@@ -218,6 +218,118 @@ var _ = Describe("Kube2e: gateway", func() {
 		})
 	})
 
+	Context("Proxy reconciliation", func() {
+
+		var (
+			virtualService *gatewayv1.VirtualService
+		)
+
+		BeforeEach(func() {
+			// Create virtual service routing directly to the testrunner service
+			dest := &gloov1.Destination{
+				DestinationType: &gloov1.Destination_Kube{
+					Kube: &gloov1.KubernetesServiceDestination{
+						Ref: &core.ResourceRef{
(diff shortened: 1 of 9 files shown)


test/kube2e/gateway/gateway_test.go
```diff
@@ -218,6 +218,118 @@ var _ = Describe("Kube2e: gateway", func() {
 		})
 	})
 
+	Context("Proxy reconciliation", func() {
```
sam-heilbron, line 221: I was able to run this test locally, against a branch without my changes (ie the latest 1.11 release) and it failed

elcasteel (approved):

PR #6483, opened 2022-05-19
Add READMEs around Envoy receiving static and dynamic configuration

sam-heilbron:
# Description
Add READMEs around where Envoy receives its static and dynamic configuration.

# Context
These are core components of Gloo Edge, and we need to have more expansive internal documentation.

## Envoyinit
Update the `envoyinit` project to include a README, add comments to code, and export functions to be shared with Solo-Projects.

## xDS
xDS is responsible for serving configuration both to service proxies (Envoy) and custom Edge service (ext-auth, rate-limit). Generate a README to make it easier to onboard into the space.

# Checklist:


jackstine (commented):
LGTM.  I checked the links in the README and all looks good.  Liking the comments!

I will leave the approve open for others to add.

projects/envoyinit/cmd/main.go
```diff
@@ -1,59 +1,105 @@
...
+	bootstrapConfig, err := getAndTransformConfig(inputPath)
 	if err != nil {
 		log.Fatalf("initializer failed: %v", err)
 	}
 
-	// best effort - write to a file for debug purposes.
-	// this might fail if root fs is read only
-	writeConfig(outCfg)
+	// 2. Write to a file for debug purposes
+	err = ioutil.WriteFile(outputPath, []byte(bootstrapConfig), 0444)
```
ianmacclancy, line 46: New to codebase so not familiar with standard practice yet - Why is this using shorthand for the word error rather than using a descriptive name like "transformConfigError"? It also looks like the same error variable is being re-assigned in different error cases which to me seems unsafe as future modifications could use the variable incorrectly.

mkazin (commented):
Only the one comment on a readme file. I'm happy if everyone else is.

projects/envoyinit/cmd/main.go
sam-heilbron: Yeah a great idea. I think we've historically just re-used err, but it's not the most explicit and can certainly cause issues with re-assignment. I'll update

projects/envoyinit/README.md
```diff
@@ -0,0 +1,98 @@
...
+It can be useful to run the Envoy proxy, without the control-plane, as a way of validating behavior. [hack/envoy.yaml](./hack/envoy.yaml) provides example bootstrap that can be used
+
+`docker run 
+    --rm 
+    -ti 
+    -p 8000:8000 
+    -p 19000:19000 
+    -v $(pwd)/hack/envoy.yaml:/etc/envoy/envoy.yaml:ro
+    -l trace 
+    quay.io/solo-io/gloo-envoy-wrapper:1.11.11
```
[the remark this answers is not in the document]
sam-heilbron: Good call! I had changed the format to make it more readable for devs, but forgot to re-test it. Will update

projects/envoyinit/README.md
```diff
@@ -0,0 +1,98 @@
...
+`
+
+After running this, you should see a lot of Envoy logs:
```
jbohanon, line 72: I've found [--component-log-level](https://www.envoyproxy.io/docs/envoy/v1.8.0/operations/cli#cmdoption-component-log-level) to be extremely handy and maybe worth calling out here

projects/envoyinit/cmd/main.go
jbohanon: It's very idiomatic Go to use `err` all over the place. It's generally assumed that the value in `err` will be assessed and used before reassignment.

projects/envoyinit/README.md
sam-heilbron: I've never used that before, but looks super useful! Ill add

jbohanon: Yeah, it helps improve the signal-to-noise when you know sorta what you're looking for 👍

projects/envoyinit/cmd/main.go
```diff
@@ -1,59 +1,105 @@
 package main
 
 import (
+	_ "github.com/solo-io/gloo/projects/envoyinit/hack/filter_types"
```
sam-heilbron, line 4: This used to live in the top level 'hack' directory. Since the filter types are specifically used for envoy serialization, I opted to move these values directly into the project. My other hope is that as we start to move resources to where they are used, we can have developers run codegen only in certain projects, as necessary.

projects/envoyinit/cmd/main.go
```diff
@@ -1,59 +1,105 @@
...
-	inputFile := inputCfg()
-	outCfg, err := utils.GetConfig(inputFile)
+	envoyExecutable := GetEnvoyExecutable()
+	inputPath := GetInputConfigPath()
+	outputPath := GetOutputConfigPath()
+
+	RunEnvoy(envoyExecutable, inputPath, outputPath)
+}
+
+// RunEnvoy run Envoy with bootstrap configuration injected from a file
```
sam-heilbron, line 37: Note: this is a non functional change. However solo-projects writes the same code, and I figured we're better off just exporting the functionality and importing it.

projects/envoyinit/cmd/main.go
sam-heilbron: Yeah I'm torn on this. I find the proposal makes the code more readable, but I'm happy to stick with go conventions. @nfuden do you have strong feelings about this?

jbohanon: I'd also point out that `errorDescriptionErr` is a common pattern when defining errors as globals to be returned and checked against

nfuden: Err is short and readable. One will usually evaluate errors soon after getting them. In the error evaluation you can see the err often being checkagainst known errorstrings that follow a naming format similar to what you are asking for. Therefore in code if an error can be errNotFound or errDenied it doesn't add much to name the error value it's own thing. Also minor mem optimization due to reuse of variable. There are cases where you should name your err something longer but that's generally in a multi err space where the error scope of usefulness is greater than its immediate environs. 

A good read on using errors as values https://go.dev/blog/errors-are-values#:~:text=all%20the%20time%2C%20something%20must,this%20pattern%2C%20and%20stop%20there.

ianmacclancy: I think in the context of only one error in the method there is no issue. Err is short and understandable but only when it's the only use and specific to a single error which seems to be the case in the document you linked. I called it out here as "err" variable is used for 2 entirely different errors (line 40 - getAndTransformConfig error and then again line 58 for syscall.exec). I have seen re-use of variable that share names cause absolute nightmares when it comes to debugging and code legibility. When it can mean different things 10 lines apart that get a bit dodgy. As I read the code I have to keep track of what the variable is now referencing. Its storing errors from two separate processes with not change in name. This is not behavior you want to encourage with any code. The extremely small memory optimization is entirely outweighed by the loss of legibility and higher likelihood that the next person editing the code accidentally misses that you re-assigned the variable. While good to assume the error is quickly evaluated and can be re-assigned - its better to enforce that it can only be used for that context. 

So I think @nfuden and I are agreeing that in a multi-error space they should not share names but I'm coming at it from a different angle? This is something I would never want in Java but maybe in go its just how its done.

nfuden: Since the scope of err is small this is safe. Here is my current rule of thumb. 
If err is not evaluated and used in the next 2 lines then it probably should have a different name.

Looking at this further the most idiomatic way if you did not have a overzealous linter would be to write this as _ = ioutil.WriteFile(outputPath, []byte(bootstrapConfig), 0444)

You would do this to call out that yes you understand there is a return and you are indicating that you are intentionally ignoring it.
 
 In otherwords the usage of err in this context is idiomatic

ianmacclancy: Fair enough! Resolving

ianmacclancy (approved):

projects/envoyinit/README.md
```diff
@@ -0,0 +1,114 @@
...
+### Dynamic
+
+Envoy receives dynamic configuration via the [xDS protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#xds-protocol). The Gloo [xDS](https://github.com/solo-io/gloo/tree/master/projects/gloo/pkg/xds) package contains relevant code for serving dynamic configuration.
+
+## Debug
+
+It can be useful to run the Envoy proxy, without the control-plane, as a way of validating proxy behavior. [hack/envoy.yaml](./hack/envoy.yaml) provides example bootstrap that can be used.
+
+`
+docker run \
```
jbohanon, line 63: The formatting here still looks weird. I think it would be better to use the ``` style code block or remove the trailing \ characters. The single ` style code block is not preserving the newlines.

projects/envoyinit/cmd/main.go
```diff
@@ -1,59 +1,102 @@
...
 	if err != nil {
 		log.Fatalf("initializer failed: %v", err)
 	}
 
-	// best effort - write to a file for debug purposes.
+	// 2. Write to a file for debug purposes
+	// since this operation is meant only for debug purposes, we ignore the error
 	// this might fail if root fs is read only
-	writeConfig(outCfg)
+	_ = ioutil.WriteFile(outputPath, []byte(bootstrapConfig), 0444)
```
jbohanon, line 48: nit: use `os` package in lieu of outmoded `io/ioutil` package ([see docs here](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#xds-protocol))

sam-heilbron: good call!

projects/gloo/pkg/xds/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## xDS Server
+
+Gloo Edge is an xDS server. It maintains a snapshot-based, in-memory cache and responds to xDS requests with the resources that are requested.
+
+The following discovery services are supported by Gloo Edge:
+
+### ListenerDiscoveryService
+
+The [ListenerDiscoveryService](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/dynamic_configuration#lds) allows Envoy to discovery Listeners at runtime.
```
jbohanon, line 15: nit: discover, not discovery

Same on lines 19, 23, 27, 31

projects/gloo/pkg/xds/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+The [AggregatedDiscoveryService](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/dynamic_configuration#aggregated-xds-ads) allows Envoy to discovery all resource types over a single stream at runtime.
+
+### SoloDiscoveryService
+
+The [SoloDiscoveryService](https://github.com/solo-io/solo-kit/blob/97bd7c2c67420a6d99bb96f220f2e1a04c6d8a0d/pkg/api/xds/solo-discovery-service.pb.go#L194) is a custom xDS service, used to serve resources of Any type.
+
+In addition to serving configuration for Envoy resources, the Gloo xDS server is also responsible for serving configuration to a number of enterprise extensions (ie `ext-auth` and `rate-limit`)
+
+The SoloDiscoveryService is required to serve these extension resources. It is largely based on the Envoy v2 API, and since it is purely an internal API, we do not need to upgrade the API as the Envoy xDS API. [This issue](https://github.com/solo-io/gloo/issues/4369) contains additional context around the reason behind this custom discovery service.
```
jbohanon, line 39: wording: we do not need to upgrade the API as the Envoy xDS API _changes_ ? or maybe we do not need to upgrade the API _with_ the Envoy xDS API ?

projects/gloo/pkg/xds/README.md
sam-heilbron: 🤦 good catch!

projects/gloo/pkg/xds/README.md
sam-heilbron: ah yeah good call here, I'll update

projects/envoyinit/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Envoy exposes an [administration interface](https://www.envoyproxy.io/docs/envoy/latest/operations/admin) which can be used to query and modify different aspects of the server. The address of this interface is defined in the [bootstrap API](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-msg-config-bootstrap-v3-admin), though it is commonly found at port `19000`. 
+If the above command succeeded, you should be able to visit [port 19000 in your browser](http://localhost:19000/) to explore the admin interface.
+
+## Useful Information
+
+### Other resources
+
+To discover further information about Envoy, there are a number of great sources:
+ - [Hoot YouTube series](https://www.youtube.com/watch?v=KsO4pw4tEGA)
```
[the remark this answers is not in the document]
sam-heilbron: yeah, great call! I ended up just linking the top level repo

projects/gloo/pkg/xds/cache.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/solo-kit/pkg/api/v1/control-plane/cache"
+)
+
+// FallbackNodeCacheKey is used to let nodes know they have a bad config
+// we assign a "fix me" snapshot for bad nodes
+const FallbackNodeCacheKey = "misconfigured-node"
+
+var (
+	// Compile-time assertion
```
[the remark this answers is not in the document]
sam-heilbron: We can do this, I had largely been following the pattern use elsewhere in our code / I had seen in the go control plane

projects/gloo/pkg/xds/envoy.go
```diff
@@ -1,76 +1,18 @@
...
 	solo_xds "github.com/solo-io/solo-kit/pkg/api/xds"
 	"google.golang.org/grpc"
 )
 
-// Returns the node.metadata.role from the envoy bootstrap config
-// if not found, it returns a key for the Fallback snapshot
-// which alerts the user their Envoy is missing the required role key.
-type ProxyKeyHasher struct{}
-
-func NewNodeHasher() *ProxyKeyHasher {
```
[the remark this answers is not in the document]
sam-heilbron: I'm not sure it's necessary, though happy to regroup on this. I think if we assume that Gloo is imported like a library, and modifying exported resources is a breaking change, then we would need to establish that pattern as a team and be consistent with it. I'm not sure we need to do that though since this is more of a standalone project, than a library that is intended to be imported.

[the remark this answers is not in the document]
nfuden: Yeah the overhead is a good callout. We should probably mainly worry about proto changes rather than the library imports.

nfuden:
I am happy with this as it currently is

jbohanon (approved):
LGTM! Thanks for making this stuff more clear!

projects/envoyinit/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+The running instance of Envoy. In Gloo Edge, this is commonly referred to as the `gateway-proxy` component.
+
+## Background
+
+The [Envoy Proxy](https://www.envoyproxy.io/) is a cloud-native, high-performance, service proxy. 
+
+### Source Code
+
+The Gloo Edge service proxies provide all the functionality of the [open source Envoy Proxy](https://github.com/solo-io/envoy-gloo), in addition to some custom extensions. The source code for these proxies is maintained at [envoy-gloo](https://github.com/solo-io/envoy-gloo)
```
jenshu, line 11: both links point to the envoy-gloo repo. did you mean to link to the upstream envoy repo for the first one?

PR #6487, opened 2022-05-20
Remove VALIDATION_MUST_START

elcasteel:
# Description
We were copying the value of `gateway.validation.enabled` into an environment variable, and checking it in the event that no other validation configuration was provided in the settings custom resource. However, when we create the settings custom resource, we never leave the validation configuration empty if validation is enabled so this code was not reachable. 

I removed the reference to the environment variable and also added a helm test for the case where validation is not enabled in the gloo deployment (which is only partially relevant here but I noticed the lack of a test while I was trying to decide what to do with `VALIDATION_MUST_START`)

# Checklist:

### DIFF
--- a/projects/gloo/pkg/syncer/setup/setup_syncer.go
+++ b/projects/gloo/pkg/syncer/setup/setup_syncer.go
@@ -1105,11 +1105,6 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
 		if validation.ValidatingWebhookKeyPath == "" {
 			validation.ValidatingWebhookKeyPath = gwdefaults.ValidationWebhookTlsKeyPath
 		}
-	} else {
-		if validationMustStart := os.Getenv("VALIDATION_MUST_START"); validationMustStart != "" && validationMustStart != "false" {
-			return bootstrap.Opts{}, errors.Errorf("VALIDATION_MUST_START was set to true, but no validation configuration was provided in the settings. "+
-				"Ensure the v1.Settings %v contains the spec.gateway.validation config", settings.GetMetadata().Ref())
-		}
 	}
 	readGatewaysFromAllNamespaces := settings.GetGateway().GetReadGatewaysFromAllNamespaces()
 	return bootstrap.Opts{
--- a/changelog/v1.12.0-beta16/validation_envvar.yaml
+++ b/changelog/v1.12.0-beta16/validation_envvar.yaml
@@ -1,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/6421
+    resolvesIssue: true
+    description: >
+      Remove the reference to the VALIDATION_MUST_START environment variable from gloo setup.
--- a/install/test/helm_test.go
+++ b/install/test/helm_test.go
@@ -4019,7 +4019,31 @@ metadata:
 						})
 						testManifest.ExpectDeploymentAppsV1(glooDeployment)
 					})
-
+					It("can disable validation", func() {
+						glooDeployment.Spec.Template.Spec.Volumes = []v1.Volume{{
+							Name: "labels-volume",
+							VolumeSource: v1.VolumeSource{
+								DownwardAPI: &v1.DownwardAPIVolumeSource{
+									Items: []v1.DownwardAPIVolumeFile{{
+										Path: "labels",
+										FieldRef: &v1.ObjectFieldSelector{
+											FieldPath: "metadata.labels",
+										},
+									}},
+								},
+							},
+						},
+						}
+						glooDeployment.Spec.Template.Spec.Containers[0].VolumeMounts = []v1.VolumeMount{
+							{Name: "labels-volume",
+								MountPath: "/etc/gloo",
+								ReadOnly:  true,
+							}}
+						prepareMakefile(namespace, helmValues{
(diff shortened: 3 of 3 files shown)


projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1105,11 +1105,6 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
 		if validation.ValidatingWebhookKeyPath == "" {
 			validation.ValidatingWebhookKeyPath = gwdefaults.ValidationWebhookTlsKeyPath
 		}
-	} else {
-		if validationMustStart := os.Getenv("VALIDATION_MUST_START"); validationMustStart != "" && validationMustStart != "false" {
```
jenshu, line 1109: is `VALIDATION_MUST_START` used anywhere now? can we remove it from the gateway-deployment helm template too?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1105,11 +1105,6 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
 		if validation.ValidatingWebhookKeyPath == "" {
 			validation.ValidatingWebhookKeyPath = gwdefaults.ValidationWebhookTlsKeyPath
 		}
-	} else {
```
jenshu, line 1108: can we still get here if the user manually removes the validation config from the settings resource?

projects/gloo/pkg/syncer/setup/setup_syncer.go
elcasteel: I have an issue open to completely remove the gateway-deployment helm template: https://github.com/solo-io/solo-projects/issues/3667

projects/gloo/pkg/syncer/setup/setup_syncer.go
elcasteel: So at first I thought that removing the validation config should be a valid way to disable validation but on thinking again, if k8s had the webhook configured on install then that's potentially an issue. I'm going to test that out and possibly update this again.

elcasteel:
/kick going to reopen the access log flake

settings.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: v1
```
jenshu, line 1: delete?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1106,9 +1106,12 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
...
 		}
 	} else {
-		if validationMustStart := os.Getenv("VALIDATION_MUST_START"); validationMustStart != "" && validationMustStart != "false" {
-			return bootstrap.Opts{}, errors.Errorf("VALIDATION_MUST_START was set to true, but no validation configuration was provided in the settings. "+
-				"Ensure the v1.Settings %v contains the spec.gateway.validation config", settings.GetMetadata().Ref())
+		// This will stop users from setting failurePolicy=fail and then removing the webhook configuration
+		if validationMustStart := os.Getenv("VALIDATION_MUST_START"); validationMustStart != "" && validationMustStart != "false" && gatewayMode {
+			return bootstrap.Opts{}, errors.Errorf("A validation webhook was configured, but no validation configuration was provided in the settings. "+
+				"Ensure the v1.Settings %v contains the spec.gateway.validation config."+
+				"If you have removed the webhook configuration since installing, set the envirnment variable VALIDATION_MUST_START=false",
```
jenshu, line 1113:
```suggestion
				"Ensure the v1.Settings %v contains the spec.gateway.validation config. "+
				"If you have removed the validation configuration since installing, set the environment variable VALIDATION_MUST_START=false",
```

elcasteel: pushed a slight variation of this to clarify that the webhook config needs to be removed from k8s.

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1106,9 +1106,13 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
 			validation.ValidatingWebhookKeyPath = gwdefaults.ValidationWebhookTlsKeyPath
 		}
 	} else {
-		if validationMustStart := os.Getenv("VALIDATION_MUST_START"); validationMustStart != "" && validationMustStart != "false" {
-			return bootstrap.Opts{}, errors.Errorf("VALIDATION_MUST_START was set to true, but no validation configuration was provided in the settings. "+
-				"Ensure the v1.Settings %v contains the spec.gateway.validation config", settings.GetMetadata().Ref())
+		// This will stop users from setting failurePolicy=fail and then removing the webhook configuration
```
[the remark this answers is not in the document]
elcasteel: as this is written we just end with gloo failing loudly as soon as the settings change rather than confusingly the next time the admission webhook is hit. It won't be possible to add new resources until the config is fixed.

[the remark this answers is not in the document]
jenshu: ah, gotcha. that sounds non-ideal (i.e. we should be able to default the validation settings to something, even if it's missing from the Settings CR), but as this was existing behavior, i think it's ok to leave it for now

jenshu (approved):

Issue #6295, opened 2022-04-11
Helm: Support Installing gateway-proxy in separate namespace from Gloo Edge Control Plane


murphye:
### Is your feature request related to a problem? Please describe.

Not able to isolate gateway-proxy into a different namespace from the Gloo Edge control plane.

### Describe the solution you'd like

Gloo Mesh Workspaces is configured by namespaces. So, if there are multiple Gloo Edge `gateway-proxy` installs with a single control plane, they need to be in different namespaces to support Gloo Mesh Workspace isolation when used with the Istio integration.

This goes a step beyond https://docs.solo.io/gloo-edge/latest/installation/advanced_configuration/multi-gw-deployment/ by allowing the `gateway-proxy` namespace to be configurable per proxy.

The `gateway-proxy` namespace would likely need to be configured in Helm under `gloo.gatewayProxies.NAME.gatewaySettings`

See the diagram attached to this issue.

### Describe alternatives you've considered

Temporarily we can use this guide for multiple control planes in different namespaces, but this is not ideal. We want a single Gloo Edge control plane.
https://docs.solo.io/gloo-edge/latest/installation/advanced_configuration/multiple-gloo-installs/

### Additional Context

This is needed for a customer install for multi-gateway Edge + Istio + Portal + Gloo Mesh.

I am hoping the implementation would be a fairly straightforward Helm chart change on where to place the gateway-proxy in the namespaces.

murphye:
![Gloo Edge Gateway Namespaces](https://user-images.githubusercontent.com/643097/162801326-b93c7580-a887-40d7-a7e4-b58a602ad789.png)

PR #6505, opened 2022-05-26
Install gateway-proxy in separate namespace from Gloo Edge Control Plane

mkazin:
# Description

Allows setting the namespace of amy named proxy (gatewayProxies.NAME) using a new 

Draft notes:
* This change does not create any of the specified non-control-plane namespaces. These must  be created manually.

# Context

This change allows Gloo Edge to be deployed in a more customized configuration. This is required when integrating proxies into Gloo Mesh Workspaces which use namespaces for isolation.
 
# Testing
Testing was done manually, verifying that XDS configuration reached proxies in both gloo-system and a custom  namespace.

Configuration used in testing (based on the [multi-gw deployment guide](https://github.com/solo-io/gloo/blob/master/docs/content/installation/advanced_configuration/multi-gw-deployment.md))
```
  gatewayProxies:
    publicGw: # Proxy name for public access (Internet facing)
      disabled: false # overwrite the "default" value in the merge step
      kind:
        deployment:
          replicas: 2
      service:
        kubeResourceOverride: # workaround for https://github.com/solo-io/gloo/issues/5297
          spec:
            ports:
              - port: 443
                protocol: TCP
                name: https
                targetPort: 8443
            type: LoadBalancer
      gatewaySettings:
        customHttpsGateway: # using the default HTTPS Gateway
          virtualServiceSelector:
            gateway-type: public # label set on the VirtualService
        disableHttpGateway: true # disable the default HTTP Gateway
    corpGw: # Proxy name for private access (intranet facing)
      disabled: false # overwrite the "default" value in the merge step
      namespace: corp-ns
      service:
        httpPort: 80
        httpsFirst: false
        httpsPort: 443
        httpNodePort: 32080 # random port to be fixed in your private network
        type: NodePort
      # Creates corp-gw-config-dump-service
      readConfig: true
      readConfigMulticluster: true
      gatewaySettings:
        customHttpGateway: {} # using the default HTTP Gateway
        disableHttpsGateway: true # disable the default HTTPS Gateway
    gatewayProxy:
      disabled: false # disable the default gateway-proxy deployment and its 2 default Gateway CRs
```

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6295

### DIFF
--- a/install/test/fixtures.go
+++ b/install/test/fixtures.go
@@ -17,7 +17,7 @@ node:
   id: "{{.PodName}}.{{.PodNamespace}}"
   metadata:
     # role's value is the key for the in-memory xds cache (projects/gloo/pkg/xds/envoy.go)
-    role: "{{.PodNamespace}}~gateway-proxy"
+    role: "gloo-system~gateway-proxy"
 static_resources:
   listeners: # if or $statsConfig.enabled (or $spec.readConfig $spec.extraListenersHelper) # $spec.extraListenersHelper
   - name: prometheus_listener
@@ -193,7 +193,7 @@ node:
   cluster: gateway
   id: '{{.PodName}}.{{.PodNamespace}}'
   metadata:
-    role: '{{.PodNamespace}}~gateway-proxy'
+    role: 'gloo-system~gateway-proxy'
 static_resources:
   clusters:
   - alt_stat_name: xds_cluster
@@ -331,7 +331,7 @@ node:
   cluster: gateway
   id: '{{.PodName}}.{{.PodNamespace}}'
   metadata:
-    role: '{{.PodNamespace}}~gateway-proxy'
+    role: 'gloo-system~gateway-proxy'
 static_resources:
   clusters:
   - alt_stat_name: xds_cluster
@@ -483,7 +483,7 @@ node:
   cluster: gateway
   id: '{{.PodName}}.{{.PodNamespace}}'
   metadata:
-    role: '{{.PodNamespace}}~gateway-proxy'
+    role: 'gloo-system~gateway-proxy'
 static_resources:
   clusters:
   - alt_stat_name: xds_cluster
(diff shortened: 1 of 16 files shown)


install/test/fixtures.go
```diff
@@ -17,7 +17,7 @@ node:
   id: "{{.PodName}}.{{.PodNamespace}}"
   metadata:
     # role's value is the key for the in-memory xds cache (projects/gloo/pkg/xds/envoy.go)
-    role: "{{.PodNamespace}}~gateway-proxy"
+    role: "gloo-system~gateway-proxy"
```
inFocus7, line 20: Probably a dumb q, but why are we hardcoding `.PodNamespace` to `gloo-system` now? (extends to this and all other areas where `.PodNamespace` was replaced)

And should we update the `id` a few lines above to match this change as well?

nfuden, line 20: is this change necessary?

nfuden (commented):
We should have a test to prove that this outputs correctly for 2 proxies in different namespaces

inFocus7:
Also agree with Nathan on testing that different proxies w/ different namespaces works! 💯

install/test/fixtures.go
mkazin: Those are good questions!

Q1:
1. This is a test fixture, so it's not so much hard-coding a value, but specifying the expected value.
2. .PodNamespace is a special value which is not translated by Helm, but rather by Envoy => we're looking at the expected Envoy config here.
3. This *was* the expected value when the namespace of gateway-proxies was identical to that of the pod they ran in, which also matched the namespace of the gloo pod.
4. That's no longer the case, as this PR allows for the deployment of a gateway-proxy in other namespaces.
5. However, we don't want to keep using the pod's namespace, as this represents the key of the xds cache under which the proxy's config is stored
6. To avoid confusion and possible upgrade issues we decided not to change the xds keys, and leaving them with whatever was used for the writeNamespace (gloo's namespace). In these tests, that's "gloo-system".

Q2: no, I believe the id is correct, as it identifies the envoy node in its installed namespace.

Nadine2016 (approved):

install/test/fixtures.go
mkazin: I think so. Please see [this response](https://github.com/solo-io/gloo/pull/6505#discussion_r883906230) I left to Fabian's similar question.

nfuden (approved):

PR #6509, opened 2022-05-27
Status field with supergraph for gateway schemas

Bslabe123:
# Description

Added a status field named "Messages" on which custom resources can set messages per resource ref that get set on the status. Messages now includes the computed supergraph of all subgraphs in a gateway schema CR.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/3338


projects/gloo/pkg/plugins/plugin_interface.go
```diff
@@ -28,6 +30,7 @@ type Plugin interface {
 type Params struct {
 	Ctx      context.Context
 	Snapshot *v1snap.ApiSnapshot
+	Messages *map[*core.ResourceRef][]string
```
kdorosh, line 33: maps are reference types, they are always passed by reference https://go.dev/blog/maps

we shouldn't need the extra pointers

Bslabe123: I thought so too but for some reason If i pass in the map and not a pointer to the map, whenver I mutate the map all the changes are lost upon exiting that function, would that mean that this is correct then or is it a bug with my logic, I can point to where I use it 
`	
ref := params.VirtualHostParams.Params.Messages
	if ref == nil {
		return eris.Wrapf(err, "params.virtualHostParams.Params.Messages cannot be written because it was never initialized")
	}
	(*ref)[gqlRef] = []string{routeConf.GetExecutableSchema().GetSchemaDefinition().String()}`

Bslabe123: HOW did i miss that, fixed!

projects/gloo/pkg/translator/translator.go
```diff
@@ -149,6 +149,15 @@ func (t *translatorInstance) Translate(
 		}
 	}
 
+	messagesMap := params.Messages
+	if messagesMap != nil && len(messagesMap) > 0 {
```
kdorosh, line 153: nit, style; `len(nil)` is zero, no need for the nil check first

kdorosh: messagesMap is only used twice, perhaps get rid of it and just use `params.Messages`

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -125,6 +127,7 @@ func (s *translatorSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapsh
 		params := plugins.Params{
 			Ctx:      proxyCtx,
 			Snapshot: snap,
+			Messages: map[*core.ResourceRef][]string{},
```
kdorosh, line 130: does this really need to be a pointer to a `core.ResourseRef`?

kdorosh: probably slightly easier to reason about without the pointer

PR #6542, opened 2022-06-08
projects: StagedFilters now require configs to be set

nfuden:
Blocker for envoy 1.22 where using the name of a config is insufficient but rather needs the typedurl

### DIFF
--- a/changelog/v1.12.0-beta19/chart-version copy.yaml
+++ b/changelog/v1.12.0-beta19/chart-version copy.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/envoy-gloo-ee/issues/260
+    resolvesIssue: false
+    description: Breaking change from envoy 1.22 typedurl needed in all filters
--- a/projects/gloo/pkg/plugins/grpcweb/plugin.go
+++ b/projects/gloo/pkg/plugins/grpcweb/plugin.go
@@ -60,7 +60,6 @@ func (p *plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) (
 	if p.isDisabled(listener) {
 		return nil, nil
 	}
-	filter, _ := plugins.NewStagedFilterWithConfig(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
 
-	return []plugins.StagedHttpFilter{filter}, nil
+	return []plugins.StagedHttpFilter{plugins.MustNewStagedFilter(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)}, nil
 }
--- a/projects/gloo/pkg/plugins/grpcweb/plugin.go
+++ b/projects/gloo/pkg/plugins/grpcweb/plugin.go
@@ -60,7 +60,7 @@ func (p *plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) (
 	if p.isDisabled(listener) {
 		return nil, nil
 	}
-	filter, _ := plugins.NewStagedFilter(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
+	filter, _ := plugins.NewStagedFilterWithConfig(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
 
 	return []plugins.StagedHttpFilter{filter}, nil
 }
--- a/projects/gloo/pkg/plugins/grpcweb/plugin.go
+++ b/projects/gloo/pkg/plugins/grpcweb/plugin.go
@@ -60,7 +60,7 @@ func (p *plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) (
 	if p.isDisabled(listener) {
 		return nil, nil
 	}
-	filter, _ := plugins.NewStagedFilterWithConfig(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
+	filter, _ := plugins.NewStagedFilter(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
 
 	return []plugins.StagedHttpFilter{filter}, nil
 }
--- a/projects/gloo/pkg/plugins/grpcweb/plugin.go
+++ b/projects/gloo/pkg/plugins/grpcweb/plugin.go
@@ -1,7 +1,9 @@
 package grpcweb
 
 import (
+	envoygrpcweb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/grpc_web/v3"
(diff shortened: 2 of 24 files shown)


changelog/v1.12.0-beta19/chart-version copy.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: nit: is this a new feature? It feels more non-user facing given that there is no impact on the user facing API

projects/gloo/pkg/plugins/staged_filter_util.go
```diff
@@ -1,17 +1,21 @@
...
-	return s
+// NewStagedFilterWithConfig is deprecated as config is now always needed
+// See the new signature of NewStagedFilter
+func NewStagedFilterWithConfig(name string, config proto.Message, stage FilterStage) (StagedHttpFilter, error) {
+	return NewStagedFilter(name, config, stage)
 }
 
-func NewStagedFilterWithConfig(name string, config proto.Message, stage FilterStage) (StagedHttpFilter, error) {
+// NewStagedFilter creates an instance of the named filter with the desired stage
+func NewStagedFilter(name string, config proto.Message, stage FilterStage) (StagedHttpFilter, error) {
```
sam-heilbron, line 18: These two functions have the same signatures and different names. Do we still need both? I could see value in deprecating one if it were possible, but we can no longer support nil config, so it seems that we're safe to just remove one. What do you think?

projects/gloo/pkg/plugins/staged_filter_util.go
```diff
@@ -20,22 +24,26 @@ func NewStagedFilterWithConfig(name string, config proto.Message, stage FilterSt
 		Stage: stage,
 	}
 
-	if config != nil {
+	if config == nil {
+		return s, errors.New("filters must have a config specified to derive its type")
```
sam-heilbron, line 28: 👏

projects/gloo/pkg/plugins/staged_filter_util.go
```diff
@@ -20,22 +24,26 @@ func NewStagedFilterWithConfig(name string, config proto.Message, stage FilterSt
...
+	}
 
-		marshalledConf, err := utils.MessageToAny(config)
-		if err != nil {
-			// this should NEVER HAPPEN!
-			return StagedHttpFilter{}, err
-		}
+	marshalledConf, err := utils.MessageToAny(config)
+	if err != nil {
+		// all config types should already be known
```
sam-heilbron, line 33: love the extra context!

projects/gloo/pkg/plugins/grpcweb/plugin.go
```diff
@@ -52,11 +54,13 @@ func (p *plugin) isDisabled(httplistener *v1.HttpListener) bool {
...
+// HttpFilters will add an empty version of the grpcweb filter to a listener
+// if it is not explicitly set.
 func (p *plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) ([]plugins.StagedHttpFilter, error) {
 	if p.isDisabled(listener) {
 		return nil, nil
 	}
-	return []plugins.StagedHttpFilter{
-		plugins.NewStagedFilter(wellknown.GRPCWeb, pluginStage),
-	}, nil
+	filter, _ := plugins.NewStagedFilterWithConfig(wellknown.GRPCWeb, &envoygrpcweb.GrpcWeb{}, pluginStage)
```
sam-heilbron, line 63: I know we historically would drop the error for nil config, and do that inside of a utility function. Now that its in the plugin, I think a comment outlining why this is safe (or why an error would occur in the first place). Alternatively, we could just return the error as well, even if we are assuming it will be nil almost every time.

projects/gloo/pkg/plugins/staged_filter_util.go
nfuden: Its a good point. I struggled with this. 
I want the long term name to be NewStagedFilter but that would have made this pull much larger.
I intend that once this is updated then switch over solo-projects' calls to only use the nicer name and then we can update gloo as well.

changelog/v1.12.0-beta19/chart-version copy.yaml
nfuden: Idk its breaking for plugin implementers right?

nfuden: Actually thats just for library users in retrospect we are good.

projects/gloo/pkg/plugins/grpcweb/plugin.go
nfuden: Yeah so we do return the error always in the WithConfig variant but didnt do it in the version without config.
I updated the comment on NewStagedFilter to reflect the error reasons.

I like the return of the error as the _ shows that we know and are asserting that it should never fail but we dont need to have a different function. If we want to mute the error then we could wrap this with an error swallowing util like
MustNewStagedFilter but in general having a single function to call seems cleaner in general.
That being said I am happy to implement it if there is still a desire.

projects/gloo/pkg/plugins/staged_filter_util.go
```diff
@@ -1,17 +1,23 @@
...
+
 	envoyhttp "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	"github.com/golang/protobuf/proto"
 	"github.com/solo-io/gloo/projects/gloo/pkg/utils"
 )
 
-func NewStagedFilter(name string, stage FilterStage) StagedHttpFilter {
-	s, _ := NewStagedFilterWithConfig(name, nil, stage)
-	return s
+// NewStagedFilterWithConfig is deprecated as config is now always needed
```
saiskee, line 11: [nit] Can we mark this as deprecated following GoDoc conventions? 

```suggestion
// Deprecated: config is now always needed
```

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -198,7 +198,9 @@ func (p *Plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) (
 		}
 		filters = append(filters, earlyFilter)
 	}
-	filters = append(filters, plugins.NewStagedFilter(FilterName, pluginStage))
+
+	emptyFilter, _ := plugins.NewStagedFilterWithConfig(FilterName, &envoytransformation.FilterTransformations{}, pluginStage)
```
saiskee, line 202: Since `NewStagedFilterWithConfig` is deprecated, can we use the non-deprecated method here.

saiskee (approved):
LGTM 🖖

nfuden:
/kick Access Log in memory Access Logs Grpc [It] can stream access logs 
Step #10 - "test": /workspace/gloo/test/e2e/access_log_test.go:173
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=c388891b-4220-41b0-b473-5d3a5bc6fa8f

PR #6558, opened 2022-06-13
Isolate Secure VirtualHosts On Separate FilterChains

sam-heilbron:
# Description

# Remaining Work
## Before Merge
- [x] Merge and release changes to schemagen library (https://github.com/solo-io/protoc-gen-openapi/pull/4)
- [x] Merge and release changes to solo-kit (https://github.com/solo-io/solo-kit/pull/482)

## After Merge
- [ ] Kube e2e test
- [ ] Add support for HybridGateway. This will likely leverage a lot of code re-use, but we don't need to add it to this first PR, which is already huge. I will not backport this feature until the full change set has been merged.
- [ ] Some of our documentation is affected by this. Sync with the docs team to make a plan around this (I'm happy to do it in the follow up)

# Solution

This is a very large PR. Below I attempt to provide an overview of the changes, broken down per component, in an order that is straightforward to follow.

We break this set of changes down into the following groups:
1. API Changes
1. Gloo Translation
1. Gateway Translation
1. End to End Testing
1. Additional Changes

### Settings API
This new behavior is gated by the the `IsolateVirtualHostsBySslConfig` field that is now available on the Settings API. Since this affects Gateway translation in particular the field is:
```
settings:
  gateway:
    isolateVirtualHostsBySslConfig
```

**This field defaults to false (or undefined)**, meaning that we will maintain the old behavior and not isolate virtual hosts. Instead, we will group all virtual hosts, and then duplicate filter chains based on the ssl config.

If you set this value to true, then you will opt into the new translation pattern. 

This is the only user facing aspect of the functionality. All behavioral changes affect the internal translation engine, and the produced Envoy configuration.

### Gateway Annotation
The Settings API allows configuring the behavior globally for a Gloo installation. However, you may want to opt-in (or out) of this behavior per-gateway, allowing you to progressively introduce this behavior into your system. We expose a new Gateway annotation with the power to override the global setting.

`gateway.solo.io/isolate_vhost` can be set to a truthy value to override the behavior for a Gateway. If the annotation is defined and valid, it will override the Gateway. If the annotation is invalid or undefined, we will respect the global behavior.

### Proxy API (*internal api*)
We currently have 3 types of Listeners on the Proxy: HttpListener, TcpListener, HybridListener. As part of this work, we add a new type of Listener, the AggregateListener. This is required, since the decision of how to group VirtualHosts must exist during the Gateway translation (and be persisted on the Proxy). **Additionally, by introducing a new Listener type, we ensure that distinct code paths are used for this functionality, to reduce the scope of changes.**

## 2. Gloo Changes
*The Gloo component is responsible for converting a Proxy into appropriate Envoy resources*

### ListenerSubsystemTranslatorFactory
To completely isolate the new functionality from the old functionality, use a new type of translator to convert the `AggregateListener` into a set of Envoy Listeners and RouteConfigurations. Since we already have small components that build these resources, most of the code was just instantiating objects properly.

I added high level tests for these translators as well. The intention was to validate the number of resources (how many filter chains within the listener? how many distinct route configurations) and the reports that were generated during translation. 

With these changes alone, we can convert an AggregateListener into the appropriate Envoy configuration as well as report errors on it accurately. We do not yet have the code to convert Gateway resources into the AggregateListener, but the explanation of those components is below.

### Testing
Added tests (`listener_subsystem_test.go`) to validate that provided a Proxy, we generate the appropriate number of Envoy resources in case of success, or the appropriate reports (warnings and errors) in the case of failure.

## 3. Gateway Changes
*The Gateway component is responsible for converting Gateways,VirtualServices and RouteTables into a Proxy CR*

### VirtualServiceTranslator
We have a multiple translators which are responsible for converting VirtualServices into VirtualHosts. It made sense to me to split off this functionality into a standalone component. This way it could be shared across each of the translators. I copy/pasted the code that was originally in the `http_translator.go` file and moved it to its own `virtual_service_translator.go`. I did not move the tests as well, since that would require refactoring to use the `TranslateVirtualService` interface. 

### HttpTranslator and HybridTranslator
Neither of these translators were changed. There are some small code changes to support the VirtualServiceTranslator, but I intentionally left these alone. The rationale behind this is that the changes we're making are fairly widespread, and it would be safer to ensure that all existing functionality is unchanged.

### AggregateTranslator
This is the new translator that is responsible for converting Gateway CRs into the new AggregateListener type.

There will be more unit tests that can be added to this component, but the envoy e2e tests validate most of the functionality already.

### Gateway Translator
The GatewayTranslator converts Gateways into Listeners, and aggregates those listeners on a single Proxy.

This translator contains a mapping of sub-translators, responsible for converting a Gateway -> Listener. 

Based on the global setting (`IsolateVirtualHostsBySslConfig`) and the Gateway annotations (`gateway.solo.io/isolate_vhost`) we pick the appropriate translator to convert the Gateway into a Listener.

### Testing
There are tests to validate that provided the combination of settings and annotations, the appropriate translator is chosen for each of the Gateway types. This confirms that all existing behavior will remain by default, and the new type of translation will only be executed if users opt into it.

### Envoy
I added some envoy e2e tests in a new `aggregate_listener_test.go` file. These tests should clarify the problem statement, as well as provide concrete examples to distinguish the old behavior from the new behavior. Without understanding how it works, these tests should confirm that the proposed changes do work.

### Kubernetes
This is more of a smoke test than anything. I wanted to validate that we witness 1 successful run, and 1 failed run in a Kubernetes cluster. 

## 5. Test Utility Code
*I made some enhancements to our testing utility code*

### VirtualServiceBuilder
We have a lot of functions that accept many properties, and then we tend to have variations of those functions, with slightly different parameters. This to me, is an example of the [telescoping constructor anti-pattern](https://codethataint.com/blog/builder-pattern/) which can be solved using the [builder pattern](https://refactoring.guru/design-patterns/builder). Since VirtualServices are frequently modified (and I was using them in a test) I decided to try to create a Builder for this object. This allows the tests that build these resources to be more explicit with the pieces, and the large object creation is abstracted from the test.

### TestUpstream
Similar to how we created VirtualServices, when executing requests against an Upstream, we tend to pass many optional parameters. I decided to create an aggregate type to represent the RequestOptions and the ResponseOptions. This way, we can slowly move towards this function signature, and clients can just provide the options that they care about.

### TestClients
TestClients are the set of resource clients used by tests to read/write resources from the cache. To avoid test pollution, I followed a new pattern, where at the beginning of each test I write a snapshot of resources, and after it I delete the snapshot. Each test within a suite may configure which resources are on the snapshot, but this ensures there is only one place that the resources are actually written to the cache.

To make this usable for other tests, I expose a 'WriteSnapshot' and 'DeleteSnapshot' method on the TestClients struct.

# Checklist:


test/e2e/aggregate_listener_test.go
```diff
@@ -0,0 +1,390 @@
...
+			It("produces a Proxy with a single HttpListener", func() {
+				proxy, err := testClients.ProxyClient.Read(defaults.GlooSystem, gatewaydefaults.GatewayProxyName, clients.ReadOpts{})
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(proxy.GetListeners()).To(HaveLen(1))
+				Expect(proxy.GetListeners()[0].GetHttpListener()).NotTo(BeNil())
+			})
+
+			It("routes requests to all routes on gateway", func() {
+				// This test demonstrates the flaw with HttpListeners:
```
ben-taussig-solo, line 352: Love the clarifying comment

projects/gloo/api/v1/settings.proto
```diff
@@ -683,6 +683,11 @@ message GatewayOptions {
     // This is set based on the install mode. It indicates to gloo whether or not it should run the gateway
     // translations and validation.
     google.protobuf.BoolValue enable_gateway_controller = 9;
+
+    // If set, group virtual hosts by matching ssl config, and isolate them on separate filter chains
+    // The default behavior is to aggregate all virtual hosts, and expose them on identical filter chains,
+    // each with a FilterChainMatch that corresponds to the ssl config.
```
ben-taussig-solo, line 689: Does it make sense to clarify in this comment that enabling this setting will leverage the new listener type? Or are we intentionally leaving that opaque to the user?

ben-taussig-solo (commented):
Couple of nits but otherwise this holds up to my first pass -- do you have time today to get on a call to discuss the ListenerSubsystemTranslatorFactory and AggregateTranslator changes?

projects/gateway/pkg/translator/ssl_configuration.go
```diff
@@ -0,0 +1,65 @@
...
+	result := map[*gloov1.SslConfig][]*v1.VirtualService{}
+	mergedSslConfig := map[string]*gloov1.SslConfig{}
+	groupedVirtualServices := map[string][]*v1.VirtualService{}
+
+	for _, virtualService := range virtualServices {
+		sslConfig := virtualService.GetSslConfig()
+		sslConfigHash := hashSslConfig(sslConfig)
+
+		if matchingCfg, ok := mergedSslConfig[sslConfigHash]; ok {
+			// there is an existing sslConfig that differs only by sni domain, update the entry
```
gunnar-solo, line 21: I saw a comment similar to this in the hash function--help me understand why differing by _only_ sni domain is important, here?

projects/gateway/pkg/translator/translator_test.go
```diff
@@ -335,6 +335,122 @@ var _ = Describe("Translator", func() {
...
+										HttpGateway: &v1.HttpGateway{},
+									},
+								},
+							},
+						},
+					},
+					BindPort: 3,
+				}
+
+				snap.Gateways = append(
```
gunnar-solo, line 374: _very plausible_ I'm reading these `BeforeEach`'s wrong:  does this mean that we are continually growing the length of snap.Gateways with each `It(...)`?

test/e2e/aggregate_listener_test.go
```diff
@@ -0,0 +1,390 @@
...
+				})
+		}
+
+		BeforeEach(func() {
+			simpleRouteName := "simple-route"
+			vsBuilder := gloohelpers.NewVirtualServiceBuilder().WithNamespace(defaults.GlooSystem)
+
+			vsEast := vsBuilder.
+				WithName("vs-east").
+				WithDomain("east.com").
```
gunnar-solo, line 179: you are right--I like this syntax a lot (in comparison to telescoping constructors)

projects/gateway/pkg/translator/translator_test.go
sam-heilbron: Luckily, the Snap is initiliazed in an outer context in a BeforeEach, so it's size is constrained. With that said, I reworked the test to use a Table, and tried to make the Snapshot initialization a bit more explicit too.

test/e2e/aggregate_listener_test.go
sam-heilbron: Me too! I'm going to start trying to use this pattern for resources that are constantly built/adjusted in tests

projects/gateway/pkg/translator/aggregate_translator.go
```diff
@@ -0,0 +1,120 @@
...
+	var aggregateListener *gloov1.AggregateListener
+	switch gateway.GetGatewayType().(type) {
+	case *v1.Gateway_HttpGateway:
+		aggregateListener = a.computeAggregateListenerForHttpGateway(params, proxyName, gateway)
+
+	case *v1.Gateway_HybridGateway:
+		aggregateListener = a.computeAggregateListenerForHybridGateway(params, proxyName, gateway)
+	}
+
+	if aggregateListener == nil {
```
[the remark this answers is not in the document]
sam-heilbron: Good call! I moved it to the default case

projects/gloo/api/v1/settings.proto
sam-heilbron: Yeah it's a great question. I opted to leave this opaque to the user, since to me that is more of an internal detail that they don't need to be aware of. Do you think there is some addition that I could make to this comment to clarify the behavior to users?

projects/gateway/pkg/translator/aggregate_translator.go
```diff
@@ -0,0 +1,120 @@
...
+func (a *AggregateTranslator) computeAggregateListenerForHttpGateway(params Params, proxyName string, gateway *v1.Gateway) *gloov1.AggregateListener {
+	var httpFilterChains []*gloov1.AggregateListener_HttpFilterChain
+	availableHttpListenerOptionsByName := make(map[string]*gloov1.HttpListenerOptions)
+	availableVirtualHostsByName := make(map[string]*gloov1.VirtualHost)
+
+	sslGateway := gateway.GetSsl()
+	httpGateway := gateway.GetHttpGateway()
+	virtualServices := getVirtualServicesForHttpGateway(params, gateway, httpGateway, sslGateway)
+
+	if gateway.GetSsl() {
```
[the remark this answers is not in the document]
sam-heilbron: I like it! I was thinking of refactoring this when I add support for HybridGateway's, since that code will require reuse of the above function. Do you mind if I table that restructuring for the follow-up PR that handles that change?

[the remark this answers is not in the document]
gunnar-solo: go for it (next iteration)

projects/gloo/api/v1/settings.proto
ben-taussig-solo: I think that's reasonable, just wanted to check in

projects/gateway/pkg/translator/ssl_configuration.go
sam-heilbron: https://www.cloudflare.com/learning/ssl/what-is-sni/

SNI helps solve the problem of allowing a single endpoint (single IP) to serve multiple certificates, and allowing clients to specify which certificate (the client specifies the domain) they want to verify during the TLS handshake.

If there are 2 SslConfigurations that differ by only domainName, it means we have identical certificates that should be served up for each of the domains. Instead of maintaining distinct configurations, we are safe to merge them and serve the shared cert for all domains that match

projects/gateway/pkg/translator/ssl_configuration.go
```diff
@@ -0,0 +1,65 @@
...
+		} else {
+			// there is no existing sslConfig, create a new entry
+			ptrToCopy := proto.Clone(sslConfig).(*gloov1.SslConfig)
+			mergedSslConfig[sslConfigHash] = ptrToCopy
+			groupedVirtualServices[sslConfigHash] = append(groupedVirtualServices[sslConfigHash], virtualService)
+		}
+	}
+
+	for sslHash, sslConfig := range mergedSslConfig {
+		result[sslConfig] = groupedVirtualServices[sslHash]
```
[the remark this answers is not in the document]
sam-heilbron: It's a good callout. I felt this was safe since the map is short lived and there's no chance of us modifying the contents. The intention is to maintain a mapping of sslConfig and a set of VirtualServices. However, we could achieve that in other ways (say, a map where the ssl hash is the key and the object is the tuple of ssl definition and virtual service list). I'm happy to adjust the code if you think it's worth it. (Also can be done in the follow-up PR)

gunnar-solo (approved):
Make it so.

PR #6566, opened 2022-06-14
Supergraph field on status/Messages on status

Bslabe123:
# Description

Added the ability to add messages to pod statuses, This was done by modifying the struct for reports and statuses. Along side this also added logic for gateway proxies to add the graphQL super-graph schema to messages.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/3338


projects/gloo/pkg/translator/translator.go
```diff
@@ -149,6 +149,13 @@ func (t *translatorInstance) Translate(
 		}
 	}
 
+	messagesMap := params.Messages
+	if len(messagesMap) > 0 {
+		for _, messages := range messagesMap {
+			reports.AddMessages(proxy, messages...)
+		}
+	}
```
kdorosh, line 157: let's also add this logic to the gateway translation loop (other implementers of the translator interface)

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	github.com/solo-io/skv2 v0.21.6
 	// Pinned to the `gloo-namespaced-statuses` tag of solo-apis
 	github.com/solo-io/solo-apis v0.0.0-20210922150112-505473b2e66c
-	github.com/solo-io/solo-kit v0.28.0
+	github.com/solo-io/solo-kit v0.28.1-0.20220615170347-2653ede69b17
```
kdorosh, line 57: let's depend on a released version of solokit!

projects/gateway/pkg/translator/translator.go
```diff
@@ -93,6 +93,13 @@ func (t *GwTranslator) Translate(ctx context.Context, proxyName string, snap *gl
 		if listener != nil {
 			listeners = append(listeners, listener)
 		}
+
+		messagesMap := params.Messages
```
sam-heilbron, line 97: Instead of accumulating these messages during translation, and then adding them to the reports, could we just pass the reports around (like we have historically) and execute AddMessage in the appropriate place? This allows our translator to have fewer dependencies

Bslabe123: That might be cleaner! Would eliminate the need for the past few commits but yes messages can definitely written to reports as is without these "boilerplate"y additions. Changing now.

projects/gloo/pkg/plugins/plugins.go
```diff
@@ -38,6 +40,7 @@ type InitParams struct {
 type Params struct {
 	Ctx      context.Context
 	Snapshot *v1snap.ApiSnapshot
+	Messages map[*core.ResourceRef][]string
```
sam-heilbron, line 43: Why do these Parameters need to include Messages now?

go.mod
```diff
@@ -50,11 +50,11 @@ require (
 	github.com/solo-io/go-utils v0.21.26
 	github.com/solo-io/k8s-utils v0.1.0
 	github.com/solo-io/protoc-gen-ext v0.0.16
-	github.com/solo-io/protoc-gen-openapi v0.0.4
+	github.com/solo-io/protoc-gen-openapi v0.1.0
```
sam-heilbron, line 53: nit: could we also add the dep bumps in the changelog for transparency?

projects/gloo/pkg/translator/translator.go
```diff
@@ -149,6 +149,13 @@ func (t *translatorInstance) Translate(
 		}
 	}
 
+	messagesMap := params.Messages
+	if len(messagesMap) > 0 {
+		for _, messages := range messagesMap {
+			reports.AddMessages(proxy, messages...)
```
sam-heilbron, line 155: This will add the messages to the Proxy (and not necessarily the underlying resource). Is that what is intended?

sam-heilbron: Unless I am missing something, I don't see the code that adds Messages. Can we add a test that should produce a message, and then validate that it appears on the status?

projects/gloo/pkg/plugins/plugins.go
Bslabe123: params.messages is side effected with the messages when we do processRoute in solo-projects https://github.com/solo-io/solo-projects/pull/3768/files

projects/gloo/pkg/translator/translator.go
Bslabe123: Skipped this test because the only message we would write would be the stitched graphql schema which we generate in solo-projects so it has its own e2e test there, https://github.com/solo-io/solo-projects/pull/3768/files

Bslabe123: We were looking at it at one point, We figured the side effect way might be a bit cleaner because if we were to expand the report struct it would be easier to just attach it to params instead of changing the interface which could be messier/harder to implement and maintain.

kdorosh: We figured the side effect was simpler / smaller change (no need to update every single plugin interface e.g. `ProcessRoute`/`ProcessEndpoint`/etc..)

The plugins today already side effect some inputs (see the `in` and `out` pointers to each). While it might be nice to rewrite the plugin framework to be more functional at some point, it did not seem worth the effort to do or start that refactor when we could just do it in GMG. Happy to discuss further in slack @sam-heilbron if you have concerns, definitely not completely sold on this implementation if you feel strongly

Also fwiw I'm not a huge fan of requiring `reports.AddWarning`/`reports.AddError` after each plugin function call, and if we made this more functional then we'd need to either replace all of these with `reports.AddWarningOrErrorAndMessages` or explore an even more involved refactor (one that seems out of scope for this issue)

Bslabe123: I could open a ticket for this refactor if we want to do that separately regardless?

projects/gloo/pkg/plugins/plugins.go
sam-heilbron: Ah I see!  love the comment that clarifies this

sam-heilbron (commented):
Looks great to me!

kdorosh (approved):

Issue #6563, opened 2022-06-14
Fix Route Sanitization and Validation


ianmacclancy:
While working on https://github.com/solo-io/gloo/issues/6098

It became apparent that our current route validation and sanitization do not work as expected. 

This task is to track the work to fix our current code before merging new route errors and validation 

Issues Discovered 
- Code errors  
  - We never process route errors onto the proxy report  - Incorrect Conversion of sanitized errors from the resource report to warnings in the proxy report - the error string used here  https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/validation/server.go#L296 doe not match the string used as a key in the map of warnings
  - Incorrect error output on ProcessingError. Reason: no destination type specified. The formatted error should include the route name and instead just lists a message of "ProcessingError. Reason: no destination type specified. Route Name: " with a missing route name
- Test Errors - Route Errors and Sanitization are Not Tested by the Server_Test.go file
  - https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/validation/server_test.go#L118 - this test, which is supposed to test the after effects of route sanitization, passes an xdsSanitizer that has a RouteReplacementSanitizer that has enabled set to false so no sanitization happens for it to test.

Design of Flow Issues
- We correct the proxy report outside of sanitization when that correction cannot happen without sanitization - rather than reconciling the two reports after sanitization - sanitization could update both 
- We sanitize outside of translation - but might be able to sanitize as part of translation rather than after 
 
Done Criteria
- K2e test with rejected resources 
- Tests correctly test route errors with enabled sanitizers 
- Proxy Report updated after sanitization 
- Design for restructure of code - we update the proxy report after sanitization when it should happen as part of sanitization and sanitization occurs outside translation

ianmacclancy:
Initial work on issue - https://github.com/solo-io/gloo/tree/issue-6563

PR #6567, opened 2022-06-14
Route error when no weights passed on multiaction route weighted Destination 

ianmacclancy:
# Context

This change is made to address - https://github.com/solo-io/gloo/issues/6098 

Summary - customers can pass no weights on weighted destinations and this causes envoy errors.

# Description

Changes
- Validation added to setWeightedClusters method 
  - Weights will default to 0 if not weight is passed and any weight less than 1 is not considered valid so a check was added and we now emit a route error if we hit that case
  - Additional comments were added to make it clear why we are setting total weight based on envoy defaults we are overriding  

These changes will add errors that customers should be able to see but will not reject the resource as we hoped. The reason for this is tracked here - https://github.com/solo-io/gloo/issues/6563

Summary - Validation is not currently working on route errors 

# Checklist:


projects/gloo/pkg/translator/route_config.go
```diff
@@ -429,6 +429,17 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
 			return err
 		}
 
+		//Catch when Customers pass weights less than 1 or forget to pass weights - this leads to envoy errors
+		if weightedDest.GetWeight() < 1 {
```
nfuden, line 433: nit: As far as I can tell route pieces with 0 weight are fine as long as the total weight for the whole multi route is greater than 0.
That being said It does seem odd to allow for routes than might never be routed to (weight 0) so I am fine with passing this as is from a safety in gloo perspective but I would be interested to hear if we did any digging into why one may want to specifiy some sub weights that are at 0 within a multi route

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-weightedcluster-clusterweight

ianmacclancy: https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/proxy.proto.sk/#weighteddestination - our docs specifically say it must be greater than 0.

projects/gloo/pkg/translator/route_config.go
```diff
@@ -429,6 +429,17 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+			weightedDestUpstream := weightedDest.GetDestination().GetUpstream()
+
+			validation.AppendRouteError(routeReport,
+				validationapi.RouteReport_Error_ProcessingError,
+				fmt.Sprintf("Incorrect configuration for Weighted Destination for upstream: %s - Weighted Destinations require a weight that is greater than 1", weightedDestUpstream.Name),
```
jenshu, line 438:
```suggestion
				fmt.Sprintf("Incorrect configuration for Weighted Destination for upstream: %s - Weighted Destinations require a weight that is greater than 0", weightedDestUpstream.Name),
```
should this say "greater than 0"? or "at least 1"?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -429,6 +429,17 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
...
+				routeName,
+			)
+		}
+
 		totalWeight += weightedDest.GetWeight()
```
jenshu, line 443: if the weightedDest weight was invalid (e.g. a negative number), do we still want to include it in the totalWeight here?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -455,6 +466,8 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
 		}
 	}
 
+	//Envoy has a default total weight of 100 and requires all weights to equal 100
```
jenshu, line 469: can you clarify if the weight total is required to equal 100, or do they just need to equal the total weight you provide?

nfuden: We likely want to reject full route which the error should do?

projects/gloo/pkg/translator/route_config.go
nfuden: We likely want to reject the entire route if any part is incorrect which in retrospect a single appended error may not be sufficient

projects/gloo/pkg/translator/route_config.go
nfuden: Envoy doesn't require total weight to be 100. We must set total weight to be higher than 0

projects/gloo/pkg/translator/route_config.go
ianmacclancy: For integer values those two statements mean the same. I worded it this way as it captures the case where someone has passed 0 as a weight. It better calls out the situation they put themselves in.

projects/gloo/pkg/translator/route_config.go
ianmacclancy: @jenshu The weight is an unsigned integer so it cannot be negative. 

@nfuden When I assign the error this way it appears on the resource report as a route error on the virtual service. I would expect this to reject the route and new virtual service it is on if the validation was working. As it is not, I was not able to test this.

projects/gloo/pkg/translator/route_config.go
ianmacclancy: I see how this could be misleading as it could be read that envoy always requires the weights to equal 100. I can update this to be clear that the total weights must equal the set total weight value.

projects/gloo/pkg/translator/route_config.go
```diff
@@ -455,6 +466,10 @@ func (h *httpRouteConfigurationTranslator) setWeightedClusters(params plugins.Ro
 		}
 	}
 
+	//Enovy has a default total weight of 100 and requires all weights to equal the current value of total weight
```
jenshu, line 469:
```suggestion
	//Envoy has a default total weight of 100 and requires all weights to equal the current value of total weight
```

projects/gloo/pkg/translator/route_config.go
nfuden: Works for me

jenshu (dismissed):

nfuden (approved):

PR #6585, opened 2022-06-21
setup/settings: Update to allow for maxvalidationbytes to take effect without manual kick

nfuden:
# Description

On settings update we check if grpc servers need a kick
Right now we only do it if the addr changes but we also need to set anything else that is passed in as a grpc opt. In this case the validation bytes.

# Context

\Users were confused when this did not take affect.

### DIFF
--- a/changelog/v1.12.0-beta23/grpcmaxvalidationbyte-refresh.yaml
+++ b/changelog/v1.12.0-beta23/grpcmaxvalidationbyte-refresh.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6564
+    description: Allow for validation bytes to be updated by setttings without a manual roll
+    resolvesIssue: false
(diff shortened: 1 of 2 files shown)


nfuden:
/kick  [Fail] CSRF defined on listener defined on listener with shadow mode config [It] should succeed with allowed origin, unsafe request

test/kube2e/gloo/setup_syncer_test.go
```diff
@@ -247,6 +264,53 @@ var _ = Describe("SetupSyncer", func() {
...
+				portFwd := exec.Command("kubectl", "port-forward", "-n", namespace,
+					"deployment/gloo", validationPort)
+				portFwd.Stdout = os.Stderr
+				portFwd.Stderr = os.Stderr
+				err := portFwd.Start()
+				Expect(err).ToNot(HaveOccurred())
+				return portFwd.Process
+			}
+
+			It("restarts validation grpc server when settings change", func() {
```
jenshu, line 294: 🎉

jenshu (approved):

PR #6612, opened 2022-06-23
add MaxCallSendMsgSize option to Validate call

bewebi:
# Description

- Add `defaultOpts` field to `connectionRefreshingValidationClient` (wrapper for the validation client) and prepend those opts to any passed on call to `Validate`
- Add `GrpcMaxSizeBytes` to `ValidationOpts` struct, assigned to value of `ValidationServerGrpcMaxSizeBytes`
- Pass `defaultOpts` containing `MaxCallRecvMsgSize` option value from `ValidationOpts.GrpcMaxSizeBytes` to `NewConnectionRefreshingValidationClient()` iff > 0

# Context

Currently setting `ValidationServerGrpcMaxSizeBytes` only has the intended behavior of setting a maximum size for responses from the validation server when the setting is less than the default (4MB)
This is because, while we had applied the setting as a [`MaxRecvMsgSize`](https://github.com/solo-io/gloo/blob/dab957fb161e7746875a058cbb6d3c10120d8807/projects/gloo/pkg/syncer/setup/setup_syncer.go#L254) option on the server, the client was still honoring the default value for maximum response size.

In this PR we use the same setting that populates the `MaxRecvMsgSize` option on the server to populate a `MaxCallRecvMsgSize` option on the client call, ensuring that if the setting is greater than the default, both the client and server will respect the higher maximum.

Given the user-facing setting does not refer specifically to gRPC receiving vs sending, it may make sense to apply `MaxSendMsgSize` on the validation server and `MaxCallSendMsgSize` on the client as well.

Correct behavior has been verified by creating a cluster which should generate a >4MB proxy CR (by applying 12k VirtualServices), setting `ValdiationServerGrpcMaxSizeBytes` to ~10MB and observing that the Gateway pod running the latest 1.9 version hits an error like:
```
{"level":"error","ts":"2022-06-23T17:39:56.342Z","logger":"gateway.v1.event_loop.gateway","caller":"syncer/setup_syncer.go:305","msg":"error: event_loop: 1 error occurred:\n\t* failed to add proxy validation results to reports: internal err: communication with proxy validation (gloo) failed: All attempts fail:\n#1: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4757282 vs. 4194304)\n\n","version":"1.9.16-2-gdab957fb1","stacktrace":"github.com/solo-io/gloo/projects/gateway/pkg/syncer.RunGateway.func1\n\t/Users/bernie/go/src/github.com/solo-io/gloo/projects/gateway/pkg/syncer/setup_syncer.go:305"}
```

On the current branch, the code successfully completes the rpc call to `Validate`. (FWIW in my local dev cluster another error is encountered when trying to write the Proxy because it exceeds the configured etcd resource size limit.)

*Note:* The bug appeared to surface in #5301, and indeed the problematic behavior only seems to occur from that point on, but this is misleading. The underlying behavior was always present, however in #5301 we begin [including entire proxy resources](https://github.com/solo-io/gloo/pull/5301/files#diff-2ab04dbc4a0dcf23297b16ea7ec39f37d76eff5eaa33fe90db0ead0b88c7b022R254) in the validation responses, dramatically increasing the possibility that they will be larger than the default `MaxCallRecvMsgSize` of 4MB. We should also investigate whether this is necessary and/or advisable.

# Checklist:


projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -23,13 +25,17 @@ type ProxyReconciler interface {
 }
 
 type proxyReconciler struct {
+	maxCallRecvMsgSize int
```
jackstine, line 28: nit: comment what is `maxCallRecvMsgSize` doing?

changelog/v1.9.18/fix-validation-server-grpc-max-size-bytes.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6596
+    resolvesIssue: false
+    description: Fix bug that caused `ValidationServerGrpcMaxSizeBytes` setting not to be honored if set >4MB.
```
sam-heilbron, line 5: nit: We may want to add more context outlining that the field was honored, but it only configured what the server can receive, and we never configured what the client can receive back. Perhaps it's context that is not relevant to the user, since all they care about is if the Proxy was >4MB then this call failed

projects/gateway/pkg/reconciler/proxy_reconciler.go
```diff
@@ -23,16 +25,18 @@ type ProxyReconciler interface {
 }
 
 type proxyReconciler struct {
-	statusClient   resources.StatusClient
-	proxyValidator validation.GlooValidationServiceClient
-	baseReconciler gloov1.ProxyReconciler
+	maxCallRecvMsgSize int // maximum size to be allowed for validation response
+	statusClient       resources.StatusClient
+	proxyValidator     validation.GlooValidationServiceClient
```
sam-heilbron, line 30: We have an existing GlooValidationServiceClient which is the interface of our validation client. I wonder if there's a way to merge this construct with our maxCallRecMsgSize since they are logically similar. To me the ProxyReconciler shouldn't care/be aware of the maxMessageSize, it should just be given a client that has that context build into it.

I guess in this proposal, we add `defaultGrpcOptions` to the validationClient and the code that executes that call can pass in others as well.

sam-heilbron (dismissed):
The proxy_reconciler is one place where we rely on a grpc request and pass a Proxy. However, we have other usages of it as well in our validation webhook that we need to update.

I think given that it's easy to miss the multiple places where we use this code, it would be valuable to construct a client at setup time with the proper options, and then rerly on that client in other areas of the code.

Code that we also use a validation client but don't pass options: https://github.com/solo-io/gloo/blob/v1.10.x/projects/gateway/pkg/validation/validator.go#L741

projects/gateway/pkg/translator/opts.go
```diff
@@ -32,4 +32,5 @@ type ValidationOpts struct {
 	AlwaysAcceptResources        bool
 	AllowWarnings                bool
 	WarnOnRouteShortCircuiting   bool
+	GrpcMaxSizeBytes             int
```
sam-heilbron, line 35: Is this used?

sam-heilbron (commented):
I think there are a few leftover variables that we no longer need.

Given that a response is both a proxy+proxyReport, the client will hit its buffer limit before the server will. It's probably ok, but worth calling out. Or we make the response contain only proxy metadata

bewebi:
`validationOpts.GrpcMaxSizeBytes` is used to populate (or not) the `MaxCallRecvMsgSize` option [here](https://github.com/solo-io/gloo/blob/2a6ab7a5a263dd23ad844691411ef7b7c8ac7953/projects/gateway/pkg/syncer/setup_syncer.go#L254)

The key change is that if the client hits a limit before the server, the user can now configure the limit to be higher!

sam-heilbron:
Looks great!

Calling out that we're intentionally keeping the Proxy in the response for this change, but I think it's something we need to revisit in the future. For example, for N proxies, we will now fail on the client when a proxy is 1/N the defined size. This may be problematic, but this PR addresses the immediate pains.

https://github.com/solo-io/gloo/pull/3407 Is where the Proxy in the response was introduced, and it seems useful for debugging. Perhaps we make it configurable

sam-heilbron:
Changes look good to me.

One nit: Let's avoid unnecessary imports, especially for something just used by a test.

sam-heilbron (commented):
Changes look good to me.

One nit: Let's avoid unnecessary imports, especially for something just used by a test.

PR #6616, opened 2022-06-23
[1.10] setup/settings: Update to allow for validationServerGrpcMaxSizeBytes to take effect without manual kick

jbohanon:
Description
On settings update we check if grpc servers need a kick
Right now we only do it if the addr changes but we also need to set anything else that is passed in as a grpc opt. In this case the validation bytes.

Context
Users were confused when the settings change was not reflected in the validation server's behavior


jackstine (approved):
Go

Issue #6476, opened 2022-05-19
HCM overrides do not support 0-value overrides


DoroNahari:
### Gloo Edge Version

1.11.x (latest stable)

### Kubernetes Version

1.22.x

### Describe the bug

This feature https://github.com/solo-io/gloo/pull/6414 presented an option to override settings with delegated gateways.
As described in the feature:
`preventChildOverrides=false | {"parent":{"foo":"bar"}, "child":{"foo":"baz"}} --> {"foo":"baz"}`

When setting a parent gateway with a boolean prop true and a child gateway with the same boolean prop false, the override is not working.

For example:
Parent gateway:
```
apiVersion: gateway.solo.io/v1
kind: Gateway
metadata:
  name: gateway-proxy-ssl
  namespace: gloo-system
spec:
  bindAddress: '::'
  bindPort: 8443
  hybridGateway:
    delegatedHttpGateways:
      httpConnectionManagerSettings:
        acceptHttp10: true
      selector:
        labels:
          hybrid-gateway: ssl
  proxyNames:
    - gateway-proxy
  ssl: true
  useProxyProto: false
```

Child gateway:
```
apiVersion: gateway.solo.io/v1
kind: MatchableHttpGateway
metadata:
  name: default-ssl-http-gateway-2
  namespace: apigw-gloo-ee
  labels:
    hybrid-gateway: ssl
spec:
  httpGateway:
    virtualServiceSelector:
      http-gateway: default-ssl
    options:
        httpConnectionManagerSettings:
          acceptHttp10: false

...
```
The observed behavior was that http1.0 yet was accepted

gunnar-solo:
this is a real ...ish issue.  It's mostly a logic problem.  Consider the field `SkipXffAppend`, which has this autogenerated getter:
```
func (x *HttpConnectionManagerSettings) GetSkipXffAppend() bool {
	if x != nil {
		return x.SkipXffAppend
	}
	return false
}
```

because booleans cannot be nil, our current implementation gives us these possibilities:
```
parent | child  | outcome
------ | ------ | -------
true   | true   | true    # as-expected
true   | false  | true    # raises issue, as _optimally_, we want "false"
true   | unset  | true    # as-expected
false  | true   | true    # as-expected
false  | false  | false   # as-expected
false  | unset  | false   # as-expected
unset  | true   | true    # as-expected
unset  | false  | false   # as-expected
unset  | unset  | false   # as-expected

[...]

PR #6622, opened 2022-06-24
Isolate Secure VirtualHosts On Separate FilterChains (Part 2)

sam-heilbron:
# Description

# Remaining Work
## Before Merge
- [x] HybridGateway.MatchedGateway
- [x] HybridGateway.DelegatedGateway
- [x] e2e test to validate MatchedGateways
- [x] e2e test to validate DelegatedGateways

## After Merge
- [ ] Public docs (This feature is documented in the API, and we can have more of a conversation around how to add it to our docs, but that is not blocked the merge of the feature)

# Solution

Builds off of work that was introduced in https://github.com/solo-io/gloo/pull/6558

## Hybrid Translator
There are some slight changes to how the HybridTranslator is implemented, but the functionality has remained the same.

Previously, we would reconcile (merge) sslConfigurations from the VirtualService and Gateway, and then apply the merged config back on the original resource, modifying it in place. I didn't write a test to verify this, but my concern was that by modifying the original Gateway resource, we would be affecting translation. I restructured the code and split off functions, since I would need these functions in my new functionality. I also ensured that the functions return instead of modifying.

Due to this change, I had to update the tests that validated this merge logic. The tests compared an expected gateway with what resulted after translation. Since we are no longer modifying the gateway in place, I updated the test to use the outputted listener and compare the fields there. The values were the same.

## Aggregate Listener
Wrote a bunch of code for this, but it's largely copy/reuse/paste from the Hybrid Translator

## Aggregate Listener Builder
An aggregate listener attempts to reduce space by storing a single copy of a resource, and then references whenever it is used in a filter chain. This means that the process of indexing the resource can be verbose and make the code harder to read. Also, since some code requires performing this action multiple times, I added a utility builder that just aggregates the state. It allows us to keep the intention of the code clearer (I think) and reduces the amount of variables we need to maintain/return.

## Proxy Reconciler
Added support for the new listener type and added corresponding tests

## Gateway Selector
Previously, it was possible to have a Gateway CR without SSL, select a MatchableHttpGateway with SSL. Now it's no longer possible. A few tests needed to be updated to reflect this.

## Envoy End to End Tests
I added test cases to demonstrate the same behavior that we previously had demonstrated for HttpGateways, but this time for:
-  Insecure HybridGateway that defines sub-gateways inline
- Secure HybridGateway that defines sub-gateway inline
- Insecure HybridGateway that selects sub-gateways using delegation
- Secure HybridGateway that selects sub-gateways using delegation

# Checklist:


projects/gateway/pkg/translator/hybrid_translator_test.go
```diff
@@ -364,24 +364,32 @@ var _ = Describe("Hybrid Translator", func() {
...
 
 					// perform translation
 					params := NewTranslatorParams(ctx, snap, reports)
-					hybridTranslator.ComputeListener(params, defaults.GatewayProxyName, snap.Gateways[0])
+					listener := hybridTranslator.ComputeListener(params, defaults.GatewayProxyName, snap.Gateways[0])
 
 					// evaluate results
-					hcm_after := child.GetHttpGateway().GetOptions().GetHttpConnectionManagerSettings()
-					ssl_after := child.GetMatcher().GetSslConfig()
+					Expect(listener).NotTo(BeNil())
```
sam-heilbron, line 380: We perform the same evaluation here. The difference, is that we compare the definition on the produced listener instead of the original gateway. This is necessary since we no longer modify the gateway

projects/gateway/pkg/translator/merge.go
```diff
@@ -87,3 +90,53 @@ func isEmptyValue(v reflect.Value) bool {
...
+		return parentClone
+	}
+	if parent == nil {
+		// use child exactly as-is
+		return childClone
+	}
+
+	if preventChildOverrides {
+		// merge, preferring parent
+		mergo.Merge(childClone, parentClone, mergo.WithOverride)
```
ben-taussig-solo, line 110: Gunnar and I have each run into issues with mergo in certain edge cases, in particular as it relates to how it handles merging nil values and how that interacts with the `WithOverride` setting. Gunnar wrote up a nice summary of this probably about 1/2 months ago. I'm going to link it here when I find it, but just want to call this out as a potential source of unexpected behavior.

ben-taussig-solo: Here's Gunnar's breakdown of the issue: https://github.com/solo-io/gloo/issues/6476#issuecomment-1132973401

sam-heilbron: This function was created by copying the code that Gunnar wrote, and moving it to a separate function so that it could be evaluated in multiple places. The only change I made was to return early if possible. All existing tests that Gunnar had written are passing. Do you have a concern with a particular case?

ben-taussig-solo: Ok, I figured this was probably the case -- In that case I think we're fine, just wanted to call it out

gunnar-solo: no issue here, @sam-heilbron.  I think @ben-taussig-solo is calling out that refactor preserves the existing bug.  Not a big deal, since that's not the point of this PR

projects/gateway/pkg/translator/merge.go
```diff
@@ -87,3 +90,53 @@ func isEmptyValue(v reflect.Value) bool {
 	}
 	return false
 }
+
+func mergeSslConfig(parent, child *v1.SslConfig, preventChildOverrides bool) *v1.SslConfig {
+	// Clone to be safe, since we will mutate it
+	parentClone := proto.Clone(parent).(*v1.SslConfig)
```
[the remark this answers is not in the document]
ben-taussig-solo: +1 to this

.github/workflows/regression-tests.yaml
```diff
@@ -2,10 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - 'master'
-  pull_request:
```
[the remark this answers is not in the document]
sam-heilbron: Handled by: https://github.com/solo-io/gloo/pull/6634

projects/gateway/pkg/translator/merge.go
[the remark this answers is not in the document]
sam-heilbron: Agreed!

projects/gateway/pkg/reconciler/proxy_reconciler_test.go
```diff
@@ -145,6 +158,24 @@ var _ = Describe("ReconcileGatewayProxies", func() {
 				Expect(vhosts).To(HaveLen(1))
 				Expect(vhosts).To(ContainName(translator.VirtualHostName(goodVs)))
 			})
+
+			It("only creates the valid virtual hosts (using IsolateVirtualHosts Feature)", func() {
```
sam-heilbron, line 162: These tests mirror the test above it. But we generate the proxy using our new feature flag, to generate a different type of listener (aggregate). The test then verifies that the listener has the same features as expected. If you modify the actual proxy_reconciler code taht I added, these tests fail (which is good!)

gunnar-solo (approved):
Make it so.

sam-heilbron:
Another instance where the final test (helm regression test) actually completed and passed, but github never signaled that the job was complete

PR #6633, opened 2022-06-24
makefile: Local trivy runs outside of go-utils update

nfuden:
Update makefile for changes to trivy. 
We normally do not exercise this code path as we have a more robust scanning for existing vuln but this is a nice one to have for local dev.

Namely newer versions of trivy require the directive `image` when scanning images.


jenshu (approved):

sam-heilbron:
Local dev only change, pushing through while we have number of PRs churning in CI

PR #6620, opened 2022-06-23
Exposing various Http1/2 settings

ianmacclancy:
# Context

Fixes for https://github.com/solo-io/gloo/issues/6466

# Description

Exposed envoy fields: 
- override_stream_error_on_invalid_http_message (added to upstream as an http2 option)
- override_stream_error_on_invalid_http_message (added to http1ProtocolOptions)
- request_headers_timeout (added to HttpConnectionManager)

From the issue, I identified and checked envoy and our configs align for these fields 
- initial_connection_window_size (already exists in upstream)
- initial_stream_window_size (already exists in upstream)
- stream_idle_timeout (already exists in HttpConnectionManager)

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #6635, opened 2022-06-24
Backport Expose various HTTP/2 settings - issue 6466

ianmacclancy:
Backporting https://github.com/solo-io/gloo/pull/6620


sam-heilbron (dismissed):

nfuden:
Looks like it needs codegen updates.

Also can we name the pr differently. The title becomes the squashed commit title which means in the future its a bit harder to understand via git blame

sam-heilbron (approved):

PR #6504, opened 2022-05-26
Clarify Gateway Translation Steps

sam-heilbron:
# Description

Add comments around aspects of gateway translation

# Context

Users ran into this bug doing ... \ Users needed this feature to ...

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #6637, opened 2022-06-26
[1.11] Isolate Secure VirtualHosts On Separate FilterChains

sam-heilbron:
# Description

# Context
This code was merged into the main branches in multiple stages. Therefore, to support on v1.11, we must backport each of the relevant pieces.

## Documentation improvements to translation https://github.com/solo-io/gloo/pull/6504
This pull request introduced code that the code to implement the isolate virtual hosts feature depended on. When attempting to cherry-pick the feature alone, there were a lot of merge commits. Since this PR is mostly non-functional I decided to backport it,  in an attempt to make the other cherry-picks easier.

## Isolate Virtual Hosts (Part 1) https://github.com/solo-io/gloo/pull/6558
Part 1 of support for isolating virtual hosts by ssl configuration

## Isolate Virtual Hosts (Part 2) https://github.com/solo-io/gloo/pull/6622
Part 2  of support for isolating virtual hosts by ssl configuration

## Additional One-Off Changes
In https://github.com/solo-io/gloo/pull/5888/ we merged the Gloo+Gateway pods. This was not backported to 1.11, but included a few fixes that were required for some of the features in this PR.
- HttpGateways are instantiated at setup time and included as part of ApiSnapshot
- Cluster Role includes those HttpGateway resources

# Checklist:


sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d4664b93-4c26-4eb3-87fb-bf03296fc1e0

install/helm/gloo/templates/21-namespace-clusterrole-ingress.yaml
```diff
@@ -19,7 +19,7 @@ rules:
   resources: ["customresourcedefinitions"]
   verbs: ["get", "create"]
 - apiGroups: ["gloo.solo.io", "enterprise.gloo.solo.io", "gateway.solo.io"]
-  resources: ["settings", "upstreams", "upstreamgroups", "proxies", "authconfigs", "virtualservices", "routetables", "virtualhostoptions", "routeoptions", "gateways"]
+  resources: ["settings", "upstreams", "upstreamgroups", "proxies", "authconfigs", "virtualservices", "routetables", "virtualhostoptions", "routeoptions", "gateways", "httpgateways"]
```
nfuden, line 22: Question should we call this out in some upgrade docs or some such? Namely we want to call out loudly when clusterroles need to be remade.

projects/gateway/pkg/translator/predicate.go
```diff
@@ -0,0 +1,57 @@
+package translator
+
+import v1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1"
+
+// Predicate is used to determine how to reduce a list of Gateways
+type Predicate interface {
```
nfuden, line 6: nit: If we are going to be filtering gateways on arbitrary rules can we not lock our selves into having the interface being a Predicate? if we need to do something that is not purely a predicate in the future then this will muddy the waters alot.

install/helm/gloo/templates/21-namespace-clusterrole-ingress.yaml
sam-heilbron: This should be abstracted by helm. On an upgrade, the resources are different and will be applied.

projects/gateway/pkg/translator/predicate.go
sam-heilbron: I'm not sure I follow entirely. This interface, though perhaps overkill, just accepts a Gateway and return whether or not to process it. Any of the details used to decide whether or not to process that Gateway won't affect the interface.

Can you give me an example of a change that might occur that would cause this interface to break/change?

sam-heilbron: This idea was based on https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/predicate/predicate.go#L32

install/helm/gloo/templates/21-namespace-clusterrole-ingress.yaml
nfuden: Right as this isnt a Custom Resource its update but we do have people using gloo and manage their clusterroles independently. it seems that we should at least call it out in the changelog

projects/gateway/pkg/translator/predicate.go
nfuden: This is purely based on the name. 

A predicate implies that one can only give something that is at the start of the string to match.
In the future we may want to enable regex matching or some such. 
We see the same issue in go-utils whereby a we have a filtering interface called predicate that we now need to update to filter on things that are not just Predicates! Therefore if you have a predicate interface and multiple implementing structs you could change the name to be more future facing while maintaining the structs naming. 

Take SingleNamespacePredicate. It doesnt care if the filtering interface it implements is called Predicate or GatewayFilter so current functionality does not break. If we create a regex filter and it implements the predicate interface now you potentially have confusion in the implementation.
Because Predicate is an exported function if someone implements their own predicates if we choose to change the name down the line our change now becomes a breaking change.

install/helm/gloo/templates/21-namespace-clusterrole-ingress.yaml
sam-heilbron: Sounds good, I can update!

Are you proposing that we call out all RBAC related config that Gloo manages moving forward? I had assume that we manage the RBAC definitions to power Gloo and that is more of an implementation detail that doesn't need to be explicitly called out. But if it's valuable, I like the idea of being explicit in changelogs moving forward

projects/gateway/pkg/translator/predicate.go
sam-heilbron: GatewayPredicate? Since it's in the gateway package, I assumed the name didn't need to include additional context, since that could be assumed from the package. Happy to update though to make it clearer.

I had was stuck between having a GatewayPredicate, with a single Read method, and having a generic predicate with a ReadGateway method. I opted for the latter, in case we want to add custom behavior to filtering out other types of resources (in which case we could add a new method to the interface)

sam-heilbron: `A predicate implies that one can only give something that is at the start of the string to match.`

To me a Predicate doesn't imply this, but instead just returns true/false to a question (https://www.baeldung.com/cs/predicates) . And the decision about how (based on namespace, regex, label, annotation...etc) is abstracted from the user.

nfuden: Ooh in gateway package so should not be gateway predicate good call. That stutters. My issue is with the word predicate. But it's also a nit so I say we go forward with this

install/helm/gloo/templates/21-namespace-clusterrole-ingress.yaml
nfuden: Yeah I'm mainly only worried in backports as sec teams will notice increased permissions (even if this isn't really that much of an increase)

sam-heilbron: This change to cluster roles isn't optional to users, as the Gloo pod will crashloopbackoff without it (on Ingress installations). Since we will start up a ResourceClient, and attempt to perform a ListOperation, which is then rejected and returns a 403. 
Again, open to publicizing this type of change if you feel that it's valuable for users to be aware of

nfuden: I think that having access changes is in general good but since it crashes immediately I am fine with it

nfuden (approved):

PR #6642, opened 2022-06-27
Backporting new status handling to v1.8

nfuden:
# Description

This allows for reverting from versions with the new status functionality.

# Context

Upgrading to 1.9 sets a new style of status, this will allow solo-kit to not error when it sees the statuses.
Gloo is the only one writing these statuses so we are safe to allow a new run of gloo to purge the status and reupdate with the current state fo the world.

To validate one should install 1.9 and setup the hello world example (ie have an upstream)
Get the definition of an object such as an upstream and verify you have the new namespaced statuses
Then uninstall gloo (not all just normal uninstall) install with this branch. 
Then verify that eventually (a little bit of time) no more errors are thrown and the object you previously got now has the old style statuses.


sam-heilbron (commented):
Can we include steps in the description that a reviewer can take to manually verify this change behaves as expected?

jackstine (dismissed):
Bump It

elcasteel (approved):

Issue #4152, opened 2021-01-19
Allow setting istio-proxy image for istio sds during a helm installation of gloo


malmanzor:
**Is your feature request related to a problem? Please describe.**

When deploying gloo via helm I can set `global.istioSDS.enabled: true` to add istio/sds sidecars to the gateway-proxy. We would like to customize the image used for the `istio-proxy container` so that it matches the version of istio that we are using rather than the default version that is set in the gloo chart https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L301. Currently, this is only possible by providing a `customSidecar` definition which would be overkill for just setting the `istio-proxy` image name.

**Describe the solution you'd like**

I would like to be able to set the istio-proxy container name via a helm value:

~~~
global:
  istioSDS:
    enabled: true
    proxyImage: docker.io/istio/proxyv2:<myversion>
~~~

**Describe alternatives you've considered**

Currently, we update the gloo helm chart as part of our build process to change the image version.

rinormaloku:
> global:
>   istioSDS:
>     enabled: true
>     proxyImage: docker.io/istio/proxyv2:<myversion>

The following PR adds this feature: https://github.com/solo-io/gloo/pull/6651. I put it under glooMtls and named it `envoySidecarImage` so that it is similar to the other options.

```
global:
  istioSDS:
    enabled: true
    customSidecars:
  glooMtls:
    sds:
      image:
        tag: 1.11.16
    envoySidecarImage: docker.io/istio/proxyv2:1.12.6
```

jenshu:
this will be available in gloo OSS v1.12.0-beta27 and v1.11.20

the new helm value is `global.glooMtls.istioProxy.image` e.g.
```
global:
  glooMtls:
    istioProxy:
      image:
        repository: proxyv2
        registry: docker.io/istio
        tag: 1.9.5
```

PR #6651, opened 2022-06-28
Parametrize the gateway proxy sidecar image

rinormaloku:
# Description

Currently, installing Gloo Edge with Istio doesn't work as the istio-proxy is a very old image. 1.9.5.

This parametrizes the image so that users can self-define the version of istio when deploying with helm.

Fixes https://github.com/solo-io/gloo/issues/4152


jenshu (changes requested):

install/helm/gloo/generate/values.go
```diff
@@ -606,6 +606,7 @@ type Mtls struct {
 	Enabled               *bool                 `json:"enabled,omitempty" desc:"Enables internal mtls authentication"`
 	Sds                   SdsContainer          `json:"sds,omitempty"`
 	EnvoySidecar          EnvoySidecarContainer `json:"envoy,omitempty"`
+	EnvoySidecarImage     *string               `json:"envoySidecarImage,omitempty" desc:"Sets the Istio-proxy container image to use for mTLS"`
```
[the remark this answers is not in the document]
nfuden: Oof that's is a good point given the scope where this is placed

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -350,7 +350,7 @@ spec:
 {{- end }}
 {{- if not $global.istioSDS.customSidecars }}
       - name: istio-proxy
-        image: docker.io/istio/proxyv2:1.9.5
+        image: {{ $global.glooMtls.envoySidecarImage }}
```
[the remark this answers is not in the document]
rinormaloku: What is the idea of the global image? If it is about gloo edge, then there is no correlation with the global image.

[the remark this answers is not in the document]
jenshu: it has defaults for some fields in case you don't provide them here. probably the only relevant one for defaulting is the `pullPolicy` since you will be overriding the registry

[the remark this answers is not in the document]
rinormaloku: I updated the code according to your suggestion. Thanks for the information, Jen!

install/helm/gloo/generate/values.go
```diff
@@ -606,6 +606,7 @@ type Mtls struct {
 	Enabled               *bool                 `json:"enabled,omitempty" desc:"Enables internal mtls authentication"`
 	Sds                   SdsContainer          `json:"sds,omitempty"`
 	EnvoySidecar          EnvoySidecarContainer `json:"envoy,omitempty"`
+	EnvoySidecarImage     *Image                `json:"envoySidecarImage,omitempty" desc:"Istio-proxy image to use for mTLS"`
```
jenshu, line 609: for the existing `EnvoySidecarContainer` above, it's used in the gloo deployment in the `envoy-sidecar` container.

how about to mirror that, we can name this new field IstioProxy
```suggestion
	IstioProxy     IstioProxyContainer               `json:"istioProxy,omitempty" desc:"Istio-proxy container"`
```
and define container:
```
type IstioProxyContainer struct {
	Image *Image `json:"image,omitempty" desc:"Istio-proxy image to use for mTLS"`
}
```

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -350,7 +350,8 @@ spec:
 {{- end }}
 {{- if not $global.istioSDS.customSidecars }}
       - name: istio-proxy
-        image: docker.io/istio/proxyv2:1.9.5
+        {{- $istioProxyImage := merge $global.glooMtls.envoySidecarImage $global.image }}
+        image: {{ template "gloo.image" $istioProxyImage }}
```
jenshu, line 354: on line 434 there's a hardcoded `imagePullPolicy`. we should use the `$istioProxyImage.pullPolicy`. can you also move the field up here to be right after `image`?

install/helm/gloo/generate/values.go
rinormaloku: @jenshu I updated the code according to your suggestion. However, I am not sure if you have seen my comment about the previous naming:

> it was dictated to me by the existing property https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L354-L356

With this update for someone to configure the istio-proxy container resources he would have to use the property: `envoySidecarResources` and to configure the image he'd have to use the `istioProxy.image`. Which are two different names envoy and istio proxy, while we mean istio-proxy in both cases.

jenshu: hmm, i think it's ok to keep the naming we have. in the gloo deployment, the envoySidecarResources is used under the `envoy-sidecar` container and the image uses `global.glooMtls.envoy.image`.
for this new one, i think `istioProxy.image` works because it matches the container name. it happens to reuse the envoySidecarResources value but i think that's ok. if we rename the envoySidecarResources in gateway-proxy-deployment, the user will have to define the resources in 2 places

rinormaloku: Sounds good to me!

jenshu (approved):

PR #6576, opened 2022-06-16
Add useBearerSchemaForAuthorization to extauth

inFocus7:
# Description

Please include a summary of the changes.

This bug fixes ... \ This new feature can be used to ...

# Context

Users ran into this bug doing ... \ Users needed this feature to ...

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #6654, opened 2022-06-30
Add useBearerSchemaForAuthorization to extauth - v1.8.x backport

inFocus7:
backport of #6576 
We currently don't do Bearer schema authorization when using "Authorization" token header.
This adds the extauth ability to configure useBearerSchemaForAuthorization. The ext auth service would then ensure the "Bearer" prefix for upstreams is configured when the header is "Authorization".
Context

Users ran into issues where Authorizations aren't in the form Bearer: <access_token>, but instead just <access_token>, so they need to use workaround on their end.


inFocus7:
codegen's bound to fail. going to see which files have a diff since my codegen had ~20ish files updated. (using go v1.17+, since m1 doesn't support lesser versions)

jbohanon (approved):

PR #6673, opened 2022-07-02
helm: fix rollout job when gateways are empty

jenshu:
Follow-up fix to https://github.com/solo-io/gloo/pull/6548

When the default gateways are disabled, no resources are rendered in the `gloo.customResources` function. We should not call kubectl apply when there is nothing to apply, as it results in an error.

### DIFF
--- a/install/test/helm_test.go
+++ b/install/test/helm_test.go
@@ -3712,6 +3712,15 @@ spec:
 							job := getJob(testManifest, namespace, "gloo-resource-rollout")
 							Expect(job.Spec.Template.Spec.Containers[0].Command[2]).NotTo(ContainSubstring("kubectl rollout"))
 						})
+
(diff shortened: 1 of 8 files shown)


install/test/helm_test.go
```diff
@@ -3712,6 +3712,15 @@ spec:
 							job := getJob(testManifest, namespace, "gloo-resource-rollout")
 							Expect(job.Spec.Template.Spec.Containers[0].Command[2]).NotTo(ContainSubstring("kubectl rollout"))
 						})
+
+						It("does not call kubectl apply when default gateways are disabled", func() {
```
elcasteel, line 3716: 🎉

elcasteel (approved):

ianmacclancy (commented):
LGTM

PR #6682, opened 2022-07-06
Fix issues with helm custom resources [1.11]

jenshu:
Backport of #6548 and #6673, minus 1.12-specific features like multi-namespace gateways


ianmacclancy (approved):
LGTM

PR #6695, opened 2022-07-07
Weighted destination test updates

ianmacclancy:
# Description

Adding a new test to cover situations where users pass weighted destinations with 0 weight with weighted destinations with weights higher than 0

This also includes a name change to a previous test to reflect that we now check total weight and will only fail if all passed destinations have 0 weight

### DIFF
--- a/projects/gloo/pkg/translator/translator_test.go
+++ b/projects/gloo/pkg/translator/translator_test.go
@@ -2244,10 +2244,11 @@ var _ = Describe("Translator", func() {
 
 		var (
 			multiActionRouteWithNoWeightPassedDest *v1.Route
-			multiActionRouteWithZeroWeightDest     *v1.Route
 			multiActionRouteWithOneValidDest       *v1.Route
+			multiActionRouteZeroAndFiveAsWeights   *v1.Route
 			expectedErrorString                    string
-			weightedDestValidWeight                *v1.WeightedDestination
+			weightedDestFiveWeight                 *v1.WeightedDestination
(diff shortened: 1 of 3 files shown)


projects/gloo/pkg/translator/translator_test.go
```diff
@@ -2298,17 +2299,48 @@ var _ = Describe("Translator", func() {
...
+			totalWeight := weightedDestFiveWeight.Weight + weightedDestZeroWeight.Weight
+			clusterNameFiveWeight := weightedDestFiveWeight.Destination.GetUpstream().Name + "_" + weightedDestFiveWeight.Destination.GetUpstream().Namespace
+			clusterNameZeroWeight := weightedDestZeroWeight.Destination.GetUpstream().Name + "_" + weightedDestZeroWeight.Destination.GetUpstream().Namespace
+
+			//There is only one route with a weight of 5 so total weight for the cluster should be 5
+			Expect(clusterAction.WeightedClusters.TotalWeight.GetValue()).To(Equal(totalWeight))
+			clusters := clusterAction.WeightedClusters.Clusters
+			Expect(clusters).To(HaveLen(2))
+
+			for _, c := range clusters {
```
jenshu, line 2331: do the clusters appear in a nondeterministic order?

ianmacclancy: I would not want a test to rely on the ordering of the clusters. It also makes sure you are explicit in which cluster you are checking for which weight.

jenshu (dismissed):

jenshu (approved):

Issue #6697, opened 2022-07-07
Default value missing for validationServerGrpcMaxSizeBytes


asayah:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

A default value is missing for validationServerGrpcMaxSizeBytes in case not provided in the setting crd, validationServerGrpcMaxSizeBytes is not a mandatory field, so anyone upgrading from a prior version that doesn't have a default in the setting will face the following bug in the gateway pod: 

```
{"level":"error","ts":"2022-07-07T22:25:45.651Z","logger":"gateway.v1.event_loop.gateway","caller":"syncer/setup_syncer.go:343","msg":"error: event_loop: 1 error occurred:\n\t* failed to add proxy validation results to reports: internal err: communication with proxy validation (gloo) failed: All attempts fail:\n#1: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (397 vs. 0)\n\n","version":"1.11.20","stacktrace":"github.com/solo-io/gloo/projects/gateway/pkg/syncer.RunGateway.func1\n\t/workspace/gloo/projects/gateway/pkg/syncer/setup_syncer.go:343"}

```

### Steps to reproduce the bug

- Install Gloo edge 1.11.20 ( or EE 1.11.24)
- Create a demo vs + upstream 
- Modify the setting CR to remove the value validationServerGrpcMaxSizeBytes
- Restart the gateway and gloo pod
- check the gatway pod log 
 

### Expected Behavior

A default value is provided

bewebi:
A fix has been merged into all Gloo LTS branches and will be released the next time those branches are released and eventually trickle to EE versions as well
If this fix urgently needs to be released in a particular version please reopen and say which version (and OSS or EE) and we will prioritize that

PR #6711, opened 2022-07-11
only set maxGrpcRecvSize if > 0

bewebi:
# Description

- Only set `maxGrpcRecvSize` on validation server if > 0

# Context

This ensures that we use the default value rather than `0` if `validationServerGrpcMaxSizeBytes` is not set, as described in #6697

The fix was manually verified e2e in a local kind cluster

# Checklist:

### DIFF
--- a/projects/gloo/pkg/syncer/setup/setup_syncer.go
+++ b/projects/gloo/pkg/syncer/setup/setup_syncer.go
@@ -266,7 +266,10 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 	if s.validationServer == emptyValidationServer {
 		// create new context as the grpc server might survive multiple iterations of this loop.
 		ctx, cancel := context.WithCancel(context.Background())
-		validationGrpcServerOpts := []grpc.ServerOption{grpc.MaxRecvMsgSize(maxGrpcRecvSize)}
+		var validationGrpcServerOpts []grpc.ServerOption
+		if maxGrpcRecvSize > 0 {
+			validationGrpcServerOpts = append(validationGrpcServerOpts, grpc.MaxRecvMsgSize(maxGrpcRecvSize))
+		}
 		s.validationServer = NewValidationServer(ctx, s.makeGrpcServer(ctx, validationGrpcServerOpts...), validationTcpAddress, true)
 		s.previousValidationServer.cancel = cancel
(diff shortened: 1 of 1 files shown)


projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -266,7 +266,10 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 	if s.validationServer == emptyValidationServer {
 		// create new context as the grpc server might survive multiple iterations of this loop.
 		ctx, cancel := context.WithCancel(context.Background())
-		validationGrpcServerOpts := []grpc.ServerOption{grpc.MaxRecvMsgSize(maxGrpcRecvSize)}
+		var validationGrpcServerOpts []grpc.ServerOption
```
jenshu, line 269: this will fall back to the default as defined by the grpc server, right?
would it be better to fall back on our default value as defined in the helm values?

bewebi: That's right

I went this this based on the assumption that if a user actually deletes this setting they do not want Gloo to be enforcing any `MaxRecvMsgSize` and that we shouldn't quietly/invisibly enforce a default that isn't visible to the user

It's a judgement call though, and I'm certainly open to using our own default...the argument there would be that it's more likely to end up with no `validationServerGrpcMaxSizeBytes` field due to upgrade from a version without a default and therefore we should use our own default

FWIW in 1.11 and earlier our default is the same as the gRPC default (4MB) though in 1.12 our default becomes 10MB

jenshu: just noticed that we are checking for size < 0 [here](https://github.com/solo-io/gloo/blob/a1b18f956843b05b4887eb949f6c562a6f946bf5/projects/gloo/pkg/syncer/setup/setup_syncer.go#L225)
can we move the == 0 check there too?
(that might simplify the 1.12 forward-port too, since both the proxy debug and validation servers use the same maxGrpcRecvSize variable, and we won't have to write the logic twice)

bewebi: Pushed that after manually verifying intended behavior, but I'm not sure it addresses your point
Perhaps connect offline to get on the same page about that if needed

jenshu (approved):

PR #6735, opened 2022-07-14
Add key_prefix to InternalSession

inFocus7:
# Description

- Added `key_prefix` for in the `InternalSession`/`cookie` `AuthConfig` resource config.

- [ext-auth-service W.I.P](https://github.com/solo-io/ext-auth-service/pull/358)
- [solo-projects W.I.P](https://github.com/solo-io/solo-projects/pull/3890)

# Context

User(s) ran into issues when creating multiple AuthConfigs w/ cookie domain + subdomains set. This led to token values colliding, since the subdomain includes both the parent domain + itself's tokens under the same cookie names (+ could also lead to infinite callback loop when logging out of subdomain).

# Checklist:


inFocus7:
looks like regressions are flaking(?), unrelated test failures

```
• Failure in Spec Setup (BeforeEach) [0.360 seconds]
Kube2e: gateway Proxy reconciliation [BeforeEach] should process proxy with deprecated label 
/home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:281
```

inFocus7:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=501fab5a-388a-43c2-8e1c-d67facd92ebc

```
Step #10 - "test": Summarizing 1 Failure:
Step #10 - "test": 
Step #10 - "test": [Fail] RetryOnUnavailableClientConstructor [It] creates a new client and closes the old one 
Step #10 - "test": /workspace/gloo/projects/gateway/pkg/validation/robust_client_test.go:74
```

changelog/v1.12.0-rc3/add-key-prefix-to-extauth-cookies.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: For enterprise only changes, I tend to make it non-user-facing in the open source repo, and then a new feature when the functionality is supported in the enterprise repo.

changelog/v1.12.0-rc3/add-key-prefix-to-extauth-cookies.yaml
```diff
@@ -0,0 +1,6 @@
...
+    issueLink: https://github.com/solo-io/solo-projects/issues/3709
```
sam-heilbron, line 3: This is a link to a private repo. Since this shows upp in the changelog, and on our docs, it will be unreachable for all public users. Is there a reason against having an open source issue to track this work?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -331,6 +331,8 @@ message RedisOptions {
 
 message UserSession {
     message InternalSession {
+        // Prefix to append to stored cookies.
```
sam-heilbron, line 334: I assume this defaults to an empty string if not provided? Are there any limitations to the characters that are available for this key?

Also, for the RedisSession, I see that cookie_name is available. As part of this change, would it make sense to make that configurable or is there no benefit?

inFocus7: Yeah it'd default to an empty string, therefore it wouldn't affect anything (and therefore wouldn't break existing setups). I'll have to check the second point... I'm not entirely sure what characters are allowed for cookie key, but looking that up now though! 👍🏼 

I don't think there'd be a benefit for that for internal cookies. The RedisSession has the cookie_name to store a reference for the redis client. Although in our case, we have multiple cookies storing the values, which is why I added the `KeyPrefix` to modify all.

changelog/v1.12.0-rc3/add-key-prefix-to-extauth-cookies.yaml
inFocus7: Just replaced it, I found a similar issue in the open-source repo 👍🏼

inFocus7:
/kick
last shown log was `Step #6 - "grab-ginkgo-container": 3926b61b7194: Download complete`
https://storage.googleapis.com/solo-public-build-logs/log-e55cb7b7-1c0d-4682-830f-30155505a6d2.txt

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
inFocus7: Update: added a link to MDN's set-cookie, which includes information on legal characters for the cookie key/name

inFocus7:
/kick [test flake](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=cd8b3c85-ebe1-4ec9-aea2-5af9ed083829)

```
Step #10 - "test": [Fail] CSRF defined on listener defined on listener with shadow mode config [It] should succeed with allowed origin, unsafe request 
Step #10 - "test": /workspace/gloo/test/e2e/csrf_test.go:260
```

inFocus7:
Rerunning failed job + created flake issue #6819

jackstine (approved):
LGTM!

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -538,12 +538,14 @@ redis socket types
...
 ```yaml
 "allowRefreshing": .google.protobuf.BoolValue
+"keyPrefix": string
 
 ```
 
 | Field | Type | Description |
 | ----- | ---- | ----------- | 
 | `allowRefreshing` | [.google.protobuf.BoolValue](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/bool-value) | When set, refresh expired id-tokens using the refresh-token. Defaults to false. Explicitly set to true to enable refreshing. |
+| `keyPrefix` | `string` | Prefix to append to cookies keys. Cookie keys will be stored in the form `<key_prefix>_<cookie_name>`. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes. |
```
artberger, line 548: A couple grammar nits, and how about adding a use case?
> Prefix to append to cookie keys, such as to enforce a specific semantic like `__Secure-`. Cookie keys are stored in the form `<key_prefix>_<cookie_name>`. For more information, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes.

inFocus7: Good callout! In our use case the `KeyPrefix` is more-so that the stored authentication cookies are unique. 

For example when we have authentication cookies stored in a domain **and** subdomain, we don't want to just set the same cookies because there'll be duplicates cookie keys in the browser (ex. `id_token = domainToken, id_token = subdomainToken`). With this `KeyPrefix`, a user can configure (per-domain/authentication configuration) and they can be stored as `<domain_key_prefix>_id_token = domainToken, <subdomain_key_prefix>_id_token=subdomainToken`

artberger: Nice, that use case makes more sense; updated my comment!

PR #6740, opened 2022-07-15
handles NPE

jackstine:
# Description

There is a NPE that crashes gloo in the `MarkPerFilterConfig()`.  This will prevent the NPE. 

# Context

Users are facing an issue when the gloo pod crashes because the destination is set to `nil`. Context as to why the destination is set to `nil` is not known yet.
# Checklist:


projects/gloo/pkg/plugins/pluginutils/per_filter_config.go
```diff
@@ -118,7 +118,11 @@ func ModifyPerFilterConfig(
 		return nil
 	}
 
-	err = errors.Errorf("unexpected destination type %v", reflect.TypeOf(inAction.GetDestination()).Name())
+	err = errors.New("unexpected destination type that is nil")
```
sam-heilbron, line 121: looks good to me!
could this also be handled by adding a default case to the above switch statement?

jackstine: Why does that matter

sam-heilbron: Oh sorry wasn't proposing that we do it that way, was more just clarifying my understanding of the change

jenshu (approved):

PR #6741, opened 2022-07-18
Gateway install cleanup

elcasteel:
# Description
Remove the helm charts that defined the gateway deployment and service accounts used by the gateway pods.

Also update the version mismatch check in glooctl check to use an annotation in the gloo deployment instead of relying on the image in the gateway pod. As part of this I removed a duplicate check from glooctl (we used to check at both glooctl startup and in a separate implementation in `glooctl check` )

# Context

This is follow up work from removing the gateway pod. At first we were setting the number of replicas of the gateway deployment to 0. That worked but given that the gateway deployment does not work it makes more sense to remove it completely. 

# Checklist:


install/helm/gloo/templates/5-gateway-deployment.yaml
```diff
@@ -1,132 +0,0 @@
-{{- define "gateway.deploymentSpec"}}
-{{- $image := .Values.gateway.deployment.image }}
```
jenshu, line 2: do we also want to remove any helm values (if any) from values.go and values-template, that are no longer used?

elcasteel: Good idea - I removed the gateway.deployment and gateway.serviceAccount fields from values.go and values-template.

elcasteel:
> 
> How does this deletion affect downgrades? How would a user upgrade to 1.12 and then downgrade from it?

There aren't any CRD changes in this PR which rules out a pretty big source of downgrade issues. The other thing that I think is worth at least a manual test is ensuring that values overrides for the values removed from values.go can be specified on a downgrade or persisted on an upgrade->downgrade

elcasteel:
I ran a manual upgrade and rollback between v1.11.23 and this version and some basic routes worked and I also found that the label I added to the gateway deployment on the original install was still present after the rollback.

Makefile
```diff
@@ -478,7 +455,7 @@ certgen-docker: $(CERTGEN_OUTPUT_DIR)/certgen-linux-$(GOARCH) $(CERTGEN_OUTPUT_D
 # Build All
 #----------------------------------------------------------------------------------
 .PHONY: build
-build: gloo glooctl gateway discovery envoyinit certgen ingress
+build: gloo glooctl discovery envoyinit certgen ingress
```
jenshu, line 458: I just remembered that glooctl [uses the gateway image tag](https://github.com/solo-io/gloo/blob/2caed6cc339515bb6f9b6ea1fd772f8e70c7e637/projects/gloo/cli/pkg/cmd/check/root.go#L947) to determine the OSS gloo version. Now that gateway is gone is there another image tag we can check instead? or maybe we should do the check in another way. Note that we specifically need the OSS image tag, not the EE tag
```
Server: {"type":"Gateway","enterprise":true,"kubernetes":{"containers":[{"Tag":"1.12.0-beta9","Name":"discovery-ee","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta9","Name":"extauth-ee","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta17","Name":"gateway","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta9","Name":"gloo-ee-envoy-wrapper","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta9","Name":"gloo-ee","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta9","Name":"observability-ee","Registry":"quay.io/solo-io"},{"Tag":"1.12.0-beta9","Name":"rate-limit-ee","Registry":"quay.io/solo-io"},{"Tag":"6.2.4","Name":"redis","Registry":"docker.io"}],"namespace":"gloo-system"}}
```

elcasteel: There aren't any other deployments that will have the OSS image tag when EE is installed. We should probably add an extra label to the gloo pod with the OSS version to use when there isn't a gateway deployment.

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -34,6 +34,7 @@ spec:
         prometheus.io/path: /metrics
         prometheus.io/port: "9091"
         prometheus.io/scrape: "true"
+        solo.io/ossImage: {{ .Values.gloo.deployment.ossImageTag }}
```
jenshu, line 37:
```suggestion
        gloo.solo.io/oss-image-tag: {{ .Values.gloo.deployment.ossImageTag }}
```
kubernetes generally seems to [use kebab-case](https://kubernetes.io/docs/reference/labels-annotations-taints/) for annotations/labels (although admittedly we are already inconsistent with our casing in gloo)

also noticed we use the prefix `gloo.solo.io` elsewhere

install/helm/gloo/templates/11-ingress-proxy-deployment.yaml
```diff
@@ -1,6 +1,6 @@
 {{- define "ingressProxy.deploymentSpec"}}
 {{- if .Values.ingress.enabled }}
-{{- $image := .Values.gateway.deployment.image }}
+{{- $image := .Values.ingressProxy.deployment.image }}
```
jenshu, line 3: good catch!

install/helm/gloo/generate/values.go
```diff
@@ -229,6 +229,7 @@ type GlooDeployment struct {
 	ExtraGlooLabels       map[string]string `json:"extraGlooLabels,omitempty" desc:"Optional extra key-value pairs to add to the spec.template.metadata.labels data of the primary gloo deployment."`
 	ExtraGlooAnnotations  map[string]string `json:"extraGlooAnnotations,omitempty" desc:"Optional extra key-value pairs to add to the spec.template.metadata.annotations data of the primary gloo deployment."`
 	LivenessProbeEnabled  *bool             `json:"livenessProbeEnabled,omitempty" desc:"Set to true to enable a liveness probe for Gloo Edge (default is false)."`
+	OssImageTag           *string           `json:"ossImageTag,omitempty"`
```
jenshu, line 232: can you add a description for this field, and say that it should not be set by the user, it's for debugging(?) purposes only? (i'm afraid users will think they can override the oss dependency by setting this helm value)

jenshu: alternatively, is there a way we can set this annotation value somewhere in code, without it being passed through helm values? (i'm not sure if there's a good way or not)

install/helm/gloo/values-template.yaml
```diff
@@ -50,6 +50,7 @@ gloo:
     replicas: 1
     runAsUser: 10101
     serviceAccount: {}
+    ossVersion:
```
jenshu, line 53: nit: can we just leave this out since it will be generated? also, shouldn't it be `ossImageTag`?

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -28,6 +28,7 @@ var (
 	KnativeUniqueContainers = []string{"knative-external-proxy", "knative-internal-proxy"}
 	IngressUniqueContainers = []string{"ingress"}
 	GlooEUniqueContainers   = []string{"gloo-ee"}
+	ossImageAnnotation      = "solo.io/ossImage"
```
jenshu, line 31: same suggestion as above - `gloo.solo.io/oss-image-tag`

projects/gloo/cli/pkg/prerun/version_warning_test.go
```diff
@@ -143,8 +144,30 @@ var _ = Describe("version command", func() {
 	It("should warn when the versions differ by major version", func() {
 		versionGetter.versions = buildContainerVersions(false, []*version.Kubernetes_Container{{
 			Tag:      v_1_0_0,
-			Name:     prerun.ContainerNameToCheck,
+			Name:     prerun.ContainerNameToCheckAnnotation,
 			Registry: "test-registry",
+			OssTag:   v_1_0_0,
```
jenshu, line 149: just curious why do the tags all use underscores in these tests?

projects/gloo/cli/pkg/prerun/version_warning.go
```diff
@@ -140,22 +143,29 @@ type ContainerVersion struct {
...
+// this is determined by looking at either the version of gateway (if the pod is present) or the annotation in the gloo pod.
 func getOpenSourceVersions(podVersions []*versiondiscovery.ServerVersion) (versions []*versionutils.Version, err error) {
 	for _, podVersion := range podVersions {
 		switch podVersion.GetVersionType().(type) {
 		case *versiondiscovery.ServerVersion_Kubernetes:
 			for _, container := range podVersion.GetKubernetes().GetContainers() {
 				containerVersion, err := versionutils.ParseVersion("v" + container.GetTag())
+				if err != nil {
+					continue
+				}
```
jenshu, line 155: this seems to only be used if the container name matches. can we move it down into the `if` block below?

projects/gloo/cli/pkg/prerun/version_warning.go
```diff
@@ -140,22 +143,29 @@ type ContainerVersion struct {
...
+				if container.GetName() == ContainerNameToCheckTag {
+					versions = append(versions, containerVersion)
```
jenshu, line 157: should we `continue` here as we don't have to check the annotation if we already found the gateway pod

projects/gloo/cli/pkg/prerun/version_warning.go
```diff
@@ -140,22 +143,29 @@ type ContainerVersion struct {
...
+				}
+				containerOssVersion, err := versionutils.ParseVersion("v" + container.GetOssTag())
 				if err != nil {
 					// if the container version doesn't match our versioning scheme
 					// (ie, as of writing this the redis container is on version "5")
 					// then just skip it
 					continue
 				}
```
jenshu, line 165: this can also be moved into the `if` block below

projects/gloo/cli/pkg/prerun/version_warning.go
```diff
@@ -140,22 +143,29 @@ type ContainerVersion struct {
...
 
-				if container.GetName() == ContainerNameToCheck {
-					versions = append(versions, containerVersion)
+				if container.GetName() == ContainerNameToCheckAnnotation {
```
jenshu, line 167: what do you think about reversing the ordering and checking ContainerNameToCheckAnnotation before ContainerNameToCheckTag, since all versions going forward will have only the annotation?

install/helm/gloo/generate/values.go
elcasteel: Helm really seems like the most appropriate place to set a field in our deployments at install time.

projects/gloo/cli/pkg/prerun/version_warning.go
elcasteel: I rearranged it in the version I'm going to push soon but ultimately we still look at every container (to allow for multiple gloo deployments) so this won't really save us any checks

projects/gloo/cli/pkg/prerun/version_warning_test.go
elcasteel: git blame dates most of this to 2019 so it might have made sense then.

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -79,8 +77,6 @@ func RootCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra.
 
 			CheckMulticlusterResources(opts)
 
-			CheckVersionsMatch(opts)
```
jenshu, line 82: with the removal of this, does `glooctl check` no longer print warnings when the oss version doesn't match?

jenshu (approved):

PR #6759, opened 2022-07-19
helm: fix argocd hanging

jenshu:
# Description

Fixes a couple of issues causing install/upgrade to fail with ArgoCD:
- The `gloo-resource-migration`, `gloo-resource-rollout`, and `gloo-resource-cleanup` jobs previously had a hard-coded `ttlSecondsAfterFinished` of 0, meaning the jobs were cleaned up immediately after they finished. This was too quick for ArgoCD, which after the job ran would seemingly try to clean up a job that was no longer there, and would be stuck syncing forever. To fix this, made the `ttlSecondsAfterFinished` configurable and defaulted to 60.
- The `gloo-resource-migration` job is only meant to run `pre-upgrade`, but since ArgoCD [maps both pre-install and pre-upgrade hooks](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks) to the same PreSync phase, it runs the migration job on both installs and upgrades. On installs, there was an ordering issue where the Gloo Edge CRDs were not installed before the migration job runs and the job would fail. Since the job is meant to only run on upgrades, added a check to only run migration steps if the CRDs are already there.

# Checklist:


install/helm/gloo/templates/5-resource-migration-job.yaml
```diff
@@ -35,6 +35,11 @@ spec:
           - /bin/sh
           - -c
           - |
+            if ! kubectl get crd upstreams.gloo.solo.io; then
+              echo "Gloo Edge crds are not installed; skipping migration job"
```
ianmacclancy, line 39: Is resource migration only for upstreams? The printout reads as if there are no crds at all and it looks like we are only checking upstreams. Might be more clear to say no upstream crds installed.

jenshu: right now it works on gateways and upstreams (and in the future any other CRs that we add to the helm chart).

i figured that just checking one arbitrary CRD would be enough, as if it's not there then it probably means Gloo Edge is not installed yet

i can update the message to be a bit more clear though - how does this sound: `Could not find Upstream CRD. Gloo Edge may not be installed yet. Skipping migration job`

jenshu: updated in 7518bfb

ianmacclancy (approved):
LGTM - kicked off flaking tests.

Issue #6143, opened 2022-03-23
Create a new unwrapAsApiGateway option to transform response payloads from AWS Lambdas


jameshbarton:
### Is your feature request related to a problem? Please describe.

Some of our clients want to migrate from AWS API Gateway to Gloo Edge. They currently have API Gateway to call a number of AWS Lambdas. There is a pre-defined response format that is expected by API Gateway, as described [here](https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html#apigateway-types-transforms).

### Describe the solution you'd like

Create a new AWS `DestinationSpec` option similar to `unwrapAsAlb`, where the following JSON payload is expected and parsed by Gloo Edge:
```
{
      "statusCode": 200,
      "headers": {
        "Content-Type": "application/json"
      },
      "isBase64Encoded": false,
      "multiValueHeaders": { 
        "X-Custom-Header": ["My value", "My other value"],
      },
      "body": "{\n  \"TotalCodeSize\": 104330022,\n  \"FunctionCount\": 26\n}"
}
```

The expectation is Gloo Edge to be able to translate this into a valid response to the client where:
response status = statusCode
response headers = headers / multiValueHeaders
response body = body

This could be an option enabled by a flag `unwrapAsApiGateway` on this [DestinationSpec](https://docs.solo.io/gloo-edge/master/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/aws/aws.proto.sk/#destinationspec).

### Describe alternatives you've considered

1. User-space transformations
2. Custom WebAssembly filters

### Additional Context

Related to #6162

jameshbarton:
The customer provided us [this generic Golang Lambda function](https://github.com/solo-io/gloo/files/8374645/test-lambda.zip) as an example we can use in our test suite to verify that this change satisfies their requirements.

chrisgaun:
Sync with @nrjpoddar and @sam-heilbron if the work on this issues goes into second week of April (week of April 11th).

nrjpoddar:
@ben-taussig-solo if you're stuck/need help please reach out to @yuval-k

Rachael-Graham:
Tracking doc changes in https://github.com/solo-io/gloo/issues/6764

PR #6762, opened 2022-07-19
API Support for AWS API Gateway transformer

ben-taussig-solo:
# Description
Expose an enterprise-only API  to support unwrapping a response as though it were an API Gateway.

# Context

## Upgrade Envoy to 1.22
It includes a number of changes outlined in https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22/v1.22.0.html.

There are a number of features that are worth exploring in a separate PR. To keep this PR tightly scoped I aimed to resolve only the problems I encountered when upgrading.

`listener: the [ipv4_compat](https://www.envoyproxy.io/docs/envoy/v1.22.2/api-v3/config/core/v3/address.proto#envoy-v3-api-field-config-core-v3-socketaddress-ipv4-compat) flag can only be set on Ipv6 address and Ipv4-mapped Ipv6 address. A runtime guard is added envoy.reloadable_features.strict_check_on_ipv4_compat and the default is true.`

In Gloo, we automatically set the `Ipv4Compat` [to true](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/translator/listener.go#L56), meaning that any code that relied on ipv4 address would now fail with the following error:

```
 gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) listener: Only IPv6 address '::' or valid IPv4-mapped IPv6 address can set ipv4_compat: 0.0.0.0:8081
```

We have 2 options for resolving this:
1. Reject Gateways that specify a bindAddress that is ipv4
2. Convert Gateways that specify an ipv4 bindAddress into the ipv6 format

I elected to move forward with Option 2, so that the impact to end users is invisible. I would like to follow up this work with a broader "Support Envoy 1.22 task" and that would include other work such as:
- Warn on gateways that use a deprecated bind address format, so that even though we convert, users can be made aware and upgrade
- Explore other envoy features that Gloo should expose/support

## Support Enterprise-Only API for Lambdas
*This is enterprise only functionality, and therefore is considered a non-user facing change*

To support the migration of executing lambdas from Gloo as as opposed to the AWS API Gateway.

## Upgrade Envoy to 1.23
We are jumping past envoy 1.22 to envoy 1.23.

Will track important updates for that bump here.

# Checklist:

- [] I have made corresponding changes to the documentation

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6143


ben-taussig-solo:
/kick Failure on helm tests. Never seen this before. https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4fa2c85f-277f-4553-8d12-4a13cc2c6157

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
```diff
@@ -104,9 +105,15 @@ func makeFallbackListenerAndCluster(
...
-		}},
+		HttpFilters: []*envoyhcm.HttpFilter{
+			{
+				Name: wellknown.Router,
+				ConfigType: &envoyhcm.HttpFilter_TypedConfig{
+					TypedConfig: &any.Any{
+						TypeUrl: "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router",
+					},
+				},
+			}},
```
jackstine, line 116: Question: why are the HttpFilters now given a typeURL, whereas before they did not have it?

projects/gloo/pkg/syncer/aws/aws_translator_syncer.go
```diff
@@ -0,0 +1,91 @@
+package aws
```
sam-heilbron, line 1: I don't believe we need this translator syncer extension. These extensions are used to set configuration in our xds cache to be served up by external services. Since extauth and ratelimit receive config over xds in edge, they each require one. However, generic plugins are used to generate config that will be served to envoy

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -202,7 +204,14 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
 	// As outlined by the Envoy docs, the last configured filter has to be a terminal filter.
 	// We set the Router filter (https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router)
 	// as the terminal filter in Gloo Edge.
-	envoyHttpFilters = append(envoyHttpFilters, &envoyhttp.HttpFilter{Name: wellknown.Router})
+	envoyHttpFilters = append(envoyHttpFilters, &envoyhttp.HttpFilter{
+		Name: wellknown.Router,
+		ConfigType: &envoyhcm.HttpFilter_TypedConfig{
+			TypedConfig: &any.Any{
```
sam-heilbron, line 210: Is this due to envoy being bumped and now requires the typeUrl to be present? Is there a wellknown constant we can reference?

projects/gloo/api/v1/options/aws/aws.proto
```diff
@@ -76,4 +76,8 @@ message DestinationSpec {
     // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html
     bool unwrap_as_alb = 7;
 
+    // Unwrap the response as if the proxy was an AWS API Gateway.
+    // Cannot be configured simultaneously with unwrap_as_alb.
```
sam-heilbron, line 80: I like this callout, though as a user I'm confused about what the impact is, if I try to configure both. Is it that one is preferred over the other? Or will I see errors? I think we can clarify this further to remove any uncertainty

projects/gloo/api/external/envoy/extensions/aws/filter.proto
```diff
@@ -116,3 +122,6 @@ message AWSLambdaConfig {
   // Defaults to not refreshing on time period. Suggested is 15 minutes.
   google.protobuf.Duration credential_refresh_delay = 4;
 }
+
+message ApiGatewayTransformation {
```
sam-heilbron, line 126: How is this used?

projects/gloo/api/external/envoy/extensions/aws/filter.proto
```diff
@@ -45,6 +47,10 @@ message AWSLambdaPerRoute {
   // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html
   // Defaults to false.
   bool unwrap_as_alb = 5;
+
+  // transformer configuration used to process response data
+  // cannot be configured simultaneously with unwrap_as_alb
+  .solo.io.envoy.config.core.v3.TypedExtensionConfig transformer_config = 6;
```
sam-heilbron, line 53: Why do we use a boolean in the destination spec, but the transformer config on the per route? Unwap as ALP uses booleans in both places.

sam-heilbron (commented):
Looking good for the most part. I think we can remove some code that isn't necessary and clarify some of the APIs, then should be good to go!

projects/gloo/api/v1/options/aws/aws.proto
ben-taussig-solo: Fair point -- unwrap_as_alb will take priority over unwrap_as_api_gateway

projects/gloo/api/external/envoy/extensions/aws/filter.proto
ben-taussig-solo: This is needed in envoy to specify the transformation type when configuring the unwrapAsApiGateway functionality -- see [here](https://github.com/solo-io/envoy-gloo/pull/167/files#diff-c3f2d22ce071bbb117a988b1bde2a069e4eed7c380783e7c6164fd38b7665f2eR25)

I think this is probably not necessary in the gloo side of things, so I will remove it

projects/gloo/pkg/syncer/aws/aws_translator_syncer.go
ben-taussig-solo: Ok, makes sense, I will remove this portion

projects/gloo/pkg/translator/network_filters.go
ben-taussig-solo: Yes, this is due to envoy being bumped + now requiring the typeUrl to be present. There is not, as far as I know, but I think it makes sense that it could go somewhere along with `wellknown.Router`

sam-heilbron:
RouteTransformation Test failed on a previous run: 
```
 Timed out after 20.000s.
Step #10 - "test":   Error: Unexpected non-nil/non-zero extra argument at index 1:
Step #10 - "test":   	<*url.Error>: &url.Error{Op:"Post", URL:"http://localhost:11139/1", Err:(*net.OpError)(0xc003580be0)}
Step #10 - "test": 
Step #10 - "test":   /workspace/gloo/test/e2e/route_transformation_test.go:115
Step #10 - "test": 
```

Re-running to see if it's a flake, will run locally as well

jackstine (commented):
LGTM

projects/gloo/pkg/plugins/enterprise_warning/plugin_test.go
```diff
@@ -366,4 +367,108 @@ var _ = Describe("enterprise_warning plugin", func() {
 		})
 	})
 
+	Context("aws", func() {
+
+		It("will not err if aws.UnwrapAsApiGateway is nt configured on single destination route", func() {
```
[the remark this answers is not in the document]
sam-heilbron: Good catch! Updated

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
sam-heilbron: This is due to a change in envoy:

https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.22/v1.22.0.html

`config: type URL is used to lookup extensions regardless of the name field. This may cause problems for empty filter configurations or mis-matched protobuf as the typed configurations. This behavioral change can be temporarily reverted by setting runtime guard envoy.reloadable_features.no_extension_lookup_by_name to false.`

ianmacclancy (dismissed):
Some Nits - LGTM

projects/gloo/cli/pkg/flagutils/route.go
```diff
@@ -46,6 +46,8 @@ func AddRouteFlags(set *pflag.FlagSet, route *options.InputRoute) {
 			"e.g. in the case of static content (images, HTML, etc.) being served by Lambda")
 	set.BoolVarP(&route.Destination.DestinationSpec.Aws.UnwrapAsAlb, "aws-alb-unwrap", "", false,
 		"Sets if gloo should handle responses as if it was an ALB. Appropriately handles the response body and sets headers.")
+	set.BoolVarP(&route.Destination.DestinationSpec.Aws.UnwrapAsAPIGateway, "aws-api-gw-unwrap", "", false,
+		"Sets if gloo should handle responses as if it was an API Gateway. Appropriately handles the response body and sets headers.")
```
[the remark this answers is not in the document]
nfuden: Should it though? its on the destination of aws and that would stutter

projects/gloo/pkg/plugins/enterprise_warning/plugin_test.go
```diff
@@ -366,4 +367,108 @@ var _ = Describe("enterprise_warning plugin", func() {
...
+		It("will err if aws.UnwrapAsApiGateway is configured on multi destination route", func() {
+			p := NewPlugin()
+
+			route := &v1.Route{
+				Action: &v1.Route_RouteAction{
+					RouteAction: &v1.RouteAction{
+						Destination: &v1.RouteAction_Multi{
+							Multi: &v1.MultiDestination{
+								Destinations: []*v1.WeightedDestination{{
+									Destination: &v1.Destination{
```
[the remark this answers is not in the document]
sam-heilbron: Yeah that's fair, for the purpose of the test theweight doesn't matter, but I like the idea of having realistic config

install/helm/gloo/crds/gateway.solo.io_v1_Gateway.yaml
```diff
@@ -3783,6 +3805,8 @@ spec:
                                                     type: boolean
                                                   unwrapAsAlb:
                                                     type: boolean
+                                                  unwrapAsApiGateway:
```
nfuden, line 3808: Is there a reason we are using a bool again rather than converting to an enum to future proof against what seems to be an ever growing set of aws services that we are mimicking?

sam-heilbron: This was inherited from the existing API decision so I went with it. I would prefer to move this to an enum or a oneof as we expect only one to be set, but given the timing of this, I'd rather keep it as is

ianmacclancy (approved):

PR #6794, opened 2022-07-24
Fix/cache max sizes

nfuden:
Allow for max sizing to be applied to the cleaner portion of the filter.


changelog/v1.12.0-rc3/fix-caching-custom-layer-bytes.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/3206
+    resolvesIssue: false
+    description: Cache max size configurable on custom layer in addition to top layer
```
sam-heilbron, line 5: nit: When you say "custom layer as opposed to top layer", are you referring to being configured at the listener level as opposed to the global level?

projects/gloo/api/external/envoy/extensions/cache/grpc/config.proto
```diff
@@ -29,4 +29,8 @@ message GrpcCacheConfig {
 
   // Connection timeout  for retrieval from cache
   google.protobuf.Duration timeout = 2;
+
+  // Max payload size to cache. This may be set to protect against large
```
sam-heilbron, line 33: Is this the internal API for the Envoy filter? What does this value default to if no size is provided?

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -134,6 +134,10 @@ spec:
   extauth:
 {{- toYaml .Values.global.extensions.extAuth | nindent 4 }}
 {{- end }}
+{{- if .Values.global.extensions.caching }}
+  caching:
```
sam-heilbron, line 138: The settings proto uses a different key `caching_server`: https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/settings.proto#L379

sam-heilbron (changes requested):
The key used in the settings need to be cachingServer to match the API. Can we add a test to verify that passing in the appropriate value would result in yaml that fits the Settings API?

projects/gloo/api/external/envoy/extensions/cache/grpc/config.proto
nfuden: This is the internal api for our custom filter. Sets to 0 if not set but we handle that in our enterprise code.

changelog/v1.12.0-rc3/fix-caching-custom-layer-bytes.yaml
nfuden: Nope upstream introduced a custom maxbytes setup in their filter which we were previously applying our custom maxbytes to but we have a cleaner implementation at the grpc filter level that we would like to use.

install/helm/gloo/templates/18-settings.yaml
nfuden: Good point, was testing it with the replace in helm rather than testing the boot post. Changing

changelog/v1.12.0-rc3/fix-caching-custom-layer-bytes.yaml
sam-heilbron: Makes sense! I think I read that as a user and am slightly confused about the user facing implications. It seems this is a non user facing change though?

nfuden: Good call ou!

sam-heilbron (approved):
LGTM

PR #6709, opened 2022-07-11
Upgrade Security Scanner 

sam-heilbron:
# Description
Upgrade the security scanner to run against all images, not just the latest patch for each minor version.

# Context
This dependency upgrade includes a couple of enhancements, each of which are outlined below

### Scan Changes
We now run scans against all images in our LTS branches, as opposed to just on the latest patch version. This was changed in the scanner utility code and brings the behavior back to its original state. 

The go-utils PR linked above has context on why this change was introduced.

### Min Version Increase
We use to use 1.7 as the minimum version. However, since 1.8 is our true maintained min version, I increase the threshold to not scan images that we aren't updating

### Image Creation
We use to create images for the latest patch release in Solo-Projects and publish to the security tab in Gloo for all versions.

This has been changed to create a github issue for the latest patch in both solo-projects and gloo.

### Local Testing
```
IMAGE_REPO=quay.io/solo-io SCAN_DIR=_output/scans VERSION_CONSTRAINT="> v1.7.4, < v1.7.7" go run generate_docs.go run-security-scan -r glooe
```

This command will generate the security scan for 1.7.5 and 1.7.6, write those outputs to the _output/scans dir, and only 1.7.6 will result in a github issue being modified

The output logs from this command are:
```text
 IMAGE_REPO=quay.io/solo-io SCAN_DIR=_output/scans VERSION_CONSTRAINT="> v1.7.4, < v1.7.7" go run generate_docs.go run-security-scan -r glooe
{"level":"debug","ts":"2022-07-11T12:14:45.448-0400","caller":"cmd/generate_docs.go:313","msg":"Scanner will write results to directory: _output/scans"}
{"level":"debug","ts":"2022-07-11T12:14:45.449-0400","caller":"securityscanutils/securityscan.go:156","msg":"Processing user defined configuration for repository (solo-io, solo-projects)"}
{"level":"debug","ts":"2022-07-11T12:14:45.449-0400","caller":"securityscanutils/securityscan.go:163","msg":"Scanning github repo for releases that match version constraint: >v1.7.4 <v1.7.7"}
{"level":"debug","ts":"2022-07-11T12:14:50.436-0400","caller":"securityscanutils/securityscan.go:174","msg":"Number of github releases to scan: 2"}
{"level":"debug","ts":"2022-07-11T12:14:50.436-0400","caller":"securityscanutils/securityscan.go:193","msg":"GithubIssueWriter configured with Predicate: &{releasesByTagName:map[v1.7.6:0xc000202320]}"}
[...]

PR #6807, opened 2022-07-26
v1.10.x: Solo-kit and go-utils update

ben-taussig-solo:
# Description

Updated solo-kit and go-utils to their latest versions

# Context

go-utils bump also fixes https://github.com/solo-io/go-utils/issues/475

# Context

Resolves https://github.com/solo-io/go-utils/pull/487 in v1.10.x

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,7 +5,7 @@ steps:
 
 # $COMMIT_SHA is a default gcloud env var, to run via cloudbuild submit use:
 # gcloud builds submit --substitutions COMMIT_SHA=<commit sha>,REPO_NAME=solo-io/gloo,_PR_NUM=<<insert PR Number here>> --project solo-public
-- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.15'
+- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.24'
   args:
     - "--repo-name"
     - "$REPO_NAME"
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -5,7 +5,7 @@ steps:
 
 # $COMMIT_SHA is a default gcloud env var, to run via cloudbuild submit use:
 # gcloud builds submit --substitutions COMMIT_SHA=<commit sha>,REPO_NAME=solo-io/gloo,_PR_NUM=<<insert PR Number here>> --project solo-public
-- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.24'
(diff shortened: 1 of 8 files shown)


ben-taussig-solo:
/kick strange go mod download issue, can't identify any particular error? https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=dd104773-2c0b-46b0-89db-ea12d02a6735

cloudbuild.yaml
```diff
@@ -5,7 +5,7 @@ steps:
 
 # $COMMIT_SHA is a default gcloud env var, to run via cloudbuild submit use:
 # gcloud builds submit --substitutions COMMIT_SHA=<commit sha>,REPO_NAME=solo-io/gloo,_PR_NUM=<<insert PR Number here>> --project solo-public
-- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.15'
+- name: 'gcr.io/$PROJECT_ID/prepare-go-workspace:0.4.24'
```
sam-heilbron, line 8: Why is this change necessary? What was the error that we would see before this change?

ben-taussig-solo: This might not be necessary -- I was experiencing strange behavior with the CI build step, where it stopped about 1:00 and the build showed as failed. Now though, there appear to be full logs for that run. I'll try reverting the change and see if I can get CI to pass.

sam-heilbron (commented):
In the original PR that bumped go-utils to pull in the logging handler npe fix (https://github.com/solo-io/gloo/pull/6709), I added logic to one of our kube2e tests to hit the logging endpoint. The idea was that this test will only pass if the request can be made. Since it's recommended in our production docs I think it's valuable that we have automated testing to confirm that it works. Can we add the same check to this branch as well?

cloudbuild.yaml
ben-taussig-solo: This is not necessary -- resolving this comment as I've reverted this to the original value.

ben-taussig-solo:
> In the original PR that bumped go-utils to pull in the logging handler npe fix (#6709), I added logic to one of our kube2e tests to hit the logging endpoint. The idea was that this test will only pass if the request can be made. Since it's recommended in our production docs I think it's valuable that we have automated testing to confirm that it works. Can we add the same check to this branch as well?

Added this test in this PR: https://github.com/solo-io/gloo/pull/6807/commits/6d994c380592dd5d07d71203755cb024968cc3a5

sam-heilbron (approved):
Looks good!

I think in the future it would be valuable to link the PR from master that we're backporting, and outline the behavrioral change in Gloo Edge and why that is added

Issue #6860, opened 2022-08-04
bindAddr: 0.0.0.0 not working anymore in clusters that don't support ipv6 


asayah:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

Creating a gateway with bindAddr: 0.0.0.0 results in the following error in clusters where ipv6 is not supported: 

```
"details": "malformed IP address: ::ffff:0.0.0.0"
     }
    },
    {
     "name": "listener-0.0.0.0-8443",
     "error_state": {
      "failed_configuration": {
       "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
       "name": "listener-0.0.0.0-8443",
       "address": {
        "socket_address": {
         "address": "::ffff:0.0.0.0",
         "port_value": 8443,
         "ipv4_compat": true
        }
```

### Steps to reproduce the bug

Install Gloo 1.12 on a cluster with ipv4 deactivated  
create/modify a gateway CR to change `:: -> 0.0.0.0`

### Expected Behavior

Working

nfuden:
Reopening until released in ee

PR #6861, opened 2022-08-04
Set Ipv4Compat only on listeners with Ipv6 bind address

sam-heilbron:
# Description

Set ipv4Compat on a listener based on the state of the listener bind address

# Context

This value was always set to true in earlier versions of Gloo. 

In Envoy 1.22, the behavior changed such that this field could only be set to true if the listener address was an ipv4 compatible address.

The previous solution was to identify any ipv4 addresses (which would be rejected in Envoy 1.22) and convert them to ipv6 addresses. However, if you are running a kube cluster with only ipv4 support this causes issues.

The new solution is to identify ipv4 addresses and turn this ipv4Compat field to false so that it will not crash. Currently this solution will continue to allow ipv4 anywhere possible.

## Testing
The cors_test (https://github.com/solo-io/gloo/blob/master/test/e2e/cors_test.go#L40) is an example of a test that sets the listener bind address to an ipv4 compatible value. 

1. Run it locally against these changes, and notice that it passes.
2. Then change the listener translation code to always set ipv4Compat to true, regardless of the bind address. It should now fail. 
This demonstrates that the field is respected by Envoy and we should not set it when an ipv4 address is identified.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6860


projects/gloo/pkg/translator/utils.go
```diff
@@ -102,17 +102,48 @@ type typedConfigObject interface {
 // GetIpv6Address returns the IPv6 Address for a provided bindAddress
 // returns an error and the original address, if the bindAddress is not a valid IP address
 func GetIpv6Address(bindAddress string) (string, error) {
```
jenshu, line 104: is this function used anymore? looks like we're not calling it from translator

changelog/v1.13.0-beta2/ipv4-legacy-support.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+- type: FIX
+  issuelink: https://github.com/solo-io/gloo/issues/6860
+  resolvesIssue: true
```
jenshu, line 4: nit: should this be false until it gets into an enterprise build?

projects/gloo/pkg/translator/utils.go
nfuden: Its not but it seemed like a nice utility. Willing to remove it though if we think it will be cleaner.

changelog/v1.13.0-beta2/ipv4-legacy-support.yaml
nfuden: Good callout!

projects/gloo/pkg/translator/utils.go
jenshu: yeah i think we should remove if it's not used

projects/gloo/pkg/translator/utils.go
```diff
@@ -102,17 +102,48 @@ type typedConfigObject interface {
...
+		return true, true, nil
 	}
+	return true, false, nil
 
-	return bindAddress, nil
+}
+
+// isIPv4NotInv6 checks the string to see if it was unmapped.
+// Used as the standard net.Parse smashes everything to ipv6
+func isIPv4NotInv6(ipString string) bool {
```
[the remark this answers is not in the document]
nfuden: I am open to a better name but basically checking if the ipv4 address is not padded and in ipv6 space 
namely detect and reject ::ffff:0.0.0.0 over 0.0.0.0

projects/gloo/pkg/translator/utils.go
```diff
@@ -102,17 +102,48 @@ type typedConfigObject interface {
...
+	}
+	if isIPv4NotInv6(bindAddress) {
+		return true, true, nil
 	}
+	return true, false, nil
 
-	return bindAddress, nil
+}
+
+// isIPv4NotInv6 checks the string to see if it was unmapped.
```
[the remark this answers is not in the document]
nfuden: Pushed an update that sets examples of what ipv4 mapped to ipv6 space looks like.
Does it explain it sufficiently?

projects/gloo/pkg/translator/utils.go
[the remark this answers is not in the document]
jenshu: gotcha. so it's rejecting if it's an ipv4 address that's been converted to ipv6 format? how about something like `isIPv4Format` for the name, as if it's regular ipv4 format it should return true?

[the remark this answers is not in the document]
nfuden: The reason that I didn't go with that name is technically it's still sorta ipv4 even when it's been mapped to ipv6. Namely in the net package all it's are automated but can still return as being ipv4 when requested as such so I found it confusing

projects/gloo/pkg/translator/utils.go
```diff
@@ -99,20 +99,39 @@ type typedConfigObject interface {
...
+
+}
 
-	return bindAddress, nil
+// isIPv4NotInv6 checks the string to see if it is
+// ipv4 and not ipv4 mapped into ipv6 space and not ipv6.
+// Used as the standard net.Parse smashes everything to ipv6.
+// Basically false if ::ffff:0.0.0.0 and true if 0.0.0.0
+func isIPv4NotInv6(ipString string) bool {
+	for i := 0; i < len(ipString); i++ {
```
jenshu, line 128: would an easier check be: `strings.Contains(ipString, ":")` ?

projects/gloo/pkg/translator/utils.go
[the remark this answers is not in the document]
jenshu: ooh actually i like your wording in the comments above "non-padded pure ipv4 address"
`isPureIpv4Address`?

projects/gloo/pkg/translator/utils.go
```diff
@@ -99,20 +99,39 @@ type typedConfigObject interface {
...
-		// Therefore, this is the case where the bindIP represents an IPv4 address and we need to convert
-		// it into an IPv4-mapped IPv6 address
-		return fmt.Sprintf("::ffff:%s", bindAddress), nil
+		// so this is not an acceptable ipv4
+		return false, false, nil
 	}
+	if isIPv4NotInv6(bindAddress) {
+		return true, true, nil
+	}
+	return true, false, nil
```
jenshu, line 119:
```suggestion
	return true, isIPv4NotInv6(bindAddress), nil
```

projects/gloo/pkg/translator/utils.go
nfuden: I think so but I haven't looked at all the possible formats but this is copied from the net.parseip code so seems safer

projects/gloo/pkg/translator/utils.go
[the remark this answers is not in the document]
nfuden: Ooh I like that!

projects/gloo/pkg/translator/utils_test.go
```diff
@@ -28,21 +28,22 @@ var _ = Describe("Plugin", func() {
 	})
 
 	DescribeTable(
-		"GetIpv6Address",
-		func(address, expectedAddress string, expectedErr error) {
-			ipv6Address, err := translator.GetIpv6Address(address)
+		"IsIpv4Address",
+		func(address string, expectedIpv4 bool, expectedErr error) {
+			_, isIpv4Address, err := translator.IsIpv4Address(address)
```
jenshu, line 33: should we test all the return values `validIpv4, strictIPv4, err`?

nfuden: Never hurts. Might also donsome docs on it or move it out from this package

nfuden:
I didnt find any places that would make sense to document. It may be something in the future where we expose an option to reject IPv4 traffic (ipcompat) but that seems out of scope for this.

nfuden:
/kick CSRF defined on listener defined on listener

jenshu (approved):

PR #6864, opened 2022-08-04
ci: added hack script to compress changelogs by minor version

gunnar-solo:
When run, the script produces a changelog directory resembling:

Screen Shot 2022-08-04 at 8 15 28 AM

This commit was made _without_ running, as the number of changes it generates, otherwise, is overwhelming to review

### DIFF
--- a/hack/squash-changelogs.sh
+++ b/hack/squash-changelogs.sh
@@ -1,0 +1,26 @@
+#!/bin/bash -e
+CURRENT_MAJOR_VERSION=1         # major semvar version of supported product
+CURRENT_MINOR_VERSION=13        # minor semvar version of supported product
+SUPPORTED_VERSIONS=4            # how many versions of product are in active support
+LEGACY_CODE_FOLDER="_archive"   # intended destination of legacy changelogs
+
+cd ../changelog
+
+for folder in v*.*.*; do
+    folder="${folder#?}"
+    semver=( ${folder//./ } )
+    major="${semver[0]}"
+    minor="${semver[1]}"
+
+    # active version, supported version, legacy version
+    if   [ $CURRENT_MAJOR_VERSION = $major ] && [ $CURRENT_MINOR_VERSION = $minor ]; then
+        continue
+    elif [ $CURRENT_MAJOR_VERSION = $major ] && [ $(($minor+$SUPPORTED_VERSIONS+1)) -gt $CURRENT_MINOR_VERSION ]; then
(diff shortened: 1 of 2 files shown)


gunnar-solo:
/kick [[Fail] CSRF defined on listener only on listener [It] should succeed with allowed origin](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=b16f0bd0-72ca-44f9-a3ec-54d0c9fe6e09)

hack/squash-changelogs.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -e
+CURRENT_MAJOR_VERSION=1         # major semvar version of supported product
```
nfuden, line 2: non-blocking: Do you think we could / should use the https://github.com/solo-io/gloo/blob/master/hack/make-changelog-entry.sh#L19 to get the most recent version so that we dont have to manually update these?

gunnar-solo: Nope!  I do not think we should use that.  Re:  I'd prefer the script work the same for all branches it's committed to.

`master`
```bash
CURRENT_MAJOR_VERSION=1
CURRENT_MINOR_VERSION=13
SUPPORTED_VERSIONS=4
```

`v1.12.x`
```bash
CURRENT_MAJOR_VERSION=1
CURRENT_MINOR_VERSION=12
SUPPORTED_VERSIONS=3
```

`v1.11.x`
```bash
CURRENT_MAJOR_VERSION=1
CURRENT_MINOR_VERSION=11
SUPPORTED_VERSIONS=2
```

`v1.10.x`
```bash
CURRENT_MAJOR_VERSION=1
CURRENT_MINOR_VERSION=10
SUPPORTED_VERSIONS=1
```

`v1.9.x`
```bash
CURRENT_MAJOR_VERSION=1
CURRENT_MINOR_VERSION=9
SUPPORTED_VERSIONS=0
```

gunnar-solo: willing to be contradicted, here.  This is just my "it works the same in all cases" logic

gunnar-solo:
/kick (Step #10 - "test": [Fail] Gateway in memory tcp gateway ssl [It] should work with ssl 
)[https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=0d3bb9f8-76df-4393-a1fe-23bf960657da]

nfuden (approved):

PR #6875, opened 2022-08-04
Revert "11 | build(changelog): archived changelogs from unsupported v…

nfuden:
Reverting as breaks changelogbot


jbohanon (approved):
Reverts addition of script as well; deviation from master reversion

PR #6876, opened 2022-08-04
Revert "12 | build(changelog): archived changelogs from unsupported v…

nfuden:
revert to unblock changelog bot


jbohanon (approved):
Reverts addition of script as well; deviation from master reversion

PR #6886, opened 2022-08-05
Add internal config field for fail-on-redirect

jbohanon:
When adding the config for ext auth server fail-on-redirect to API, I missed the internal message which also needs the field added. This PR corrects that issue.

Original PR: https://github.com/solo-io/gloo/pull/6844


jbohanon:
/kick [should work with ssl invalid resource version](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=bb30b7a5-3498-46a2-9911-f81383a4a3de)

nfuden (approved):
👍

PR #6891, opened 2022-08-05
Back1.12/ipv4 only compat

nfuden:
Description
Set ipv4Compat on a listener based on the state of the listener bind address

Context
This value was always set to true in earlier versions of Gloo.

In Envoy 1.22, the behavior changed such that this field could only be set to true if the listener address was an ipv4 compatible address.

The previous solution was to identify any ipv4 addresses (which would be rejected in Envoy 1.22) and convert them to ipv6 addresses. However, if you are running a kube cluster with only ipv4 support this causes issues.

The new solution is to identify ipv4 addresses and turn this ipv4Compat field to false so that it will not crash. Currently this solution will continue to allow ipv4 anywhere possible.


jenshu (approved):

nfuden:
/kick fails in the same way: CSRF defined on listener defined on listener with shadow mode config [It]

nfuden:
/kick  CSRF defined on listener defined on listener we need to fix CSRF

PR #6895, opened 2022-08-05
Set MaxCallSendMsgSize on proxy debug client

Bslabe123:
# Description

Proxy debug client now applies the CallOption grpc.MaxCallRecvMsgSize to GetProxies

# Context

Some customers have proxies that exceed the 100Mb default size limit which would cause 'kubectl get proxies' to error out. We fixed this issue by setting the size limit to the maximum possible (MaxInt32).

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/pull/6627


changelog/v1.13.0-beta3/proxy-max-size-checking.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/pull/6627
+    description: Added a check for proxy size before returning via grpc
```
sam-heilbron, line 4: What was the default message size that was being used before this change? How would this error manifest to a user?

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +162,14 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
 	return list, nil
 }
 func getProxiesFromGrpc(name string, namespace string, opts *options.Options, proxyEndpointPort string) (gloov1.ProxyList, error) {
+	settingsClient := helpers.MustNamespacedSettingsClient(opts.Top.Ctx, opts.Metadata.GetNamespace())
+	settings, err := settingsClient.Read(opts.Metadata.GetNamespace(), defaults.SettingsName, clients.ReadOpts{Ctx: opts.Top.Ctx})
+
+	options := []grpc.CallOption{}
+	if settings.GetGateway().GetValidation() != nil {
```
sam-heilbron, line 169: Is there a situation where a user would want to get the Proxy but not have validation enabled? If so, how should we handle that? If not, can we outline in a comment why we believe using this setting is ok?

elcasteel: This change is basically a repeat of this work: https://github.com/solo-io/gloo/pull/6627 and this will only come up for users that have very large proxies (pretty sure there are only a couple that need to override this for validation). I think we strongly encourage always using validation in production.
The fix would be either a true breaking change to the settings API to move the existing field out of Validation or add another field and then have two fields that mean basically the same thing. 
We won't actually need both fields set at the same time because we don't send proxies over grpc for validation when they are persisted in memory (and glooctl needs to get them over grpc). The biggest problem would be user confusion about where to set the value.

Bslabe123: I could make the changelog a bit more clear about where to set the value.

changelog/v1.13.0-beta3/proxy-max-size-checking.yaml
Bslabe123: The default size is 1.5Mb, just added a comment to outline that! Also the error would be produced by the function consuming these options `r, err := pxClient.GetProxies(opts.Top.Ctx, &debug.ProxyEndpointRequest{
				Name:      name,
				Namespace: opts.Metadata.GetNamespace(),
			}, options...)`

projects/gloo/cli/pkg/common/get.go
sam-heilbron: Yeah I figured it would only apply to users who have configured this. Do you think it makes sense to just set it to the max value for the CLI or is there a downside to that?

changelog/v1.13.0-beta3/proxy-max-size-checking.yaml
sam-heilbron: Excellent! I think this type of information would be valabue to include in the PR context to outline what exactly we're changing. I could see value in the changelog to call out the 100MB default

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +162,17 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
 	return list, nil
 }
 func getProxiesFromGrpc(name string, namespace string, opts *options.Options, proxyEndpointPort string) (gloov1.ProxyList, error) {
+	settingsClient := helpers.MustNamespacedSettingsClient(opts.Top.Ctx, opts.Metadata.GetNamespace())
+	settings, err := settingsClient.Read(opts.Metadata.GetNamespace(), defaults.SettingsName, clients.ReadOpts{Ctx: opts.Top.Ctx})
```
sam-heilbron, line 166: Should we do something with this error?

projects/gloo/cli/pkg/common/get.go
elcasteel: Yeah, actually it seems reasonable to just set this value to max_int. The go grpc implementation doesn't do anything silly like always allocating the max message size. Also the default maximum size for sent messages is max_int (which we don't override) so it's fine if we don't make the same change in the server.

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +162,17 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
...
+
+	// Translates/Stores all grpc options defined in settings.Gateway
+	options := []grpc.CallOption{}
+	if settings.GetGateway().GetValidation() != nil {
+		//Both of these default to 100MB (100 * 1024 * 1024)
+		options = append(options, grpc.MaxCallSendMsgSize(int(settings.GetGateway().GetValidation().GetValidationServerGrpcMaxSizeBytes().GetValue())))
```
elcasteel, line 172: We only need to set MaxCallRecvMsgSize because the size of the messages we send from this client is small.

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +162,17 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
...
+		options = append(options, grpc.MaxCallRecvMsgSize(int(settings.GetGateway().GetValidation().GetValidationServerGrpcMaxSizeBytes().GetValue())))
```
nfuden, line 173: Since this is just cli and a user could stop it whenever they want how do we feel about just well slapping max int like maxcallsend defaults to in the go grpc client.  https://github.com/grpc/grpc-go/blob/master/server.go#L59
The maxrecv size seems to have been defaulted to 4mb to protect low bandwidth  https://github.com/grpc/grpc-java/issues/1676

Bslabe123: Would there be any performance drawbacks to doing it this way?

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +163,21 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
 	return list, nil
 }
 func getProxiesFromGrpc(name string, namespace string, opts *options.Options, proxyEndpointPort string) (gloov1.ProxyList, error) {
+	settingsClient := helpers.MustNamespacedSettingsClient(opts.Top.Ctx, opts.Metadata.GetNamespace())
```
sam-heilbron, line 166: I had thought the change was to just use the max size and not read the settings at all. Is there a particular reason to maintain the settings definition since this is a CLI call?

projects/gloo/cli/pkg/common/get.go
sam-heilbron: Bump on this, I had thought the outcome was to just use maxInt regardless of the settings

changelog/v1.13.0-beta4/proxy-max-size-checking.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/pull/6627
+    description: Added logic to check proxy size against ValidationServerGrpcMaxSizeBytes before returning via grpc
+    resolvesIssue: true
```
sam-heilbron, line 5: Do we plan on backporting this fix?

changelog/v1.13.0-beta4/proxy-max-size-checking.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/pull/6627
+    description: Added logic to check proxy size against ValidationServerGrpcMaxSizeBytes before returning via grpc
```
sam-heilbron, line 4: Let's call out specifically the change in the user experience. I think we should write these from the perspective of a user unfamiliar with the code changes, who is interested in understanding what is changed. in this case, I think we need to call out the cli command that is used, the previous behavior and the new behavior.

changelog/v1.13.0-beta4/proxy-max-size-checking.yaml
elcasteel: Would be good to have in 1.12

Bslabe123: Can definitely backport!

changelog/v1.13.0-beta4/proxy-max-size-checking.yaml
Bslabe123: Can do! Could you link to where this should be called out?

projects/gloo/cli/pkg/common/get.go
Bslabe123: Missed this, fixing now!

projects/gloo/cli/pkg/common/get.go
Bslabe123: Missed that nit above, its now permanently fixed to MaxInt

changelog/v1.13.0-beta4/proxy-max-size-checking.yaml
sam-heilbron: If we're backporting, shouldn't this be resolvesIssue=false?

projects/gloo/cli/pkg/common/get.go
```diff
@@ -162,6 +163,12 @@ func getProxiesFromK8s(name string, opts *options.Options) (gloov1.ProxyList, er
 	return list, nil
 }
 func getProxiesFromGrpc(name string, namespace string, opts *options.Options, proxyEndpointPort string) (gloov1.ProxyList, error) {
+
+	// Translates/Stores all grpc options defined in settings.Gateway
```
sam-heilbron, line 167: Is this comment accurate?

We could add a comment about how Proxies get large, and can exceed the default grpc msg size (100mb) and we set it to the max size to avoid that.

sam-heilbron (approved):
Nice! Love the extra comments in the code for clarification around why we set the MsgSize to the maxInt, and the changelog is clear what the user impact is

Issue #6782, opened 2022-07-21
AutoMapFromMetadata feature in AuthConfig not working


pszeto:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

When AutoMapFromMetadata is configured, dynamic metadata values in the specified namespace shouldl be mapped to query params on requests made to the OIDC auth endpoint and the OIDC token endpoint. However, I'm not seeing those query params added.

### Steps to reproduce the bug

1.  Create AuthConfig:
```
apiVersion: enterprise.gloo.solo.io/v1
kind: AuthConfig
metadata:
  name: oauth-httpbin
  namespace: gloo-system
spec:
  configs:
  - oauth2:
      oidcAuthorizationCode:
        appUrl: https://httpbin.mysolo.io/
        authEndpointQueryParams:
          authEndPointKey: authEndpointValue
        autoMapFromMetadata:
          namespace: solo_authconfig_oidc
        callbackPath: /callback
        clientId: <redacted>
        clientSecretRef:
          name: oauth
          namespace: gloo-system
        headers:
          idTokenHeader: jwt
        issuerUrl: http://35.243.161.53:8080/auth/realms/master/
        logoutPath: /logout
        scopes:
        - email
        session:
          cookieOptions:
            domain: mysolo.io
          failOnFetchFailure: true
          redis:
            cookieName: session
            options:
              host: redis.gloo-system.svc.cluster.local:6379
        tokenEndpointQueryParams:
          tokenEndpointKey: tokenEndpointValue
status:
  statuses:
    gloo-system:
      reportedBy: gloo
      state: 1
```
2. Create VS. 
```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: httpbin-vs
  namespace: gloo-system
spec:
  sslConfig:
    secretRef:
      name: upstream-tls
      namespace: gloo-system
  virtualHost:
    domains:
    - httpbin.mysolo.io
    options:
      extauth:
        configRef:
          name: oauth-httpbin
          namespace: gloo-system
      stagedTransformations:
        early:
          requestTransforms:
          - requestTransformation:
              transformationTemplate:
                dynamicMetadataValues:
                - key: tenantid
                  metadataNamespace: solo_authconfig_oidc
                  value:
                    text: mytenantid
                - key: spaceid
                  metadataNamespace: solo_authconfig_oidc
                  value:
                    text: myspaceid
                extractors:
                  spaceid:
                    header: :path
[...]

PR #6896, opened 2022-08-05
AutoMapForMetaData Fixes

ianmacclancy:
# Description

Updates to config generation to allow access ext-auth access to specified namespaces
# Context

https://github.com/solo-io/gloo/issues/6782


projects/gloo/pkg/plugins/extauth/config_generator.go
```diff
@@ -214,7 +216,58 @@ func BuildStagedHttpFilters(configurationGenerator func() ([]*envoyauth.ExtAuthz
 	return filters, nil
 }
 
-func GenerateEnvoyConfigForFilter(settings *extauthv1.Settings, upstreams v1.UpstreamList) (*envoyauth.ExtAuthz, error) {
+func getMetaDataNamespacesFromTransforms(listener *v1.HttpListener) []string {
```
jackstine, line 219: nit: comment please

projects/gloo/pkg/plugins/extauth/config_generator.go
```diff
@@ -214,7 +216,58 @@ func BuildStagedHttpFilters(configurationGenerator func() ([]*envoyauth.ExtAuthz
...
+						namespacesSet[namespace] = voidMember
+					}
+				}
+			}
+		}
+	}
+	return namespacesSet
+}
+
+func GenerateEnvoyConfigForFilter(settings *extauthv1.Settings, listener *v1.HttpListener, upstreams v1.UpstreamList) (*envoyauth.ExtAuthz, error) {
```
jackstine, line 270: would this be a breaking change?

ianmacclancy: Its additional information that was already available. It is now being passed where this method is called. So there are required changes in solo-projects

jbohanon: The concern I'd have is that since this is OSS code and an exported function, community users could have calls to this function that would break in a dep-bump.

jackstine: Yeah I think it would be good to change the changlog to `BREAKING_CHANGE `

projects/gloo/pkg/plugins/extauth/config_generator.go
```diff
@@ -214,7 +216,58 @@ func BuildStagedHttpFilters(configurationGenerator func() ([]*envoyauth.ExtAuthz
...
+			}
+		}
+	}
+	namespaces := make([]string, len(namespacesSet))
+	i := 0
+	for namespace, _ := range namespacesSet {
+		namespaces[i] = namespace
+		i++
+	}
+	return namespaces
```
jenshu, line 237: this section can be shortened/simplified with the use of [sets.String](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String) to build the set, and then
[UnsortedList](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.UnsortedList) (or List if you need it sorted) to get the string slice

jbohanon (commented):
Just nits, happy to approve

projects/gloo/pkg/plugins/extauth/config_generator.go
jackstine: We have decided that the additional changes will be only written to enterprise solution, and use a wrapper to set the `MetadataContextNamespaces` when calling `GenerateEnvoyConfigForFilter`

ianmacclancy: Based on the discussion on this comment the code for adding metaData to namespace will be moved to solo projects in this PR - https://github.com/solo-io/solo-projects/pull/3954 and I will close out this PR

Issue #6538, opened 2022-06-07
Istio Compat: Routes to Services In Workspace with Service Isolation Enabled Fail Without Workaround


bgottfried91:
### Gloo Edge Version

1.12.x (beta)

### Kubernetes Version

1.23.x

### Describe the bug

#6195 and #6196 added support for full compatibility between a GE Gateway Proxy and services in the Mesh, but it turns out the support only works for cross-workspace traffic (e.g. from a `gateways` workspace to the `bookinfo` workspace) for workspaces that do not have `Service Isolation` enabled. If you do enable `Service Isolation`, any GE route that does not do a `hostRewrite`* will fail when attempting to access a service in another workspace that has `Service Isolation` enabled.

*You can work around this by setting a `hostRewrite` of `servicename.namespace` for each route. See additional context for why this isn't a good long-term solution.

### Steps to reproduce the bug

1. Deploy Gloo Mesh and register a cluster with it.
2. Deploy Gloo Edge on the target cluster. Make sure to set `global.istioIntegration.enableIstioSidecarOnGateway` in the Helm values file and ensure that the Gloo Edge gateway proxy has a sidecar.
3. Deploy an application on the target cluster that Edge will target (e.g. `bookinfo` )
4. Create two workspaces on the registered cluster, one of which covers the Gloo Edge namespace and the other which covers the application namespace. Ensure that `Service Isolation` is enabled for the application workspace.
5. Create a Virtual Service for Gloo Edge that targets an application in the application namespace (e.g. `productpage`)
6. Try to reach the application through that VS on the Gloo Edge proxy URL. It should fail with an error `upstream connect error or disconnect/reset before headers. reset reason: connection termination`

### Expected Behavior

When creating a route for a service that is in the Mesh and Gloo Edge is configured for compatibility with Gloo Mesh, you should not have to do any special configuration for the routes exposed by Gloo Edge beyond pointing them to the service upstreams.

### Additional Context

[...]

PR #6817, opened 2022-07-27
6470 hcm internal address config

ashishb-solo:
# Description

The new feature exposes `extensions.filters.network.http_connection_manager.v3.HttpConnectionManager.InternalAddressConfig`

# Context

Users requested this feature in the linked issue

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6470

PR #6897, opened 2022-08-05
Rewrite hosts in mesh

elcasteel:
# Description
This change adds a new route action plugin that runs when istio integration is enabled. It looks at the destination for routes and if they are kubernetes services, rewrites the host to be `servicename.namespace` and sets the x-forwarded-host header.

It also bumps the dependency for envoyproxy go-control-plane which also resulted in bumping a version for `protoc-gen-validate` and changed a lot of generated files. I might separate these two changes to make the PRs easier to read.

Future change: Set of istio integration tests.

This change also bumps the version of go-control-plane which resulted in an update to protobuf which resulted in small changes to a lot of generated files. 

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6538


projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -0,0 +1,99 @@
...
+			// kube upstreams are the only things that we need to access within the mesh
+			return "", nil
+		}
+		name = kubeUs.Kube.GetServiceName()
+		namespace = kubeUs.Kube.GetServiceNamespace()
+	} else if d, ok := dest.Single.GetDestinationType().(*v1.Destination_Kube); ok {
+		name = d.Kube.GetRef().GetName()
+		namespace = d.Kube.GetRef().GetNamespace()
+	}
+	return fmt.Sprintf("%s.%s", name, namespace), nil
```
jenshu, line 98: if name / namespace were never set, this would return "." (rather than "")

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -0,0 +1,99 @@
...
+
+// When istio integration is enabled, we need to access k8s services using a host that istio will recognize (servicename.namespace)
+// We do this by adding a hostRewrite for kube destinations kube upstreams. In case the upstream also wants the original host,
+// we also set x-forwarded-host
+// We ignore other destinations and routes that already have a rewrite applied and return an error if we can't look up an Upstream.
+func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	//check if this is a kubernetes destination
+	dest, ok := in.GetRouteAction().GetDestination().(*v1.RouteAction_Single)
+	if !ok {
+		// TODO: possibly do something with other types of destinations
```
jenshu, line 53: should we handle this TODO?

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -0,0 +1,99 @@
...
+	ExtensionName = "istio_integration"
+)
+
+// Handles transformations required to integrate with Istio
+type plugin struct {
+	usClient v1.UpstreamClient
+}
+
+func NewPlugin(ctx context.Context, upstreams factory.ResourceClientFactory) *plugin {
+	// TODO: handle situation when we can't create an upstream client
```
jenshu, line 28: should we handle this TODO?

projects/gloo/pkg/plugins/istio_integration/plugin.go
elcasteel: We've already returned in the cases where name/namespace aren't set (lines 85, 90)

projects/gloo/pkg/plugins/istio_integration/plugin.go
elcasteel: We only need to handle k8s destinations for the current use case and I don't know what the correct rewrite would be for other types of destinations. I can clarify the comment and remove the TODO part

projects/gloo/pkg/plugins/istio_integration/plugin.go
jenshu: but that's only if it's of the type Destination_Upstream.
i think for any other types (i.e. consul) we should fall back and return ""

also, for the upstream/kube cases above, do we need to check that name/namespace are not empty? if they're empty/invalid we should probably just skip them or print an error

elcasteel: yeah good point on the other types, I pushed up a check for unset name/namespace. There isn't a meaningful validity check for non-empty name/namespace values so we have to just go with whatever values are there.

projects/gloo/pkg/plugins/istio_integration/plugin.go
elcasteel: replaced with a log

projects/ingress/pkg/api/v1/translator_snapshot_emitter.sk.go
```diff
@@ -23,25 +23,25 @@ import (
...
 var (
 	// Deprecated. See mTranslatorResourcesIn
-	mTranslatorSnapshotIn = stats.Int64("translator.ingress.solo.io/emitter/snap_in", "Deprecated. Use translator.ingress.solo.io/emitter/resources_in. The number of snapshots in", "1")
+	mTranslatorSnapshotIn = stats.Int64("translator_ingress_solo_io/emitter/snap_in", "Deprecated. Use translator_ingress_solo_io/emitter/resources_in. The number of snapshots in", "1")
 
 	// metrics for emitter
-	mTranslatorResourcesIn    = stats.Int64("translator.ingress.solo.io/emitter/resources_in", "The number of resource lists received on open watch channels", "1")
-	mTranslatorSnapshotOut    = stats.Int64("translator.ingress.solo.io/emitter/snap_out", "The number of snapshots out", "1")
-	mTranslatorSnapshotMissed = stats.Int64("translator.ingress.solo.io/emitter/snap_missed", "The number of snapshots missed", "1")
+	mTranslatorResourcesIn    = stats.Int64("translator_ingress_solo_io/emitter/resources_in", "The number of resource lists received on open watch channels", "1")
```
sam-heilbron, line 29: If these are required, I think we should call out loudly in the changelog this change. The resource metrics are heavily used by clients to track that Gloo is behaving appropriately, and they'll need to update their dashboards or raise questions about why the counts have gone to 0

projects/gloo/pkg/plugins/hcm/testing/expected_hcm_fields.json
```diff
@@ -45,5 +45,7 @@
     "StripPortMode",
     "StreamErrorOnInvalidHttpMessage",
     "PathNormalizationOptions",
-    "StripTrailingHostDot"
+    "StripTrailingHostDot",
+    "ProxyStatusConfig",
+    "TypedHeaderValidationConfig"
```
sam-heilbron, line 50: I think this test was intended to identify when new fields are added to the HCM, as a way of proactively also exposing them in our API. I don't think it makes sense to do as part of this work (scope creep), but just wanted to call out that we'll probably want to do https://github.com/solo-io/gloo/pull/6817/files and expose these other fields as a fast follow

sam-heilbron (commented):
Looking good! Small questions and one comment about changing stats

projects/gloo/pkg/plugins/registry/registry.go
```diff
@@ -110,7 +113,11 @@ func Plugins(opts bootstrap.Opts) []plugins.Plugin {
 	if opts.Consul.ConsulWatcher != nil {
 		glooPlugins = append(glooPlugins, consul.NewPlugin(opts.Consul.ConsulWatcher, consul.NewConsulDnsResolver(opts.Consul.DnsServer), opts.Consul.DnsPollingInterval))
 	}
-
+	lookupResult, found := os.LookupEnv("ENABLE_ISTIO_INTEGRATION")
```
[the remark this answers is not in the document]
elcasteel: The environment variable is set via helm/values-override file (introduced here: https://github.com/solo-io/gloo/pull/6312/files#diff-fb80295aabf60cfc9f228596c24465c7c0e065906264f26f6320d6fb3b33a605R607)

sam-heilbron (dismissed):

sam-heilbron (approved):

elcasteel:
/kick build-bot didn't restart after changelog-bot's merge

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -0,0 +1,103 @@
...
+// we also set x-forwarded-host
+// We ignore other destinations and routes that already have a rewrite applied and return an error if we can't look up an Upstream.
+func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	//check if this is a kubernetes destination
+	dest, ok := in.GetRouteAction().GetDestination().(*v1.RouteAction_Single)
+	if !ok {
+		// TODO: possibly do something with other types of destinations
+		return nil
+	}
+	hostInMesh, err := GetHostFromDestination(dest, p.usClient)
```
murphye, line 61: Possible Bug

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -0,0 +1,103 @@
...
+		us, err := usClient.Read(single.Upstream.GetNamespace(), single.Upstream.GetName(), clients.ReadOpts{})
+		if err != nil {
+			return "", err
+		}
+		kubeUs, ok := us.GetUpstreamType().(*v1.Upstream_Kube)
+		if !ok {
+			// kube upstreams are the only things that we need to access within the mesh
+			return "", nil
+		}
+		name = kubeUs.Kube.GetServiceName()
```
murphye, line 97: Possible Bug

Issue #6799, opened 2022-07-25
VirtualService sslConfig pointing to a SDS not generating correct envoy config


pszeto:
### Gloo Edge Version

1.11.x (latest stable)

### Kubernetes Version

1.22.x

### Describe the bug

Configuring a Virtual Service to use an SDS to get certificates not generating correct envoy config.

### Steps to reproduce the bug

1. Create the following VS.

```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: vs-direct
  namespace: gloo-system
spec:
  sslConfig:
    sds:
      targetUri: unix:/tmp/uds_path
      certificatesSecretName: upstream-tls
  virtualHost:
    domains:
      - '*'
    routes:
      - matchers:
         - prefix: /
        directResponseAction:
          status: 200
          body: "Hello, world!"
```

The envoy config has:

```
transport_socket:
                  name: envoy.transport_sockets.tls
                  typed_config:
                    '@type': >-
                      type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext
                    common_tls_context:
                      tls_params: {}
                      alpn_protocols:
                        - h2
                        - http/1.1
                      tls_certificate_sds_secret_configs:
                        - name: upstream-tls
                          sds_config:
                            api_config_source:
                              api_type: GRPC
                              grpc_services:
                                - envoy_grpc:
                                    cluster_name: gateway_proxy_sds
                              transport_api_version: V3
                            resource_api_version: V3
```

### Expected Behavior

Expected the envoy config to look similiar :

https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret.html

```
      typed_config:
        "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
        common_tls_context:
          tls_certificate_sds_secret_configs:
          - name: client_cert
            sds_config:
              resource_api_version: V3
              api_config_source:
                api_type: GRPC
                transport_api_version: V3
                grpc_services:
                  google_grpc:
                    target_uri: unix:/tmp/uds_path
```

nfuden:
See linked zendesk for more info, we this is more about secret retrieval from vault which then leads to this

kdorosh:
another option @pszeto to unblock on the current version is to define an upstream with the pipe location to the SDS server, and then configure SDS to use the generated envoy cluster (with that pipe)

pszeto:
@kdorosh 
[...]

PR #6917, opened 2022-08-09
buildSds() creates GoogleGrpc TargetSpecifier if TargetUri is not empty

bewebi:
# Description

- When generating SDS config, if `TargetUri` is specified and `ClusterName` is not specified, create a `GrpcService` with a `GoogleGrpc` `TargetSpecifier`
  - Use the `CertificateSecretName` or `ValidationContextName`, depending on which we are generating config for, as the `StatPrefix` 

# Context

This has always been expected behavior, and the fact that we do not create the expected `GoogleGrpc` `TargetSpecifier` until now is considered a bug

In theory `TargetUri` and `ClusterName` are mutually exclusive and could/should be part of the `SdsBuilder` `oneof` in the `SDSConfig` proto
In order to prevent a breaking change we opt not to modify the proto and prioritize `ClusterName` if bother are specified in order to maintain previous behavior of ignoring `TargetUri` in that case

I've manually verified that the expected config is generated given the VS from the original [issue](https://github.com/solo-io/gloo/issues/6799).

# Checklist:

### DIFF
--- a/projects/gloo/pkg/utils/ssl_test.go
+++ b/projects/gloo/pkg/utils/ssl_test.go
@@ -354,7 +354,7 @@ var _ = Describe("Ssl", func() {
 			cert := c.TlsCertificateSdsSecretConfigs[0]
 			Expect(vctx.Name).To(Equal("ValidationContextName"))
 			Expect(cert.Name).To(Equal("CertificatesSecretName"))
-			// If they are no equivalent, it means that any serialization is different.
+			// If they are not equivalent, it means that any serialization is different.
 			// see here: https://github.com/envoyproxy/go-control-plane/pull/158
 			// and here: https://github.com/envoyproxy/envoy/pull/6241
 			// this may lead to envoy updates being too frequent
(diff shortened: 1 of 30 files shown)


projects/gloo/pkg/utils/ssl_test.go
```diff
@@ -400,6 +398,32 @@ var _ = Describe("Ssl", func() {
 
 		})
 
+		When("TargetUri is specified", func() {
+			BeforeEach(func() {
+				sdsConfig.TargetUri = "targetUri"
+			})
+
+			It("should have a sds setup with a default cluster name", func() {
```
kdorosh, line 406: nit, name needs to be updated

kdorosh (commented):
looks reasonable to me!

projects/gloo/pkg/utils/ssl_test.go
```diff
@@ -400,6 +398,69 @@ var _ = Describe("Ssl", func() {
...
 		})
 
+		Context("TargetUri is specified", func() {
+			BeforeEach(func() {
+				sdsConfig.TargetUri = "targetUri"
+			})
+
+			When("only TargetUri is specified", func() {
+
+				FIt("should have a sds setup with a GoogleGrpc TargetSpecifier with the expected TargetUri", func() {
```
kdorosh, line 408: unfocus

bewebi: 🤦‍♂️

nfuden (approved):

nfuden (approved):

PR #6791, opened 2022-07-21
Remove Error from Translation Interfaces (Gloo Edge HA)

sam-heilbron:
# Blocked By
-  [x] https://github.com/solo-io/solo-projects/pull/3905. Since this breaks a number of pulbic methods in Gloo OSS, the linked Enterprise PR demonstrates how to incorporate these changes.

# Description

Gloo translation loop always completes and updates internal xDS snapshot

# Context

There are a lot of affected files here, but the changes are fairly trivial, and fall into a couple of categories. 

## Plugin.Init does not error
Plugin.Init is called [every envoy translation loop](https://github.com/solo-io/gloo/blob/7d372917821cb06d1b77599ea78942c98d8c2ba3/projects/gloo/pkg/translator/translator.go#L110). If initialization returns an error, we would never process the configuration and generate an xDS snapshot. There are limited cases that needed to be updated (most returned nil). I opted to either log a warning or return the error during construction of the plugin.

## Translate does not error
The only error that could occur was from plugin initialization (solved above). With that interface updated to never return an error, the Translator could be similarly updated.

## TranslatorSyncerExtensions do not error
After translation of envoy resources, we execute translator syncer extensions, to update the xDS cache for custom syncers (in our case, extauth and ratelimit and other enterprise functionality which relies on xds). We similarly update this interface to not return an error. Instead, the assumption is that errors which are encountered are placed on resource reports.

## TranslatorSyncerExtensionFactory do not error
The factory used to generate TranslatorSyncerExtensions has an error in the return signature, but no instances were using this, and the setup loop would just log this error and do nothing with it. I removed the error as it was unnecessary and misleading.

## TranslatorSyncerExtension reconciliation removed from OSS
Gloo can be run with default extensions (used by OSS to warn about enterprise config) or custom extensions (supplied by EE). Gloo OSS would preform reconciliation to determine if an enterprise extension would override a open source extension. This relationship felt backwards, and I opted to move that reconciliation into the enterprise repo.

Now, the open source repo just accepts what is provided as that state of the world. WHen running open source gloo, it provides the open source extensions explicitly.

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
[...]

PR #6919, opened 2022-08-10
Restore PluginRegistryFactory

elcasteel:
# Description
Use factories to create plugin registries and translator instances on each run of translation. We used this pattern until this change: https://github.com/solo-io/gloo/pull/6791/files#diff-7274e84d88447a5ce54e45c3e8f869dbfb3d2c9cb95198a0791fad46282cd6f5

# Context
Users have seen plugins crashing due to simultaneous writes to maps.

# Checklist:

### DIFF
--- a/projects/gloo/pkg/translator/translator.go
+++ b/projects/gloo/pkg/translator/translator.go
@@ -38,7 +38,7 @@ type Translator interface {
 }
 
 var (
-	_ Translator = new(translatorInstance)
+	_ Translator = new(translatorFactory)
 )
 
 // translatorInstance is the implementation for a Translator used during Gloo translation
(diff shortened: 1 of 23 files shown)


projects/gloo/pkg/translator/translator.go
```diff
@@ -49,20 +49,41 @@ type translatorInstance struct {
 	listenerTranslatorFactory *ListenerSubsystemTranslatorFactory
 }
 
+type translatorFactory struct {
+	pluginRegistryFactory plugins.PluginRegistryFactory
```
sam-heilbron, line 53: One idea: What if a translator factory is instantiated with a pluginRegistryFactory (so that the registry it creates is unique) and then it creates the Registry and stores it. Then all future translation runs use the same registry

elcasteel: What if the validation server is handling multiple requests at once and has multiple translation loops running as a result? Would it be possible for updates from one run to interfere with the results from another run? What if we have an update from an invalid resource that is ultimately rejected? I kind of think that caches that persist across multiple runs will need to be handled differently by validation and I want to put some time into doing that properly.

sam-heilbron: That's fair!

nfuden (approved):

Issue #6472, opened 2022-05-18
Create unique virtual services in regression tests


elcasteel:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

In the gateway regression tests, we have a lot of tests that create virtual services and then verify that curl works in various circumstances. 
We have a helper function to create a virtual service that uses a hard coded name and the domain "*". https://github.com/solo-io/gloo/blob/master/test/kube2e/gateway/gateway_test.go#L2630

The hardcoded name and domain mean that only one virtual service at a time can be created with this helper and therefore tests need to defensively clean up resources and wait for resources from previous tests to be deleted or the tests are flaky. 

We should replace the hardcoded names and domains with uniquely generated ones. 

### Steps to reproduce the bug

Run CI. 

### Expected Behavior

CI should be deterministic.

Nadine2016:
Some logs where I could not pass this test for my doc PR https://github.com/solo-io/gloo/runs/6756308289?check_suite_focus=true. This was the PR that I tried to merge in https://github.com/solo-io/gloo/pull/6522

ianmacclancy:
Hitting this on - https://github.com/solo-io/gloo/pull/6653

ianmacclancy:
Hitting this on - https://github.com/solo-io/gloo/runs/7235963658?check_suite_focus=true#step:11:163

elcasteel:
saw this here: https://github.com/solo-io/gloo/runs/7394019423?check_suite_focus=true

elcasteel:
happens  frequently still: https://github.com/solo-io/gloo/runs/7775959992?check_suite_focus=true

PR #6926, opened 2022-08-10
Introduce Leader Election for Gloo 

sam-heilbron:
# Description

Introduce leader election to Gloo component

# Context
Introduce leader election if using Kubernetes as the source of truth for resources.

## Solution
Initially I place the leader election code inside the Gloo SetupFunc, which is executed each time Settings change. The effect was that each time the SetupFunc was called, the election process was cancelled (since we use a new context for each setup), a new leader was elected, and inconsistent behavior was observed.

I moved this election logic to the main setup portion so that it happens once during startup, and if leadership is lost, we kill the application.

## Technical Debt
This is called out loudly in a code comment in https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/api/converters/kube/artifact_converter.go, where the debt is incurred.

We need to ignore the configmap (or whatever kube resource maintains the state of the leader) during translation. Since it is updated on an interval (2 seconds) if it's processed by Gloo controllers, we will resync the entire state of the world continually.

Ideally, we ignore configmaps with a particular label, but that isn't supported in solo-kit. The faster solution, is to ignore it explicitly in code for now, and handle the more robust solution in a follow-up.

## Follow Up Work
1. Solo-kit enhancements to resource filtering
2. Helm hardening around other HA features

## Enterprise Follow-Up
To support this in Enterprise, the following work will be required: https://github.com/solo-io/solo-projects/pull/3974

# Checklist:


pkg/bootstrap/leaderelector/election_factory.go
```diff
@@ -0,0 +1,15 @@
+package leaderelector
+
+import "context"
+
+type ElectionConfig struct {
```
nfuden, line 5: nit: Add comment on exported name?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -575,6 +579,11 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 
 	startRestXdsServer(opts)
 
+	identity, err := startLeaderElection(opts.WatchOpts.Ctx, opts.LeaderElectionFactory)
```
nfuden, line 582: Do we want to guard this more to make it optional like in gloo mesh? https://github.com/solo-io/gloo-mesh-enterprise/blob/main/pkg/bootstrap/bootstrap.go#L159

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return identity, err
+	}
+
+	l, err := k8sleaderelection.NewLeaderElector(
+		k8sleaderelection.LeaderElectionConfig{
+			Lock:          resourceLock,
+			LeaseDuration: 15 * time.Second, // Default value according to docs
+			RenewDeadline: 10 * time.Second, // Default value according to docs
+			RetryPeriod:   2 * time.Second,  // Default value according to docs
```
jackstine, line 49: docs, what docs? do you have the link

sam-heilbron: Good call! Updated the comment to include a link to the actual docs

pkg/bootstrap/leaderelector/election_factory.go
sam-heilbron: good call! updated to include context about this and other exported types

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: Yeah so in mesh, most components share startup/setup functionality, and one of the options to initialize startup is this flag. I would have to double check but I believe that the users don't set this flag, and instead developers do for the agent (which supports leader election). 

In Edge, each setup is distinct so by adding this only to the gloo setup, it means that only the gloo component will start elections. The added complexity is that we can run gloo outside of kube, so I ensured that since we're just adding leader election support based on kube, that we only make it available if kube is already used as the config source (defined in settings)

EItanya: So in Mesh it is configurable, but that setting is meant mostly for developers. It's configurable using a flag though so you have to restart the deployment.

I definitely agree with the overall approach though. I think the factory here for testing means the flag shouldn't be necessary.

test/kube2e/gateway/gateway_suite_test.go
```diff
@@ -129,6 +129,9 @@ settings:
   replaceInvalidRoutes: true
 gateway:
   persistProxySpec: true
+gloo:
+  deployment:
+    replicas: 2
```
jenshu, line 134: will you be adding some e2e testing around making sure that leader election works with the 2 replicas?

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -0,0 +1,81 @@
...
+					leader.Store(true)
+					config.OnStartedLeading(callbackCtx)
+				},
+				OnStoppedLeading: func() {
+					contextutils.LoggerFrom(ctx).Error("Stopped Leading")
+					leader.Store(false)
+					config.OnStoppedLeading()
+				},
+				OnNewLeader: func(identity string) {
+					contextutils.LoggerFrom(ctx).Debugf("New Leader Elected with Identity: %s", identity)
```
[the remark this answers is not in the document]
sam-heilbron: Since this is the base implementation, I tried to keep it as lightweight as possible. I could certainly, add information about the identity of the leader in here, or we could configure Gloo elections to define their "OnNewLeader" callback to print this information. What do you think?

pkg/utils/namespaces.go
```diff
@@ -29,3 +33,10 @@ func ProcessWatchNamespaces(watchNamespaces []string, writeNamespace string) []s
 
 	return watchNamespaces
 }
+
+func GetPodNamespace() string {
+	if podNamespace := os.Getenv("POD_NAMESPACE"); podNamespace != "" {
+		return podNamespace
+	}
+	return "gloo-system"
```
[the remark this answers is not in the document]
sam-heilbron: Yeah the challenge I ran into, is that that constant is defined in some `projects/gloo/pkg/utils`, which causes a circular dependency. I would like in the future (and I tried in my other cleanup PR) to define constants at the top level, without dependencies so they can be imported easily.

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -0,0 +1,81 @@
...
+
+	leOpts := leaderelection.Options{
+		LeaderElection:          true,
+		LeaderElectionID:        config.Id,
+		LeaderElectionNamespace: config.Namespace,
+	}
+	// Create the resource Lock interface necessary for leader election.
+	// Controller runtime requires an event handler provider, but that package is
+	// internal so for right now we pass a noop handler.
+	resourceLock, err := leaderelection.NewResourceLock(f.restCfg, NewNoopProvider(), leOpts)
```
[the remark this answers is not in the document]
sam-heilbron: A noop handler is just the implementation for leader election that does nothing (ie what we have today, where we only support a single replica). Is there anything I can do to help make that clearer?

test/kube2e/gateway/gateway_suite_test.go
sam-heilbron: Yeah first I want to make sure that existing regression tests just all work with multiple replicas.

Ideally I'd like to come up with a more explicit strategy to test that leader election is working, but I'm still thinking about that.

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -389,13 +394,19 @@ func (s *statusSyncer) syncStatus(ctx context.Context) error {
...
+			}
 		} else {
-			// The inputResource's status was successfully written, update the cache and metric with that status
-			status := s.reporter.StatusFromReport(subresourceStatuses, currentStatuses)
-			localInputResourceLastStatus[inputResource] = status
+			contextutils.LoggerFrom(ctx).Debugf("Not a leader, skipping reports writing")
 		}
+
 		status := s.reporter.StatusFromReport(subresourceStatuses, currentStatuses)
 		s.statusMetrics.SetResourceStatus(ctx, inputResource, status)
```
elcasteel, line 411: is statusMetrics what we're reporting to prometheus, etc? Should this be in the leader-only block?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -760,7 +771,16 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
...
+		gwTranslatorSyncer = gwsyncer.NewTranslatorSyncer(
+			opts.WatchOpts.Ctx,
+			opts.WriteNamespace,
+			proxyClient,
+			proxyReconciler,
+			rpt,
+			gatewayTranslator,
+			statusClient,
+			statusMetrics,
+			identity)
```
jackstine, line 783: Just a question in general, for long function signatures, should we try to consolidate them with structs?

projects/gateway/pkg/syncer/translator_syncer.go
sam-heilbron: It is what we're reporting to prometheus, but prometheus will handle the differentiation for us, with unique labels per pod.

https://github.com/solo-io/gloo/pull/6926#discussion_r942799235

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: We could in the future. I don't want to do it as part of this work since it's tangential, but I like the idea of reducing the number of dependencies

test/kube2e/gateway/robustness_test.go
```diff
@@ -238,7 +238,7 @@ var _ = Describe("Robustness tests", func() {
 			// Wait for the proxy to be accepted
 			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
 				return proxyClient.Read(testHelper.InstallNamespace, defaults.GatewayProxyName, clients.ReadOpts{Ctx: ctx})
-			}, 60*time.Second, 1*time.Second)
+			}, 120*time.Second, 1*time.Second)
```
elcasteel, line 241: Are you planning to keep this? Does it make the test more stable? (Could we reduce the time again if we fixed this: https://github.com/solo-io/gloo/issues/6472)

sam-heilbron: I can remove. I found that locally we were near the limit, but even with this bump it's still failing on CI more consistently than I'd like

elcasteel: If there are times when it passes after more than 60 but less than 120 seconds then it makes sense to keep it. I'm mostly concerned about waiting longer for CI to fail (and also a bit concerned about why it would take a minute for a control plane change to propagate in CI)
Would it also make sense (for future CI work) to make these timeouts configurable via environment variable or something for local testing?

sam-heilbron: Yeah my bump ended up being incorrect, and the 60 seconds isn't the limiting factor. In the end, it has to do with how this test works (scaling down deployments) and how it takes time for the lease to be chosen after that step. What is happening is that we scale up the deployment, start election, but that will take 15 seconds for the leader to diagnosed as unavailable and a new one chosen. In the meantime process the VS changes (but don't persist a status change since there is no leader yet). 

The status is only affected by this test because of the scaleDown/scaleUp work, so I opted to just remove the status check. in the end, we perform a query which will return true only if translation has occurred.

test/kube2e/util.go
```diff
@@ -163,11 +182,16 @@ func getSnapOut(metricsPort string) string {
 		return bodyResp
 	}, "5s", "1s").ShouldNot(BeEmpty())
 
-	Expect(bodyResp).To(ContainSubstring("api_gloosnapshot_gloo_solo_io_emitter_snap_out"))
-	findSnapOut := regexp.MustCompile("api_gloosnapshot_gloo_solo_io_emitter_snap_out ([\\d]+)")
+	findSnapOut, err := regexp.Compile("api_gloosnapshot_gloo_solo_io_emitter_snap_out ([\\d]+)")
+	if err != nil {
+		// No snapOut metrics were found, still starting up
+		return 0
```
elcasteel, line 188: If we can't find the snapOut metrics at all, the consistency tests succeed and we move on? Is this change necessary as part of leader election or just a general stability change? Will this cause us to either miss errors caused by breaking metrics (this check has caught lost metrics for me) or start the rest of the test suite too early (and make it flake later because gloo wasn't ready)?

sam-heilbron: I could probably revert this change but it originally worked by:
1. get the metrics once
2. eventually they're consistent
3. consistently they're the same

I changed it to:
1. eventually they are positive AND consistent for n times in a row

I had found that metrics would jitter between steps 2 and 3 causing us to think we had reached stability before we had.

In the end, this was due to a bug in the leader election implementation (not ignoring the configmap storing the state of the leader). So these changes are not necessary, but I think they achieve the same intention: assure metrics are stable.

sam-heilbron: https://github.com/solo-io/gloo/pull/6926/files#diff-6ae69a54709841b3a5db49fcd47d092708bec5326d9e5dde896a3f7e30bbd451R152 we require that the metrics are non-zero

sam-heilbron: but I'm happy to revert these changes to what they were originally

elcasteel: gotcha the two calls to getSnapOut threw me (I searched for one and didn't scroll up further). This seems good (I like failing differently for different problems) but what would be really cool would be to have a failure message that said one of "metrics not found" or "snapshots never stabilized).

sam-heilbron: i like it! updated

projects/gloo/pkg/syncer/translator_syncer.go
```diff
@@ -96,15 +102,22 @@ func (s *translatorSyncer) Sync(ctx context.Context, snap *v1snap.ApiSnapshot) e
 		reports.Merge(intermediateReports)
 	}
 
-	if err := s.reporter.WriteReports(ctx, reports, nil); err != nil {
-		logger.Debugf("Failed writing report for proxies: %v", err)
-		multiErr = multierror.Append(multiErr, eris.Wrapf(err, "writing reports"))
+	if s.identity.IsLeader() {
+		// Only leaders will write reports
```
kdorosh, line 106: i would think only leaders would do translation etc too? otherwise we may use a bunch of CPU / RAM for no reason? also a bunch of checks like this all over the codebase will be very very hard to maintain rather than a single check at the top level that disables everything else until elected as leader

should we even be serving validation requests either? do we really want to have watches duplicated everywhere (can we stress out or require k8s or dependencies to scale?) simultaneously? i thought most of the value of having the follower ready was to ensure that the pod was ready / was already scheduled

kdorosh: open to more discussion here; there are some valid points against but i think it will be easier and perhaps better to just short circuit everything at the top level

nfuden: Being able to scale validation seems pretty nice but the point about watches is a good one it might even be behavior one could opt into. That being said if the watches arent in place is the pod really ready to go?

kdorosh: that's a fair point; probably good to have the watches ready and warmed before being marked as ready and thus available for failover. i may just be paranoid; depends on how much strain gloo alone puts on the k8s apiserver itself and thus scaling it could become an issue. might be nice to get input from the field here to determine if this is warranted / should be configurable behavior

EItanya: Everyone should be doing translation and validation. Envoy/k8s will reach out to the service and the request will be Load Balanced, so all must be ready to serve the latest.

pkg/bootstrap/leaderelector/election_factory.go
```diff
@@ -0,0 +1,29 @@
...
+	// Callback function that is executed when the current component stops leading
+	OnStoppedLeading func()
+	// Callback function that is executed when a new leader is elected
+	OnNewLeader func(leaderId string)
+	// ReleaseOnCancel should be set true if the lock should be released
+	// when the run context is cancelled. If you set this to true, you must
+	// ensure all code guarded by this lease has successfully completed
+	// prior to cancelling the context, or you may have two processes
+	// simultaneously acting on the critical path.
+	ReleaseOnCancel bool
```
EItanya, line 22: Are we using this, if not why is it exposed?

projects/gloo/pkg/api/converters/kube/artifact_converter.go
```diff
@@ -27,6 +29,21 @@ func (c *converter) FromKubeConfigMap(_ context.Context, rc *skcfgmap.ResourceCl
...
+		// This is a temporary (hacky) solution
+		// ConfigMaps are used by our Kubernetes LeaderElection implementation to store that state of the
+		// current leader. This means that every election (2s interval) the ConfigMap is updated.
+		// Gloo Edge will recognize this change, and re-process the state of the world.
+		// The result is that the system does not become consistent.
+		//
+		// A preferred solution would be to filter out this resouce (either by annotation or label)
+		// Solo-Kit only supports inclusive label selection, not exclusive label selection.
+		// In a follow-up, we will extend the filtering of solo-kit to enable filtering out certain
+		// labels. But for now, we return nil, meaning that this resource will be skipped by our Watch.
```
EItanya, line 43: I'm a little worried this may be a blocker, this means the system will never reach a steady state.

pkg/bootstrap/leaderelector/election_factory.go
sam-heilbron: In gloo We do opt into it. I could probably just not expose it and make it the default

EItanya: Why do we do this? I'm not sure we should. I'd prefer the safe route where the leader loses the lease and the next pod picks it up a few seconds later.

projects/gloo/pkg/api/converters/kube/artifact_converter.go
sam-heilbron: I'm our kube tests we ensure that metrics stabilize, before we run the tests. This is what failed originally and clued me into the need for this change.

Do you not think this change solves the problem?

EItanya: Spoke with @nfuden offline, I misunderstood the comment. This is fine by me

pkg/bootstrap/leaderelector/election_factory.go
sam-heilbron: That's fair. I had assumed that releasing the lock on cancel would actually allow the transition to be quicker, since not releasing the lock would mean we have to wait the leaseDuration (15second default) for new leaders to take over.

projects/gloo/pkg/syncer/translator_syncer.go
kdorosh: are we solving for scaling gloo, or just HA? I agree, if we want to scale gloo and load balance validation / xds requests we need to run translation (really everything but status reporting) on every pod. that said, i'm also nervous about scaling gloo given how many requests it can make, particularly in the consul case. I don't think having 3-5 gloo pods will be all that aggressive on the k8s apiserver because of the natural watch api, but for backends like consul we implement watches by querying for every single endpoint individually this puts a lot of strain on the consul server, and scaling gloo to multiple replicas could cause problems there when we may just want to solve for HA / ensuring pods can be scheduled.

really feels like more of a product requirement decision here than an engineering one to me. is the customer concerned about just scheduling pods? or about scaling gloo itself (is CPU / RAM pegged?)

kdorosh: also i don't know enough about the leader election libraries used here, but have we reasoned about what happens in error scenarios? usually in failure there can be (briefly) two or zero leaders..

sam-heilbron: I don't believe it's possible for their to be 2 at once, but zero can occur when the leader dies and there is a delay (configurable) for a new leader to be elected. Given that we are only performing reporting on statuses, it's ok if there is this delay. We can change the duration that it reuquires for a new leader to be elected to reduce that gap however

kdorosh:
before we merge i'd recommend reading https://aws.amazon.com/builders-library/leader-election-in-distributed-systems/ and following best practices here, in particular:
- add metrics for leaders
- reason about lease expiration, especially in the case of garbage collection (which is a real concern for us as a cpu intensive application)

kdorosh:
I think the defaults are sane; i don't think we will hit multi second issues unless we have deadlock, kernel errors, almost total network failure.. the kinds of things that should result in a new leader regardless

test/kube2e/gateway/gateway_test.go
```diff
@@ -293,7 +293,7 @@ var _ = Describe("Kube2e: gateway", func() {
 					return nil, eris.Errorf("Expected %d routes to test runner service, but found %d", expectedRoutesToTestRunner, actualRoutesToTestRunner)
 				}
 				return proxy, nil
-			}, 60*time.Second, 1*time.Second)
```
nfuden: Is 30 really enough?

sam-heilbron: It shouldn't take longer than 30 seconds to process a proxy status I don't think

sam-heilbron: Running locally it takes under 15 seconds. The challenge I think, is that some of our tests scale down and up our gloo instances, which causes leader election to adjust. This is good for general testing, but also means that the first test to run after the scaling up, must wait for the new leader to be elected before the status will be updated.

nfuden:
I am content with current state.

nfuden (approved):

pkg/bootstrap/leaderelector/kube/metrics.go
```diff
@@ -0,0 +1,37 @@
+package kube
```
timflannagan, line 1: @sam-heilbron Do you remember why we manually vendored this file from the c-r or client-go repository?

Issue #6821, opened 2022-07-29
Helm - add timeout to Jobs (activeDeadlineSeconds)


bcollard:
### Is your feature request related to a problem? Please describe.

Customer asked us to add `activeDeadlineSeconds` to all our Kubernetes Jobs.

PR #6951, opened 2022-08-15
Remove default activeDeadlineSeconds from jobs

jenshu:
Only set the value if it's explicitly passed in through helm values.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6821


jackstine (approved):
LGTM!

PR #6958, opened 2022-08-16
ISSUEFORM: Update for 1.12 release

nfuden:
Update the request forms


jenshu (approved):

nfuden:
/skip-ci CSRF defined on listener only on listener [It] should succeed with allowed origin also flaked?

jenshu:
(adding wip label so cve PR merges first)

PR #6962, opened 2022-08-16
Fix namespace check when registering cluster

jenshu:
Use the correct namespace to check for gloo installation when registering clusters for Gloo Fed.

### DIFF
--- a/projects/gloo/cli/pkg/cmd/federation/register/register.go
+++ b/projects/gloo/cli/pkg/cmd/federation/register/register.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/solo-io/gloo/pkg/cliutil"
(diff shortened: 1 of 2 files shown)


projects/gloo/cli/pkg/cmd/federation/register/register.go
```diff
@@ -33,7 +32,7 @@ func installCrdsToRemote(context string) error {
 
 	_, err = helmInstall.Run(chartObj, nil)
 	if err != nil {
-		_, _ = fmt.Fprintf(os.Stderr, "\ngloo failed to install CRDs! Detailed logs available at %s.\n", cliutil.GetLogsPath())
```
jenshu, line 36: the logs didn't seem to get written to this file, so just printed the error instead

nfuden (approved):

PR #6979, opened 2022-08-17
Back 1.12/lambda role based

nfuden:
When using resource based policies for cross account calls to lambda one specifies a whole function arn. This means that you dont have to do so on upstreams with the accountid specified.


changelog/v1.12.8/lambda-by-resource-policy.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NEW_FEATURE
```
ben-taussig-solo, line 2: Do we want to change this bc this is a backport?

nfuden: Oh my I should have known when I saw the error message good call

ben-taussig-solo: Changed to `FIX` here: https://github.com/solo-io/gloo/pull/6979/commits/9f56bb1afb5d960311ebc8baebffee9bd4384d5f

ben-taussig-solo (approved):
Looks good!

Issue #6977, opened 2022-08-17
[1.12.x] Duplicated "VolumeMounts" Keys in 1-gloo-deployment.yaml


vmarlier:
### Gloo Edge Version

1.12.x (latest stable)

### Kubernetes Version

1.22.x

### Describe the bug

Hello,
During our attempt to update Gloo to 1.12.6 with our Fluxv2 stack, we noticed an issue outputted by the `Helm-Controller`.
Here is the error:
```
Helm upgrade failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
        line 101: mapping key "volumeMounts" already defined at line 75
```
Basically Helm is complaining about a duplicated `volumeMounts` key.

The issue is only firing when we want to install through the flux stack since they added a check for duplated keys. When using the helm cli, it seems helm do not complain but take randomly one of the VolumeMounts key definition. So it is still an issue.

When taking a look at the file for 1.12.x I can see a duplication [here](https://github.com/solo-io/gloo/blob/e46bcf6ccf791ff65b0b73f67dec8e2f39bbd4e2/install/helm/gloo/templates/1-gloo-deployment.yaml#L195) and [here](https://github.com/solo-io/gloo/blob/e46bcf6ccf791ff65b0b73f67dec8e2f39bbd4e2/install/helm/gloo/templates/1-gloo-deployment.yaml#L253).

It is possible to fix this please ?

### Steps to reproduce the bug

1. Install Gloo via Fluxv2 stack with a version prior to 1.12.x
2. Update to 1.12.x
3. Should see an issue with the fluxv2 helm-controller.

### Expected Behavior

Being able to install it through fluxv2 without any blocker.

kevin-shelaga:
looks to be an issue when `gateway.validation.enabled` and `gateway.enabled` are true

sam-heilbron:
This has been resolved in affected versions (1.12, 1.3) of Gloo.

PR #6981, opened 2022-08-17
Remove Duplicate VolumeMount from Gloo Deployment

sam-heilbron:
# Description

Fix helm template for gloo deployment to not re-define VolumeMounts twice

# Context

## Duplicate VolumeMount (https://github.com/solo-io/gloo/issues/6977)

If upgrading Gloo using the Fluxv2 stack, the Helm-Controller would emit an error about duplicate keys, and fail the upgrade. Anyone using the Helm CLI would not encounter an error, but would have one of the VolumeMount defintions applied, which could lead to errors.

## Test Pollution
Previously, Helm tests would share the previous `TestManifest` if they did not define one themselves. This could cause some sneaky bugs. Shoutout to @jenshu for identifying this.

# Checklist:


sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=794d1e85-dce1-4583-857e-1f0e368e51d1 haven't seen this error before

pkg/utils/syncutil/log_redactor_test.go
```diff
@@ -56,6 +67,13 @@ var _ = Describe("Log Redacter", func() {
 		Expect(s).NotTo(ContainSubstring(privateKey))
 	})
 
+	It("contains redacted content when artifacts are present", func() {
```
elcasteel, line 70: Nit: this description threw me off. Maybe switch to something like "cert replaced with expected value"

install/test/helm_test.go
```diff
@@ -1392,6 +1392,10 @@ metadata:
 					})
 
 					It("disabling http gateway disables corresponding service port", func() {
+						prepareMakefile(namespace, helmValues{
+							valuesArgs: []string{"gatewayProxies.gatewayProxy.gatewaySettings.disableHttpGateway=true"},
+						})
```
jenshu, line 1397: how were these tests previously working without the value passed in?

pkg/utils/syncutil/log_redactor_test.go
sam-heilbron: I like it! I'll update this. I had lazily copied the existing tests for secret, but I'll change both since I agree they're much clearer the way you outline it

install/test/helm_test.go
sam-heilbron: Ha i was wondering the same thing. I didn't look too closely so I'm not sure. But the test claims to set a value which we weren't setting.

elcasteel (approved):

install/test/helm_test.go
jenshu: ah ok i think it was previously passing because the tests carried over the values from the `prepareMakefile` from the previous tests

install/test/helm_suite_test.go
```diff
@@ -110,36 +116,50 @@ var _ ChartRenderer = &helm3Renderer{}
...
 
 	for _, hook := range hooks {
 		manifest := hook.Manifest
-		_, err = f.Write([]byte("\n---\n" + manifest))
+		_, err = testManifestFile.Write([]byte("\n---\n" + manifest))
 		Expect(err).NotTo(HaveOccurred(), "Should be able to write the hook manifest to the temp file for the helm unit tests")
 	}
 
-	return NewTestManifest(f.Name()), nil
+	return NewTestManifest(testManifestFile.Name()), nil
```
jenshu, line 159: do we need to close the files when we're done?

jenshu (dismissed):

install/test/helm_test.go
sam-heilbron: ohhh good find! Is there a change I can make to resolve that?

install/test/helm_suite_test.go
sam-heilbron: good call! Updated

jenshu (approved):

PR #7010, opened 2022-08-23
Validation Webhook: Auto-Accept Status Updates

sam-heilbron:
# Description

Auto-accept status updates, instead of requiring a full translation run

# Context

The validation webhook executes a full translation run to determine whether a proposed change is valid. Status only changes will not affect the validity of the proposed change, and therefore it is both unnecessary and expensive to run translation for status only changes.

# Checklist:


projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,126 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+	// If it is, we do not need to validate the resource
+	var newResource, oldResource crdv1.Resource
+	if err := json.Unmarshal(admissionRequest.Object.Raw, &newResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+	if err := json.Unmarshal(admissionRequest.OldObject.Raw, &oldResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+
+	specsAreEqual := reflect.DeepEqual(newResource.Spec, oldResource.Spec)
```
EItanya, line 563: `proto.Equal`?

EItanya: Also, what about metadata, do we not care about label changes, etc.?

sam-heilbron: Yeah I was thinking about this and erred on the side of "if it's not a spec change, allow it". One example of a problematic change is:
You could change the label of a VirtualService, which is selected by a Gateway, and that would mean that the Gateway no longer selects any VS's and should reject the change.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,126 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
 }
+
+func (wh *gatewayValidationWebhook) shouldValidateResource(ctx context.Context, admissionRequest *v1beta1.AdmissionRequest, resource resources.InputResource) (bool, error) {
+	logger := contextutils.LoggerFrom(ctx)
+
+	if err := protoutils.UnmarshalResource(admissionRequest.Object.Raw, resource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+	if skipValidationCheck(resource.GetMetadata().GetAnnotations()) {
+		logger.Debugf("Skipping validation. Reason: detected skip validation annotation")
```
nfuden, line 545: nit: can we make this log line more usable? We have information on the resource perhaps we can decide on some subset of interest things to print out in this log? Otherwise this could get lost.

sam-heilbron: What additional data do you think would be valuable to include in the log?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
jenshu: > You could change the label of a VirtualService, which is selected by a Gateway, and that would mean that the Gateway no longer selects any VS's and should reject the change.

wouldn't this be a reason for *not* skipping validation when there's a metadata-only change?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,126 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+	}
+
+	if admissionRequest.Operation != v1beta1.Update {
+		return true, nil
+	}
+
+	// For update requests, we check to see if this is a status update
+	// If it is, we do not need to validate the resource
+	var newResource, oldResource crdv1.Resource
+	if err := json.Unmarshal(admissionRequest.Object.Raw, &newResource); err != nil {
```
nfuden, line 556: Question: Should we add a comment detailing that unknown fields wont trigger translation?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
nfuden: Perhaps admissionRequest.Name or admissionRequest.Kind + admissionRequest.Resource

nfuden: Just imagine a large deployment with multiple skip annotations. It might be interesting to see what thing is skipping validation.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: Unknown fields will be rejected by the validation schema, not the validation webhook, so I don't think we need to worry about them. We have a test https://github.com/solo-io/gloo/blob/master/test/kube2e/gateway/gateway_test.go#L2229 which proves this

nfuden: Right but thats based on the CRD itself right?

So if I have this correct then the following is true: 
1. kubernetes itself validates that the input corresponds to CRD and rejects 
2. we unmarshal to spec from incoming request ignoring unknown fields

So this means that if the CRD is say from a newer version that actually has some new fields then we wont perform validation which also saves unnecessary compute

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: Should that go with https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go#L276, which we log at the beginning of every request? I imagined the metadata about a request is logged early, and the decision alone can be logged later. happy to add additional metadata however, if you think it's valuable

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,132 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
 }
+
+func (wh *gatewayValidationWebhook) shouldValidateResource(ctx context.Context, admissionRequest *v1beta1.AdmissionRequest, resource resources.InputResource) (bool, error) {
+	logger := contextutils.LoggerFrom(ctx)
+
+	if err := protoutils.UnmarshalResource(admissionRequest.Object.Raw, resource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+	if skipValidationCheck(resource.GetMetadata().GetAnnotations()) {
+		logger.Debugf("Skipping validation. Reason: detected skip validation annotation")
```
jenshu, line 545: think @nfuden also mentioned this, but can we add the resource name/namespace and kind?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,132 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+
+	// A status-only update will only differ by status and generation, so we nil both fields before comparing
+	newResource.Status = nil
+	oldResource.Status = nil
+	newResource.Generation = 0
+	oldResource.Generation = 0
+
+	specsAreEqual := reflect.DeepEqual(newResource, oldResource)
+	if specsAreEqual {
+		logger.Debugf("Skipping validatation. Reason: status only update")
```
jenshu, line 571: add resource name/namespace/kind here too?

jenshu, line 571:
```suggestion
		logger.Debugf("Skipping validation. Reason: status only update")
```

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
nfuden: Once we start interleaving logs then it gets lost fast but Im fine with this staying as is if we want. Perhaps we need to have some of the metadata appended to all logs within a given sync run but thats likely out of scope.

nfuden: Thought: What if we just make a new context that has a logger with the additional fields?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +451,132 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+	// If it is, we do not need to validate the resource
+	var newResource, oldResource crdv1.Resource
+	if err := json.Unmarshal(admissionRequest.Object.Raw, &newResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+	if err := json.Unmarshal(admissionRequest.OldObject.Raw, &oldResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+
+	// A status-only update will only differ by status and generation, so we nil both fields before comparing
```
jenshu, line 563: does resourceVersion not change on status updates?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: I like it!

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: Yeah good call. I tried adding a kube2e test, but it's tricky to test, given that we still update generation and other metadata fields, so its hard to identify if the auto-accept of status does anything. Ill keep working on one siince I think it will be necessary to prove out/ensure we don't have a regression.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
nfuden: https://github.com/solo-io/gloo/pull/7013/files

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: yeah good call, it's now included in the context for the logger

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +460,130 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+
+	// For update requests, we check to see if this is a status update
+	// If it is, we do not need to validate the resource
+	var newResource, oldResource crdv1.Resource
+	if err := json.Unmarshal(admissionRequest.Object.Raw, &newResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+	if err := json.Unmarshal(admissionRequest.OldObject.Raw, &oldResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
```
kdorosh, line 570: rather than do it this way... should we hash it instead and compare hashes? that is what hashing was intended for

kdorosh: this should ignore the status but also ignore any fields on the spec we want opted out here as well (which could cause similar issues)

nfuden: If we do that and there are new fields on the CRD that our current version of gloo doesnt care about then it will still generate validations.

sam-heilbron: Yeah I think that's cleaner. I think I'll need to pass in the typed resource into the function for the 'oldResource'. I'll update!

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=f78e19e9-dcd9-4065-9a5b-d14f4bff3bf2 consul works as expected

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -450,75 +456,128 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
...
+		return true, nil
+	}
+
+	// For update requests, we check to see if this is a status update
+	// If it is, we do not need to validate the resource
+	if err := protoutils.UnmarshalResource(admissionRequest.OldObject.Raw, oldResource); err != nil {
+		return false, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
+	}
+
+	equalResources := resource.MustHash() == oldResource.MustHash()
```
jenshu, line 570: MustHash will panic on error. do we want that, or should we just handle the errors?

jenshu, line 570: Hash uses the Metadata which includes resourceVersion and generation. so for a status-only change, it looks like the hash could change?

sam-heilbron: Yeah so part of the benefit of converting to the gloo types (https://github.com/solo-io/solo-kit/blob/master/api/v1/metadata.proto#L48) is that the core.Metadata type ignores certain fields when hashing.

jenshu: oh nice, didn't notice those annotations!

sam-heilbron: In our emitter, we panic if the hash cannot complete (https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/api/v1/gloosnapshot/api_snapshot_emitter.sk.go#L883) so I chose to follow that pattern

jenshu:
LGTM as well

nfuden (dismissed):

nfuden (approved):

PR #7028, opened 2022-08-25
fixes #5961 disable strict response headers to allow websocket connections

### LINKED ISSUE #5961: Disable require_strict_1xx_and_204_response_headers in envoy configuration
### Version

_No response_

### Is your feature request related to a problem? Please describe.

When Gloo edge is used alongside the Istio ingress gateway, web socket connections (verified with SignalR 5.x) are rejected with a 502 Bad Gateway, and with the following log in the `gateway-proxy`:
```
Error dispatching received data: http/1.1 protocol error: transfer encoding not allowed in 1xx or 204
```

This happens when a 101 status code is returned to upgrade the connection to web sockets. The istio-ingress gateway adds the following header `key=transfer-encoding value=chunked` as shown below

```
[2022-02-22 13:55:50.465][20][trace][http] [external/envoy/source/common/http/http1/codec_impl.cc:499] [C327] 
➥ completed header: key=transfer-encoding value=chunked
```

Which makes the envoy fail with: 
```
Error dispatching received data: http/1.1 protocol error: transfer encoding not allowed in 1xx or 204
```

### Describe the solution you'd like

Envoy from version 1.15 stopped allowing upstream 1xx or 204 responses with Transfer-Encoding or non-zero Content-Length headers. So now the default is to reject such requests. This however can be reverted by setting the following setting on envoy boostrap configuration to false (the default is true)
```
envoy.reloadable_features.send_strict_1xx_and_204_response_headers: false
```

The same is configured in the Istio side: https://github.com/istio/istio/blob/master/pkg/bootstrap/config.go#L282
To resolve those issues: [#28433](https://github.com/istio/istio/issues/28433), [#36711](https://github.com/istio/istio/issues/36711) 

### Describe alternatives you've considered

_No response_

### Additional Context

We only need to update our helm chart https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/12-ingress-proxy-configmap.yaml#L20

To add the following config:
```
    layered_runtime:
      layers:
      - name: static_layer
        static_layer:
          overload:
            global_downstream_max_connections: 250000
        envoy.reloadable_features.require_strict_1xx_and_204_response_headers: false    # 1
      - name: admin_layer
        admin_layer: {}
```
1#  The change

Paschi28:
# Description

This pr fixes #5961 by adding the config line to disable require_strict_1xx_and_204_response_headers by default. 
The disabling is necessary to connect websocket connections through the gateway .

# Context

Users ran into this bug doing websocket connections e.g. SignalR.

# Checklist:

### DIFF
--- a/install/helm/gloo/templates/12-ingress-proxy-configmap.yaml
+++ b/install/helm/gloo/templates/12-ingress-proxy-configmap.yaml
@@ -18,6 +18,7 @@ data:
         static_layer:
           overload:
             global_downstream_max_connections: 250000
+          envoy.reloadable_features.require_strict_1xx_and_204_response_headers: false
       - name: admin_layer
         admin_layer: {}
     node:


Paschi28:
Hi @rinormaloku,
could you please have a look.

rinormaloku:
The change is correct.
@kdorosh can you please review this change

install/helm/gloo/templates/12-ingress-proxy-configmap.yaml
```diff
@@ -18,6 +18,7 @@ data:
         static_layer:
           overload:
             global_downstream_max_connections: 250000
+          envoy.reloadable_features.require_strict_1xx_and_204_response_headers: false
```
kdorosh, line 21: this is not compliant with the HTTP spec and has been disabled since envoy 1.22 https://github.com/projectcontour/contour/issues/4458#issuecomment-1111217672

i think we should add something like this to the other configmaps as an escape hatch for clients that aren't spec compliant https://github.com/solo-io/gloo/blob/df2dc06df85dd5e030ba6f5d53c1d666926ce6d1/install/helm/gloo/templates/9-gateway-proxy-configmap.yaml#L30-L32

but we should not allow all clients to make spec non-compliant requests

kdorosh (changes requested):
let's do a simpler helm change as an escape hatch rather than allow non spec-compliant clients (including istio) by default

Paschi28:
I figured out, that the needed change is possible using the helm spec `gloo.gatewayProxies.gatewayProxy.customStaticLayer`

PR #7035, opened 2022-08-25
validation: Update a couple debug logs that can be large with large p…

nfuden:
# Description


projects/gateway/pkg/validation/validator.go
```diff
@@ -706,9 +706,11 @@ func (v *validator) ValidateDeleteUpstream(ctx context.Context, upstreamRef *cor
 			return err
 		}
 	}
-	logger.Debugf("Got response from GlooValidationService: %s", response.String())
 
-	_, err = v.getReportsFromGlooValidationResponse(response)
+	reports, err := v.getReportsFromGlooValidationResponse(response)
+	// reports are gauranteed to not be nil at this point and we can more easily not report on the proxy
+	// than on the response itself without more loops on non-debug calls
+	logger.Debugf("Got response from GlooValidationService: UpstreamReports: %v, ProxyReports: %v", *reports.UpstreamReports, *reports.ProxyReports)
```
elcasteel, line 713: This might be too paranoid but given that we've also seen very large reports/statuses on proxies maybe we should only log the state and/or move all the proxy contents logging to "trace"

nfuden: Yeah so even proxyreports could be super large. Ill check it out

elcasteel: Also, will this construct the (potentially large) formatted string regardless of whether we're logging at debug level? Is this a case where we'd we'd want to wrap in `if debugEnabled` (maybe go's logger is smarter than log4j, I'm not sure)

nfuden: zap is smart enough to check that as the first statement prior to the execution checks the not exposed level. Given that I didnt want to make a loop gated by level as it wouldnt be pulling from the logger itself.

sam-heilbron (commented):
Each of the validator methods seem to have a debug log line where we log the entire response (ValidateDeleteSecret: logger.Debugf("Got response from GlooValidationService: %s", response.String())). Do we need to update these instances as well?

Side note: Given the ease of using the String method for the ValidationRequest and ValidationResponse, would it make sense to implementer some stringer method on those objects which can always be called and contain only the relevant details?

nfuden:
The problem is that it is the stringer and the appropriate small subobjects portion would have a decent loop overhead. We can address that at a later point.

nfuden:
/kick [Fail] Access Log in memory Access Logs File [It] can create string access logs

sam-heilbron (approved):

PR #5421, opened 2021-09-29
Node selector overwrite

bewebi:
# Description

Set custom gatwayProxies' nodeSelector to its original values, if present, undoing the mergeOverwrite with the default gatewayProxy for that field

# Context

Test custom gatewayProxies with and without nodeSelector in separate unit tests and together, with default declared third, in another unit test which would catch a bug I encountered during development where default values were overwritten when custom gatewayProxies with nodeSelector set were declared first.

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5286

PR #7052, opened 2022-08-29
Small helm fixes and test updates to enable istio integration

elcasteel:
# Description

1. Allow users to set an istio revision tag instead of the default.
This is done by adding the field `values.global.istioIntegration.istioSidecarRevTag`. When that field is set, all the places where we were previously setting `istio-injection: enabled` or `sidecar.istio.io/inject: "true"` we instead set `istio.io/rev:<revTag>`
2. Use the default values for excludeInboundPorts in custom gateways when no custom value is set. (And tests to verify).

# Context

These are two install bugs that customers have hit while trying to use Istio integration settings in Gloo.

# Checklist:


install/helm/gloo/templates/3-discovery-deployment.yaml
```diff
@@ -27,7 +27,9 @@ spec:
         {{ $key }}: {{ $value | quote }}
         {{- end }}
         {{- end }}
-        {{- if .Values.global.istioIntegration.whitelistDiscovery }}
+        {{- if and .Values.global.istioIntegration.whitelistDiscovery .Values.global.istioIntegration.istioSidecarRevTag }}
+        istio.io/rev: {{ .Values.global.istioIntegration.istioSidecarRevTag }}
+        {{- else if .Values.global.istioIntegration.whitelistDiscovery }}
         sidecar.istio.io/inject: "true"
```
jenshu, line 33: should `sidecar.istio.io/inject` also be set to true in the `if` case above?

elcasteel: Going off this documentation from istio, I don't think so: https://istio.io/latest/blog/2021/revision-tags/

elcasteel: yeah, another set of docs says that `sidecar.istio.io/inject` takes precedence over the revision tag (so we shouldn't set both): https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/

jenshu: ah ok. looks like there are 2 injection labels: `istio-injection=enabled` for namespaces, and `sidecar.istio.io/inject=true` for pods. do we also need to support the revision tag in the [namespace template](https://github.com/solo-io/gloo/blob/c8d021be733b9c30c9b917a6ff0cc0af6cf2456f/install/helm/gloo/templates/0-namespace.yaml#L9)?

elcasteel: updated the namespace template and added a test

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -11,8 +11,10 @@
...
 {{- $spec := deepCopy $gatewaySpec | mergeOverwrite (deepCopy $gatewayProxy) -}}
 {{- $ports := list }}
+{{- if not (empty $spec.podTemplate) }}
+  {{- $ports = (list $spec.podTemplate.httpPort $spec.podTemplate.httpsPort $spec.podTemplate.extraPorts) }}
+{{- end }}
 {{- if not (empty $gatewaySpec.podTemplate) }}
-  {{- $ports = (list $gatewaySpec.podTemplate.httpPort $gatewaySpec.podTemplate.httpsPort $gatewaySpec.podTemplate.extraPorts) }}
   {{- if not (empty $gatewaySpec.podTemplate.nodeSelector) }}
     {{- $_ := set $spec.podTemplate "nodeSelector" $gatewaySpec.podTemplate.nodeSelector }}
   {{- end }}
```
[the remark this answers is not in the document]
elcasteel: No, there are existing tests that verify this behavior: (all added in this PR: https://github.com/solo-io/gloo/pull/5421)
This block is explicitly checking to see if custom nodeSelectors exist and ensure that they are set.

install/test/helm_test.go
```diff
@@ -938,6 +940,7 @@ var _ = Describe("Helm Test", func() {
 							"global.istioIntegration.enableIstioSidecarOnGateway=true",
 							fmt.Sprintf("gatewayProxies.gatewayProxy.podTemplate.httpPort=%d", httpPort),
 							fmt.Sprintf("gatewayProxies.gatewayProxy.podTemplate.httpsPort=%d", httpsPort),
+							fmt.Sprintf("gatewayProxies.namedGatewayProxy.disabled=false"),
```
jenshu, line 943: why was the addition of disabled=false needed?

install/test/helm_test.go
```diff
@@ -969,10 +972,77 @@ var _ = Describe("Helm Test", func() {
...
+					httpsPort := 8443
+					secondDeploymentHttpPort := 1337
+					secondDeploymentHttpsPort := 1338
+					prepareMakefile(namespace, helmValues{
+						valuesArgs: []string{"global.istioIntegration.whitelistDiscovery=true",
+							"global.istioIntegration.enableIstioSidecarOnGateway=true",
+							"global.istioIntegration.istioSidecarRevTag=some-revision",
+							fmt.Sprintf("gatewayProxies.gatewayProxy.podTemplate.httpPort=%d", httpPort),
+							fmt.Sprintf("gatewayProxies.gatewayProxy.podTemplate.httpsPort=%d", httpsPort),
+							fmt.Sprintf("gatewayProxies.namedGatewayProxy.disabled=false"),
```
jenshu, line 994: same here - why is disabled=false needed, shouldn't that be the default?

i suspect this is to fix an existing bug where the helm chart isn't rendering some of the resources (deployments/services/configmaps) associated with a custom gateway proxy unless we explicitly enable the custom gw proxy. should we just fix that bug in the helm chart?

elcasteel: I replaced this with setting the logLevel and a comment to make it clear that this is a one-line way to put a proxy with a custom name in the map of gatewayProxies.

jenshu (approved):

Issue #7038, opened 2022-08-26
Gateway robustness e2e Test Cleanup


nfuden:
### Gloo Edge Version

1.12.x (latest stable)

### Describe the bug

Recent flake count is high for  Robustness tests Updates Envoy endpoints, even if proxy is invalid works, even with deleted services [It] works, even with deleted services 

This makes all CI and therefore all work take much longer.

### Steps to reproduce the bug

Run a pr through CI on 1.12 or 1.13 post leader election. 

### Expected Behavior

No subtest fails

elcasteel:
We have a separate issue open for this, but I'm seeing the domain conflict issue relatively frequently during robustness test runs: https://github.com/solo-io/gloo/runs/8078063464?check_suite_focus=true

Issue #6988, opened 2022-08-18
Kube2e: gateway Proxy reconciliation [It] should process proxy with deprecated label


ashishb-solo:
### Gloo Edge Version

1.13.x (beta)

### Describe the bug

```
2022-08-18T22:16:09.6277613Z LOGS FROM gloo-system.testrunner: STARTING SLEEP! Access me.
2022-08-18T22:16:09.6278027Z ** End Kubernetes Dump ** 
2022-08-18T22:16:09.6278344Z [AfterEach] Proxy reconciliation
2022-08-18T22:16:09.6278828Z   /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:250
2022-08-18T22:16:09.6279220Z [AfterEach] Kube2e: gateway
2022-08-18T22:16:09.6279656Z   /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:206
2022-08-18T22:16:09.6280053Z SSSSSSSSSSSSSSSSSSSSSSSSSSSS
2022-08-18T22:16:09.6282167Z JUnit report was created: /home/runner/work/gloo/gloo/test/kube2e/gateway/junit.xml
2022-08-18T22:16:09.6282481Z 
2022-08-18T22:16:09.6282490Z 
2022-08-18T22:16:09.6282797Z Summarizing 1 Failure:
2022-08-18T22:16:09.6282991Z 
2022-08-18T22:16:09.6283282Z [Fail] Kube2e: gateway Proxy reconciliation [It] should process proxy with deprecated label 
2022-08-18T22:16:09.6283809Z /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:283
2022-08-18T22:16:09.6284051Z 
2022-08-18T22:16:09.6284196Z Ran 6 of 35 Specs in 205.443 seconds
2022-08-18T22:16:09.6284713Z FAIL! -- 5 Passed | 1 Failed | 0 Pending | 29 Skipped
2022-08-18T22:16:09.6285136Z --- FAIL: TestGateway (206.14s)
2022-08-18T22:16:09.6285407Z FAIL
2022-08-18T22:16:09.6829303Z 
2022-08-18T22:16:09.6830078Z Ginkgo ran 1 suite in 16m13.703594998s
2022-08-18T22:16:09.6834399Z Test Suite Failed
2022-08-18T22:16:09.6979952Z make: *** [run-tests] Error 1
[...]

Issue #6932, opened 2022-08-10
Regression test flake: linkerd enabled updates routes...


elcasteel:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

`[Fail] Kube2e: gateway tests with virtual service [AfterEach] linkerd enabled updates routes with appended headers appends linkerd headers when linkerd is enabled 
/home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:341`

Context:

```
ail handler msg Unexpected error:
    <*errors.withStack | 0xc0019b3ea8>: {
        error: <*errors.withMessage | 0xc00114e5a0>{
            cause: <*errors.StatusError | 0xc001ab1360>{
                ErrStatus: {
                    TypeMeta: {Kind: "", APIVersion: ""},
                    ListMeta: {
                        SelfLink: "",
                        ResourceVersion: "",
                        Continue: "",
                        RemainingItemCount: nil,
                    },
                    Status: "Failure",
                    Message: "admission webhook \"gloo.gloo-system.svc\" denied the request: resource incompatible with current Gloo snapshot: [Gateway validation is yet not available. Waiting for first snapshot]",
                    Reason: "",
                    Details: {
                        Name: "vs",
                        Group: "gateway.solo.io",
                        Kind: "VirtualService",
                        UID: "",
                        Causes: [
                            {
                                Type: "",
                                Message: "Error Gateway validation is yet not available. Waiting for first snapshot",
                                Field: "",
                            },
                        ],
                        RetryAfterSeconds: 0,
                    },
                    Code: 400,
                },
            },
            msg: "deleting resource vs",
        },
        stack: [0x3070bc5, 0x3070b5a, 0x3bcff26, 0x4a61d2c, 0x1ca8596, 0x1ca7c98, 0x1ca8a15, 0x1cb1229, 0x1cb0b8c, 0x1cb02ec, 0x1ce567b, 0x1ce5276, 0x1ce44d7, 0x1ce7c56, 0x1cfd893, 0x1cfd5f8, 0x4a1acde, 0x1858614, 0x16f3d01],
    }
    deleting resource vs: admission webhook "gloo.gloo-system.svc" denied the request: resource incompatible with current Gloo snapshot: [Gateway validation is yet not available. Waiting for first snapshot]

```

### Steps to reproduce the bug

Example run https://github.com/solo-io/gloo/runs/7776341150?check_suite_focus=true
I've seen this at least twice today.

### Expected Behavior

CI should pass

Issue #6493, opened 2022-05-23
CI test flake: Kube2e: gateway tests with virtual service with a mix of valid and invalid virtual services


elcasteel:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

Error: 
```
BeforeEach] propagates the valid virtual services to envoy 
/home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:707

  Expected an error, got nil

  /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:668
```
These tests enable and disable strict validation and sometimes the updated setting is not propagated by the time we check to make sure it is enabled. 

The test tries to write an invalid resource and expects it to fail so I think if we wanted to fix it with an `Eventually` we would need to do something like:
```
Eventually:
    try to write the invalid resource
    if rejected: success
    if accepted: delete the resource, retry
```

### Steps to reproduce the bug

Run Ci

### Expected Behavior

deterministic tests

jenshu:
https://github.com/solo-io/gloo/runs/7339699207?check_suite_focus=true

elcasteel:
https://github.com/solo-io/gloo/runs/7399154173?check_suite_focus=true

ashishb-solo:
https://github.com/solo-io/gloo/runs/7452064700?check_suite_focus=true

```
2022-07-21T15:23:45.1269030Z [Fail] Kube2e: gateway tests with virtual service with a mix of valid and invalid virtual services [BeforeEach] propagates the valid virtual services to envoy 
2022-07-21T15:23:45.1273642Z /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:660
```

PR #7067, opened 2022-08-31
CI Stability: Reduce Gateway Kube2e Test Flakes

sam-heilbron:
# Description

Improve our Gateway Kube2e tests to increase speed and reduce flakes

# Context

### Type 1: Test Pollution
Test pollution is the source of some flakes in our tests. This occurs when one test creates resources, applies it to the cluster, and then doesn't remove them after the test completes.

For example, vs-east is created with domain=* and then in a new test, vs-west is created with domain=*. If the two tests are completely isolated, there will be no issue. However, if vs-east is not cleaned up, then when vs-west is applied, it will be rejected by the webhook due to https://github.com/solo-io/gloo/blob/e545d7c1fa3d182f333eb51bbe0b37068ad34654/projects/gateway/pkg/translator/virtual_service_translator.go#L32.

This type of pollution may not impact tests every time because they execute in a random order.

To avoid, tests no longer write resources (at least in the gateway kube tests). Instead, tests just add resources to a snapshot in a BeforeEach, and that snapshot is applied before the test, and deleted after the test. That way, each test operates against a clean cluster.

### Type 2: Eventual Consistency
There are certain actions which should be work immediately (a Write succeeds) and others that should work eventually (a status on a resource is updated).

Most cases of eventual consistency were already handled.

However, one case that was not was around our High Availability support. When the Gloo component is scaled to 0 (as happens in some of our tests), and then scaled back up, leader election is executed to determine the new leader. However, the old lease will remain for the 'lease duration' before a new candidate can be elected. Therefore, this is a period of time where the Gloo controller is running but not performing leader duties (ie writing statuses).
Our tests would flake because after any tests that involve scaling down/up Gloo, the status updates weren't occurring. 

The solution, was to add a step where we write/update a placeholder resource consistently until a status is identified. This way we know that a new leader is elected and the test can proceed.

## Speed Increase
By removing the recursive ginkgo flag (-r, ./...), ginkgo no longer spends time compiling packages that are irrelevant to the kube2e test suite.

Regression tests take 7-10 minutes shorter as a result.

## Testing/Validation
I pushed 6 empty commits to this branch, and confirmed that all pass CI.

## Follow-Up Work
- Apply the Snapshot writing framework to our other regression tests. They are not as flaky, but would be valuable to have consistency across how our tests are written
- Determine if there are sources of optimization for the Helm regression test suite. This currently takes ~10 minutes longer than all others. One idea, would be to move any upgrade/downgrade tests into a suite that runs nightly
- E2E (in memory) test improvements

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6988
resolves https://github.com/solo-io/gloo/issues/7038
resolves https://github.com/solo-io/gloo/issues/6932
resolves https://github.com/solo-io/gloo/issues/6493
resolves https://github.com/solo-io/gloo/issues/6472
resolves https://github.com/solo-io/gloo/issues/5149
resolves https://github.com/solo-io/gloo/issues/5495
resolves https://github.com/solo-io/gloo/issues/2484


test/kube2e/gateway/gateway_test.go
```diff
@@ -387,138 +397,37 @@ var _ = Describe("Kube2e: gateway", func() {
 			Entry("can route to upstreams", false),
 			Entry("can route to upstreams with compressed proxy", true))
 
-		Context("routing directly to kubernetes services", func() {
```
sam-heilbron, line 390: This test already existed in the "can route to upstream": we create a VS to a kube service, and then route requests to that service

test/kube2e/gateway/gateway_test.go
```diff
@@ -2458,82 +2365,6 @@ spec:
 		})
 	})
 
-	Context("compressed spec is working", func() {
```
sam-heilbron, line 2461: We have a can route to upstream test, which runs against both compressedProxySpec = {false,true}

.github/workflows/regression-tests.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       matrix:
         # knative support has been deprecated: https://github.com/solo-io/gloo/issues/5707
         # We have removed it from our CI regression tests
-        kube-e2e-test-type: ['gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'glooctl']
+        kube-e2e-test-type: ['gateway']
```
elcasteel, line 67: Calling out what should be a temporary change so we don't forget to revert

Makefile
```diff
@@ -159,22 +159,28 @@ install-go-tools: mod-download
...
+	$(DEPSGOBIN)/ginkgo -ldflags=$(LDFLAGS) -r -failFast -trace -progress -race -compilers=4 -failOnPending -noColor $(TEST_PKG)
 endif
 
 .PHONY: run-ci-regression-tests
-run-ci-regression-tests: TEST_PKG=./test/kube2e/...
+run-ci-regression-tests: TEST_PKG=./test/kube2e
 run-ci-regression-tests: RUN_REGRESSION_TESTS=true
-run-ci-regression-tests: install-go-tools run-tests
+run-ci-regression-tests: install-test-tools
+	# We intentionally leave out the `-r` ginkgo flag, since we are specifying the exact package that we want run
```
gunnar-solo, line 182: this is a very nice find

test/helpers/snapshot_writer.go
```diff
@@ -0,0 +1,206 @@
...
+		return true
+	}
+
+	// When we apply a Snapshot, parents resources may fail due to child resources still being created
+	// To get around this we retry applying the entire snapshot, but some resources may already exist
+	return errors.IsExist(writeError)
+}
+
+// DeleteSnapshot deletes all resources in the ApiSnapshot from the cache
+func (s snapshotWriterImpl) DeleteSnapshot(snapshot *gloosnapshot.ApiSnapshot, deleteOptions clients.DeleteOpts) error {
```
gunnar-solo, line 131: "everything must go".

gunnar-solo (commented):
II see a lot of work around test helpers in general.  I _assume_ this is to combat some bad test patterns that leave pollution around.

Do you find any _particular_ pattern(s) that was causing strife?

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -310,7 +310,9 @@ func (v *VirtualServiceTranslator) mergeDelegatedVirtualHostOptions(vs *v1.Virtu
 	for _, optionRef := range optionRefs {
 		vhOption, err := options.Find(optionRef.GetNamespace(), optionRef.GetName())
 		if err != nil {
-			reports.AddError(vs, err)
+			// missing refs should only result in a warning
+			// this allows resources to be applied asynchronously
+			reports.AddWarning(vs, err.Error())
```
elcasteel, line 315: Will this cause us to accept resources that may have previously been rejected? If so, does it warrant a particular call out in the changelog or other documentation?

sam-heilbron: Yeah it's a good callout. I went back and forth on whether to call this out.

I ended up copying the logic (warning instead of error, and the exact comment from our route table delegation code), so that this delegation logic would behave the same way. I think this is the behavior we intended to make originally, so that eventual consistency works. This error/warning was actually contributing to flakes in our tests as well, because we needed to ensure that the new CR was in the snapshot before applying resources that referenced that CR.

The reason I opted to not call out this change is because a previously rejected resource (referencing a CR that was not yet in the snapshot) will now be accepted, but with the same outcome (the config from that CR will not be applied). 

However, even though this is the preferred behavior for users, it still may be valuable to call out in the changelog. What do you think @elcasteel ?

sam-heilbron: Opted to include it since we're better off being loud rather than not: https://github.com/solo-io/gloo/pull/7067/commits/d99affcdf6f9c06108efaac85bb90103504623aa

Makefile
```diff
@@ -159,22 +159,28 @@ install-go-tools: mod-download
...
+	RUN_REGRESSION_TESTS=$(RUN_REGRESSION_TESTS) $(DEPSGOBIN)/ginkgo -ldflags=$(LDFLAGS) -failFast -trace -progress -race -compilers=4 -failOnPending -noColor $(TEST_PKG)/$(KUBE2E_TESTS)
```
kdorosh, line 183: does it take this long when we use `-r` for the unit tests? curious if this is the best solution to the problem, or if any alternatives were considered

changelog/v1.13.0-beta12/ci-gateway-kube-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+    issueLink: https://github.com/solo-io/gloo/issues/6493
+    description: Fix kube2e test "gateway tests with virtual service with a mix of valid and invalid virtual services"
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/6472
+    description: Ensure tests create VirtualServices without wildcard domains
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/5149
+    description: Fix kube2e test "can modify upstreams repeatedly, and endpoints don't lag via EDS"
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/5495
```
kdorosh, line 21: nit; can we change test fixes to non user facing?

projects/gateway/pkg/translator/converter.go
```diff
@@ -199,7 +199,9 @@ func (rv *routeVisitor) visit(
 		for _, optionRef := range optionRefs {
 			routeOpts, err := reporterHelper.snapshot.RouteOptions.Find(optionRef.GetNamespace(), optionRef.GetName())
 			if err != nil {
-				reporterHelper.addError(resource.InputResource(), err)
+				// missing refs should only result in a warning
+				// this allows resources to be applied asynchronously
```
kdorosh, line 203: how does this allow resources to be applied asynchronously? you mean just in the case when validation is set to allow warnings but reject errors?

test/helpers/snapshot_writer.go
```diff
@@ -0,0 +1,206 @@
...
+
+	for {
+		// to account for writing latency, we inject a back-off strategy for retrying the snapshot write
+		mostRecentResult := s.doWriteSnapshot(snapshot, writeOptions)
+		if mostRecentResult == nil {
+			return nil
+		}
+		logger.Warnf("Failed to write snapshot on attempt #%d. Reason: %+v", attempt, mostRecentResult)
+
+		shouldContinue := s.backoffStrategy(attempt)
```
kdorosh, line 43: have we considered something like https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/plugins/consul/eds.go#L339-L356 to make this cleaner?

projects/gateway/pkg/translator/converter.go
sam-heilbron: yeah its a good callout I can can certainly improve the comment. tbh, I copied the comment from our route table delegation code and was too lazy to make more accurate. I can update all instances to clarify this behavioral difference since I think its valuable to codify

test/helpers/snapshot_writer.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (s snapshotWriterImpl) isContinuableWriteError(writeError error) bool {
+	if writeError == nil {
+		return true
+	}
+
+	// When we apply a Snapshot, parents resources may fail due to child resources still being created
+	// To get around this we retry applying the entire snapshot, but some resources may already exist
+	return errors.IsExist(writeError)
+}
```
kdorosh, line 128: do we want anything for resource version errors?

test/helpers/snapshot_writer.go
sam-heilbron: Ohh I like it! I had not, will update

Makefile
sam-heilbron: I didn't dig too deep into the unit tests, though i imagine some improvements around skipping certain packages from being compiled, and using other ginkgo flags to speed up compilation would be valuable. Since these tests only run in a single package, this seemed to be the most efficient solution for these tests. Did you have anythign else in mind?

test/helpers/virtual_services.go
```diff
@@ -65,34 +83,111 @@ func (b *virtualServiceBuilder) getOrDefaultRoute(routeName string) *v1.Route {
...
+	}
+	return nil
+}
+
 func (b *virtualServiceBuilder) Build() *v1.VirtualService {
+	if err := b.errorIfInvalid(); err != nil {
+		// We error loudly here
+		// These types of errors are intended to prevent developers from creating resources
+		// which are semantically correct, but lead to test flakes/confusion
+		log.Panic(err)
```
kdorosh, line 188: prefer `Fail()`?

test/kube2e/gateway/gateway_suite_test.go
```diff
@@ -92,6 +106,22 @@ func StartTestHelper() {
...
+	resourceClientset, err = kube2e.NewKubeResourceClientSet(ctx, cfg)
+	Expect(err).NotTo(HaveOccurred())
+
+	snapshotWriter = helpers.NewSnapshotWriter(resourceClientset, func(attempt int) bool {
+		if attempt > 3 {
+			// only retry 3 times
+			return false
+		}
+
+		time.Sleep(time.Duration(attempt^2) * time.Second)
```
kdorosh, line 122: same nit about retry library

test/helpers/snapshot_writer.go
sam-heilbron: I opted not to since I think if we're getting version errors, it means somethign else is going wrong

Makefile
sam-heilbron: Yeah I like that approach for improving the unit tests

test/helpers/virtual_services.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7067/commits/db68455fb550b82b3289ea86b4b5f769f08f6751

test/helpers/snapshot_writer.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7067/commits/db68455fb550b82b3289ea86b4b5f769f08f6751

projects/gateway/pkg/translator/converter.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7067/commits/badea643770b582e92baf4cbf2b96836384e4c77

changelog/v1.13.0-beta12/ci-gateway-kube-tests.yaml
sam-heilbron: https://github.com/solo-io/gloo/pull/7067/commits/0086d75f9a588f4be26d74db3d44bf6598c58d43

Makefile
kdorosh: my only objection to this plan is ensuring that the new command here fails if it finds zero suites (e.g. new dev forgets to ensure file and test names match). i just want to avoid the situation where ci passes but runs zero suites

test/helpers/snapshot_writer.go
```diff
@@ -0,0 +1,203 @@
...
+		retryOptions:      append(defaultRetryOptions, retryOptions...),
+	}
+}
+
+// WriteSnapshot writes all resources in the ApiSnapshot to the cache
+func (s snapshotWriterImpl) WriteSnapshot(snapshot *gloosnapshot.ApiSnapshot, writeOptions clients.WriteOpts) error {
+	return retry.Do(func() error {
+		if writeOptions.Ctx.Err() != nil {
+			// intentionally return early if context is already done
+			// this is a backoff loop; by the time we get here ctx may be done
```
kdorosh, line 46: i think if we want to take this approach then we should return nil for error

kdorosh: otherwise we should add this logic / check to the `RetryIf`

sam-heilbron: ohhhh you're totally right!!

sam-heilbron: https://github.com/solo-io/gloo/pull/7067/commits/502c5b9441ac63c9ed42a88a3f6076d6ce0a22ba

Makefile
sam-heilbron: I think we're actually protected against this.

When I originally added this change, the gloo mTLS suite failed with: https://github.com/solo-io/gloo/runs/8165246152?check_suite_focus=true


Commit before error: https://github.com/solo-io/gloo/pull/7067/commits/2ca26a9576c977b915c76b4eb00a9dc9186052c9
Commit fixing error: https://github.com/solo-io/gloo/pull/7067/commits/d148a36b203b8ca8cebb25977902a7f5d7710a8f

so when we encounter no test suites, ginkgo errors for us

kdorosh: sounds good. just saving the output of the test above for posterity:
```shell
 GOBIN=/home/runner/work/gloo/gloo/_output/.bin go install github.com/onsi/ginkgo/ginkgo
# We intentionally leave out the `-r` ginkgo flag, since we are specifying the exact package that we want run
RUN_REGRESSION_TESTS=true /home/runner/work/gloo/gloo/_output/.bin/ginkgo -ldflags="-X github.com/solo-io/gloo/pkg/version.Version=0.0.0-fork" -failFast -trace -progress -race -compilers=4 -failOnPending -noColor ./test/kube2e/gloomtls
Found no test suites
For usage instructions:
	ginkgo help
make: *** [Makefile:183: run-ci-regression-tests] Error 1
Error: Process completed with exit code 2.
```

kdorosh (commented):
love this approach, lgtm!

gunnar-solo (approved):
Make it so!

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a79fed23-214a-4a76-97b9-7c189a416f18

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4494eb48-53d5-4947-a17f-342dec33f14a

PR #7069, opened 2022-09-01
Make sure plugins use `Init()` correctly

saiskee:


elcasteel (approved):
LGTM 
I assume we need a similar check on the plugins defined in solo-projects

PR #7074, opened 2022-09-02
11 | validation: Update a couple debug logs that can be large with large p…

gunnar-solo:
# Description


nfuden (approved):

gunnar-solo:
/kick [[Fail] RetryOnUnavailableClientConstructor [It] creates a new client and closes the old one](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=2dabd8cf-9abd-47fc-865f-da0c5ed54a9c)

PR #7083, opened 2022-09-02
Use annotated gloo pod to get OSS image tag

elcasteel:
# Description

Export the functionality from the glooctl client/server version mismatch check and use it in the `glooctl istio inject` implementation to get the correct tag value for the SDS sidecar. 
I updated a unit test as part of this change to show that we need to check pods with the names "gloo" and "gloo-ee." A regression test for this type of behavior would need to be in solo-projects.

# Context

Previously this worked by checking the tag on the gateway pod but since it was removed there aren't any deployments in enterprise environments that are built as part of OSS releases. 

# Checklist:


projects/gloo/cli/pkg/cmd/istio/util.go
```diff
@@ -116,30 +112,23 @@ func getJWTPolicy(pilotContainer corev1.Container) string {
 // getGlooVersion gets the version of gloo currently running
 // in the given namespace, by checking the gloo deployment.
 func getGlooVersion(ctx context.Context, namespace string) (string, error) {
-	client := helpers.MustKubeClient()
-	_, err := client.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	sv := versioncmd.NewKube(namespace, "")
+	clientServerVersions, err := versioncmd.GetClientServerVersions(ctx, sv)
```
bewebi, line 116: (Why) do we need to call `GetClientServerVersions()`? 
We only use `GetServer()` from the returned `Version` object, which is populated from `sv.Get()`
Since we have `sv` it seems like we could call `sv.Get()` directly?

Makefile
```diff
@@ -34,7 +34,7 @@ ORIGIN_URL ?= $(shell git remote get-url origin)
 UPSTREAM_ORIGIN_URL ?= git@github.com:solo-io/gloo.git
 UPSTREAM_ORIGIN_URL_HTTPS ?= https://www.github.com/solo-io/gloo.git
 ifeq ($(filter "$(ORIGIN_URL)", "$(UPSTREAM_ORIGIN_URL)" "$(UPSTREAM_ORIGIN_URL_HTTPS)"),)
-	VERSION := 0.0.0-fork
+	VERSION := 0.0.1-fork
```
bewebi, line 37: (Why) is this a necessary change?

projects/gloo/cli/pkg/cmd/istio/util.go
elcasteel: Updated - I don't think there's any particular future value in using `GetClientServerVersions()` as a layer of abstraction.

Makefile
elcasteel: Our version checking code doesn't treat `0.0.0-` as a valid version so this was failing the tests I added.

Issue #6820, opened 2022-07-28
e2e test: CSRF defined on listener only on listener [It] should succeed with allowed origin


ashishb-solo:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

e2e test flake in CSRF module:

```
Step #10 - "test": Summarizing 1 Failure:
Step #10 - "test": 
Step #10 - "test": [Fail] CSRF defined on listener only on listener [It] should succeed with allowed origin 
Step #10 - "test": /workspace/gloo/test/e2e/csrf_test.go:215
Step #10 - "test": 
Step #10 - "test": Ran 37 of 157 Specs in 157.814 seconds
Step #10 - "test": FAIL! -- 36 Passed | 1 Failed | 0 Pending | 120 Skipped
Step #10 - "test": --- FAIL: TestE2e (160.67s)
Step #10 - "test": FAIL
Step #10 - "test": 
Step #10 - "test": Ginkgo ran 127 suites in 22m58.47111392s
```

[Full build log](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=836ec9c7-071b-4fb5-93bc-545f95842de1)

### Steps to reproduce the bug

Run CI tests on a PR.

### Expected Behavior

It should pass deterministically

Issue #6819, opened 2022-07-28
CI Test Flake: [It] Should be able to rotate a secret referenced on a sslConfig on a kube upstream 


inFocus7:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

Test flake in during CI's `CI / k8s regression tests (gloo, false)` job.

```
• Failure [280.034 seconds]
GlooResourcesTest rotating secrets on upstream sslConfig [It] Should be able to rotate a secret referenced on a sslConfig on a kube upstream 
/home/runner/work/gloo/gloo/test/kube2e/gloo/gloo_resources_test.go:253

  Expected
      <string>: no healthy upstream
  to contain substring
      <string>: Hello, world!

  /home/runner/work/gloo/gloo/test/kube2e/gloo/gloo_resources_test.go:284
```

### Steps to reproduce the bug

Run CI tests on a PR.

### Expected Behavior

It should pass.

### Additional Context

- [suite log](https://github.com/solo-io/gloo/runs/7563853161?check_suite_focus=true)

elcasteel:
another instance of this https://github.com/solo-io/gloo/runs/8005352415?check_suite_focus=true

PR #7102, opened 2022-09-07
CI Stability: Use Snapshot Model In Other Kube2e Tests

sam-heilbron:
# Description

Follow up to: https://github.com/solo-io/gloo/pull/7067

# Context

Apply the snapshot model, introduced in our earlier PR, to the other kube end-to-end test suites

### Resources Test
The Gloo kube tests attempt to setup a VS that routes traffic to an Upstream, with traffic encrypted by keys defined in a secret.
We previously wrote a bunch of resources (pods, service accounts, secrets...etc). 
I re-used the TestRunner service, which we deploy at the beginning of all kube tests, and added a simple VS to route to this Service.

I think this simplifies the test, because it ensures we create the minimum required resources.

### Setup Syncer Tests
Tests previously had both kube tests and non kube tests. The kube tests only run if you provide the RUN_KUBE_TESTS env variable, which we do not do in our Github actions. These tests therefore have not been running.

I decided to move any non-kube tests out of this suite, and into the package co-located with the code it is unit testing.

# CSRF Tests
I wrapped the entire check (request succeeds, metrics are what we expect) in an Eventualy, to provide startup time. This seemed to solve the flake for me. Though I couldn't reliably produce the flake locally so I can't be certain.

# Follow-Up
- Improve our Helm tests, which currently take the longest to run, because they install/uninstall gloo on each iteration, even when its unnecessary.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6819
resolves https://github.com/solo-io/gloo/issues/6820


test/kube2e/gloo/gloo_resources_test.go
```diff
@@ -1,264 +1,121 @@
...
+
+			// Modify the VirtualService to include the necessary SslConfig
+			testRunnerVs.SslConfig = &gloov1.SslConfig{
+				SslSecrets: &gloov1.SslConfig_SecretRef{
+					SecretRef: &core.ResourceRef{
+						Name:      tlsSecret.GetName(),
+						Namespace: tlsSecret.GetNamespace(),
+					},
+				},
+			}
```
jackstine, line 107: This particular error occurred when attaching the sslConfig to the Upstream which is connected to the Virtual Service. I don't think this is producing the exact behavior that we were initially looking for.

sam-heilbron: oh good catch! 
Out of curiosity what is the difference between the two? Aren't we confirming that we can change data in a resource, and that dependents get updated appropriately?

sam-heilbron:
Build-bot flaked with https://github.com/solo-io/gloo/issues/6820, I have an idea for how to fix and will try to resolve as part of this PR.

test/kube2e/gateway/gateway_test.go
```diff
@@ -2282,31 +2227,6 @@ func getRouteWithDelegateRef(delegate string, path string) *gatewayv1.Route {
...
-		Action: &gatewayv1.Route_DelegateAction{
-			DelegateAction: &gatewayv1.DelegateAction{
-				DelegationType: &gatewayv1.DelegateAction_Selector{
-					Selector: &gatewayv1.RouteTableSelector{
-						Namespaces: []string{testHelper.InstallNamespace},
-						Labels:     labels,
-					},
-				},
-			},
-		},
```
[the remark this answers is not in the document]
sam-heilbron: Yeah I agree. I've been trying to conslidate them in: https://github.com/solo-io/gloo/blob/master/test/helpers/virtual_services.go

test/kube2e/gateway/gateway_test.go
```diff
@@ -2030,43 +2033,29 @@ spec:
...
+					testRunnerVs.GetMetadata().GetNamespace(),
+					testRunnerVs.GetMetadata().GetName(),
+					clients.ReadOpts{Ctx: ctx})
+				Expect(err).NotTo(HaveOccurred())
 
-				err := resourceClientset.VirtualServiceClient().Delete(vs.Metadata.Namespace, vs.Metadata.Name, clients.DeleteOpts{Ctx: ctx})
-				Expect(err).ToNot(HaveOccurred())
+				vs.VirtualHost.Options = &gloov1.VirtualHostOptions{Transformations: invalidTransform}
+				_, err = resourceClientset.VirtualServiceClient().Write(vs, clients.WriteOpts{Ctx: ctx, OverwriteExisting: true})
+				Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
sam-heilbron: I like it!

test/kube2e/gloo/gloo_resources_test.go
jackstine: I am not sure what the VS translation creates itself, but the Upstream should translate the need for the Envoy Cluster and the SSLConfig creates the Envoy Cluster Load Assignment (CLA). The behavior that we need to test is that the update for the Envoy Cluster occurs at an earlier time than the CLA, else this will cause the upstream to be unhealthy.

test/kube2e/gloo/gloo_resources_test.go
```diff
@@ -96,20 +98,56 @@ var _ = Describe("GlooResourcesTest", func() {
...
+			upstreamSslConfigString, err := json.Marshal(upstreamSslConfig)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Annotate the kube service, so that discovery applies the ssl configuration to the generated upstream")
+			Eventually(func(g Gomega) {
+				testRunnerService, err := resourceClientset.KubeClients().CoreV1().Services(testHelper.InstallNamespace).Get(ctx, helper.TestrunnerName, metav1.GetOptions{})
+				g.Expect(err).NotTo(HaveOccurred())
+
+				testRunnerService.Annotations[serviceconverter.DeepMergeAnnotationPrefix] = "true"
+				testRunnerService.Annotations[serviceconverter.GlooAnnotationPrefix] = fmt.Sprintf("{sslConfig: %s}", upstreamSslConfigString)
```
jackstine, line 118: ok so this is creating a discovered upstream, with the annotations of the `sslConfig` added to it. 
As described [here](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/destination_types/discovered_upstream/discovered-upstream-configuration/).

I assume this is ok, since the upstream is created by the discovery of the kube resource.

jackstine: thanks for the share

ianmacclancy (approved):

PR #7118, opened 2022-09-08
Fix gloo-resource-cleanup failure when CRD doesn't exist

jenshu:
Exit early from the `gloo-resource-cleanup` job (instead of erroring) if the Gloo Edge CRDs have already been deleted.

### DIFF
--- a/changelog/v1.13.0-beta13/cleanup-ignore-crd-errors.yaml
+++ b/changelog/v1.13.0-beta13/cleanup-ignore-crd-errors.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7109
+    resolvesIssue: false
+    description: Exit early from the `gloo-resource-cleanup` job (instead of erroring) if the Gloo Edge CRDs have already been deleted.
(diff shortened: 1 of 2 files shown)


install/helm/gloo/templates/5-resource-cleanup-job.yaml
```diff
@@ -45,6 +45,12 @@ spec:
           - -c
           - |
             kubectl delete --ignore-not-found validatingwebhookconfigurations.admissionregistration.k8s.io gloo-gateway-validation-webhook-{{ .Release.Namespace }} || exit $?
+
+            if ! kubectl get crd upstreams.gloo.solo.io; then
```
jenshu, line 49: added a similar check to the one we have in the [migration job](https://github.com/solo-io/gloo/blob/de0861a5642fce08649a16c33185aec605f885d5/install/helm/gloo/templates/5-resource-migration-job.yaml#L47-L50)

changelog/v1.13.0-beta13/cleanup-ignore-crd-errors.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7109
+    resolvesIssue: false
+    description: Exit early from the `gloo-resource-cleanup` job (instead of erroring) if the Gloo Edge CRDs have already been deleted.
```
sam-heilbron, line 5: When does the resource cleanup run and when does this situation occur?

I'm curious why CRDs would be deleted before the job runs

sam-heilbron (approved):

changelog/v1.13.0-beta13/cleanup-ignore-crd-errors.yaml
jenshu: (spoke offline: this job is run on helm uninstall; a customer ran into an error when they had deleted the crds before uninstalling)

PR #7146, opened 2022-09-14
Reduce Clone() calls in translation

elcasteel:
# Description
I noticed that `proto.Clone` was a hotspot in a couple of spots on a translation flame graph. 
One of the places was in the route_replacing_sanitizer where it's ok to modify the snapshot (it's run in sequence with the upstream sanitizer which also modifies the snapshot). 
I also found a place where we cloned a virtual host but didn't modify anything except to sanitize the name. I updated this to only clone when the name is updated.

# Context

# Checklist:


projects/gateway/pkg/translator/merge.go
```diff
@@ -95,38 +95,35 @@ func isEmptyValue(v reflect.Value) bool {
...
 	return childClone
 }
 
 func mergeHCMSettings(parent, child *hcm.HttpConnectionManagerSettings, preventChildOverrides bool) *hcm.HttpConnectionManagerSettings {
 	// Clone to be safe, since we will mutate it
-	parentClone := proto.Clone(parent).(*hcm.HttpConnectionManagerSettings)
-	childClone := proto.Clone(child).(*hcm.HttpConnectionManagerSettings)
-
-	if childClone == nil {
+	if child == nil {
```
elcasteel, line 115: This is rearranged so we avoid cloning resources that won't be used

projects/gateway/pkg/translator/merge.go
```diff
@@ -95,38 +95,35 @@ func isEmptyValue(v reflect.Value) bool {
...
 		// use parent exactly as-is
-		return parentClone
+		return proto.Clone(parent).(*hcm.HttpConnectionManagerSettings)
```
elcasteel, line 117: I'm not 100% sure these are necessary, we could remove them if we can show that the returned result isn't modified later.

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
```diff
@@ -301,11 +299,11 @@ func (s *RouteReplacingSanitizer) replaceRoutes(
 				}
 				vh.GetRoutes()[j] = route
 			}
-			sanitizedRouteConfig.GetVirtualHosts()[i] = vh
+			cfg.GetVirtualHosts()[i] = vh
```
kdorosh, line 302: this is similar in spirit to what we do with the route with `vh.GetRoutes()[j] = route` a couple lines before. now that we are updating the route on the resource during the loop, i dont think we even need this line at all (this line basically says `vh = vh`)

projects/gloo/pkg/translator/translator.go
```diff
@@ -312,15 +312,19 @@ func MustEnvoyCacheResourcesListToHash(resources []envoycache.Resource) uint64 {
...
+func MakeRdsResources(routeConfigs []*envoy_config_route_v3.RouteConfiguration, cloneRouteConfigs bool) envoycache.Resources {
 	var routesProto []envoycache.Resource
 
 	for _, routeCfg := range routeConfigs {
 		// don't add empty route configs, envoy will complain
 		if len(routeCfg.GetVirtualHosts()) < 1 {
 			continue
 		}
-		routesProto = append(routesProto, resource.NewEnvoyResource(proto.Clone(routeCfg)))
+		if cloneRouteConfigs {
```
kdorosh, line 323: a comment here would be nice explaining why we do this

kdorosh: although i don't think we need the clone here at all, why don't we just remove it?

kdorosh: we also do this in `generateXdsSnapshot`:

```golang
	for _, ep := range endpoints {
		endpointsProto = append(endpointsProto, resource.NewEnvoyResource(proto.Clone(ep)))
	}
	for _, cluster := range clusters {
		clustersProto = append(clustersProto, resource.NewEnvoyResource(proto.Clone(cluster)))
	}
	for _, listener := range listeners {
		// don't add empty listeners, envoy will complain
		if len(listener.GetFilterChains()) < 1 {
			continue
		}
		listenersProto = append(listenersProto, resource.NewEnvoyResource(proto.Clone(listener)))
	}
```

and can make similar optimizations there

elcasteel: Is it ok if the original proxy here gets mutated? https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/envoy_translator_syncer.go#L129
Although I guess it might be faster to clone the proxy once rather than clone all of these resources

kdorosh: unrelated but can we fix this while we are here https://github.com/solo-io/gloo/blob/523361fe08a7793a8f4943feca71673618bbfde7/projects/gloo/pkg/translator/translator.go#L260-L263

kdorosh: lets hop on a call to discuss whether it's safe; i think so, but not sure what you mean by the original proxy

elcasteel: what's the fix needed for `// don't add empty listeners ...` ?

kdorosh: oh brain fog.. i thought that said less than zero not less than one.. :facepalm:

projects/gateway/pkg/translator/merge.go
elcasteel: Regardless this isn't a hotspot right now. I found it by searching for `proto.Clone` although it's something we can keep in mind for future optimizations.

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
elcasteel: yeah, I was able to delete both those lines.

kdorosh (commented):
lgtm

kdorosh (approved):

PR #7104, opened 2022-09-07
[1.12] instantiate plugins once

saiskee:
…ook (#7080)

* add changelog

* update

* Revert "Restore PluginRegistryFactory (#6919)"

This reverts commit 4af358588dfe5c8291e6754b0e0e2a040278e0fe.

* goimports

* update

* goimports

* deflake robust_client_test

* add back validation translator

* goimpotrs

* add changelog

* yaml

* fix changelog

* update

* goimports

* update comment and uncomment validation translator

Co-authored-by: soloio-bulldozer[bot] <48420018+soloio-bulldozer[bot]@users.noreply.github.com>

# Description

Please include a summary of the changes.

This bug fixes ... \ This new feature can be used to ...

# Context

Users ran into this bug doing ... \ Users needed this feature to ...

# Testing

Scale tests were run to ensure that no memory leaks would result as having long-living plugins. Results are reported as below

Scale testing was done over a 2 hour time frame, in which a 100 httpbin services and deployments were created. During this time, memory usage was monitored and no continuous memory leak was detected. 
# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #7164, opened 2022-09-16
[1.12] Use init correctly backport

saiskee:
# Description

When backporting the change to instantiate plugins only once (https://github.com/solo-io/gloo/pull/7104), I needed to backport the fix in which we make sure that all plugins are using `Init()` correctly, as this resets the state of the plugins. This change mitigates a leak in which the grpc plugin's state is never reset on `Init()`.
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7165


changelog/v1.12.18/plugins-use-init.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+- type: NON_USER_FACING
+  description: "Small change to make sure that plugins are using the Init method in the plugin interface correctly and resetting fields."
```
kdorosh, line 3: user-facing description?

elcasteel (approved):

PR #7172, opened 2022-09-17
ratelimit_config: Update to no longer panic and only add a todo

nfuden:
Remove panic for now to restore operations.
Glooctl check will always pass as has been the case since namespaced statuses existed for rate limit configs.


jbohanon (approved):
LGTM

PR #7180, opened 2022-09-19
gateway/translator:  Update grouping for ssl configs to emit stable order

nfuden:
For SSL isolation we emit a series of filterchains that are entirely separate. To do this we have a helper function that creates series of virtual services per ssl config. Today this is returned as a map which then will be iterated over to generate final httpfilter chains. 
Given that all our translation needs to create a stable snapshot (for its hash) this needs to be set to be ordered.
In this pr we change the base function to emit an ordering which can be followed to generate filterchains or whatever else needs this information in a stable format.

Testing for this change is implemented in a unit test whereby we create a snapshot with a Gateway that we expect to generate the multiple filterchains, then call `ComputeListener` on that config repeatedly to ensure the ordering remains consistent.


projects/gateway/pkg/translator/aggregate_translator_test.go
```diff
@@ -0,0 +1,94 @@
...
+		})
+		genProxyWithIsolatedVirtualHosts()
+		proxyName := proxy.Metadata.Name
+		aggregateTranslator := &AggregateTranslator{VirtualServiceTranslator: &VirtualServiceTranslator{}}
+		// run 100 times to ensure idempotency
+		// not sure if 100 times is valid; in anecdotal testing it tended to fail in under 20
+		for i := 0; i < 100; i++ {
+			l := aggregateTranslator.ComputeListener(NewTranslatorParams(ctx, snap, reports), proxyName, snap.Gateways[0])
+			Expect(l).NotTo(BeNil())
+			Expect(l.GetAggregateListener().HttpFilterChains[0].GetMatcher().GetSslConfig().GetSniDomains()[0]).To(Equal("sni"))
```
jenshu, line 91: do we also want to test the values from the other filter chains?

projects/gateway/pkg/translator/aggregate_translator.go
```diff
@@ -70,12 +70,13 @@ func (a *AggregateTranslator) computeAggregateListenerForHttpGateway(params Para
 	builder := newBuilder()
 	if gateway.GetSsl() {
 		// for an ssl gateway, create an HttpFilterChain per unique SslConfig
-		virtualServicesBySslConfig := groupVirtualServicesBySslConfig(virtualServices)
-		for sslConfig, virtualServiceList := range virtualServicesBySslConfig {
+		vsOrdering, virtualServicesBySslConfig := groupVirtualServicesBySslConfig(virtualServices)
```
jenshu, line 73:
```suggestion
		orderedSslConfigs, virtualServicesBySslConfig := groupVirtualServicesBySslConfig(virtualServices)
```
nit: naming?

projects/gateway/pkg/translator/aggregate_translator_test.go
```diff
@@ -0,0 +1,94 @@
...
+	It("Computes listener idempotently when provided different ssl configs", func() {
+		gw1 := snap.Gateways[1]
+		gw := gw1.GetHttpGateway()
+		gw.VirtualServiceExpressions = nil
+		gw.VirtualServiceSelector = nil
+		gw.VirtualServices = append(gw.VirtualServices, &core.ResourceRef{
+			Name:      "ssl-vs",
+			Namespace: ns,
+		}, &core.ResourceRef{
+			Name:      "different-ssl-vs",
```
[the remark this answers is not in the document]
jbohanon: What do you mean be check that the first vs is the same resource ref? I  don't see a way to access vs resource refs on the computed listener?

projects/gateway/pkg/translator/aggregate_translator_test.go
jbohanon: @jenshu Are the changes in https://github.com/solo-io/gloo/pull/7180/commits/a60611b70d5801a25c59ca573899b06e146f7b05 what you had in mind?

projects/gateway/pkg/translator/aggregate_translator_test.go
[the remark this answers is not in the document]
nfuden: ah fair enough didnt check the accessors.

jenshu (approved):

jbohanon (approved):

PR #7181, opened 2022-09-19
ratelimit_config: Update to no longer panic and only add a todo 

nfuden:
Remove panic for now to restore operations but is explicitly a bandaid
Glooctl check will always pass as has been the case since namespaced statuses existed for rate limit configs.

This is an up-port from 1.12 which was the only other affected version as this occured due to status optimization work and partial work with namespaced statuses.

Without this gloo pod will crash once on anything that would update the rate limit cr status.


kdorosh (approved):

changelog/v1.13.0-beta16/rate-limit-status-hotfix.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/4069
```
kdorosh, line 3: nit; prefer open source links in our open source code for community users

changelog/v1.13.0-beta16/rate-limit-status-hotfix.yaml
```diff
@@ -0,0 +1,7 @@
...
+    resolvesIssue: false
+    description: |
```
kdorosh, line 5: nit; prefer `>-` formatting in changelog release notes or else we keep newlines in our release notes and generated changelogs that look silly.

these can be manually fixed in the release afterward, but id like to start pushing people to use this formatting in PRs too.

see https://github.com/solo-io/gloo/releases/tag/v1.13.0-beta16

PR #7182, opened 2022-09-19
Remove panic() calls

Bslabe123:
# Description

Removed all calls to `panic()` and replaced them with `contextutils.LoggerFrom(ctx).DPanic`.

# Context

Calling panic() can cause unexpected crashes so replacing it with a safer function will prevent a panic from bringing down a given pod.

# Checklist:


docs/content/guides/dev/example-proxy-controller.go
```diff
@@ -211,6 +212,6 @@ func makeDesiredProxy(upstreams v1.UpstreamList) *v1.Proxy {
 // make our lives easy
 func must(err error) {
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(params.Ctx).DPanic(err)
```
ben-taussig-solo, line 215: `params` doesn't exist in this context -- I also don't think it's necessary to replace these calls to `panic` in our guides

ben-taussig-solo: I think we probably want to leave this as `panic` -- I just skimmed over `example-proxy-controller.md` so I'm not completely sure of this, but I don't think that we run gloo with the logger in development mode in the example setup, so I think this change alters the behavior of the guide in a way that we probably don't want to

projects/gloo/pkg/validation/server.go
```diff
@@ -67,7 +67,7 @@ func (s *validator) shouldNotify(snap *v1snap.ApiSnapshot) bool {
 
 		hash, err := hashutils.HashAllSafe(nil, toHash...)
 		if err != nil {
-			panic("this error should never happen, as this is safe hasher")
+			contextutils.LoggerFrom(nil).DPanic("this error should never happen, as this is safe hasher")
```
ben-taussig-solo, line 70: I think we should return from here

test/e2e/aggregate_listener_test.go
```diff
@@ -214,8 +214,8 @@ var _ = Describe("Aggregate Listener", func() {
 	Context("Secure HttpGateway", func() {
 
 		var (
-			eastCert, eastPK = gloohelpers.Certificate(), gloohelpers.PrivateKey()
-			westCert, westPK = gloohelpers.GetCerts(gloohelpers.Params{
+			eastCert, eastPK    = gloohelpers.Certificate(), gloohelpers.PrivateKey()
+			westCert, westPK, _ = gloohelpers.GetCerts(gloohelpers.Params{
```
ben-taussig-solo, line 218: Any reason we don't check for an error here?

test/helpers/certs.go
```diff
@@ -164,7 +168,7 @@ var (
 
 func gencerts() {
 
-	cert, privKey = GetCerts(Params{
+	cert, privKey, _ = GetCerts(Params{
```
ben-taussig-solo, line 171: Should we check for an error here?

ben-taussig-solo (commented):
Left a couple of small comments. I have a couple of broader questions:
 - Is it necessary to make these changes in tests/examples/utilities (i.e. parts of the codebase that aren't used while the application is running)? 
   - Does this affect the behavior of the code in these locations in undesirable ways?
 - How do we make sure that we handle the `DPanic` situations appropriately in the portions of the codebase that are used while the application is running?
   - Are we returning from the function we're in in these cases? Do we return an error? If not, are we sure that the values we're returning aren't going to cause issues elsewhere (i.e., NPEs)?

docs/examples/grpc-json-transcoding/bookstore/go.mod
```diff
@@ -5,7 +5,6 @@ go 1.14
 require (
 	github.com/golang/protobuf v1.3.5
 	github.com/pkg/errors v0.8.1
-	golang.org/x/net v0.0.0-20190311183353-d8887717615a
```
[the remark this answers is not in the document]
Bslabe123: Forgot to go mod tidy after codegen, easy fix

projects/gloo/pkg/translator/translator.go
```diff
@@ -293,11 +293,11 @@ func MustEnvoyCacheResourcesListToFnvHash(resources []envoycache.Resource) uint6
 		// another path to further improve performance here.
 		out, err := mo.MarshalAppend(buf, proto.MessageV2(r.ResourceProto()))
 		if err != nil {
-			panic(errors.Wrap(err, "marshalling envoy snapshot components"))
+			contextutils.LoggerFrom(nil).DPanic(errors.Wrap(err, "marshalling envoy snapshot components"))
```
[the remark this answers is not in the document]
Bslabe123: I would but considering this is a must~~ function should we just DPanic?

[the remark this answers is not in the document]
ben-taussig-solo: In that case I think it's fine if we just return an obviously invalid value

projects/gloo/pkg/utils/any.go
```diff
@@ -44,7 +47,7 @@ func MustAnyToMessage(a *pany.Any) proto.Message {
 	var x ptypes.DynamicAny
 	err := ptypes.UnmarshalAny(a, &x)
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(nil).DPanic(err)
```
ben-taussig-solo, line 50: Same here as above -- I think we should return `nil` from here because I suspect that `return x.Message` below this will throw a runtime error if we try to call it after failing to unmarshal

test/services/gateway.go
```diff
@@ -457,15 +457,15 @@ func newServiceClient(ctx context.Context, memFactory *factory.MemoryResourceCli
 	if kube := runOpts.KubeClient; kube != nil {
 		kubeCache, err := cache.NewKubeCoreCache(ctx, kube)
 		if err != nil {
-			panic(err)
+			contextutils.LoggerFrom(ctx).DPanic(err)
```
ben-taussig-solo, line 460: I fear that the changes in this file could lead to confusing behavior in the case that our tests are misconfigured -- how do you feel about leaving these as just `panic` instead of `DPanic`?

test/v1helpers/test_grpc_service/testgrpcservice.go
```diff
@@ -33,7 +35,7 @@ import (
 func RunServer(ctx context.Context) *TestGRPCServer {
 	lis, err := net.Listen("tcp", ":0")
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(ctx).DPanic(err)
```
ben-taussig-solo, line 38: Same here as in the last test file -- I think we should consider reverting these changes to just `panic`, as this might cause unexpected behavior in our tests. Basically I just worry that we sometimes *want* the test GRPC service to panic, or at least that it doing so might be useful for a developer iterating on a set of tests that use it

test/v1helpers/test_upstream.go
```diff
@@ -167,18 +169,18 @@ func runTestServerWithHealthReply(ctx context.Context, reply, healthReply string
 
 	listener, err := net.Listen("tcp", ":0")
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(ctx).DPanic(err)
```
ben-taussig-solo, line 172: Same as the other test files -- I think we should consider reverting the changes here

docs/content/guides/dev/example-proxy-controller.go
ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/7182/commits/fc5b38b0fee0f666b8f4095214882075f36dc583

pkg/utils/settingsutil/settings.go
```diff
@@ -23,7 +25,8 @@ func FromContext(ctx context.Context) *v1.Settings {
 		return settings
 	}
 	// we should always have settings when this method is called.
-	panic("no settings on context")
+	contextutils.LoggerFrom(nil).DPanic("no settings on context")
+	return nil
```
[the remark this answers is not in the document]
ben-taussig-solo: Update: settings now returns an error

projects/discovery/pkg/fds/syncer/discovery_syncer.go
```diff
@@ -68,7 +68,8 @@ func selectUpstreamsForDiscovery(fdsMode v1.Settings_DiscoveryOptions_FdsMode, u
 	case v1.Settings_DiscoveryOptions_WHITELIST:
 		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces)
 	}
-	panic("invalid fds mode: " + fdsMode.String())
+	contextutils.LoggerFrom(nil).DPanic("invalid fds mode: " + fdsMode.String())
+	return nil
```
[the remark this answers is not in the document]
ben-taussig-solo: Update: `selectUpstreamsForDiscovery` now returns an error`

test/services/gateway.go
ben-taussig-solo: Could you do the same with the other `DPanic`s in this file?

test/v1helpers/test_grpc_service/testgrpcservice.go
ben-taussig-solo: Could you do the same with the other `DPanic`s in this file?

test/v1helpers/test_upstream.go
ben-taussig-solo: Could you do the same with the other `DPanic`s in this file?

test/services/envoy.go
```diff
@@ -275,7 +277,7 @@ func NewEnvoyFactory() (*EnvoyFactory, error) {
 
 		envoyImageTag := getEnvoyImageTag()
 		if envoyImageTag == "" {
-			panic("Must set the ENVOY_IMAGE_TAG env var. Find valid tag names here https://quay.io/repository/solo-io/gloo-envoy-wrapper?tab=tags")
+			contextutils.LoggerFrom(nil).DPanic("Must set the ENVOY_IMAGE_TAG env var. Find valid tag names here https://quay.io/repository/solo-io/gloo-envoy-wrapper?tab=tags")
```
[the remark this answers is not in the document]
ben-taussig-solo: Could you do the same with the other `DPanic`s in this file?

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -142,7 +142,7 @@ func (h *hcmNetworkFilterTranslator) ComputeNetworkFilter(params plugins.Params)
 	// 4. Generate the typedConfig for the HCM
 	hcmFilter, err := NewFilterWithTypedConfig(wellknown.HTTPConnectionManager, httpConnectionManager)
 	if err != nil {
-		panic(errors.Wrapf(err, "failed to convert proto message to struct"))
+		contextutils.LoggerFrom(nil).DPanic(errors.Wrapf(err, "failed to convert proto message to struct"))
```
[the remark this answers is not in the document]
ben-taussig-solo: `ComputeNetworkFilters` now returns an error

test/v1helpers/test_upstream.go
```diff
@@ -173,12 +175,12 @@ func runTestServerWithHealthReply(ctx context.Context, reply, healthReply string
 	addr := listener.Addr().String()
 	_, portStr, err := net.SplitHostPort(addr)
 	if err != nil {
-		panic(err)
```
ben-taussig-solo, line 176: Sorry -- missed this earlier. Can we revert to using panics here as well?

ben-taussig-solo (dismissed):
Looks good to me! Major takeaways are:
 - We're focusing on removing panics in control plane code, leaving panics in certain test clients and examples where that behavior is potentially expected or desireable
 - Control plane methods are configured to do one of the following in the former-`panic` cases:
   - return an error (in non-translation code)
   - return a nil/zero values

pkg/cliutil/nsselect/select.go
```diff
@@ -115,7 +117,7 @@ func validateResourceRefForStaticMode(typeName string, menuDescription string, r
 				return fmt.Errorf("Please specify a valid %v name. %v not found in namespace %v.", resRef.GetName(), menuDescription, resRef.GetNamespace())
 			}
 		default:
-			panic(fmt.Errorf("typename %v not recognized", typeName))
+			contextutils.LoggerFrom(nil).DPanic(fmt.Errorf("typename %v not recognized", typeName))
```
kdorosh, line 120: loggerfrom nil? this will also panic?

projects/discovery/pkg/fds/syncer/discovery_syncer.go
```diff
@@ -64,11 +68,12 @@ func selectUpstreamsForDiscovery(fdsMode v1.Settings_DiscoveryOptions_FdsMode, u
...
 	switch fdsMode {
 	case v1.Settings_DiscoveryOptions_BLACKLIST:
-		return selectUpstreamsBlacklist(upstreams, blacklistNamespaces)
+		return selectUpstreamsBlacklist(upstreams, blacklistNamespaces), nil
 	case v1.Settings_DiscoveryOptions_WHITELIST:
-		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces)
+		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces), nil
 	}
-	panic("invalid fds mode: " + fdsMode.String())
+	contextutils.LoggerFrom(nil).DPanic("invalid fds mode: " + fdsMode.String())
```
kdorosh, line 75: this will also panic (`LoggerFrom(nil)` will panic i think)? better to not log and just return the error (which itself may also be logged)

we need to see how errors are handled, not just blindly log them and return.

kdorosh (changes requested):
we need to reason about the panics more and remove all `LoggerFrom(nil)`

pkg/cliutil/nsselect/select.go
ben-taussig-solo: `LoggerFrom(nil)` is fine, it just returns the default logger. see here: https://github.com/solo-io/go-utils/blob/08d4573d2b396fb02898f1c25a08423e07912e0c/contextutils/context_and_logging.go#L120-L127

ben-taussig-solo: That being said, I can't find any examples of `LoggerFrom(nil)` in gloo or solo-projects, so I agree that we should probably change this

ben-taussig-solo: All `LoggerFrom(nil)`s have been removed, closing this conversation

projects/discovery/pkg/fds/syncer/discovery_syncer.go
ben-taussig-solo: All `LoggerFrom(nil)`s have been removed, closing this conversation

Bslabe123:
/kick Secret TLS [It] should work as with just root ca

projects/gloo/cli/pkg/cmd/create/secret/secret_test.go
```diff
@@ -191,6 +192,7 @@ metadata:
 
 			rootca := mustWriteTestFile("foo")
 			err := testutils.Glooctl("create secret tls valid --namespace gloo-system --rootca " + rootca)
+			err = testutils.Glooctl("create secret tls valid --namespace gloo-system --rootca " + rootca)
```
ben-taussig-solo, line 195: Is this intentional? Looks like it might be an accidental addition

projects/gloo/cli/pkg/cmd/version/cmd.go
```diff
@@ -168,7 +171,7 @@ func GetYaml(pb proto.Message) []byte {
 	jsn := GetJson(pb)
 	data, err := yaml.JSONToYAML(jsn)
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(context.Background()).DPanic(err)
```
ben-taussig-solo, line 174: I think we probably want to match the pattern established above and return `[]byte{}` in the error case, instead of returning `data` later on

projects/gloo/pkg/plugins/registry/registry_test.go
```diff
@@ -291,24 +292,28 @@ func TestPluginsHttpFilterUsefulness(t *testing.T) {
...
+					t.Fatalf("plugin http filter failed", err)
+					return
 				}
 				emptyListenerFilterCount += len(filters)
 			}
 
 			// Validate that the emptyListener filter count and configuredListener filter count are different
 			if emptyListenerFilterCount != len(knownBaseFilters) {
-				t.Fatalf("Found %d filters that were configured, but expected %d", emptyListenerFilterCount, len(knownBaseFilters))
+				t.Fatalf(fmt.Sprintf("Found %d filters that were configured, but expected %d", emptyListenerFilterCount, len(knownBaseFilters)))
```
ben-taussig-solo, line 310: Nit: the change here, and the one on line 315 do `t.Fatalf(fmt.Sprintf("..."))`, when `t.Fatalf("...")` accomplishes the same thing -- the inner `fmt.Sprintf()` isn't necessary

ben-taussig-solo (commented):
Overall looks pretty good! A little concerned about functions that compute hashes returning `0` -- is there appropriate error/nil checking to identify those cases and prevent us from interpreting `0` as the genuine hash value?

projects/gloo/cli/pkg/cmd/create/secret/secret_test.go
Bslabe123: Yep it is, fixing

ben-taussig-solo (dismissed):
Looks good to me. I don't love the hash functions potentially returning 0, but It does seem like we handle that case appropriately everywhere they are called, so I'm fine with it

docs/cmd/generate_docs.go
```diff
@@ -419,7 +419,10 @@ func generateSecurityScanGloo(ctx context.Context) error {
 		}
```
kdorosh, line 419: i see more instances of fatalf in this file? https://github.com/solo-io/gloo/blob/743580f2a8ff8bc8762b6d61dc88011fd51766d2/docs/cmd/generate_docs.go#L324-L328

@ben-taussig-solo while reviewing this PR, can we check out the repo and ensure nothing is broken as a form of omission?

projects/gloo/pkg/translator/translator.go
```diff
@@ -298,11 +298,13 @@ func MustEnvoyCacheResourcesListToFnvHash(resources []envoycache.Resource) uint6
...
 		if err != nil {
-			panic(errors.Wrap(err, "marshalling envoy snapshot components"))
+			contextutils.LoggerFrom(context.Background()).DPanic(errors.Wrap(err, "marshalling envoy snapshot components"))
+			return 0
 		}
 		_, err = hasher.Write(out)
 		if err != nil {
-			panic(errors.Wrap(err, "constructing hash for envoy snapshot components"))
+			contextutils.LoggerFrom(context.Background()).DPanic(errors.Wrap(err, "constructing hash for envoy snapshot components"))
+			return 0
```
kdorosh, line 307: i don't think it's safe to return 0. this is the same as an empty snapshot. we should instead update the caller to handle an error and change the xds cache key to differentiate in the case of error (and test this)

kdorosh: i.e. write a test with a buggy hasher and test it vs empty snapshot

projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -107,7 +110,7 @@ func (s *TranslatorSyncer) Sync(ctx context.Context, snap *gloov1snap.ApiSnapsho
 func (s *TranslatorSyncer) UpdateProxies(ctx context.Context) {
 	s.statusSyncer.handleUpdatedProxies(ctx)
 }
-func (s *TranslatorSyncer) GeneratedDesiredProxies(ctx context.Context, snap *gloov1snap.ApiSnapshot) (reconciler.GeneratedProxies, reconciler.InvalidProxies) {
+func (s *TranslatorSyncer) GeneratedDesiredProxies(ctx context.Context, snap *gloov1snap.ApiSnapshot) (reconciler.GeneratedProxies, reconciler.InvalidProxies, error) {
```
elcasteel, line 113: This always returns nil so I think we can remove it

projects/gloo/pkg/plugins/kubernetes/client.go
```diff
@@ -41,7 +43,7 @@ func getInformerFactory(ctx context.Context, client kubernetes.Interface, watchN
 	}
 	kubePluginSharedFactory := startInformerFactory(ctx, client, watchNamespaces)
 	if kubePluginSharedFactory.initError != nil {
-		panic(kubePluginSharedFactory.initError)
+		contextutils.LoggerFrom(context.Background()).DPanic(kubePluginSharedFactory.initError)
 	}
 	return kubePluginSharedFactory
```
elcasteel, line 48: Will we wind up with a nil pointer error later on when we try to use the factory?

Bslabe123:
```var (
	// This logger is used when there is no logger attached to the context.
	// Rather than returning nil and causing a panic, we will use the fallback
	// logger.
	fallbackLogger *zap.SugaredLogger
	// The atomic level set for any logger built here. Accessing this atomic level
	// and calling set level will change the log output dynamically.
	level zap.AtomicLevel
)```

**Review by elcasteel — APPROVED** [review]

File: docs/cmd/generate_docs.go [hunk 25]
```diff
@@ -441,6 +441,9 @@ func generateSecurityScanGlooE(ctx context.Context) error {
 
 	githubutils.SortReleasesBySemver(allReleases)
 	versionsToScan := getVersionsToScan(allReleases)
+	if err != nil {
+		return err
+	}
```
kdorosh, line 446: remove, this does not do anything. please do another self review @Bslabe123

docs/cmd/generate_docs.go
ben-taussig-solo: The assumption I've been working under is that `Fatalf`/`panic` is fine and possibly desirable outside of control plane code -- I don't think we need to be concerned about these lines

projects/gloo/pkg/utils/any.go
```diff
@@ -29,7 +32,8 @@ func MessageToAny(msg proto.Message) (*pany.Any, error) {
 func MustMessageToAny(msg proto.Message) *pany.Any {
```
nfuden, line 32: nit: can we add a method comment that indicates that despite the name this will only panic in dev builds?

projects/gateway/pkg/validation/validator.go
```diff
@@ -201,7 +201,8 @@ func (v *validator) gatewayUpdate(snap *gloov1snap.ApiSnapshot) bool {
 		toHash = append(toHash, snap.VirtualHostOptions.AsInterfaces()...)
 		hash, err := hashutils.HashAllSafe(nil, toHash...)
 		if err != nil {
-			panic("this error should never happen, as this is safe hasher")
+			contextutils.LoggerFrom(context.Background()).DPanic("this error should never happen, as this is safe hasher")
+			return 0
```
ben-taussig-solo, line 205: We should change this hasher to match the changes made to the other hasher if possible -- it's not safe to return 0

projects/gloo/pkg/upstreams/snapshot.go
```diff
@@ -43,7 +46,8 @@ func (s *hybridUpstreamSnapshot) hash() uint64 {
 	allUpstreams.Sort()
 	hash, err := hashutils.HashAllSafe(nil, allUpstreams.AsInterfaces()...)
 	if err != nil {
-		panic("this error should never happen, as it is in a safe hasher")
+		contextutils.LoggerFrom(context.Background()).DPanic("this error should never happen, as it is in a safe hasher")
+		return 0
```
ben-taussig-solo, line 50: We should have this hasher return an error

projects/gloo/pkg/validation/server.go
```diff
@@ -69,7 +69,8 @@ func (s *validator) shouldNotify(snap *v1snap.ApiSnapshot) bool {
 
 		hash, err := hashutils.HashAllSafe(nil, toHash...)
 		if err != nil {
-			panic("this error should never happen, as this is safe hasher")
+			contextutils.LoggerFrom(context.Background()).DPanic("this error should never happen, as this is safe hasher")
+			return 0
```
ben-taussig-solo, line 73: We should have this hasher return an error

docs/cmd/generate_docs.go
ben-taussig-solo: There remain a couple of `Fatalf`s that can potentially occur during setup due to bad parameters:

These two seem fine to me -- we exit if setup fails
https://github.com/solo-io/gloo/blob/removepanic/projects/ingress/cmd/main.go#L10
https://github.com/solo-io/gloo/blob/removepanic/pkg/utils/setuputils/main_setup.go#L93

I don't really know anything about leader election so I'm not sure what to take from this one:
https://github.com/solo-io/gloo/blob/removepanic/projects/gloo/pkg/setup/setup.go#L48
Looks like the only place it's called is here: 
https://github.com/solo-io/gloo/blob/removepanic/pkg/bootstrap/leaderelector/kube/factory.go#L61-L65

@Bslabe123 could you reach out to Sam to determine if this is safe?

pkg/utils/settingsutil/settings.go
```diff
@@ -14,16 +18,17 @@ func WithSettings(ctx context.Context, settings *v1.Settings) context.Context {
...
 	settings := MaybeFromContext(ctx)
 	if settings != nil {
-		return settings
+		return settings, nil
 	}
 	// we should always have settings when this method is called.
-	panic("no settings on context")
+	contextutils.LoggerFrom(ctx).DPanic("no settings on context")
+	return nil, eris.Errorf("no settings on context")
 }
```
kdorosh, line 32: we can remove this function altogether, which would be much cleaner. it is only called in the aws eds plugin; we can just pass in the settings from here instead https://github.com/solo-io/gloo/blob/bd5f0751d2dfe96fc35f0146d5458a828f7e463e/projects/gloo/pkg/plugins/aws/ec2/plugin.go#L40

projects/discovery/pkg/fds/syncer/discovery_syncer.go
```diff
@@ -64,11 +68,12 @@ func selectUpstreamsForDiscovery(fdsMode v1.Settings_DiscoveryOptions_FdsMode, u
...
 	case v1.Settings_DiscoveryOptions_BLACKLIST:
-		return selectUpstreamsBlacklist(upstreams, blacklistNamespaces)
+		return selectUpstreamsBlacklist(upstreams, blacklistNamespaces), nil
 	case v1.Settings_DiscoveryOptions_WHITELIST:
-		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces)
+		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces), nil
 	}
-	panic("invalid fds mode: " + fdsMode.String())
+	contextutils.LoggerFrom(context.Background()).DPanic("invalid fds mode: " + fdsMode.String())
+	return nil, fmt.Errorf("invalid fds mode: %v", fdsMode.String())
```
kdorosh, line 76: rather than error here, i think it makes more sense to log the dpanic error and then default to blacklist for fds (the default) so things still work rather than being completely broken

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
```diff
@@ -95,7 +97,8 @@ func (r *RateLimitConfig) convertSoloKitStatusToRateLimitConfigStatus(status *co
 		outputState = types.RateLimitConfigStatus_REJECTED
 	case core.Status_Warning:
 		// should lever happen
-		panic("cannot set WARNING status on RateLimitConfig resources")
+		contextutils.LoggerFrom(context.Background()).DPanic("cannot set WARNING status on RateLimitConfig resources")
+		return nil
```
kdorosh, line 101: nil is dangerous here; a better safety fallback is to log like you do already and then set to rejected (so we don't return nil)

projects/gloo/cli/pkg/cmd/version/cmd.go
```diff
@@ -159,7 +161,8 @@ func getDistributionName(name string, enterprise bool) string {
 func GetJson(pb proto.Message) []byte {
 	data, err := protoutils.MarshalBytes(pb)
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(context.Background()).DPanic(err)
+		return []byte{}
```
kdorosh, line 165: returning empty bytes makes no sense for the caller; instead the caller should be aware an error happened and handle it properly

projects/gloo/pkg/plugins/aws/ec2/eds.go
```diff
@@ -39,11 +43,14 @@ type edsWatcher struct {
...
-func newEndpointsWatcher(watchCtx context.Context, writeNamespace string, upstreams v1.UpstreamList, secretClient v1.SecretClient, parentRefreshRate time.Duration) *edsWatcher {
+func newEndpointsWatcher(watchCtx context.Context, writeNamespace string, upstreams v1.UpstreamList, secretClient v1.SecretClient, parentRefreshRate time.Duration) (*edsWatcher, error) {
 	var namespaces []string
 
 	// We either watch all namespaces, or create individual watchers for each namespace we watch
-	settings := settingsutil.FromContext(watchCtx)
+	settings, err := settingsutil.FromContext(watchCtx)
+	if err != nil {
+		return nil, err
+	}
```
kdorosh, line 53: again lets remove the `FromContext` function and instead pass the settings in here

projects/gloo/pkg/plugins/kubernetes/client.go
kdorosh: bump; it looks like this is still unfixed

kdorosh (changes requested):
only got halfway through the review; there are several cases that haven't been reasoned about the caller and how to handle a potential logic error. let's go over those again (and if unsure lets hop on a call and talk over) before we mark this as ready for review again. also there was a PR comment that was requested but not addressed -- lets make sure all outstanding comments are addressed before marking as LFR

projects/gloo/cli/pkg/cmd/version/cmd.go
nfuden: This is never called by the control plane does it make sense to still fail here?

projects/gloo/pkg/plugins/aws/ec2/eds.go
nfuden: Does that make semantic sense? What other things might need to be piped through to this in the future?

projects/gateway/pkg/validation/validator.go
```diff
@@ -201,11 +201,14 @@ func (v *validator) gatewayUpdate(snap *gloov1snap.ApiSnapshot) bool {
...
+			contextutils.LoggerFrom(context.Background()).DPanic("this error should never happen, as this is safe hasher")
+			return 0, errors.New("this error should never happen, as this is safe hasher")
 		}
-		return hash
+		return hash, nil
 	}
-	hashChanged := hashFunc(v.latestSnapshot) != hashFunc(snap)
+	oldHash, oldHashErr := hashFunc(v.latestSnapshot)
+	newHash, newHashErr := hashFunc(snap)
+	hashChanged := oldHash != newHash && oldHashErr != nil && newHashErr != nil
```
kdorosh, line 211: change to ORs with the errors being non-nil

projects/gloo/pkg/plugins/aws/ec2/eds.go
```diff
@@ -26,7 +27,11 @@ func (p *plugin) WatchEndpoints(writeNamespace string, unfilteredUpstreams v1.Up
 			ec2Upstreams = append(ec2Upstreams, upstream)
 		}
 	}
-	return newEndpointsWatcher(opts.Ctx, writeNamespace, ec2Upstreams, p.secretClient, opts.RefreshRate).poll()
+	epWatcher, err := newEndpointsWatcher(opts.Ctx, writeNamespace, ec2Upstreams, p.secretClient, opts.RefreshRate)
```
kdorosh, line 30: `p.settings // already here`

projects/gloo/pkg/plugins/aws/ec2/eds.go
```diff
@@ -39,11 +44,15 @@ type edsWatcher struct {
...
 }
 
-func newEndpointsWatcher(watchCtx context.Context, writeNamespace string, upstreams v1.UpstreamList, secretClient v1.SecretClient, parentRefreshRate time.Duration) *edsWatcher {
+func newEndpointsWatcher(watchCtx context.Context, writeNamespace string, upstreams v1.UpstreamList, secretClient v1.SecretClient, parentRefreshRate time.Duration) (*edsWatcher, error) {
 	var namespaces []string
 
 	// We either watch all namespaces, or create individual watchers for each namespace we watch
-	settings := settingsutil.FromContext(watchCtx)
+
+	settings := settingsutil.MaybeFromContext(watchCtx)
```
kdorosh, line 52: we can remove this

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -89,7 +89,10 @@ func (p *plugin) WatchEndpoints(writeNamespace string, upstreamsToTrack v1.Upstr
 func newEndpointWatcherForUpstreams(kubeFactoryFactory func(ns []string) KubePluginSharedFactory, kubeCoreCache corecache.KubeCoreCache, writeNamespace string, upstreamsToTrack v1.UpstreamList, opts clients.WatchOpts) (*edsWatcher, error) {
 	var namespaces []string
 
-	settings := settingsutil.FromContext(opts.Ctx)
+	settings := settingsutil.MaybeFromContext(opts.Ctx)
```
kdorosh, line 92: we also have p.settings we can remove this

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -108,6 +111,9 @@ func newEndpointWatcherForUpstreams(kubeFactoryFactory func(ns []string) KubePlu
 	}
 
 	kubeFactory := kubeFactoryFactory(namespaces)
+	if kubeFactory == nil {
+		return nil, errors.New("Error in kubeFactory")
+	}
```
kdorosh, line 116: we should update the factory to return error and check it earlier so its harder to misuse

projects/gloo/pkg/plugins/pluginutils/destination_upstreams.go
```diff
@@ -31,7 +34,8 @@ func DestinationUpstreams(snap *v1snap.ApiSnapshot, in *v1.RouteAction) ([]core.
 		}
 		return destinationsToRefs(upstreamGroup.GetDestinations())
 	}
-	panic("invalid route")
+	contextutils.LoggerFrom(context.Background()).DPanic("invalid route")
```
kdorosh, line 37: this is a bit noisy, we should remove

projects/gloo/pkg/plugins/registry/registry_test.go
```diff
@@ -291,24 +293,28 @@ func TestPluginsHttpFilterUsefulness(t *testing.T) {
 			for _, routePlugin := range pluginRegistry.GetRoutePlugins() {
 				err := routePlugin.ProcessRoute(routeParams, emptyRoute, &envoy_config_route_v3.Route{})
 				if err != nil {
-					t.Fatalf("plugin route filter failed %v", err)
+					Fail(fmt.Sprintf("plugin route filter failed %v", err))
+					return
```
kdorosh, line 297: redundant

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -82,8 +82,8 @@ type httpFilterChainTranslator struct {
 
 func (h *httpFilterChainTranslator) ComputeFilterChains(params plugins.Params) []*envoy_config_listener_v3.FilterChain {
 	// 1. Generate all the network filters (including the HttpConnectionManager)
-	networkFilters := h.networkFilterTranslator.ComputeNetworkFilters(params)
-	if len(networkFilters) == 0 {
+	networkFilters, err := h.networkFilterTranslator.ComputeNetworkFilters(params)
+	if err != nil || len(networkFilters) == 0 {
```
kdorosh, line 86: i think should at least log an error here

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -142,10 +147,11 @@ func (h *hcmNetworkFilterTranslator) ComputeNetworkFilter(params plugins.Params)
 	// 4. Generate the typedConfig for the HCM
 	hcmFilter, err := NewFilterWithTypedConfig(wellknown.HTTPConnectionManager, httpConnectionManager)
 	if err != nil {
-		panic(errors.Wrapf(err, "failed to convert proto message to struct"))
+		contextutils.LoggerFrom(context.Background()).DPanic(errors.Wrapf(err, "failed to convert proto message to struct"))
```
kdorosh, line 150: remove this log line, let handler decide what to do with the error (make sure it logs)

projects/gloo/pkg/translator/translator.go
```diff
@@ -270,20 +270,39 @@ func (t *translatorInstance) generateXDSSnapshot(
...
+		listenersNew = envoycache.NewResources(fmt.Sprintf("%v", listenersVersion), listenersProto)
+	}
 	return xds.NewSnapshotFromResources(
-		envoycache.NewResources(fmt.Sprintf("%v-%v", clustersVersion, endpointsVersion), endpointsProto),
-		envoycache.NewResources(fmt.Sprintf("%v", clustersVersion), clustersProto),
+		endpointsNew,
+		clustersNew,
 		MakeRdsResources(routeConfigs),
-		envoycache.NewResources(fmt.Sprintf("%v", listenersVersion), listenersProto))
+		listenersNew)
```
kdorosh, line 302: don't put empty endpoints/clusters/listiners etc. this is the cache key in xds cache, usually we just hash and use the hash value as the key (e.g. 129812472139182)

if we can't hash, we should still have it stored and served (we have valid config) but we need a cache key, let's say (`endpoints-hasherr`). this can overwrite any previous hash errors for the same key

projects/gloo/pkg/translator/translator.go
```diff
@@ -329,7 +351,11 @@ func MakeRdsResources(routeConfigs []*envoy_config_route_v3.RouteConfiguration)
 
 	}
 
-	routesVersion := MustEnvoyCacheResourcesListToFnvHash(routesProto)
+	routesVersion, err := MustEnvoyCacheResourcesListToFnvHash(routesProto)
+	if err != nil {
+		contextutils.LoggerFrom(context.Background()).DPanic(fmt.Sprintf("error trying to hash routesProto: %v", err))
+		return envoycache.Resources{}
```
kdorosh, line 357: same note here, return the routes but with the hash error in the key

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -352,6 +353,18 @@ var _ = Describe("Translator", func() {
...
 
+	translateWithBuggyHasher := func() *validation.ProxyReport {
+		buggyHasher := func(resources []envoycache.Resource) (uint64, error) {
+			return 0, errors.New("This is a buggy hasher error")
+		}
+		translator = NewTranslatorWithHasher(glooutils.NewSslConfigTranslator(), settings, registry.NewPluginRegistry(registeredPlugins), buggyHasher)
+		snapshot, _, report := translator.Translate(params, proxy)
+		Expect(snapshot.GetResources(types.EndpointTypeV3).Items).To(BeNil())
+		Expect(snapshot.GetResources(types.ClusterTypeV3).Items).To(BeNil())
+		Expect(snapshot.GetResources(types.ListenerTypeV3).Items).To(BeNil())
```
kdorosh, line 364: needs to be updated

projects/gloo/pkg/upstreams/base_client.go
```diff
@@ -23,7 +27,8 @@ func (c *readOnlyUpstreamBaseClient) Kind() string {
 }
 
 func (c *readOnlyUpstreamBaseClient) NewResource() resources.Resource {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return nil
```
kdorosh, line 31: let's return an empty / placeholder upstream so the caller has something not nil to work with

kdorosh: lets leave the dpanic

projects/gloo/pkg/upstreams/consul/upstream_client.go
```diff
@@ -22,23 +26,28 @@ type consulUpstreamClient struct {
 }
 
 func (*consulUpstreamClient) BaseClient() skclients.ResourceClient {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return nil
```
kdorosh, line 30: i have the same concern about nil here, can we do better?

kdorosh: let's create a mock client to return here that just returns errors

projects/gloo/pkg/upstreams/kubernetes/upstream_client.go
```diff
@@ -22,23 +25,28 @@ type kubernetesUpstreamClient struct {
 }
 
 func (c *kubernetesUpstreamClient) BaseClient() skclients.ResourceClient {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return nil
```
kdorosh, line 29: same here about mock base client

projects/gloo/pkg/utils/any.go
```diff
@@ -26,10 +29,12 @@ func MessageToAny(msg proto.Message) (*pany.Any, error) {
 	}, nil
 }
 
+// This will only DPanic in dev builds
 func MustMessageToAny(msg proto.Message) *pany.Any {
```
kdorosh, line 33: I think it's a bug to call this in production code. frankly we should remove all calls for this and replace with `MessageToAny` and force the caller to handle errors, rather than return an empty any

projects/gloo/pkg/validation/server.go
```diff
@@ -69,12 +70,14 @@ func (s *validator) shouldNotify(snap *v1snap.ApiSnapshot) bool {
...
+			return 0, errors.New("this error should never happen, as this is safe hasher")
 		}
-		return hash
+		return hash, nil
 	}
-
-	hashChanged := hashFunc(s.latestSnapshot) != hashFunc(snap)
+	oldHash, oldHashErr := hashFunc(s.latestSnapshot)
+	newHash, newHashErr := hashFunc(snap)
+	hashChanged := oldHash != newHash && oldHashErr == nil && newHashErr == nil
```
kdorosh, line 80: same note about switching to OR

projects/gloo/pkg/translator/clusters.go
```diff
@@ -110,9 +111,13 @@ func (t *translatorInstance) initializeCluster(
 		if err != nil {
 			reports.AddError(upstream, err)
 		} else {
+			typedConfig, err := utils.MessageToAny(cfg)
+			if err != nil {
+				typedConfig, _ = utils.MessageToAny(&envoyauth.UpstreamTlsContext{})
+			}
```
kdorosh, line 117: an empty transport socket will definitely break envoy. we should `reports.AddError()` and skip tls for now as a best faith effort. lets talk over tomorrow just like we did this morning

elcasteel (dismissed):
lgtm

projects/gloo/pkg/translator/clusters.go
kdorosh: we spoke about this, better to return error (and have placeholder cluster) than use a potentially insecure cluster when tls was intended

nfuden:
Let's write out the consumers are for each nil return and state why it's not going to generate a later panic

pkg/utils/settingsutil/settings.go
```diff
@@ -14,18 +14,6 @@ func WithSettings(ctx context.Context, settings *v1.Settings) context.Context {
 	return context.WithValue(ctx, settingsKey, settings)
 }
 
-func FromContext(ctx context.Context) *v1.Settings {
```
nfuden, line 17: I dont think we should remove this. Its technically an exported function and if used as a library this is a breaking change

elcasteel (dismissed):

kdorosh (changes requested):
i still had some outstanding comments (and nathan)

i think for a first iteration we should just merge the low risk ones (e.g. not implemented when we could just return an error) and leave high risk changes alone for now (e.g. hashing / xds snapshot / protobuf marshall errors)

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
```diff
@@ -95,7 +97,8 @@ func (r *RateLimitConfig) convertSoloKitStatusToRateLimitConfigStatus(status *co
 		outputState = types.RateLimitConfigStatus_REJECTED
 	case core.Status_Warning:
 		// should lever happen
-		panic("cannot set WARNING status on RateLimitConfig resources")
+		contextutils.LoggerFrom(context.Background()).DPanic("cannot set WARNING status on RateLimitConfig resources")
+		outputState = types.RateLimitConfigStatus_REJECTED
```
nfuden, line 101: Can we state what happens on this rejection and why it makes sense for a warning?

projects/gloo/pkg/plugins/als/plugin.go
```diff
@@ -117,10 +117,14 @@ func copyFileSettings(cfg *envoyalfile.FileAccessLog, alsSettings *als.AccessLog
...
 	query := &envoy_req_without_query.ReqWithoutQuery{}
+	typedConfig, err := utils.MessageToAny(query)
+	if err != nil {
+		return err
+	}
 	formatterExtensions := []*envoycore.TypedExtensionConfig{
 		{
 			Name:        "envoy.formatter.req_without_query",
-			TypedConfig: utils.MustMessageToAny(query),
+			TypedConfig: typedConfig,
```
nfuden, line 127: What happens if this is nil?

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -108,9 +108,13 @@ func (p *Plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
 		Sni: lambdaHostname,
 	}
+	typedConfig, err := utils.MessageToAny(tlsContext)
+	if err != nil {
+		return err
+	}
 	out.TransportSocket = &envoy_config_core_v3.TransportSocket{
 		Name:       wellknown.TransportSocketTls,
-		ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: utils.MustMessageToAny(tlsContext)},
+		ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: typedConfig},
```
nfuden, line 117: What happens if this typedConfig is nil?

projects/gloo/pkg/validation/server.go
```diff
@@ -69,12 +70,14 @@ func (s *validator) shouldNotify(snap *v1snap.ApiSnapshot) bool {
...
+			return 0, errors.New("this error should never happen, as this is safe hasher")
 		}
-		return hash
+		return hash, nil
 	}
-
-	hashChanged := hashFunc(s.latestSnapshot) != hashFunc(snap)
+	oldHash, oldHashErr := hashFunc(s.latestSnapshot)
+	newHash, newHashErr := hashFunc(snap)
+	hashChanged := oldHash != newHash || oldHashErr == nil || newHashErr == nil
```
nfuden, line 80: Question if new hash is nil should we attempt to keep the old hash instead?

Bslabe123: The hash is generated from the current snapshot regardless right?

projects/discovery/pkg/fds/syncer/discovery_syncer.go
```diff
@@ -68,7 +67,8 @@ func selectUpstreamsForDiscovery(fdsMode v1.Settings_DiscoveryOptions_FdsMode, u
 	case v1.Settings_DiscoveryOptions_WHITELIST:
 		return selectUpstreamsWhitelist(upstreams, whitelistNamespaces, blacklistNamespaces)
 	}
-	panic("invalid fds mode: " + fdsMode.String())
+	contextutils.LoggerFrom(context.Background()).DPanic("invalid fds mode: " + fdsMode.String())
```
kdorosh, line 70: let's update this log line to mention that we are falling back to blacklist fds mode

projects/gateway/pkg/validation/validator.go
```diff
@@ -201,11 +201,14 @@ func (v *validator) gatewayUpdate(snap *gloov1snap.ApiSnapshot) bool {
...
+			contextutils.LoggerFrom(context.Background()).DPanic("this error should never happen, as this is safe hasher")
+			return 0, errors.New("this error should never happen, as this is safe hasher")
 		}
-		return hash
+		return hash, nil
 	}
-	hashChanged := hashFunc(v.latestSnapshot) != hashFunc(snap)
+	oldHash, oldHashErr := hashFunc(v.latestSnapshot)
+	newHash, newHashErr := hashFunc(snap)
+	hashChanged := oldHash != newHash || oldHashErr != nil || newHashErr != nil
```
kdorosh, line 211: i think this deserves a comment for the next reader: basically saying that if we cannot hash then we choose to treat them as different hashes since this is just a performance optimization. in worst case we'd prefer correctness

projects/gloo/pkg/translator/clusters.go
```diff
@@ -110,9 +110,14 @@ func (t *translatorInstance) initializeCluster(
 		if err != nil {
 			reports.AddError(upstream, err)
 		} else {
-			out.TransportSocket = &envoy_config_core_v3.TransportSocket{
-				Name:       wellknown.TransportSocketTls,
-				ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: utils.MustMessageToAny(cfg)},
+			typedConfig, err := utils.MessageToAny(cfg)
+			if err != nil {
+				reports.AddError(upstream, err)
```
kdorosh, line 115: i think it would be prudent for security to also ensure that this upstream becomes a direct response action. it could be a security risk to allow a secured upstream to silently become unencrypted

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -82,8 +85,14 @@ type httpFilterChainTranslator struct {
 
 func (h *httpFilterChainTranslator) ComputeFilterChains(params plugins.Params) []*envoy_config_listener_v3.FilterChain {
 	// 1. Generate all the network filters (including the HttpConnectionManager)
-	networkFilters := h.networkFilterTranslator.ComputeNetworkFilters(params)
+	networkFilters, err := h.networkFilterTranslator.ComputeNetworkFilters(params)
+	if err != nil {
+		contextutils.LoggerFrom(context.Background()).DPanic(err)
+		return nil
```
kdorosh, line 91: the caller here will append a nil filter chain rather than skipping it; is this correct?

projects/gloo/pkg/translator/translator.go
```diff
@@ -269,20 +269,41 @@ func (t *translatorInstance) generateXDSSnapshot(
 	}
 	// construct version
 	// TODO: investigate whether we need a more sophisticated versioning algorithm
-	endpointsVersion := t.hasher(endpointsProto)
-	clustersVersion := t.hasher(clustersProto)
-	listenersVersion := t.hasher(listenersProto)
+	endpointsVersion, endpointsErr := t.hasher(endpointsProto)
+	if endpointsErr != nil {
+		contextutils.LoggerFrom(context.Background()).DPanic(fmt.Sprintf("error trying to hash endpointsProto: %v", endpointsErr))
```
kdorosh, line 274: nit; can we easily pass context in here?

projects/gloo/pkg/translator/translator.go
```diff
@@ -269,20 +269,41 @@ func (t *translatorInstance) generateXDSSnapshot(
...
+	}
+	listenersVersion, listenersErr := t.hasher(listenersProto)
+	if listenersErr != nil {
+		contextutils.LoggerFrom(context.Background()).DPanic(fmt.Sprintf("error trying to hash listenersProto: %v", listenersErr))
+	}
 
 	// if clusters are updated, provider a new version of the endpoints,
 	// so the clusters are warm
+	endpointsNew := envoycache.NewResources("endpoints-hashErr", endpointsProto)
+	if endpointsErr == nil {
```
kdorosh, line 288: this should also check that clustersErr is nil

projects/gloo/pkg/translator/translator.go
```diff
@@ -297,23 +318,26 @@ func MustEnvoyCacheResourcesListToFnvHash(resources []envoycache.Resource) uint6
...
+			return 0, errors.Wrap(err, "constructing hash for envoy snapshot components")
 		}
 	}
-	return hasher.Sum64()
+	return hasher.Sum64(), nil
 }
 
 // deprecated, slower than MustEnvoyCacheResourcesListToFnvHash
-func MustEnvoyCacheResourcesListToHash(resources []envoycache.Resource) uint64 {
+func MustEnvoyCacheResourcesListToHash(resources []envoycache.Resource) (uint64, error) {
```
kdorosh, line 334: must implies no error. i think we should deprecate the existing function, add a new one with a different name, and use the different one everywhere

projects/gloo/pkg/upstreams/hybrid_client.go
```diff
@@ -153,8 +158,8 @@ func (c *hybridUpstreamClient) Watch(namespace string, opts clients.WatchOpts) (
 
 		// return success for the sync (ie if there still needs changes its a false)
 		syncFunc := func() bool {
-			currentHash := current.hash()
-			if currentHash == previousHash {
+			currentHash, err := current.hash()
+			if currentHash == previousHash && err == nil {
```
kdorosh, line 162: we should also add a comment here to explain why this check is correct for the next dev who sees this

projects/gloo/pkg/utils/any.go
```diff
@@ -44,18 +45,18 @@ func MustAnyToMessage(a *pany.Any) proto.Message {
 	var x ptypes.DynamicAny
 	err := ptypes.UnmarshalAny(a, &x)
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(context.Background()).DPanic(err)
+		return proto.MessageV1("")
```
kdorosh, line 49: does this even work? what type / proto does this return? perhaps we should just return `nil, someErr`

projects/gloo/pkg/utils/any_test.go
```diff
@@ -40,4 +41,9 @@ var _ = Describe("Any", func() {
 		Expect(msg2.(*structpb.Struct).Fields["test"].GetStringValue()).To(Equal(msg.Fields["test"].GetStringValue()))
 	})
 
+	It("fails with nil message", func() {
+		var msg proto.Message
+		_, err := MessageToAny(msg)
+		Expect(err).To(HaveOccurred())
```
kdorosh, line 47: let's make a stronger assertion about the error (what kind of error, what does it say?). e.g. `Expect(err).To(MatchErr(ContainSubstring("...")))

kdorosh (changes requested):
still several bugs; let's get those addressed before we attempt a merge

projects/gloo/pkg/utils/any.go
Bslabe123: Actually since we only use this function in tests, might be worth it to just mark it as deprecated and revert

projects/gloo/pkg/translator/filter_chain.go
Bslabe123: Yep! This mirrors how we would handle the `if len(networkFilters) == 0` case

projects/gloo/pkg/translator/clusters.go
Bslabe123: To adress later per slack conversaion

projects/gloo/pkg/validation/server.go
Bslabe123: My mistake! This should be fixed now

projects/gloo/pkg/translator/clusters.go
Bslabe123: Updated this comment, may be a bit more involved than we thought, would it be worth to have a stopgap solution here for the meantime and open a separate issue for this?

changelog/v1.13.0-beta22/remove-panic.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7174
+    resolvesIssue: false
+    description: >
+      Removed all calls to potentially dangerous `panic()` replaced with `contextutils.LoggerFrom(ctx).DPanic()`
```
kdorosh, line 6: nit; we did not remove all calls

projects/gloo/pkg/plugins/kubernetes/client.go
```diff
@@ -35,15 +37,16 @@ type KubePluginListers struct {
...
-func getInformerFactory(ctx context.Context, client kubernetes.Interface, watchNamespaces []string) *KubePluginListers {
+func getInformerFactory(ctx context.Context, client kubernetes.Interface, watchNamespaces []string) (*KubePluginListers, error) {
 	if len(watchNamespaces) == 0 {
 		watchNamespaces = []string{metav1.NamespaceAll}
 	}
 	kubePluginSharedFactory := startInformerFactory(ctx, client, watchNamespaces)
 	if kubePluginSharedFactory.initError != nil {
-		panic(kubePluginSharedFactory.initError)
+		contextutils.LoggerFrom(context.Background()).DPanic(kubePluginSharedFactory.initError)
+		return nil, kubePluginSharedFactory.initError
```
kdorosh, line 47: i actually think this is one of those places where panic is better than silently allowing us to go on. this is an unrecoverable error (no shared informer factory means all of kube EDS won't work, which is probably the most valuable / important role for gloo) and i'd want to make sure users know immediately about e.g. any rbac errors preventing this from working rather than hiding this in the logs

my only other thought here is that we also use shared informer in our emitter watches, perhaps we can only initialize once and reuse. that seems out of scope for this issue; i think this panic should be left here with a detailed comment why

projects/gloo/pkg/translator/clusters.go
```diff
@@ -110,9 +110,17 @@ func (t *translatorInstance) initializeCluster(
...
 		} else {
-			out.TransportSocket = &envoy_config_core_v3.TransportSocket{
-				Name:       wellknown.TransportSocketTls,
-				ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: utils.MustMessageToAny(cfg)},
+			typedConfig, err := utils.MessageToAny(cfg)
+			if err != nil {
+				// TODO: Need to change the upstream to use a direct response action instead of leaving the upstream untouched
+				// Difficult because direct response is not on the upsrtream but on the virtual host
+				// The fallback listener would take much more piping as well
+				reports.AddError(upstream, err)
```
kdorosh, line 118: this is another one where i think until we have direct response action implemented, we have to leave the panic. with the panic, we'll never allow unsecured traffic to backend; but now if we just add an error we may allow this to happen (security risk)

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -82,8 +85,13 @@ type httpFilterChainTranslator struct {
...
+	}
 	if len(networkFilters) == 0 {
+		contextutils.LoggerFrom(context.Background()).DPanic("networkFilters cannot be empty")
```
kdorosh, line 94: is this really a dpanic? is it possibly expected behavior that this is just empty but not errored?

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -182,22 +194,25 @@ func newSslFilterChain(
...
-) *envoy_config_listener_v3.FilterChain {
+) (*envoy_config_listener_v3.FilterChain, error) {
 
 	// copy listenerFilter so we can modify filter chain later without changing the filters on all of them!
 	clonedListenerFilters := cloneListenerFilters(listenerFilters)
-
+	typedConfig, err := utils.MessageToAny(downstreamTlsContext)
+	if err != nil {
+		return nil, err
+	}
```
kdorosh, line 204: same note here about direct response action; i think until we have this we need to leave this panic

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -141,11 +149,15 @@ func (h *httpFilterChainTranslator) createFilterChainsFromSslConfiguration(
...
-		filterChain := newSslFilterChain(
+		filterChain, err := newSslFilterChain(
 			downstreamTlsContext,
 			sslConfig.GetSniDomains(),
 			networkFilters,
 			sslConfig.GetTransportSocketConnectTimeout())
+		if err != nil {
+			validation.AppendListenerError(h.parentReport, validationapi.ListenerReport_Error_SSLConfigError, err.Error())
+			continue
+		}
```
kdorosh, line 160: same note here about direct response action; i think until we have this we need to leave this panic

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -142,10 +146,10 @@ func (h *hcmNetworkFilterTranslator) ComputeNetworkFilter(params plugins.Params)
 	// 4. Generate the typedConfig for the HCM
 	hcmFilter, err := NewFilterWithTypedConfig(wellknown.HTTPConnectionManager, httpConnectionManager)
 	if err != nil {
-		panic(errors.Wrapf(err, "failed to convert proto message to struct"))
+		return nil, errors.Wrapf(err, "failed to convert proto message to struct")
```
kdorosh, line 149: dpanic here?

projects/gloo/pkg/translator/translator.go
```diff
@@ -269,20 +270,50 @@ func (t *translatorInstance) generateXDSSnapshot(
...
+		endpointsNew = envoycache.NewResources(fmt.Sprintf("%v-%v", clustersVersion, endpointsVersion), endpointsProto)
+	}
+	clustersNew := envoycache.NewResources("clusters-hashErr", endpointsProto)
+	if clustersErr == nil {
+		clustersNew = envoycache.NewResources(fmt.Sprintf("%v", clustersVersion), clustersProto)
+	}
+	listenersNew := envoycache.NewResources("listeners-hashErr", listenersProto)
+	if listenersErr == nil {
+		listenersNew = envoycache.NewResources(fmt.Sprintf("%v", listenersVersion), listenersProto)
+	}
```
kdorosh, line 299: super nit; i think this is easier to read if the logic is inverted -- default to normal behavior and update cache key if err is not nil

projects/gloo/pkg/translator/translator.go
```diff
@@ -297,23 +328,25 @@ func MustEnvoyCacheResourcesListToFnvHash(resources []envoycache.Resource) uint6
...
+			return 0, errors.Wrap(err, "constructing hash for envoy snapshot components")
 		}
 	}
-	return hasher.Sum64()
+	return hasher.Sum64(), nil
 }
 
 // deprecated, slower than MustEnvoyCacheResourcesListToFnvHash
-func MustEnvoyCacheResourcesListToHash(resources []envoycache.Resource) uint64 {
+func MustEnvoyCacheResourcesListToHash(resources []envoycache.Resource) (uint64, error) {
```
kdorosh, line 344: bump on https://github.com/solo-io/gloo/pull/7182/files#r987927412 still looks unaddressed (function should be deprecated but keep original signature)

projects/gloo/pkg/upstreams/base_client.go
```diff
@@ -23,7 +27,8 @@ func (c *readOnlyUpstreamBaseClient) Kind() string {
 }
 
 func (c *readOnlyUpstreamBaseClient) NewResource() resources.Resource {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return resources.Resource(nil)
```
kdorosh, line 31: let's update nil to instead be some kind of placeholder upstream (e.g. name/namespace coming from install namespace, name = "dev-error-placeholder")

projects/gloo/pkg/upstreams/consul/upstream_client.go
```diff
@@ -22,23 +26,28 @@ type consulUpstreamClient struct {
 }
 
 func (*consulUpstreamClient) BaseClient() skclients.ResourceClient {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return skclients.ResourceClient(nil)
```
kdorosh, line 30: let's update nil to instead be some kind of placeholder upstream (e.g. name/namespace coming from install namespace, name = "dev-error-placeholder")

projects/gloo/pkg/upstreams/kubernetes/upstream_client.go
```diff
@@ -22,23 +25,28 @@ type kubernetesUpstreamClient struct {
 }
 
 func (c *kubernetesUpstreamClient) BaseClient() skclients.ResourceClient {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return skclients.ResourceClient(nil)
```
kdorosh, line 29: same note about a placeholder instead of nil (called will expect this to be non-nil and likely panic upon access)

kdorosh (changes requested):
very close! just one or two panics i think we should revert since they are truly unrecoverable (or we need direct response action implemented) as well as a couple nils in base client that should have more complete placeholders returned

projects/gloo/pkg/plugins/kubernetes/client.go
Bslabe123: Adapted this and used it as a code comment

go.sum
```diff
@@ -1,3 +1,4 @@
+bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898 h1:SC+c6A1qTFstO9qmB86mPV2IpYme/2ZoEQ0hrP+wo+Q=
```
kdorosh, line 1: why so many changes to this file if no changes to go.mod?

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -107,7 +106,10 @@ func newEndpointWatcherForUpstreams(kubeFactoryFactory func(ns []string) KubePlu
 		}
 	}
 
-	kubeFactory := kubeFactoryFactory(namespaces)
+	kubeFactory, err := kubeFactoryFactory(namespaces)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error in kubeFactory")
```
kdorosh, line 111: we can undo this

projects/gloo/pkg/upstreams/base_client.go
```diff
@@ -23,7 +29,13 @@ func (c *readOnlyUpstreamBaseClient) Kind() string {
 }
 
 func (c *readOnlyUpstreamBaseClient) NewResource() resources.Resource {
-	panic(notImplementedErrMsg)
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return &v1.Upstream{
+		Metadata: &core.Metadata{
+			Name:      "dev-error-placeholder",
+			Namespace: "install",
```
kdorosh, line 36: let's make this the same as the gloo installation namespace (which we know actually exists). `utils.GetPodNamespace(),` in "github.com/solo-io/gloo/pkg/utils"

projects/gloo/pkg/upstreams/NoOpUpstreamClient/noop_client.go
```diff
@@ -0,0 +1,61 @@
...
+
+const notImplementedErrMsg = "this operation is not supported by this client"
+
+type NoOpUpstreamClient struct {
+	clientMap map[string]v1.UpstreamClient
+}
+
+func (c *NoOpUpstreamClient) NewResource() resources.Resource {
+	contextutils.LoggerFrom(context.Background()).DPanic(notImplementedErrMsg)
+	return nil
```
kdorosh, line 21: let's make this return a placeholder upstream as well

projects/gloo/pkg/upstreams/NoOpUpstreamClient/noop_client.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (c *NoOpUpstreamClient) Kind() string {
+	return ""
```
kdorosh, line 25: this should be nonempty, i.e. `reflect.TypeOf(resource).String()` where resource is an upstream

kdorosh (commented):
just a couple nits; 95% lgtm!

projects/gloo/pkg/upstreams/NoOpUpstreamClient/noop_client.go
```diff
@@ -0,0 +1,69 @@
...
+	return &v1.Upstream{
+		Metadata: &core.Metadata{
+			Name:      "dev-error-placeholder",
+			Namespace: utils.GetPodNamespace(),
+		},
+	}
+}
+
+func (c *NoOpUpstreamClient) Kind() string {
+	return reflect.TypeOf(c).String()
```
kdorosh, line 33: this shouldn't be the type of c; this should be the type of `v1.Upstream`

kdorosh: i would read https://book.kubebuilder.io/cronjob-tutorial/gvks.html to understand what `Kind()` is meant to do for the client here

kdorosh (approved):
looks great! :tada:

nfuden (approved):

PR #7198, opened 2022-09-20
Optimize rate limit hash calculation

yuval-k:


yuval-k:
note: the test fail is due to data race that is fixed in a different PR

kdorosh (approved):

sam-heilbron (approved):
LGTM

Issue #7197, opened 2022-09-20
default liveness probe (readiness only enabled by default) is trigger happy


kdorosh:
### Gloo Edge Version

1.12.x (latest stable)

### Kubernetes Version

1.23.x

### Describe the bug

the readiness and liveness probes are the same (if using the following helm values):

```yaml
gloo:
  deployment:
    livenessProbeEnabled: true
```

which depend on the xds server to determine pod health: https://github.com/solo-io/gloo/blob/v1.13.0-beta15/install/helm/gloo/templates/1-gloo-deployment.yaml#L92-L105

this is WRONG for liveness probes, as in large environments gloo may take a while to start the xds server before relying on the watches to maintain happy state quickly. we don't want the liveness probe to fail falsely and restart the pod while it's still getting ready

### Steps to reproduce the bug

install with the helm values above

### Expected Behavior

we should have a simple liveness probe that checks an http handler that just returns "OK".

this could even fire if gloo gets stuck / slow (e.g. gc hell) as it's not simple to signal the pod is in deadlock scenario.

### Additional Context

generally it is not recommended to run gloo with a liveness probe but some organizations require it, so we will have this simple one as our best offering (still disabled by default)

kdorosh:
reopening until this is also added to enterprise

kdorosh:
resolved in https://github.com/solo-io/solo-projects/pull/4141 and already released in 1.12 enterprise

PR #7199, opened 2022-09-20
Default Gloo Liveness Probe Shouldn't Wait for XDS

kdorosh:
# Description

Updates the Gloo liveness probe to use a new http server that always returns http 200 "ok".

# Context

Gloo liveness probes are disabled by default, and are not generally recommended (as there is not a good way for us to signal we are in deadlock; rather prefer on healthchecks on the node and readiness).

Regardless, some organizations require liveness and readiness probes for all pods; these users opt into our liveness probe which is occasionally trigger-happy (particularly at scale). The old probe would wait for xds before signalling health, which in large environments could lag while EDS is warming. This could, in rare cases, cause the gloo pod to needless bounce while it was warming.

## Testing

automated: added this flag to our integration tests to ensure liveness probe is being hit and passing

manual:
- `kind create cluster --image kindest/node:v1.20.15@sha256:6f2d011dffe182bad80b85f6c00e8ca9d86b5b8922cdf433d53575c4c5212248`
- `VERSION=0.0.0-kdorosh make clean build-test-chart push-kind-images -B`
- `glooctl install gateway -f _test/gloo-0.0.0-kdorosh.tgz --values vals.yaml` where vals.yaml is:

```yaml
gloo:
  deployment:
    livenessProbeEnabled: true
```

Then confirmed liveness probes were passing (and failing once the path/port was updated)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7197

### DIFF
--- a/install/helm/gloo/templates/1-gloo-deployment.yaml
+++ b/install/helm/gloo/templates/1-gloo-deployment.yaml
@@ -97,11 +97,12 @@ spec:
           failureThreshold: 3
 {{- if .Values.gloo.deployment.livenessProbeEnabled }}
         livenessProbe:
-          tcpSocket:
-            port: {{ .Values.gloo.deployment.xdsPort }}
+          httpGet:
+            path: /healthz
+            port: 8765
           initialDelaySeconds: 3
           periodSeconds: 10
-          failureThreshold: 10
+          failureThreshold: 3
 {{- end}}
         volumeMounts:
         - mountPath: /etc/envoy
@@ -242,8 +243,9 @@ spec:
           failureThreshold: 3
 {{- if .Values.gloo.deployment.livenessProbeEnabled }}
         livenessProbe:
-          tcpSocket:
-            port: {{ .Values.gloo.deployment.xdsPort }}
+          httpGet:
+            path: /healthz
+            port: 8765
           initialDelaySeconds: 3
           periodSeconds: 10
           failureThreshold: 3
--- a/pkg/utils/probes/healthz.go
+++ b/pkg/utils/probes/healthz.go
@@ -1,0 +1,48 @@
+package probes
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+func StartLivenessProbeServer(ctx context.Context) {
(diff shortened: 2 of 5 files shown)


install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -97,11 +97,12 @@ spec:
           failureThreshold: 3
 {{- if .Values.gloo.deployment.livenessProbeEnabled }}
         livenessProbe:
-          tcpSocket:
-            port: {{ .Values.gloo.deployment.xdsPort }}
+          httpGet:
+            path: /healthz
+            port: 8765
```
jenshu, line 102: any need to make this port configurable?

pkg/utils/probes/healthz.go
```diff
@@ -0,0 +1,48 @@
...
+		mux := new(http.ServeMux)
+		mux.HandleFunc("/healthz", okHandler)
+		server = &http.Server{
+			Addr:    fmt.Sprintf(":%d", 8765),
+			Handler: mux,
+		}
+		contextutils.LoggerFrom(ctx).Infof("healthz server starting at %s", server.Addr)
+		err := server.ListenAndServe()
+		if err == http.ErrServerClosed {
+			contextutils.LoggerFrom(ctx).Info("Stats server closed")
```
jenshu, line 26: should this say "healthcheck server" instead of "stats server"?
same with the other occurrences in this file

jenshu (commented):
thanks for the useful PR description!

install/helm/gloo/templates/1-gloo-deployment.yaml
kdorosh: i can't think of a good reason; figured better to have less noise / toggles in our helm chart unless we actually need it

pkg/utils/probes/healthz.go
kdorosh: good call, updated in https://github.com/solo-io/gloo/pull/7199/commits/b83dd9127fb8d4d98270ed1808a960494f627dda

jenshu (approved):

Bslabe123 (approved):

Issue #7202, opened 2022-09-21
Race condition caused by same translator being used in `proxyReconciler` and `validating_admission_webhook`


saiskee:
### Gloo Edge Version

1.12.x (latest stable)

### Kubernetes Version

1.24.x

### Describe the bug

Because plugins are now long living (after gloo v1.12.15), a single translator instance should not be shared amongst different goroutines in order to prevent a race condition where multiple loops are trying to write to the plugin's persisted state (e.g. caches). 

The validationTranslator is used [here](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/setup/setup_syncer.go#L750) and the same validator is used in the [validation server](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/setup/setup_syncer.go#L750) as well as the [proxyReconciler](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/setup/setup_syncer.go#L762) and the [gwValidationSyncer](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/setup/setup_syncer.go#L778). 

### Steps to reproduce the bug

Build gloo with `-race` and make multiple updates to a virtual service with jwt config simultaneously (using a bash script). This will cause a DATA RACE log in the gloo pod.

### Expected Behavior

--

PR #7207, opened 2022-09-21
Resolve various data races in translation and validation

kdorosh:
# Description

Ran Gloo in a large environment under race detector, which identified some races which are fixed in this PR:
- Updates validation server to use a shared translator that is synchronized which prevents data races across translation
- Updates locking in validation server to prevent concurrent access to last snapshot
- Updates `RateLimitConfig` client construction to happen during `init()` so it cannot race with global scheme access by k8s shared informers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7202
resolves https://github.com/solo-io/skv2/issues/375
resolves https://github.com/solo-io/gloo/issues/7213


kdorosh:
/kick docker build flake? https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=3283b0be-0b04-4280-a5f1-bbc82ce7b19d

elcasteel:
This will definitely fix the problem with concurrent writes in plugins. I'm pretty confident that it won't deadlock. Are we also planning to share a single translator between validation and translation?

kdorosh:
@elcasteel I think that's a good idea and the right direction forward; torn on whether we want to make that specific change in our backport however. Still reasoning about the shared memory here

elcasteel:
> @elcasteel I think that's a good idea and the right direction forward; torn on whether we want to make that specific change in our backport however. Still reasoning about the shared memory here

We shared the translator before we started reusing plugins so I don't believe there are any other issues with shared memory. 
I think if we are trying to avoid waiting on locks the best set up would be to give the validation webhook its own translator (and own validator) and then to have a separate validator for the proxy reconciler that can share a translator with translation. 
I think it's worth scale testing with a single translator for everything.

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
```diff
@@ -49,6 +50,9 @@ var (
 )
 
 type RouteReplacingSanitizer struct {
+	// note to devs: this can be called in parallel by the validation webhook and main translation loops at the same time
+	// any stateful fields should be protected by a mutex
+	lock             sync.RWMutex
```
elcasteel, line 55: None of these fields are modified except in NewRouteReplacingSanitizer so we don't need the mutex here yet. I do agree that having a note is good so devs are aware going forward that things are shared between multiple goroutines and may need to be protected.

kdorosh: I think that's a fair point, let's wait until we need synchronization here https://github.com/solo-io/gloo/pull/7207/commits/3fd22c3234ba94c3e610830c41b67c1e4d3dc973

projects/gloo/pkg/translator/translator.go
```diff
@@ -43,6 +44,7 @@ var (
 
 // translatorInstance is the implementation for a Translator used during Gloo translation
 type translatorInstance struct {
+	lock                      sync.RWMutex
```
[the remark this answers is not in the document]
kdorosh: https://github.com/solo-io/gloo/pull/7207/commits/d4fd93716b1d3649bb208532561e887aa92c84db

projects/gloo/pkg/translator/translator.go
```diff
@@ -68,6 +71,8 @@ func (t *translatorInstance) Translate(
 	proxy *v1.Proxy,
 ) (envoycache.Snapshot, reporter.ResourceReports, *validationapi.ProxyReport) {
 	// setup tracing, logging
+	t.lock.Lock()
```
yuval-k, line 74: how much time does translation take? could this be a problem if this take a long time?

kdorosh: that and potential scaling concerns are why I don't want to rush this change out. I'd like to run this in your environment and benchmark more thoroughly before merging.

tldr; i don't know

elcasteel: If we see a big slowdown after this, we can give the validation webhook one translator and translation+proxy_reconciler another translator which will separate the two truly independent uses of the lock.

EItanya: Agreed, I thought that's what `translatorInstance` was for, to instantiate one for each.

kdorosh: Yes; although I think it's a bit better to have validation and core translation use the same if possible. For example, think about two plugins resolving DNS async .. imagine if one copy of translator has DNS entries and the other does not, or something to this effect. Validation translation may succeed using its cached DNS data but then fail on the real translation. This type of bug is why I think they should be the same.

kdorosh:
/kick look into https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=25a10be6-8fc4-4a45-92ba-e3bad5e9d94c

projects/gloo/pkg/translator/translator.go
```diff
@@ -320,7 +331,9 @@ func MakeRdsResources(routeConfigs []*envoy_config_route_v3.RouteConfiguration)
 		if len(routeCfg.GetVirtualHosts()) < 1 {
 			continue
 		}
-		routesProto = append(routesProto, resource.NewEnvoyResource(routeCfg))
+		// important to clone, ownership will be transferred to goroutine serving xds snapshot.
+		// marshalling a proto to serve over grpc xds can mutate it, which can race with other reads (e.g. hashing last seen snapshot)
+		routesProto = append(routesProto, resource.NewEnvoyResource(proto.Clone(routeCfg)))
```
elcasteel, line 336: This is called from two places and I don't think we need the clone when it's called from route_replacing_sanitizer so we could pass in a bool to decide whether or not to clone (had this in a previous iteration of the clone removal change).

kdorosh: @yuval-k is considering another solution to this that updates hashstructure to be read-only so that we can keep these as they are, rather than add clones

yuval-k (approved):

projects/gloo/pkg/syncer/sanitizer/route_replacing_sanitizer.go
```diff
@@ -49,6 +49,8 @@ var (
 )
 
 type RouteReplacingSanitizer struct {
+	// note to devs: this can be called in parallel by the validation webhook and main translation loops at the same time
```
sam-heilbron, line 52: 👏

sam-heilbron (commented):
LGTM on first pass!

sam-heilbron (approved):

PR #7229, opened 2022-09-23
Add race detector Make target

elcasteel:
# Description

This change is mostly copied from: https://github.com/solo-io/solo-projects/pull/4093
Adding a make target to build gloo in a docker container with CGO and -race enabled to make it easy for developers without linux machines to test with the race detector. 
We will also publish the images at gloo:$VERSION-race so anyone can test a specific release.

Tested locally by running the new targets.


Makefile
```diff
@@ -395,6 +396,47 @@ gloo-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfi
 		-t $(IMAGE_REPO)/gloo:$(VERSION) $(QUAY_EXPIRATION_LABEL)
 
 #----------------------------------------------------------------------------------
+# Gloo with race detection enabled.
+# This is intended to be used to aid in local debugging by swapping out this image in a running gloo instance
+#----------------------------------------------------------------------------------
+GLOO_RACE_OUT_DIR=$(OUTPUT_DIR)/gloo-race
```
kdorosh, line 402: do you think it makes sense to always build gloo in docker? then we can consolidate make targets and then only have a flag / env var to build with race

elcasteel: it's a lot slower to build in docker and building gloo is on the critical path for most dev workflows so I don't want to add unnecessary overhead

Makefile
```diff
@@ -395,6 +396,47 @@ gloo-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfi
...
+_ := $(shell mkdir -p $(GLOO_RACE_OUT_DIR))
```
jenshu, line 403: can the `mkdir` be moved to within the `$(GLOO_RACE_OUT_DIR)/Dockerfile.build` target so we only create it if needed?

elcasteel: updated in both PRs

Makefile
```diff
@@ -395,6 +396,48 @@ gloo-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfi
...
 #----------------------------------------------------------------------------------
+# Gloo with race detection enabled.
+# This is intended to be used to aid in local debugging by swapping out this image in a running gloo instance
+#----------------------------------------------------------------------------------
+GLOO_RACE_OUT_DIR=$(OUTPUT_DIR)/gloo-race
+
+$(GLOO_RACE_OUT_DIR):
+	mkdir -p $@
+$(GLOO_RACE_OUT_DIR)/Dockerfile.build: $(GLOO_DIR)/Dockerfile
+	cp $< $@
```
jenshu, line 407:
```suggestion
$(GLOO_RACE_OUT_DIR)/Dockerfile.build: $(GLOO_DIR)/Dockerfile
        mkdir -p $(GLOO_RACE_OUT_DIR)
	cp $< $@
```
i think the mkdir can just be added to this target since all the other targets call this one

Makefile
```diff
@@ -395,6 +396,48 @@ gloo-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfi
...
+		.
+	touch $@
+# Build inside container as we need to target linux and must compile with CGO_ENABLED=1
+# We may be running Docker in a VM (eg, minikube) so be careful about how we copy files out of the containers
+$(GLOO_RACE_OUT_DIR)/gloo-linux-$(GOARCH): $(GLOO_RACE_OUT_DIR)/.gloo-race-docker-build
+	docker create -ti --name gloo-race-temp-container $(IMAGE_REPO)/gloo-race-build-container:$(VERSION) bash
+	docker cp gloo-race-temp-container:/gloo-linux-$(GOARCH) $(GLOO_RACE_OUT_DIR)/gloo-linux-$(GOARCH)
+	docker rm -f gloo-race-temp-container
+
+.PHONY: gloo-race
```
jenshu, line 427: can you add some comments on when to use gloo-race vs gloo-race-docker?

projects/gloo/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+#   anymore. Using this container you can build the binary on any architecture.
+#
+#####################################################################################################################
+ARG GO_BUILD_IMAGE
+FROM $GO_BUILD_IMAGE as build-env
+
+ARG VERSION
+ARG GCFLAGS
+ARG LDFLAGS
+ARG GITHUB_TOKEN
```
jenshu, line 16: is GITHUB_TOKEN used?

projects/gloo/Dockerfile
```diff
@@ -0,0 +1,40 @@
+#####################################################################################################################
```
jenshu, line 1: would it be confusing that there's both a `projects/gloo/Dockerfile` and `projects/gloo/cmd/Dockerfile` now? should this be moved to a subdir like `projects/gloo/race/Dockerfile`?

elcasteel: This is the same pattern we use in solo-projects where the Dockerfile for the containers we ship are in projects/$(PROJECT)/cmd and the Dockerfile for the build containers is in the root of the project directory. Both files have comments explaining what they're for so it doesn't seem particularly confusing to me.

jenshu (approved):

nfuden (approved):

PR #7259, opened 2022-09-28
Truncate status

kdorosh:
# Description

Truncate size of statuses written to k8s / etcd to protect against large keys stored in k8s backend. Max reason size is now 1kb and for subresource statuses altogether the max is around 1.36kb.

If truncation fails and the status is still too large, we will skip writing the status altogether to protect etcd / k8s backing storage.

# Context

Testing gloo edge at massive scale we saw the status written at times be much larger than was reasonable. This caused errors getting resources from the k8s apiserver (gateway errors over max grpc request size) which would in turn crash gloo. Since the statuses are read-only for users, if they get too large we now truncate to keep the resources stored in etcd a reasonable size.

Picks up behavior in https://github.com/solo-io/solo-kit/pull/524 and https://github.com/solo-io/solo-kit/pull/525 and https://github.com/solo-io/solo-kit/pull/526

Note that users can revert to the old original behavior by setting these env vars on the gloo pod:
- `DISABLE_MAX_STATUS_SIZE`="true"
- `DISABLE_TRUNCATE_STATUS`="true"

# Checklist:


kdorosh:
WIP until we can also pick up https://github.com/solo-io/solo-kit/pull/525

sam-heilbron (approved):

nfuden (approved):

PR #7277, opened 2022-10-02
Move status updates to their own goroutine

kdorosh:
# Description

This PR makes the following changes:
- moves gloo status reporting to its own goroutine so it's not on the hot path for translation. this means k8s api calls / rate limiting no longer affect or slow down updating our internal xds snapshot
- adds a new field to the Gloo `Settings` which allows users to disable the validation server even if `gateway.validation.disableTransformationValidation` is already set to true
- add metrics to transformation validation cache so we can determine how helpful disabling transformation validation will be

# Context

These optimizations came as a result of internal scale testing at 1k VS / RLC / AC / 3k US. When benchmarked, I noticed that the profile was really not all that hot, and came to realize most of the time was spent doing nothing / waiting for api calls to come back. In benchmarking at this scale, I saw speedups of a bit over 2x in time from when a new set of CRs were applied and to when this config was reflected in envoy and routeable.

# Testing

Manually verified metrics were working, deployed this image (enterprise version) in load testing env to validate performance improvements.

Relying on extensive regression testing of statuses (most of our tests wait for status to continue forward) as well.

# Checklist:


projects/gateway/pkg/syncer/translator_syncer.go
```diff
@@ -296,10 +296,11 @@ func (s *statusSyncer) setStatuses(list gloov1.ProxyList) {
 }
 
 func (s *statusSyncer) forceSync() {
-	select {
-	case s.syncNeeded <- struct{}{}:
-	default:
+	if len(s.syncNeeded) > 0 {
```
nfuden, line 299: nit: lets add a comment on this. Not because its often needed but many engineers coming to go for the first time may be confused about the difference in cap and len for buffered channels and may make an error on this.

nfuden (approved):
I like this. It does make me feel we need to start using a wrapper for time so we can simulate it for tests (time.after on the retry channel) but thats not needed for this pr so I am happy with it as is.
My one other statement is really a nit so I leave it to you and the second reviewer to decide if we care about it

sam-heilbron (approved):

Issue #6777, opened 2022-07-20
Provide validation for the Ratelimit CRs


asayah:
### Is your feature request related to a problem? Please describe.

Today we enforce schemas validation for RateLimit CRs 
The need is to support Server validation for Rate limiting CR 

This work should also enable the metric for ratelimitconfigs:  https://docs.solo.io/gloo-edge/master/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/settings.proto.sk/#observabilityoptions

krisztianfekete:
Duplicate of https://github.com/solo-io/gloo/issues/6533

nfuden:
Part of https://github.com/solo-io/gloo/issues/2797

nfuden:
@jenshu  has context and can help to pair on this

jackstine:
Here is a list of issues that we can add to make the necessary changes to the validation webhook for Rate Limit and future resources.
[] https://github.com/solo-io/gloo/issues/4783
[] https://github.com/solo-io/gloo/issues/7271
[] https://github.com/solo-io/gloo/issues/7272
[] https://github.com/solo-io/solo-projects/issues/3558
[] https://github.com/solo-io/gloo/issues/2797

jackstine:
Validation for Rate Limit Configs is now available in V1.13.0

bdecoste:
Seeing an issue with the RLC validation in 1.13.10:

```
$ cat ratelimitconfig-validation.json
{
    "request": {
        "uid": "abc",
        "dryRun": true,
        "kind": {
            "group": "ratelimit.solo.io",
            "version": "v1alpha1",
            "kind": "RateLimitConfig"
        },
        "namespace": "gloo-system",
        "object": {
            "apiVersion": "ratelimit.solo.io/v1alpha1",
            "kind": "RateLimitConfig",
            "metadata": {
                "labels": {
                    "central-id": "200004162",
                    "proxy": "abc"
                },
                "name": "abc",
                "namespace": "gloo-system"
            },
            "spec": {
                "rawsinvalid": {
                    "descriptors": [
                        {
                            "key": "generic_key",
                            "value": "abc",
                            "rateLimit": {
                                "requestsPerUnit": 50,
                                "unit": "SECOND"
                            }
                        }
                    ],
                    "rateLimit": [
                        {
                            "action": [
                                {
                                    "genericKey": {
                                        "descriptorValues": "abc"
                                    }
                                }
                            ]
                        }
[...]

Issue #2797, opened 2020-04-15
Provide webhook validation for all Gloo resources


lgadban:
Ideally, the existing validating webhook should provide validation for some of the resources defined in the `gloo.solo.io` api group.
Validation could be useful for:

- [ ]  `upstreamgroups.gloo.solo.io`
- [x]  `upstreams.gloo.solo.io` (https://github.com/solo-io/gloo/issues/5173)
- [ ]  `settings.gloo.solo.io` (Possibly?)
- [ ] `enterprise.gloo.solo.io`
- [x] Secrets (https://github.com/solo-io/gloo/pull/5470)
- [ ] ConfigMaps
- [ ] https://github.com/solo-io/gloo/issues/5636
- [ ] https://github.com/solo-io/gloo/issues/5810

lgadban:
Somewhat related to https://github.com/solo-io/gloo/issues/2114

lgadban:
This means that incorrect rate limit settings being applied can bring down the control plane

lgadban:
Edited the initial issue to include the `enterprise.gloo.solo.io` api group, mainly to perform validation on `AuthConfig` resources.

For example, this could prevent an existing [issue with malformed `AuthConfigs` causing segfaults](https://github.com/solo-io/gloo/issues/3358) (although we should fix that issue as well, but validation will still be useful)

sam-heilbron:
Part 1, Upstreams: https://github.com/solo-io/gloo/issues/5173

antonioberben:
Hi, there is this issue, related with the topic of this milestone but not covered by the requirements.

https://github.com/solo-io/gloo/issues/5636

Should it be part of this? Thanks you

kdorosh:
further considerations:
- requires helm hook on upgrade to ensure valid with both new and old gloo validation loops
- users may find webhook preventing deletion of k8s services not acceptable?
  - we can add feature to require label on svcs to be seen (and validated) by gloo to limit impact here

as part of https://github.com/solo-io/gloo/issues/5721#issuecomment-1069500839

nfuden:
Perhaps we split each resource into its own issue for now and link em here

chrisgaun:
ratelimiting config and root action are most important ones that have come up recently.

chrisgaun:
Related issue. Validation becomes slow at scale https://github.com/solo-io/solo-projects/issues/4031

PR #7284, opened 2022-10-04
Generic Validation Webhook

jackstine:
# Description

Providing Generic Validation of resources.

Using the Hashable Resource, we can use a generic type to validate resources in the API Snapshot.
Both gloo and gateway resources now have to use full translation to make validations. All proxies are validated, regarless of resource that is added, even if it does not impact the proxy. As opposed to before, we had filtered the proxies based off the resource that we were modifying.

This has uncovered other errors in the translation that were not captured before in testing. Some testing has been changed to validate these errors that are now properly being reported on the resource validation.

Currently, resources have to be added to the GVK supported maps, for the resource to be accepted for validation.
`GvkSupportedDeleteGatewayResources`, ` gloovalidation.GvkToSupportedDeleteGlooResources`, `GvkSupportedValidationGatewayResources`, and `gloovalidation.GvkToSupportedGlooResource`.

# Context

Should be able to add new resources to gloo just by changing the helm chart values in the admission `install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml` and adding the gvk to the support list.  This will be done in a future issue  https://github.com/solo-io/gloo/issues/6777

# Checklist:

- [x] bump solo kit


projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,15 +394,103 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
...
 	case ListGVK:
 		return wh.validateList(ctx, admissionRequest.Object.Raw, dryRun)
+		/*
+			 	Currently having issues with the following resources as they are not Input Resources
+				// Artifacts          gloo_solo_io.ArtifactList
+				// Endpoints          gloo_solo_io.EndpointList
+				// Secrets            gloo_solo_io.SecretList
+				// Ratelimitconfigs   github_com_solo_io_gloo_projects_gloo_pkg_api_external_solo_ratelimit.RateLimitConfigList
+
+				// will have to check out RateLimitConfigs, not sure what causes it to not be a Hashable Input Resource
```
kdorosh, line 426: this is because this CR is generated by skv2 instead of solo-kit :cry:

jackstine: ah

projects/gateway/pkg/validation/gloo_resource_validator.go
```diff
@@ -0,0 +1,36 @@
...
+// DeleteGlooResourceValidator allows validation of gloo.solo.io resources
+type GlooResourceValidator interface {
+	CreateModifiedRequest(resource resources.Resource) *validation.GlooValidationServiceRequest
+}
+
+var GvkToGlooValidator = map[schema.GroupVersionKind]GlooResourceValidator{
+	gloov1.UpstreamGVK: UpstreamValidator{},
+}
+
+var GvkToDeleteGlooValidator = map[schema.GroupVersionKind]DeleteGlooResourceValidator{
```
kdorosh, line 33: to make this more generic, can we have a single interface that calls a full gloo translation (gateway -> gloo in a row, just like the gloo syncer today)? this way, we just register the gvk (regardless of which interface), add to the snapshot, and then translate and use resource reports and owner refs to determine if our resource was errored / problematic with the current snapshot

projects/gateway/pkg/validation/route_table_validator.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	var proxiesToConsider []string
+	for proxy := range affectedProxies {
+		proxiesToConsider = append(proxiesToConsider, proxy)
+	}
+	sort.Strings(proxiesToConsider)
+
+	return proxiesToConsider
+}
```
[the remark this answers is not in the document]
nfuden: Shouldnt a given gvk also be able to note if it needs a full translation run? I am wondering purely for the extensions where they need to be translated but may not be needed if it is a new resource rather than a delete or modify and purely needs to run the rate limit server specific translation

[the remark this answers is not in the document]
jackstine: are you proposing that we validate all the proxies at once, instead of filtering them?
just iterate over the map of `proxyname -> gateway list` (`gatewaysByProxy`)?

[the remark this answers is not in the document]
jackstine: are you proposing we iterate over the entire proxy list instead of filtering them?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,160 +394,68 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
...
-	if !shouldValidate {
-		return nil, nil
-	}
-
-	if reports, err = wh.validator.ValidateGateway(ctx, &gw, isDryRun(admissionRequest)); err != nil {
-		return reports, &multierror.Error{Errors: []error{errors.Wrapf(err, "Validating %T failed", gw)}}
+func (wh *gatewayValidationWebhook) deleteRef(ctx context.Context, gvk schema.GroupVersionKind, ref *core.ResourceRef, admissionRequest *v1beta1.AdmissionRequest) (*validation.Reports, *multierror.Error) {
+	err := wh.validator.ValidateDeleteRef(ctx, gvk, ref, isDryRun(admissionRequest))
+	if err != nil {
+		return nil, &multierror.Error{Errors: []error{errors.Wrapf(err, "failed validatin deletion of resource namespace: %s name: %s", ref.GetNamespace(), ref.GetName())}}
```
nfuden, line 410:
```suggestion
		return nil, &multierror.Error{Errors: []error{errors.Wrapf(err, "failed validating the deletion of resource namespace: %s name: %s", ref.GetNamespace(), ref.GetName())}}
```

projects/gateway/pkg/validation/gloo_resource_validator.go
jackstine: For Gloo Resources, we could just add the resource or delete the reference 

```golang
	snapCopy := s.latestSnapshot.Clone() // cloning can mutate so we need a write lock
	s.lock.Unlock()
	if req.isDelete {
		snapCopy.remove(ref)
	} else {
		snapCopy.AddToResourceList(req.resource)
	}
```

projects/gateway/pkg/validation/validator.go
```diff
@@ -358,12 +332,45 @@ func (v *validator) validateSnapshot(ctx context.Context, apply applyResource, d
...
+func (v *validator) ValidateDeletedGvk(ctx context.Context, gvk schema.GroupVersionKind, resource resources.Resource, dryRun bool) error {
+	_, supportedGatewayResource := GvkSupportedDeleteGatewayResources[gvk]
+	_, supportedGlooResource := gloovalidation.GvkToSupportedDeleteGlooResources[gvk]
+	if supportedGatewayResource || supportedGlooResource {
+		_, err := v.validateGatewayResource(ctx, resource, true, dryRun, true)
+		return err
+	}
+	contextutils.LoggerFrom(ctx).Debugf("unsupported validation for resource delete ref namespace [%s] name [%s] group [%s] kind [%s]", resource.GetMetadata().GetNamespace(), resource.GetMetadata().GetName(), gvk.Group, gvk.Kind)
+	return nil
+}
```
jackstine, line 354: both gloo and gateway resources must be validated against all the proxies and gateways

projects/gateway/pkg/validation/README.md
```diff
@@ -0,0 +1,6 @@
+
+When creating a new resource Group extend the Validator.go method`ValidationIsSupported()`
```
kdorosh, line 2: i'm wondering if in the final version we can derive this from the generated gloo snapshot struct (everything in there gets automatically added) so as we add new types we don't need to update this function. although what you have here is a million times better than what we have today so i'm not considering this blocking in favor of incremental improvements

projects/gateway/pkg/validation/validator.go
```diff
@@ -81,57 +80,50 @@ var _ Validator = &validator{}
...
-	ValidateVirtualService(ctx context.Context, vs *v1.VirtualService, dryRun bool) (*Reports, error)
-	ValidateDeleteVirtualService(ctx context.Context, vs *core.ResourceRef, dryRun bool) error
-	ValidateRouteTable(ctx context.Context, rt *v1.RouteTable, dryRun bool) (*Reports, error)
-	ValidateDeleteRouteTable(ctx context.Context, rt *core.ResourceRef, dryRun bool) error
-	ValidateUpstream(ctx context.Context, us *gloov1.Upstream, dryRun bool) (*Reports, error)
-	ValidateDeleteUpstream(ctx context.Context, us *core.ResourceRef, dryRun bool) error
-	ValidateDeleteSecret(ctx context.Context, secret *core.ResourceRef, dryRun bool) error
+	ValidateModifiedGvk(ctx context.Context, gvk schema.GroupVersionKind, resource resources.Resource, dryRun bool) (*Reports, error)
+	ValidateDeletedGvk(ctx context.Context, gvk schema.GroupVersionKind, resource resources.Resource, dryRun bool) error
+	ValidationIsSupported(gvk schema.GroupVersionKind) bool
```
kdorosh, line 85: i can't understate how happy i am about this :100:

projects/gateway/pkg/validation/validator.go
```diff
@@ -375,22 +382,20 @@ func (v *validator) ValidateList(ctx context.Context, ul *unstructured.Unstructu
 	)
 
 	v.lock.Lock()
-	defer v.lock.Unlock()
 	originalSnapshot := v.latestSnapshot.Clone()
+	defer v.lock.Unlock()
```
kdorosh, line 386: why move this?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook_test.go
```diff
@@ -581,12 +572,11 @@ type mockValidator struct {
 	fValidateList                 func(ctx context.Context, ul *unstructured.UnstructuredList, dryRun bool) (*validation.Reports, *multierror.Error)
 	fValidateGateway              func(ctx context.Context, gw *v1.Gateway, dryRun bool) (*validation.Reports, error)
 	fValidateVirtualService       func(ctx context.Context, vs *v1.VirtualService, dryRun bool) (*validation.Reports, error)
-	fValidateDeleteVirtualService func(ctx context.Context, vs *core.ResourceRef, dryRun bool) error
+	fValidateDeleteVirtualService func(ctx context.Context, rt *core.ResourceRef, dryRun bool) error
```
kdorosh, line 575: why these changes? i thought vs was correct?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,160 +394,75 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
 	isDelete := admissionRequest.Operation == v1beta1.Delete
 	dryRun := isDryRun(admissionRequest)
 
-	switch gvk {
-	case ListGVK:
+	if gvk == ListGVK {
 		return wh.validateList(ctx, admissionRequest.Object.Raw, dryRun)
```
kdorosh, line 398: perhaps we need to open an issue for support for delete list? realizing that may not be supported today? (out of scope obviously for this ticket)

jackstine: Yeah I need to clean up everything from the `validating_admission_webhook_test.go`.  All the methods were mocked to make the tests pass.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook_test.go
jackstine: Yeah I need to clean up everything from the validating_admission_webhook_test.go. All the methods were mocked to make the tests pass.

projects/gateway/pkg/validation/README.md
jackstine: do we want to support all resources?
https://github.com/solo-io/gloo/blob/386d929e897f5f8e0fe0980f33ecd46113916a9e/projects/gloo/pkg/api/v1/gloosnapshot/api_snapshot.sk.go#L895

kdorosh: i think we should. i don't have a strong preference if we add it here or in a follow up; but if we add extra unit tests to prove the other types and this PR gets large then we may want to split the PR in two (refactor to generic, then add new functionality)

jackstine: I am good with splitting them up, since we have an issue for this
https://github.com/solo-io/gloo/issues/2797

changelog/v1.13.0-beta25/generic_validation_webhook.yaml
```diff
@@ -0,0 +1,12 @@
+changelog:
+  - type: DEPENDENCY_BUMP
+    description: >
+      To support a generic validation webhook, solo kit will have to be bumped with the necessary changes
+      to make resources interface the Hashable Input Resource.
+    dependencyOwner: solo-io
+    dependencyRepo:  solo-kit
+    dependencyTag: todo
+  - type: BREAKING_CHANGE
+    description: generic validation webhook
```
kdorosh, line 10: this is a tricky one; i know from a go library perspective it is a breaking change (and thus in solo-kit this is how i'd like to designate it). since this is customer facing, and from a user's perspective there are no breaking changes, I'd prefer to mark this as `NON_USER_FACING` (since no action required of them)

for more detailed discussion on this see https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1666133024820829

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
kdorosh: i opened https://github.com/solo-io/gloo/issues/7352 to track

projects/gateway/pkg/validation/validator.go
```diff
@@ -215,82 +241,65 @@ func (v *validator) gatewayUpdate(snap *gloov1snap.ApiSnapshot) bool {
...
+	} else {
+		snapshotClone.AddOrReplaceToResourceList(opts.Resource)
+	}
 
-	gatewaysByProxy := utils.GatewaysByProxyName(snapshotClone.Gateways)
+	// only set the glooResource if we are validating a gloo resource
+	var glooResource resources.Resource
+	if opts.Gvk.Group == gloovalidation.GlooGroup {
+		glooResource = opts.Resource
+	}
```
kdorosh, line 292: for now this assumption holds, but i don't think we want to do this long term or for an entirely generic solution. IIRC other resources in a gloo snapshot don't have the gloo group (e.g. secrets, configmaps, auth configs, rate limit configs, etc)

kdorosh: to be clear, i think we should remove this check in this PR since we shouldn't really need to do it

projects/gateway/pkg/validation/validator.go
kdorosh: bump on this

projects/gateway/pkg/validation/validator.go
```diff
@@ -425,365 +472,82 @@ func (v *validator) processItem(ctx context.Context, item unstructured.Unstructu
...
-	if !v.ready() {
-		return errors.Errorf("Gateway validation is yet not available. Waiting for first snapshot")
-	}
-	snap := v.latestSnapshot.Clone()
-
-	vs, err := snap.VirtualServices.Find(vsRef.Strings())
-	if err != nil {
-		// if it's not present in the snapshot, allow deletion
-		return nil
+func (v *validator) copySnapshot(ctx context.Context, dryRun bool) (*gloosnapshot.ApiSnapshot, error) {
```
kdorosh, line 486: just for future devs can we comment this noting that it's not thread safe / latest snapshot needs to be cloned under lock (this will both read and write latest snapshot)

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -492,7 +486,14 @@ var _ = Describe("Validator", func() {
...
 			It("accepts the vs", func() {
-				v.validationFunc = failProxy
+				// TODO-JAKE not sure if we want this to be a ValidateAccept
+				v.glooValidator = ValidateAccept
+				/*
+					So it looks like the gateways have to have selectors that match the VS labels and be in it's ns
+					if this is not true, then we do not need to validate the virtual service against this gateway.
+
+					So the question is why is the VS failing?
+				*/
```
kdorosh, line 496: did we sort this?

kdorosh: if not, may be worth a huddle

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -746,76 +748,86 @@ var _ = Describe("Validator", func() {
...
-				err = v.ValidateDeleteVirtualService(context.TODO(), ref, false)
+				err = v.ValidateDeletedGvk(context.TODO(), v1.VirtualServiceGVK, snap.VirtualServices[0], false)
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring(
-					VirtualServiceDeleteErr([]*core.ResourceRef{
-						{Name: defaults.GatewayProxyName, Namespace: ns},
-						{Name: defaults.GatewayProxyName + "-ssl", Namespace: ns},
-						{Name: defaults.GatewayProxyName + "-hybrid", Namespace: ns},
-					}).Error()))
+				// TODO-JAKE probably not the error we want to be having.  This looks like the default error.
```
kdorosh, line 753: same note about todo being sorted and potential huddle

projects/gateway/pkg/validation/validator_test.go
jackstine: I think this is acceptable because now we should have a report for each proxy that we validate.

projects/gateway/pkg/validation/validator_test.go
jackstine: I think this is acceptable as well.

projects/gateway/pkg/validation/validator.go
jackstine: this is correct we have `enterprise.gloo.solo.io` and `ratelimit.solo.io` for authConfig and ratelimit respectively.  We could use the `IsSupportive()` functions.  I was not aware of this.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,160 +394,75 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
...
-	if err := ul.UnmarshalJSON(rawJson); err != nil {
-		return nil, &multierror.Error{Errors: []error{WrappedUnmarshalErr(err)}}
-	}
-	if reports, errs = wh.validator.ValidateList(ctx, &ul, dryRun); errs != nil {
-		return reports, errs
+	if isDelete {
+		return wh.deleteRef(ctx, gvk, ref, admissionRequest)
+	} else {
+		return wh.validateGvk(ctx, gvk, ref, admissionRequest)
 	}
```
nfuden, line 404: nit: its more go like not to nest this final else
```suggestion
	} 
	return wh.validateGvk(ctx, gvk, ref, admissionRequest)
	
```

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,160 +394,75 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
...
-	shouldValidate, shouldValidateErr := wh.shouldValidateResource(ctx, admissionRequest, &vs, &oldVs)
-	if shouldValidateErr != nil {
-		return nil, &multierror.Error{Errors: []error{shouldValidateErr}}
-	}
-	if !shouldValidate {
-		return nil, nil
+func (wh *gatewayValidationWebhook) validateGvk(ctx context.Context, gvk schema.GroupVersionKind, ref *core.ResourceRef, admissionRequest *v1beta1.AdmissionRequest) (*validation.Reports, *multierror.Error) {
+	var reports *validation.Reports
+	if !wh.validator.ModificationIsSupported(gvk) {
+		contextutils.LoggerFrom(ctx).Debugf("unsupported validation for resource namespace [%s] name [%s] group [%s] kind [%s]", ref.GetNamespace(), ref.GetName(), gvk.Group, gvk.Kind)
```
nfuden, line 425: Question are we guaranteed that ref is non-nil?

projects/gateway/pkg/validation/validator.go
jackstine: I think I was thinking at the moment the only thing that needs to be locked is the cloning of the snapshot. So I unlocked it after the `Clone()`  But if you look later we add/delete the resource to the latest snapshot. So I came back and added the `defer` statement, but did not move it back to after the `Lock()`.

projects/gateway/pkg/validation/validator.go
jackstine: nice call out.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
jackstine: line 327

```go
	ref := &core.ResourceRef{
		Namespace: req.Namespace,
		Name:      req.Name,
	}

	reports, validationErrs := wh.validateAdmissionRequest(ctx, gvk, ref, req)
```

projects/gateway/pkg/validation/validator.go
jackstine: I think we just need to exclude any gateway resources
```go
	var glooResource resources.Resource
	if opts.Gvk.Group != GatewayGroup {
		glooResource = opts.Resource
	}
```

jackstine: I also wonder what happens if we propagate all resources to the gloo validator.

jackstine: Yeah looks like we can get rid of this entirely.  I will see if it passes CI in s-p

jackstine:
Another thing we can add is that `ignoreGlooValidationFailure` is deprecated.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -395,160 +394,74 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
...
-
-	shouldValidate, shouldValidateErr := wh.shouldValidateResource(ctx, admissionRequest, &vs, &oldVs)
-	if shouldValidateErr != nil {
-		return nil, &multierror.Error{Errors: []error{shouldValidateErr}}
-	}
-	if !shouldValidate {
-		return nil, nil
+func (wh *gatewayValidationWebhook) validateGvk(ctx context.Context, gvk schema.GroupVersionKind, ref *core.ResourceRef, admissionRequest *v1beta1.AdmissionRequest) (*validation.Reports, *multierror.Error) {
+	var reports *validation.Reports
+	if !wh.validator.ModificationIsSupported(gvk) {
```
sam-heilbron, line 423: I like the explicit check in code here. It also feels like this is the responsibility of the webhook configuration itself: https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml#L26

What do you see the benefit of having this check in code, in addition to the webhook config?

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/enterprise_snapshot.sk.go
```diff
@@ -52,6 +54,53 @@ func (s EnterpriseSnapshot) HashFields() []zap.Field {
...
+		for i, res := range s.AuthConfigs {
+			if refKey == res.GetMetadata().Ref().Key() {
+				s.AuthConfigs[i] = typed
+				updated = true
+			}
+		}
+		if !updated {
+			s.AuthConfigs = append(s.AuthConfigs, typed)
+		}
+		s.AuthConfigs.Sort()
```
sam-heilbron, line 97: Why is the sort necessary?

projects/gloo/pkg/validation/gloo_validator.go
```diff
@@ -0,0 +1,93 @@
...
+
+const GlooGroup = "gloo.solo.io"
+
+var GvkToSupportedDeleteGlooResources = map[schema.GroupVersionKind]bool{
+	gloov1.UpstreamGVK: true,
+	gloov1.SecretGVK:   true,
+}
+
+var GvkToSupportedGlooResources = map[schema.GroupVersionKind]bool{
+	gloov1.UpstreamGVK: true,
```
sam-heilbron, line 26: It may be useful to callout why Secrets are not supported as a modifiable GVK. As a developer I see this difference and wonder "why can we support validating a secret deletion but not modification"

projects/gloo/pkg/validation/gloo_validator.go
```diff
@@ -0,0 +1,93 @@
...
+		// if no proxy was passed in, call translate for all proxies in snapshot
+		proxiesToValidate = snapshot.Proxies
+	}
+
+	if len(proxiesToValidate) == 0 {
+		// This can occur when a Gloo resource (Upstream), is modified before the ApiSnapshot
+		// contains any Proxies. Orphaned resources are never invalid, but they may be accepted
+		// even if they are semantically incorrect.
+		// This log line is attempting to identify these situations
+		contextutils.LoggerFrom(ctx).Warnf("found no proxies to validate, accepting update without translating Gloo resources")
```
sam-heilbron, line 71: nit: can we just return early here?

projects/gloo/pkg/validation/gloo_validator.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	params := plugins.Params{
+		Ctx:      ctx,
+		Snapshot: snapshot,
+	}
+	for _, proxy := range proxiesToValidate {
+		xdsSnapshot, resourceReports, proxyReport := gv.glooTranslator.Translate(params, proxy)
```
sam-heilbron, line 79: nit: A comment above this loop may be helpful to link the code that we use for validation to the code that we use for translation. It might make it easier for developers to find what we're trying to mirror, and also identify situations where we need to update this validator code

projects/gloo/pkg/validation/server.go
```diff
@@ -187,6 +191,7 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
 
 	var validationReports []*validation.ValidationReport
 	var proxiesToValidate v1.ProxyList
+
```
sam-heilbron, line 194: Can we call ValidateGloo from within this method since they should be doing the same thing?

sam-heilbron (commented):
This looks awesome! Mostly minor comments

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/enterprise_snapshot.sk.go
jackstine: I do not know why. But all lists on the snapshots are sorted. 

I am just keeping the expected state of the snapshot, as every author before me has.  From what I have seen, whenever an element has been added to the snapshot lists, they have always been sorted.

Snapshots come from the emitter, and are sorted when they are added to the snapshot.
https://github.com/solo-io/solo-kit/blob/master/pkg/code-generator/codegen/templates/snapshot_emitter_template.go#L348

projects/gloo/pkg/validation/gloo_validator.go
jackstine: I will delegate this to the other issue https://github.com/solo-io/gloo/issues/6777.  I do not know why only the deletion of secrets were supported and not the modification as well.

I assume that they will work equally expected with the generic changes. but would need valid testing.

projects/gloo/pkg/validation/gloo_validator.go
jackstine: Yeah I like that, I had initially had this, when the code was in server.go, but had gotten rid of it from some refactoring.

projects/gloo/pkg/validation/server.go
jackstine: ` Validate(ctx context.Context, req *validation.GlooValidationServiceRequest) ` is not generic.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
jackstine: I will delegate this to the other issue https://github.com/solo-io/gloo/issues/6777. See conversion that Kevin asked as well. 
https://github.com/solo-io/gloo/pull/7284#discussion_r999849312

projects/gateway/pkg/validation/validator.go
```diff
@@ -80,58 +100,66 @@ var _ Validator = &validator{}
...
-	ValidateVirtualService(ctx context.Context, vs *v1.VirtualService, dryRun bool) (*Reports, error)
-	ValidateDeleteVirtualService(ctx context.Context, vs *core.ResourceRef, dryRun bool) error
-	ValidateRouteTable(ctx context.Context, rt *v1.RouteTable, dryRun bool) (*Reports, error)
-	ValidateDeleteRouteTable(ctx context.Context, rt *core.ResourceRef, dryRun bool) error
-	ValidateUpstream(ctx context.Context, us *gloov1.Upstream, dryRun bool) (*Reports, error)
-	ValidateDeleteUpstream(ctx context.Context, us *core.ResourceRef, dryRun bool) error
-	ValidateDeleteSecret(ctx context.Context, secret *core.ResourceRef, dryRun bool) error
+	// ValidateModifiedGvk validate the deletion of a resource.
+	ValidateModifiedGvk(ctx context.Context, gvk schema.GroupVersionKind, resource resources.Resource, dryRun bool) (*Reports, error)
+	// ValidateDeletedGvk validate the creation or update of a resource.
```
kdorosh, line 107: nit; i think these comments are switched

projects/gloo/pkg/validation/gloo_validator.go
kdorosh: agree with the above here. in first pass, prefer to keep behavior the same and then add testing / support for add and deletion of all resources

kdorosh: i suspect the lack of support for the modification of secrets was just planned future work

projects/gloo/pkg/validation/server.go
```diff
@@ -220,8 +229,37 @@ func (s *validator) Validate(ctx context.Context, req *validation.GlooValidation
...
+	// the gateway validator will call this function to validate Gloo resources.
+	s.lock.Lock()
+	// we may receive a Validate call before a Sync has occurred
+	if s.latestSnapshot == nil {
+		s.lock.Unlock()
+		return nil, eris.New("proxy validation called before the validation server received its first sync of resources")
+	}
+	snapCopy := s.latestSnapshot.Clone() // cloning can mutate so we need a write lock
+	s.lock.Unlock()
+	if resource != nil {
```
kdorosh, line 245: do we want dpanic if resource is nil?

kdorosh (commented):
lgtm, just non blocking nits / comment nits

projects/gloo/pkg/validation/server.go
jackstine: What is the reason to have a panic here if the resource is nil? It would have been validating the proxy, and current snapshot if the resource is nil.  This should be expected behavior.
Validate the snapshot, without adding/deleting/modifying a resource.

kdorosh: The idea was that if resource is nil here, it's probably a developer error (today) since this is only called as part of the validation webhook AFAIK. Although I'm on board with keeping this as is, that's reasonable behavior too. If we wanted to assert the validation webhook resource being validated is not nil it may be better to do that there.

kdorosh (approved):

projects/gloo/pkg/api/v1/enterprise/options/extauth/v1/enterprise_snapshot.sk.go
kdorosh: We recommend sorting resources before / after all modifications to the snapshot to ensure idempotency of generated xds config. This matters more when we update and notice new resources off our k8s watches, but is generally a good practice during our translation itself too

projects/gloo/pkg/validation/server.go
sam-heilbron: Oh yeah, sorry I couldn't find a way to comment on the code that is 10 lines lower.

This method is the interface for the grpc client, but it's implementation is to receive the request, and then perform validation. That validation seems to duplicate what ValidateGloo does (unless im missing something). So i fiugred instead of having duplicate code that could drift, we just call ValidateGloo from within this method

projects/gloo/pkg/validation/gloo_validator.go
sam-heilbron: Oh yeah I agree that it's just work we hadn't done, more raising that a comment might be helpful since my natural instinct is to see the two lists and wonder why they're different.

projects/gloo/pkg/validation/server.go
jackstine: gotcha.  Yeah that makes sense

sam-heilbron (approved):
🎉 awesome work!

Issue #7301, opened 2022-10-06
Onboarding / Dev documentation updates


ashishb-solo:
A low priority timeboxed space to update various documentation pain points found during onboarrding.

PR #7312, opened 2022-10-10
Added `make help` target

Bslabe123:
# Description

Running `make help` now lists descriptions for each make target

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7148

PR #7318, opened 2022-10-12
build-test-assets make target removal and changing build tool to buildx

Bslabe123:
# Description

UX improvements to our makefile and changing of the build tool to buildx.

# Context

Some of our make targets are not intended for developer use and are only used by the build system/CI, only make targets intended for developer use should have comments.

Fix for issue introduced in https://github.com/solo-io/gloo/pull/7312 which breaks build-test-assets make target: Instead of fixing this bug we removed build-test-assets due to it not being used anywhere and replicable by running `make build-test-chart glooctl`.

The changes can be proven safe by running them manually.

Many of the arch specific issues encountered in enterprise were fixed by using the buildx plugin which has support for multi-platform builds. This is a safe change since we don't modify any existing `docker build` behavior.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7301
resolves https://github.com/solo-io/gloo/issues/7384


nfuden:
Lets get approval from an arm user and an amd user for this as well

Makefile
```diff
@@ -663,8 +663,11 @@ ifeq ($(CREATE_ASSETS),"true")
 	DOCKER_IMAGES := docker
 endif
 
+# Tag all existing images and retags each one with the repo specified in {IMAGE_REPO}
+# this ensures we have redundancy and can pull images from multiple registries
+# in our case {RETAG_IMAGE_REGISTRY} and {IMAGE_REPO}
 .PHONY: docker-push-retag
-docker-push-retag: ## check if all images are already built for {RETAG_IMAGE_REGISTRY}, and build/retag them respectively with tags for the repository specified by {IMAGE_REPO}.
```
ianmacclancy, line 667: Can we add a note to the PR why we are reverting the move to ## with make help? This will now move us to two different commenting schemes in the make file

sam-heilbron: I think we have 2 commenting schemes because there are two benefits to it. 
The existing commenting scheme, which this target uses, tries to clarify what targets are intended to do.
make help was introduced to try to be the introduction point to targets that developers should use. It alone can't provide all the necessary information.

I think both styles are valuable (make help, as a high level, this is a useful command to devs) and inline comments to clarify how it should be run.  `docker-push-retag` for example should not be run by developers and is just used by CI, that's why I recommended we not include it in make help

nfuden: We should consider laying this out in a style guide for our make files too

Bslabe123: Would it make sense to have a convention for differentiating between make targets that are and aren't meant to be used by developers?

Makefile
```diff
@@ -768,9 +771,6 @@ push-docker-images-arm-to-kind-registry:
 # The following targets are used to generate the assets on which the kube2e tests rely upon.
 # The Kube2e tests will use the generated Gloo Edge Chart to install Gloo Edge to the GKE test cluster.
 
-.PHONY: build-test-assets
```
sam-heilbron, line 771: We are removing a Make target. This could impact other developers, so I think we need to call out loudly in the PR context, what this target did and why we feel safe removing it

sam-heilbron (dismissed):

Bslabe123:
/kick : [Fail] Leader Startup Action [It] StartupAction executed once, if elected

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d8466d5f-5cd0-4446-b9b5-8cb5e994e42b

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=74386a44-4b91-48e0-9563-fe7013a9e838

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=35f79123-f621-4a07-82c3-e38aec23eceb

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=073d5fe6-f2f9-48bf-a3c0-74dcf04706ff

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4c733b12-7554-4d1b-a2f7-4091adc1fa41

sam-heilbron (commented):
This looks generally good! I think there is some overlap between this issue and the license key work. For example, in the testing steps there is a reference to supplying a license key to gloo-fed which I feel is irrelevant to this work. Also, the changelog implies we are fixing this make target instead of removing it.

Let's huddle when you get a chance just to clarify these parts and make a plan.

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=516db051-9e48-4512-9517-08daefe51159

jackstine:
if we are moving toward using `build buildx` then we need to add in the changes for the `extended-docker.sh`
https://github.com/solo-io/solo-projects/blob/master/ci/extended-docker/extended-docker.sh

jackstine:
Can you add some notes to the makefile to make it clear that to run the make file to mimic the effects of CI that the environment variables `TAGGED_VERSION` and the `CREATE_TEST_ASSETS` need to be set.

Also can something similar to the following be added to the extended file. So that others can run locally. Obviously, the IMAGE_REPO can be anything.

```
VERSION=0.0.1-jake TAGGED_VERSION=0.0.1-jake IMAGE_REPO=localhost:5000 ci/extended-docker/extended-docker.sh
```

Note that `CREATE_TEST_ASSETS` has to be set because we are pushing assets locally.

ci/extended-docker/README.md
```diff
@@ -0,0 +1,6 @@
+# Extended Docker Script
+
+## Running Locally
+
+If needed for testing, run from the root of this repo:
+VERSION=<your-version> TAGGED_VERSION=<your-version> IMAGE_REPO=localhost:5000 ci/extended-docker/extended-docker.sh
```
jackstine, line 6: nice!

jackstine:
Looking good, I was able to build this locally after merging this PR.
The Race image cannot be built as is on M1 machines.
https://github.com/solo-io/gloo/pull/7374

Makefile
```diff
@@ -730,7 +753,7 @@ ifeq ($(CREATE_ASSETS), "true")
 	docker push $(IMAGE_REPO)/sds:$(VERSION) && \
 	docker push $(IMAGE_REPO)/access-logger:$(VERSION)
 endif
-
+# To mimic the effects of CI, both TAGGED_VERSION and CREATE_TEST_ASSETS need to be set
```
sam-heilbron, line 756: nit: It looks like this target is gated by `CREATE_ASSETS`, can we include that in the description? Also, it may be helpful to outline what extended images are

ci/extended-docker/README.md
sam-heilbron, line 6: Why do we need to specify VERSION and TAGGED_VERSION?

sam-heilbron: I like this addition of the readme! Do we want developers to run this script or use the make target? Also, what situation would someone need to execute this?

Bslabe123: Jake might be a better person to contact as per https://github.com/solo-io/gloo/pull/7318/#issuecomment-1289475499 but im still interested in the answer as wel

Bslabe123: If we dont specify these two then CREATE_ASSETS will be set to false and gate the whole make target

jackstine: To answer Sam's question. We do not have to set `VERSION` and `TAGGED_VERSION` at the same time. 

The command I wrote to test this locally using `TAGGED_VERSION` because I want to push the images.
I set `VERSION` because I want to set the value, instead of having it overwritten.

`VERSION` is optional, but if not set should equal `TAGGED_VERSION` - first character.

https://github.com/solo-io/gloo/blob/validate-rate-limit-config/Makefile#L67

jackstine: to effectively test the file, you have to push the images.

Makefile
```diff
@@ -114,12 +114,22 @@ LDFLAGS := "-X github.com/solo-io/gloo/pkg/version.Version=$(VERSION)"
...
+# if `GO_ARCH` is set, then it will keep its value. Else, it will be changed based off the machine's host architecture.
+# if the machines architecture is set to arm64 then we want to set the appropriate values, else we only support amd64
+IS_ARM_MACHINE := $(or	$(filter $(UNAME_M), arm64), $(filter $(UNAME_M), aarch64))
+ifneq ($(IS_ARM_MACHINE), )
+	PLATFORM := --platform linux/amd64
+	ifeq ($(GOARCH), amd64)
+		GOARCH := amd64
+	else
+		GOARCH := arm64
+		PLATFORM := --platform linux/arm64
```
jackstine, line 126: nit: set Platform here twice to the same value

jackstine (approved):

Bslabe123:
/kick [logs](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=32f225a3-5e01-4bc8-8a8d-f2ebe477252e)

jackstine (dismissed):

Makefile
```diff
@@ -143,6 +151,22 @@ TEST_ASSET_DIR := $(ROOTDIR)/_test
 # in the tree rooted at that directory that match the given criteria.
 get_sources = $(shell find $(1) -name "*.go" | grep -v test | grep -v generated.go | grep -v mock_)
 
+# If both GCLOUD_PROJECT_ID and BUILD_ID are set, define functions that take a docker image name
```
sam-heilbron, line 154: I see this is needed by the race image, and exists in the enterprise repo. I don't fully follow what this is responsible for doing. Could you help me understand?

ci/extended-docker/README.md
```diff
@@ -0,0 +1,8 @@
...
```
sam-heilbron, line 6: Why are both version and tagged version necessary? It may be worth reading https://github.com/solo-io/gloo/blob/master/Makefile#L57 to see how they relate to eachother

Makefile
Bslabe123: Were pushing this image to GCR unlike all the other images in this makefile, this $(call .....) just allows us to tag the image as (i.e. `gcr.io/<FIRST_PARAMETER_HERE>:test-<BUILD_ID`>.

ci/extended-docker/README.md
Bslabe123: VERSION isn't necessary but TAGGED_VERSION is, we need TAGGED_VERSION set or else RELEASE gets set to false and gates the entire make target, however if VERSION is already set then this takes precedence.

Makefile
sam-heilbron: We push all our images to quay and grc, does this image get pushed to both? Why do we publish with a tag that is prefixed by test, isn't this just an extended gloo image that can be used in our tests?

Bslabe123: It does get retagged and pushed to quay if `docker-push` gets called, but when created it gets named gcr.io/gloo:test-<BUILD_ID>. I took a second look and I dont see where we push to GCR though

ci/extended-docker/README.md
sam-heilbron: I ran this script locally:
```
docker run -d \
  -e REGISTRY_HTTP_ADDR=0.0.0.0:5001 \
  -p 5001:5001 \
  --name registry-test \
  registry:2
VERSION=sh TAGGED_VERSION=sh IMAGE_REPO=localhost:5001 ci/extended-docker/extended-docker.sh
```

and the output was:
```
... lots of build logs
...
..
docker create -ti --name gloo-race-temp-container localhost:5001/gloo-race-build-container:0.0.1-fork bash
Unable to find image 'localhost:5001/gloo-race-build-container:0.0.1-fork' locally
Error response from daemon: manifest for localhost:5001/gloo-race-build-container:0.0.1-fork not found: manifest unknown: manifest unknown
make: *** [/Users/samheilbron/go/src/github.com/solo-io/gloo/_output/gloo-race/gloo-linux-amd64] Error 1
>> Temporary output file /var/folders/xb/94sy9wvs26n7km68nb3y29b40000gn/T/tmp.2UA1B3Jn
>> Removing /var/folders/xb/94sy9wvs26n7km68nb3y29b40000gn/T/tmp.2UA1B3Jn
```

Right now the images are being built with a different tag than supplied which seems to be causing issues.

I think it would be valuable since we reference a localhost:5000 (local registry) that we include in the steps how a developer would start this up

jbohanon: I believe the `docker run` command needs to port `5001:5000` as it listens on 5000 in the container by default ([ref](https://docs.docker.com/registry/deploying/#customize-the-published-port))

That said, I agree a tutorial or script would help tremendously. A modified version of [this script](https://github.com/solo-io/gloo/blob/master/ci/deploy-to-kind-cluster.sh) is what I use, with the command
```bash
kind delete cluster --name kind || true; \
~/Downloads/kind-with-registry.sh
```

Makefile
```diff
@@ -422,40 +450,40 @@ $(GLOO_RACE_OUT_DIR)/Dockerfile.build: $(GLOO_DIR)/Dockerfile
...
-gloo-race-docker: $(GLOO_RACE_OUT_DIR)/.gloo-race-docker ## gloo-race-docker
-$(GLOO_RACE_OUT_DIR)/.gloo-race-docker: $(GLOO_RACE_OUT_DIR)/gloo-linux-amd64 $(GLOO_RACE_OUT_DIR)/Dockerfile
-	docker build $(call get_test_tag_option,gloo) $(GLOO_RACE_OUT_DIR) \
-		--build-arg ENVOY_IMAGE=$(ENVOY_GLOO_IMAGE) --build-arg GOARCH=amd64 $(PLATFORM) \
+gloo-race-docker: $(GLOO_RACE_OUT_DIR)/gloo-linux-amd64 $(GLOO_RACE_OUT_DIR)/Dockerfile ## gloo-race-docker
+	docker buildx build --load $(PLATFORM) $(GLOO_RACE_OUT_DIR) -f $(GLOO_RACE_OUT_DIR)/Dockerfile.build \
+		--build-arg ENVOY_IMAGE=$(ENVOY_GLOO_IMAGE) \
+		--build-arg GOARCH=amd64 $(PLATFORM) \
+		--build-arg GO_BUILD_IMAGE=$(GOLANG_VERSION) \
+		--build-arg VERSION=$(VERSION)
```
jackstine, line 486: Missing `PLATFORM`

Makefile
```diff
@@ -422,40 +450,40 @@ $(GLOO_RACE_OUT_DIR)/Dockerfile.build: $(GLOO_DIR)/Dockerfile
...
 		-t $(IMAGE_REPO)/gloo:$(VERSION)-race $(QUAY_EXPIRATION_LABEL)
```
jackstine, line 487: missing `\` need to include the tag

projects/gloo/Dockerfile
```diff
@@ -26,7 +26,7 @@ ADD . /go/src/github.com/solo-io/gloo
 WORKDIR /go/src/github.com/solo-io/gloo
 
 # Build gloo with CGO and -race enabled
-RUN CGO_ENABLED=1 GOARCH=${GOARCH} GOOS=linux go build \
+RUN GO111MODULE=off CGO_ENABLED=1 GOARCH=${GOARCH} GOOS=linux go build \
```
jackstine: why was this added?

Bslabe123:
Split into two PRs: [buildx ](https://github.com/solo-io/gloo/pull/7394) and [build-test-assets](https://github.com/solo-io/gloo/pull/7395)

PR #7322, opened 2022-10-13
fix(Upstream): allow use_downstream_protocol when useHttp2 is set

gunnar-solo:
# Description
When using `USE_DOWNSTREAM_PROTOCOL`, we found that we _weren't_ respecting a nested envoy field as `USE_DOWNSTREAM`.  Rather, it was previously hardcoded to always `Explicit`ly use http2.  Below are reproduction/verification steps

## Environment setup steps, pre-fix
[[referenced files](https://drive.google.com/drive/folders/1T-aK7woIe_fwY5JYMIpBP7OBQUB9X4Mr?usp=sharing)]
```bash
kind delete cluster && kind create cluster

helm install -n gloo-system gloo gloo/gloo --create-namespace --version 1.13.0-beta24

kubectl apply -f ./upstream.yaml
kubectl apply -f ./virtual-service.yaml
kubectl -n gloo-system delete gateway gateway-proxy
kubectl create secret tls static-httpbin --key ./key.pem --cert ./cert.pem --namespace gloo-system
```

## Environment setup steps, post-fix
[[referenced files](https://drive.google.com/drive/folders/1T-aK7woIe_fwY5JYMIpBP7OBQUB9X4Mr?usp=sharing)]
```bash
kind delete cluster && kind create cluster

make install-go-tools push-kind-images package-chart
helm install -n gloo-system gloo _output/helm/charts/gloo*.tgz --create-namespace

kubectl apply -f ./upstream.yaml
kubectl apply -f ./virtual-service.yaml
kubectl -n gloo-system delete gateway gateway-proxy
kubectl create secret tls static-httpbin --key ./key.pem --cert ./cert.pem --namespace gloo-system
```

## test/verification commands
```bash
# 3 different processes
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443
kubectl port-forward -n gloo-system deploy/gateway-proxy 19000
watch 'curl -s http://localhost:19000/stats | grep "cluster.static-httpbin_gloo-system.upstream_cx_http"'

# curl with http1 || http2 and observe the "watch" command increment
curl https://127.0.0.1:8443/headers -vk --http1.1
curl https://127.0.0.1:8443/headers -vk --http2
```

https://user-images.githubusercontent.com/92050522/197213895-c62bb4d8-7b43-474f-b466-9de0dee13ef7.mov


projects/gloo/pkg/plugins/protocoloptions/plugin.go
```diff
@@ -83,20 +84,40 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+
+		http1ProtocolOptions := &envoy_config_core_v3.Http1ProtocolOptions{}
+		if in.GetConnectionConfig().GetHttp1ProtocolOptions() != nil {
+			http1ProtocolOptions, err = upstreamconn.ConvertHttp1ProtocolOptions(*in.GetConnectionConfig().GetHttp1ProtocolOptions())
+			if err != nil {
+				return err
+			}
+		}
+
+		protobuf.UpstreamProtocolOptions = &envoy_extensions_upstreams_http_v3.HttpProtocolOptions_UseDownstreamProtocolConfig{
```
bewebi, line 114: Would it make sense to refactor so that we only set `protobuf.UpstreamProtocolOptions` once (depending on the `ProtocolSelection`) rather than set it and then overwrite if it's `USE_DOWNSTREAM_PROTOCOL`?

gunnar-solo:
## Environment setup steps, pre-fix
```bash
kind delete cluster && kind create cluster
helm install -n gloo-system gloo gloo/gloo --create-namespace --version 1.13.0-beta24

# see attached for files
kubectl apply -f ./upstream.yaml
kubectl apply -f ./virtual-service.yaml
kubectl -n gloo-system delete gateway gateway-proxy
kubectl create secret tls static-httpbin --key ./key.pem --cert ./cert.pem --namespace gloo-system
```

## Environment setup steps, post-fix
```bash
kind delete cluster && kind create cluster

cd ..
make install-go-tools push-kind-images package-chart
helm install -n gloo-system gloo _output/helm/charts/gloo*.tgz --create-namespace
cd -

kubectl apply -f ./upstream.yaml
kubectl apply -f ./virtual-service.yaml
kubectl -n gloo-system delete gateway gateway-proxy
kubectl create secret tls static-httpbin --key ./key.pem --cert ./cert.pem --namespace gloo-system
```

## test/verification commands
```bash
# 3 different processes
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443
kubectl port-forward -n gloo-system deploy/gateway-proxy 19000
watch 'curl -s http://localhost:19000/stats | grep "cluster.static-httpbin_gloo-system.upstream_cx_http"'

# curl with http1 || http2 and observe the "watch" command increment
curl https://127.0.0.1:8443/headers -vk --http1.1
curl https://127.0.0.1:8443/headers -vk --http2
```

https://user-images.githubusercontent.com/92050522/197213895-c62bb4d8-7b43-474f-b466-9de0dee13ef7.mov

projects/gloo/pkg/plugins/protocoloptions/plugin.go
gunnar-solo: dealer's choice?  I prefer _not_ nesting, by default.  Ex:

### my preferred
```python
a = 'foo'
if (cond) {
  a = 'bar'
}
```

### comment request
```python
if (cond) {
  a = 'bar'
} else {
  a = 'foo'
}
```

🤷 I'll change it if _really_ wanted

ianmacclancy: I agree with @bewebi On this. 

If options not nil - get them 
else - create an empty options struct 

It slightly memory optimized and reads cleaner

bewebi: There's definitely some unnecessary processing, but for me it's mainly about how it reads
IMO the processing is negligible enough to prioritize legibility

An alternative might be to add in-line comments to describe what's happening along the lines of 
```
// Usually UpstreamProtocolOptions should be...
< a = 'foo' >
// Except when USE_DOWNSTREAM_PROTOCOL
< a = 'bar' >
```
In other words clarify that the initial setting may not actually be necessary, because as I read it I would assume everything that's being done is necessary and then I wonder why we set the field only to overwrite it

bewebi: An alternative that may may it clearer which case is more common might be a `switch` on `in.GetProtocolSelection()` with the original behavior as `default`

projects/gloo/pkg/plugins/protocoloptions/plugin.go
```diff
@@ -83,19 +84,39 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+		OverrideStreamErrorOnInvalidHttpMessage: ose,
+	}
+
+	protobuf := &envoy_extensions_upstreams_http_v3.HttpProtocolOptions{}
+	if in.GetProtocolSelection() == v1.Upstream_USE_DOWNSTREAM_PROTOCOL {
+		var err error
+
+		http1ProtocolOptions := &envoy_config_core_v3.Http1ProtocolOptions{}
+		if in.GetConnectionConfig().GetHttp1ProtocolOptions() != nil {
+			http1ProtocolOptions, err = upstreamconn.ConvertHttp1ProtocolOptions(*in.GetConnectionConfig().GetHttp1ProtocolOptions())
```
nfuden, line 100: Should we be relying on one plugin from another? This seems like a bad dependency chain waiting to happen.

projects/gloo/pkg/plugins/upstreamconn/plugin.go
```diff
@@ -128,7 +128,7 @@ func convertHttpProtocolOptions(hpo protocol.HttpProtocolOptions) (*envoy_config
 	return out, nil
 }
 
-func convertHttp1ProtocolOptions(hpo protocol.Http1ProtocolOptions) (*envoy_config_core_v3.Http1ProtocolOptions, error) {
+func ConvertHttp1ProtocolOptions(hpo protocol.Http1ProtocolOptions) (*envoy_config_core_v3.Http1ProtocolOptions, error) {
```
nfuden, line 131: If we decide that this should be exported (see question on inter plugin dependency above) can we have a comment in proper go format?

projects/gloo/pkg/plugins/protocoloptions/plugin.go
gunnar-solo: good on calling this out--it's now been lifted into a `utils` folder

gunnar-solo:
/kick [[Fail] gzip filter defined [It] should return compressed json](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=46e5402e-d618-4137-9818-4e1969411f8f)

gunnar-solo:
/kick [Zipkin config loading dynamic tracing with static collector cluster [It] should send trace msgs with valid zipkin provider](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=99ee9094-8920-44b6-8c13-3e79036ddb13)

gunnar-solo:
/kick [identical zipkin failure](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=b7c025c2-c294-4763-b480-05b79becf765)

bewebi (approved):

projects/gloo/pkg/plugins/utils/httpprotocolhelpers/http1_utils.go
```diff
@@ -0,0 +1,46 @@
...
+
+	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_extensions_http_header_formatters_preserve_case_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/http/header_formatters/preserve_case/v3"
+)
+
+const (
+	PreserveCasePlugin = "envoy.http.stateful_header_formatters.preserve_case"
+)
+
+func ConvertHttp1ProtocolOptions(hpo protocol.Http1ProtocolOptions) (*envoy_config_core_v3.Http1ProtocolOptions, error) {
```
nfuden, line 15: nit: exported function comment?

nfuden (approved):

bewebi (approved):

PR #7338, opened 2022-10-18
add lottery to review selection

ianmacclancy:
# Description

This PR is not ready for review and is posted so that I can view the results of github workflows. 

# Context 
During a team sync it came up that we were looking for a new way to find reviewers and get people familiar / looking at new areas of the codebase

Concept: Add a review lottery to our PR process that randomly chooses a team member to take a look at the code

It is not a replacement for SMEs or Code Owner reviews.


kdorosh (commented):
this is fun i like this

.github/workflows/reviewer-lottery.yaml
```diff
@@ -0,0 +1,13 @@
+name: "Reviewer lottery"
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, reopened]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
kdorosh, line 10: why v1 instead of v2?

ianmacclancy: Copy and paste from their recommended yaml. Just playing around with it for now

nrjpoddar:
Is this the right approach to get good PR reviews from experts? If we are doing 2 reviewers they need to be SMEs or folks who bring in different usual perspective. I feel that code owners is the right approach rather than random but I'm happy to learn why this is a good starting point.

kdorosh:
@nrjpoddar I think the idea was to have two reviewers: one SME (to offer different perspective) and one random (an opportunity to onboard / increase code ownership)

Would love to hear more perspectives here, may be worth taking ton an internal :thread:

sam-heilbron:
potential concern. The idea was that this would provide an random aspect for finding a second reviewer and the 

> @nrjpoddar I think the idea was to have two reviewers: one SME (to offer different perspective) and one random (an opportunity to onboard / increase code ownership)
> 
> Would love to hear more perspectives here, may be worth taking ton an internal 🧵

Yeah agreed with Kevin. The team raised a similar concern to what @nrjpoddar says. We hoped however that this would help distribute the load of reviews, spread knowledge across different features within the team, and still require the author to identify 1 or 2 SMEs in addition to review

kdorosh (approved):

Issue #7349, opened 2022-10-20
Better default/configurability for TCP Keepalive time


ashishb-solo:
### Version

1.12.x (beta)

### Is your feature request related to a problem? Please describe.
Envoy connection to gloo XDS server currently uses default tcp keepalive settings.
This value for TCP keepalive is too long (>2 hours) and cannot be changed at installation time.

### Describe the solution you'd like

1. Provide a better default value for TCP keepalive. Current suggestion is 10 minutes
2. Provide a way to configure this in our helm templates

### Describe alternatives you've considered

Current workaround is to modify our envoy bootstrap configuration and restart all gateway-proxy instances.
This can be done by editing the configmap.

ashishb-solo:
Envoy documentation on TCP keepalive [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-tcpkeepalive)

Notable quote:

> Default is to use the OS level configuration (unless overridden, Linux defaults to 7200s (i.e., 2 hours.)

PR #7358, opened 2022-10-21
Add TcpKeepaliveTimeSeconds to helm templates

ashishb-solo:
# Description

Allow users to configure TCP keepalive time and set the default value to 1 minute.

Envoy exposes this option [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-tcpkeepalive). This commit is an effort to expose this configuration to users.

# Context

When services fail, no Gloo updates are possible until the TCP connection is removed, so we need to reduce the default and make it configurable to users. Additionally, users would like to be able to change this value at the time of installation via helm, so we choose to modify the bootstrap Envoy configuration in our helm templates.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7349


install/helm/gloo/templates/9-gateway-proxy-configmap.yaml
```diff
@@ -170,7 +170,8 @@ data:
                     port_value: {{ empty $spec.xdsServicePort | ternary .Values.gloo.deployment.xdsPort $spec.xdsServicePort }}
         http2_protocol_options: {}
         upstream_connection_options:
-          tcp_keepalive: {}
+          tcp_keepalive:
+            keepalive_time: {{ $spec.tcpKeepaliveTimeSeconds | default 60}}
```
jenshu, line 174: i think we can define this default in values-template.yaml - but we should also make sure that custom gateway proxies inherit the default

it would be good to add a helm unit test that checks (for both the default gateway proxies and custom gateway proxies) that 1) we get the default value if nothing is passed in and 2) can override the default

install/helm/gloo/generate/values.go
```diff
@@ -416,6 +416,7 @@ type GatewayProxy struct {
 	LogLevel                       *string                      `json:"logLevel,omitempty" desc:"Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
 	XdsServiceAddress              *string                      `json:"xdsServiceAddress,omitempty" desc:"The k8s service name for the xds server. Defaults to gloo."`
 	XdsServicePort                 *uint32                      `json:"xdsServicePort,omitempty" desc:"The k8s service port for the xds server. Defaults to the value from .Values.gloo.deployment.xdsPort, but can be overridden to use, for example, xds-relay."`
+	TcpKeepaliveTimeSeconds        *uint32                      `json:"tcpKeepaliveTimeSeconds,omitempty" desc:"The amount of time in seconds for connections to be idle before sending keep-alive probes. See here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-tcpkeepalive"`
```
ben-taussig-solo, line 419: Nit: can we add "Defaults to 60 seconds" somewhere here?

ben-taussig-solo (commented):
Looks pretty good to me but could we add a test that confirms that the default applies to custom gateway proxies like Jenny said? Apologies if it's there and I missed it

install/test/helm_test.go
```diff
@@ -4658,6 +4658,28 @@ metadata:
 					"gateway-proxy-id": "gateway-proxy",
 				}
 
+				It("should allow customising tcp keepalive", func() {
+					prepareMakefile(namespace, helmValues{
+						valuesArgs: []string{"",
+							"gatewayProxies.gatewayProxy.tcpKeepaliveTimeSeconds=30"},
```
jenshu, line 4664: can we also add a test to make sure custom gateways can override the defaults, e.g. "gatewayProxies.anotherGatewayProxy.tcpKeepaliveTimeSeconds=..."

jenshu (approved):

nfuden (approved):

PR #7360, opened 2022-10-21
Deflake Tracing e2e test

sam-heilbron:
# Description

# Context

# Checklist:


sam-heilbron:
/skip-changelog there is no current github issue tracking this flake

nfuden (approved):
This seems reasonable to me. I am still not sure from the example logs that it will fully be fixed by this but it seems to at least make it much less flakey.

test/e2e/zipkin_test.go
```diff
@@ -114,304 +108,278 @@ var _ = Describe("Zipkin config loading", func() {
...
-
-			// write gateways and wait for them to be created
-			err := gloohelpers.WriteDefaultGateways(writeNamespace, testClients.GatewayClient)
-			Expect(err).NotTo(HaveOccurred(), "Should be able to write default gateways")
-			Eventually(func() (gatewayv1.GatewayList, error) {
-				return testClients.GatewayClient.List(writeNamespace, clients.ListOpts{})
-			}, "10s", "0.1s").Should(HaveLen(2), "Gateways should be present")
-
-			// run envoy
-			err = envoyInstance.RunWithConfigFile(testClients.GlooPort, "./envoyconfigs/zipkin-static-cluster.yaml")
```
gunnar-solo, line 340: what was the rationale behind moving from a config yaml to in-momory declaration?

test/e2e/gzip_test.go
```diff
@@ -55,152 +57,118 @@ var _ = Describe("gzip", func() {
...
-		}, "5s", "0.1s").ShouldNot(BeNil())
-	}
-
-	checkVirtualService := func(testVs *gatewayv1.VirtualService) {
-		Eventually(func() (*gatewayv1.VirtualService, error) {
-			return testClients.VirtualServiceClient.Read(testVs.Metadata.GetNamespace(), testVs.Metadata.GetName(), clients.ReadOpts{})
-		}, "5s", "0.1s").ShouldNot(BeNil())
-	}
-
-	testRequest := func(jsonStr string) string {
```
gunnar-solo, line 97: it seems to me that the bulk of the fix comes from wrapping this in an `Eventually`.  I see that we're also pulling in a bunch of "previously discussed best practices" as well

(just an observation)

test/e2e/gzip_test.go
```diff
@@ -55,152 +57,118 @@ var _ = Describe("gzip", func() {
...
 
-		// write a test upstream
 		// this is the upstream that will handle requests
-		testUs := v1helpers.NewTestHttpUpstream(ctx, envoyInstance.LocalAddr())
-		up = testUs.Upstream
-		_, err = testClients.UpstreamClient.Write(up, clients.WriteOpts{OverwriteExisting: true})
-		Expect(err).NotTo(HaveOccurred())
+		testUpstream = v1helpers.NewTestHttpUpstream(ctx, envoyInstance.LocalAddr())
+
+		vsToTestUpstream := gloohelpers.NewVirtualServiceBuilder().
```
gunnar-solo, line 70: I appreciate this helper method builder its miles

bewebi (approved):
Looks good

Github really made a mess of the diff, so it was hard to follow
Part of this is seemingly due to some refactoring/cleanup that wasn't explicitly called out in the PR description
It would be good to mention _all_ changes made to code (ie including refactoring, renaming, etc.) in the description in order to make it easier for reviewers to parse the diff

test/e2e/zipkin_test.go
sam-heilbron: There we 2 tests that did the same thing so I just removed one. One test wrote a cluster to the bootstrap config of envoy (to validate that we can reference the cluster by name), the other write an upstream and then referenced that upstream by cluster name.

I prefer the second type of test, since it validate the gloo pipeline (translation) and is more similar to what a customer would use

test/e2e/gzip_test.go
sam-heilbron: Yes that's fair. The eventually is the retry logic that we may need, but also the way that configuration is applied caused the flake. Before we would write a bunch of resources and then modify one. Since we previously had valid resources (gw, vs, upstream) requests would still go through using the old config. The new format is to just write resources once, and only with the config we need.

So yeah, theoretically we could have just used the eventually, but I figured moving more of our tests towards the snapshot model would be easier to reason through

test/e2e/gzip_test.go
sam-heilbron: ha me too! I foud that we have so many combinations of VS config, that we tended to have a telescoping constructor anti-pattern:

https://knowledgebasement.com/avoid-the-telescoping-anti-pattern-in-java/

sam-heilbron:
> Looks good
> 
> Github really made a mess of the diff, so it was hard to follow Part of this is seemingly due to some refactoring/cleanup that wasn't explicitly called out in the PR description It would be good to mention _all_ changes made to code (ie including refactoring, renaming, etc.) in the description in order to make it easier for reviewers to parse the diff

Yeah that's fair! The refactoring isn't new per se, in that its a pattern used in other tests that I was trying to bring to these. But good callout that I can make that more explicit in the PR context

PR #7374, opened 2022-10-24
Race image arm ci support

jackstine:
# Description

Can't build the race image. Also building race image on `arm64` machines takes 10 minutes to build on M1 machines, so opt out to build it by default on M1 machines.

# Context

Don't need to build the race image every time I need to build Gloo. Only needs to be built during CI builds.

# Checklist:


Makefile
```diff
@@ -427,14 +428,15 @@ $(GLOO_RACE_OUT_DIR)/.gloo-race-docker-build: $(GLOO_SOURCES) $(GLOO_RACE_OUT_DI
...
+		--build-arg GOARCH=amd64 \
+		$(PLATFORM) \
 		.
 	touch $@
 # Build inside container as we need to target linux and must compile with CGO_ENABLED=1
 # We may be running Docker in a VM (eg, minikube) so be careful about how we copy files out of the containers
 $(GLOO_RACE_OUT_DIR)/gloo-linux-$(GOARCH): $(GLOO_RACE_OUT_DIR)/.gloo-race-docker-build
 	docker create -ti --name gloo-race-temp-container $(IMAGE_REPO)/gloo-race-build-container:$(VERSION) bash
-	docker cp gloo-race-temp-container:/gloo-linux-$(GOARCH) $(GLOO_RACE_OUT_DIR)/gloo-linux-$(GOARCH)
+	docker cp gloo-race-temp-container:/gloo-linux-amd64 $(GLOO_RACE_OUT_DIR)/gloo-linux-amd64
```
ianmacclancy, line 439: Can you add a comment as to why the arch is hardcoded here and not in other locations. I don't want us to get bit by this again.

Makefile
```diff
@@ -706,10 +708,20 @@ ifeq ($(RELEASE), "true")
...
-docker: discovery-docker gloo-docker gloo-race-docker \
+docker: docker-local docker-ci
+
+.PHONY: docker-local
+docker-local: discovery-docker gloo-docker  \
 		gloo-envoy-wrapper-docker certgen-docker sds-docker \
 		ingress-docker access-logger-docker kubectl-docker
 
+.PHONY: docker-ci
+ifeq ($(UNAME_M), arm64)
```
ianmacclancy, line 719: Why is this separated into its own make command?

ianmacclancy: That makes a lot of sense to me - this looks like what it does is not build the race image on arm architecture. Which means it will still build locally for a lot of devs. So its an arm vs non-arm rather than a dev vs ci check.

jackstine:
Support for CI will be included in https://github.com/solo-io/gloo/issues/7375

Makefile
```diff
@@ -419,6 +420,7 @@ $(GLOO_RACE_OUT_DIR)/Dockerfile.build: $(GLOO_DIR)/Dockerfile
 	mkdir -p $(GLOO_RACE_OUT_DIR)
 	cp $< $@
 
+# image is built with amd64 architecture only because it wil build on arm
```
elcasteel, line 423: this comment is a little confusing, how about:

Hardcode GOARCH for targets that are both built and run entirely in amd64 docker containers

jackstine: eh, my bad.

Makefile
elcasteel: Do we have an environment variable we can check to see if we are running in CI? Then we can check that as part of `docker push` to decide whether to add the -race image instead of checking the architecture. 
With this set up, people on Intel Macs can avoid building the -race image by calling `make docker-local` but the naming and logic is kind of confusing.

ianmacclancy: Ethan and I are saying similar things. 

Its confusing to call this target "ci" as its just checking architecture 
For now can we change the name to docker-arch-check or something along those lines?

We can refactor it to a better name when we get checks on if we are running in CI, or as suggested in standup, when we split out CI commands into separate makefiles

jackstine: this might be a larger effort than just fixing the immediate problem.

elcasteel: I'd be good with keeping the code as-is and changing the names of targets to reflect that the check is based on architecture

ianmacclancy: Yup, only change for now would be a name that indicates its an architecture check

ianmacclancy (approved):

elcasteel (approved):

PR #7376, opened 2022-10-25
v1.11.x: bump golang.org/x/text to v0.4.0

ben-taussig-solo:
# Description

 - Bumps `golang.org/x/text` to `v0.4.0` to resolve [CVE-2022-32149](https://github.com/advisories/GHSA-69ch-w2m2-3vjp)


jenshu (approved):

go.mod
```diff
@@ -81,10 +81,11 @@ require (
 	go.opencensus.io v0.23.0
 	go.uber.org/multierr v1.6.0
 	go.uber.org/zap v1.19.1
-	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3
+	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4
 	golang.org/x/net v0.0.0-20220906165146-f3363e06e74c // indirect
-	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	golang.org/x/tools v0.1.10
+	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4
+	golang.org/x/text v0.4.0 // indirect
```
inFocus7, line 87: Since `x/text` wasn't originally part of the go.mod list, but needed the bump, did you just do a `go get golang.org/x/text@v0.4.0`?

Asking, since in [mine](https://github.com/solo-io/dev-portal/pull/2448/files), I pinned it. Although if it's preferred to `go get` it I'll update mine 👀

inFocus7 (commented):
Everything looks good, left a question and I'll approve right after! 👍🏼

go.mod
ben-taussig-solo: Yeah, I did exactly that! `go get golang.org/x/text@v0.4.0`

I don't think it should matter either way -- I slightly prefer the version in this PR because I think it's clearer that golang.org/x/text is an indirect dependency, but if we've pinned indirect deps to resolve CVEs in the past then I think we should do that for the sake of consistency. Looking into some previous CVE PRs now and I'll report back with an update

inFocus7: Ah I haven't checked previous PRs actually, so your way could be the preferred!

ben-taussig-solo: Struggling to find one that bumps an indirect dep.....

ben-taussig-solo: Seems like the approach in this PR has been used in the past: https://github.com/solo-io/gloo/pull/7187/files

inFocus7: It looks like that one just updates the existing `x/net`, although I think your approach is better. It has the `// indirect` tag and is clearer than pinning it!

inFocus7 (approved):
🚀

PR #7385, opened 2022-10-27
Add Extensions to Gloo Validation: Rate Limit Configs

jackstine:
# Description

Add Extensions to Gloo Validation.
Rate Limit Configs are added to the webhook.

`TranslatorSyncerExtensionParams` has been changed so that the `RateLimitServiceSettings` is now a pointer to the `ServiceSettings`.

Bulk validation of resources may be slower or less performant, with the addition and value gained from validating Rate Limit Configs. Please use the list GVK to bulk validate resources in a single API call.

The Validator interface has changed, we have deleted the `ModificationIsSupported`, and `DeletionIsSupported` methods, as they are no longer needed.  All supported resources are now mentioned in the `gateway-validation-webhook-configuration.yaml`.

The `IgnoreProxyValidationFailure` flag was intended to ignore network issues between the Gateway server and the Gloo server. Since the Gateway server and the Gloo server have merged, this is no longer an issue that can occur, and the field is now deprecated.

We have also made suggestions about adding in metrics for validation, such as time spent validating, or time spent locked. These are referenced in the Issue [tagged on this PR](https://github.com/solo-io/gloo/issues/6777#issuecomment-1263872978).

# Context

Adding Extensions to Gloo Validation. This will allow for validation of Rate Limit Configs.

# Checklist:


projects/gateway/pkg/validation/validator.go
```diff
@@ -427,6 +427,9 @@ func (v *validator) ValidateModifiedGvk(ctx context.Context, gvk schema.GroupVer
 
 func (v *validator) validateModifiedResource(ctx context.Context, gvk schema.GroupVersionKind, resource resources.Resource, dryRun, acquireLock bool) (*Reports, error) {
 	var reports *Reports
+	// TODO-JAKE
+	// so it looks like we might need to extend the helm chart in enterprise to add the rate limit resources
```
sam-heilbron, line 431: I'm a little confused by this comment: it refers to helm charts and rate limit resources. Perhaps it will be clearer in the future when you embark on it, but what is the change that will be necessary and why?

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
```diff
@@ -79,6 +79,7 @@ func (r *RateLimitConfig) GetNamespacedStatuses() *core.NamespacedStatuses {
 }
 
 func (r *RateLimitConfig) SetNamespacedStatuses(status *core.NamespacedStatuses) {
+	contextutils.LoggerFrom(context.Background()).DPanic(status.Statuses)
```
sam-heilbron, line 82: We can sync on this, but this line should never be executed. have you noticed that it is?

projects/gloo/pkg/syncer/extauth/extauth_translator_syncer.go
```diff
@@ -41,6 +42,19 @@ func (s *translatorSyncerExtension) ID() string {
 	return ServerRole
 }
 
+// TODO-JAKE later
+func (s *translatorSyncerExtension) Translate(
```
sam-heilbron, line 46: See the earlier proposal to handle this: Instead of implementing a new method, just re-using Sync but passing parameters that do nothing to the snapshot cache

projects/gloo/pkg/syncer/ratelimit/ratelimit_translator_syncer.go
```diff
@@ -39,6 +41,23 @@ func (s *translatorSyncerExtension) ID() string {
 	return ServerRole
 }
 
+func (s *translatorSyncerExtension) Translate(
```
sam-heilbron, line 44: Not required for this PR:

https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/syncer/ratelimit/ratelimit_translator_syncer.go#L28

I think we could remove this open source syncer extension entirely. Its purpose is to warn users about using enterprise features that they don't have access to. We introduced the [enterprise warning plugin](https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/plugins/enterprise_warning/plugin.go) to solve this problem. So you could add a rate limit check to that plugin and delete all of this code if you wanted.

sam-heilbron (commented):
Looking good! Left some initial thoughts. Happy to discuss them in person too if you'd like.

Other thought: The webhook API has some fields that we may want to expose to users (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#validatingwebhook-v1-admissionregistration-k8s-io) to make it easier to opt into (or out of) certain fields. We don't need to do it as part of this PR, but I think its important to outline the impact o this new step that occurs during validation, to potentially give users a way to opt out if at first it's not performant enough.

projects/gloo/pkg/syncer/syncer_extensions.go
```diff
@@ -30,6 +30,10 @@ type TranslatorSyncerExtension interface {
 		settings *v1.Settings,
 		snapshotSetter SnapshotSetter,
 		reports reporter.ResourceReports)
+
+	// Translate will process the ApiSnapshot and updates the reports with Errors/Warnings that it encouters
+	// unlike Sync() this will not update the SnapshotCache.
+	Translate(ctx context.Context, snap *v1snap.ApiSnapshot, proxies v1.ProxyList, reports reporter.ResourceReports) error
```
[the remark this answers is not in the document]
jackstine: No I had not consider using a no-op for the snapshot setter.  A concern though was that the list of proxies came from the snapshot itself.  Since we validate one proxy at a time, I wanted a way to validate a separate proxy individually. This could be done by switching the list of proxies in the snapshot.

projects/gateway/pkg/validation/validator.go
jackstine: this comment can be ignored now, because I was able to just add the resource GVK to the helm charts in Gloo.

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
jackstine: Yeah this line was/is being called. I have not checked it recently, but it was showing in the logs earlier.

projects/gloo/pkg/validation/gloo_validator.go
```diff
@@ -5,43 +5,56 @@ import (
...
+	ratelimit "github.com/solo-io/gloo/projects/gloo/pkg/api/external/solo/ratelimit"
 	gloo_translator "github.com/solo-io/gloo/projects/gloo/pkg/translator"
 )
 
 const GlooGroup = "gloo.solo.io"
 
 var GvkToSupportedDeleteGlooResources = map[schema.GroupVersionKind]bool{
-	gloov1.UpstreamGVK: true,
-	gloov1.SecretGVK:   true,
+	gloov1.UpstreamGVK:           true,
```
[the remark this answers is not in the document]
jackstine: yeah, I was thinking about this as well, we could see about adding the validation for the extensions to the Gateway validator. That way we could check all the proxies at once, after we have checked all the gateway and gloo resources. This would keep the gloo resources separated.

projects/gloo/pkg/validation/server.go
```diff
@@ -40,15 +49,23 @@ type validator struct {
...
+	}
 	return &validator{
-		translator:    translator,
-		glooValidator: NewGlooValidator(translator, xdsSanitizer),
+		translator:    config.Translator,
+		glooValidator: NewGlooValidator(gvc),
 		notifyResync:  make(map[*validation.NotifyOnResyncRequest]chan struct{}, 1),
-		ctx:           ctx,
-		xdsSanitizer:  xdsSanitizer,
+		ctx:           config.Ctx,
```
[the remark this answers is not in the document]
jackstine: Not sure how the Ctx is nil?
For your second point, we use ctx in a lot of structs in the setup_syncer. Maybe we could make a refactor.

projects/gateway/pkg/validation/validator.go
sam-heilbron: Which GVK did you adjust in the chart? Are you referring to the validation webhook config to accept rate limits?

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
sam-heilbron: "this line was/is being called'- is this true before your changes or just in this PR?

projects/gloo/pkg/validation/server.go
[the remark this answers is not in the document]
sam-heilbron: Oh my mistake, I was  confusing the validatorConfig and glooValidatorConfig, one of which has a ctx property and the other does not

projects/gateway/pkg/validation/validator.go
jackstine: yeah 
https://github.com/solo-io/gloo/blob/7f492fdbb3021bada8ed71552e11c0e489a8b61d/install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml#L39

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -36,6 +36,10 @@ webhooks:
     apiGroups: ["gloo.solo.io"]
     apiVersions: ["v1"]
     resources: ["secrets"]
+  - operations: [ "CREATE", "UPDATE", "DELETE" ]
```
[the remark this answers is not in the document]
jackstine: They should apply all their rate limit configs first, before applying vs and route tables that will reference the RLC.  Where is a good place to highlight this information?

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -27,7 +27,11 @@ webhooks:
   - operations: [ "CREATE", "UPDATE", "DELETE" ]
     apiGroups: ["gateway.solo.io"]
     apiVersions: ["v1"]
-    resources: ["*"]
+    resources: ["virtualservices", "routetables"]
+  - operations: [ "CREATE", "UPDATE"]
+    apiGroups: ["gateway.solo.io"]
+    apiVersions: ["v1"]
+    resources: ["gateways"]
```
kdorosh, line 34: why remove delete for gateways?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2086,14 +2101,65 @@ spec:
...
+						settings.GetGateway().GetValidation().EnableExtensionValidation = &wrappers.BoolValue{Value: false}
+					}, testHelper.InstallNamespace)
+				})
 
+				JustBeforeEach(func() {
+					verifyValidationWorks()
+				})
+
+				It("rejects bad resources", func() {
+					testCases := []testCase{{
```
kdorosh, line 2135: nit; perhaps we can add this as a test case to the existing validation tests? (basically where this was copied from)

jackstine:
I added in commits [2b5ae741012f32cdb4bb5e0c2f1b298b18c07433](https://github.com/solo-io/gloo/pull/7385/commits/2b5ae741012f32cdb4bb5e0c2f1b298b18c07433) to get rid of the support methods in the validation webhook.  This also updated the helm webhook configuration so that the webhook would only support the options that were initially intended. We can get rid of this commit, if it seems to cause a problem.

Changes to the webhook configuration can be added later in another PR to support all resources.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
jackstine: This was the initial support.  The gateway would not be supported for delete operations, only create/update.

https://github.com/solo-io/gloo/pull/7385#issuecomment-1307323152

jenshu: why did we remove validation of Gateway deletion? wouldn't that be a regression since we used to validate that, but now we don't?

jackstine: deletion of gateways were never supported. 
https://github.com/solo-io/gloo/blob/v1.12.x/projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go#L401

test/kube2e/gateway/gateway_test.go
jackstine: I separated it because we have to set `EnableExtensionValidation = true`

projects/gloo/pkg/validation/gloo_validator.go
```diff
@@ -79,12 +92,22 @@ func (gv glooValidator) Validate(ctx context.Context, proxy *gloov1.Proxy, snaps
...
 
+		// TODO-JAKE we might want to set the proxies of the snapshot here so that they align with the proxiesToValidate list above...
+		for _, ex := range gv.extensions {
+			err := ex.Translate(ctx, snapshot, v1.ProxyList{proxy}, resourceReports)
+			// TODO-JAKE not sure if we want to have an error here, because these errors are not really respective of the proxy resource
+			if err != nil {
+				resourceReports.AddError(proxy, err)
+			}
+		}
+
```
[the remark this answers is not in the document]
jackstine: I moved the code to its own validator. This way we validate all the proxies at once for the extensions.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
jenshu: ah gotcha, forgot about that!

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -27,7 +27,11 @@ webhooks:
   - operations: [ "CREATE", "UPDATE", "DELETE" ]
     apiGroups: ["gateway.solo.io"]
     apiVersions: ["v1"]
-    resources: ["*"]
```
sam-heilbron, line 30: In the [gateway API group](https://github.com/solo-io/gloo/tree/master/projects/gateway/api/v1), we have: gateway, rt, vs, and matchable_http_gateway. I'm not sure if we support validating that 4th option in the code, but I did want to call out that this change will mean those matchable http gateways will not be subject to the validation webhook at all. Is that intentional?

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -27,7 +27,11 @@ webhooks:
...
+    resources: ["virtualservices", "routetables"]
+  - operations: [ "CREATE", "UPDATE"]
```
sam-heilbron, line 31: Given the uncertainty around lack of delete support for gateways, do you think it's worth it to explicitly comment in the code outlining why its not supported?

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
```diff
@@ -79,6 +79,8 @@ func (r *RateLimitConfig) GetNamespacedStatuses() *core.NamespacedStatuses {
 }
 
 func (r *RateLimitConfig) SetNamespacedStatuses(status *core.NamespacedStatuses) {
+	// TODO-JAKE find out what is calling this, how this is being called.
```
sam-heilbron, line 82: Should we remove this? Have we identified why this code was being executed?

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -68,6 +68,7 @@ func measureResource(ctx context.Context, resource string, len int) {
 	}
 }
 
+// syncEnvoy this will translate, sanatize, and set the snapshot for each of the proxies, all while merging all the reports into allReports.
```
sam-heilbron, line 71: nit:
```suggestion
// syncEnvoy will translate, sanatize, and set the snapshot for each of the proxies, all while merging all the reports into allReports.
```

projects/gloo/pkg/syncer/validation/validator.go
```diff
@@ -0,0 +1,42 @@
...
+
+type validator struct {
+	extensions []syncer.TranslatorSyncerExtension
+	settings   *v1.Settings
+}
+
+// Validate will sync the extensions with the snapshot. This uses a no-op snapshot setter, so that no changes to the
+// snapshot occur.
+func (v *validator) Validate(ctx context.Context, snapshot *gloosnapshot.ApiSnapshot) reporter.ResourceReports {
+	noOp := &syncer.NoOpSnapshotSetter{}
```
sam-heilbron, line 34: nit: Since this is a no-op, we could instantiate it once on the struct, instead of creating it each time we call validate

test/kube2e/gateway/gateway_test.go
```diff
@@ -1252,60 +1292,71 @@ var _ = Describe("Kube2e: gateway", func() {
 		})
 	})
 
-	Context("tests with RateLimitConfigs", func() {
-
-		var rateLimitConfig *v1alpha1skv1.RateLimitConfig
+	Context("validation server is disabled", func() {
```
sam-heilbron, line 1295: How is the validation server disabled for these tests?

projects/gloo/pkg/syncer/syncer_extensions.go
```diff
@@ -34,7 +40,7 @@ type TranslatorSyncerExtension interface {
 
 type TranslatorSyncerExtensionParams struct {
 	Hasher                   func(resources []envoycache.Resource) (uint64, error)
-	RateLimitServiceSettings ratelimit.ServiceSettings
+	RateLimitServiceSettings *ratelimit.ServiceSettings
```
sam-heilbron, line 43: This is a change to an exported field of an exported struct, so I think it would be considered breaking. Since this code is used internally (ie with the enterprise repo), it's not an issue per se, but I think we should call this out loudly in the PR. Ideally we merge the enterprise code right after this is released, but we should do our best to notify other devs of this change since they may run into it

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -772,32 +788,27 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 		logger.Debugf("Gateway translation is disabled. Proxies are provided from another source")
 	}
 
-	gwValidationSyncer := gwvalidation.NewValidator(gwvalidation.NewValidatorConfig(
-		gatewayTranslator,
-		validator.ValidateGloo,
-		ignoreProxyValidationFailure,
-		allowWarnings,
-	))
+	syncerValidator := syncerValidation.NewValidator(syncerExtensions, opts.Settings)
```
sam-heilbron, line 791: It may be helpful to outline in a comment what the syncerValidator is, since the setup code is responsible for kicking off so many components. 

Right now we instantiate the validator with the same syncer extensions that are provided at runtime. For enterprise, doesn't this include extauth configs as well? So even though we may be validating a rate limit config, now we'll translate the extauth configs as well. Or am I missing something?

sam-heilbron, line 791: nit: I think syncerValidator could be made a bit more concrete. We have "syncers" in Gloo Edge, which are responsible for syncing the state of resources. The glooValidator for exampe, validates gloo resources, I think a more appropriate name might be syncerExtensionValidator, or extensionValidator since it's responsible for validating the syncer extensions

projects/gateway/pkg/validation/validator.go
```diff
@@ -585,10 +504,22 @@ func (v *validator) getReportsFromGlooValidation(reports []*gloovalidation.GlooV
...
 		}
 	}
-	return &Reports{
-		ProxyReports: &proxyReports,
-		Proxies:      proxies,
-	}, errs
+	return errs
+}
+
+func (v *validator) getErrorsFromResourceReports(reports reporter.ResourceReports) error {
```
sam-heilbron, line 510: Could we use the `validate` and `validateStrict` methods that are exposed on the ResourceReports https://github.com/solo-io/solo-kit/blob/master/pkg/api/v2/reporter/reporter.go#L196?

projects/gloo/pkg/syncer/validation/validator.go
```diff
@@ -0,0 +1,42 @@
...
+// NewValidator will create a new validator for the purpose of validating extensions.
+func NewValidator(extensions []syncer.TranslatorSyncerExtension, settings *v1.Settings) validator {
+	return validator{
+		extensions: extensions,
+		settings:   settings,
+	}
+}
+
+type Validator interface {
+	// Validate will validate the proxies of a snapshot, and return the resource reports corresponding to the
```
sam-heilbron, line 21: nit: It's not that we're validating the proxies of the snapshot, we're more validating certain resources (the extended resources) of the snapshot.

projects/gloo/pkg/syncer/validation/validator.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/gloosnapshot"
+	"github.com/solo-io/gloo/projects/gloo/pkg/syncer"
+	"github.com/solo-io/solo-kit/pkg/api/v2/reporter"
+)
+
+// NewValidator will create a new validator for the purpose of validating extensions.
+func NewValidator(extensions []syncer.TranslatorSyncerExtension, settings *v1.Settings) validator {
```
sam-heilbron, line 13: nit: return the reference?

projects/gloo/pkg/syncer/ratelimit/ratelimit_translator_syncer.go
jackstine: So it is still being called,  Looks like the [UnmarshalResource](https://github.com/solo-io/solo-kit/blob/master/pkg/utils/protoutils/marshal.go#L188) in solo-kit needs to conform to both CustomInputResources and InputResources.  this can be taken care of with an update.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
jackstine: Should be all good, Here is the [GVK](https://github.com/solo-io/gloo/blob/validate-rate-limit-config/projects/gateway/pkg/api/v1/matchable_http_gateway.sk.go#L175) of the MatchableHttpGateway, which was initially ignored.  See [this comment](https://github.com/solo-io/gloo/pull/7385#discussion_r1014172135) for more details.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
jackstine: that is fair for now.

projects/gloo/api/external/solo/ratelimit/rate_limit_config.go
jackstine: just made an update to the comment, we will have to make a change to solo-kit I believe.

sam-heilbron: Let's chat in our sync this morning. The rateLimitConfig is different than most other gloo resources, so I wouldn't expect solo-kit to be required to support this

projects/gloo/pkg/syncer/setup/setup_syncer.go
jackstine: yes that is correct.

projects/gloo/pkg/syncer/syncer_extensions.go
jackstine: We should make that a practice before we merge PRs.

elcasteel:
On the performance side, it's important to note that we only validate one resource at a time so validating more types of resources has the potential for a noticeable impact on the performance of applying a lot of config at once. I'm especially thinking of the particular use case where a customer has ratelimit and authconfig for each VS.

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1106,7 +1117,6 @@ func constructOpts(ctx context.Context, clientset *kubernetes.Interface, kubeCac
 			ValidatingWebhookPort:        gwdefaults.ValidationWebhookBindPort,
 			ValidatingWebhookCertPath:    validationCfg.GetValidationWebhookTlsCert(),
 			ValidatingWebhookKeyPath:     validationCfg.GetValidationWebhookTlsKey(),
-			IgnoreProxyValidationFailure: validationCfg.GetIgnoreGlooValidationFailure(),
```
elcasteel, line 1109: This was exposed as a setting before right? If so, we should call out in the changelog that it's no longer supported, why and what setting to use instead

jackstine: I can add it to the changelog as well, the docs are covered in these changes.  There is no replacement for this field.

jackstine:
Changes that I have made to silence the errors from Ext-Auth.
[Delete the Error for a missing Ext-Auth Config](https://github.com/solo-io/gloo/pull/7385/commits/fe1599609430c0ef02c2d3a588bfb9621ad31237)
Changes made to support changes in tests [one](https://github.com/solo-io/gloo/pull/7385/commits/fb9b69c725e1395332580a82573951c9876294e3) [two](https://github.com/solo-io/gloo/pull/7385/commits/3faf329c904a5891170133a3940232a7009470a4) [three](https://github.com/solo-io/gloo/pull/7385/commits/24cf81d239cc85bf254012c0d2c11bb5cf74b5c3) [four](https://github.com/solo-io/gloo/pull/7385/commits/53840e865f870d181076a8520948a782e2875194)

jackstine:
Here is the performance test of applying 800 resources.

Settings:
```
    validation:
      allowWarnings: true
      alwaysAccept: false
      disableTransformationValidation: false
      proxyValidationServerAddr: gloo:9988
      serverEnabled: true
      validationServerGrpcMaxSizeBytes: 104857600
      warnRouteShortCircuiting: false
```

The only errors from the validation were those related to the missing Secret. This is because the applier does not create the secret.

```
objects start:  2022-11-10T15:32:54-06:00
7 / 800 [->_____________________________________________________________________________________________________________________________________________________________] 0.88% ? p/s ?First error encountered at index: 0 admission webhook "gloo.gloo-system.svc" denied the request: resource incompatible with current Gloo snapshot: [Validating *v1.VirtualService failed: 1 error occurred:
	* Validating *v1.VirtualService failed: validating *v1.VirtualService name:"vs-0"  namespace:"gloo-system": failed to validate Proxy [namespace: gloo-system, name: gateway-proxy] with gloo validation: Listener Error: SSLConfigError. Reason: SSL secret not found: list did not find secret gloo-system.client-tls-secret

]
800 / 800 [--------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% 3 p/s 5m0s
objects done:  2022-11-10T15:37:54-06:00
```

jackstine:
not sure what is causing this issue, but it looks like maybe the upstream group `(Name: test, Namespace: gloo-system)` was never deleted in the previous test?

```
{"level":"error","ts":"2022-11-11T14:21:31.763Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:354","msg":"Validation failed: 1 error occurred:\n\t* Validating *v1.VirtualService failed: 1 error occurred:\n\t* Validating *v1.VirtualService failed: validating *v1.VirtualService name:\"missing-upstream\" namespace:\"gloo-system\": failed gloo validation resource reports: 2 errors occurred:\n\t* invalid resource gloo-system.gateway-proxy\n\t* upstream group not found, (Name: test, Namespace: gloo-system)
```

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -73,22 +73,6 @@ func (n *httpNetworkFilterTranslator) ComputeNetworkFilters(params plugins.Param
 	// was removed. If we want to support other network filters, we would process
 	// those plugins here.
 
-	// Check that we don't refer to nonexistent auth config
```
sam-heilbron, line 76: How did you confirm/test that the behavior this protected against is still functional?

projects/gateway/pkg/validation/validator.go
```diff
@@ -338,11 +284,7 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
 		glooReports, err := v.glooValidator(ctx, proxy, opts.Resource, opts.Delete)
 		if err != nil {
 			err = errors.Wrapf(err, failedGlooValidation)
-			if v.ignoreProxyValidationFailure {
-				contextutils.LoggerFrom(ctx).Error(err)
-			} else {
-				errs = multierr.Append(errs, err)
-			}
```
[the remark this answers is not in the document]
jackstine: The error that is returned now is if the snapshot has not been received by the gloo validator.
Let me add more context around the deprecated field.
https://github.com/solo-io/gloo/blob/validate-rate-limit-config/projects/gloo/pkg/validation/server.go#L210

The `ignoreProxyValidationFailure` flag was intended to ignore network issues between the Gateway server and the Gloo server.  Since the Gateway server and the Gloo server have merged, this is no longer an issue that can occur, and the field is now deprecated.

projects/gateway/pkg/validation/validator.go
```diff
@@ -585,31 +493,30 @@ func (v *validator) getReportsFromGlooValidation(reports []*gloovalidation.GlooV
...
+		return reports.ValidateStrict()
 	}
-	return multiErr
+	return reports.Validate()
 }
 
-func getErrors(err error) []string {
-	if err == nil {
-		return []string{}
+// UnmarshalResource is the same as the solo-kit pkg/utils/protoutils.Unmarshal() except it does not set the status of the resource
```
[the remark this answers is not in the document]
jackstine: let me check that out, and we can make a PR to change solo-kit

projects/gloo/pkg/translator/network_filters.go
jackstine: this should only occur in the extensions. This is a link to the test that should hit in Gloo.
https://github.com/solo-io/gloo/blob/10fc47fb95863982bb9a74bb9318d83bcb1785dd/test/kube2e/gateway/gateway_test.go#L2081

 I will add a test in the enterprise to show that this is occurring.

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -455,7 +455,7 @@ var _ = Describe("Translator", func() {
 	})
 
 	Context("Auth configs", func() {
-		It("will error if auth config is missing", func() {
+		It("will not error if auth config is missing", func() {
```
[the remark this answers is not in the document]
jackstine: this should only be called out in enterprise.  I can add to the changelog, if you like. But Ext-Auth should not be used in Gloo.  If there is an existing AuthConfig that is referenced, then it will report an error stating that auth config is enterprise only.  If the config is missing, then nothing occurs in Gloo.  Else in Enterprise it will report a warning on the missing Auth Config.

test/kube2e/gateway/gateway_test.go
```diff
@@ -1252,60 +1292,71 @@ var _ = Describe("Kube2e: gateway", func() {
 		})
 	})
 
-	Context("tests with RateLimitConfigs", func() {
-
-		var rateLimitConfig *v1alpha1skv1.RateLimitConfig
+	Context("validation will not always accept", func() {
```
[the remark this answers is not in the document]
jackstine: Yeah the wording is incorrect.

I think there is already enough validation in enterprise.

changelog/v1.13.0-beta27/extension-validation.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NEW_FEATURE
+    description: >
+      Validation of extensions. This can be used to validate Rate Limit Configs and Ext-Auth Configs.
```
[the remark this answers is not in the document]
jackstine: I like that idea.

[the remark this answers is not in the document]
jackstine: I will just say Validation of `TranslatorSyncerExtensions` is now enabled. since that is what we are doing.

And move the other resources to Enterprise Changelog.

projects/gloo/pkg/translator/translator_test.go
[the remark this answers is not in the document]
jackstine: Since Auth-Configs are enterprise only, we should not this in the enterprise changelog.

projects/gateway/pkg/validation/validator.go
```diff
@@ -533,48 +474,15 @@ func (v *validator) validateResource(opts *validationOptions) (*Reports, error)
...
-) {
-	var (
-		errs         error
-		proxyReports ProxyReports
-		proxies      []*gloov1.Proxy
-	)
+// getErrorsFromGlooValidation returns an error comprising of the gloo reports. The errors will include warnings if
+// allowWarnings is not set.
+func (v *validator) getErrorsFromGlooValidation(reports []*gloovalidation.GlooValidationReport) error {
+	var errs error
```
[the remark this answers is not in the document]
jackstine: I am sorry I do not follow.
1. what is the "Report type"?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -397,9 +395,16 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
 	if gvk == ListGVK {
 		return wh.validateList(ctx, admissionRequest.Object.Raw, dryRun)
 	}
+
+	if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk]; !hit {
+		contextutils.LoggerFrom(ctx).Debugf("unsupported validation for resource namespace [%s] name [%s] group [%s] kind [%s]", ref.GetNamespace(), ref.GetName(), gvk.Group, gvk.Kind)
```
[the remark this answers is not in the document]
elcasteel: If users are looking at logs from an aggregator rather than directly on the pod, they don't always have previous lines to provide context. As a general principle I like to treat this kind of debug message as something that needs to stand alone. (which is to say leave this code as is, assuming everything here is helpful for understanding why a this resource wasn't validated)

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -10,14 +10,16 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/solo-io/gloo/pkg/utils"
-	gatewayv1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1"
```
elcasteel, line 13: thanks for cleaning this up!

changelog/v1.13.0-beta27/extension-validation.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: NEW_FEATURE
+    description: >
+      Validation of `TranslatorSyncerExtensions` is now enabled.
```
kdorosh, line 4: since the changelog is user facing and read by our customers, i prefer this reads as it will affect them. more technical details you have written here (e.g. interface changes) should go in the PR description.

i think a user really just cares that we added support for ratelimit and extauth validation

projects/gateway/pkg/validation/validator.go
```diff
@@ -338,11 +284,7 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
 		glooReports, err := v.glooValidator(ctx, proxy, opts.Resource, opts.Delete)
 		if err != nil {
 			err = errors.Wrapf(err, failedGlooValidation)
-			if v.ignoreProxyValidationFailure {
```
kdorosh, line 341: I don't think we can remove this. we still support deploying the gateway and gloo pods (even if the default changed) and until we remove that functionality we should not remove this config

projects/gateway/pkg/validation/validator.go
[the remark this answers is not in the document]
kdorosh: i don't think we can remove this, see https://github.com/solo-io/gloo/pull/7385/files#r1024550463

projects/gateway/pkg/validation/validator.go
elcasteel: We don't support deploying the gateway pod at all. We support user supplied proxies but this validation webhook is never making a network call.

test/kube2e/gateway/gateway_test.go
```diff
@@ -77,6 +76,46 @@ var _ = Describe("Kube2e: gateway", func() {
...
+		}
+		attempt := 0
+		Eventually(func(g Gomega) bool {
+			placeholderUs.Metadata.Name = fmt.Sprintf("invalid-placeholder-us-%d", attempt)
+
+			_, err := resourceClientset.UpstreamClient().Write(placeholderUs, clients.WriteOpts{Ctx: ctx})
+			if err != nil {
+				// We have successfully rejected an invalid upstream
+				// This means that the webhook is fully warmed, and contains a Snapshot with a Proxy
+				return true
```
kdorosh, line 105: this can also happen if the error is a network error. we should ensure the error message is a validation rejection by checking the error message

test/kube2e/gateway/gateway_test.go
```diff
@@ -77,6 +76,46 @@ var _ = Describe("Kube2e: gateway", func() {
...
+				Static: &static.UpstreamSpec{
+					Hosts: []*static.Host{{
+						Addr: "~",
+					}},
+				},
+			},
+		}
+		attempt := 0
+		Eventually(func(g Gomega) bool {
+			placeholderUs.Metadata.Name = fmt.Sprintf("invalid-placeholder-us-%d", attempt)
```
kdorosh, line 99: nit; you probably want a newline here

kdorosh (commented):
lgtm overall!

projects/gateway/pkg/validation/validator.go
kdorosh: oh interesting, ok nevermind then i like this change. thanks for sharing @elcasteel

projects/gloo/pkg/translator/translator_test.go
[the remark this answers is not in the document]
elcasteel: I'm still confused about the reasoning behind this behavior change. Is there a corresponding update in solo-projects to error on missing auth configs? Is there something about the change to validation that required moving that check?

projects/gateway/pkg/validation/validator.go
jackstine: How can you deploy the gateway and gloo pods?

elcasteel: there is never a gateway pod (post 1.12)

projects/gloo/pkg/translator/network_filters.go
jackstine: brought this back in

projects/gloo/pkg/translator/translator_test.go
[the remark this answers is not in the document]
jackstine: Leaving this behavior as it was before

jackstine:
[xds-relay test is failing](https://github.com/solo-io/gloo/actions/runs/3491260446/jobs/5843688618)

projects/gateway/pkg/validation/validator.go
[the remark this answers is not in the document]
nfuden: Can we link the PR?

changelog/v1.13.0-beta27/extension-validation.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NEW_FEATURE
+    description: >
+      Validation webhook will now validate Auth Configs and Rate Limit Configs.
```
nfuden, line 4: are we always validating them? Is this togglable?

jackstine: Validated only if referenced on a VS or Route, is that some detail that we should include?

projects/gateway/pkg/validation/validator.go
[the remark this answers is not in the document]
nfuden: https://github.com/solo-io/gloo/pull/7385/files/5023798220783e2cdbea6d78bf004677e9e444ec#diff-abf1ea0c5351ae08b421977226f5aeddce902eb5caba0aa9beef54837a68bc6eR482 
The report that error came from I assume

projects/gateway/pkg/validation/validator.go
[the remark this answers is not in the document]
jackstine: here is the issue 
https://github.com/solo-io/solo-kit/issues/528

jackstine:
Ext-Auth configs in the following commits
https://github.com/solo-io/gloo/pull/7385/commits/417aab48e46b06a521a1319dfefe879f4c86994e
https://github.com/solo-io/gloo/pull/7385/commits/f5244f3b957ff86e3f11b8a5f7ea76eac84810bc

sam-heilbron (approved):

kdorosh (approved):

Issue #6933, opened 2022-08-11
Expose Retry Backoff Settings


guydc:
### Is your feature request related to a problem? Please describe.

Currently, envoy retries are fired in [0, 250]ms [intervals](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router-x-envoy-max-retries). By exposing the retry policy [backoff settings](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy-retrybackoff), it's possible to change the base and max interval. 

### Describe the solution you'd like

Expose `base_interval` and `max_interval` in the [gloo retry policy config](https://docs.solo.io/gloo-edge/master/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/retries/retries.proto.sk/).

irateswami:
👀

PR #7404, opened 2022-11-07
Expose retry backoff settings

irateswami:
# Description

Added the base and max interval for the retry policy backoff settings, pretty handy! 

- [ x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x ] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ x I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6933


projects/gloo/pkg/plugins/basicroute/plugin.go
```diff
@@ -250,9 +254,18 @@ func convertPolicy(policy *retries.RetryPolicy) *envoy_config_route_v3.RetryPoli
 		numRetries = 1
 	}
 
+	/** TODO possibly set defaults here?
+	FROM THE ENVOY DOCS: given the default interval, the first retry will be delayed randomly by 0-24ms, the 2nd by 0-74ms, the 3rd by 0-174ms, and so on. The interval is capped at a maximum interval, which defaults to 10 times the base interval (250ms). The default base interval (and therefore the maximum interval) can be manipulated by setting the upstream.base_retry_backoff_ms runtime parameter.
+	**/
+	retryBackOffPolicyIntervals := &routev3.RetryPolicy_RetryBackOff{
```
nfuden, line 260: What happens if max is less than min? It looks like it would cause invalid envoy config

projects/gloo/api/v1/options/retries/retries.proto
```diff
@@ -22,4 +22,10 @@ message RetryPolicy {
 
     // Specifies a non-zero upstream timeout per retry attempt. This parameter is optional.
     google.protobuf.Duration per_try_timeout = 3;
+
+    // Specifies the base interval for a retry
+    uint32 base_interval = 4;
```
nfuden, line 27: nit: add some of the caveats on the intended type for this?
Ie this currently looks (at least on phone) that it's setting an int that becomes time.duration.
If I remember correctly that's a nanoseconds?

projects/gloo/api/v1/options/retries/retries.proto
```diff
@@ -22,4 +22,10 @@ message RetryPolicy {
...
+
+    // Specifies the max interval for a retry
+    uint32 max_interval = 5;
```
nfuden, line 30: According to upstream proto there are some caveats, might as well copy those.

For example if requested duration is less than 1ms what happens.

projects/gloo/api/v1/options/retries/retries.proto
nfuden: Also should this be a uint32 or should we just have a protobuf duration?
Probably fine either way just wondering on the choice.

nfuden (commented):
Nice start! Some edge cases to address but otherwise seems reasonable. While we are here does it make sense to also consider the ratelimited retry? If we think they don't share enough concerns I'm happy punting it but since we are in the space the route proto had the field and without digging deeper it seemed interesting in context of what the user may want to be setting when they are setting a retry policy

projects/gloo/api/v1/options/retries/retries.proto
Bslabe123: Shouldnt it be an int64 as per https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/types/known/durationpb/duration.pb.go#L148?

Bslabe123: Shouldnt it be an int32 as per https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/types/known/durationpb/duration.pb.go#L148

irateswami: @nfuden I'm fine with it being either a `uint32` or a `google.protobuf.Duration`, do we want to have some kind of sane default? Or is this going to be a required field moving forward?

My concern is that a) why would a negative be a valid number, and b) if we don't supply a sane default, configs that have been previously written will now be invalid if the fields are not supplied.

nfuden: I don't see what configs would have been previously written?

I think sane guard rails make sense. In gloo code we can check and enforce. For example what envoy does with 1ms and under is interesting and should likely be handled on our end.

nfuden: Also for requires versus not check out nullability of protobuf duration

PR #7389, opened 2022-10-28
Upgrade tests

ianmacclancy:
# Description
Adding upgrade tests that will run on each PR
Issue: https://github.com/solo-io/solo-projects/issues/4189
Stability Milestone: https://github.com/solo-io/solo-projects/milestone/18 

# Context
Tests will be 
- Last minor latests patch release -> PR code 
- Current Minor last patch release -> PR code

PR #7408, opened 2022-11-08
Makefile m1update

ianmacclancy:
# Description

Remove conditional for architecture on push images. This would prevent m1 machines from pushing docker images when they are needed for use in kind cluster deployments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -745,7 +745,6 @@ docker-push: docker-push-local docker-push-non-arm
 
 .PHONY: docker-push-local
 docker-push-local: $(DOCKER_IMAGES)
-ifeq ($(CREATE_ASSETS), "true")
 	docker push $(IMAGE_REPO)/ingress:$(VERSION) && \
 	docker push $(IMAGE_REPO)/discovery:$(VERSION) && \
 	docker push $(IMAGE_REPO)/gloo:$(VERSION) && \
@@ -754,7 +753,6 @@ ifeq ($(CREATE_ASSETS), "true")
 	docker push $(IMAGE_REPO)/kubectl:$(VERSION) && \
 	docker push $(IMAGE_REPO)/sds:$(VERSION) && \
 	docker push $(IMAGE_REPO)/access-logger:$(VERSION)
-endif
 
 .PHONY: docker-push-non-arm
 docker-push-non-arm:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -745,7 +745,6 @@ docker-push: docker-push-local docker-push-non-arm
 
 .PHONY: docker-push-local
 docker-push-local: $(DOCKER_IMAGES)
-ifeq ($(CREATE_ASSETS), "true")
```
jenshu, line 748: instead of removing this check, would it work if you just set `CREATE_ASSETS ` or `CREATE_TEST_ASSETS` to true when you call this?

ianmacclancy: Sam and I discussed and this seemed to be the best option. The check is not needed based on the way we are setting and changing the set values of create assets and create test assets and we are looking into refactoring the make file as having create assets and create test assets end up more confusing than helpful. They were added to prevent unnecessary builds and values based on the location we are running but in this case, it's needed for both local and ci so the check should not exist.

jackstine: Are we sure that this will not have any side effects during releases/CI builds?

sam-heilbron: This CREATE_ASSETS guard was introduced a while back as a way of ensuring developers do not push to remote repositories that are meant for released versions. There is more we can do to simplify the Makefile I think, and prevent devs from executing the wrong commands, as well as ensure that our remote repos are not accessible locally without proper credentials.

As for this target, it's named `...-local` and so it feels strange that we need to guard against creating assets in the first place. Since this isn't used in CI, this seemed like the easiest way to unblock local development.

ianmacclancy: I have not seen any failures in the CI runs that have used it. It's being pulled out of https://github.com/solo-io/gloo/pull/7389 which has run successfully.

jackstine: I think you might find that we are pushing these images to quay.io every time we run CI. Whereas before we were not [logs](https://storage.googleapis.com/solo-public-build-logs/log-e1ea178b-32b5-462f-9466-31a263a59802.txt)

`docker push quay.io/solo-io/discovery:1.13.0-beta25-7408`

sam-heilbron: We have always pushed assets to quay during CI (https://github.com/solo-io/gloo/blob/master/cloudbuild.yaml#L124). This was done so we could depend on unreleased assets in solo-projects. 'CREATE_TEST_ASSETS' is meant to be the guard that protects against that.

jackstine (approved):
LGTM

jenshu (approved):

PR #7439, opened 2022-11-16
Fix for setting XDS_RELAY in regression matrix

elcasteel:
# Description
We want run regression tests with and without  XDS relay enabled. We configure this with an environment variable that is checked by the test setup code. However the github actions were only setting this when we create the cluster, not when we install gloo and run the tests.


elcasteel:
/skip-ci  GHA change won't affect the in-memory tests

nfuden (approved):

gunnar-solo (approved):

Issue #7398, opened 2022-11-02
expose jwt clockskew


kevin-shelaga:
### Version

1.12.x (latest stable)

### Is your feature request related to a problem? Please describe.

expose jwt clockskew so that jwts can remian valid longer than 60s after expiry

### Describe the solution you'd like

expose jwt clockskew

### Describe alternatives you've considered

passthrough auth

### Additional Context

done in mesh https://github.com/solo-io/gloo-mesh-enterprise/issues/4235

sam-heilbron:
Envoy API: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#extensions-filters-http-jwt-authn-v3-jwtprovider

Gloo API: https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/enterprise/options/jwt/jwt.proto#L49

Definition of Done: `clock_skew_seconds` exposed in the Gloo API

nfuden:
It seems reasonable to me that this issue should actually be to add all likely relevant fileds from the jwt_authn proto to be configurable in gloo. 
It MUST have a way to configure clock_skew_seconds, it MAY have other options configurable and it MUST have a rationale for each extant field and why it should or shouldnt have a bespoke toggle in gloo.

irateswami:
@nfuden Something I'm not understanding: [jwts are only in the enterprise version](https://docs.solo.io/gloo-edge/latest/guides/security/auth/jwt/). Does the entire `jwt` proto need to be implemented?

irateswami:
Whoops! Didn't mean to close

nfuden:
Reopening until this is released in enterprise

PR #7441, opened 2022-11-17
exposed the clock skew endpoint in the proto

irateswami:
# Description

exposing the jwt clock skew so that a jwt can last longer than 60s after expiration

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7398


nfuden:
Can we add a description or link to where the provider is passed as a whole to envoy?
Basically where does this value get consumed today?

nfuden (approved):

sam-heilbron (approved):

PR #7462, opened 2022-11-22
Glooctl improvements

sam-heilbron:
# Description
Small fixes for glooctl

# Context
## --with-gloo-fed=false
While testing installations of Gloo Federation, I was unable to use the `--with-gloo-fed=false` flag. As a quick workaround, I ensure that the flag only takes effect if the value is true and I supply the proper helm values in the `--values` flag.

To be clear, there are likely other fixes for this flag more holistically, but at least when you supply `--with-gloo-fed=false` the value is not supplied and the default is inherited (which is false).

### How to test?
1. `make glooctl`

## waitForCrds
Ensure that after performing an installation, we at least wait for the Settings CRD to be registered.

I had noticed during some gloo fed testing:
```
Gloo failed to install! Detailed logs available at /home/runner/.gloo/debug.log.
      Error: installing Gloo Edge Enterprise in gateway mode: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Settings" in version "gloo.solo.io/v1"
      : exit status 1
  occurred
```

# Checklist:


elcasteel:
For clarity, is there going to be a change in behavior when a user does `glooctl install gateway enterprise` without trying to set `--with-gloo-fed`?

projects/gloo/cli/pkg/cmd/install/installer.go
```diff
@@ -173,13 +177,34 @@ func (i *installer) installFromConfig(installerConfig *InstallerConfig) error {
...
 }
 
+func (i *installer) waitForCrdsToBeRegistered(ctx context.Context) error {
+	requiredCrds := []string{
+		"settings.gloo.solo.io",
+	}
+	apiExts := helpers.MustApiExtsClient()
+	logger := contextutils.LoggerFrom(ctx)
+	for _, crdName := range requiredCrds {
+		logger.Debugw("waiting for crd to be registered", zap.String("crd", crdName))
```
nfuden, line 199: I like that this is debug level but do you think there is value in being loud that we are in the crd registration wait?
Like an infof up at the start of the function? I know we have the failure at 201 but was just wondering if we wanted any other normal output?

projects/gloo/cli/pkg/cmd/install/enterprise.go
```diff
@@ -20,13 +20,17 @@ func enterpriseCmd(opts *options.Options) *cobra.Command {
...
-			}
 			if opts.Install.LicenseKey == "" {
 				return eris.New("No license key provided, please re-run the install with the following flag `--license-key=<YOUR-LICENSE-KEY>")
 			}
+
+			extraValues := map[string]interface{}{
+				"license_key": opts.Install.LicenseKey,
+			}
+			if opts.Install.WithGlooFed {
+				extraValues["gloo-fed.enabled"] = true
```
jenshu, line 31: Do these values get merged with the default values.yaml (which i believe has gloo-fed.enabled=true)? If so, how would we explicity set --with-gloo-fed=false?

sam-heilbron:
These changes aren't strictly necessary anymore, and it will take more effort to actually fix the with-gloo-fed flag correctly.

PR #7466, opened 2022-11-23
Istio Integration: Wrong Port When Service TargetPort Is Different Fix

inFocus7:
# Description

Fixes issue seen with Istio integration. When a `TargetPort` is set, specifically one that differs from the service's `Port`, the gateway proxy contacts the wrong port.

Moved our `istioIntegrationEnabled` `endpointsMap[]` logic outside `eps` loop, as that creates multiple upstream ref copies for the same `Epkey`.

Also added more logging to the kube dump on prefail for tests. Great for seeing recent events if an issue happens, it's how we noticed the issues with the tests were due to insufficient cpu for scheduling.

# Context

This bug was noticed when a user set differing `Port` and `TargetPort`. When `curl`ing to the service, the gateway proxy logs showed a failed (`503`) request to `ServiceIP:TargetPort`, and the app's Istio proxy didn't receive a request. The app's Istio proxy listens to `ServiceIP:ServicePort`, therefore we need to make sure that endpoint is created for istio integration.

# TODO
- [x] What if istio integration is enabled, **_but_** we're trying to hit an app that doesn't have a istio proxy injected? Make sure this still works, if not, then I'll need to narrow down this change to those with `TargetPort`s set up.
    - A simple route test on a static upstream ([here](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/destination_types/static_upstream/#create-upstream)) passed.

## Final Updates/Thoughts
- I noted that when `istioIntegrationEnabled()`, we'd always be routing to the `<ServiceIP>:<ServicePort>` even for pods that don't need it (Istio wouldn't (need to) be listening in on them). This shouldn't lead to an issue and it should still correctly resolve, therefore not taking care of this case, and just fixing the initial issue of needing to user `ServicePort` for istio integration.

# Reproduction Steps (post-fix)
1. Set up kind cluster + local images: `VERSION=0.0.0-kind ci/setup-kind.sh`
2. Installed istio onto cluster with demo profile: `istioctl install -y --set profile=demo`
3. Setup HTTPBin app with istio injection (this automatically sets it up with the differing port + targetPort which caused the issue) 
    -  `kubectl create ns httpbin && kubectl label namespaces httpbin istio-injection=enabled && kubectl apply -f https://raw.githubusercontent.com/istio/istio/master/samples/httpbin/httpbin.yaml`
4. Install Gloo Edge with Istio integration enabled (`global.istioIntegration.enableIstioSidecarOnGateway=true`) through helm values ([more info](https://docs.solo.io/gloo-edge/latest/guides/integrations/service_mesh/istio/))
5. Enabled access logging ([more info](https://docs.solo.io/gloo-edge/master/guides/security/access_logging/#:~:text=An%20access%20log%20may%20be,modifying%20the%20Gateway%20Custom%20Resource.))
6. Created a VirtualService to route to the HTTPBin upstream
7. Port-forwarded gateway-proxy
8. `curl`-ed the service and got an `OK`. `gateway-proxy` logs shows the request being routed to the proper port `ServiceIP:ServicePort`, and the httpbin istio-proxy logged a `200` response.

When this failed (pre-fix). The `curl` resulted in a `503`, with it being logged in the `gateway-proxy` but the istio-proxy didn't log a request/response since it didn't reach it.

# Checklist:

- [X] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.


projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -317,14 +317,15 @@ func filterEndpoints(
...
+				port := findFirstPortInEndpointSubsets(subset, singlePortService, kubeServicePort, istioIntegrationEnabled)
 				if port == 0 {
 					warnsToLog = append(warnsToLog, fmt.Sprintf("upstream %v: port %v not found for service %v in endpoint %v", usRef.Key(), spec.GetServicePort(), spec.GetServiceName(), subset))
 					continue
 				}
 
 				if istioIntegrationEnabled {
 					hostname := fmt.Sprintf("%v.%v", spec.GetServiceName(), spec.GetServiceNamespace())
+					// todo - could also just not `findFirstPortInEndpointSubsets` and use `kubeServicePort` here
 					key := Epkey{hostname, port, spec.GetServiceName(), spec.GetServiceNamespace(), usRef}
```
inFocus7, line 329: seeing as how we're just using the service's port, might be more performant/better to move the istio integration outside the subset loop, as we don't care about them. 
trying that in my current local changes. 
will need to check if there was any specific reason we were using the subset ports previously, and if that's still necessary. i could try testing locally with a setup that requires `targetPort`, and one which doesn't.

inFocus7: manually tested with following conditions:
- only a port: `PASS`
- port + target port different: `PASS`
- port + target port same: `PASS`

inFocus7:
making ready-for-review so all tests run

inFocus7:
update: limiting resources on just gloo doesn't solve limited cpu issue when deploying gloo pod. trying limiting on gateway as well (although, not sure of its initial value since it isn't set in helm), and removing ALS so there's more cpu for gloo deployment 🤞🏼

inFocus7:
note: last two commits are locally untested (since i get errors running built glooctl (although CI doesn't seem to face it)). The idea is to install the istio proxies _after_ the initial gloo setup using resources. not 100% certain it will fix the issue, but should hopefully separate needing resources while gloo installs and injecting the istio proxies later on.

inFocus7:
update time:
istio proxy was deployed in all services installed, but wasn't necessary so just installed on the important pods for these tests: `gateway-proxy` & `testrunner`. So now 2 less istio proxies consuming resources 🤞🏼

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -311,6 +311,16 @@ func filterEndpoints(
 			errorsToLog = append(errorsToLog, fmt.Sprintf("upstream %v: port %v not found for service %v", usRef.Key(), spec.GetServicePort(), spec.GetServiceName()))
 			continue
 		}
+
+		// Istio uses the service's port for routing requests
+		if istioIntegrationEnabled {
+			hostname := fmt.Sprintf("%v.%v", spec.GetServiceName(), spec.GetServiceNamespace())
+			key := Epkey{hostname, uint32(kubeServicePort.Port), spec.GetServiceName(), spec.GetServiceNamespace(), usRef}
```
kdorosh, line 318: does this use the kube service port or target port?

kdorosh: i thought istio typically just routes directly to pods to skip any k8s load balancing. so i would expect the service port to be ignored, and prefer the target port

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Labels:    map[string]string{"gloo": helper.TestrunnerName},
+			},
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{
+					{
+						Name:     "http",
+						Port:     port,
+						Protocol: corev1.ProtocolTCP,
+					},
+				},
```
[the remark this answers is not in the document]
inFocus7: On mobile atm, so not 100%, but i believe target port is set below. Basically so i can create and test a service with multiple configs if port + targetPort (second optional)

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: It uses the kube service port. I'll do a bit more investigation, but what we saw was that with istio integration, we need to route to the service port in order to hit the istio proxy (sidecar?). 
When the target port and port differed (as is with the HTTPBin example), we had issues since the pod's port was used (targetPort), but the istio proxy didnt receive a req.

We didn't have any tests beforehand, so hard to be 100% certain on what we previously expected. Which is why i created multiple tests

inFocus7: I could also walk through this locally if that helps. Currently taking linch, but can set up environment pre-this-fix and post, to see the requests on gateway-proxy + istio.

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Selector: map[string]string{"gloo": helper.TestrunnerName},
+			},
+		}
+		// set a targetPort if needed
+		if targetPort != -1 {
+			service.Spec.Ports[0].TargetPort = intstr.FromInt(targetPort)
+		}
```
kdorosh, line 57: super nit; i slightly prefer doing this logic before the service struct is initialized. it's a bit easier to read. e.g.

```golang
configuredTargetPort := 0 // or nil, whatever the default is
if targetPort != -1 {
  configuredTargetPort = targetPort
} 

svc := &service{
  // use configuredTargetPort here
}
```

projects/gloo/pkg/plugins/kubernetes/eds.go
kdorosh: ah this makes sense after re-reading. the map is to upstream refs, not endpoints

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Method:            "GET",
+				Host:              helper.TestrunnerName,
+				Service:           gatewayProxy,
+				Port:              gatewayPort,
+				ConnectionTimeout: 10,
+				Verbose:           true,
+				WithoutStats:      true,
+			}, fmt.Sprintf("HTTP/1.1 %d", expected), 1, time.Minute*1)
+		},
+			table.Entry("with non-matching, yet valid, port and target (app) port", int32(8000), helper.TestRunnerPort, http.StatusOK),
```
kdorosh, line 152: to ensure they are different, can we use say `helper.TestRunnerPort + 1` instead of `8000`?

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,159 @@
...
+			table.Entry("with matching port and target port", int32(80), helper.TestRunnerPort, http.StatusOK),
+			table.Entry("without target port, and port matching pod's port", int32(helper.TestRunnerPort), -1, http.StatusOK),
+			table.Entry("without target port, and port not matching app's port", int32(8000), -1, http.StatusServiceUnavailable),
```
kdorosh, line 155: same about not hardcoding `8000` here and instead incrementing the helper test runner port

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,174 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	skhelpers.RegisterPreFailHandler(helpers.KubeDumpOnFail(GinkgoWriter, testHelper.InstallNamespace))
+
+	// Install Gloo
+	values, cleanup := getHelmOverrides()
+	defer cleanup()
+
+	_ = testutils.Kubectl("create", "ns", testHelper.InstallNamespace)
+	_ = testutils.Kubectl("label", "namespace", testHelper.InstallNamespace, "istio-injection=enabled")
```
kdorosh, line 75: we should check errors?

test/kube2e/istio/istio_integration_test.go
inFocus7: sounds good! From a service config i set up without a target port (+ this [unofficial k8s doc](https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/services-networking/service/)), the default value for a TargetPort is the Port.

test/kube2e/istio/istio_suite_test.go
inFocus7: good call, added error checks for the kubectl commands

test/kube2e/istio/istio_integration_test.go
kdorosh: yeah we can set the default to nil and only override if we have provided a target port in the test

inFocus7: Tried but doesn't work, the `TargetPort` is of type `IntOrString` which doesn't allow a `nil` value

projects/gloo/pkg/plugins/kubernetes/eds.go
ben-taussig-solo, line 318: Nit: For readability, could we call out the field names in this key?
```suggestion
                        key := Epkey{
				Address:     hostname,
				Port:        uint32(kubeServicePort.Port),
				Name:        spec.GetServiceName(),
				Namespace:   spec.GetServiceNamespace(),
				UpstreamRef: usRef,
			}
			copy
```

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,176 @@
...
+	_, err = values.Write([]byte(`
+global:
+  istioIntegration:
+    disableAutoinjection: true
+    labelInstallNamespace: true
+gloo:
+  deployment:
+    resources:
+      requests:
+        cpu: 100m
```
ben-taussig-solo, line 158: is it necessary to set these resource limits on the Gloo deployment and Gateway Proxy podTemplate? If possible, could we reduce the helm overrides to just the istioIntegration settings?

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// patch only the gateway-proxy to be istio inject-able
+	err = testutils.Kubectl("patch", "-n", testHelper.InstallNamespace, "deployment", "gateway-proxy", "--patch", "{\"spec\": {\"template\": {\"metadata\": {\"labels\": {\"sidecar.istio.io/inject\": \"true\"}}}}}")
+	Expect(err).NotTo(HaveOccurred())
+
+	// Check that everything is OK
+	kube2e.GlooctlCheckEventuallyHealthy(1, testHelper, "90s")
+	// TODO(marco): explicitly enable strict validation, this can be removed once we enable validation by default
+	// See https://github.com/solo-io/gloo/issues/1374
+	kube2e.UpdateAlwaysAcceptSetting(ctx, false, testHelper.InstallNamespace)
```
ben-taussig-solo, line 87: Is this still relevant? I see that the issue linked on line 86 is closed

test/kube2e/istio/istio_suite_test.go
inFocus7: It _might_ be possible, I added this before limiting the istio proxies deployed which ultimately fixed resrouce limitation. Pushing latest PR to see if managing the resources doesn't matter anymore 👍🏼

inFocus7: Update: Will need these resource limitations. Removing them causes
```
Warning   FailedScheduling    pod/testrunner                        0/1 nodes are available: 1 Insufficient cpu.
```
(as seen in the [kube dumps](https://github.com/solo-io/gloo/actions/runs/3651018863/jobs/6167742200))

test/kube2e/istio/istio_suite_test.go
inFocus7: latest commit removes this. unable to test locally at the moment, so using CI to make sure it still runs as expected 👍🏼

ben-taussig-solo (dismissed):
Just validated this manually using the steps in the description. Looks good and thanks for the detailed testing steps!

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -311,6 +311,22 @@ func filterEndpoints(
...
+
+		// Istio uses the service's port for routing requests
+		if istioIntegrationEnabled {
+			hostname := fmt.Sprintf("%v.%v", spec.GetServiceName(), spec.GetServiceNamespace())
+			key := Epkey{
+				Address:     hostname,
+				Port:        uint32(kubeServicePort.Port),
+				Name:        spec.GetServiceName(),
+				Namespace:   spec.GetServiceNamespace(),
+				UpstreamRef: usRef,
```
kdorosh, line 323: i think this should be the copy ref, or else the pointer will update and be wrong at the end of the loop. same trick we do with the endpoints map

ideally this is tested and caught by a unit test for eds.go

kdorosh (changes requested):
apparent bug, or at least simplification (if `copyRef` is unneeded. but i think it is) is required

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+			err = resourceClientSet.UpstreamClient().Delete(upstreamRef.Namespace, upstreamRef.Name, clients.DeleteOpts{
+				IgnoreNotExist: true,
+			})
+			helpers.EventuallyResourceDeleted(func() (resources.InputResource, error) {
+				return resourceClientSet.UpstreamClient().Read(upstreamRef.Namespace, upstreamRef.Name, clients.ReadOpts{})
+			})
+		})
+
+		table.DescribeTable("should act as expected with varied ports", func(port int32, targetPort int, expected int) {
```
kdorosh, line 146: love this test very compact :clap:

sam-heilbron (changes requested):
I checked out your changes, undid the eds.go changes  (just using the version of the file on the master branch) and then ran your regression tests (thank you for adding them!) and they all passed.

Given that, it seems these tests are a good addition to validate some cases, but don't test the new behavior we've introduced. Can you add a test which passes with your changes but will fail without them?

Also, I agree with Kevin's point that our EDS logic is very complex and deserves some more enhanced unit tests to validate the logic.

inFocus7:
> Can you add a test which passes with your changes but will fail without them?

I tried it too and got the same results. The `with non-matching, yet valid, port and target (app) port` should be the port that determines it's fixed, so unsure why it's passing on testrunner. I'll have to test it locally to see why `testrunner` doesn't have the issue, but when I set it up (locally) with `httpbin`, we do see the issue.

inFocus7:
Self-performed local tests to find out what the issue is:
- [ ] using httpbin, instead of “testrunner”
- [X] For httpbin deployment, `ports.containerPort` is set. What if it's set in testrunner pod as well? (investigated by adding containerPort to the pods creation; not issue)
- [ ] maybe due to install gloo enterprise locally, but tests use OSS?
- [ ] Is the `curl`? hitting the gateway proxy service directly, instead of the IP?
- [X] test uses istio 1.11.4, i use 1.15.1 (looked into it; not issue)
- [X] I have autoinjection enabled locally (looked into it; not issue)
- [X] Maybe the existing services aren’t getting deleted/created? (looked into it; not issue)

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: Ah great catch! Just pushed the change!

inFocus7: resolving again. False-flag ^. It was working since, after moving stuff back, i mistakenly still kept "uint32(kubeServicePort.Port)" as the Port which is the fix.

inFocus7:
Latest changes (past two commits) have been reformating so that we only use the istio eds logic (adding endpoint with service information) when adding pods that have an `istio-proxy` container listening. 

Originally we didn't care for it, and did the istio eds logic as long as `isIstioIntegrationEnabled()` was true. If we want/need to keep that approach, then I'll simply revert the latest changes so we're _only_ updating the port to be the service port (what istio listens on) without caring for `istio-proxy` deployment.

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			helpers.EventuallyResourceDeleted(func() (resources.InputResource, error) {
+				return resourceClientSet.VirtualServiceClient().Read(virtualServiceRef.Namespace, virtualServiceRef.Name, clients.ReadOpts{})
+			})
+
+			err = resourceClientSet.ServiceClient().Delete(serviceRef.Namespace, serviceRef.Name, clients.DeleteOpts{
+				IgnoreNotExist: true,
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(func() bool {
```
ben-taussig-solo, line 132: Nit: is there a particular reason we don't use `helpers.EventuallyResourceDeleted` here?

inFocus7: Good call, I forgot the reason earlier but after looking into it: 
`helper.EventuallyResourceDeleted` takes in a function parameter that returns a `resources.InputResource`. That is a type/interface we own and our input resources (VirtualServices, Clients, etc.) implement.
 
_But_ `ServiceClient` is a k8s resource, therefore it doesn't implement our `InputResource` interface, and we can't use it with `helper.EventuallyResourceDeleted`.

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+				Host:              helper.TestrunnerName,
+				Service:           gatewayProxy,
+				Port:              gatewayPort,
+				ConnectionTimeout: 10,
+				Verbose:           false,
+				WithoutStats:      true,
+				ReturnHeaders:     true,
+			}, fmt.Sprintf("HTTP/1.1 %d", expected), 1, time.Minute*1)
+		},
+			table.Entry("with non-matching, yet valid, port and target (app) port", int32(helper.TestRunnerPort+1), helper.TestRunnerPort, http.StatusOK),
```
sam-heilbron, line 162: This is a great example of both positive and negative test cases!!

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+		// glooctl should delete the namespace. we do it again just in case it failed
+		// ignore errors
+		_ = testutils.Kubectl("delete", "namespace", testHelper.InstallNamespace)
+
+		EventuallyWithOffset(1, func() error {
+			return testutils.Kubectl("get", "namespace", testHelper.InstallNamespace)
+		}, "60s", "1s").Should(HaveOccurred())
+
+		cancel()
```
sam-heilbron, line 126: I think that cancel should happen regardless of the TEAR_DOWN env

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,174 @@
...
+	skhelpers.RegisterCommonFailHandlers()
+	skhelpers.SetupLog()
+	_ = os.Remove(cliutil.GetLogsPath())
+	junitReporter := reporters.NewJUnitReporter("junit.xml")
+	RunSpecsWithDefaultAndCustomReporters(t, "Istio Suite", []Reporter{junitReporter})
+}
+
+var (
+	testHelper  *helper.SoloTestHelper
+	ctx, cancel = context.WithCancel(context.Background())
```
sam-heilbron, line 52: This should probably be initialized in the BeforeSuite since we clean it up in the AfterSuite

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,174 @@
...
+    resources:
+      requests:
+        cpu: 100m
+        memory: 256Mi
+gatewayProxies:
+  gatewayProxy:
+    podTemplate:
+      resources:
+        requests:
+          cpu: 100m
```
sam-heilbron, line 163: As we move towards a more standard way of setting resource limits for tests, I think it would be valuable to call these out loudly in a comment. Perhaps indicating the behavior we noticed when we did not set resource limits, and even perhaps a tip in the PR outlining how developers can debug that behavior.

sam-heilbron (commented):
Changes look good to me and love the new tests!

Thanks for calling out:
```
I noted that when istioIntegrationEnabled(), we'd always be routing to the <ServiceIP>:<ServicePort> even for pods that don't need it (Istio wouldn't (need to) be listening in on them). This shouldn't lead to an issue and it should still correctly resolve, therefore not taking care of this case, and just fixing the initial issue of needing to user ServicePort for istio integration.
```

Though this works, I wonder if this behavior is preferred. Since this is the existing behavior (before your changes) I don't think we need to change it as part of this work, but could you reach out to the field/product to clarify this behavior of the flag, and perhaps write up an issue for a new feature if we need to be able to dynamically determine if istio integration is enabled?

sam-heilbron:
/kick

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=203cf664-80f9-4e81-a057-fadb86d79b76

```
Fail] Consul e2e with envoy and gloo [It] resolves consul services with hostname addresses (as opposed to IPs addresses) 
Step #10 - "test": /workspace/gloo/test/e2e/consul_test.go:237
```

I did not see an issue tracking this flake. If we see this more frequently we can create one

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,176 @@
...
+	var err error
+
+	ctx, cancel = context.WithCancel(context.Background())
+
+	testHelper, err = kube2e.GetTestHelper(ctx, namespace)
+	Expect(err).NotTo(HaveOccurred())
+
+	skhelpers.RegisterPreFailHandler(helpers.KubeDumpOnFail(GinkgoWriter, testHelper.InstallNamespace))
+
+	// Install Gloo
```
sam-heilbron, line 69: nit: this comment is inaccurate. I think it's meant to be a few lines lower

sam-heilbron (approved):
Great work!

kdorosh (approved):

test/kube2e/istio/istio_integration_test.go
```diff
@@ -0,0 +1,169 @@
...
+			table.Entry("with matching port and target port", int32(helper.TestRunnerPort), helper.TestRunnerPort, http.StatusOK),
+			table.Entry("without target port, and port matching pod's port", int32(helper.TestRunnerPort), -1, http.StatusOK),
+			table.Entry("without target port, and port not matching app's port", int32(helper.TestRunnerPort+1), -1, http.StatusServiceUnavailable),
+			table.Entry("pointing to the wrong target port", int32(8000), helper.TestRunnerPort+1, http.StatusServiceUnavailable), // or maybe 404?
```
nfuden, line 166: Do we have certainty here? I dont love an or as a comment here as it sounds like a potential flake. 
Perhaps we have gotten clarity since that merits the removal of this comment?

sam-heilbron (approved):

test/kube2e/istio/istio_suite_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/solo-io/gloo/test/helpers"
+	"github.com/solo-io/go-utils/log"
+	"github.com/solo-io/go-utils/testutils"
+	"github.com/solo-io/k8s-utils/testutils/helper"
+	skhelpers "github.com/solo-io/solo-kit/test/helpers"
+)
+
+const (
+	gatewayProxy = gatewaydefaults.GatewayProxyName
+	gatewayPort  = int(80)
```
nfuden, line 33: we seem to use this alot of places is it time to move this to gatewaydefaults? 
gatewaydefaults.DefaultGatewayPort?

I just dislike magic numbers proliferating

PR #7474, opened 2022-11-28
Feat/lambda with basic defaults

nfuden:
# Description

New Lambda functionality to enable route level overrides as well as a way to use the first function for uses cases such as older portal that cant set all the fields for aws specs on routes.

# Context

This PR does not attempt to change anything about envoy configuration for those who have not opted in or set new options.
The ability to use the first function should only occur if and when helm configuration is set as it is a behavior change that we likely don't want in general. 
https://github.com/solo-io/solo-projects/pull/4313 implements the breaking function signature change in ee.

# Testing Notes
The first consideration in this PR is routes without destinationspec for functions. This should only occur if the helm setting is enabled.
To test take a valid setup with 1 function. Go to the route and delete the function specifier. Expect an error. Update settings to allow for defaults expect to get back to the same envoy configuration.

The second consideration is defaults. Take a valid setup with a destinationspec that has at least 1 non-default value, for example request transformation. 
1. Set that non-default value on the upstream's defaults. Nothing should change. 
2. Set that non-default value back to default. Nothing should change
3. Set the non-default value back to default on the upstream defaults. The configuration should change.


projects/gloo/api/v1/options/aws/aws.proto
```diff
@@ -42,6 +42,9 @@ message UpstreamSpec {
    
     // Optional override to disable role chaining;
     bool disable_role_chaining = 6;
+    
+    // Default Destination settings. Will override any settings on normal destinations with any non-empty defaults.
```
ben-taussig-solo, line 46: Nit: what do we mean by `normal destinations` -- should it be just `destinations` or `destinations pointing to this upstream` instead?

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -117,45 +126,23 @@ func (p *Plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 		ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: typedConfig},
 	}
 
-	var accessKey, sessionToken, secretKey string
+	// To utilize the aws lambda plugin much of the power comes via its secret management
+	// Check that one of the supported auth paradigms in enabled.
+	// Currently: static secret ref, credential discovery or ServiceAccountCreds such in eks
```
ben-taussig-solo, line 131: Super nit: can we add a date to this comment, I.E.: `Currently as of 11/29/2022:`

ben-taussig-solo: I guess git history makes this clear, closing the conversation

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -334,35 +321,100 @@ func (p *Plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
...
+		// TransformerConfig is intentionally not included as that is an enterprise only feature
+		TransformerConfig: nil,
+	}
+
+	return lambdaRouteFunc, nil
+
+}
+
+// deriveStaticSecret from ingest if we are using a kubernetes secretref
+// TODO(nfuden): find a way to pull out params from this.
```
ben-taussig-solo, line 389: What is the status of this TODO?

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -334,35 +321,100 @@ func (p *Plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
...
+// Named returns with the derived string contents or an error due to retrieval or format.
+func deriveStaticSecret(params plugins.Params, secretRef *core.ResourceRef) (access, session, secret string, err error) {
+	glooSecret, err := params.Snapshot.Secrets.Find(secretRef.Strings())
+	if err != nil {
+		err = errors.Wrapf(err, "retrieving aws secret")
+		return
```
ben-taussig-solo, line 395: Is this identical to `return "", "", "", err`?

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -334,35 +321,100 @@ func (p *Plugin) HttpFilters(_ plugins.Params, _ *v1.HttpListener) ([]plugins.St
 	return filters, nil
 }
 
-func GenerateAWSLambdaRouteConfig(destination *aws.DestinationSpec, upstream *aws.UpstreamSpec) (*AWSLambdaPerRoute, error) {
+// GenerateAWSLambdaRouteConfig is an overridable way to handle destination logic for lambdas.
+// Passed in at plugin creation as it fulfills PerRouteConfigGenerator interface
+func GenerateAWSLambdaRouteConfig(options *v1.GlooOptions_AWSOptions, destination *aws.DestinationSpec, upstream *aws.UpstreamSpec) (*AWSLambdaPerRoute, error) {
+
+	// merge the non-default values (trues and non-zeros) from default onto destination
```
ben-taussig-solo, line 328: The comment suggests that you're aware of this, but Gunnar and I have each dealt with some frustrating behavior related to merging explicitly specified default values using mergo (AFAIK this is not a failure of the library, but moreso an issue with using non-nil-able datatypes). To what extent do we expect this to be an issue for us? Are there usecases we want to support which might be blocked by this limitation?

Off the top of my head, I feel like most lambda destination settings require a user to set a flag to `true`, so we might be safe

nfuden: We chatted in slack around this and decided that this felt safe for the current fields and that also as we add more fields we should always try to have them be nullable otherwise this type of operation is potentially dangerous

projects/gloo/api/v1/options/aws/aws.proto
nfuden: Does the updated wording work for you?

changelog/v1.14.0-beta1/lambda-no-function.yaml
```diff
@@ -0,0 +1,11 @@
+changelog:
+  - type: NEW_FEATURE
+    description: >
+      Add option that allows usage of lambda upstreams without specifying a function.
+      Affects custom extensions of the lambda extension built in go that use gloo as a library.
```
kdorosh, line 5: nit; can we add a bit more information for the user so they can decide at a glance if this is an interesting feature for them or not? i.e. what's the UX for this new feature how would a user interact and use this?

projects/gloo/api/v1/options/aws/aws.proto
```diff
@@ -42,6 +42,9 @@ message UpstreamSpec {
    
     // Optional override to disable role chaining;
     bool disable_role_chaining = 6;
+    
+    // Default Destination settings. Will override any settings on specified destinations with any non-default settings specified.
+    DestinationSpec default_destination_settings = 7;
```
kdorosh, line 47: our typical model is more specific resources override more generic / parental ones. thus i think the name `default_destination_settings` is a bit redundant. perhaps just `destination_overrides`?

projects/gloo/api/v1/settings.proto
```diff
@@ -533,6 +533,13 @@ message GlooOptions {
         // Does not affect the default filewatch for service account only augments it.
         // Defaults to not refreshing on time period. Suggested is 15 minutes.
         google.protobuf.Duration credential_refresh_delay = 4;
+
+        // Sets the unsafe behavior where a route can specify a lambda upstream
```
kdorosh, line 537: can we call out for users who may not be as security focused why this is unsafe?

projects/gloo/api/v1/options/aws/aws.proto
nfuden: Given that our proto is not nullable I wanted to call out loudly that we dont have the cleanest of behaviors here.
I am happy to make the name change if we prefer

kdorosh: perhaps we can make the name change and update the comment to explicitly call out that all fields will be overwritten (no merging)?

projects/gloo/api/v1/options/aws/aws.proto
ben-taussig-solo: Currently the comment says `Will override any settings on specified destinations with any non-default settings specified.`

This kind of leaves me with two questions:
 - _which_ settings get overridden?
 - how is a destination specified

I think the answers to these questions are:
 - The AWS DestinationSpec gets overridden
 - This occurs for any route that points to this upstream which specifies an AWS destinationSpec

With that in mind, could we change this to something like:

`Specifies default AWS DestinationSpec configuration for any route targeting this upstream. Note that the route in question must have an AWS DestinationSpec to be affected.`

^^^ I don't love this, but I'm trying to eliminate ambiguity

projects/gloo/api/v1/settings.proto
nfuden: I think the lines below represent the risk on it.

nfuden: Would you prefer if we prefix that line with something along the lines of `What makes this dangerous` and then go into the explanation?

projects/gloo/api/v1/options/aws/aws.proto
nfuden: Yeah I feel it I put an adapted version of that in the comment.

ben-taussig-solo: Looks good!

changelog/v1.14.0-beta1/lambda-no-function.yaml
nfuden: Yeah I dont want to call it out too loudly but I updated the changelog to describe the use case a bit more explicitly. 
Do you feel the new format covers it?

kdorosh: I would just mention that it's Gloo Portal not just Portal but yes lgtm

projects/gloo/api/v1/settings.proto
kdorosh: yeah i think that extra prefix helps, but upon second read this isn't totally misleading

changelog/v1.14.0-beta1/lambda-no-function.yaml
nfuden: Yeah good point

ben-taussig-solo (approved):

kdorosh (approved):

PR #7483, opened 2022-11-28
Goroutine Leak

Bslabe123:
# Description

Changes to checkResources to prevent accumulating goroutines on opts.Top.ctx if called repeatedly.

# Context

When running our tests with the race flag, we are limited to 8192 goroutines. Currently the `checkResources` function creates ~2000 per run and if we accumulate these all on the same context by running the function multiple times as we do in `GlooctlCheckEventuallyHealthy`, then we will hit that limit and error. To fix this, the function now creates a new context with cancel from opts.Top.ctx and runs all goroutines on that before cleaning itself up.

# Checklist:


projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -97,7 +97,9 @@ func RootCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra.
 func CheckResources(opts *options.Options) error {
 	var multiErr *multierror.Error
 
-	err := checkConnection(opts.Top.Ctx, opts.Metadata.GetNamespace())
+	ctx, cancel := context.WithCancel(opts.Top.Ctx)
```
ianmacclancy, line 100: ctx is used throughout this file but now we are passing two contexts to most functions - one as ctx and the other through opts as opts.Top.ctx. Could we name the context with cancel in a way that indicates its different than opts.Top.ctx - could use ctxWithCancel

ianmacclancy (approved):

nfuden (approved):

Issue #5462, opened 2021-10-12
glooctl check needs a "readonly" mode


jdef:
**Is your feature request related to a problem? Please describe.**

We want to give users the ability to run `glooctl check ...` in order to verify their configuration. This is complicated by the fact that "check" requires `create` permission for the `pods/portforward` subresource.

**Describe the solution you'd like**

We want to be able to specify a set of readonly RBAC settings so that users can run, without errors, `glooctl check ...` in order to validate their gloo configuration(s).

**Describe alternatives you've considered**

Forking glooctl ?

**Additional context**

PR #7492, opened 2022-12-01
`glooctl check` `--read-only` flag

Bslabe123:
# Description

Added a `--read-only` flag to `glooctl check` which disables any port forward gets and any checking of resulting resources.

# Context

Whenever we go a `kubectl port-forward` we create a new resource in our cluster to handle the routing of traffic. This is a problem when we want to check resources that require port-forward GETs if we want to have a check that doesn't modify the cluster whatsoever (i.e. a check that wouldnt require RBAC create permission). This change now allows us to have a version of glooctl check that doesn't require RBAC create permissions albeit by bypassing some checks.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5462


projects/gloo/cli/pkg/cmd/options/options.go
```diff
@@ -451,4 +451,7 @@ type Register struct {
 type Check struct {
 	// The maximum length of time to wait before giving up on a secret request. A value of zero means no timeout.
 	SecretClientTimeout time.Duration
+
+	// Makes check read only by skipping any checks that create resources in the cluster
+	ReadOnly bool
```
nfuden, line 456: Should this be at the root level? ie should this be only for glooctl check or for other glooctl operations as well.

projects/gloo/cli/pkg/cmd/check/proxy_prom_stats.go
```diff
@@ -24,16 +26,21 @@ const promStatsPath = "/stats/prometheus"
...
 	var multiWarn *multierror.Error
 	for _, deployment := range deployments.Items {
 		if deployment.Labels["gloo"] == "gateway-proxy" || deployment.Name == "gateway-proxy" || deployment.Name == "ingress-proxy" || deployment.Name == "knative-external-proxy" || deployment.Name == "knative-internal-proxy" {
 			gatewayProxyDeploymentsFound++
 			if *deployment.Spec.Replicas == 0 {
 				multiWarn = multierror.Append(multiWarn, eris.New("Warning: "+deployment.Namespace+":"+deployment.Name+" has zero replicas"))
-			} else if err := checkProxyPromStats(ctx, glooNamespace, deployment.Name); err != nil {
-				return err, multiWarn
+			} else if opts.Check.ReadOnly {
+				multiWarn = multierror.Append(multiWarn, eris.New("Warning: checking proxies with port forwarding is disabled"))
```
nfuden, line 38: From a ux perspective does it make sense to do this per proxy? It seems to me that we should call out replica counts of 0 more loudly than this. Therefore I propose moving of this return to outside the for loop and simplifying your current else to be an else if

projects/gloo/cli/pkg/cmd/check/gloo_stats.go
```diff
@@ -79,36 +79,36 @@ func checkXdsMetrics(ctx context.Context, opts *options.Options, glooNamespace s
...
-		fmt.Println(errMessage)
-		return fmt.Errorf(errMessage)
-	}
-
-	for _, deployment := range deployments.Items {
-		if deployment.Name == rateLimitDeployment {
-			printer.AppendCheck("Checking rate limit server... ")
-			if !RateLimitIsConnected(stats) {
-				return fmt.Errorf("rate limit server is not connected")
+	if opts.Check.ReadOnly {
```
nfuden, line 82: why is this such a big if wrapper?
https://danp.net/posts/reducing-go-nesting/

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -871,18 +872,21 @@ func checkProxies(ctx context.Context, opts *options.Options, namespaces []strin
...
 		return multiErr
 	}
 	printer.AppendStatus("proxies", "OK")
 	if warnings != nil && warnings.Len() != 0 {
-		printer.AppendCheck(warnings.Error() + "\n")
+		accumulatedWarning := ""
+		for _, warning := range warnings.Errors {
+			accumulatedWarning += warning.Error() + "\n"
+		}
+		printer.AppendCheck(accumulatedWarning)
```
jackstine, line 889: why was this necessary?

Bslabe123: This is mostly just to make the warnings print nicely, theres also an issue where if we use ther Error() function then we also print `n errors occured:` even though were printing warnings.

projects/gloo/cli/pkg/cmd/options/options.go
Bslabe123: I can definitely be convinced of this after looking at some of the behavior in `glooctl get`. Adding

projects/gloo/cli/pkg/cmd/check/gloo_stats.go
Bslabe123: Oversight, I see what you mean! Fixed

projects/gloo/cli/pkg/cmd/check/proxy_prom_stats.go
Bslabe123: The only reason I have this check inside the for loop is so that we can accumulate both the 'no replicas' error and the new warning while also making sure that we dont call checkProxyPromStats if we have read only enabled.

nfuden (approved):

nfuden (approved):

projects/gloo/cli/pkg/cmd/check/proxy_prom_stats.go
```diff
@@ -24,23 +26,28 @@ const promStatsPath = "/stats/prometheus"
...
 	var multiWarn *multierror.Error
+	var readOnlyErr error
 	for _, deployment := range deployments.Items {
 		if deployment.Labels["gloo"] == "gateway-proxy" || deployment.Name == "gateway-proxy" || deployment.Name == "ingress-proxy" || deployment.Name == "knative-external-proxy" || deployment.Name == "knative-internal-proxy" {
 			gatewayProxyDeploymentsFound++
 			if *deployment.Spec.Replicas == 0 {
 				multiWarn = multierror.Append(multiWarn, eris.New("Warning: "+deployment.Namespace+":"+deployment.Name+" has zero replicas"))
-			} else if err := checkProxyPromStats(ctx, glooNamespace, deployment.Name); err != nil {
-				return err, multiWarn
+			} else if opts.Top.ReadOnly {
```
sam-heilbron, line 38: I haven't read through this function exactly, but if we're in read-only mode, does this function ever do something? If not, I think it would be simpler to pull this logic out of the for loop, and perhaps even out of the function

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=06f78f22-f77a-4f8e-a47d-5cd993d07e62

projects/gloo/cli/pkg/cmd/check/gloo_stats.go
```diff
@@ -79,6 +79,10 @@ func checkXdsMetrics(ctx context.Context, opts *options.Options, glooNamespace s
 	localPort := strconv.Itoa(freePort)
 	adminPort := strconv.Itoa(int(defaults.GlooAdminPort))
 	// stats is the string containing all stats from /stats/prometheus
+	if opts.Top.ReadOnly {
```
sam-heilbron, line 82: Similar for this one, if in read-ony mode, does checkXDS metric ever do anything?

sam-heilbron (commented):
I like this addition! I think the changelog clearly outlines the new behavior, and the tests verify the functionality. 
What is a little unknown to me (perhaps my ignorance): we have some metrics checks that require port-forwarding, which we're allowing to be skipped. Would it make more sense, instead of handling that skip logic nested in those functions, handling the guard in the top level check function?

To me CheckResources perform a set of steps, and each step either require read privilege or write privileges. I could picture we guard the read-only mode once in that function, and inside an if statement, perform all the check steps that require write privilege.

projects/gloo/cli/pkg/cmd/check/proxy_prom_stats.go
Bslabe123: Its a little subtle but we do still want to accumulate and report errors if we have no deployments found or deployments with zero replicas and if we did pull this functionaltiy out of the loop we would still need to keep `else if opts.Top.ReadOnly` to prevent `checkProxyPromStats` from being called.

projects/gloo/cli/pkg/cmd/check/gloo_stats.go
Bslabe123: in read only mode the function only does the below which i would believe is worth keeping but I could be convinced otherwise 
```	
if deployments == nil {
		fmt.Println("Skipping due to an error in checking deployments")
		return fmt.Errorf("xds metrics check was skipped due to an error in checking deployments")
	}
```

Bslabe123:
> I like this addition! I think the changelog clearly outlines the new behavior, and the tests verify the functionality. What is a little unknown to me (perhaps my ignorance): we have some metrics checks that require port-forwarding, which we're allowing to be skipped. Would it make more sense, instead of handling that skip logic nested in those functions, handling the guard in the top level check function?
> 
> To me CheckResources perform a set of steps, and each step either require read privilege or write privileges. I could picture we guard the read-only mode once in that function, and inside an if statement, perform all the check steps that require write privilege.

That was one of the approaches I considered! The only issue with that is that it would essentially be an alias for `glooctl check -x proxies,xds-metrics`. It would bypass the warning and/or erroring if we have incorrect gateway-proxy deployments.

projects/gloo/cli/pkg/cmd/check/gloo_stats.go
sam-heilbron: Makes sense, I was picturing we could have checkXdsMetrics and checkXdsMetricsReadOnly and that way each could do 1 thing. But not tied to the idea!

sam-heilbron (approved):
Awesopme work!

PR #7503, opened 2022-12-02
[DNM]run CI with k8s 1.23.14

elcasteel:
For release testing, show that our k8s based regressions work on 1.23


elcasteel:
I accidentally opened against master instead of 1.13

Issue #2340, opened 2020-01-31
Feat: Expose envoy's`config.filter.accesslog.v3.AccessLogFilter`


NelsonJeppesen:
### Use-case
I want to enable Access Logs for tracking but the number of logs is overwhelming. I'd like to use `config.filter.accesslog.v3.RuntimeFilter` to log just a percentage of the logs
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/accesslog/v3/accesslog.proto#envoy-v3-api-msg-config-accesslog-v3-runtimefilter

### Feature request
Expose envoy's `config.filter.accesslog.v3.RuntimeFilter` in Gloo's Gateway CRD

Something like this:
```diff
apiVersion: gateway.solo.io/v1
kind: Gateway
metadata:
  annotations:
    origin: default
  name: gateway
  namespace: gloo-system
spec:
  bindAddress: '::'
  bindPort: 8080
  proxyNames: 
    - gateway-proxy
  httpGateway: {}
  useProxyProto: false
  options:
    accessLoggingService:
      accessLog:
      - fileSink:
          path: /dev/stdout
+         filter:
+           runtime_filter:
+             runtime_key: somekey
+             percent_sampled: 
+                 numerator: "..."
+               denominator: "..."
+           use_independent_randomness": true
          stringFormat: ""
```

delta50:
I am also interested in the ability to use envoy's access log filter. I was using it before we moved to gloo from a custom control plane to log requests that resulted in 4xx or 5xx responses.

alex-mozejko:
this would be super helpful

bcollard:
got a new ask about this feature, from a prospect

kevin-shelaga:
reopening for backport to 1.13

nfuden:
backported to v1.13.19 adn v1.14.5

PR #7511, opened 2022-12-05
Expose config filter accesslog v3 access log

ghost:
# Description

Added the option to use runtime filter on access logs. 

This is one of the access log filters that is supported by envoy. 

Manual e2e test:

Build the gloo docker package: VERSION=some-version make gloo-docker -B
Run Gloo locally backed by a filesystem: https://docs.solo.io/gloo-edge/latest/installation/gateway/development/docker-compose-file/
Include the GLOO_VERSION in the up command: GLOO_VERSION=some-version docker-compose up
Update install/docker-compose-file/data/config/gateways/gloo-system/gateway-proxy.yaml
by adding a filter to an access log:

       filter:
        status_code_filter:
          comparison:
            op: GE
            value:
              runtime_key: "400"
and validate you can see in the config dump

# Context

User requested this feature to to make it easy to track access logs. 

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/2340


projects/gloo/api/v1/options/als/als.proto
```diff
@@ -40,6 +42,10 @@ message FileSink {
         // https://www.envoyproxy.io/docs/envoy/v1.14.1/configuration/observability/access_log#format-dictionaries
         google.protobuf.Struct json_format = 3;
     }
+
+    // Filter which is used to determine if the access log needs to be written.
+    AccessLogFilter filter = 4;
```
jenshu, line 47: how do you feel about moving this field up into the `AccessLog` message, so we don't have to repeat it for each output destination type?

docs/data/ProtoMap.yaml
```diff
@@ -14,15 +14,51 @@ apis:
   als.options.gloo.solo.io.AccessLog:
```
ashishb-solo, line 14: out of curiosity, what is the empty `docker-local` file that was added?

test/e2e/access_log_test.go
```diff
@@ -227,6 +275,100 @@ var _ = Describe("Access Log", func() {
...
+														Kind: &structpb.Value_StringValue{
+															StringValue: "%REQ(:METHOD)%",
+														},
+													},
+												},
+											},
+										},
+									},
+								},
+								Filter: &als.AccessLogFilter{
```
ashishb-solo, line 308: test is a little hard to read cause the indentation is so deep. could you perhaps just break the or-filter into a separate variable?

test/e2e/access_log_test.go
```diff
@@ -140,6 +142,52 @@ var _ = Describe("Access Log", func() {
...
+											},
+											UseIndependentRandomness: true,
+										},
+									},
+								},
+							},
+						},
+					},
+				}
+				_, err = testClients.GatewayClient.Write(gw, clients.WriteOpts{Ctx: ctx, OverwriteExisting: true})
```
sam-heilbron, line 179: We noticed that updating (patching) a resource occasionally can fail due to a potential bug in our codebase. Therefore, we introduced https://github.com/solo-io/gloo/blob/master/test/helpers/util.go#L20, which performs the patch and handles retries internally.

test/e2e/access_log_test.go
```diff
@@ -140,6 +142,52 @@ var _ = Describe("Access Log", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				Eventually(func(g Gomega) {
```
sam-heilbron, line 182: I think when using the nested Gomega matchers, we need to ensure that the outermost Eventually succeeds. I think we need to have `Eventually(....).Should(Succeed())`.

You may want to try this out to confirm my suspicion, but I fear that the following test will pass:
```
Eventually(func(g Gomega) {
  g.Expect(1).To(Equal(2))
})
```
whereas, the following update will make it fail correctly
```
Eventually(func(g Gomega) {
  g.Expect(1).To(Equal(2))
}).Should(Succeed())
```

test/e2e/access_log_test.go
```diff
@@ -140,6 +142,52 @@ var _ = Describe("Access Log", func() {
...
+					TestUpstreamReachable()
+
+					var entry *envoy_data_accesslog_v3.HTTPAccessLogEntry
+					g.Eventually(msgChan, 2*time.Second).Should(Receive(&entry))
+					g.Expect(entry.CommonProperties.UpstreamCluster).To(Equal(translator.UpstreamToClusterName(tu.Upstream.Metadata.Ref())))
+				}, time.Second*21, time.Second*2)
```
sam-heilbron, line 188: I assume the other tests use these numbers, but any indication what the magic of 21 seconds is?

sam-heilbron (commented):
Looking good! A couple of small comments about the tests, but appreciate having those e2e tests written.

Given this new functionality that we're exposing in the Gloo API, is it worth (perhaps in a follow up PR) outlining in our docs some valuable use cases of this filter?

projects/gloo/api/v1/options/als/als.proto
```diff
@@ -56,3 +62,207 @@ message GrpcService {
...
+    // reason about from a probability perspective (i.e., setting to true will
+    // cause the filter to behave like an independent random variable when
+    // composed within logical operator filters).
+    bool use_independent_randomness = 3;
+  }
+
+// Performs a logical “and” operation on the result of each filter in filters.
+// Filters are evaluated sequentially and if one of them returns false, the
+// filter returns false immediately.
+message AndFilter {
```
nfuden, line 175: Looking at the name here it raises a question: If the and filter has 0 sub filters how does envoy behave? Does it reject this configuration? If so we should consider adding a guard rail in the plugins/als/plugin.go

test/e2e/access_log_test.go
jmassasolo: Yes, Your assumption is correct, tested it out. I will add it.

test/e2e/access_log_test.go
jmassasolo: Not sure. I used the timeout and polling interval used in other access log tests.

projects/gloo/api/v1/options/als/als.proto
jmassasolo: It takes it. filters is just set to null and is ignored.

jbohanon (changes requested):
I don't think we're actually translating the new settings here. I think we need to implement a `copyFilterSettings` function in plugin.go to get the settings from our proto type to Envoy's and put them in the new config, then we need to update the tests to properly check for the filters config

test/e2e/access_log_test.go
sheidkamp: Test updated

jbohanon (commented):
Dismissing request change review

test/e2e/access_log_test.go
```diff
@@ -194,10 +198,166 @@ var _ = Describe("Access Log", func() {
 				}, time.Second*30, time.Second/2).Should(Succeed())
 			})
 
+			It("can create json access logs with multiple filters", func() {
+				gw := gwdefaults.DefaultGateway(writeNamespace)
```
sam-heilbron, line 202: We can talk more about this but there are 2 ways we can apply resources to our cache:
1. Use the resourcesToCreate object, which are written for us in the JustBeforeEach of the TestContext. For this to work however, you'll need to define those resources in a BeforeEach (https://github.com/solo-io/gloo/blob/master/test/e2e/example_test.go#L117)
2. Patch the resource directly within the test (https://github.com/solo-io/gloo/blob/master/test/e2e/example_test.go#L146)

test/e2e/access_log_test.go
```diff
@@ -194,10 +198,166 @@ var _ = Describe("Access Log", func() {
...
+			Eventually(func(g Gomega) {
+				req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("http://%s:%d/1", "localhost", defaults.HttpPort), nil)
+				g.Expect(err).NotTo(HaveOccurred())
+				req.Host = e2e.DefaultHost
+				g.Expect(http.DefaultClient.Do(req)).Should(matchers.HaveOkResponse())
+
+				// We can get a 404 by not setting the Host header.
+				req, err = http.NewRequest(http.MethodPost, fmt.Sprintf("http://%s:%d/BAD/HOST", "localhost", defaults.HttpPort), nil)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(http.DefaultClient.Do(req)).ShouldNot(matchers.HaveOkResponse())
```
sam-heilbron, line 350: Do you think it's worth it to make this assertion even more explicit, that we're expecting a 404 response instead of not a 200?

test/e2e/access_log_test.go
```diff
@@ -194,10 +198,166 @@ var _ = Describe("Access Log", func() {
...
+
+				logs, err := testContext.EnvoyInstance().Logs()
+				g.Expect(err).To(Not(HaveOccurred()))
+				g.Expect(logs).To(Not(ContainSubstring(`"POST /1 HTTP/1.1" 200`)))
+				g.Expect(logs).To(ContainSubstring(`"POST /BAD/HOST HTTP/1.1" 404`))
+
+			}, time.Second*10, time.Second/2).Should(Succeed()) // Todo, restore to 30
```
sam-heilbron, line 357: Just a reminder about the TODO here

test/e2e/access_log_test.go
```diff
@@ -194,10 +198,166 @@ var _ = Describe("Access Log", func() {
...
+					g.Expect(logs).To(ContainSubstring(`"method":"POST"`))
+					g.Expect(logs).To(ContainSubstring(`"protocol":"HTTP/1.1"`))
+				}, time.Second*30, time.Second/2).Should(Succeed())
+			})
 		})
 
 	})
 
+	Context("Test Filters (use string format)", func() {
+		BeforeEach(func() {
```
sam-heilbron, line 302: 👏

test/e2e/access_log_test.go
```diff
@@ -194,10 +198,166 @@ var _ = Describe("Access Log", func() {
...
+						},
+					},
+				},
+			}
+			testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+				gw,
+			}
+		})
+
+		It("can create string access logs", func() {
```
sam-heilbron, line 340: I like this test a lot. What do you think about getting a 200, validating the logs don't have the request included (because of the filter), then do what you did to get a 404. To me that will explicitly demonstrate that this filter is filtering out the non-404s correctly

projects/gloo/pkg/plugins/als/plugin.go
```diff
@@ -63,33 +67,133 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, parentListener *
...
 	return logCfg, nil
 }
 
+// Since we are using the same proto def, marshal out of gloo format and unmarshal into envoy format
+func translateFilter(inFilter *als.AccessLogFilter) (*envoyal.AccessLogFilter, error) {
+	if inFilter == nil {
+		return nil, nil
+	}
+
+	if err := validateFilterEnums(inFilter); err != nil {
```
sam-heilbron, line 121: nit: I remember we were surprised that the proto.Marshal wouldn't error with invalid enums. It might be helpful to call out in a comment why we validate the enums specifically and what can happen if we don't

projects/gloo/api/v1/options/als/als.proto
```diff
@@ -56,3 +63,207 @@ message GrpcService {
...
+    repeated AccessLogFilter filters = 1 [(validate.rules).repeated = {min_items: 2}];
+}
+  
+// Performs a logical “or” operation on the result of each individual filter.
+// Filters are evaluated sequentially and if one of them returns true, the
+// filter returns true immediately.
+message OrFilter {
+    repeated AccessLogFilter filters = 2 [(validate.rules).repeated = {min_items: 2}];
```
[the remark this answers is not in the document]
sheidkamp: Because that's how they do it here: https://github.com/envoyproxy/envoy/blob/317efa53c6add02fa9533fc8f93f10f590077dca/api/envoy/config/accesslog/v3/accesslog.proto#L24

But followup question... Do we need to explicitly acknowledge the reuse of the Envoy API code? It looks like these sections were copy and pasted.

projects/gloo/api/v1/options/als/als.proto
```diff
@@ -26,6 +30,9 @@ message AccessLog {
         // Send access logs to gRPC service
         GrpcService grpc_service = 3;
     }
+    // Filter which is used to determine if the access log needs to be written.
+    //.solo.io.envoy.config.core.v3.AccessLogFilter filter = 4;
```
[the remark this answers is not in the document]
sheidkamp: Removed

test/e2e/access_log_test.go
sheidkamp: I will take a closer look... this was an "inherited" test I didn't pay much attention to

test/e2e/access_log_test.go
sheidkamp: good call

test/e2e/access_log_test.go
sheidkamp: good catch

test/e2e/access_log_test.go
sheidkamp: I was putting the 200 call in the "Eventually" because I thought we needed to wait for the server to come up and accept requests. I put the 404 call after the 200 call so that if there was a delay in the logs becoming available, having the 404 in the log means that the earlier 200 should have been logged if it was going to be log.

I have updated this to a single 200 call before the eventually, but am also keeping the "not 200 in the logs" test in the eventually

projects/gloo/pkg/plugins/als/plugin.go
sheidkamp: comment added

test/e2e/access_log_test.go
sheidkamp: Some questions about this (for Tuesday) about the best way to approach this. There's a shared "before each", so modifying directly within the test seems appropriate, but the example is patching the virtual service, and I want to modify the Gateway which doesn't seem to have a patch function.

projects/gloo/api/v1/options/als/als.proto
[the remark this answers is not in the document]
jbohanon: I would think it would be good to at least include somewhere in the proto or the PR comment the release or commit where the file was copied from. Could save a lot of digging later on if/when we need to update to match upstream

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -1,15 +1,23 @@
 package als_test
 
 import (
+	envoyal "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
+	envoy_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	envoy_v32 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
```
jbohanon, line 6: rename to something more descriptive such as `envoy_routeconfig`

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -1,15 +1,23 @@
...
 	envoyalfile "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/file/v3"
 	envoy_extensions_filters_network_http_connection_manager_v3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	envoy_v31 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
```
jbohanon, line 9: rename to something more descriptive such as `envoy_types`

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -1,15 +1,23 @@
...
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	v31 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/config/core/v3"
```
jbohanon, line 14: rename to something more descriptive such as `gloo_envoy`

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -1,15 +1,23 @@
...
+	v32 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/config/route/v3"
```
jbohanon, line 15: rename to something more descriptive such as `gloo_envoy_routeconfig`

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -1,15 +1,23 @@
...
+	v3 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/type/v3"
```
jbohanon, line 16: rename to something more descriptive such as `gloo_envoy_types`

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -22,10 +30,456 @@ import (
...
+
+		DescribeTable("Test We Correctly Handle Bad ENUM",
+			func(glooInputFilter *accessLogService.AccessLogFilter, searchStr string) {
+
+				accessLog := alsSettings.GetAccessLog()[0]
+				accessLog.Filter = glooInputFilter
+
+				accessLogConfigs, err = ProcessAccessLogPlugins(alsSettings, nil)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(searchStr))
```
jbohanon, line 355: If you change the error pattern in the translation, you can use the matcher `MatchError` here which will recursively check for the desired error

projects/gloo/pkg/plugins/als/plugin.go
```diff
@@ -63,33 +67,136 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, parentListener *
...
+	return outFilter, nil
+}
+
+var (
+	InvalidEnumValueError = func(filterName string, fieldName string, value string) error {
+		return errors.Errorf("Invalid value of %s in Enum field %s of %s", value, fieldName, filterName)
+	}
+	NestedError = func(filterName string, err error) error {
+		return errors.Errorf("%s, inside an %s", err, filterName)
+	}
```
jbohanon, line 147: By doing this, we can check for error via `errors.Is(err, InvalidEnumValueError)` or in tests with `Expect(err).Should(MatchError(importname.InvalidEnumValueError))` instead of relying on raw string comparison
```suggestion
var (
        InvalidEnumValueError = errors.New("invalid value in Enum Field")
	WrapInvalidEnumValueError = func(filterName string, fieldName string, value string) error {
		return errors.Wrap(InvalidEnumValueError, "invalid value of %s in Enum field %s of %s", value, fieldName, filterName)
	}
```

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -22,10 +30,456 @@ import (
...
+								Statuses: []envoyal.GrpcStatusFilter_Status{1, 2},
+								Exclude:  false,
+							},
+						},
+					}),
+			)
+
+		})
+
+		DescribeTable("Test We Correctly Handle Bad ENUM",
```
jbohanon, line 347: nit: 
```suggestion
		DescribeTable("Test We Correctly Handle Bad Enum",
```

projects/gloo/pkg/plugins/als/plugin_test.go
```diff
@@ -227,20 +680,72 @@ var _ = Describe("Plugin", func() {
...
-										ServiceRef: &als.GrpcService_StaticClusterName{
+										ServiceRef: &accessLogService.GrpcService_StaticClusterName{
 											StaticClusterName: translatorutil.UpstreamToClusterName(usRef),
 										},
 										AdditionalRequestHeadersToLog:   extraHeaders,
 										AdditionalResponseHeadersToLog:  extraHeaders,
 										AdditionalResponseTrailersToLog: extraHeaders,
 									},
 								},
+								// TODO - hardcoded values
```
jbohanon, line 736: Is this TODO going to be addressed in this PR? If not, please put your name or handle on it and a little more description

test/e2e/access_log_test.go
```diff
@@ -194,8 +198,167 @@ var _ = Describe("Access Log", func() {
...
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(logs).To(ContainSubstring(`"method":"POST"`))
+					g.Expect(logs).To(ContainSubstring(`"protocol":"HTTP/1.1"`))
+				}, time.Second*30, time.Second/2).Should(Succeed())
+			})
+		})
+
+	})
+
+	Context("Test Filters (use string format)", func() {
```
jbohanon, line 301: Can we not nest this context in the `String Format` context beginning on L103?

jbohanon (commented):
Looks great so far!

jbohanon (approved):

projects/gloo/api/v1/options/als/als.proto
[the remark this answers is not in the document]
sheidkamp: Added comment at the top of the file with (perma)link

projects/gloo/pkg/plugins/als/plugin_test.go
sheidkamp: todo addressed, removed

test/e2e/access_log_test.go
sheidkamp: We could, but we're not really testing the string formatter - that's somewhat incidental. I think the filter testing deserves its own "top level" entry independent of the output format. Let me know what you think.

projects/gloo/pkg/plugins/als/plugin.go
```diff
@@ -63,33 +67,136 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, parentListener *
...
+
+		// Make the "base" config with output destination
 		switch cfgType := al.GetOutputDestination().(type) {
 		case *als.AccessLog_FileSink:
 			var cfg envoyalfile.FileAccessLog
-			if err := copyFileSettings(&cfg, cfgType); err != nil {
+			err := copyFileSettings(&cfg, cfgType)
+			if err != nil {
 				return nil, err
 			}
```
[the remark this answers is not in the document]
sheidkamp: I think this was from some debugging, reverting

projects/gloo/pkg/plugins/als/plugin.go
```diff
@@ -63,33 +67,136 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, parentListener *
...
+		results = append(results, &newAlsCfg)
+
 	}
+
 	logCfg = append(logCfg, results...)
 	return logCfg, nil
 }
 
+// Since we are using the same proto def, marshal out of gloo format and unmarshal into envoy format
+func translateFilter(inFilter *als.AccessLogFilter) (*envoyal.AccessLogFilter, error) {
```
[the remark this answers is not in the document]
sheidkamp: My gut is that the best answer is "use OO/interfaces and don't pass anything either way", but going with passing as the pointer to the struct to keep things consistent across the code, as the "pass a pointer as an argument" seems to be a very common pattern.

test/e2e/access_log_test.go
sheidkamp: Per our conversation removed this test.

test/e2e/access_log_test.go
sheidkamp: My question is "how valuable is this"... it looks like a substantial amount of work of get this going, to test what is really Envoy functionally (which is why I put in a smoke test instead of something comprehensive).

The biggest issue is going to be writing the matcher against the logs... we can't use a JSON matcher because we don't have JSON, we have the raw logs, and we can't use a string matcher because the JSON lines can come out in any order.

I'll take a swing at this this afternoon, but talking with Sam it seemed like his preference was to get this ticket closed (I'm out the rest of the week) than spending too much time on testing the actual functions of the filters

jbohanon: I guess I would just argue that if we need the strings formatter tests at all, they should be in the appropriate context since we put the json tests in the json context. If we don't need them, I'm fine with that also

sheidkamp: I've updated the test names and added a comment to make clear what we're testing there... does this work?

jbohanon: Did you push these changes? I don't see these changes or any of the other conversations you've resolved.

jbohanon (changes requested):
Previous review has not been addressed

jbohanon (changes requested):
Previous review has not been addressed

test/e2e/access_log_test.go
sheidkamp: They are showing up now.

sam-heilbron (approved):

sam-heilbron (approved):

sam-heilbron (approved):

sam-heilbron (approved):

jbohanon (approved):

jbohanon:
closed in favor of https://github.com/solo-io/gloo/pull/7627 due to administrative difficulties with this PR.

PR #7508, opened 2022-12-05
7406 trace context support

ashishb-solo:
# Description

Implement the OpenTelemetry envoy plugin in our tracing plugin.

OpenTelemetry plugin API: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/opentelemetry.proto

# Context

Enable W3C-compliant tracing headers

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7406

PR #7521, opened 2022-12-07
7406 trace context support (#7508)

ashishb-solo:
* Cherry-pick c511b2f6 onto master branch

Initial commit of tracing plugin changes

At this point, the plugin should in theory have all translation implemented. However, there is no error checking present in the plugin. A basic unit test exists for this work, but all it does it ensure that no errors occurred. The unit test also does not yet validate the content of the output object.

* Only set GrpcService if OcagentAddress is empty

This was creating an error in Envoy - if the GrpcService object being sent to Envoy had an empty string, then it would not accept the configuration.

* Final version of opencensus changes

This commit includes a translation fix and simplification of the API. Originally, we were outputting an object of type GrpcService_EnvoyGrpc_ for the grpcAddress option, but the Envoy plugin only supports the GrpcService_GoogleGrpc_ type. So we fix that here. We also simplify the protobuf API to use a oneof for specifying HTTP or GRPC.

I'm committing this now because we have decided to drop using OpenCensus due to the fact that it cannot be reconfigured. A lot of the work on this effort will be discarded, so I am saving it here for posterity.

* Rename all strings in opencensus.proto to opentelemetry

* Rename opencensus.proto to opentelemetry.proto

* Rewrite everything to use opentelemetry instead

So far, I have only tested manually, but it seems to work okay. One odd behaviour is that Envoy only seems to report information to the echo server on shutdown.

Unit tests aren't doing anything serious at the moment and need to be improved. It will also be necessary to write an e2e test for this. But otherwise, things seem to be working much more smoothly than they did with opencensus.

* Replace "Census" with "Telemetry" in function name

* Remove autogenerated opencensus documentation file

* Update unit test

* Add changelog

* Add e2e test for opentelemetry

* Update docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/external/envoy/config/trace/v3/opentelemetry.proto.sk.md

Co-authored-by: Art <artberger@users.noreply.github.com>

* Incorporate some changes from Sam

* Remove race conditions from e2e tests

The major issue here was a pair of variables with near-global scope: collectorApiHit and collectionURLPath. Both of these were triggering go's race condition flags, probably because they were being written to and read from ginkgo's setup and teardown callbacks.

There was also an issue in shutting down the collection server, which is addressed by a comment around where the change was made.

* Some comments and docstring fixes

* Add default statements in tracing plugin type translation

* Changelog improvements

Co-authored-by: soloio-bulldozer[bot] <48420018+soloio-bulldozer[bot]@users.noreply.github.com>
Co-authored-by: Art <artberger@users.noreply.github.com>

# Description

# Context

# Checklist:


nfuden:
/kick  Aggregate Listener Secure HybridGateway (Matched) IsolateVirtualHostsBySslConfig = true [It] produces a Proxy with a single AggregateListener

nfuden (approved):

sam-heilbron (approved):

Issue #4204, opened 2021-01-27
glooctl check configurable timeouts


law10k:
**Is your feature request related to a problem? Please describe.**
Right now `glooctl check` has hardocded timeouts that may not be long for slow network connections and large results.
This should be user configurable.

**Describe the solution you'd like**
An env var or `glooctl.config` option to tuning the read timeouts of the various `glooctl check` calls

**Describe alternatives you've considered**
N/A (?)

**Additional context**

PR #7523, opened 2022-12-07
`glooctl check --check-connection-timeout <string>`

Bslabe123:
# Description

Added  `--check-connection-timeout` flag to `glooctl check` to adjust the cluster connection check timeout.

# Context

Prior to this change we hardcoded the timeout to five seconds which was not sufficient for large clusters given the mechanics of the check. To do this check, we create a settings client and get settings in the cluster and assure that an error has not occurred. The check would ideally be O(1) but in reality is O(#settings in cluster).

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/4204


changelog/v1.14.0-beta2/glooctl-check-timeout.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    description: Added  --timeout flag to `glooctl check` to adjust the cluster connection check timeout.
```
sam-heilbron, line 3: 👏 love the clarity of this changelog!

docs/content/reference/cli/glooctl_check.md
```diff
@@ -24,6 +24,7 @@ glooctl check [flags]
   -p, --pod-selector string               Label selector for pod scanning (default "gloo")
       --read-only                         only do checks that dont require creating resources (i.e. port forwards)
   -r, --resource-namespaces stringArray   Namespaces in which to scan gloo custom resources. If not provided, all watched namespaces (as specified in settings) will be scanned.
+      --timeout float                     cluster connection check timeout (default 5)
```
sam-heilbron, line 27: Sometimes I find timeouts confusing because it's unclear what units the timeout uses, and what it applies to. Can we clarify those two points in our docs so users know when to use this and what impact it has?

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -929,12 +930,12 @@ func renderNamespaceName(namespace, name string) string {
 
 // Checks whether the cluster that the kubeconfig points at is available
 // The timeout for the kubernetes client is set to a low value to notify the user of the failure
-func checkConnection(ctx context.Context, ns string) error {
-	client, err := helpers.GetKubernetesClientWithTimeout(5 * time.Second)
+func checkConnection(ctx context.Context, opts *options.Options) error {
+	client, err := helpers.GetKubernetesClientWithTimeout(time.Duration(int(opts.Check.Timeout * float64(time.Second.Nanoseconds()))))
```
sam-heilbron, line 934: We're converting this to a duration. Any reason we wouldn't just make the timeout variable a string that we unmarshal into a duration. So the user could supply "5s" or "1m" depending on their needs?

projects/gloo/cli/pkg/cmd/options/options.go
```diff
@@ -452,4 +452,7 @@ type Register struct {
 type Check struct {
 	// The maximum length of time to wait before giving up on a secret request. A value of zero means no timeout.
 	SecretClientTimeout time.Duration
+
+	// Timeout for connecting to the cluster
+	Timeout float64
```
nfuden, line 457: It looks like this is not the only timeout.
Have we looked at secretclient timeout and how that is set and how it may or may not interact with this?

It seems confusing to me that we have this timeout that seems to set a default timeout and then a seperate secretclienttimeout. 
What happens if both are set and secret client timeout is higher than this timeout?

Should this be configurable in all the locations that secretclient timeout is configurable?

nfuden (changes requested):
The general idea seems good, however I am concerned with how this diverges from the secretclienttimeout and makes no mention of it.

While I am comfortable with taking a different approach it feels like the prior art was not considered in this case and would want to understand why we are taking the different approach as well as how the two play together.

projects/gloo/cli/pkg/cmd/check/root.go
Bslabe123: That would be more intuitive! Added

Bslabe123:
See: https://github.com/solo-io/gloo/pull/7535

Issue #6771, opened 2022-07-20
expose Envoy route option host_rewrite_path_regex


bcollard:
### Is your feature request related to a problem? Please describe.

Expose Envoy route option `host_rewrite_path_regex`
It allows to dynamically replace the Host value with some part of the path. 

This is a demand from the community.

PR #7566, opened 2022-12-19
6771 expose host rewrite path regex

sheidkamp:
# Description

This ticket exposes Envoy route option `host_rewrite_path_regex`

Manual e2e test:

- Build the gloo docker package: `VERSION=some-version make gloo-docker -B`
- Run Gloo locally backed by a filesystem: https://docs.solo.io/gloo-edge/latest/installation/gateway/development/docker-compose-file/
  - Include the GLOO_VERSION in the `up` command: `GLOO_VERSION=some-version docker-compose up`

Update install/docker-compose-file/data/config/virtualservices/gloo-system/default.yaml
by adding a rewrite rule to the `options` section of a route:

```
      hostRewritePathRegex:
        pattern:
          googleRe2: {}
          regex: ^/(.+)/.+$
        substitution: \1
```

and validate you can see in the config dump
# Context

This option was not exposed. We are now exposing it.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6771


projects/gloo/pkg/plugins/basicroute/plugin_test.go
```diff
@@ -545,6 +545,45 @@ var _ = Describe("host rewrite", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(routeAction.GetAutoHostRewrite().GetValue()).To(Equal(true))
 	})
+
+	It("rewrites using regex", func() {
```
sam-heilbron, line 549: This is a great positive test case to verify that when provided a valid regex, we can convert it and place it on the Envoy route. Could we also add a negative test case that verifies when an invalid regex is provided we handle the error cleanly?

projects/gloo/api/v1/options.proto
```diff
@@ -336,6 +336,10 @@ message RouteOptions {
         // This setting is only honored for upstreams that use DNS resolution (i.e., their generated Envoy cluster is
         // of type STRICT_DNS or LOGICAL_DNS -- think aws, azure, or static upstreams with hostnames)
         google.protobuf.BoolValue auto_host_rewrite = 19;
+
+        // Indicates that during forwarding, the host header will be swapped with the result of the regex 
+        // substitution executed on path value with query and fragment removed.
+        .solo.io.envoy.type.matcher.v3.RegexMatchAndSubstitute host_rewrite_path_regex = 146;
```
sam-heilbron, line 342: Any particular reason for this field number?

changelog/v1.14.0-beta2/host_rewrite_header_regex.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/6771
+    description: Expose Envoy route option host_rewrite_path_regex
```
sam-heilbron, line 4: nit: Since users rely on these changelogs, we try to make them as informative as possible. Perhaps include why a user would use this field (basically grabbing the comment on the proto definition)

sam-heilbron (commented):
Looks great!

Can we also include some e2e testing for this feature? It could either be an automated in memory e2e test that verifies when we configure the Gloo API, we can query the Envoy Admin Endpoint and confirm that config_dump includes this field. Or a manual test with steps outlined on the PR for a reviewer to follow

projects/gloo/api/v1/options.proto
sheidkamp: The highest number I found in the file was 145:

`ratelimit.options.gloo.solo.io.RateLimitConfigRefs rate_limit_regular_configs = 145;`

So I went one higher

projects/gloo/pkg/plugins/basicroute/plugin_test.go
sheidkamp: Assuming you mean an invalid regex string, like "\\" as the matcher, then the negative tests pass (passing in the wrong type of object makes the compiler cranky and just won't run). Looks like they never get compiled/executed, just have their strings pulled out to recast from Gloo format to Envoy format. I don't see a validation function in the regex utils that we could apply in the regex utils (https://github.com/solo-io/gloo/blob/master/pkg/utils/regexutils/regex.go)

Do we have an existing function that would validate the regex beyond the basic .proto format?

changelog/v1.14.0-beta2/host_rewrite_header_regex.yaml
sheidkamp: updated

projects/gloo/api/external/envoy/api/v2/route/route.proto
```diff
@@ -537,7 +537,7 @@ message RouteAction {
         // downstream or `custom (config_http_conn_man_headers_custom_request_headers)` header.
         // If header value is empty, host header is left intact.
         //
-                //
+        //
```
ianmacclancy, line 540: Nit: This file does not need a modification in this PR

ianmacclancy (commented):
Looks good - will wait for you and Sam to resolve conversations and then ping me for an approval

projects/gloo/api/external/envoy/api/v2/route/route.proto
sheidkamp: What's our general approach in this situation? Revert and leave the bad formatting because the file isn't relevant, or keep the non-functional formatting fix ?

sam-heilbron: I'm in favor of a "see it, fix it" style of approach. Other changes which can cause noise we may want to do in a separate PR, but this doesn't cause noise I don't think

projects/gloo/api/v1/options.proto
sam-heilbron: ah yeah seems like we jump around a bit in this file. Probably fine to leave as is

projects/gloo/pkg/plugins/basicroute/plugin_test.go
sam-heilbron: Oh sorry I meant how can we trigger the case where "ConvertRegexMatchAndSubstitute" returns an error. But it seems like that isn't possible

projects/gloo/api/v1/options.proto
sheidkamp: Looking a little closer at the numbers in that file, it looks like the 141-145 numbers should be read as 14.1-14.5, like sub-bullets. To make the number on this a little less random, I changed it to 10.1 = 10. I don't love it, but I dislike it less.

ianmacclancy (dismissed):

projects/gloo/pkg/plugins/basicroute/plugin_test.go
sheidkamp: Found a way to do it, half for the practice. I don't know if we have a standard practice/library we use, but my monkeypatch solution is in the newest commit.

sam-heilbron (approved):
I followed the steps laid out in the PR: built the images, ran gloo backed by a file system using the images I built, modified the virtual service config, and confirmed that the config dump at the admin endpoint contained the processed settings.

sam-heilbron (approved):
Everything was passing before changelog-bot pushed an update

ianmacclancy (approved):

PR #7570, opened 2022-12-20
Extra Containers and InitContainers

jackstine:
# Description

Including helm properties to include InitContainers and Containers to the gateway proxy deployment.
By applying the following properties to the helm chart `gatewayProxies.gatewayProxy.podTemplate.extraInitContainers` and `gatewayProxies.gatewayProxy.podTemplate.extraContainers` initContainers and containers, respectively, can be attached to the gateway proxy deployment.

Initcontainers has also been added to PodSpecs.

# Context

Users were trying to use `extraInitContainersHelper` which used as a call to a template in helm. What they really need is a property to allow them to add new containers.

# Checklist:


changelog/v1.14.0-beta2/extra-helm-containers.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NEW_FEATURE
```
kdorosh, line 2: we have a `HELM` changelog type fyi, some customers in the past asked us to separate helm changes / breaks from the core product

sam-heilbron (commented):
Spoke offline, plan is to:
1. Add testing to verify the template changes
2. Add support for initContainers on other deployments (since this only adds support for gateway proxy) since our hope is it will be a quick win while we're in the space and maintain parity across the deployments

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -159,11 +159,19 @@ spec:
...
       terminationGracePeriodSeconds: {{ $spec.podTemplate.terminationGracePeriodSeconds }}
       {{- end }}
-      {{- if $spec.extraInitContainersHelper }}
+      {{- if (or $spec.podTemplate.extraInitContainers $spec.extraInitContainersHelper) }}
       initContainers:
+      {{- if $spec.podTemplate.extraInitContainers }}
+{{ toYaml $spec.podTemplate.extraInitContainers | indent 6 }}
+      {{- end }}
+      {{- if $spec.extraInitContainersHelper }}
       {{- include $spec.extraInitContainersHelper . | nindent 6 }}
```
ben-taussig-solo, line 168: does this need to be wrapped in an `{{- if ... }}` block? or does the `extraInitContainersHelper` necessarily exist if `extraInitContainers` exists?

install/helm/gloo/generate/values.go
```diff
@@ -492,6 +493,8 @@ type GatewayProxyPodTemplate struct {
 	CustomReadinessProbe          *appsv1.Probe         `json:"customReadinessProbe,omitempty"`
 	CustomLivenessProbe           *appsv1.Probe         `json:"customLivenessProbe,omitempty"`
 	ExtraGatewayProxyLabels       map[string]string     `json:"extraGatewayProxyLabels,omitempty" desc:"Optional extra key-value pairs to add to the spec.template.metadata.labels data of the gloo edge gateway-proxy deployment."`
+	ExtraContainers               []interface{}         `json:"extraContainers,omitempty" desc:"Extra [containers](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers) to be added to the slice of containers on the gateway proxy deployment."`
```
ben-taussig-solo, line 496: Super nit: Is slice the language the K8s uses for lists/arrays/collections of containers? This is really not a super important point but it might be good for clarity to use the language that K8s users.

[Here](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers) it appears that they use the term "array" -- how would you feel about changing this line and the next one to use "array" instead of "slice"?

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
jackstine: not sure what you mean, these are two different values.

jackstine: one of them can exist without the other

install/helm/gloo/generate/values.go
jackstine: cool.  Yeah Helm uses slice

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
ben-taussig-solo: On 162 we do `{{- if (or $spec.podTemplate.extraInitContainers $spec.extraInitContainersHelper) }}`

I'm concerned about the case where `$spec.podTemplate.extraInitContainers` is set and `$spec.extraInitContainersHelper` is not -- in that case, should we wrap `{{- include $spec.extraInitContainersHelper . | nindent 6 }}` on line 168 in an `if` block?

ben-taussig-solo: Oh wait there already is an if block.... sorry
resolving this thread

ben-taussig-solo (approved):
Tested e2e with the following steps and everything looks good!

# Testing steps

1. Create kind cluster
   - `kind create cluster --name kind`
2. Prepare helm chart
   - `make clean` to delete all previously created assets
   - `TAGGED_VERSION=v1.14.0-beta1 make gneerate-helm-files`
     - I am using an already existing tag so that I won't have to recreate the docker images. This is not necessary, but it saves time.
3. Prepare Helm values file:
   - ```yaml
     gatewayProxies:
     gatewayProxy:
         podTemplate:
         extraInitContainers:
             - name: init
             image: busybox
             command: ["sh", "-c", "echo 'hello world'"]
        extraContainers:
        # make a long running container because we get a crash loop backoff if the container exits
            - name: long-running
            image: busybox
            command: ["sh", "-c", "while true; do echo 'hello world'; sleep 10; done"]
     ```
   - Save this file as `values.yaml`
4. Install Gloo OSS using newly created helm chart
   - `helm install -n gloo-system --create-namespace gloo install/helm/gloo -f debug_testing/values.yaml`
5. Get all running containers from the `gateway-proxy` deployment and confirm that the long-running container is running
   - `kubectl get pods -n gloo-system -l gloo=gateway-proxy -o jsonpath='{.items[0].status.containerStatuses[*].name}'`
   - `kubectl get pods -n gloo-system -l gloo=gateway-proxy -o jsonpath='{.items[0].status.containerStatuses[*].state}'`
6. Get the completed init container
   - `kubectl get pods -n gloo-system -l gloo=gateway-proxy -o jsonpath='{.items[0].status.initContainerStatuses[0].name}'`
   - `kubectl get pods -n gloo-system -l gloo=gateway-proxy -o jsonpath='{.items[0].status.initContainerStatuses[0].state}'`

ashishb-solo (approved):

PR #7580, opened 2022-12-21
k8s 1.24 ci updates [1.13]

jenshu:
more k8s 1.24 updates:
- ci
- docs
- kubectl image

### DIFF
--- a/jobs/kubectl/Dockerfile
+++ b/jobs/kubectl/Dockerfile
@@ -1,4 +1,4 @@
-FROM bitnami/kubectl:1.22.12 as kubectl
+FROM bitnami/kubectl:1.24.7 as kubectl
 
 FROM alpine:3.15.6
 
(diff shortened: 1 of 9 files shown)


jobs/kubectl/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM bitnami/kubectl:1.22.12 as kubectl
+FROM bitnami/kubectl:1.24.7 as kubectl
```
sam-heilbron, line 1: Do we want this version to match our client libs or are server version? I had imagined we would keep this in line with our client libraries

nfuden (approved):

jobs/kubectl/Dockerfile
nfuden: I am unsure on this but we arent bumping k8s due to go dependency bump but at the same time if being deployed in ci as 1.24 env it seems to make sense to be deployed as that?

jenshu: spoke offline; we decided to keep this change so it matches the server version we are testing against

elcasteel (approved):

Issue #7410, opened 2022-11-09
e2e flake in aggregate_listener_test.go


ashishb-solo:
### Gloo Edge Version

1.13.x (beta)

### Describe the bug

Observed a flake in this test:

```
Step #10 - "test": • Failure [2.028 seconds]
Step #10 - "test": Aggregate Listener Secure HybridGateway (Matched) IsolateVirtualHostsBySslConfig = false [It] produces a Proxy with a single HybridListener 
Step #10 - "test": /workspace/gloo/test/e2e/aggregate_listener_test.go:549
Step #10 - "test": 
Step #10 - "test":   Expected
Step #10 - "test":       <[]*v1.Listener | len:0, cap:0>: nil
Step #10 - "test":   to have length 1
Step #10 - "test": 
Step #10 - "test":   /workspace/gloo/test/e2e/aggregate_listener_test.go:553
Step #10 - "test": 
Step #10 - "test":   Full Stack Trace
Step #10 - "test":   github.com/solo-io/gloo/test/e2e_test.glob..func2.8.3.2()
Step #10 - "test":      /workspace/gloo/test/e2e/aggregate_listener_test.go:553 +0x192
Step #10 - "test":   github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc0007976e0)
Step #10 - "test":      /go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/leafnodes/runner.go:113 +0xf6
Step #10 - "test":   github.com/onsi/ginkgo/internal/leafnodes.(*runner).run(0xc0007976e0)
Step #10 - "test":      /go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/leafnodes/runner.go:64 +0x158
Step #10 - "test":   github.com/onsi/ginkgo/internal/leafnodes.(*ItNode).Run(0xc0006b4b20)
Step #10 - "test":      /go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/leafnodes/it_node.go:26 +0x95
Step #10 - "test":   github.com/onsi/ginkgo/internal/spec.(*Spec).runSample(0xc00032d860, 0x0?, {0x713f500, 0xc000384640})
Step #10 - "test":      /go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/spec/spec.go:215 +0x35a
[...]

PR #7583, opened 2022-12-22
Fix destinationOverrides not being applied to empty aws destination

inFocus7:
# Description

This fixes an issue with `destinationOverrides` not being set if the destination was unset for aws routes.
Adds tests to check the outroute has the applied overrides, and moves tests with "no destinationSpec" to own context.

# Context

While testing more upstream & virtual service routing edge cases I fell upon this.

# Testing

1. Install Gloo with custom `values.yaml` to disable lambda discovery & with fallbackToFirstFunction enabled: `discovery.fdsMode: DISABLED; settings.aws.fallbackToFirstFunction: true`
2. Create an AWS secret 
3. Create an Upstream with a lambda function manually added to it and a `destinationOverride` set (ex. `destinationOverrides.requestTransformation: true`)
4. Create a Virtual Service pointing to upstream, **without** a `destinationSpec set`
5. Port-forward gloo-proxy to 8080
6. `curl -X POST -d '{"foo": "hello world"}' localhost:8080/`

Without this fix you'll see the request go through, but the `destinationOverride` will not take effect.
With this fix, you'll see the request go through as well as the `destinationOverride` taking effect.

# Checklist:

### DIFF
--- a/projects/gloo/pkg/plugins/aws/plugin.go
+++ b/projects/gloo/pkg/plugins/aws/plugin.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"unicode/utf8"
 
+	"github.com/hashicorp/go-multierror"
+
 	envoy_config_cluster_v3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_config_route_v3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
@@ -14,7 +16,6 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/golang/protobuf/ptypes/wrappers"
-	"github.com/hashicorp/go-multierror"
 	"github.com/imdario/mergo"
 	. "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/aws"
 	envoy_transform "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/transformation"
@@ -164,11 +165,20 @@ func (p *Plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 func (p *Plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
 	err := pluginutils.MarkPerFilterConfig(params.Ctx, params.Snapshot, in, out, FilterName,
 		func(spec *v1.Destination) (proto.Message, error) {
+			dest := spec.GetDestinationSpec()
 			// check if it's aws destination
-			if spec.GetDestinationSpec() == nil {
-				return nil, nil
+			// it is incorrect to set lambda functionality on routes that do not have a lambda function specified unless we fallback
+			if dest == nil {
+				if p.settings.GetFallbackToFirstFunction().GetValue() == false {
+					return nil, nil
+				}
+				// if we fallback to first, we need to have a valid destination
+				contextutils.LoggerFrom(params.Ctx).Debug("no destinationSpec set with fallbackToFirstFunction enabled, processing as aws route")
+				dest = &v1.DestinationSpec{
+					DestinationType: &v1.DestinationSpec_Aws{},
+				}
 			}
-			awsDestinationSpec, ok := spec.GetDestinationSpec().GetDestinationType().(*v1.DestinationSpec_Aws)
+			awsDestinationSpec, ok := dest.GetDestinationType().(*v1.DestinationSpec_Aws)
 			if !ok {
 				return nil, nil
 			}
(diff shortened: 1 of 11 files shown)


projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -175,7 +175,9 @@ func (p *Plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 				// if we fallback to first, we need to have a valid destination
 				contextutils.LoggerFrom(params.Ctx).Debug("no destinationSpec set with fallbackToFirstFunction enabled, processing as aws route")
 				dest = &v1.DestinationSpec{
-					DestinationType: &v1.DestinationSpec_Aws{},
+					DestinationType: &v1.DestinationSpec_Aws{
+						Aws: &aws.DestinationSpec{},
```
inFocus7, line 179: Adding this `Aws`, so that when it's passed to `perRouteConfigGenerator`, it isn't `nil` and the destinationOverrides get applied to it.

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -199,7 +201,15 @@ func (p *Plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 				return nil, err
 			}
 			// should be aws upstream
-			return p.perRouteConfigGenerator(p.settings, awsDestinationSpec.Aws, lambdaSpec)
+			awsLambda, err := p.perRouteConfigGenerator(p.settings, awsDestinationSpec.Aws, lambdaSpec)
+			if err == nil {
+				// set the `destinationSpec` to be `awsDestinationSpec` after being modified in `GenerateAWSLambdaRouteConfig`
+				// doing it this way allows us to keep the original destinationSpec with the updated awsDestinationSpec added onto it
+				dest.DestinationType = awsDestinationSpec
+				spec.DestinationSpec = dest
```
inFocus7, line 209: Setting the `spec`s `DestinationSpec` to the updated values so it still takes effect. 

TODO: think about logic. when `destinationSpec` _was_ set, the destinationOverrides were set correctly without needing to do this. So maybe (some of) this logic isn't necessary?

inFocus7: Did not need to set `dest.DestinationType = awsDestinationSpec`, as it's modified when passed to `perRouteConfigGenerator`

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -199,7 +201,13 @@ func (p *Plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 				return nil, err
 			}
 			// should be aws upstream
-			return p.perRouteConfigGenerator(p.settings, awsDestinationSpec.Aws, lambdaSpec)
+			awsLambda, err := p.perRouteConfigGenerator(p.settings, awsDestinationSpec.Aws, lambdaSpec)
+			if err == nil {
+				// set the `destinationSpec` to be the original destinationSpec with the updated awsDestinationSpec added onto it
+				spec.DestinationSpec = dest
```
nfuden, line 207: I like that this solves the issue we have been seeing but in general side affects of this manner are a bit scary.
It seems that this is correct given what we talked about above where we know that this is definitely an aws destination (as the upstream is aws) but can we call this out a bit more loudly in the function signature or on the plugin?

inFocus7: Not sure if you're also content with the updated comment you left. I am, so if you are as well we can resolve this, else I can add some more to it!

inFocus7:
/kick flake #7410 [Details](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=928c3782-5478-44bd-b982-2ccbe8e6288f)

projects/gloo/pkg/plugins/aws/plugin.go
```diff
@@ -165,40 +165,60 @@ func (p *Plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+			// local variable to avoid side effects for calls that are not to aws upstreams
 			dest := spec.GetDestinationSpec()
-			// check if it's aws destination
-			// it is incorrect to set lambda functionality on routes that do not have a lambda function specified unless we fallback
-			if dest == nil {
-				if p.settings.GetFallbackToFirstFunction().GetValue() == false {
-					return nil, nil
-				}
-				// if we fallback to first, we need to have a valid destination
+			tryingNonExplicitAWSDest := spec.GetDestinationSpec() == nil && p.settings.GetFallbackToFirstFunction().GetValue()
```
jenshu, line 170: when fallback is true, are we validating anywhere that there exists only one lambda function? if there's more than one, should we error?

inFocus7: I don't believe we check that there's only one lambda fn. The fallback falls back to first lambda function found on upstream, so it doesn't necessarily matter if there's more than one, as long as there is 1 to fall back on.

relevant code
- [ensuring there is at least 1 function](https://github.com/solo-io/gloo/blob/c16e282531adba0ae40558c227dc875719327e3e/projects/gloo/pkg/plugins/aws/plugin.go#L367-L369)
- [using the first function found as fallback](https://github.com/solo-io/gloo/blob/c16e282531adba0ae40558c227dc875719327e3e/projects/gloo/pkg/plugins/aws/plugin.go#L380-L389)

jenshu, line 170:
```suggestion
			tryingNonExplicitAWSDest := dest == nil && p.settings.GetFallbackToFirstFunction().GetValue()
```

jenshu (dismissed):

inFocus7:
/kick flake #6108 [Details](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=91dd2f75-fa56-4015-893f-d71ad91a71fa)

nfuden (approved):

PR #7594, opened 2022-12-23
`glooctl check-crds --version <string> --local-chart <string> --show-yaml <bool>`

Bslabe123:
# Description

When upgrading Gloo, we don’t have any automated processes to update our CRDs. However, the install will still be considered complete even with outdated CRDs. This presents an issue when we try to create custom resources with outdated resource definitions yielding non descriptive errors. We would like a way to check that deployed CRDs are of the same version of Gloo that is deployed.

# Context

We would like to create a new glooctl command (possibly “glooctl check-crd --version <string> —local-chart <string>”). This command should get all CRDs in the current cluster and compare them against the latest published (or `—version`s) helm chart for gloo OSS. If `—local-chart` is set then the helm chart pulled from `—local-chart` will be used in place of the one mentioned above. The CRDs from the helm charts are preprocessed to add the defaults that would have been seen if the chart was deployed (i.e. CustomResourceDefinition.Spec.Names.Singular is set to the lower case version of CustomResourceDefinition.Spec.Names.Kind, CustomResourceDefinitionCustomResourceConversion.Strategy has a default value that gets set) to avoid comparison errors. The name of each resource that doesn't exactly match (including missing resources) the one found in the helm chart will be reported.

# Testing
Two new tests added in `test/kube2e/glooctl/glooctl_test.go`, one successful case and one failure case.

# Caveats

Due to k8s preprocessing of yaml before deploying, there may be some unanticipated and/or inconsequential differences between deployed and accepted CRDs (i.e. ordering of keys differing). Should a difference due to preprocessing be suspected it is heavily suggested to use the `--show-yaml` flag and diff the `yaml`s manually

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/4193


Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=488cc78f-c3a1-4ad8-b22a-f15a8bc61d88

test/kube2e/glooctl/glooctl_test.go
```diff
@@ -507,6 +507,18 @@ var _ = Describe("Kube2e: glooctl", func() {
 			Expect(err.Error()).To(ContainSubstring("No namespaces specified are currently being watched (defaulting to 'gloo-system' namespace)"))
 		})
 	})
+	Context("check-crds", func() {
+		It("validates correct CRDs", func() {
+			chartUri := filepath.Join(testHelper.RootDir, testHelper.TestAssetDir, testHelper.HelmChartName+"-"+testHelper.ChartVersion()+".tgz")
+			_, err := runGlooctlCommand("check-crds", "--local-chart", chartUri)
```
elcasteel, line 513: This check should also handle the nightly test case when we're using a released version. We can check to see if `testHelper.ReleasedVersion` is set and if it is, run `glooctl check-crds --version testHelper.ReleasedVersion`

projects/gloo/cli/pkg/cmd/check-crds/root.go
```diff
@@ -0,0 +1,164 @@
...
+			diffs = append(diffs, crd.Name)
+		} else {
+			acceptedCrdBytes, _ := yaml.Marshal(acceptedCrd.Spec)
+			if string(clusterCrdBytes) != string(acceptedCrdBytes) {
+				diffs = append(diffs, crd.Name)
+			}
+		}
+	}
+	if len(diffs) != 0 {
+		return eris.New("Diffs detected on the following CRDs:\n\t" + strings.Join(diffs, "\n\t"))
```
elcasteel, line 89: Consider adding something to this message to help users figure out how to fix the issue (although it might be more helpful to have that as part of the help message, I'm not sure)

projects/gloo/cli/pkg/cmd/check-crds/root.go
```diff
@@ -0,0 +1,164 @@
...
+	acceptedCRDs, err := getCRDsFromHelm(chartPath)
+	if err != nil {
+		return eris.Wrapf(err, "Error getting CRDs from %s", chartPath)
+	}
+	clusterCRDs, err := getCRDsInCluster()
+	if err != nil {
+		return eris.Wrapf(err, "Error getting CRDs in current cluster")
+	}
+
+	lookupTable := map[string]apiextv1.CustomResourceDefinition{}
```
elcasteel, line 71: nit: make the naming on this lookup table a little clearer to show that it represents the specs of the crds in the chart

projects/gloo/cli/pkg/cmd/check-crds/root.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, crd := range acceptedCRDs {
+		lookupTable[crd.Name] = crd
+	}
+
+	diffs := []string{}
+	for _, crd := range clusterCRDs {
```
elcasteel, line 77: Will this catch a situation where there is a CRD in the chart that doesn't exist at all in the cluster?

projects/gloo/cli/pkg/cmd/check-crds/root.go
```diff
@@ -0,0 +1,164 @@
...
+		clusterCrdBytes, _ := yaml.Marshal(crd.Spec)
```
elcasteel, line 78: Are there situations where yaml.Marshall might put fields in different orders to make it look like equivalent specs are different or make a diff look bigger than it actually is?

projects/gloo/cli/pkg/cmd/check-crds/root.go
Bslabe123: Fixing by changing the set of CRDs in the map to chart CRDs, made the name more descriptive as per comment above

projects/gloo/cli/pkg/cmd/check-crds/root.go
Bslabe123: since we marshall the same data type (`apiextv1.CustomResourceDefinitionSpec`) with the same function (`yaml.Marshal`) would it be safe to assume that the ordeing will be identical for both?

elcasteel: I'm not sure. We don't necessarily add the fields to the specs in the exact same order and I don't know if the marshaller makes any guarantees about the outputs.

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=70bd37b3-3d9a-4e99-beef-b2decda74807

elcasteel (approved):

projects/gloo/cli/pkg/cmd/check-crds/root.go
elcasteel: We decided to warn users about the possibility of false positives.

nfuden (approved):

PR #7555, opened 2022-12-16
In Memory E2E Test Enhancements

sam-heilbron:
# Description

A set of changes to allow tests to run locally, or clarify which ones cannot be run locally

# Context

### Http Matcher (http.go)
We frequently verify that an HTTP request returned a status code and a body. Instead of replicating code throughout the codebase, I created a simple gomega Matcher, composed of 2 existing matchers.

This matcher is quite powerful, but to make it easier to use, I exposed some utility methods (thanks @ianmacclancy for the suggestion) that are frequently used (200 response with empty body, some status code response..etc).

There are more places in the code where we can use this matcher that I'll try to handle in a follow up. But this sets the standard for a single matcher we can use whenever asserting http responses have particular properties.

Signed off by:
- [ ] @ashishb-solo 
- [ ] @ianmacclancy 
- [ ] @jackstine 

### Test Requirements (requirements.go)
Certain tests require environmental conditions (like only running on linux). I introduced a mechanism to define these. By explicitly defining these, it is easier for developers to understand why tests are failing locally for them. Before, it was unclear that certain factors needed to be met, and we would spend hours trying to debug a test.

With requirements defined, we have the ability to notify ginkgo to either skip the tests, fail the tests  or run the tests anyway based on an environment variable: `INVALID_TEST_REQS={run,skip,fail}`. Now when running tests locally, developers can set it to `skip` and all tests should pass on their local machine.

Signed off by:
- [ ] @ashishb-solo 
- [ ] @ianmacclancy 
- [ ] @jackstine

### Gateway Test Service (gateway.go)
Settings are the root object from which all components are configured. We read the settings object and then build a bootstrap options object, which is just more explicit configuration derived from the settings.

Our in memory tests cheat, in that instead of defining Settings, and using the setup logic to derive bootstrap objects, we define the bootstrap options directly. We also use properties from the Settings object to initialize certain behaviors.

This tries to move our tests towards a world where all configuration is derived from the Settings object, which mirrors our live code. We define defaults, and allow tests to override them.

Signed off by:
- [ ] @ashishb-solo 
- [ ] @ianmacclancy 
- [ ] @jackstine

### Test Context (test_context.go)
All of our in memory e2e tests work by:
1. Creating an Envoy instance and running it
2. Running gloo locally
3. Writing a standard set of resources (gateway, vs, upstream) to the cache that Gloo watches (in memory cache)
4. Modifying the resources and asserting behavior when querying envoy
5. Cleaning up all resources

[...]

PR #7596, opened 2022-12-23
In Memory E2E Test Enhancements (Part 2)

sam-heilbron:
# Description

Follow up to: https://github.com/solo-io/gloo/pull/7555

# Context

## Http Response Matcher
Updated the response matcher to include header matching, as well as a custom extension point so that developers can easily validate additional criteria that is not yet supported.

I use this custom matcher to perform enhanced validation in the staged transformation test

## Proxy Protocol Test
Update the test to use the new `TestContext` framework

## Gzip Test
Introduce a transform to verify the response body is compressed correctly. These transforms are powerful to use with matchers so I opted to introduce it here as an example

## Gateway Test
Moved a test that really belongs in the `grpc_web` test files.
Removed unnecessary configuration that had no effect, and would likely just confuse developers

## Staged Transformations Test
Update the test to use the new `TestContext` framework

# Testing
## E2E Tests
Ensure that all e2e tests can pass locally, skipping tests whose requirements are not met
```
ENVOY_IMAGE_TAG=1.14.0-beta1 TEST_PKG=./test/e2e/... WAIT_ON_FAIL=1 DEBUG=1 INVALID_TEST_REQS=skip make test
```

## AWS Credentials
Ensure that all e2e tests can pass locally, skipping tests whose requirements are not met
```
TEST_PKG=./test/aws_credentials  WAIT_ON_FAIL=1 DEBUG=1 INVALID_TEST_REQS=skip make test
```

*Note: I could not figure out how to run these tests locally, even supplying the same credentials that we use in CI. I spent a bit of time investigating and decided to hold off on putting more effort into this, as we'll need to expand our aws integration tests in the future*
# Checklist:


test/helpers/requirements.go
```diff
@@ -148,6 +156,20 @@ func (r RequiredConfiguration) validateTruthyEnv() error {
...
+		// credentials not required
+		return nil
+	}
+
+	if r.awsCredentials.GetValue() == true {
+		// credentials required, but defined
+		return nil
+	}
+
+	return fmt.Errorf("aws credentials needs to be set, but are not")
```
nfuden, line 170: why is this an errorf if there is nothing being formatted here?

test/helpers/requirements.go
```diff
@@ -188,3 +210,14 @@ func Vault() Requirement {
 		TruthyEnv("RUN_VAULT_TESTS")(configuration)
 	}
 }
+
+func Aws() Requirement {
```
nfuden, line 214: nit: When adding new exported functions can we make sure that they have a function comment?

test/helpers/util.go
```diff
@@ -12,12 +12,12 @@ import (
...
 
 // PatchResource mutates an existing resource, retrying if a resourceVersionError is encountered
-func PatchResource(ctx context.Context, resourceRef *core.ResourceRef, mutator func(resource resources.Resource), client clients.ResourceClient) error {
+func PatchResource(ctx context.Context, resourceRef *core.ResourceRef, mutator func(resource resources.Resource) resources.Resource, client clients.ResourceClient) error {
 	return PatchResourceWithOffset(1, ctx, resourceRef, mutator, client)
 }
 
 // PatchResourceWithOffset mutates an existing resource, retrying if a resourceVersionError is encountered
-func PatchResourceWithOffset(offset int, ctx context.Context, resourceRef *core.ResourceRef, mutator func(resource resources.Resource), client clients.ResourceClient) error {
+func PatchResourceWithOffset(offset int, ctx context.Context, resourceRef *core.ResourceRef, mutator func(resource resources.Resource) resources.Resource, client clients.ResourceClient) error {
```
nfuden, line 20: It looks like we are changing this as implementors sometimes created functions that did not fully reflect their changes in the input resource? 
Is there a different reason for us doing this?
Given that I dont have full clarity on this and the function comment claims it mutates the existing resource perhaps some extra clarity there would make this change be clearer.

test/matchers/http.go
```diff
@@ -42,25 +44,61 @@ func HavePartialResponseBody(substring string) types.GomegaMatcher {
...
+	// Custom is a generic matcher that can be applied to validate any other properties of an http.Response
+	// Optional: If not provided, does not perform additional validation
+	Custom types.GomegaMatcher
 }
 
 // HaveHttpResponse returns a GomegaMatcher which validates that an http.Response contains
 // particular expected properties (status, body..etc)
 // If an expected body isn't defined, we default to expecting an empty response
+// NOTE TO DEVELOPERS: The underlying matchers cache properties, so if you are wrapping this
+// matcher in an Eventually, be sure to create a new matcher each iteration.
```
nfuden, line 78: Since it seems like reuse is an implementation error but we own the sub match call do we think we should store an extra attr on MatchHttpResponseMatcher and use it to work around the error?

sam-heilbron: I like the idea! Though I'm not sure I entirely follow. Are you proposing modifying the match implementation to construct a new body matcher each time that Match is called?

test/helpers/util.go
sam-heilbron: Yeah good call, I can clarify more. I encountered this when modifying virtual services. The implementation I had used to simplify the construction of these virtual services was the virtual service builder, which under the hood clones the resources before returning the new one (https://github.com/solo-io/gloo/blob/master/test/helpers/virtual_services.go#L235). Therefore, I needed to return the modified resources instead of changing it in place.

I could probably find a workaround if you have strong feelings around keeping this function signature the same. What do you think?

test/helpers/requirements.go
sam-heilbron: Yeah good catch, this is actually removed in a follow-up commit that I'm about to push

test/e2e/aws_credentials/aws_credentials_suite_test.go
```diff
@@ -3,14 +3,25 @@ package aws_credentials
...
 // These tests are isolated from the other e2e tests because they do not need envoy to be running
 func TestAwsCredentials(t *testing.T) {
 	helpers.RegisterCommonFailHandlers()
 	junitReporter := reporters.NewJUnitReporter("junit.xml")
 	RunSpecsWithDefaultAndCustomReporters(t, "AWS Credentials Suite", []Reporter{junitReporter})
 }
+
+var _ = BeforeSuite(func() {
+	testhelpers.ValidateRequirementsAndNotifyGinkgo(testhelpers.Aws())
+})
```
jackstine, line 27: Nice way to handle the requirements.

test/e2e/aws_credentials/ec2_test.go
```diff
@@ -25,31 +25,23 @@ import (
...
-		// role arn format: "arn:aws:iam::[account_number]:role/[role_name]"
-		roleArn = os.Getenv("AWS_ARN_ROLE_1")
-		if roleArn == "" {
-			Skip("no AWS role ARN available")
-		}
-		var opts clients.WriteOpts
-
-		accessKey := v.AccessKeyID
-		secretKey := v.SecretAccessKey
+		Expect(err).NotTo(HaveOccurred())
```
jackstine, line 44: I feel like the comments are only usful for the PR, and serve no purpose after the PR merges.
```
		// In the BeforeSuite, we validate that credentials are present
		// Therefore, if somehow we have reached here, we error loudly
```

test/e2e/cors_test.go
```diff
@@ -91,24 +92,33 @@ var _ = Describe("CORS", func() {
 			preFlightRequest.Host = allowedOrigin
 
 			By("Request with allowed origin")
+			req := preFlightRequest
 			Eventually(func(g Gomega) {
-				headers := executeRequestWithAccessControlHeaders(preFlightRequest, allowedOrigins[0], "GET")
-				v, ok := headers[requestACHMethods]
-				g.Expect(ok).To(BeTrue())
-				g.Expect(strings.Split(v[0], ",")).Should(ConsistOf(allowedMethods))
-
```
jackstine, line 99: Maybe a comment as to why `req := preFlightRequest`

test/e2e/aws_credentials/ec2_test.go
nfuden: idk I like having them so that if it errors there is context readily available to the implementor

test/e2e/cors_test.go
```diff
@@ -91,24 +92,33 @@ var _ = Describe("CORS", func() {
...
+				req.Header.Set("Access-Control-Request-Method", http.MethodGet)
+				req.Header.Set("Access-Control-Request-Headers", "X-Requested-With")
+
+				resp, err := http.DefaultClient.Do(req)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(resp).Should(matchers.HaveOkResponseWithHeaders(map[string]interface{}{
+					requestACHMethods: MatchRegexp(strings.Join(allowedMethods, ",")),
+					requestACHOrigin:  Equal(allowedOrigins[0]),
+				}))
+			}).Should(Succeed())
```
jackstine, line 107: Maybe a comment as to why `req := preFlightRequest`

sam-heilbron: Yeah good idea.  im going to update this test to not reuse that request object, and just build one entirely

test/e2e/aws_credentials/ec2_test.go
sam-heilbron: Yeah I'm partial to keeping it just so it's clear why we are certain about that expectation. Perhaps instead of a comment, I include that message in the Assertion description?

test/matchers/http.go
sam-heilbron: Updated to include an internal attribute that we use to fail loudly if a match has been evaluated more than once. I moved this comment to that failure line. Let me know if this is clear or if there's more we can do to help developers understand

test/matchers/http.go
```diff
@@ -14,15 +16,15 @@ import (
...
+		Body:       gomega.BeEmpty(),
 	})
 }
 
 // HaveStatusCode expects an http response with a particular status code and an empty body
 func HaveStatusCode(statusCode int) types.GomegaMatcher {
 	return HaveHttpResponse(&HttpResponse{
 		StatusCode: statusCode,
-		Body:       "",
+		Body:       gomega.BeEmpty(),
```
jackstine, line 27: Nice, I'll try to remember this

test/matchers/http.go
```diff
@@ -42,25 +44,61 @@ func HavePartialResponseBody(substring string) types.GomegaMatcher {
...
+			Header: headerName,
+			Value:  headerMatch,
+		})
+	}
+
+	expectedCustomMatcher := expected.Custom
+	if expected.Custom == nil {
+		// Default to an always accept matcher
+		expectedCustomMatcher = gstruct.Ignore()
+	}
```
jackstine, line 100: nit: shouldn't this be in reverse?  set the default value, if not nil then set the value?

sam-heilbron: We can reorder the implementation, though the behavior will be the same: if a user defines a custom matcher we use it, if not we use an Ignore matcher. I'll update

sam-heilbron: Turns out reordering this check (set the default value, and then override if the user provides one) gets challenging with type checks. The custom matcher is always a GomegaMatcher so we can switch it, but the expectedbody can be a string or matcher, so if we default it to a string, it causes challenges. If it's ok with you, I'm going to leave this order as is

jackstine: good to know

test/e2e/cors_test.go
```diff
@@ -136,24 +146,8 @@ var _ = Describe("CORS", func() {
...
 				g.Expect(cfg).To(MatchRegexp(corsFilterString))
 				g.Expect(cfg).NotTo(MatchRegexp(corsActiveConfigString))
-			}).ShouldNot(HaveOccurred())
+			}).Should(Succeed())
 		})
 	})
 
 })
-
-func executeRequestWithAccessControlHeaders(req *http.Request, origin, method string) http.Header {
```
ianmacclancy, line 145: This looks like removal of shared code and a move to duplicated code. Can you help me understand this change?

test/e2e/gateway_test.go
```diff
@@ -129,60 +119,6 @@ var _ = Describe("Gateway", func() {
 				}
 			})
 
-			It("should disable grpc web filter", func() {
```
ianmacclancy, line 132: Are we losing coverage with the removal of this test?

test/e2e/grpcweb_test.go
```diff
@@ -50,6 +52,33 @@ var _ = Describe("Grpc Web", func() {
...
+					Disable: true,
+				},
+			}
+
+			testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+				gw,
+			}
+		})
+
+		It("can disable grpc web filter", func() {
```
ianmacclancy, line 70: this was moved from gateway to this test file - I figure this is just moving it to where it makes more sense?

nfuden (approved):

test/e2e/cors_test.go
sam-heilbron: This method takes a request, adds some headers, executes the request and returns the headers from the response. Since our http matcher now includes header matching, I wanted to use our custom matcher. In addition, I felt that since this was used in just 2 places, having the explicit request building (host, headers..etc) in the test itself, allowed it to be clearer what we were testing

test/e2e/grpcweb_test.go
sam-heilbron: Yeah most other tests in this file were around handling traffic. Since we have an explicit grpc web test file, I figured it was better suited to live there.

gunnar-solo (approved):

jackstine (approved):

Issue #7153, opened 2022-09-15
Improve glooctl check reporting about virtual services and referenced resources


elcasteel:
### Is your feature request related to a problem? Please describe.

Glooctl check looks at all the resources referenced by a virtual service and reports the same "missing" error if a resource is missing, has a warning or rejected status or doesn't have a status at all. This can be confusing for users especially as gloo is doing its initial sync and status for resources have not all been written (statuses are written async and can take some time to appear in large deployments). 
A very quick fix would be to make this warning more descriptive (couple of lines) 
It would also be fairly quick (maybe a day of work) to check the status of each resource and give a separate warning for missing ref/rejected/unknown status.

### Describe the solution you'd like

The more thorough fix would require updating the interface of these `checkResourceX` methods so they return statuses and names: https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/cmd/check/root.go#L370

And then looking at those statuses in `checkVirtualServices`: https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/cmd/check/root.go#L627

I haven't looked into the tests that need to be updated.

PR #7607, opened 2022-12-29
`glooctl check` virtual service reporting fixes/improvements

Bslabe123:
# Description

For some of our sub checks (checkResourceX functions), if we see that the status hasn’t been set on a given resource, we don’t add it to the list of known resources. This is an issue whenever resources exist in our cluster but Gloo hasn’t reported their status yet as these resources will not be listed as known for other checks that rely on them.
# Context

We want to append the resource to the list of known resources in any case, not just if resource.GetNamespacedStatuses() != nil. Also adding a new corresponding error message for the (resource.GetNamespacedStatuses() == nil) case.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7153

### DIFF
--- a/changelog/v1.14.0-beta4/glooctl-check-virtual-services
+++ b/changelog/v1.14.0-beta4/glooctl-check-virtual-services
@@ -1,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7153
+    description: >
+    	Glooctls `checkResourceX` functions now report all resources regardless if resource.GetNamespacedStatuses() != nil. Added error messages for resources with no statuses.
+
--- a/changelog/v1.14.0-beta4/glooctl-check-virtual-services
+++ b/changelog/v1.14.0-beta4/glooctl-check-virtual-services
@@ -1,5 +1,5 @@
 changelog:
   - type: FIX
     issueLink: https://github.com/solo-io/gloo/issues/7153
-    description: Glooctls `checkResourceX` functions now report all resources regardless if resource.GetNamespacedStatuses() != nil. Added error messages for resources with no statuses.
+    description: Glooctl check functions now report all resources regardless if they have statuses reported yet or not. Added error messages for resources with no statuses.
 
--- a/changelog/v1.14.0-beta4/glooctl-check-virtual-services
+++ b/changelog/v1.14.0-beta4/glooctl-check-virtual-services
@@ -1,6 +1,5 @@
 changelog:
   - type: FIX
     issueLink: https://github.com/solo-io/gloo/issues/7153
-    description: >
-    	Glooctls `checkResourceX` functions now report all resources regardless if resource.GetNamespacedStatuses() != nil. Added error messages for resources with no statuses.
+    description: Glooctls `checkResourceX` functions now report all resources regardless if resource.GetNamespacedStatuses() != nil. Added error messages for resources with no statuses.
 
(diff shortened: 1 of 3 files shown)


Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=ccaa7a81-c5f8-4439-89a6-96383dc8c7d9

changelog/v1.14.0-beta4/glooctl-check-virtual-services
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7153
+    description: Glooctls `checkResourceX` functions now report all resources regardless if resource.GetNamespacedStatuses() != nil. Added error messages for resources with no statuses.
```
nfuden, line 4: nit: This is should be user facing and they dont necessarily know or care about what a checkResourceX function may be.

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -428,8 +428,11 @@ func checkUpstreams(ctx context.Context, opts *options.Options, namespaces []str
 						multiErr = multierror.Append(multiErr, errors.New(errMessage))
 					}
 				}
-				knownUpstreams = append(knownUpstreams, renderMetadata(upstream.GetMetadata()))
+			} else {
+				errMessage := fmt.Sprintf("Found upstream with no status: %s\n", renderMetadata(upstream.GetMetadata()))
+				multiErr = multierror.Append(multiErr, fmt.Errorf(errMessage))
```
nfuden, line 433: Why are we fmt errorf after having already done a sprintf?

Bslabe123: Good callout, we have this in other places in this file should we fix these as well?

elcasteel (approved):

nfuden (approved):

projects/gloo/cli/pkg/cmd/check/check_test.go
```diff
@@ -0,0 +1,148 @@
...
+	})
+
+	AfterEach(func() {
+		cancel()
+	})
+
+	Context("glooctl check", func() {
+		It("should error if resource has no status", func() {
+
+			client := helpers.MustKubeClient()
```
sam-heilbron, line 42: Why do we set UseMemoryClients in the beforeEach but then call this line here?

projects/gloo/cli/pkg/cmd/check/check_test.go
```diff
@@ -0,0 +1,148 @@
...
+			client.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: defaults.GlooSystem,
+				},
+			}, metav1.CreateOptions{})
+
+			appName := "default"
+			client.AppsV1().Deployments("gloo-system").Create(ctx, &appsv1.Deployment{
```
sam-heilbron, line 50: Do other tests in this package require kube clients? If so, this seems safe to do, but if not, we could probably just use inMemory clients to read/write resources

sam-heilbron (commented):
This could cause `glooctl check` to return an error after upgrading to this version of Gloo when previously it did not. Perhaps the PR is enough, but are there other ways we could notify clients of this change in behavior? I know we don't recommend it, but there's a chance that people consume the output of `glooctl check` to build alerts.

We also have some features in our API where we process resources with particular labels. Is there a chance that the Gloo controllers are configured to ignore certain resources but then we report an error now?

projects/gloo/cli/pkg/cmd/check/check_test.go
nfuden: https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/helpers/clients.go#L94

PR #7623, opened 2023-01-03
`glooctl check --context <string>`

Bslabe123:
# Description

Currently we dont have any mechanism for setting the kubecontext to use when performing `glooctl check` and we default to the kubecontext currently selected. We would like a mechanism to select the kubecontext used when scanning, preferably with a similar UX to `kubectl --context <string>`

# Context

A new `--context <string>` was added to select the context used for scanning. Changes were made to each of our clients which originally created a rest config with the kubecontext being hardcoded to an empty string (corresponding to the currently selected kubecontext), each of these clients were changed to now create rest config with the kubecontext being set by the `--context` flag value.

# Checklist:

### DIFF
--- a/projects/gloo/cli/pkg/cmd/options/options.go
+++ b/projects/gloo/cli/pkg/cmd/options/options.go
@@ -50,6 +50,7 @@ type Top struct {
 	PodSelector        string   // label selector for pod scanning
 	ResourceNamespaces []string // namespaces in which to check custom resources
 	ReadOnly           bool     // Makes check read only by skipping any checks that create resources in the cluster
+	KubeContext        string   // K8s Context to run glooctl on
 }
 
 type HelmInstall struct {
--- a/projects/gloo/cli/pkg/flagutils/metadata.go
+++ b/projects/gloo/cli/pkg/flagutils/metadata.go
@@ -38,3 +38,8 @@ func AddExcludeCheckFlag(set *pflag.FlagSet, strarrptr *[]string) {
 func AddReadOnlyFlag(set *pflag.FlagSet, readOnly *bool) {
 	set.BoolVarP(readOnly, "read-only", "", false, "only do checks that dont require creating resources (i.e. port forwards)")
 }
+
+// AddReadOnlyFlag adds a flag to our flag set that indicates we shouldn't do anything that requires RBAC create permissions
+func AddKubeContextFlag(set *pflag.FlagSet, KubeContext *string) {
+	set.StringVarP(KubeContext, "context", "", "", "kube context to use when interacting with kubernetes")
+}
--- a/projects/gloo/cli/pkg/helpers/clients.go
+++ b/projects/gloo/cli/pkg/helpers/clients.go
@@ -140,7 +140,7 @@ func KubeClient() (kubernetes.Interface, error) {
 }
 
 func MustGetNamespaces(ctx context.Context) []string {
-	ns, err := GetNamespaces(ctx)
+	ns, err := GetNamespaces(ctx, "")
 	if err != nil {
 		log.Fatalf("failed to list namespaces")
 	}
@@ -148,17 +148,13 @@ func MustGetNamespaces(ctx context.Context) []string {
 }
 
 // Note: requires RBAC permission to list namespaces at the cluster level
-func GetNamespaces(ctx context.Context) ([]string, error) {
+func GetNamespaces(ctx context.Context, kubecontext string) ([]string, error) {
 	customFactory := getConfigClientFactory()
 	if customFactory != nil {
 		return []string{"default", defaults.GlooSystem}, nil
 	}
+	kubeClient, err := GetKubernetesClient(kubecontext)
 
-	cfg, err := kubeutils.GetConfig("", "")
-	if err != nil {
-		return nil, errors.Wrapf(err, "getting kube config")
-	}
-	kubeClient, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
 		return nil, errors.Wrapf(err, "getting kube client")
 	}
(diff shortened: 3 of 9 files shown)


projects/gloo/cli/pkg/helpers/clients.go
```diff
@@ -598,21 +594,21 @@ func MustNamespacedAuthConfigClient(ctx context.Context, ns string) extauth.Auth
...
+	client, err := AuthConfigClient(ctx, namespaces, "")
 	if err != nil {
 		log.Fatalf("failed to create auth config client: %v", err)
 	}
 	return client
 }
 
 // provide "" (metav1.NamespaceAll) to get a cluster-scoped authConfig client
-func AuthConfigClient(ctx context.Context, namespaces []string) (extauth.AuthConfigClient, error) {
+func AuthConfigClient(ctx context.Context, namespaces []string, kubecontext string) (extauth.AuthConfigClient, error) {
```
nfuden, line 605: Question: Is there a reason we are passing this as an extra arg rather than doing something where we stuff it into the go context?

projects/gloo/cli/pkg/cmd/options/options.go
```diff
@@ -50,6 +50,7 @@ type Top struct {
 	PodSelector        string   // label selector for pod scanning
 	ResourceNamespaces []string // namespaces in which to check custom resources
 	ReadOnly           bool     // Makes check read only by skipping any checks that create resources in the cluster
+	KubeContext        string   // K8s Context to run glooctl check with
```
nfuden, line 53: is this only for glooctl check or could it be used by other glooctl commands?

projects/gloo/cli/pkg/helpers/clients.go
Bslabe123: Considering that context is concurrency feature would it be appropriate to put something like opts.Top.KubeContext in there? I could be convinced otherwise

projects/gloo/cli/pkg/cmd/options/options.go
Bslabe123: This just reflects the current behavior, do you have an idea for a better comment?

projects/gloo/cli/pkg/helpers/clients.go
nfuden: Context is a concurrency feature? I dont quite understand what you are saying here. Can you expand on that?

elcasteel (dismissed):
lgtm

projects/gloo/cli/pkg/cmd/options/options.go
nfuden: We should probably try to support it for all glooctl operations. I know this is an expansion in scope but it seems odd that one could run glooctl install --context and then have it be a recognized argument but only work for glooctl check

projects/gloo/cli/pkg/helpers/clients.go
Bslabe123: As per discussion opts.Top now has ContextAccessible fields that get passed into the context. But also see that the motivation behind contexts isn't entirely concurrency although it is one of the motivations (i.e. cancel signals)

projects/gloo/cli/pkg/cmd/options/options.go
Bslabe123: Added!

projects/gloo/cli/pkg/cmd/options/contextoptions/contextoptions.go
```diff
@@ -0,0 +1,44 @@
...
+type Consul struct {
+	UseConsul       bool // enable consul config clients
+	RootKey         string
+	AllowStaleReads bool
+	Client          func() (*api.Client, error)
+}
+
+func ContextAccessibleFrom(ctx context.Context) (ContextAccessible, error) {
+	if ctx != nil {
+		if logger, ok := ctx.Value("top").(ContextAccessible); ok {
```
nfuden, line 31: What is the logger here?

Bslabe123: Typo! Should be `contextAccessible`, I used the `loggerFrom` function in go-utils as a template

projects/gloo/cli/pkg/flagutils/metadata.go
```diff
@@ -38,3 +38,8 @@ func AddExcludeCheckFlag(set *pflag.FlagSet, strarrptr *[]string) {
 func AddReadOnlyFlag(set *pflag.FlagSet, readOnly *bool) {
 	set.BoolVarP(readOnly, "read-only", "", false, "only do checks that dont require creating resources (i.e. port forwards)")
 }
+
+// AddReadOnlyFlag adds a flag to our flag set that indicates we shouldn't do anything that requires RBAC create permissions
+func AddKubeContextFlag(set *pflag.FlagSet, KubeContext *string) {
```
nfuden, line 43: Comment naming mismatch

elcasteel (dismissed):

docs/content/reference/cli/glooctl.md
```diff
@@ -21,6 +21,7 @@ glooctl is the unified CLI for Gloo.
       --consul-root-key string     key prefix for for Consul key-value storage. (default "gloo")
       --consul-scheme string       URI scheme for the Consul server. Use with --use-consul (default "http")
       --consul-token string        Token is used to provide a per-request ACL token which overrides the agent's default token. Use with --use-consul
+      --context string             kube context to use when interacting with kubernetes
```
kdorosh, line 24: i think this flag should be `--kube-context`, the current name is not descriptive enough imo

Bslabe123: Even if were looking to mirror the UX of `kubectl --context`?

kdorosh: yes, i'd say so. the context is implied when it's `kubectl` but as is it kind of implies this is a `glooctl` context which is misleading

nfuden: I like it. I can see either way but definitely this seems cleaner

elcasteel (approved):

nfuden (approved):

PR #7626, opened 2023-01-03
fix a typo

peterj:
# Description

Fixes a "Virtual Server" to "Virtual Service" typo.


docs/content/guides/traffic_management/destination_selection/path_matching/_index.md
```diff
@@ -241,7 +241,7 @@ It will now return results.
 ]
 ```
 
-You can try any number of different combination to see how the exact match option works. When you're done, let's clean up the exact match Virtual Server and check out the regex matcher.
+You can try any number of different combination to see how the exact match option works. When you're done, let's clean up the exact match Virtual Service and check out the regex matcher.
```
jenshu, line 244:
```suggestion
You can try any number of different combinations to see how the exact match option works. When you're done, let's clean up the exact match Virtual Service and check out the regex matcher.
```

sam-heilbron (approved):

jenshu (approved):

PR #7630, opened 2023-01-04
Enhancements to Test Utilities

sam-heilbron:
# Description
Enhance testing utilities

# Context
Used by: https://github.com/solo-io/solo-projects/pull/4394

## SnapshotWriter
Expand the Snapshot Writer to support AuthConfig objects

## Upgrade Gomega
Upgrade the Gomega dependency to match the version used in Gloo Enterprise 

## Gomega Utilities
Build out a series of Matchers, Transforms and Assertions which can be used more broadly in tests. These previously existed spread out in the Gloo codebase, and by placing them in a shared place, we can use them more broadly.

*Note: This involved moving the matchers package down a level (into the gomega package), which resulted in a number of e2e test files having their import path changed*

# Checklist:


test/kube2e/util.go
```diff
@@ -152,78 +137,6 @@ func ExpectEqualProtoMessages(g Gomega, a, b proto.Message, optionalDescription
 	g.Expect(a.String()).To(Equal(b.String()), optionalDescription...)
 }
 
-func EventuallyMetricsBecomeConsistent(offset int, metricsPort int) {
```
sam-heilbron, line 155: This logic was just moved to a shared assertion/matcher package since it's useful in other tests to perform similar validation that certain statistics stabilize

test/e2e/staged_transformation_test.go
```diff
@@ -120,7 +121,7 @@ var _ = Describe("Staged Transformation", func() {
 			Eventually(func(g Gomega) {
 				res, err := http.DefaultClient.Do(req)
 				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(res).Should(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
+				g.Expect(res).Should(matchers2.HaveHttpResponse(&matchers2.HttpResponse{
```
jenshu, line 124: nit: testmatchers (what it was before) seems like a better import name

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+
+	By("Perform the assertions while the port forward is open")
+	for _, assertion := range assertions {
+		assertion.WithOffset(offset + 1).ShouldNot(HaveOccurred())
+	}
+}
+
+// IntStatisticReachesConsistentValueAssertion returns an assertion that a prometheus stats has reached a consistent value
+// It optionally returns the value of that statistic as well
+func IntStatisticReachesConsistentValueAssertion(prometheusStat string, inARow int) (types.AsyncAssertion, *int) {
```
jenshu, line 83: what does inARow mean?

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+// a series of Asynchronous assertions. The fort-forward is cleaned up with the function returns
+func EventuallyWithOffsetStatisticsMatchAssertions(offset int, statsPortFwd StatsPortFwd, assertions ...types.AsyncAssertion) {
+	portForward, err := cliutil.PortForward(
+		statsPortFwd.ResourceNamespace,
+		statsPortFwd.ResourceName,
+		fmt.Sprintf("%d", statsPortFwd.LocalPort),
+		fmt.Sprintf("%d", statsPortFwd.TargetPort),
+		false)
+
+	defer func() {
```
jenshu, line 57: since `portForward` can be nil in case of error, should this defer be moved to after we ensure there's no error (i.e. after the Expect below)?

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+		if portForward.Process != nil {
+			portForward.Process.Kill()
+			portForward.Process.Release()
+		}
+	}()
+	ExpectWithOffset(offset+1, err).NotTo(HaveOccurred())
```
jenshu, line 63: why are we adding 1 to the offset everywhere?

test/kube2e/util.go
```diff
@@ -117,30 +112,20 @@ gatewayProxies:
...
-	err := portFwd.Start()
-	Expect(err).ToNot(HaveOccurred())
-
-	defer func() {
-		if portFwd.Process != nil {
-			portFwd.Process.Kill()
-		}
-	}()
+	// We port-forward the Gloo deployment stats port to inspect the metrics and log settings
+	glooStatsForwardConfig := assertions.DefaultStatsPortFwd
```
jenshu, line 116: will `DefaultStatsPortFwd` be used anywhere besides here? i'm wondering if we need the default or if we can just construct the StatsPortFwd here (since we are overriding namespace anyway)

test/kube2e/util.go
```diff
@@ -117,30 +112,20 @@ gatewayProxies:
...
+	glooStatsForwardConfig.ResourceNamespace = installNamespace
 
 	// Gloo components are configured to log to the Info level by default
-	EventuallyLogLevel(metricsPort, zapcore.InfoLevel)
+	logLevelAssertion := assertions.EventuallyLogLevel(zapcore.InfoLevel)
+
+	// The emitter at some point should stabilize and not continue to increase the number of snapshots produced
+	emitterMetricAssertion, _ := assertions.IntStatisticReachesConsistentValueAssertion("api_gloosnapshot_gloo_solo_io_emitter_snap_out", 4)
```
jenshu, line 123: where is the name of this stat defined? do we know it won't change?

test/gomega/assertions/logging.go
```diff
@@ -0,0 +1,25 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+// EventuallyLogLevel returns an Assertion to verify that the dynamic log level matches the provided value
+func EventuallyLogLevel(logLevel zapcore.Level) types.AsyncAssertion {
+	request, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://localhost:%d/logging", stats.DefaultPort), nil)
+	Expect(err).NotTo(HaveOccurred())
+
+	return Eventually(func(g Gomega) {
+		g.Expect(http.DefaultClient.Do(request)).Should(matchers.HaveHttpResponse(&matchers.HttpResponse{
```
jenshu, line 20: are we closing the response bodies somewhere?

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+
+	By("Ensure port-forward is open before performing assertions")
+	statsRequest, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://localhost:%d/", statsPortFwd.LocalPort), nil)
+	ExpectWithOffset(offset+1, err).NotTo(HaveOccurred())
+	EventuallyWithOffset(offset+1, func(g Gomega) {
+		g.Expect(http.DefaultClient.Do(statsRequest)).To(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
```
jenshu, line 69: are we closing the response bodies somewhere?

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	var (
+		currentlyInARow   = 0
+		previousStatValue = 0
+		currentStatValue  = pointer.Int(0)
+	)
+
+	return Eventually(func(g Gomega) {
+		g.Expect(http.DefaultClient.Do(metricsRequest)).To(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
```
jenshu, line 100: are we closing the response bodies somewhere?

test/e2e/staged_transformation_test.go
sam-heilbron: Agreed! This was auto-changed when I moved the package, will update

test/gomega/assertions/stats.go
sam-heilbron: InARow is the number of consecutive times the statistic must be the same value for it to be considered "consistent". We periodically check the stat value (every few seconds), and if we get (9,9,9,9) that would satisfy 4 in a row. Is there a better name for this variable or perhaps just a better comment explaining it. Alternatively, if we got (8,8,9,9) then we would only have 2 in a row, and would keep polling

test/gomega/assertions/stats.go
sam-heilbron: Yeah I like it

test/gomega/assertions/stats.go
sam-heilbron: The idea was that the caller provides the offset, and this function is 1 more than whatever the caller provided (each fucnction call is worth 1 on the call stack).

test/kube2e/util.go
sam-heilbron: We can certainly just construct it, I'm in favor of that. The default port foward is used elsewhere in the codebase, so I figured it could be a useful utility. I'll update here to just explicitly construct all the properties

test/kube2e/util.go
sam-heilbron: Yeah this was pulled from existing code, but just made more general since I perform a similar assertion against a statistic of a different name. The statistic https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/api/v1/gloosnapshot/api_snapshot_emitter.sk.go#L58 comes from our emitter. The purpose is that this stat should stabliize, meaning that gloo has reached a consistent state.

If the stat is changed accidentally (it should't because clients may alert on it), then tests will fail because we expect this stat to be non-zero

test/gomega/assertions/stats.go
jenshu: gotcha, i think just a comment on the function would help

test/gomega/assertions/logging.go
sam-heilbron: Yeah the shared http matcher that we use handles all the complexity for us. Under the hood we just rely on gomega's matcher: https://github.com/solo-io/gloo/blob/master/test/matchers/http.go#L112 which handles body reading/closing.

This is why I think it would be good to migrate more of our tests to use this shared utility, because it simplifies many of the http assertions/matches we want to perform

test/gomega/assertions/stats.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7630#discussion_r1063594796

test/gomega/assertions/stats.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7630#discussion_r1063594796

test/gomega/transforms/regex.go
```diff
@@ -0,0 +1,25 @@
+package transforms
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+// IntRegexTransform parses a http.Response body
```
ben-taussig-solo, line 9: I think this comment is a bit misleading right now -- could we change this to something like
```
// IntRegexTransform returns a function that can be used to transform a byte array (e.g., an http.Response body) into an int
// by applying a regular expression to the byte array. The regular expression must have exactly one capture group.
// The function will return the integer value of the capture group.
```

test/gomega/transforms/regex.go
```diff
@@ -0,0 +1,25 @@
...
+// and returns the integer value of the provided regular expression
+func IntRegexTransform(regexp *regexp.Regexp) func(body []byte) (int, error) {
+	return func(body []byte) (int, error) {
+		matches := regexp.FindAllStringSubmatch(string(body), -1)
+		if len(matches) != 1 {
+			return 0, fmt.Errorf("found %d matches, expected 1", len(matches))
+		}
+
+		matchCount, conversionErr := strconv.Atoi(matches[0][1])
```
ben-taussig-solo, line 18: Could we add some commentary around this line? Something like

```
		// matches[0] is the first match
		// matches[0][1] is the first capture group
		matchCount, conversionErr := strconv.Atoi(matches[0][1])
```

test/gomega/transforms/regex.go
ben-taussig-solo: Additionally, I'm questioning whether this method should just live as a helper in `test/gomega/assertions/stats.go`, which is the only place it is currently used.

sam-heilbron: I like it!

test/gomega/transforms/regex.go
sam-heilbron: Good call, will update

test/kube2e/util.go
ben-taussig-solo: I think 4 is a bit of a magic number here -- could we parameterize that value and leave a comment explaining why seeing the same value of the metric 4 times in a row gives us confidence that the statistic has reached a consistent value?

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	"github.com/solo-io/gloo/pkg/cliutil"
+	"github.com/solo-io/go-utils/stats"
+)
+
+const (
+	TimeToSyncStats     = time.Second * 5 // Metrics reporting occurs at 5s intervals
+	SafeTimeToSyncStats = TimeToSyncStats + time.Second*2
```
ben-taussig-solo, line 24: Could we explain this constant in a comment? FWIW I have noticed that metrics updates take ~7 seconds to be propagated, so this makes intuitive sense to me, but I think it's valuable to call that out

test/gomega/assertions/stats.go
```diff
@@ -0,0 +1,117 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	"github.com/solo-io/gloo/pkg/cliutil"
+	"github.com/solo-io/go-utils/stats"
+)
+
+const (
+	TimeToSyncStats     = time.Second * 5 // Metrics reporting occurs at 5s intervals
```
ben-taussig-solo, line 23: Can a user adjust the metrics reporting interval? If so, can we derive this value from that setting?

test/gomega/transforms/regex.go
ben-taussig-solo: Comment added here: https://github.com/solo-io/gloo/pull/7630/commits/f650904ba066f9866360c410b52c069447fbb88d#diff-85c7d2d5a72ce65a52b67fffe00aea286281c6966b17872a50ba2e24d0fca99d

Resolving this thread

test/gomega/assertions/stats.go
ben-taussig-solo: Comment added here: https://github.com/solo-io/gloo/pull/7630/commits/f650904ba066f9866360c410b52c069447fbb88d#diff-85c7d2d5a72ce65a52b67fffe00aea286281c6966b17872a50ba2e24d0fca99d

Resolving this thread

test/gomega/transforms/regex.go
ben-taussig-solo: Comment added here: https://github.com/solo-io/gloo/pull/7630/commits/f650904ba066f9866360c410b52c069447fbb88d#diff-85c7d2d5a72ce65a52b67fffe00aea286281c6966b17872a50ba2e24d0fca99d

Resolving this thread

test/kube2e/util.go
sam-heilbron: Yeah happy to provide a comment oulining why 4 is chosen. When you say parametrize that value, in what sense do you mean? The utility Assertion has that as a parameter. I figured we don't want to pass it into "EventuallyReachesConsistentState" since the definition of what is considered "consistent state" can be maintained within that function and users don't need to know what it requires

test/gomega/assertions/stats.go
sam-heilbron: Sure thing!

test/gomega/assertions/stats.go
sam-heilbron: The reporting interval is based on the libraries we choose, and we configure the values. I believe we just use the defaults

test/gomega/assertions/stats.go
ben-taussig-solo: Updated [here](https://github.com/solo-io/gloo/pull/7630/commits/0e2824c88597cf641c956b22a2bb345507a538f3), looks great!

test/gomega/assertions/logging.go
jenshu: agreed, we should use this in all our tests (not necessarily in this PR)

test/kube2e/util.go
ben-taussig-solo: Adding a comment is fine! I was just thinking we could make a constant like
`timesInARow := 4` for added clarity, but I think a comment is preferable

jenshu (approved):

test/kube2e/util.go
sam-heilbron: Oh nice! yeah thats waht I ended up doing

ben-taussig-solo (approved):

PR #7633, opened 2023-01-05
Feature/expose suppress envoy headers v2

irateswami:
# Description

This change exposes the suppress envoy headers field, within a router object. Previously the router object had no args but was present, so this is the first field to be exposed from that. 

# Checklist:


changelog/v1.14.0-beta5/suppress-envoy-headers.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/6713
+    resolvesIssue: false
+    description: Sometimes its desirable to call another instance of Gloo and not pass any of the x-envoy headers, this allows that.
```
gunnar-solo, line 5: nit: this is a user-facing changelog.  Make it declarative (also fix `its` typo)

projects/gloo/api/v1/options.proto
```diff
@@ -54,6 +54,13 @@ import "github.com/solo-io/solo-kit/api/external/envoy/api/v2/core/socket_option
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
+// Router is an extension of the envoy http filters
+// Maps to https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto
+message Router{
```
sam-heilbron, line 59: Let's define this in a separate router package, like we do for our other filters. See: https://github.com/solo-io/gloo/tree/master/projects/gloo/api/v1/options for some examples

projects/gloo/api/v1/options.proto
```diff
@@ -160,6 +167,8 @@ message HttpListenerOptions {
     google.protobuf.BoolValue leftmost_xff_address = 16;
 
     dfp.options.gloo.solo.io.FilterConfig dynamic_forward_proxy = 28;
+
+    Router router = 18;
```
gunnar-solo, line 171: I was _just about_ to complain about having an object with only 1 field, but then saw that there were several more fields that envoy stuffs into its `Router` that we may want to have later

sam-heilbron, line 171: Can we add an explanatory comment about what the Router properties are and perhaps link to envoy documents? These proto comments are published in our public api reference docs (https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk/#httplisteneroptions) and are generally the way that users can understand how/when to use these settings

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -160,6 +160,71 @@ var _ = Describe("Listener Subsystem", func() {
...
+
+				hcm := typedConfig.(*envoy_http_connection_manager_v3.HttpConnectionManager)
+				Expect(hcm.HttpFilters).To(HaveLen(2))
+				ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+				routeFilter := hcm.GetHttpFilters()[1]
+				_ = routeFilter
+				Expect(routeFilter.GetName()).To(Equal("envoy.filters.http.router"))
+
+				/**
```
sam-heilbron, line 222: Why is this commented out?

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,6 +209,18 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
 	// As outlined by the Envoy docs, the last configured filter has to be a terminal filter.
 	// We set the Router filter (https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router)
 	// as the terminal filter in Gloo Edge.
+	if router := h.listener.Options.GetRouter(); router != nil {
```
sam-heilbron, line 212: Could we combine these two cases: router options defined, router options not defined so that we just build the router config (with or without options) and then append the Router filter with that configuration always?

projects/gloo/api/v1/options.proto
```diff
@@ -54,6 +54,13 @@ import "github.com/solo-io/solo-kit/api/external/envoy/api/v2/core/socket_option
 import "google/protobuf/duration.proto";
 import "google/protobuf/wrappers.proto";
 
+// Router is an extension of the envoy http filters
```
sam-heilbron, line 57: nit: The Router is an http filter

sam-heilbron (commented):
Looking great! 

Let's definitely add an e2e test to demonstrate this functionality since anytime we're messing with building a new filter config we want to verify it is accepted by envoy.

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,6 +209,18 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
...
+		routerFilter := plugins.MustNewStagedFilter(
+			wellknown.Router,
+			router,
+			plugins.AfterStage(plugins.RouteStage),
+		)
+
+		envoyHttpFilters = append(envoyHttpFilters, routerFilter.HttpFilter)
+
+		return envoyHttpFilters
```
gunnar-solo, line 221: nit:  I prefer the Terminal filter logic to be more explicitly grouped.  The code as-is could potentially be separated by someone not being careful on future edits.  Specifically:
```go
if router != nil {
  // append your changes
} else {
  // append default router
}

return envoyHttpFilters
```

gunnar-solo (changes requested):
Some small nits which would be nice to address.  Overall looks good!

Looking for a finished e2e test before full approval, though

projects/gloo/api/v1/options.proto
irateswami: Thanks!

projects/gloo/pkg/translator/network_filters.go
irateswami: Yeah that's a good point, I can change that

changelog/v1.14.0-beta5/suppress-envoy-headers.yaml
irateswami: Ah that pesky apostrophe. Can you clarify what you mean by declarative?

projects/gloo/pkg/translator/listener_subsystem_test.go
irateswami: forgot to delete

test/e2e/happypath_test.go
```diff
@@ -184,6 +185,48 @@ var _ = Describe("Happy path", func() {
 					Expect(statsString).To(ContainSubstring("vhost.virt1.vcluster.test-vc."))
 				})
 
+				It("it correctly passes the suppress envoy headers config", func() {
+					proxy := getTrivialProxyForUpstream(defaults.GlooSystem, envoyPort, up.Metadata.Ref())
+
+					// Set a virtual cluster matching everything
```
sam-heilbron, line 191: This comment more accurately describes the test above "correctly configures envoy to emit virtual cluster statistics". What we're really doing here is configuring an http listener option to set suppressEnvoyHeaders to true.

test/e2e/happypath_test.go
```diff
@@ -184,6 +185,48 @@ var _ = Describe("Happy path", func() {
...
+					proxy.Listeners[0].GetHttpListener().Options = &gloov1.HttpListenerOptions{
+						Router: &gloov1.Router{
+							SuppressEnvoyHeaders: wrapperspb.Bool(true),
+						},
+					}
+
+					_, err := testClients.ProxyClient.Write(proxy, clients.WriteOpts{})
```
sam-heilbron, line 198: nit: can we include the Ctx in the clients.WriteOpts?

test/e2e/happypath_test.go
```diff
@@ -184,6 +185,48 @@ var _ = Describe("Happy path", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+
+					// This will hit the virtual host with the above virtual cluster config
```
sam-heilbron, line 201: What is the purpose of this line? I imagine it is to ensure that in addition to Envoy containing the configuration we expect (which we assert a few lines later), it's that we can route traffic to the upstream (envoy cluster) when we set suppressEnvoyHeaders. This assertion is probably good enough, though if we wanted to make it even more expressive, we could probably assert that a response doesn't return any of the envoy headers.

test/e2e/happypath_test.go
```diff
@@ -184,6 +185,48 @@ var _ = Describe("Happy path", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+
+					// We expect the envoy configuration to contain these properties in the configuration dump
+					Expect(cfg).To(MatchRegexp("\"suppress_envoy_headers\": true"))
+				})
+
+				It("it correctly DID NOT pass the suppress envoy headers config", func() {
+					proxy := getTrivialProxyForUpstream(defaults.GlooSystem, envoyPort, up.Metadata.Ref())
+
+					// Set a virtual cluster matching everything
```
sam-heilbron, line 214: Can we update this comment to clarify what we're configuring? It seems that we're explicitly not defining config for the http listener options

test/e2e/happypath_test.go
```diff
@@ -184,6 +185,48 @@ var _ = Describe("Happy path", func() {
...
+					_, err := testClients.ProxyClient.Write(proxy, clients.WriteOpts{})
+					Expect(err).NotTo(HaveOccurred())
+
+					// This will hit the virtual host with the above virtual cluster config
+					TestUpstreamReachable()
+
+					cfg, err := envoyInstance.ConfigDump()
+					Expect(err).NotTo(HaveOccurred())
+
+					// We expect the envoy configuration to contain these properties in the configuration dump
```
sam-heilbron, line 226: I think what you meant to say is: "We expect the envoy configuration NOT to contain these properties in the configuration dump"

projects/gloo/api/v1/options/router/router.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package gloo.solo.io;
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1";
```
sam-heilbron, line 3: I think we want the go_package to match the location of the generated go code. See https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/options/protocol_upgrade/protocol_upgrade.proto#L5 for an example.

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,14 +208,19 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
...
-		Name: wellknown.Router,
-		ConfigType: &envoyhcm.HttpFilter_TypedConfig{
-			TypedConfig: &any.Any{
-				TypeUrl: "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router",
-			},
-		},
-	})
+	routerV3 := routerv3.Router{}
+
+	if routerFilter := h.listener.GetOptions().GetRouter(); routerFilter != nil {
```
sam-heilbron, line 213: nit: We can consolidate these nested conditionals, since the Getter methods protect us from NPEs:
```go
if h.listener.GetOptions().GetRouter().GetSuppressEnvoyHeaders().GetValue() {
  // custom logic if user set "successEnvoyHeaders" to true
}
```

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,14 +208,19 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
...
+		if routerFilter.GetSuppressEnvoyHeaders().GetValue() {
+			routerV3.SuppressEnvoyHeaders = true
+		}
+	}
+
+	envoyHttpFilters = append(envoyHttpFilters, plugins.MustNewStagedFilter(
```
sam-heilbron, line 219: In our other plugins, which are used to generate Http Filters, we use NewStagedFilter (example: https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/plugins/gzip/plugin.go#L58) and if an error occurs during marshaling (it shouldn't) we return the error. We may want to follow a similar paradigm for consistency. MustNewStagedFilter is a little misleading of a name since if an error occurs, instead of panicing we just drop the error on the floor.

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -173,6 +173,52 @@ var _ = Describe("Listener Subsystem", func() {
...
+				hcmFilter := filterChain.GetFilters()[0]
+				typedConfig, err := sslutils.AnyToMessage(hcmFilter.GetConfigType().(*envoy_config_listener_v3.Filter_TypedConfig).TypedConfig)
+				_ = typedConfig
+
+				hcm := typedConfig.(*envoy_http_connection_manager_v3.HttpConnectionManager)
+				Expect(hcm.HttpFilters).To(HaveLen(2))
+				ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+				routeFilter := hcm.GetHttpFilters()[1]
+				Expect(routeFilter.GetName()).To(Equal("envoy.filters.http.router"))
```
ianmacclancy, line 219: Are we skipping validation on the boolean value as it is returned in a format we cannot easily validate?

ianmacclancy: I dont see any added comments or validation on the string value existing - opening this back up

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -173,6 +173,47 @@ var _ = Describe("Listener Subsystem", func() {
 				ExpectWithOffset(1, routeConfig.GetName()).To(Equal(hcmRouteConfigName))
 			},
 		),
+		Entry(
```
sam-heilbron, line 176: We can keep this here, but I'm not sure this is the best place for this test. At the top of these tests I see:
```
// These tests validate that the ListenerSubsystemTranslatorFactory produces Translators
// which in turn create Envoy Listeners and RouteConfigurations with expected values
// The tests are non-exhaustive, as we expect each translator to more rigorously test the
// edge cases. Instead, these tests focus on the high level Envoy resources that are created.
```

Again, we can keep them in here, but this test seems like it fits more with the tests defined in translator_test.go, which perform similar validation to this test

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -173,6 +173,47 @@ var _ = Describe("Listener Subsystem", func() {
...
+			func(listener *envoy_config_listener_v3.Listener, routeConfigs []*envoy_config_route_v3.RouteConfiguration) {
+				By("Should be able to add and translate the router to an envoy config")
+				ExpectWithOffset(1, listener.GetFilterChains()).To(HaveLen(1))
+				lf := listener.GetFilterChains()
+				_ = lf
+				filterChain := listener.GetFilterChains()[0]
+				ExpectWithOffset(1, filterChain.GetFilterChainMatch()).To(BeNil())
+
+				hcmFilter := filterChain.GetFilters()[0]
+				typedConfig, err := sslutils.AnyToMessage(hcmFilter.GetConfigType().(*envoy_config_listener_v3.Filter_TypedConfig).TypedConfig)
```
sam-heilbron, line 206: I would check out https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/translator/translator_test.go#L395 for an example of unmarshaling this configuration to the proper types. It uses a utility to simplify this, as well as allows us to assert values on types, not just typeUrls

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -132,7 +131,11 @@ func (h *hcmNetworkFilterTranslator) ComputeNetworkFilter(params plugins.Params)
 	httpConnectionManager := h.initializeHCM()
 
 	// 2. Apply HttpFilters
-	httpConnectionManager.HttpFilters = h.computeHttpFilters(params)
+	var err error
```
sam-heilbron, line 134: I don't think this is a safe change. 
Previously we would compute all http filters, and if we encountered an error we would add it to a report (put the status on the top level gateway resource) and continue. This way all valid http filters would be placed on the HCM.

Now if we encounter an error, we return an empty network filter.

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,16 +212,25 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
...
+	if h.listener.GetOptions().GetRouter().GetSuppressEnvoyHeaders().GetValue() {
+		routerV3.SuppressEnvoyHeaders = true
+	}
+
+	newStagedFilter, err := plugins.NewStagedFilter(
+		wellknown.Router,
+		&routerV3,
+		plugins.AfterStage(plugins.RouteStage),
+	)
+	if err != nil {
```
sam-heilbron, line 226: We should be following the same pattern that we do above:

1. Create the filter
2. If it succeeds, append it
3. If it fails, add it as an error to the report

sam-heilbron (changes requested):
I think we've subtly changed the way we translate the HCM http filters, and we should probably just keep it as it was before. Let's hop on a call if you have any questions about this

projects/gloo/pkg/translator/network_filters.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7633#discussion_r1065905640

We wouldn't be appending a filter in a bad state, we would be adding that to the report.

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -132,6 +131,7 @@ func (h *hcmNetworkFilterTranslator) ComputeNetworkFilter(params plugins.Params)
 	httpConnectionManager := h.initializeHCM()
 
 	// 2. Apply HttpFilters
+	var err error
```
sam-heilbron, line 134: I think we can delete this as it is unused

test/e2e/happypath_test.go
sam-heilbron: We can sync tomorrow on this. The example_test (https://github.com/solo-io/gloo/blob/master/test/e2e/example_test.go#L57) provides an example for how we can route to a destination.

It works because in the setup we initialize a server, and an upstream referencing that server: https://github.com/solo-io/gloo/blob/master/test/e2e/test_context.go#L75.

`TestUpstreamReachable()` performs a request that hits a service as well

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -173,6 +174,53 @@ var _ = Describe("Listener Subsystem", func() {
...
+				ExpectWithOffset(1, filterChain.GetFilterChainMatch()).To(BeNil())
+
+				hcmFilter := filterChain.GetFilters()[0]
+				_, err := sslutils.AnyToMessage(hcmFilter.GetConfigType().(*envoy_config_listener_v3.Filter_TypedConfig).TypedConfig)
+				Expect(err).NotTo(HaveOccurred())
+
+				hcm := &envoy_http_connection_manager_v3.HttpConnectionManager{}
+				err = translator.ParseTypedConfig(hcmFilter, hcm)
+				ExpectWithOffset(1, err).NotTo(HaveOccurred())
+				Expect(hcm.HttpFilters).To(HaveLen(2))
```
sam-heilbron, line 218: We should be consistent with the Offset we use in a test. https://onsi.github.io/gomega/#making-assertions-in-helper-functions

```
To get around this, Gomega provides versions of Expect, Eventually and Consistently named ExpectWithOffset, EventuallyWithOffset and ConsistentlyWithOffset that allow you to specify an offset in the call stack. 
```

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -173,6 +174,53 @@ var _ = Describe("Listener Subsystem", func() {
...
+					VirtualHostRefs: []string{"vhost-ref"},
+				}},
+			},
+			func(listener *envoy_config_listener_v3.Listener, routeConfigs []*envoy_config_route_v3.RouteConfiguration) {
+				By("Should be able to add and translate the router to an envoy config")
+				ExpectWithOffset(1, listener.GetFilterChains()).To(HaveLen(1))
+				lf := listener.GetFilterChains()
+				_ = lf
+				filterChain := listener.GetFilterChains()[0]
+				ExpectWithOffset(1, filterChain.GetFilterChainMatch()).To(BeNil())
```
sam-heilbron, line 209: It doesn't hurt to have this, but is this validating something in particular? This seems irrelevant to what we're validating

sam-heilbron (commented):
LGTM! Mostly nits. I think we can clarify some of the testing challenges in a huddle tomorrow

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -209,14 +209,23 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
...
+		wellknown.Router,
+		&routerV3,
+		plugins.AfterStage(plugins.RouteStage),
+	)
+
+	if err != nil {
+		validation.AppendHTTPListenerError(h.report, validationapi.HttpListenerReport_Error_ProcessingError, err.Error())
+	}
+
+	envoyHttpFilters = append(envoyHttpFilters, newStagedFilter.HttpFilter)
```
gunnar-solo, line 228: thanks for the adjustment, here (as opposed to the original iteration).  I enjoy the choice of assembly pattern, rather than having 2 different `return` paths

test/e2e/happypath_test.go
```diff
@@ -184,6 +186,54 @@ var _ = Describe("Happy path", func() {
...
+					Expect(cfg).To(MatchRegexp("\"suppress_envoy_headers\": true"))
+
+				})
+
+				It("it correctly DID NOT pass the suppress envoy headers config", func() {
```
gunnar-solo, line 216: chuckling at this assertion text

test/e2e/happypath_test.go
```diff
@@ -184,6 +186,65 @@ var _ = Describe("Happy path", func() {
...
+						Ctx: ctx,
+					})
+					Expect(err).NotTo(HaveOccurred())
+					TestUpstreamReachable()
+
+					// This will hit the virtual host with the above virtual cluster config
+					response, err := http.Get(fmt.Sprintf("http://%s:%d/", "localhost", defaults.HttpPort))
+					Expect(err).NotTo(HaveOccurred())
+
+					headers := response.Header
```
ianmacclancy, line 236: Extra lines from testing - need to pull these out

ianmacclancy (changes requested):

ianmacclancy (approved):
LGTM!

sam-heilbron (approved):

PR #7638, opened 2023-01-06
Support storing protoDescriptorBin in a ConfigMap

nfuden:
Configmap support for proto descriptors when using grpcjson plugin. This reduces the size of the gateway object definition which previously could break the rollout job.

To use this in helm, old helm values in the shape of:
```
gatewayProxies:
  gatewayProxy:
    gatewaySettings:
      customHttpGateway:
        options:
          grpcJsonTranscoder:
            protoDescriptorBin: <proto-desc-content...>
```
should be changed to:
```
gatewayProxies:
  gatewayProxy:
    gatewaySettings:
      customHttpGateway:
        options:
          grpcJsonTranscoder:
            protoDescriptorConfigMap:
              configMapRef:
                name: my-config-map
                namespace: gloo-system  # optional, defaults to install namespace
              key: my-key  # optional if there is only one key in the configmap
global:
  configMaps:
  - name: my-config-map
     namespace: gloo-system
     data:
       my-key: <proto-desc-content...>
```


projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -100,35 +118,46 @@ func translateGlooToEnvoyPrintOptions(options *grpc_json.GrpcJsonTranscoder_Prin
...
 }
 
-func translateConfigMapToProtoBin(ctx context.Context, snap *gloosnapshot.ApiSnapshot, configRef *grpc_json.GrpcJsonTranscoder_DescriptorConfigMap) (out []byte) {
+// get the proto descriptor data from a ConfigMap
+func translateConfigMapToProtoBin(ctx context.Context, snap *gloosnapshot.ApiSnapshot, configRef *grpc_json.GrpcJsonTranscoder_DescriptorConfigMap) ([]byte, error) {
+	if configRef.GetConfigMapRef() == nil {
+		return nil, NoConfigMapRefError()
+	}
 
+	// make sure the referenced configmap exists and has data
```
nfuden, line 127: Nit:Exists in gloo snapshot

jenshu:
undrafting so ci runs

install/helm/gloo/generate/values.go
```diff
@@ -552,10 +553,6 @@ type AccessLogger struct {
 	*DeploymentSpec
 }
 
-type GatewayProxyConfigMap struct {
```
jenshu, line 555: this struct was not used/referenced anywhere

test/e2e/grpc_json_test.go
```diff
@@ -92,69 +89,119 @@ var _ = Describe("GRPC to JSON Transcoding Plugin - Envoy API", func() {
 		}
 	}
 
-	It("Routes to GRPC Functions", func() {
+	testRequest := func(path string, shouldMatch bool) {
+		body := []byte(`"foo"`) // this is valid JSON because of the quotes
```
elcasteel, line 93: nit: `"foo"` could be a constant that we use in all the comparisons.

elcasteel: Although now I see that this is moved rather than new so even less pressing.

jenshu: 2a2e564

jenshu (approved):
approving on behalf of @nfuden

PR #7647, opened 2023-01-09
build: allow VERSION var to be set in local dev

gunnar-solo:
# Description
Changed a _very heavily used_ Makefile variable, `VERSION`, from being always hardcoded to `0.0.1-fork` when doing local development

# Context
I was chasing a subtle bug when working on some EE actions, and found that I could not build arbitrary versions on the `1.12.x` OSS branch


nfuden (commented):
As part of validation for this issue let's open a pr from a fork and make sure there isn't anything we are missing

gunnar-solo:
replaced by https://github.com/solo-io/gloo/pull/7650

PR #7658, opened 2023-01-11
Add authentication to Vault via AWS IAM and EC2

jbohanon:
# Description

This new feature can be used to authenticate to HashiCorp Vault via AWS when Vault is used as secret store.

# Validation

Run a development Vault server with docker 
```shell
docker run -d \
--cap-add=IPC_LOCK \
-e 'VAULT_LOCAL_CONFIG={"storage": {"file": {"path": "/vault/file"}}, "listener": [{"tcp": { "address": "0.0.0.0:8200", "tls_disable": true}}], "default_lease_ttl": "168h", "max_lease_ttl": "720h", "ui": true}' \
-p 8200:8200 \
--name vault \
--restart unless-stopped \
vault server -dev
```
We can get the root token created for the Vault server via `docker logs vault | grep 'Root Token'`, then exec into the container via `docker exec -it vault /bin/sh` where we can execute commands against the server with the `vault` command. First we will need to set some variables:
```shell
# needed because vault cli defaults this value to https
export VAULT_ADDR=http://127.0.0.1:8200
export VAULT_TOKEN=<root token>
```

Create a development admin policy 
```shell
vault policy write "admin" - <<EOF
path "*" {
  capabilities = ["create", "read", "update", "delete", "list", "patch", "sudo"]
}
EOF
```

Now we can enable and configure the AWS auth method and create a role called `vault-role`
```shell
vault auth enable aws
vault write auth/aws/config/client secret_key=<secret access key> access_key=<access key id>
vault write auth/aws/role/vault-role auth_type=iam bound_iam_principal_arn=<arn> policies=admin
```

It is important the the AWS user represented by <arn> have the appropriate AWS policy. A recommended policy can be found in the guide [here](https://developer.hashicorp.com/vault/docs/auth/aws#recommended-vault-iam-policy). 

It appears that using the IAM method, the value for the `bound_iam_principal_arn` must be or include the exact principal ARN that is provided in the request. That is to say, even if an AWS role trusts the AWS user, we cannot provide the user to the vault login method if the only the role is specified in `bound_iam_principal_arn`.

Create a test secret called creds with a single value where key=password and vault = foo: `vault kv put -mount=secret creds password=foo`

Now we can exit the container's shell, and we should be able to access the vault programmatically in Go as long as we set our environment variables correctly. A test using the Go client can be performed with the following:

<details>
<summary>Click for Go example</summary>

```go
package main

import (
	"context"
	"fmt"
	"os"

	vault "github.com/hashicorp/vault/api"
	auth "github.com/hashicorp/vault/api/auth/aws"
)

func main() {
        // make sure we have the correct values in the environment variables
        // this is not necessary if they are set in the calling process or if using a credentials file
	os.Setenv("AWS_ACCESS_KEY_ID", "<access key id>")
	os.Setenv("AWS_SECRET_ACCESS_KEY", "<secret access key>")
	os.Setenv("AWS_SESSION_TOKEN", "<session token>")
	
        fmt.Println(getSecretWithAWSAuthIAM())
}

// Fetches a key-value secret (kv-v2) after authenticating to Vault via AWS IAM,
// one of two auth methods used to authenticate with AWS (the other is EC2 auth).
func getSecretWithAWSAuthIAM() interface{} {
	config := vault.DefaultConfig() // modify for more granular configuration
	config.Address = "http://localhost:8200"

	client, err := vault.NewClient(config)
	if err != nil {
		return fmt.Errorf("unable to initialize Vault client: %w", err)
	}

	awsAuth, err := auth.NewAWSAuth(
		auth.WithRole("vault-role"), // must match the name of the vault role previously created
	)
	if err != nil {
		return fmt.Errorf("unable to initialize AWS auth method: %w", err)
	}

        // the SDK called here relies on one of a few different credentials methods. see https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config
	authInfo, err := client.Auth().Login(context.Background(), awsAuth)
	if err != nil {
		return fmt.Errorf("unable to login to AWS auth method: %w", err)
	}
	if authInfo == nil {
		return fmt.Errorf("no auth info was returned after login")
	}

	// get secret from the default mount path for KV v2 in dev mode, "secret"
	secret, err := client.KVv2("secret").Get(context.Background(), "creds")
	if err != nil {
		return fmt.Errorf("unable to read secret: %w", err)
	}

	// data map can contain more than one key-value pair,
	// in this case we're just grabbing one of them
	value, ok := secret.Data["password"].(string)
	if !ok {
		return fmt.Errorf("value type assertion failed: %T %#v", secret.Data["password"], secret.Data["password"])
	}

	return value
}
```
</details>

# Checklist:


jbohanon:
Separating work into separate PRs for IAM and EC2 implementations.

PR #7669, opened 2023-01-13
Add authentication to Vault via AWS IAM

jbohanon:
# Description

This new feature can be used to authenticate to HashiCorp Vault via AWS when Vault is used as secret store.

# Validation

Run a development Vault server with docker 
```shell
docker run -d \
--cap-add=IPC_LOCK \
-e 'VAULT_LOCAL_CONFIG={"storage": {"file": {"path": "/vault/file"}}, "listener": [{"tcp": { "address": "0.0.0.0:8200", "tls_disable": true}}], "default_lease_ttl": "168h", "max_lease_ttl": "720h", "ui": true}' \
-p 8200:8200 \
--name vault \
--restart unless-stopped \
vault server -dev
```
We can get the root token created for the Vault server via `docker logs vault | grep 'Root Token'`, then exec into the container via `docker exec -it vault /bin/sh` where we can execute commands against the server with the `vault` command. First we will need to set some variables:
```shell
# needed because vault cli defaults this value to https
export VAULT_ADDR=http://127.0.0.1:8200
export VAULT_TOKEN=<root token>
```

Create a development admin policy 
```shell
vault policy write "admin" - <<EOF
path "*" {
  capabilities = ["create", "read", "update", "delete", "list", "patch", "sudo"]
}
EOF
```

Now we can enable and configure the AWS auth method and create a role called `vault-role`
```shell
vault auth enable aws
vault write auth/aws/config/client secret_key=<secret access key> access_key=<access key id>
vault write auth/aws/role/vault-role auth_type=iam bound_iam_principal_arn=<arn> policies=admin
```

It is important the the AWS user represented by <arn> have the appropriate AWS policy. A recommended policy can be found in the guide [here](https://developer.hashicorp.com/vault/docs/auth/aws#recommended-vault-iam-policy). 

It appears that using the IAM method, the value for the `bound_iam_principal_arn` must be or include the exact principal ARN that is provided in the request. That is to say, even if an AWS role trusts the AWS user, we cannot provide the user to the vault login method if the only the role is specified in `bound_iam_principal_arn`.

Create a test secret called creds with a single value where key=password and vault = foo: `vault kv put -mount=secret creds password=foo`

Now we can exit the container's shell, and we should be able to access the vault programmatically in Go as long as we set our environment variables correctly. A test using the Go client can be performed with the following:

<details>
<summary>Click for Go example</summary>

```go
package main

import (
	"context"
	"fmt"
	"os"

	vault "github.com/hashicorp/vault/api"
	auth "github.com/hashicorp/vault/api/auth/aws"
)

func main() {
        // make sure we have the correct values in the environment variables
        // this is not necessary if they are set in the calling process or if using a credentials file
	os.Setenv("AWS_ACCESS_KEY_ID", "<access key id>")
	os.Setenv("AWS_SECRET_ACCESS_KEY", "<secret access key>")
	os.Setenv("AWS_SESSION_TOKEN", "<session token>")
	
        fmt.Println(getSecretWithAWSAuthIAM())
}

// Fetches a key-value secret (kv-v2) after authenticating to Vault via AWS IAM,
// one of two auth methods used to authenticate with AWS (the other is EC2 auth).
func getSecretWithAWSAuthIAM() interface{} {
	config := vault.DefaultConfig() // modify for more granular configuration
	config.Address = "http://localhost:8200"

	client, err := vault.NewClient(config)
	if err != nil {
		return fmt.Errorf("unable to initialize Vault client: %w", err)
	}

	awsAuth, err := auth.NewAWSAuth(
		auth.WithRole("vault-role"), // must match the name of the vault role previously created
	)
	if err != nil {
		return fmt.Errorf("unable to initialize AWS auth method: %w", err)
	}

        // the SDK called here relies on one of a few different credentials methods. see https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config
	authInfo, err := client.Auth().Login(context.Background(), awsAuth)
	if err != nil {
		return fmt.Errorf("unable to login to AWS auth method: %w", err)
	}
	if authInfo == nil {
		return fmt.Errorf("no auth info was returned after login")
	}

	// get secret from the default mount path for KV v2 in dev mode, "secret"
	secret, err := client.KVv2("secret").Get(context.Background(), "creds")
	if err != nil {
		return fmt.Errorf("unable to read secret: %w", err)
	}

	// data map can contain more than one key-value pair,
	// in this case we're just grabbing one of them
	value, ok := secret.Data["password"].(string)
	if !ok {
		return fmt.Errorf("value type assertion failed: %T %#v", secret.Data["password"], secret.Data["password"])
	}

	return value
}
```
</details>

# Checklist:


projects/gloo/pkg/bootstrap/vault.go
```diff
@@ -20,62 +27,159 @@ func NewVaultSecretClientFactory(client *api.Client, pathPrefix, rootKey string)
...
 		}
-		tlsCfg.Insecure = insecure.GetValue()
+		client.SetToken(token)
+		return client, nil
 	}
+}
 
-	client, err := api.NewClient(cfg)
+func configureAwsAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*api.Client, error) {
+	return configureAwsIamAuth(aws, client)
```
ben-taussig-solo, line 127: So the idea here is that we could potentially expand `configureAwsAuth` to handle multiple types of AWS authentication? Do you expect that we would introduce a new `AuthMethod` on `v1.Settings_VaultSecrets` for a new AWS auth scheme? In that case I think we should consider renaming `v1.Settings_VaultSecrets_Aws` to `v1.Settings_VaultSecrets_AwsIam` and calling directly `configureAwsIamAuth` from `configureVaultAuth`. If not then this looks good to me!

jbohanon: So the API changes I anticipate if we are to support EC2 in addition to IAM can be seen in [this PR](https://github.com/solo-io/gloo/pull/7672/files). I was thinking of leaving the `v1.Settings_VaultSecrets_Aws` and adding a field to it to specify AWS-specific `login_method`. Correct, this is the reason for the indirect call to `configureAwsIamAuth` via `configureAwsAuth`.

In any case, it would be worth commenting why this indirection exists

projects/gloo/pkg/bootstrap/vault.go
```diff
@@ -20,62 +27,159 @@ func NewVaultSecretClientFactory(client *api.Client, pathPrefix, rootKey string)
...
+	awsAuth, err := awsauth.NewAWSAuth(loginOptions...)
 	if err != nil {
 		return nil, err
 	}
-	if tlsCfg != nil {
-		if err := cfg.ConfigureTLS(tlsCfg); err != nil {
-			return nil, err
-		}
+
+	// TODO(jbohanon) set up auth token refreshing with client.NewLifetimeWatcher()
```
nfuden, line 175: So this should purely be for where we reload settings and or if we are pointing to secret info in a kubernetes secret right?

jbohanon: No, this is more to refresh the token that authenticates to Vault which we get through the AWS auth process. The client should handle this via the `NewLifetimeWatcher` method mentioned. It protects against Vault revoking the token or it expiring. The token expiry is defined when the Vault role which the client is trying to login as is [configured](https://developer.hashicorp.com/vault/api-docs/auth/aws#create-role).

projects/gloo/pkg/bootstrap/vault.go
```diff
@@ -20,62 +27,159 @@ func NewVaultSecretClientFactory(client *api.Client, pathPrefix, rootKey string)
...
 		}
-		tlsCfg.TLSServerName = tlsServerName
+		addSettingFunc(b.GetValue())
 	}
-	if insecure := vaultSettings.GetInsecure(); insecure != nil {
-		if tlsCfg == nil {
-			tlsCfg = &api.TLSConfig{}
+
+	// Add our settings to the vault TLS config, preferring settings set in the
+	// new TlsConfig field to those in the deprecated fields
```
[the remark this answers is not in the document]
jbohanon: I agree. Would it be more appropriate perhaps to _first_ nil-check the `TlsConfig` field and set based on deprecated if nil, otherwise use the new field's values?

projects/gloo/pkg/bootstrap/vault.go
```diff
@@ -20,62 +27,159 @@ func NewVaultSecretClientFactory(client *api.Client, pathPrefix, rootKey string)
...
+	// this is a secret value and should not exist in the Settings in plaintext
+	if secretAccessKey := aws.GetSecretAccessKey(); secretAccessKey == "" {
+		return nil, errors.New("secret access key must be defined for AWS IAM auth")
+	} else {
+		os.Setenv("AWS_SECRET_ACCESS_KEY", secretAccessKey)
+	}
+
+	loginOptions := []awsauth.LoginOption{awsauth.WithIAMAuth()}
+
+	if role := aws.GetRole(); role != "" {
```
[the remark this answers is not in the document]
jbohanon: Actually, renaming the import here would probably make it clearer without adding redundantly specific naming in the proto

projects/gloo/api/v1/settings.proto
```diff
@@ -83,45 +83,101 @@ message Settings {
...
+    // For more info see https://developer.hashicorp.com/vault/docs/auth/aws
+    message VaultAwsAuth {
+        // The Vault role we are trying to authenticate to.
+        // This is not necessarily the same as the AWS role to which the Vault role is configured.
+        string vault_role = 1;
+        string region = 2;
+        string iam_server_id_header = 3;
+        string mountPath = 4;
+        string access_key_id = 5;
+        // This is a secret but can't be a secret ref because we are configuring the secrets engine...
```
jbohanon, line 152: remove

projects/gloo/api/v1/settings.proto
```diff
@@ -83,45 +83,101 @@ message Settings {
...
+        string secret_access_key = 6;
+        // This is a secret but can't be a secret ref because we are configuring the secrets engine...
```
jbohanon, line 154: remove

test/e2e/vault_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/factory"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+
+	vaultapi "github.com/hashicorp/vault/api"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = FDescribe("Vault e2e", func() {
```
jbohanon, line 24: Remove test focus

test/services/vault.go
```diff
@@ -152,14 +168,51 @@ func (i *VaultInstance) Token() string {
...
 
+func (i *VaultInstance) EnableAWSAuthMethod(settings *v1.Settings_VaultSecrets) error {
+	// Enable the AWS auth method
+	_, err := i.Exec("auth", "enable", "aws")
+	if err != nil {
+		return err
+	}
+
+	// Add our admin policy
+	tmpFileName := fmt.Sprintf("%s/policy.json", i.tmpdir)
```
jbohanon, line 191: nit: use `filepath.Join()`

projects/gloo/api/v1/settings.proto
```diff
@@ -83,45 +83,99 @@ message Settings {
...
+        }
+    }
+
+    // Configure Vault client to authenticate to server via AWS auth (IAM only).
+    // For more info see https://developer.hashicorp.com/vault/docs/auth/aws
+    message VaultAwsAuth {
+        // The Vault role we are trying to authenticate to.
+        // This is not necessarily the same as the AWS role to which the Vault role is configured.
+        string vault_role = 1;
+        string region = 2;
```
ben-taussig-solo, line 148: Can you add descriptions for the fields here?

projects/gloo/api/v1/settings.proto
```diff
@@ -83,45 +83,99 @@ message Settings {
...
+        string iam_server_id_header = 3;
+        string mountPath = 4;
```
ben-taussig-solo, line 150: Nit: should this be `mount_path` to match the other fields here?

test/e2e/vault_test.go
```diff
@@ -0,0 +1,172 @@
...
+		cancel        context.CancelFunc
+		vaultInstance *services.VaultInstance
+		err           error
+	)
+
+	BeforeEach(func() {
+		Skip("until AWS creds are in cloudbuild")
+
+		helpers.ValidateRequirementsAndNotifyGinkgo(
+			helpers.Vault(),
```
ben-taussig-solo, line 36: Not fully in the loop on our new requirements, but I think we should probably add `helpers.AwsCredentials()` here

test/e2e/vault_test.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		eventuallyReadsTestSecret := func() {
+			Eventually(func(g Gomega) {
+				sec, err := secretClient.Read("gloo-system", "test-secret", clients.ReadOpts{}.WithDefaults())
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(sec.String()).NotTo(BeEmpty())
+			}, "5s", ".5s").Should(Succeed())
+		}
+
+		testWriteSecret := &gloov1.Secret{
```
ben-taussig-solo, line 76: nit: I think the name of this secret is a little confusing because we also have the `writeTestSecret` method -- could you leave a comment clarifying the purpose of this secret?

test/e2e/vault_test.go
```diff
@@ -0,0 +1,172 @@
...
+			BeforeEach(func() {
+				vaultFactoryConfig = &services.VaultFactoryConfig{}
+				startVault()
+				settings = &gloov1.Settings_VaultSecrets{
+					Address: "http://localhost:8200",
+					AuthMethod: &gloov1.Settings_VaultSecrets_Aws{
+						Aws: &gloov1.Settings_VaultAwsAuth{
+							VaultRole:       "vault-role",
+							Region:          "us-east-1",
+							AccessKeyId:     os.Getenv("AWS_ACCESS_KEY_ID"),
```
ben-taussig-solo, line 129: Do you have to set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in your environment to run these tests? If so we should consider adapting what we do in the AWS tests here https://github.com/solo-io/gloo/blob/276bf78fbf133bfd2b9345b4b75823d68853bead/test/e2e/aws_test.go#L541-L550 to extract credentials from the local filesystem. If we do this, the vault tests should run properly in CI, although broadly speaking AWS tests will fail, as the default Access key ID/Secret Key have been updated in CI, but certain other AWS data has not

jbohanon: Yeah so the opacity in the client libraries gave me a bit of a headspin here. We should be able to use that `credentials.NewSharedCredentials("", "")` to access the values regardless if they are in the shared creds file or the environment variables, correct? So getting them from here will harden against different creds injection methods?

test/services/vault.go
```diff
@@ -5,32 +5,40 @@ import (
...
+	"github.com/solo-io/go-utils/log"
 
 	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega/gexec"
 )
 
-const defaultVaultDockerImage = "vault:1.1.3"
+const defaultVaultDockerImage = "vault:1.12.2"
 const defaultAddress = "127.0.0.1:8200"
+const DefaultVaultToken = "root"
```
nfuden, line 20: nit: Probably shouldnt be exported?

projects/gloo/api/v1/settings.proto
ben-taussig-solo: Update looks good!

ben-taussig-solo (approved):
Looks good, I approve! CI will not pass until this PR goes through: https://github.com/solo-io/gloo/pull/7698 .

test/services/vault.go
jbohanon: This was originally exported to be used in the test. I clearly forgot to do that and hardcoded it in the test... if it is blocking I can either remove the export or use the value in the test

jbohanon: I elected to use the variable 91dde48e97377d894f80df9fe205b802032a7a47

ben-taussig-solo (approved):

nfuden (approved):

Issue #7377, opened 2022-10-25
Make http port optional in K8s manifests


guydc:
### Version

1.12.x (latest stable)

### Is your feature request related to a problem? Please describe.

Currently, the gloo edge chart enables and http port (default: 80) on both gateway-proxy service and pods:

- https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/8-gateway-proxy-service.yaml
- https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L227

While it's possible to disable the envoy http listener, the existence of the additional http port in k8s manifests leads to the creation of undesirable resources (e.g. hyperscaler firewall rules). 

### Describe the solution you'd like

It should be possible to disable the http port in the proxy service and pod manifests with appropriate helm chart values.

kitarp29:
Can I take this up? @guydc @bobzoller 
Please assign it to me.

SuperAayush:
Hey @guydc @chrisgaun!!
I would love to help with this issue, can you please assign it to me?
Also can you give me a lead on how can I get things easy for me?

nfuden:
@SuperAayush we are happy to talk implementation details with you on this one it has a few moving parts that could be hard to hit without a measured survey. If you want we could talk through the process used once we complete this or if you are in our open source slack drop me a dm and we can talk about what you might want to do / if there is any other guidance we can give you. 
I admit to not checking the slack messages there often enough but it might be our best communication strategy.
Of course if you feel strongly about this issue in particular (the above text was in case it wasnt about the particular problem) we definitely can instead figure out how to best help this be tackled instead.

irateswami:
👀

SuperAayush:
> 

This is really nice of you @nfuden, I have dropped a message to you on Slack.
Thank You!!

Bslabe123:
As per standup heres the proposal:

# Problem Summary:
When we create the k8s manifest for gloo via helm we expose all ports that could be used by default. In certain situations it may be prudent to not allow traffic on certain ports (i.e. HTTPS but no HTTP traffic). There also may be situations in which leaving a port open may lead to unneeded resources being created (as mentioned in the issue, hyperscaler firewall rules). We would like to introduce a toggle/convention to allow anyone building gloo helm charts to disable the exposing of certain ports.
# Design:
[...]

PR #7687, opened 2023-01-17
Optional Ports for gateway-proxy pods and services

Bslabe123:
# Description

When we create Gloos k8s manifest via helm we expose all ports on gateway-proxy pods/services by default. In certain situations it may be prudent to not allow traffic on certain ports (i.e. HTTPS but no HTTP traffic). There also may be situations in which leaving a port open may lead to unneeded resources being created (as mentioned in the issue, hyperscaler firewall rules). We would like to introduce a toggle/convention to allow anyone building helm charts to disable the exposing of certain ports on gateway-proxy pods/services.

# Context

We have decided to make 0 a reserved value indicating that the port should be disabled, when rendering the helm chart we wrap the port declaration in an if(port != 0)/else statement. We also call this out loudly in the field description since its a less explicit toggle.

# Checklist:

### DIFF
--- a/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
+++ b/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
@@ -236,6 +236,7 @@ spec:
             - NET_BIND_SERVICE
           {{- end}}
         ports:
+{{- if $spec.podTemplate.httpPort }}
         - containerPort: {{ $spec.podTemplate.httpPort }}
           name: http
           protocol: TCP
@@ -244,6 +245,8 @@ spec:
           hostPort: {{ $spec.podTemplate.httpPort }}
           {{- end}}
           {{- end}}
+{{- end}}
+{{- if $spec.podTemplate.httpsPort}}
         - containerPort: {{ $spec.podTemplate.httpsPort }}
           name: https
           protocol: TCP
@@ -252,6 +255,7 @@ spec:
           hostPort: {{ $spec.podTemplate.httpsPort }}
           {{- end}}
           {{- end}}
+{{- end}}
        {{- if and $statsConfig.enabled $statsConfig.podMonitorEnabled }}
         - name: http-monitoring
           containerPort: 8081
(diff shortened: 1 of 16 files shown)


ben-taussig-solo (approved):

install/helm/gloo/templates/8-gateway-proxy-service.yaml
```diff
@@ -34,7 +34,7 @@ spec:
   # port order matters due to this issue: https://github.com/solo-io/gloo/issues/2571
   ports:
 {{- if $spec.service.httpsFirst }}
-  {{- if not $gatewaySettings.disableHttpsGateway }}
+  {{- if and (not $gatewaySettings.disableHttpsGateway) $spec.service.httpsPort }}
```
sam-heilbron, line 37: Is it the service httpsPort or the podTemplate httpsPort that should be part of the conditional?

install/helm/gloo/templates/8-gateway-proxy-service.yaml
```diff
@@ -139,9 +139,11 @@ spec:
   selector:
     gloo: gateway-proxy
     gateway-proxy-id: {{ $name | kebabcase }}
+{{- if and .Values.global.glooStats.enabled .Values.global.glooStats.serviceMonitorEnabled }}
```
sam-heilbron, line 142: Is this related to the http(s) port disabling or just an additional change?

sam-heilbron (changes requested):
@jenshu raises a good point about the relationship between `disableHttpGateway` and setting httpPort=0 on gateway-proxy. Why do we need both flags?

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -361,6 +364,7 @@ spec:
           initialDelaySeconds: 3
           periodSeconds: 10
           failureThreshold: 3
+{{- end }}
```
jenshu, line 367: i think this change should be undone

Bslabe123:
See: https://github.com/solo-io/gloo/issues/7377#issuecomment-1387445788

PR #7692, opened 2023-01-19
Rollout job configmap

jenshu:
This fixes the `argument list too long` errors that can be seen in the resource rollout job during helm installs/upgrades when the Gateway CR yaml gets too long (e.g. when a grpc-json proto descriptor is set). The Gateway yaml is now stored in a ConfigMap which is mounted as a file by the resource rollout job. The job then calls `kubectl apply` on the file, instead of applying the entire yaml inline.

### DIFF
--- a/install/helm/gloo/templates/5-resource-configmap.yaml
+++ b/install/helm/gloo/templates/5-resource-configmap.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gloo-custom-resource-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: gloo
+    gloo: custom-resources
+data:
+  {{- $cr := include "gloo.customResources" . | indent 4 -}}
+  {{- $hasCr := contains "kind:" $cr -}}{{/* whether there are any custom resources */}}
+  custom-resources: |
+    {{- $cr }}
+  has-custom-resources: "{{- $hasCr -}}"
--- a/install/helm/gloo/templates/5-resource-rollout-job.yaml
+++ b/install/helm/gloo/templates/5-resource-rollout-job.yaml
@@ -40,10 +40,26 @@ spec:
       {{- include "gloo.pullSecret" $image | nindent 6 -}}
       serviceAccountName: gloo-resource-rollout
       {{- include "gloo.podSpecStandardFields" .Values.gateway.rolloutJob | nindent 6 -}}
+      volumes:
+        - name: custom-resource-config-volume
+          configMap:
+            name: gloo-custom-resource-config
+            items:
+              - key: custom-resources
+                path: custom-resources
       containers:
         - name: kubectl
           image: {{template "gloo.image" $image}}
           imagePullPolicy: {{ $image.pullPolicy }}
+          env:
+            - name: HAS_CUSTOM_RESOURCES
+              valueFrom:
+                configMapKeyRef:
+                  name: gloo-custom-resource-config
+                  key: has-custom-resources
+          volumeMounts:
+            - name: custom-resource-config-volume
+              mountPath: /etc/gloo-custom-resources
           securityContext:
             runAsNonRoot: true
             {{- if not .Values.gateway.rolloutJob.floatingUserId }}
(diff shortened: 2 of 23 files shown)


install/helm/gloo/templates/5-resource-rollout-job.yaml
```diff
@@ -40,10 +40,26 @@ spec:
       {{- include "gloo.pullSecret" $image | nindent 6 -}}
       serviceAccountName: gloo-resource-rollout
       {{- include "gloo.podSpecStandardFields" .Values.gateway.rolloutJob | nindent 6 -}}
+      volumes:
```
nfuden, line 43: Does the role have access to the config map by default? Is there anything we need to callout for customers who may take a more locked down approach to rbac? 
Since the configmap creation is handled by helm it shouldnt be an issue but accessing it may?

install/helm/gloo/templates/5-resource-configmap.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gloo-custom-resource-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: gloo
+    gloo: custom-resources
+data:
+  {{- $cr := include "gloo.customResources" . | indent 4 -}}
```
gunnar-solo, line 10: trying to understand, here.  If this block is defining `cr :=`, why is it indented?

install/test/helm_test.go
```diff
@@ -2317,17 +2302,16 @@ spec:
...
 						})
 						It("uses merged values for the gateway", func() {
-							job := getJob(testManifest, namespace, "gloo-resource-rollout")
-							Expect(job.Spec.Template.Spec.Containers[0].Command[2]).To(ContainSubstring(`apiVersion: gateway.solo.io/v1
+							expectCustomResourceSubstring(`apiVersion: gateway.solo.io/v1
 kind: Gateway
 metadata:
   name: another-gateway-proxy
-  namespace: ` + namespace + `
+  namespace: `+namespace+`
```
gunnar-solo, line 2309: help me out--what are these `+` signs?

install/test/helm_test.go
```diff
@@ -4183,13 +4162,24 @@ metadata:
 							Expect(job.Spec.Template.Spec.Containers[0].Command[2]).NotTo(ContainSubstring("kubectl rollout"))
 						})
 
-						It("does not call kubectl apply when default gateways are disabled", func() {
+						It("configmap should contain gateway yaml when default gateways are enabled", func() {
+							prepareMakefile(namespace, helmValues{valuesArgs: []string{}})
```
gunnar-solo, line 4166: could we explicitly set `gatewaySettings.enabled=true`, here?

Right now, this test relies upon a default behavior which "might" change

gunnar-solo (commented):
a couple questions/nit

install/test/helm_test.go
```diff
@@ -99,6 +99,15 @@ var _ = Describe("Helm Test", func() {
 			statsAnnotations map[string]string
 		)
 
+		expectCustomResourceSubstring := func(substr string, contains bool) {
```
sheidkamp, line 102: nice

sheidkamp (approved):
LGTM!

install/helm/gloo/templates/5-resource-configmap.yaml
jenshu: the output of the `gloo.customResources` helper function is the Gateway yaml and needs to be stuffed into the value of the `custom-resources` key (which has indentation 2) below. The inline yaml seems to need indentation 4 in order to be treated as all part of the configmap value (otherwise you get syntax errors). See the [mtls configmap](https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/19-gloo-mtls-configmap.yaml) for another example - all of the envoy yaml there is indented 4 spaces

install/test/helm_test.go
jenshu: they're just concatenating the namespace (variable) to the yaml string

install/test/helm_test.go
jenshu: good call, i can do that

jenshu: 1f8597e

install/helm/gloo/templates/5-resource-rollout-job.yaml
jenshu: as far as i can tell, if we're just referencing the configmap via built-in fields on the k8s resource (as we are here), we don't need any additional permissions. we only need to add permissions to the serviceaccount's role if we're calling k8s apis directly (e.g. via the `kubectl` commands below)

gunnar-solo (approved):

sheidkamp (approved):

jenshu:
/kick
```
[Fail] Consul e2e with envoy and gloo [It] resolves consul services with hostname addresses (as opposed to IPs addresses) 
```

Issue #7149, opened 2022-09-14
Expose http health check method from envoy in Gloo API


bdecoste:
### Version

1.11.x

### Is your feature request related to a problem? Please describe.

I'd like to expose the method param from envoy's http cluster health check in the Gloo API so I can send health checks other than GET:

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#envoy-v3-api-msg-config-core-v3-healthcheck-httphealthcheck

https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/external/envoy/api/v2/core/health_check.proto.sk/#httphealthcheck

### Describe the solution you'd like

expose the method param from envoy's http cluster health check in the Gloo API

### Describe alternatives you've considered

None

PR #7695, opened 2023-01-19
Expose `config.core.v3.HealthCheck.HttpHealthCheck.method` as `v3.HealthCheck.HttpHealthCheck.method`

Bslabe123:
# Description

Whenever we do health checks via http since we have no way to set the method, we default to the GET method. We would like to expose envoys `config.core.v3.HealthCheck.HttpHealthCheck.method` to allow us to set the method used when performing health checks

# Context

This new field `v3.HealthCheck.HttpHealthCheck.method` corresponds to envoys `config.core.v3.HealthCheck.HttpHealthCheck.method`. Currently its functionality is verified for POST and GET methods. We have a negative test for CONNECT as well.

NOTE: Some corrections have been made to unit tests, particularly around changing `Equals()` matchers to `MatchProto()` matchers wherever appropriate. This is because we compare the state field which is not guaranteed to be equivalent.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7149


Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=79d1b0b4-98e7-46b6-9007-8ed2094d3619

ben-taussig-solo:
In the test, can we validate that the request the upstream receives uses the specified method?

projects/gloo/api/external/envoy/api/v2/core/health_check.proto
```diff
@@ -136,6 +137,13 @@ message HealthCheck {
         // (Enterprise Only): If defined, the response health check rules take precedence over the http
         // `expected_statuses`.
         .advancedhttp.options.gloo.solo.io.ResponseAssertions response_assertions = 10;
+
+
+        // HTTP Method that will be used for health checking, default is "GET".
+        // GET, HEAD, POST, PUT, DELETE, OPTIONS, TRACE, PATCH methods are supported, but making request body is not supported.
+        // CONNECT method is disallowed because it is not appropriate for health check request.
+        // If a non-200 response is expected by the method, it needs to be set in :ref:`expected_statuses <envoy_v3_api_field_config.core.v3.HealthCheck.HttpHealthCheck.expected_statuses>`.
+        solo.io.envoy.config.core.v3.RequestMethod method = 11 [(validate.rules).enum = {defined_only: true not_in: 6}];
```
sam-heilbron, line 146: what is the purpose of the "not in" part here?

test/helpers/default_gateways.go
```diff
@@ -18,3 +18,16 @@ func WriteDefaultGateways(writeNamespace string, gatewayClient v1.GatewayClient)
 
 	return err
 }
+
+func WriteDefaultGatewaysWithHealthChecks(writeNamespace string, gatewayClient v1.GatewayClient) error {
```
sam-heilbron, line 22: Similar to other default thoughts. I think we can use the default gateway, add the extensions and write it in the test. I think the creation and writing are two separate concepts, and we can have a utility that writes a list of gateways, but we don't need unique utilities to build and write each variation we would want

test/e2e/health_checks_test.go
```diff
@@ -181,6 +197,122 @@ var _ = Describe("Health Checks", func() {
 			})
 		}
 
+		Context("passes health checks with different methods", func() {
+			BeforeEach(func() {
+				envoyInstance.EnablePanicMode()
```
sam-heilbron, line 202: nit: I think a comment on this line would be helpful as this setup is a critical part of testing health checks but not something we use for most tests

test/e2e/health_checks_test.go
```diff
@@ -181,6 +197,122 @@ var _ = Describe("Health Checks", func() {
...
+
+				var hosts []*static_plugin_gloo.Host
+				hosts = append(hosts, &static_plugin_gloo.Host{
```
sam-heilbron, line 205: This works! An alternative would be to use our TestUpstream concept https://github.com/solo-io/gloo/blob/master/test/v1helpers/test_upstream.go#L102 which wraps a locally running server with the upstream CR configuration.


https://github.com/solo-io/gloo/blob/master/test/e2e/test_context.go#L75 has an example of creating one, and https://github.com/solo-io/gloo/blob/master/test/e2e/test_context.go#L87 demonstrates how to create a simple VS that routes to it

test/e2e/health_checks_test.go
```diff
@@ -181,6 +197,122 @@ var _ = Describe("Health Checks", func() {
...
+						HealthChecker: &envoy_config_core_v3.HealthCheck_HttpHealthCheck_{
+							HttpHealthCheck: &envoy_config_core_v3.HealthCheck_HttpHealthCheck{
+								Path:   "some-invalid-path",
+								Method: envoy_config_core_v3.RequestMethod_GET,
+							},
+						},
+					},
+				},
+			}
+			for _, envoyHealthCheckTest := range tests {
```
sam-heilbron, line 301: These are awesome! I think we could convert this into a Table driven test: https://pkg.go.dev/github.com/onsi/ginkgo/extensions/table

sam-heilbron (commented):
Loooking great! Mostly some comments around structuring the tests/code, happy to huddle in person if you have questions

test/e2e/health_checks_test.go
```diff
@@ -39,9 +44,9 @@ var _ = Describe("Health Checks", func() {
 	)
 
 	BeforeEach(func() {
-		helpers.ValidateRequirementsAndNotifyGinkgo(
-			helpers.LinuxOnly("Relies on FDS"),
-		)
+		//helpers.ValidateRequirementsAndNotifyGinkgo(
```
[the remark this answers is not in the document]
Bslabe123: Good callout! The linux check now happens in the before each of every subset of tests except the ones I added

projects/gloo/api/external/envoy/api/v2/core/health_check.proto
Bslabe123: Since this field is just a passthrough, the description and field type is copy-pasted from https://github.com/envoyproxy/envoy/blob/91e048a6fef965390a2cc65963a37a086fd8217a/api/envoy/config/core/v3/health_check.proto#L161-L165. Theres more specifics about semantics here as well https://github.com/bufbuild/protoc-gen-validate

test/e2e/health_checks_test.go
```diff
@@ -39,10 +43,6 @@ var _ = Describe("Health Checks", func() {
 	)
 
 	BeforeEach(func() {
-		helpers.ValidateRequirementsAndNotifyGinkgo(
```
sam-heilbron, line 42: Love the idea to move this to a more focused subset of the tests! 

My one concern is that FDS is enabled in the RunOptions, which occurs in this outer beforeEach. If FDS is not needed for a test, we may need to change up the setup code.

Perhaps we do the following:
- Undo this change in the current PR, so the PR only contains the health check changes
- Comment on the PR how someone can run the focused health check tests locally (setting INVALID_TEST_REQS=run will allow us to bypass this ValidateRequirements method)
- We huddle and create a follow up PR which moves the validation to more focused tests, and we only run FDS for tests that require it

test/e2e/health_checks_test.go
```diff
@@ -209,9 +211,85 @@ var _ = Describe("Health Checks", func() {
 		})
 	})
 
+	table.DescribeTable("translates and persists health checkers with different methods",
+		func(check *envoy_config_core_v3.HealthCheck,
+			expectedConfig string,
+			expectedLog string) {
+			tu = v1helpers.NewTestHttpUpstream(ctx, envoyInstance.LocalAddr())
```
sam-heilbron, line 218: Love this test! It appears to do the following:
1. Create a test upstream server (TestUpstream creates both the server and the upstream resource)
2. Update the upstream health checks
3. Write the upstream
4. Wait for it to be accepted
5. Create a VS to that upstream
6. Write it and wait for it to be accepted
7. Perform some assertions


I think we can simplify this to do the following:
1. Before the table, create the necessary upstream, vs
2. In the table, patch the Upstream using PatchResource: https://github.com/solo-io/gloo/blob/master/test/helpers/util.go#L16
3. Assert that Eventually the config dump contains the expected regex

I think by only containing the changes that are relevant to the table test (ie patching the upstream) its easier to follow what we are actually testing.

test/e2e/health_checks_test.go
```diff
@@ -209,9 +211,85 @@ var _ = Describe("Health Checks", func() {
...
+			_, err = testClients.VirtualServiceClient.Write(vs, clients.WriteOpts{})
+			Expect(err).NotTo(HaveOccurred())
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				return testClients.VirtualServiceClient.Read(vs.Metadata.Namespace, vs.Metadata.Name, clients.ReadOpts{})
+			})
+
+			envoyConfig, err := envoyInstance.ConfigDump()
+			Expect(err).To(Not(HaveOccurred()))
+
+			// Get "http_health_check" and its contents out of the envoy config dump
```
sam-heilbron, line 247: I like these assertions. Can you help me understand how this assertion proves that health checks work as expected? Should we not always be executing some requests to validate that the upstream is considered healthy?

projects/gloo/api/external/envoy/api/v2/core/health_check.proto
sam-heilbron: Makes sense! I don't believe the validate.rules are actually interpreted by our proto complier however, so I don't think they do anything. It may be worth poking around the codebase to see if we have examples of this in use.

If we want to enforce that CONNECT is disallowed, I think we'll need to do that directly in the control plane translation code.

changelog/v1.14.0-beta8/envoy-http-healthcheck-method.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/7149
+    description: Expose `config.core.v3.HealthCheck.HttpHealthCheck.method` as `v3.HealthCheck.HttpHealthCheck.RequestMethod`
+  - type: DEPENDENCY_BUMP
+    dependencyOwner: solo-io
+    dependencyTag: v1.24.0-patch2
+    dependencyRepo: envoy-gloo
+  - type: DEPENDENCY_BUMP
+    dependencyOwner: envoyproxy
+    dependencyTag: v0.10.3-0.20220803141441-64b2023d5b70
```
sam-heilbron, line 11: Looks like a new version was released last night: https://github.com/envoyproxy/go-control-plane/releases/tag/v0.11.0

Can we check to see if it works and if so, upgrade to this one? If not, happy with this change as we have done this in the past

pkg/utils/api_conversion/health_check.go
```diff
@@ -139,6 +140,7 @@ func ToEnvoyHealthCheck(check *envoycore_gloo.HealthCheck, secrets *v1.SecretLis
 			RequestHeadersToAdd:    requestHeadersToAdd,
 			RequestHeadersToRemove: typed.HttpHealthCheck.GetRequestHeadersToRemove(),
 			ExpectedStatuses:       ToEnvoyInt64RangeList(typed.HttpHealthCheck.GetExpectedStatuses()),
+			Method:                 envoy_config_core_v3.RequestMethod(typed.HttpHealthCheck.GetMethod()),
```
sam-heilbron, line 143: In the API we callout that the CONNECT method is disallowed. Can we test that the control plane will reject this? My guess is that our proto validation will not handle this and we need to explicitly perform validation in the control plane. We can test this by installing gloo, set the connect method and seeing how the proxy behaves. If we need to add validation, we should also add a unit test to confirm

sam-heilbron (commented):
Looking good!
Small comments around testing strategy, and potential edge case we should verify.

projects/gloo/api/external/envoy/api/v2/core/health_check.proto
Bslabe123: Added a new translation test in, should be in the next commit

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,80 @@ var _ = Describe("Health Checks", func() {
 		})
 	})
 
+	// This test can be run locally by setting INVALID_TEST_REQS=run, to bypass this ValidateRequirements method in the BeforeEach
+	FDescribe("translates and persists health checkers", func() {
```
jbohanon, line 219: Remove test focus

test/e2e/health_checks_test.go
```diff
@@ -39,9 +44,9 @@ var _ = Describe("Health Checks", func() {
 	)
 
 	BeforeEach(func() {
-		helpers.ValidateRequirementsAndNotifyGinkgo(
-			helpers.LinuxOnly("Relies on FDS"),
-		)
+		//helpers.ValidateRequirementsAndNotifyGinkgo(
+		//	helpers.LinuxOnly("Relies on FDS"),
+		//)
```
jbohanon, line 49: We should either remove this or uncomment it. Wasn't this moved into the two contexts that actually need it in a previous iteration?

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,80 @@ var _ = Describe("Health Checks", func() {
...
+			Eventually(func(g Gomega) {
+				envoyConfig, err := envoyInstance.ConfigDump()
+				g.Expect(err).To(Not(HaveOccurred()))
+
+				// Get "http_health_check" and its contents out of the envoy config dump
+				http_health_check := regexp.MustCompile(`(?sU)("http_health_check": {).*(})`).FindString(envoyConfig)
+				g.Expect(http_health_check).To(ContainSubstring(expectedConfig))
+			}, "10s", "1s").ShouldNot(HaveOccurred())
+		}
+		FIt("with different methods", func() {
```
jbohanon, line 264: Remove test focus

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,80 @@ var _ = Describe("Health Checks", func() {
...
+				return testClients.UpstreamClient.Read(tu.Upstream.Metadata.Namespace, tu.Upstream.Metadata.Name, clients.ReadOpts{})
+			})
+
+			vs := getTrivialVirtualServiceForUpstream(writeNamespace, tu.Upstream.Metadata.Ref())
+			_, err = testClients.VirtualServiceClient.Write(vs, clients.WriteOpts{})
+			Expect(err).NotTo(HaveOccurred())
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				return testClients.VirtualServiceClient.Read(vs.Metadata.Namespace, vs.Metadata.Name, clients.ReadOpts{})
+			})
+			By("default", func() { patchUpstreamAndCheckConfig(v3.RequestMethod_METHOD_UNSPECIFIED, `"path": "health`) })
```
jbohanon, line 279: nit: a newline here would help readability by separating the test cases from the setup.

jbohanon (commented):
Looks great!

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,81 @@ var _ = Describe("Health Checks", func() {
 		})
 	})
 
+	// This test can be run locally by setting INVALID_TEST_REQS=run, to bypass this ValidateRequirements method in the BeforeEach
+	Describe("translates and persists health checkers", func() {
```
jbohanon, line 219: Conventionally we use one `Describe` per test file and use `Context` blocks within that. Not sure functionally if it matters, but it's how I've always seen it

go.mod
```diff
@@ -68,9 +68,9 @@ require (
 	go.uber.org/zap v1.19.1
 	golang.org/x/sync v0.1.0
 	golang.org/x/tools v0.3.0
-	google.golang.org/genproto v0.0.0-20220329172620-7be39ac1afc7
-	google.golang.org/grpc v1.45.0
-	google.golang.org/protobuf v1.28.0
+	google.golang.org/genproto v0.0.0-20221118155620-16455021b5e6
+	google.golang.org/grpc v1.52.0
+	google.golang.org/protobuf v1.28.1
```
sam-heilbron, line 73: Is this necessary? This comes with a lot of noise in the PR, so Im good with it if it's necessary, but if not I think this PR could be cleaner without it

go.mod
```diff
@@ -68,9 +68,9 @@ require (
 	go.uber.org/zap v1.19.1
 	golang.org/x/sync v0.1.0
 	golang.org/x/tools v0.3.0
-	google.golang.org/genproto v0.0.0-20220329172620-7be39ac1afc7
-	google.golang.org/grpc v1.45.0
-	google.golang.org/protobuf v1.28.0
+	google.golang.org/genproto v0.0.0-20221118155620-16455021b5e6
+	google.golang.org/grpc v1.52.0
```
sam-heilbron, line 72: Why is this upgraded? In our replace section, we still pin this to 1.40.0.

pkg/utils/api_conversion/health_check.go
```diff
@@ -133,12 +135,16 @@ func ToEnvoyHealthCheck(check *envoycore_gloo.HealthCheck, secrets *v1.SecretLis
 		if err != nil {
 			return nil, err
 		}
+		if typed.HttpHealthCheck.GetMethod().Number() == envoy_config_core_v3.RequestMethod_CONNECT.Number() {
```
sam-heilbron, line 138: 👏 Nice!

This seems like something that can be easily unit tested, can we add one?

projects/gateway/pkg/translator/http_translator_test.go
```diff
@@ -1124,8 +1126,11 @@ var _ = Describe("Http Translator", func() {
 							},
 						},
 					},
-				}))
-				Expect(httpListener.VirtualHosts[1].Routes).To(Equal([]*gloov1.Route{
+				}
```
sam-heilbron, line 1129: I followed up in Slack, but I'd like to understand the functional difference between these assertions. And if it is different, we should call this out loudly in the PR context, as it's a notable change that others should be aware of.

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -1082,6 +1083,14 @@ var _ = Describe("Translator", func() {
 				msgList = append(msgList, v)
 			}
 			Expect(cluster.HealthChecks).To(ConsistOfProtos(msgList...))
+
```
sam-heilbron, line 1086: I like this! It may make sense to make this a standalone test, as it is verifying a particular case: that a user cannot set the CONNECT method on an http health check. 
nit: We may want to move this unit test closer to the code responsible for this validation

test/benchmark/snapshot_benchmark_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	matchers2 "github.com/solo-io/gloo/test/gomega/matchers"
+	matchers_sk "github.com/solo-io/gloo/test/gomega/matchers"
```
sam-heilbron, line 8: Why this rename? I'm in favor of something more expressive than matchers2, but matchers_sk implies solo-kit and these matchers live in the gloo repo. Perhaps gloomatchers?

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,81 @@ var _ = Describe("Health Checks", func() {
...
+				envoyConfig, err := envoyInstance.ConfigDump()
+				g.Expect(err).To(Not(HaveOccurred()))
+
+				// Get "http_health_check" and its contents out of the envoy config dump
+				http_health_check := regexp.MustCompile(`(?sU)("http_health_check": {).*(})`).FindString(envoyConfig)
+				g.Expect(http_health_check).To(ContainSubstring(expectedConfig))
+			}, "10s", "1s").ShouldNot(HaveOccurred())
+		}
+		It("with different methods", func() {
+			tu = getUpstreamWithMethod(v3.RequestMethod_METHOD_UNSPECIFIED)
```
sam-heilbron, line 265: I like this! One tip:  to avoid too many "assert resource accepted" calls, I tend to write a bunch of resources, and then wait for a Proxy to be accepted. That is the core resource that indicates:
1. We have configured the relevant VS and Gateway
2. Gateway translation has completed successfully
3. Gloo translation has processed the proxy and marked it as valid

test/e2e/health_checks_test.go
```diff
@@ -209,9 +215,81 @@ var _ = Describe("Health Checks", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				return testClients.VirtualServiceClient.Read(vs.Metadata.Namespace, vs.Metadata.Name, clients.ReadOpts{})
+			})
+
+			By("default", func() { patchUpstreamAndCheckConfig(v3.RequestMethod_METHOD_UNSPECIFIED, `"path": "health`) })
+			By("POST", func() { patchUpstreamAndCheckConfig(v3.RequestMethod_POST, `"method": "POST"`) })
+			By("GET", func() { patchUpstreamAndCheckConfig(v3.RequestMethod_GET, `"method": "GET"`) })
+
+			//We expect a health checker with the CONNECT method to be rejected and the prior health check to be retained
```
sam-heilbron, line 284: Great explanatory comment!

go.mod
Bslabe123: This bump is a consequence of bumping go-control-plane

go.mod
Bslabe123: Same as above, something about bumping go-control-plane, it had lots of dependency changes

pkg/utils/api_conversion/health_check.go
Bslabe123: Already added! https://github.com/solo-io/gloo/pull/7695/files#diff-3c7f4aae01d1c96ed3ed400a892439c433f031224c2ede80fc473adbbc860f10R1087-R1093

projects/gateway/pkg/translator/http_translator_test.go
Bslabe123: On it! The difference is that instead of directly comparing the lists of structs we should be comparing only the relevant proto fields, we shouldn't be comparing anything regarding `state`

projects/gloo/pkg/translator/translator_test.go
Bslabe123: Would it make sense to add a `By()` statement to make the test a bit more readable? And should we since we already have all tests relating to health checks within this context?

test/e2e/health_checks_test.go
Bslabe123: Good call!! Added it in

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=c2fb65ca-802a-4ed9-b349-5e8e508b6ae6

sam-heilbron (approved):

jbohanon (approved):

PR #7698, opened 2023-01-19
Temp/no aws

nfuden:
Temporarily disable aws tests due to aws account migration.
In the interim we intend for solo eng to manually run the tests with new configuration until all ci has the new credential setup

### DIFF
--- a/test/aws_credentials/aws_credentials_suite_test.go
+++ b/test/aws_credentials/aws_credentials_suite_test.go
@@ -1,6 +1,7 @@
 package aws_credentials
 
 import (
+	"os"
 	"testing"
 
 	testhelpers "github.com/solo-io/gloo/test/helpers"
@@ -16,7 +17,9 @@ const (
 )
 
 func TestAwsCredentials(t *testing.T) {
-	return
+	if os.Getenv("SKIP_TEMP_DISABLED") == "1" {
+		Skip("Skipping Temporarily due to aws issues")
+	}
 	helpers.RegisterCommonFailHandlers()
 	junitReporter := reporters.NewJUnitReporter("junit.xml")
 	RunSpecsWithDefaultAndCustomReporters(t, "AWS Credentials Suite", []Reporter{junitReporter})
--- a/test/aws_credentials/aws_credentials_suite_test.go
+++ b/test/aws_credentials/aws_credentials_suite_test.go
@@ -16,6 +16,7 @@ const (
 )
 
 func TestAwsCredentials(t *testing.T) {
+	return
 	helpers.RegisterCommonFailHandlers()
 	junitReporter := reporters.NewJUnitReporter("junit.xml")
 	RunSpecsWithDefaultAndCustomReporters(t, "AWS Credentials Suite", []Reporter{junitReporter})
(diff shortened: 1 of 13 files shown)


test/aws_credentials/aws_credentials_suite_test.go
```diff
@@ -16,6 +16,7 @@ const (
 )
 
 func TestAwsCredentials(t *testing.T) {
+	return
```
jenshu, line 19: i believe you can also just `Skip()` tests

ben-taussig-solo (dismissed):
works for me!

ben-taussig-solo:
/kick CI build doesn't seem to be running

sam-heilbron (commented):
Not sure if it would work (would be quick to verify), an alternative is that we update the requirements definintion, to just skip tests:

https://github.com/solo-io/gloo/blob/master/test/helpers/requirements.go#L199, we could put the skip directly in this method, since any test  that requires AWS creds should be calling out to this function anyway

test/aws_credentials/aws_credentials_suite_test.go
ianmacclancy: Resolve outdated

jbohanon (approved):

sam-heilbron:
> > Not sure if it would work (would be quick to verify), an alternative is that we update the requirements definintion, to just skip tests:
> > https://github.com/solo-io/gloo/blob/master/test/helpers/requirements.go#L199, we could put the skip directly in this method, since any test that requires AWS creds should be calling out to this function anyway
> 
> I'm not sure that this would work -- we could call `Skip()` in the requirements, which would prevent AWS tests from running, but `Skip()` directly causes a panic, which causes our CI builds to fail. I think the easiest way to temporarily skip AWS tests is just to return immediately from them

Since this is passing CI and has been reviewed i'm good to move forward with the current implementation. I wouldn't expect Skip to cause a panic. If anything, tests that validate requirements, call ginkgo.Skip or ginkgo.Fail depending on the configuration present: https://github.com/solo-io/gloo/blob/master/test/helpers/requirements.go#L27

ben-taussig-solo:
> > > Not sure if it would work (would be quick to verify), an alternative is that we update the requirements definintion, to just skip tests:
> > > https://github.com/solo-io/gloo/blob/master/test/helpers/requirements.go#L199, we could put the skip directly in this method, since any test that requires AWS creds should be calling out to this function anyway
> > 
> > 
> > I'm not sure that this would work -- we could call `Skip()` in the requirements, which would prevent AWS tests from running, but `Skip()` directly causes a panic, which causes our CI builds to fail. I think the easiest way to temporarily skip AWS tests is just to return immediately from them
> 
> Since this is passing CI and has been reviewed i'm good to move forward with the current implementation. I wouldn't expect Skip to cause a panic. If anything, tests that validate requirements, call ginkgo.Skip or ginkgo.Fail depending on the configuration present: https://github.com/solo-io/gloo/blob/master/test/helpers/requirements.go#L27

Skip causes a panic -- we had to revert the change to use `Skip()` instead of just returning directly in this PR because it was causing CI to fail

The source of `Skip` is as follows:

```go
//Skip notifies Ginkgo that the current spec was skipped.
func Skip(message string, callerSkip ...int) {
	skip := 0
	if len(callerSkip) > 0 {
		skip = callerSkip[0]
	}

	global.Failer.Skip(message, codelocation.New(skip+1))
	panic(GINKGO_PANIC)
}
```

PR #7717, opened 2023-01-23
added the envoyLogLevel field, along with it's tests

irateswami:
# Description

The `gloo.gatewayProxies.gatewayProxy.logLevel` field in helm did not do anything. To set the gateway proxy log level, the user must use the `envoyLogLevel` field. This commit adds that field and it's corresponding tests, which are passing. 

# Checklist:


changelog/v1.14.0-beta7/change-loglevel-to-envoyloglevel.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: HELM
+    description: >
+       Added `gloo.gatewayProxies.gatewayProxy.envoyLogLevel` to the helm values and deprecated `gloo.gatewayProxies.gatewayProxy.logLevel` as the latter has no use in setting the gateway-proxy log level.
```
sam-heilbron, line 4: I think we should refer to this just as `gatewayProxies.gatewayProxy.envoyLogLevel`. The `gloo` prefix is used when referencing a value in a Helm subchart, which is what an enterprise user would do, but this is the open source repo so I think we're good to just have the values match what open source users would configure

docs/content/reference/values.txt
```diff
@@ -794,7 +795,8 @@
 |gatewayProxies.gatewayProxy.istioMetaMeshId|string||ISTIO_META_MESH_ID Environment Variable. Defaults to "cluster.local"|
 |gatewayProxies.gatewayProxy.istioMetaClusterId|string||ISTIO_META_CLUSTER_ID Environment Variable. Defaults to "Kubernetes"|
 |gatewayProxies.gatewayProxy.istioDiscoveryAddress|string||discoveryAddress field of the PROXY_CONFIG environment variable. Defaults to "istiod.istio-system.svc:15012"|
-|gatewayProxies.gatewayProxy.logLevel|string||Level at which the pod should log. Options include "info", "debug", "warn", "error", "panic" and "fatal". Default level is info|
+|gatewayProxies.gatewayProxy.logLevel|string||DEPRECATED, use envoyLogLevel instead. Level at which the pod should log. Options include "info", "debug", "warn", "error", "panic" and "fatal". Default level is info|
```
sam-heilbron, line 798: I would recommend trying out what happens if we delete this field, and seeing how it impacts users. If we claim that this field, if configured, had no effect on a Gloo installation, let's verify the behavior if we just delete the field altogether. For the API of our Gloo Edge resources (Gateway, VirtualService...etc) there are some stricter requirements, but I think for our Helm API we may be safe to delete the field

install/helm/gloo/generate/values.go
```diff
@@ -423,7 +423,8 @@ type GatewayProxy struct {
 	IstioMetaMeshId                *string                      `json:"istioMetaMeshId,omitempty" desc:"ISTIO_META_MESH_ID Environment Variable. Defaults to \"cluster.local\""`
 	IstioMetaClusterId             *string                      `json:"istioMetaClusterId,omitempty" desc:"ISTIO_META_CLUSTER_ID Environment Variable. Defaults to \"Kubernetes\""`
 	IstioDiscoveryAddress          *string                      `json:"istioDiscoveryAddress,omitempty" desc:"discoveryAddress field of the PROXY_CONFIG environment variable. Defaults to \"istiod.istio-system.svc:15012\""`
-	LogLevel                       *string                      `json:"logLevel,omitempty" desc:"Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
+	LogLevel                       *string                      `json:"logLevel,omitempty" desc:"DEPRECATED, use envoyLogLevel instead. Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
```
sam-heilbron, line 426: Where in the project is this value consumed or do we not have any helm templates which tried to parse this value?

sam-heilbron (commented):
Looking good! I think the unit tests are a great way to validate that a user defined value gets injected into the helm template properly. However, I think the area that we're still missing for a test (even a manual one), is demostrating that the user defined value actually changes the log level on the running instance. Can we either include some manual test in the PR outline to demonstrate what steps a user could take to validate this, or an automated test that perhaps looks for a certain log line that will only be available with debug logging enabled?

docs/content/reference/values.txt
nfuden: If it did nothing and does not break anything if it is specified even after removal. Then there is no reason to leave the tombstone imo

changelog/v1.14.0-beta7/change-loglevel-to-envoyloglevel.yaml
nfuden: Should this have an issuelink?

install/test/helm_test.go
```diff
@@ -4732,7 +4790,10 @@ metadata:
 					})
 
 					It("has a creates a deployment", func() {
-						prepareMakefile(namespace, helmValues{})
+						prepareMakefile(namespace, helmValues{
+							valuesArgs: []string{
+								"gatewayProxies.gatewayProxy.envoyLogLevel=debug",
```
nfuden, line 4795: What are we testing with this extra value?

install/helm/gloo/generate/values.go
```diff
@@ -423,7 +423,8 @@ type GatewayProxy struct {
...
+	EnvoyLogLevel                  *string                      `json:"envoyLogLevel,omitempty" desc:"Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
```
nfuden, line 427: These dont look like the actual supported log levels in envoy https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/run-envoy

install/helm/gloo/generate/values.go
sam-heilbron: Given that it's consumed in the template to set an env variable, can we confirm if setting the env variable has any effect (we can compare the bheavior with and without it set), and if it does nothing, I think we should just delete it

install/test/helm_test.go
nfuden: Ok so what we are saying is then is the test changes that were added are just no ops that help with dependencies or something?
I just dont quite understand but if we all agree the tests arent meant to be doing things then fine

install/helm/gloo/generate/values.go
nfuden: I am confused this conversation is resolved and it says we are removing it. However I see the field still. Is this intentional?

install/test/helm_test.go
```diff
@@ -4784,17 +4823,6 @@ metadata:
 
 					})
 
-					It("can set log level env var", func() {
```
sam-heilbron, line 4787: Why are we deleting this test? This test seems to test that we set the log level of the discovery deployment (via the LOG_LEVEL env variable) appropriately, which we still do. Your changes affect a different component/deployment, so I would not expect this test to be affect. Am I missing something?

install/test/helm_test.go
```diff
@@ -4554,17 +4601,6 @@ metadata:
 						})
 
 					})
-
-					It("can set log level env var", func() {
```
sam-heilbron, line 4558: Same as below, I don't think this deployment should be affected

install/test/helm_test.go
```diff
@@ -3515,11 +3551,11 @@ spec:
 					It("can set log level env var", func() {
 						gatewayProxyDeployment.Spec.Template.Spec.Containers[0].Env = append(
 							gatewayProxyDeployment.Spec.Template.Spec.Containers[0].Env,
-							GetLogLevelEnvVar(),
```
sam-heilbron, line 3518: Given that we're deleting this, should we be changing the name of this test? This test seems to test that we can set the log level to a new value, which now is independent of env var, so seems like "can set log level env var" is no longer accurate

install/test/helm_test.go
```diff
@@ -2661,7 +2653,7 @@ spec:
 						}
 						container := GetQuayContainerSpec("gloo-envoy-wrapper", version, GetPodNamespaceEnvVar(), podname)
 						container.Name = "gateway-proxy"
-						container.Args = []string{"--disable-hot-restart"}
+						container.Args = []string{"--disable-hot-restart", "--log-level debug"}
```
sam-heilbron, line 2656: This seems to set the log level to debug always. Should we keep it as info (of just non-existent as the default behavior is) and then we can assert that when we supply a helm value, the container arg is set?

sam-heilbron (commented):
Couple small aspects about the tests to validate this change, but overall changes look good!

changelog/v1.14.0-beta7/change-loglevel-to-envoyloglevel.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: HELM
+    description: >
+       Added `gatewayProxies.gatewayProxy.envoyLogLevel` to the helm values and deprecated `gloo.gatewayProxies.gatewayProxy.logLevel` as the latter has no use in setting the gateway-proxy log level.
```
jenshu, line 4:
```suggestion
       Added `gatewayProxies.gatewayProxy.envoyLogLevel` to the helm values and removed `gloo.gatewayProxies.gatewayProxy.logLevel` as the latter has no use in setting the gateway-proxy log level.
```
```suggestion
       Added `gatewayProxies.gatewayProxy.envoyLogLevel` to the helm values and deprecated `gloo.gatewayProxies.gatewayProxy.logLevel` as the latter has no use in setting the gateway-proxy log level.
```

install/test/helm_test.go
```diff
@@ -3513,12 +3511,12 @@ spec:
 					})
 
 					It("can set log level env var", func() {
```
jenshu, line 3513:
```suggestion
					It("can set envoy log level", func() {
```

jenshu (approved):

jenshu (approved):

ianmacclancy (approved):
Looks good!

nfuden:
What version do we intend to backport this to? It feels like a bug and therefore a backport to 1.11 prior to having resolve issue set

PR #7707, opened 2023-01-20
Ci Aws Cred update

yuval-k:
Update aws creds to unblock novel aws location internally
Update gke via cloud builders for novel gke kubectl auth plugin features

PR #7726, opened 2023-01-24
change aws test to check role assumption

ianmacclancy:
Pr to validate ci with new role - no changes beyond this. Once validated this will be taken down and rewritten


ianmacclancy:
Changes confirmed - removing and merging changes into https://github.com/solo-io/gloo/pull/7707

Issue #7659, opened 2023-01-11
Disable core dump for envoy process


nikitakogan1:
### Version

1.13.x (latest stable)

### Is your feature request related to a problem? Please describe.

Hi,

I am running Gloo in a k8s cluster.
I would like to disable core dump for the envoy process, because of potential sensitive data exposure.

### Describe the solution you'd like

I would like to have the ability to disable core dump for the envoy process using a flag/parameter in the gateway proxy deployment / DockerFile level. 

### Describe alternatives you've considered

I can run a separate command inside the container before Envoy is started.
The issue is that this requires us to override the "EXEC" in the original image with our own "command" field on the pod, which means I need to specify the exact command in the original image on our pod and adjust it with every gloo upgrade, since it could potentially change the EXEC in the gateway-proxy image.
This process is something I would very much like to avoid.

I was able to make it work with changing the k8s gateway-proxy deployment to:

```
- args:
        - ulimit -c 0 ; exec /usr/bin/dumb-init -- /docker-entrypoint.sh
        command:
        - /bin/sh
        - -c
```

PR #7740, opened 2023-01-26
7659 disable envoy core dumps

ashishb-solo:
# Description

Provide a means to disable core dumps by setting an environment variable `DISABLE_CORE_DUMPS`. This value can also be set at the time of installation by running `helm install --set gatewayProxies.gatewayProxy.disableCoreDumps=true ...`

Steps to reproduce the issue can be found in [this commit](https://github.com/solo-io/gloo/pull/7740/commits/b5d933f6413812e7113fb9a3469fb3de090dbdd7). Reproduced here in case it gets lost in the squash:

1. temporarily change the location core dumps are written to: `echo '/dev/shm/core.%e.%p' | sudo tee /proc/sys/kernel/core_pattern` - this probably only works on linux and hopefully it doesn't break your operating system, but i can't make any guarantees here

2. start a new cluster

3. `kubectl -n gloo-system edit deployments/gateway-proxy`

4. add the DISABLE_CORE_DUMPS environment variable and set it to "false" (quotes required)

5. `kubectl -n gloo-system exec -it deployments/gateway-proxy sh`

6. send a signal to envoy signalling it to abort and leave a core dump `kill -SIGABRT $(pidof envoy)`

7. if you got kicked out of the container, run `docker exec -it deployments/gateway-proxy sh` again

8. `ls /dev/shm/` - should see the core dump file there

9. repeat steps 3-8 with DISABLE_CORE_DUMPS set to "true" and this time, no core files should appear in step 8

note that you can run this to check the status of whether core dumps are enabled: `cat /proc/$(pidof envoy)/environ | tr '\0' '\n' | grep CORE`

also note that in order to enable this, i had to put some debug code into `docker-entrypoint.sh` which i will remove in my next commit

# Context

Users needed this feature to disable core dumps

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7659


install/test/helm_test.go
```diff
@@ -2754,6 +2760,29 @@ spec:
...
+
+								Expect(structuredDeployment.Spec.Template.Spec.Containers).To(HaveLen(1))
+								for _, envvar := range structuredDeployment.Spec.Template.Spec.Containers[0].Env {
+									if envvar.Name != "DISABLE_CORE_DUMPS" {
+										continue
+									}
+									Expect(envvar.Value).To(Equal("true"))
+								}
+							})
+							// testManifest.Expect("Deployment", "Spec.Template.Spec.Containers[0].Env", "Value").To(Equal("true"))
```
ben-taussig-solo, line 2783: Should this be uncommented?

ben-taussig-solo:
changes look good, evaluating manually before proceeding with approval

install/test/helm_test.go
ashishb-solo: nee, it should be deleted. thanks for the catch

ashishb-solo: https://github.com/solo-io/gloo/pull/7740/commits/2cdb1cadb87a980a83e4e4f092e0696dab436c77

install/test/helm_test.go
```diff
@@ -2754,6 +2760,28 @@ spec:
 							testManifest.Expect("DaemonSet", gatewayProxyDeployment.Namespace, gatewayProxyDeployment.Name).To(BeEquivalentTo(daemonSet))
 						})
 
+						It("can set the DISABLE_CORE_DUMPS environment variable", func() {
```
nfuden, line 2763: Do we see passing any other env vars in the future like this?
If so do we think it make sense to parameterize this? 

Also happy to punt on that aspect as its definitely not needed

projects/envoyinit/cmd/docker-entrypoint.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/sh
-set -e
+set -eu
```
nfuden, line 2: Is there a specific reason we are using u? Or just doing this to align to best practices?

ashishb-solo: I tend to prefer strictness in bash scripts, but was also a bit ambivalent on this, since it's a fairly straightforward start-up script and if it fails due to some unknown reason, then the deployment will just never start. I have tested it extensively though. Thoughts? I'm happy to revert this.

ashishb-solo: On the other hand, `-u` only raises an error if there are undefined variables, and there are only 2 variables in the script, both of which have default values, so I don't think it adds any harm to have `-u` in there for now.

nfuden (dismissed):

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
```diff
@@ -200,6 +200,12 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: DISABLE_CORE_DUMPS
+          value: {{ $spec.disableCoreDumps | quote }}
+      {{- if $spec.logLevel }}
```
sam-heilbron, line 205: I think this was unintentional and should be removed?

ashishb-solo: I don't think so. I'm not sure why they're showing up in the diff, but those lines have been there since 2021: https://github.com/solo-io/gloo/blame/4e0a6f170cbc8670557092b151fd8edca3438411/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L205

ashishb-solo: And this seems to be present in our other deployments too, so I think it belongs there:
https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/1-gloo-deployment.yaml#L229
https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/3-discovery-deployment.yaml#L90

jenshu: it was removed recently via https://github.com/solo-io/gloo/pull/7717

probably just a merge issue caused it to get added back

jenshu (changes requested):
need to remove the logLevel

install/helm/gloo/templates/7-gateway-proxy-deployment.yaml
ashishb-solo: Got it, thanks! I've removed it: https://github.com/solo-io/gloo/pull/7740/commits/9a8e4d4eeba5235b50e314cc6ff63cb1d436bf1f

ben-taussig-solo (approved):
Got on a call with Ashish who demo'd this functionality working as expected in the PR description

ashishb-solo:
Thanks! Just need @jenshu to resolve the review and @nfuden to approve again

jenshu (approved):

PR #7741, opened 2023-01-26
Expose respectDnsTtl and dnsRefreshRate fields

irateswami:
# Description

Exposes the RespectDnsTtl and DnsRefreshRate fields in the upstream proto. For DnsRefreshRate, If the DNS refresh rate is specified and the cluster type is either STRICT_DNS, or LOGICAL_DNS, this value is used as the cluster’s DNS refresh rate. The RespectDnsTtl is an optional configuration for setting cluster’s DNS refresh rate. If the value is set to true, cluster’s DNS refresh rate will be set to resource record’s TTL which comes from DNS resolution. 


changelog/v1.14.0-beta7/expose-RespectDnsTtl-and-DnsRefreshRate.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/6402
+    resolvesIssue: false
+    description: Exposes the RespectDnsTtl and DnsRefreshRate fields in the upstream proto. For DnsRefreshRate, If the DNS refresh rate is specified and the cluster type is either STRICT_DNS, or LOGICAL_DNS, this value is used as the cluster’s DNS refresh rate. The RespectDnsTtl is an optional configuration for setting cluster’s DNS refresh rate. If the value is set to true, cluster’s DNS refresh rate will be set to resource record’s TTL which comes from DNS resolution.
```
sam-heilbron, line 5: nice! 👏

projects/gloo/api/v1/upstream.proto
```diff
@@ -161,6 +163,14 @@ message Upstream {
     // (bool) If set to true, Envoy will ignore the health value of a host when processing its removal from service discovery.
     // This means that if active health checking is used, Envoy will not wait for the endpoint to go unhealthy before removing it.
     google.protobuf.BoolValue ignore_health_on_host_removal = 22;
+
+    // (bool) If set to true, Service Discovery update period will be triggered once the TTL is expired.
+	// If minimum TTL of all records is 0 then dns_refresh_rate will be used.
+	google.protobuf.BoolValue respect_dns_ttl= 29;
+
+    // (Duration) Service Discovery DNS Refresh Rate, if unset it defaults to 5 seconds.
+	// Minimum value is 1 ms
```
sam-heilbron, line 172: I like this callout. Is this lower bound something that is enforced by Envoy? Should we have logic in the control plane to validate that the provided duration is above the lower bound?

projects/gloo/pkg/plugins/kubernetes/uds_convert_test.go
```diff
@@ -500,6 +501,24 @@ var _ = Describe("UdsConvert", func() {
 						Value: true,
 					},
 				}),
+				Entry("Using RespectDnsTtl", map[string]string{
```
sam-heilbron, line 504: Love a good table driven test!

sam-heilbron (commented):
Looking great!
1. Do you have any questions about how the DNS refresh rate affects behavior in the data plane?
2. The unit tests are great. Do you think there is a way (or is it worth it) to validate this behavior end to end to ensure that when a field is set on the GLoo API it is transformed and sent to the gateway-proxy via xDS? If an automated test is overkill, perhaps the manual steps to verify by inspecting the config dump

projects/gloo/api/v1/upstream.proto
```diff
@@ -161,6 +163,14 @@ message Upstream {
     // (bool) If set to true, Envoy will ignore the health value of a host when processing its removal from service discovery.
     // This means that if active health checking is used, Envoy will not wait for the endpoint to go unhealthy before removing it.
     google.protobuf.BoolValue ignore_health_on_host_removal = 22;
+
+    // (bool) If set to true, Service Discovery update period will be triggered once the TTL is expired.
+	// If minimum TTL of all records is 0 then dns_refresh_rate will be used.
+	google.protobuf.BoolValue respect_dns_ttl= 29;
```
bewebi, line 169:
```suggestion
    // If minimum TTL of all records is 0 then dns_refresh_rate will be used.
    google.protobuf.BoolValue respect_dns_ttl= 29;
```
Standardize spaces vs tabs

bewebi (commented):
LGTM, one of the nittiest nits

projects/gloo/api/v1/upstream.proto
sam-heilbron: If envoy is enforcing the bounds, what is the behavior if a user sets a value below that lower bound? I think when there are requirements that Envoy has, we as the control-plane should perform that validation in order to reject resources which Envoy will reject

nfuden:
can we change the title of this pr? The title becomes the commit message therefore should indicate what was done

projects/gloo/api/v1/upstream.proto
sam-heilbron: I think we want to avoid situations where Envoy is silently ignoring invalid configuration. One of the goals of the control plane is to provide feedback to users, so I think we should validate this and add an error to the report if the provided duration is outside the range that Envoy can support. @ianmacclancy what do you think?

nfuden: Did we check for nacks? Envoy doesn't apply the config but it will update with new config or does it nack it?

ianmacclancy: I agree that we should add validation on dnsRefreshRate less than 1ms.

While not mentioned yet - I don't think we need to validate if the cluster type is STRICT_DNS or LOGICAL_DNS. If the cluster is not one of those types dnsRefreshRate is ignored but if we could ignore the error in cases where the cluster type is not strict/logical DNS that would be a nice add to the error checking.

projects/gloo/pkg/translator/clusters.go
```diff
@@ -102,6 +102,8 @@ func (t *translatorInstance) initializeCluster(
 		ConnectTimeout:            ptypes.DurationProto(ClusterConnectionTimeout),
 		Http2ProtocolOptions:      getHttp2options(upstream),
 		IgnoreHealthOnHostRemoval: upstream.GetIgnoreHealthOnHostRemoval().GetValue(),
+		RespectDnsTtl:             upstream.GetRespectDnsTtl().GetValue(),
+		DnsRefreshRate:            upstream.GetDnsRefreshRate(),
```
ianmacclancy, line 106: We should add the check here and report an error - adding it in update_upstream_spec will hide the failure to set as it doesn't report incorrect config, it just silently ignores it.

projects/gloo/pkg/translator/clusters.go
```diff
@@ -89,6 +90,21 @@ func (t *translatorInstance) initializeCluster(
 	}
 
 	circuitBreakers := t.settings.GetGloo().GetCircuitBreakers()
+
+	var dnrP *durationpb.Duration
+
+	if dnrP == nil {
```
nfuden, line 96: Woulldnt this always be true as this is defined 2 lines above?

sam-heilbron:
I am going to close this as we are adding this functionality as part of https://github.com/solo-io/gloo/pull/7834

PR #7746, opened 2023-01-27
back1.12: Tests now support aws again in ci

nfuden:
Backport renablement of auto ci for aws


ben-taussig-solo (approved):
I saw that we're skipping the changelog entry in the 1.13.x PR -- If we're doing that here, this looks good to me

PR #7765, opened 2023-01-30
7659 disable envoy core dumps (#7740)

ashishb-solo:
* Initial implementation with successful manual test

We create an environment variable in the gateway-proxy deployment and pass it to the `docker-entrypoint.sh` script which is the entrypoint for the deployment to start the envoy process. If this value is set to true, we call `ulimit -c 0` which disables core dumps. If it is set to false, then it is not called and core dumps can be created at will.

I have verified that this works as follows:

1. temporarily change the location core dumps are written to: `echo '/dev/shm/core.%e.%p' | sudo tee /proc/sys/kernel/core_pattern` - this probably only works on linux and hopefully it doesn't break your operating system, but i can't make any guarantees here

2. start a new cluster

3. `kubectl -n gloo-system edit deployments/gateway-proxy`

4. add the DISABLE_CORE_DUMPS environment variable and set it to "false" (quotes required)

5. `docker exec -it deployments/gateway-proxy sh`

6. send a signal to envoy signalling it to abort and leave a core dump `kill -SIGABRT $(pidof envoy)`

7. if you got kicked out of the container, run `docker exec -it deployments/gateway-proxy sh` again

8. `ls /dev/shm/` - should see the core dump file there

9. repeat steps 3-8 with DISABLE_CORE_DUMPS set to "true" and this time, no core files should appear in step 8

note that you can run this to check the status of whether core dumps are enabled: `cat /proc/$(pidof envoy)/environ | tr '\0' '\n' | grep CORE`

also note that in order to enable this, i had to put some debug code into `docker-entrypoint.sh` which i will remove in my next commit

* Remove debug code for testing

* Wrap errors to provide a big more context

This was made necessary after a long debugging session yesterday. We were having trouble identifying from where errors were being generated and thrown, so these new contexts will hopefully make it easier for developers to track these errors down.

* Fix broken unit tests

$spec.disableCoreDumps allows marshalling the boolean value in go to/from the string value in teh environment variables array of the container. Additionally, add this fragment of the template to the expected version in the helm test.

* Add a new unit test

* Provide a default value for DISABLE_CORE_DUMPS

This isn't technically necessary, but helpful just in case something goes wrong. In the previous iteration, the script would fail and envoy would not start if DISABLE_CORE_DUMPS was not set. Technically, this should never happen because the helm templates always provide this value. But if the entrypoint is reached some other way (ie. not from the helm templates provided, or if the environment variables in the deployment are removed for whatever reason), then this script would fail and envoy would never start. So it seems safer to just provide a default here.

* Remove commented line

* Add `exec` back into docker-entrypoint.sh

* Add changelog

* Remove LOG_LEVEL

---------

Co-authored-by: soloio-bulldozer[bot] <48420018+soloio-bulldozer[bot]@users.noreply.github.com>

# Description

# Context

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7659


changelog/v1.13.5/disable-core-dumps.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NON_USER_FACING
```
jenshu, line 2: can we change the type to `HELM` so it shows up in the changelogs?

ashishb-solo: https://github.com/solo-io/gloo/pull/7765/commits/baf4b248a52482731c560adad7fbc3061a86a42b

jenshu (approved):

ben-taussig-solo (approved):

PR #7804, opened 2023-02-07
Fix Istio Crash When Using Headless Services

inFocus7:
# Description

- Fixes issue with our istio integration code by only using the Service/Cluster IP if istio integration is enabled **and** it's not a headless service (No cluster ip).
- Moves istioIntegration logic back to original place.
  - I originally moved it outside of our subset logic in #7466; it seemed as a minor performance bump, since we only cared about cluster ip + port, not pod/subsets _at the time_. Now that we require the setup of pods for headless services, I moved it back to its original place.
- Gated the validation webhook start based on if the values is `true` or ` `. The blank one is because, previously, we didn't care for it therefore treating that as ` `. But open to changing it.

# Context

Users where running into an issue when enabling istioIntegration `enableIstioSidecarOnGateway` with headless services in their environment, where their `gateway-proxy` logs were flooded with:
```
[warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment rejected: malformed IP address: None. Consider setting resolver_name or setting cluster type to 'STRICT_DNS' or 'LOGICAL_DNS'
```

# TODO
- [ ] Write tests

# Checklist:


projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -269,6 +269,8 @@ type Epkey struct {
 	Name        string
 	Namespace   string
 	UpstreamRef *core.ResourceRef
+	// While we can use the upstream ref to get the upstream and service, if there are too many upstreams that could be slow.
+	IsHeadless bool
```
kdorosh, line 273: this doesn't seem like it's used at all. the only place where this can be set to true is gated by an if statement that will ensure we don't hit this code

inFocus7: Ah I added it specifically for [these lines here](https://github.com/solo-io/gloo/pull/7804/files#diff-34a26dd6e5aa3b257ff21ab1575da112448f84063d4cbcfc01cc8e85a0b512bdR442-R443), as it's a separate function and I wanted to make sure we don't create the final endpoint wrong

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -339,8 +335,26 @@ func filterEndpoints(
 					continue
 				}
 
-				warnings := processSubsetAddresses(subset, spec, podMap, usRef, port, endpointsMap)
-				warnsToLog = append(warnsToLog, warnings...)
+				svc := getServiceFromUpstreamSpec(spec, services)
+				isHeadlessSvc := svc.Spec.ClusterIP == "None"
```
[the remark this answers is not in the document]
inFocus7: Would you happen to know where/if the deprecation is documented? I found that [LoadBalancerIP](https://github.com/kubernetes/kubernetes/pull/107235/files) was deprecated, although can't find any references to ClusterIP being deprecated

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -322,6 +336,7 @@ func filterEndpoints(
 				Name:        spec.GetServiceName(),
 				Namespace:   spec.GetServiceNamespace(),
 				UpstreamRef: &copyRef,
+				IsHeadless:  isHeadlessSvc,
```
kdorosh, line 339: this is the only place where we can set `IsHeadless` to true, but we never will because it's gated by the `if istioIntegrationEnabled && !isHeadlessSvc` check

projects/gloo/pkg/plugins/kubernetes/eds.go
kdorosh: i still don't understand, https://github.com/solo-io/gloo/pull/7804/files#r1098929238

projects/gloo/pkg/plugins/kubernetes/eds.go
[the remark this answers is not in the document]
kdorosh: i tried finding it but the first thing i found was that openshift doc i sent. i suggest reading the api docs for the latest k8s apis, e.g. https://github.com/kubernetes/api

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: Ah okay gotcha. I could hard-set it to `false` there since that's the value it'll always be.
It's still useful in the case of [here](https://github.com/solo-io/gloo/pull/7804/files#diff-34a26dd6e5aa3b257ff21ab1575da112448f84063d4cbcfc01cc8e85a0b512bdR357) - where it can be true - and [here](https://github.com/solo-io/gloo/pull/7804/files#diff-34a26dd6e5aa3b257ff21ab1575da112448f84063d4cbcfc01cc8e85a0b512bdR444-R453). When we create the endpoints from the epkey->ref map, we'd want to know if it's part of a headless service to know whether or not to  use the ClusterIP, and that's why I stored it in the key

projects/gloo/pkg/plugins/kubernetes/eds.go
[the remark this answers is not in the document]
inFocus7: Thanks! Looked through the api, as of `master` right now, ClusterIP is not marked as deprecated [serviceSpec.clusterIp](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4529-L4545)

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: https://github.com/solo-io/gloo/pull/7804/files#r1098947452

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: I could remove it, but then we'd have to do the `getServiceFromUpstreamSpec()` loop twice. 

Once [here](https://github.com/solo-io/gloo/pull/7804/files#diff-34a26dd6e5aa3b257ff21ab1575da112448f84063d4cbcfc01cc8e85a0b512bdR329-R330) to know if I should store the endpoint key using service information, and another time [here](https://github.com/solo-io/gloo/pull/7804/files#diff-34a26dd6e5aa3b257ff21ab1575da112448f84063d4cbcfc01cc8e85a0b512bdR443-R444) so we don't use the `ClusterIP` for the generated endpoint.

kdorosh: ah its line 395

inFocus7:
marking as ready-for-review to make sure all tests pass + hopefully catch a istio test flake

changelog/v1.14.0-beta10/fix-istio-integration-headless-service-crash.yaml
```diff
@@ -0,0 +1,11 @@
+changelog:
+  - type: FIX
+    description: Use Pod IPs for headless services when the gloo deployment's `ENABLE_ISTIO_INTEGRATION` is `true`.
+    issueLink: https://github.com/solo-io/gloo/issues/7491
+    resolvesIssue: false
+  - type: FIX
+    description: |
+      Fix a bug where the validation webhook would still start even if the `VALIDATION_MUST_START` environment variable was `false`.
```
kdorosh, line 8: nit, just for clarity
```suggestion
      Fix a bug where the validation webhook server on the gloo pod would still start even if the `VALIDATION_MUST_START` environment variable was `false`.
```

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -312,8 +323,11 @@ func filterEndpoints(
 			continue
 		}
 
+		svc := getServiceFromUpstreamSpec(spec, services)
+		isHeadlessSvc := svc.Spec.ClusterIP == "None"
```
kdorosh, line 327: can we link the open GH issue in a comment here for looking into clusterips?

kdorosh (commented):
lgtm overall

projects/gloo/pkg/plugins/kubernetes/eds.go
inFocus7: Just added a comment with the issue linked!

projects/gloo/pkg/plugins/kubernetes/eds.go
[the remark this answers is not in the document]
inFocus7: https://github.com/solo-io/gloo/pull/7804#discussion_r1102891723

kdorosh (approved):

nfuden (approved):

PR #7819, opened 2023-02-08
[v1.10.x] Update gopkg.in/yaml.v3 to resolve CVE

ben-taussig-solo:
# Description
 - Bump gopkg.in/yaml.v3 to 3.0.1 to resolve https://avd.aquasec.com/nvd/2022/cve-2022-28948/


go.mod
```diff
@@ -91,6 +91,7 @@ require (
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/AlecAivazis/survey.v1 v1.8.7
 	gopkg.in/src-d/go-git.v4 v4.13.1 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
jbohanon, line 94: Since this is an indirect dependency, are we certain that our direct dependency which requires/consumes this is using that version as well?

jbohanon: never mind, this is precisely how to fix that issue

jbohanon (approved):

Issue #7517, opened 2022-12-06
Kube2e Suite Test Flakes (Upgrade)


ashishb-solo:
### Gloo Edge Version

1.13.x (beta)

### Kubernetes Version

None

### Describe the bug

Log of the error can be found [here](https://github.com/solo-io/gloo/actions/runs/3632980984/jobs/6129464157)

Relevant portion below:

```
Random Seed: 1670357302
Will run 6 of 6 specs

"Tue, 06 Dec 2022 20:12:15 UTC: /home/runner/go/pkg/mod/github.com/solo-io/k8s-utils@v0.1.1/testutils/helper/install.go:253"	found Helm index file at: "/home/runner/work/gloo/gloo/_test/index.yaml"
"Tue, 06 Dec 2022 20:12:15 UTC: /home/runner/go/pkg/mod/github.com/solo-io/k8s-utils@v0.1.1/testutils/helper/install.go:262"	version of ["gloo"] Helm chart is: "1.0.0-ci"
make: *** [Makefile:226: run-ci-regression-tests] Error 1
/home/runner/work/gloo/gloo/test/kube2e/upgrade/upgrade_utils.goFail handler msg Unexpected error:
    <*eris.rootError | 0xc000254a80>: {
        global: false,
        msg: "error listing releases",
        ext: <*github.ErrorResponse | 0xc0003f4140>{
            Response: {
                Status: "401 Unauthorized",
                StatusCode: 401,
                Proto: "HTTP/2.0",
                ProtoMajor: 2,
                ProtoMinor: 0,
                Header: {
                    "X-Content-Type-Options": ["nosniff"],
                    "Content-Security-Policy": ["default-src 'none'"],
                    "Vary": [
                        "Accept-Encoding, Accept, X-Requested-With",
                    ],
                    "Date": [
                        "Tue, 06 Dec 2022 20:12:15 GMT",
                    ],
                    "X-Github-Media-Type": [
                        "github.v3; format=json",
                    ],
      	/home/runner/go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/leafnodes/runner.go:64 +0x158
      github.com/onsi/ginkgo/internal/leafnodes.(*SetupNode).Run(0xc000010158)
      	/home/runner/go/pkg/mod/github.com/onsi/ginkgo@v1.16.5/internal/leafnodes/setup_nodes.go:15 +0x95
[...]

PR #7821, opened 2023-02-08
Upgrade Tests: Add checks to github tags 

ianmacclancy:
# Context
Fixing flakes on upgrade tests

# Description
Adding 3 checks when fetching versions from github
- Version tag cannot be pre-release
- Version tag cannot have a build failure
- Version tag must be 30 minutes or older (avoids potential lag fetching the built code)

# Testing
Ran all tests locally and saw successful runs

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7517


test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -106,9 +107,13 @@ func getLatestReleasedVersion(ctx context.Context, majorVersion, minorVersion in
 		for _, release := range releases {
 			v, err := versionutils.ParseVersion(*release.Name)
 			if err != nil {
+				return nil, errors.Wrapf(err, "error parsing version from release")
+			}
+			if release.GetPrerelease() || // we don't want a prerelease version
+				strings.Contains(release.GetBody(), "This release build failed") || // we don't want a failed build
+				!release.GetPublishedAt().Before(time.Now().In(time.UTC).Add(time.Duration(-30)*time.Minute)) { // releases take time to propagate - image might not be available for 30 minutes
```
jenshu, line 114: 🙌 do you think 30 min is enough? i've seen helm charts not available for up to an hour due to gcloud caching

sam-heilbron (commented):
We have some utility methods which seem to achieve what you're looking for. Instead of redefining it, could we use https://github.com/solo-io/go-utils/blob/main/githubutils/repo.go#L174 or one of the other "Find" methods?

Side note: I notice that getLatestReleasedVersion makes a call to the GitHub API (necessary to determine latest release). However this method is called by GetUpgradeVersions, which is executed in every BeforeEach. Could we instead move this to the BeforeSuite so that it happens once and then we can re-use the results?

test/kube2e/upgrade/upgrade_utils.go
ianmacclancy: I can move it to an hour but let's talk through why it is taking so long and maybe see if we can speed up the to to availability for helm

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -85,43 +86,41 @@ func getLastReleaseOfCurrentMinor(repoName string) (*versionutils.Version, error
...
-	// We also assume that the first release that is not strict semver is technically the largest
-	for i := 0; i < 5; i++ {
-		// Get the next page of
-		listOpts := github.ListOptions{Page: i, PerPage: 10} // max per request
-		releases, _, err := client.Repositories.ListReleases(ctx, "solo-io", "gloo", &listOpts)
+	releases, err := githubutils.GetRepoReleasesWithPredicateAndMax(ctx, client, "solo-io", "gloo", newLatestPatchForMinorPredicate(versionPrefix), 2)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting releases from github")
+	}
+	for _, release := range releases {
```
sam-heilbron, line 113: I think the idea of the predicate is to filter out releases that are returned. Any reason we have a predicate that performs that filtering and then this additional filtering? Could we instead just say that the predicate only allows non-prerelease, releases before a certain time window?

test/kube2e/upgrade/upgrade_test.go
```diff
@@ -85,6 +82,14 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
 		}
 	})
 
+	// setup for all tests
+	BeforeEach(func() {
+		var err error
```
sam-heilbron, line 87: Is this necessary? We setup the context and testHelper in the BeforeSuite and then override those in the BeforeEach. Could we have just one of these do that?

ianmacclancy: Depends on how you want to handle context in the tests. Both context and testhelper are needed for the before suit and by overriding context and testhandler in the beforeEach each test will have its own context and the testhelper used for that test will have the tests context. 

If we want all the tests to share the same context it can be removed from the before each

changelog/v1.14.0-beta13/upgrade-test-flakes.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/7517
```
sam-heilbron, line 3: I see that we're closing out this issue with this PR. In the linked issue, there were an issue with these tests running on a community PR (https://github.com/solo-io/gloo/issues/7517#issuecomment-1358315840). Can you verify that with these changes that doesn't happen before we close out the issue?

test/kube2e/upgrade/README.md
```diff
@@ -0,0 +1,25 @@
+# Test Overview
```
sam-heilbron, line 1: Love this outline!

test/kube2e/upgrade/README.md
```diff
@@ -0,0 +1,25 @@
+# Test Overview 
+- All upgrade tests run twice. 
+    - current minor last patch -> current code (ex: 1.12.3 -> 1.12.4)
+    - previous minor latest patch -> current code (ex 1.11.latest -> 1.12.4)
+- Each upgrade test installs gloo, upgrades CRDs and then upgrades gloo
+- Each test runs a data setup and validation on installation of gloo and on upgrade of gloo
+
+# Running Tests
+- You will need to generate the `_test` and `_output` folders in order to run these tests locally
+  - running `./ci/setup-kind` is an easy way to set up your environment to run the tests
```
sam-heilbron, line 10: This isn't entirely accurate. It may help to just reference the top level "kube2e" README. The script isn't called setup-kind, it's https://github.com/solo-io/gloo/blob/master/ci/deploy-to-kind-cluster.sh. nit: If I reference a file within a readme, I try to link it so that it's easy to navigate to.

test/kube2e/upgrade/README.md
```diff
@@ -0,0 +1,25 @@
...
+- If you are debugging and cancel the test early there may be leftover pods and namespace resources, these commands can help clean things up for you
+
+# remove gloo-system namespace resources and crds
```
sam-heilbron, line 13: Out of curiosity, when would you recommend someone do this? It's definitely a helpful tip, but it's unclear to me as someone following this README when/why I'd want to do this

test/kube2e/upgrade/upgrade_test.go
sam-heilbron: Yeah that's totally fair. It seems that we use the testHelper in the BeforeSuite purely to prepare the chartUri that we'll pull from. I'm happy to defer to you on the maintenance of the context, and whether there is 1 for the entire suite, or 1 per test. If we continue with the pattern you have here, can we at least have a suiteCtx and ctx so that we ensure we cancel each and there is a clear distinction?

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -85,43 +86,36 @@ func getLastReleaseOfCurrentMinor(repoName string) (*versionutils.Version, error
 	return versions[len(versions)-2], nil
 }
 
+type latestPatchForMinorPredicate struct {
```
sam-heilbron, line 89: 👏

test/kube2e/upgrade/README.md
ianmacclancy: If you install gloo and then gloo edge and have cleaned up the namespace and the crds these roles will be left behind and if you go to install again they can interfere with installation as we will try to use / create them and not have the correct ownership

test/kube2e/upgrade/upgrade_test.go
```diff
@@ -85,6 +82,14 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
...
+		ctx, cancel = context.WithCancel(context.Background())
```
nfuden, line 89: Should this be a wrapper around the context setup in the beforesuite?

test/kube2e/upgrade/README.md
ianmacclancy: Copy and paste mistake - I was pulling this in from solo-projects and missed this. We don't align on the names so this just needs to be updated for oss

test/kube2e/upgrade/README.md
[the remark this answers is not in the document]
ianmacclancy: Do we specify all others tests can run offline? If this test suite is unique in needing the internet then yes - I think it makes sense to include it.

test/kube2e/upgrade/upgrade_test.go
ianmacclancy: beforeSuiteCtxCancel

ianmacclancy: There are now two distinct contexts in the file - before suite has its own context that does not need to be used elsewhere and is canceled once we have the version numbers. Before each will create a context for each test as before.

test/kube2e/upgrade/upgrade_test.go
ianmacclancy: Nope, different use and I have changed the usage. I think the current implementation addresses this comment. Each test has its own context that is canceled on gloo uninstall.

changelog/v1.14.0-beta13/upgrade-test-flakes.yaml
ianmacclancy: This is a large enough testing effort the it can be split off. https://github.com/solo-io/gloo/issues/7867
No need to hold up this PR

sam-heilbron: Sounds good! I'm surprised it will be a large effort but happy to have it done in a follow-up PR if that's how you'd like to approach it

test/kube2e/upgrade/upgrade_test.go
```diff
@@ -64,25 +63,33 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
 		firstReleaseOfMinor                bool
 	)
 
-	// setup for all tests
-	BeforeEach(func() {
-		ctx, cancel = context.WithCancel(context.Background())
+	BeforeSuite(func() {
 		var err error
-		testHelper, err = kube2e.GetTestHelper(ctx, namespace)
+		beforeSuiteCtx, beforeSuiteCtxCancel := context.WithCancel(context.Background())
```
sam-heilbron, line 68: nit: Some high level thoughts on the usage of the ctx in these tests, but nothing blocking:

I think the separation of a suiteCtx and a testCtx makes sense. I would check out https://pkg.go.dev/context or more details. I think the main benefit/purpose the context provides us in tests is a way to pass cancellation signals. Therefore, if we want to move forward with a suiteCtx (as you have it), I would expect that we: (A) call cancel in the AfterSuite (B) create test contexts from the suiteContext so that if the suiteCOntext is cancelled it ripples out to all other contexts (C) cancel the test context in a AfterEach if we create it in a beforeEach

test/kube2e/upgrade/README.md
```diff
@@ -0,0 +1,26 @@
...
+- Internet needed for tests to run
+- All upgrade tests run twice. 
+    - current minor last patch -> current code (ex: 1.12.3 -> 1.12.4)
+    - previous minor latest patch -> current code (ex 1.11.latest -> 1.12.4)
+- Each upgrade test installs gloo, upgrades CRDs and then upgrades gloo
+- Each test runs a data setup and validation on installation of gloo and on upgrade of gloo
+
+# Running Tests
+- You will need to generate the `_test` and `_output` folders in order to run these tests locally
+  - running [./ci/deploy-to-kind-cluster.sh](/ci/deploy-to-kind-cluster.sh) is an easy way to set up your environment to run the tests
```
sam-heilbron, line 11: Will this link to the proper folder? I imagine we'd need a github.com/soloio/gloo prefixing the url or if its relative then I think we're off by a few folders

sam-heilbron (approved):
Non-blocking comments that I think can get handled in the follow up PR to make these tests pass on community PRs

test/kube2e/upgrade/README.md
ianmacclancy: You can see that it works if you click on view rich diff at the top of the file diff

test/kube2e/upgrade/upgrade_test.go
nfuden: I agree with the general comment on this I would argue that while for now the main benefit of the contet is to pass cancellation it is also important in terms of tracing and how we might in the future better be able to observe the tests progression if we have correctly named spans throughout

nfuden (approved):

test/kube2e/upgrade/README.md
sam-heilbron: Oh nice! I didn't realize github handles that routing for us, neat

PR #7824, opened 2023-02-09
Prep Makefile for Ginkgo v2

sam-heilbron:
# Description

Prep Makefile and Cloudbuild for ginkgo v2 changes

# Context

We have upcoming work to upgrade Ginkgo from v1 to v2. As part of these changes, there are improvements to the Makefile and Cloudbuild.yaml that were identified. To isolate changes as much as possible, these changes are introduced independently.

The main changes were:
1. Add ginkgo randomize flags
2. Cleanup up Cloudbuild.yaml to define env variables in a single place
3. Define ginkgo flags in a single place in the makefile

## Ginkgo Randomize
It is preferable to run our tests with randomize flags included. This was done with our ginkgo v2 update in other libraries. To reduce noise, I introduced this change as part of this PR and plan to fix all tests which fails as a result.

https://onsi.github.io/ginkgo/#spec-randomization

This is recommended by ginkgo, because tests should be independent of one another, and having tests which rely on being executed in a certain order, could mean:
1. We don't fully understand what we are testing and how to do it.
2. The underlying code doesn't work, since it depends on incorrect assumptions in the test (less likely, though possible)

## Consolidate Make Targetes
Previously we had different targets execute ginkgo, with different flags. Now we have a single target `make test`, which is run with the same flags.

We still expose extra targets as syntactic sugar to run unit tests or regression tests, but now under the hood they all rely on the same `test` target.

This allows us to manage our ginkgo configuration and flags in a single place

## Ginkgo Report and Coverage Flags
In ginkgo v2 there are useful flags for generating reports and coverage results. I included them in this change, but commented them out.

### Gateway Translator Tests
Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a436b5dc-57b6-4a73-89d6-8895c64ebd29

These tests attempt to verify that when running a Gateway object through translation, with a variety of global settings and annotations applied, that we produce the proper Proxy object. 

When we changed the order (randomize) it illustrated that we were sharing state between tests causing test pollution. This is now fixed

### Proxy Reconciler Tests
Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=a2086be9-fd95-4fb0-8fdc-59a726a7d9e2

We were sharing a Proxy client (with an in memory cache) across all tests. This meant we weren't isolating tests properly. I moved the initialization of the client into a BeforeEach so it is clean for each test.

### Stats Plugin Tests
Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=c490ab4b-f8ae-4eba-bca3-07365ad3d9be

Moving setup code to beforeEach solved the problem

### Glooctl Tests
I ran CI a few times and noticed a bunch of races in these tests:

Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=ce15dba6-2957-4813-8c78-810e6646a94f

Instead of using CI to identify these, I ran
```
 TEST_PKG=projects/gloo/cli/pkg/cmd/... make run-tests
```
until there were no more races.

In General, the reason for the race was the usage of a shared client. The solution was to ensure that our utility methods receive shared variables as arguments.

### Hybrid Client Tests
Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=c06f429a-3921-4f73-9bd1-9732695980a6

Reset the shared singleton to nil before each test, since one of the test overrides the value. This wasn't an issue before because we ran tests in the same order.

### Aws Plugin Tests
These weren't affected by randomize, but they depended on an envoy binary which isn't present for local development. I updated the test to be able to run locally. Huddled with @ben-taussig-solo @nfuden @jbohanon to confirm the changes.

## Makefile Changes for Forks
I ran the make tests locally (https://github.com/solo-io/gloo/tree/master/test/make) and noticed certain steps were skipped because the environment thought it was running from a fork. 

I updated the logic in our makefile and our test to correctly identify forks using the SSH remote url.

To verify this, I forked gloo, and from within that fork performed:
```
git remote get-url origin
```

the output was
```
git@github.com:sam-heilbron/gloo.git
```

# Checklist:


Makefile
```diff
@@ -201,36 +201,59 @@ install-go-tools: mod-download install-test-tools ## Download and install Go dep
...
 
 .PHONY: run-ci-regression-tests
 run-ci-regression-tests: install-test-tools  ## Run the Kubernetes E2E Tests in the {KUBE2E_TESTS} package
-	# We intentionally leave out the `-r` ginkgo flag, since we are specifying the exact package that we want run
-	$(GINKGO_ENV) $(DEPSGOBIN)/ginkgo -ldflags=$(LDFLAGS) -randomizeSuites -randomizeAllSpecs -failFast -trace -progress -race -failOnPending -noColor ./test/kube2e/$(KUBE2E_TESTS)
-
-.PHONY: check-format
-check-format:
-	NOT_FORMATTED=$$(gofmt -l ./projects/ ./pkg/ ./test/) && if [ -n "$$NOT_FORMATTED" ]; then echo These files are not formatted: $$NOT_FORMATTED; exit 1; fi
+run-ci-regression-tests: TEST_PKG = ./test/kube2e/$(KUBE2E_TESTS)
```
bewebi, line 253: What's this syntax where the target is repeated? Any why use it?

projects/gloo/pkg/upstreams/hybrid_client_test.go
```diff
@@ -81,6 +81,9 @@ var _ = Describe("Hybrid Upstream Client", func() {
 			&api.QueryMeta{LastIndex: 100},
 			nil,
 		).AnyTimes()
+
+		// In certain test we override this, so we need to default to nil before each test
```
bewebi, line 85:
```suggestion
		// In certain tests we override this, so we need to default to nil before each test
```

Makefile
sam-heilbron: My understanding is that this is one way to enable writing multi-line recipes, specifically enabling us to execute one make target from another.

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=30deaa28-2a64-4599-824b-3e66426816ee

make test failure, retrying while I investigate locally

bewebi (dismissed):
LGTM

bewebi (dismissed):

bewebi (approved):

gunnar-solo (approved):
Make it so.

PR #7834, opened 2023-02-10
Support RespectDnsTtl and DnsRefreshRate in Upstream API

sam-heilbron:
# Description

- Expose `RespectDnsTtl` and `DnsRefreshRate` in Upstream API
- Add `-failFast` and `failOnPending` back to Ginkgo test flags

# Context
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery

## Respect Dns Ttl
The RespectDnsTtl is an optional configuration for setting cluster’s DNS refresh rate. If the value is set to true, cluster’s DNS refresh rate will be set to resource record’s TTL which comes from DNS resolution.

## Dns Refresh Rate
For DnsRefreshRate, If the DNS refresh rate is specified and the cluster type is either STRICT_DNS, or LOGICAL_DNS, this value is used as the cluster’s DNS refresh rate.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-dns-refresh-rate

## Ginkgo Failure Flags
As part of recent work (https://github.com/solo-io/gloo/pull/7824/files) I had inadvertently removed a couple of useful ginkgo flags(-failOnPending -failFast). The result of this was that when build-bot failed, on e2e tests, we flooded the logs with Envoy information, because we didn't immediately fail all the tests in that suite (which run in parallel).

### Unit Testing
- Tests to validate that translator updates cluster fields
- Tests to validate that our upstream updater handles these fields during discovery

### E2E Testing
- Tests to validate that when configured on an Upstream, these get accepted by the proxy and are available via the configuration dump

# Checklist:


projects/gloo/api/v1/upstream.proto
```diff
@@ -161,6 +162,14 @@ message Upstream {
     // (bool) If set to true, Envoy will ignore the health value of a host when processing its removal from service discovery.
     // This means that if active health checking is used, Envoy will not wait for the endpoint to go unhealthy before removing it.
     google.protobuf.BoolValue ignore_health_on_host_removal = 22;
+
+    // If set to true, Service Discovery update period will be triggered once the TTL is expired.
+    // If minimum TTL of all records is 0 then dns_refresh_rate will be used.
+    google.protobuf.BoolValue respect_dns_ttl= 29;
+
+    // Service Discovery DNS Refresh Rate.
+    // Minimum value is 1 ms. Values below the minimum are considered invalid.
```
ben-taussig-solo, line 171: Nit: Can we include that the cluster type must be either STRICT_DNS, or LOGICAL_DNS for this setting to take effect?

test/e2e/dns_test.go
```diff
@@ -0,0 +1,101 @@
...
+		testContext.JustBeforeEach()
+	})
+
+	JustAfterEach(func() {
+		testContext.JustAfterEach()
+	})
+
+	Context("Defined on an Upstream", func() {
+		// It would be preferable to assert behaviors
+		// However, in the short term, we assert that the configuration has been received by the gateway-proxy
```
ben-taussig-solo, line 39: I understand that it might not be possible because we're validating the presence of the envoy config and not the DNS refresh behavior, but if possible could you add a test to confirm that `DnsRefreshRate` only applies if the cluster type is STRICT_DNS, or LOGICAL_DNS?

projects/gloo/api/v1/upstream.proto
```diff
@@ -161,6 +162,14 @@ message Upstream {
     // (bool) If set to true, Envoy will ignore the health value of a host when processing its removal from service discovery.
     // This means that if active health checking is used, Envoy will not wait for the endpoint to go unhealthy before removing it.
     google.protobuf.BoolValue ignore_health_on_host_removal = 22;
+
+    // If set to true, Service Discovery update period will be triggered once the TTL is expired.
+    // If minimum TTL of all records is 0 then dns_refresh_rate will be used.
+    google.protobuf.BoolValue respect_dns_ttl= 29;
+
+    // Service Discovery DNS Refresh Rate.
```
ashishb-solo, line 170: nit: Would be helpful to be explicitly mention that the unit is milliseconds

test/e2e/dns_test.go
sam-heilbron: Yeah it's an interesting case. Per the docs, the value is ignored if it's set but the cluster type isn't correct. So I figure we have 2 options:
1. Have the control plane prevent against that explicitly
2. Document the behavior and let people figure it out for themselves

I'm happy to move forward with option #1. I figure instead of an e2e test, I could add the logic and validate it with a unit test. How does that sound?

projects/gloo/api/v1/upstream.proto
sam-heilbron: Yeah good call, I'll update

projects/gloo/api/v1/upstream.proto
sam-heilbron: We expect a duration, so I'm not sure it needs to be milliseconds. Am I missing something?

test/e2e/dns_test.go
ben-taussig-solo: That sounds good to me!

projects/gloo/api/v1/upstream.proto
ashishb-solo: you're right! silly of my to miss that

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3518,6 +3519,58 @@ var _ = Describe("Translator", func() {
 			table.Entry("When value=nil", nil, false))
 	})
 
+	FContext("DnsRefreshRate", func() {
```
ashishb-solo, line 3522: gotta remove the focus here

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3518,6 +3519,58 @@ var _ = Describe("Translator", func() {
...
+				cluster = clusterResource.ResourceProto().(*envoy_config_cluster_v3.Cluster)
+				Expect(cluster.GetDnsRefreshRate()).To(refreshRateMatcher)
+			},
+			table.Entry("DnsRefreshRate=nil", nil, BeNil(), BeNil()),
+			table.Entry("DsnRefreshRate valid", &duration.Duration{Seconds: 1}, MatchProto(&duration.Duration{Seconds: 1}), BeNil()),
+			table.Entry("DsnRefreshRate=0", &duration.Duration{Seconds: 0}, BeNil(), MatchError(ContainSubstring("dnsRefreshRate was set below minimum requirement"))),
+		)
+	})
+
+	Context("DnsRefreshRate", func() {
```
ben-taussig-solo, line 3542: This context and the table contained within it have the same names as those above -- could we update one of them?

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3518,6 +3519,58 @@ var _ = Describe("Translator", func() {
...
+		table.DescribeTable("Sets DnsRefreshRate on Cluster",
+			func(staticUpstream bool, refreshRate *duration.Duration, refreshRateMatcher types2.GomegaMatcher, reportMatcher types2.GomegaMatcher) {
+				if staticUpstream {
```
ben-taussig-solo, line 3545: Nit: Can we convert this to just `if !statusUpstream {` for brevity?

test/e2e/dns_test.go
ben-taussig-solo: New tests + functionality looks good! I left a couple of nits around the organization of the new translator tests

projects/gloo/pkg/translator/translator_test.go
sam-heilbron: good call! UPdated

projects/gloo/pkg/translator/translator_test.go
sam-heilbron: Good catch. I had copied thse with the intention of making different tests and then forgot about it. Deleted the unnecessary duplicate

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=5f24d1f2-c319-4826-b36e-f558993b114b

These test flakes are being addressed in https://github.com/solo-io/gloo/pull/7857

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=df0c2cb3-015d-4cf9-99c9-21945cd00a56 dfp failure again

ben-taussig-solo (dismissed):

ashishb-solo (approved):

ashishb-solo:
/kick - buildbot timed out connection to `postman-echo`

ashishb-solo (approved):

ben-taussig-solo (approved):

PR #7846, opened 2023-02-13
Ginkgo v2

sam-heilbron:
# Description

Update Ginkgo to use v2

# Context
This PR cannot merge until the following work is complete:
- [x] https://github.com/solo-io/solo-projects/pull/4584  Gloo Enterprise uses Ginkgo v2 passing CI

There are a lot of repeated changes in this PR. Below are the main types of changes that exist

I followed https://onsi.github.io/ginkgo/MIGRATING_TO_V2

## Makefile
Instead of specifying a Path for all commands, export the Path to include the local bin directory where we install required dependencies for codegen

Update the Ginkgo version to v2, and all of the flags to the new v2 syntax

## go.mod/go.sum
All of our tests which use ginkgo v2 must reference libraries which also use ginkgo v2. Therefore, we had previously merged work in:
- https://github.com/solo-io/go-utils/pull/506
- https://github.com/solo-io/k8s-utils/pull/32
- https://github.com/solo-io/solo-kit/pull/537

We updated to use these dependencies, and then pinned versions of Kubernetes, since there are broader changes required to support kube 1.24 or 1.25.

## Ginkgo v2
A simple find/replace of the ginkgo library import

## Custom Reporters
The reporting infrastructure (https://onsi.github.io/ginkgo/MIGRATING_TO_V2#improved-reporting-infrastructure) improved. It also no longer supports custom reporters, so I removed any instances of these.

## Cleanup Other Tests
I encountered the following logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=f63d7b3a-1f94-48ef-b97a-ff74e40c8cc3 which demonstrated that certain conditions for tests which were valid before, are not valid in v2.

Specifically, you cannot define a BeforeSuite within a Node, so for a variety of tests, I had to move the Suite setup/teardown into the Suite file itself (as most of our other tests already do)

Example commit: https://github.com/solo-io/gloo/pull/7846/commits/05f47bb0902144480cbe6a64e3978d1c77178db8

## Go Leak Detection
We have a utility that can be leveraged to identify goroutine leaks. In this function, we ignore the top ginkgo goroutine that exists when running tests. In the v1 -> v2 upgrade, the path to this goroutine changed and needed to be updated. Without the update, we would fail the test since we identified one more goroutine after the test than before it.

Relevant commit: https://github.com/solo-io/gloo/pull/7846/commits/5121bb402c9bc100adcfe0604b1ecf0be795e5ef

## Ginkgo Parallel Processing
In Gloo, as part of a final cleanup of ginkgo config (https://github.com/solo-io/gloo/pull/7846/commits/659b6cd79455fd14d0d36c17eac3cccc6791f77f) I moved towards using the GinkgoParallelProcess which is the recommended API for parallelizing specs as opposed to the old config.GinkgoConfig.ParallelNode.

I then encountered races in tests (https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4daef080-c518-4f7f-b228-3f9b26829a30)

The solution was to move our logic into a single place that is loaded once at init, and then reference it where we need this logic.

Relevant commit: https://github.com/solo-io/gloo/pull/7846/commits/c45e12ac8faf3acda34702d3fe3eb83c1b046930

# Checklist:


go.mod
```diff
@@ -343,10 +339,23 @@ replace (
...
+
+	// Pin Kube libraries to v0.23
+	// These should be upgraded collectively
+	k8s.io/api => k8s.io/api v0.23.16
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.23.16
+	k8s.io/apimachinery => k8s.io/apimachinery v0.23.16
+	k8s.io/cli-runtime => k8s.io/cli-runtime v0.23.16
+	k8s.io/client-go => k8s.io/client-go v0.23.16
+	k8s.io/code-generator => k8s.io/code-generator v0.23.16
+	k8s.io/component-base => k8s.io/component-base v0.23.16
```
jackstine, line 352: just for context, how did you know that these contain the full set of dependencies?

jackstine (commented):
Looking good to me 👍

go.mod
sam-heilbron: I tried to pin all the kube libraries, and when I encountered build issues I would add some. The goal was just to try to keep all the libraries in sync

projects/gloo/cli/pkg/cmd/upgrade/upgrade_suite_test.go
```diff
@@ -1,15 +1,117 @@
...
 )
 
 func TestCheck(t *testing.T) {
 	RegisterFailHandler(Fail)
-	junitReporter := reporters.NewJUnitReporter("junit.xml")
-	RunSpecsWithDefaultAndCustomReporters(t, "Upgrade Suite", []Reporter{junitReporter})
+	RunSpecs(t, "Upgrade Suite")
 }
+
+var (
```
gunnar-solo, line 24: can you help me understand the context behind this file's changes?

test/ginkgo/parallel/ports.go
```diff
@@ -0,0 +1,23 @@
...
+var (
+	parallelGinkgoProcesses int
+)
+
+// GetParallelProcessCount returns the parallel process number for the current ginkgo process
+func GetParallelProcessCount() int {
+	return parallelGinkgoProcesses
+}
+
+// GetPortOffset returns the number of parallel Ginkgo processes * 1000
```
bewebi, line 18: nit/non-blocker: Any particular reason to multiply by 1000 as opposed to any other method?
It seems like these ports may be more likely to conflict with other services being run locally by a user (ie I've run stuff locally on 3000 and 8000 before)
It may be possible to choose more obscure ports by doing `+ 7089` or something else arbitrary like that

projects/gateway/pkg/translator/hybrid_translator_test.go
```diff
@@ -412,20 +411,20 @@ var _ = Describe("Hybrid Translator", func() {
...
-							* set		| object exists and has populated subfield						|	ssl.OneWayTls.Value = true
-						Since our merge operations take in (childSsl, parentSsl, PreventChildOverrides), this leaves 4*4*2 = 36 possible combinations to test.
-
-						When performing all of said combinations on a _nested_ field, such as ssl.OneWayTls.Value, there are 4 possible results:
-							* True		| ssl.OneWayTls.Value is true
-							* False		| ssl.OneWayTls.Value is false
-							* Nil		| ssl or OneWayTls is nil
-							* Error		| translation failed; when either parent XOR child has ssl defined, we should yield an error on the report
-						(┏(･o･)┛♪┗ (･o･) ┓			┏(･o･)┛♪┗ (･o･) ┓			┏(･o･)┛♪┗ (･o･) ┓			┏(･o･)┛♪┗ (･o･) ┓			┏(･o･)┛♪┗ (･o･) ┓)*/
+						  I am sitting here, possibly having descended into madness.  In my madness, I have found it necessary/expedient to construct a 36-entry table.
```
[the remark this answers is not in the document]
sam-heilbron: ha! no its a great table/comment, I just wanted to be apart of it and take credit for it

test/ginkgo/parallel/ports.go
sam-heilbron: Yeah I just took the existing solution which used the 1000 multiplier, but you're right the offset doens't need to be that

projects/gloo/cli/pkg/cmd/upgrade/upgrade_suite_test.go
sam-heilbron: Sure thing! This pertains to the "Cleanup Other Tests" section of the PR.

You'll notice this code was moved from the `upgrade_test` file into the`upgrade_suite_test`. Previously, we were defining BeforeSuite and AfterSuite functionality in the test itself, which ginkgo v2 complains about. So this change. isactually just a cut/paste into the appropriate place.

in the "CLeanup Other tests" section there is an example lg output from a failure and a commit with an example of the specfic fix

projects/gateway/pkg/translator/opts.go
```diff
@@ -23,7 +23,7 @@ type Opts struct {
 	DevMode                        bool
 	ReadGatewaysFromAllNamespaces  bool
 	Validation                     *ValidationOpts
-	ConfigStatusMetricOpts         map[string]*metrics.Labels
+	ConfigStatusMetricOpts         map[string]*metrics.MetricLabels
```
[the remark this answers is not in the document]
sam-heilbron: I can try to find the exact commit, but I noticed when upgrading our libraries that the type `Labels` conflicted with an existing metrics library. So I updated our defined type. It's just used internally so I believe it is a safe change

test/ginkgo/parallel/ports.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7846/commits/c45e12ac8faf3acda34702d3fe3eb83c1b046930 is the commit that made this change so you can see the previous code that was in place

sam-heilbron: I agree that we could probably find a better way to offset ports, and for the most part this was introduced so we can run tests in parallel. Though to be honest, I don't think a lot of this port offset work is even necessary. but I figured I would laeve the code as close to what is was like as part of this work

gunnar-solo (approved):

test/ginkgo/parallel/ports.go
bewebi: Yeah I did see that that was previously defined
If it ain't broke no need to fix it

bewebi (approved):

sam-heilbron:
/kick restart CI, claiming there was a build error due to multiple merges

gunnar-solo (approved):

Issue #7845, opened 2023-02-13
Update trivy version to latest


nfuden:
### Version

None

### Is your feature request related to a problem? Please describe.

we are missing some checks that may be caught with the newer trivy

PR #7866, opened 2023-02-20
Trivy version

Bslabe123:
# Description

`security-scan-scheduled` now always gets the latest version of Trivy. To test, run each of these lines manually (Note: dpkg isn't on Mac by default)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7845


.github/workflows/trivy-analysis-scheduled.yaml
```diff
@@ -37,10 +37,11 @@ jobs:
 
           echo "After clearing disk space:"
           df -h
-      - name: Install Trivy (0.30.3)
+      - name: Install Trivy (latest)
         run: |
-          wget https://github.com/aquasecurity/trivy/releases/download/v0.30.3/trivy_0.30.3_Linux-64bit.deb
-          sudo dpkg -i trivy_0.30.3_Linux-64bit.deb
+          TRIVY_VERSION=$(curl --silent "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
```
ben-taussig-solo, line 42: Interested in input from others, but I think it might be preferable to just upgrade to a more recently released fixed version of trivy than to always get the newest version. I think it can be convenient to know exactly which version of trivy CI is running

nfuden: The case where this would occur is mainly if we are trying to understand a difference between local and remote scans right? So I propose we do the following. Add a log line (echo) the trivy version so that we can tell in the case we do need to debug but still have this pull latest. That way we dont have to do more bumps over time

ianmacclancy: I agree with the echo - Once thats in can approve

Bslabe123: Added!

nfuden (approved):

ianmacclancy (approved):

Issue #7867, opened 2023-02-20
Upgrade test failures on community PRs


ianmacclancy:
### Gloo Edge Version

1.14.x (beta)

### Kubernetes Version

None

### Describe the bug

Community PR test failures on upgrade tests 
https://github.com/solo-io/gloo/actions/runs/3727614502/jobs/6337457950
another instance: https://github.com/solo-io/gloo/actions/runs/3792280442/jobs/6455349446

This looks to be a failure to instantiate the test helper pod. 

Splitting off these issues from https://github.com/solo-io/gloo/issues/7517
### Steps to reproduce the bug

In order to replicate this - create a fork off of the gloo repo and put a pr in against master

### Expected Behavior

Test run

PR #7869, opened 2023-02-20
Fix Community PR tests

ianmacclancy:
# Description
Update tests to add github token for upgrade testing and so that tests run locally on forks
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7867


test/kube2e/upgrade/upgrade_test.go
```diff
@@ -77,9 +77,18 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
...
 
-		LastPatchMostRecentMinorVersion, CurrentPatchMostRecentMinorVersion, err = upgrade.GetUpgradeVersions(beforeSuiteCtx, "gloo")
-		if err != nil && strings.Contains(err.Error(), upgrade.FirstReleaseError) {
-			firstReleaseOfMinor = true
+		lastPatchMostRecentMinorVersion, currentPatchMostRecentMinorVersion, err = upgrade.GetUpgradeVersions(beforeSuiteCtx, "gloo")
+		fmt.Println("-------------------------------")
+		fmt.Println("lastPatchMostRecentMinorVersion: " + lastPatchMostRecentMinorVersion.String())
+		fmt.Println("currentPatchMostRecentMinorVersion: " + currentPatchMostRecentMinorVersion.String())
+		fmt.Println("-------------------------------")
+		if err != nil {
```
jenshu, line 85: should we do the error checking before trying to print the lines above, since those values could be nil

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -46,18 +46,18 @@ func GetUpgradeVersions(ctx context.Context, repoName string) (lastMinorLatestPa
...
 
-func getLastReleaseOfCurrentMinor(repoName string) (*versionutils.Version, error) {
+func getLastReleaseOfCurrentMinor() (*versionutils.Version, error) {
 	// pull out to const
 	_, filename, _, _ := runtime.Caller(0) //get info about what is calling the function
-	fmt.Printf(filename)
 	fParts := strings.Split(filename, string(os.PathSeparator))
 	splitIdx := 0
-	//we can end up in a situation where the path contains the repo_name twice when running in ci - keep going until we find the last use ex: /home/runner/work/gloo/gloo/test/kube2e/upgrade/junit.xml
+	//In all cases the home of the project will be one level above test - this handles forks as well as the standard case /home/runner/work/gloo/gloo/test/kube2e/upgrade/junit.xml
```
jenshu, line 54: would it be more future-proof for the caller of this func to pass in the relative path from itself to the changelog dir?
e.g. if this is called from test/kube2e/util.go it would pass in `../../changelog`

test/kube2e/upgrade/upgrade_test.go
ianmacclancy: Good idea! I'll move them

test/kube2e/upgrade/upgrade_test.go
```diff
@@ -77,9 +77,18 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
...
+			if strings.Contains(err.Error(), upgrade.FirstReleaseError) {
+				firstReleaseOfMinor = true
+			} else {
+				fmt.Println(err.Error())
+				return
+			}
```
nfuden, line 91:
```suggestion
			if !strings.Contains(err.Error(), upgrade.FirstReleaseError) {
				fmt.Println(err.Error())
				t.Fail(err)
				return
			}
			firstReleaseOfMinor = true
```

test/kube2e/upgrade/upgrade_utils.go
nfuden: that might be safest as we tried to be tricky such that this could be moved around inside of our repo and not break but obviously it has issues. Also while we are here side not lets make this comment a comment and not a directive by adding a space after the //

ianmacclancy: Yup! Can add the space. Working off test as the location is fairly future-proof as I cannot see a case where we re-org the main packages in the repo. Happy to try the relative path suggestion

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -103,13 +92,13 @@ func newLatestPatchForMinorPredicate(versionPrefix string) *latestPatchForMinorP
 	}
 }
 
-func getLatestReleasedVersion(ctx context.Context, majorVersion, minorVersion int) (*versionutils.Version, error) {
+func getLatestReleasedVersion(ctx context.Context, repoName string, majorVersion, minorVersion int) (*versionutils.Version, error) {
```
ianmacclancy, line 95: This will allow re-use in enterprise

.github/actions/regression-tests/action.yaml
```diff
@@ -58,6 +58,7 @@ runs:
     run: ./ci/deploy-to-kind-cluster.sh
   - name: Testing - kube e2e regression tests
     env:
+      GITHUB_TOKEN: ${{ github.token }}
```
sam-heilbron, line 61: What impact does this have? This change should affect anytime this job runs, so if our regression tests passed from the Gloo repository without this, can you help me understand why this is necessary for tests run from forks?

test/kube2e/upgrade/upgrade_utils.go
sam-heilbron, line 95: If we intend to use this from within Enterprise, we'll need to export this method as well

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -33,34 +33,23 @@ func (a ByVersion) Less(i, j int) bool {
...
-	//we can end up in a situation where the path contains the repo_name twice when running in ci - keep going until we find the last use ex: /home/runner/work/gloo/gloo/test/kube2e/upgrade/junit.xml
-	for idx, dir := range fParts {
-		if dir == repoName {
-			splitIdx = idx
-		}
-	}
-	pathToChangelogs := filepath.Join(fParts[:splitIdx+1]...)
-	pathToChangelogs = filepath.Join(pathToChangelogs, changelogutils.ChangelogDirectory)
-	pathToChangelogs = string(os.PathSeparator) + pathToChangelogs
+	pathToChangelogs := filepath.Join(filename, fmt.Sprintf("../../../../%s", changelogutils.ChangelogDirectory))
```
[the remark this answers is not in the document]
ianmacclancy: It's relative for upgrade_utils to the main package. This should always be true unless we change the package structure itself.

.github/actions/regression-tests/action.yaml
ianmacclancy: It adds the token to the running tests - Im assuming this is available in our normal runs but on a community PR its not present. I validated that the token is not present without this addition.

test/kube2e/upgrade/upgrade_utils.go
ianmacclancy: this is called by GetUpgradeVersions which is exported. Previously GetUpgradeVersions did not pass the value to getLatestReleasedVersion

sam-heilbron: ah thanks, I missed that

.github/actions/regression-tests/action.yaml
sam-heilbron: Yeah that makes sense, but I'm curious about `assuming this is available in our normal runs but on a community PR its not present`. Why is this the case since the workflow file is the same for both

test/kube2e/upgrade/upgrade_utils.go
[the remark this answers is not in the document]
ianmacclancy: Reverting no non-relative path - spoke offline. We are fine with the original solution

.github/actions/regression-tests/action.yaml
sam-heilbron: I think it's important to figure out because we have https://github.com/solo-io/gloo/blob/master/.github/workflows/nightly-tests.yaml#L4. So it seems that it's not immediately clear what part is necessary.

My guess:
The GITHUB_TOKEN at the top level needs to be defined. In forks, we don't have acccess to secrets, so we use the `github.token`. I imagine if we just need the top level definition, and it needs to be updated to ensure the token is not pulled from a secret so that forks work

ianmacclancy: We are going to test this out in another PR as a fast follow

sam-heilbron (approved):

jenshu (approved):

PR #7881, opened 2023-02-23
fix incorrect merge with master

ianmacclancy:
# Description
Fixup some issues that occurred during recent merges into the master branch

# Context
## Makefile
In a recent PR, https://github.com/solo-io/gloo/pull/7846, we updated our Makefile to no longer explicitly specify the PATH to use for running executables. Instead, we exported the path and then calling the executable directly.

This worked in CI and locally for @sam-heilbron . However, we determined that it worked locally, because I had updated the PATH variable in my `.zshrc`.

For those who had not, the following error would appear, when running:
```
make install-go-tools generated-code -B
``` 

```
make: gettercheck: No such file or directory
```

The solution was to update our references to prefix the path with the `DEPSGOBIN`.

In addition, I made some changes to break down the `generated-code` recipe into smaller targets, and remove and unncessary call to `gofmt` which is handled by `goimports`.

## Upgrade Tests
Cleanup unused methods 
GinkoV2 merge missed from master

# Testing
## Makefile
1. Pull down this branch
2. Inspect your local $PATH and confirm that there is no `_output/.bin` included
3. Run `make -B install-go-tools generated-code`
4. Confirm there are no errors


elcasteel:
looks like codegen is failing with format issues which I assume is from not running it before pushing and not the changes to the makefile.

elcasteel (approved):

gunnar-solo (approved):

sam-heilbron:
Adding WIP label to prevent merges. We still need to update the PR context to outline what the issue was in Upgrade tests

Issue #7887, opened 2023-02-24
DOCS: Helm enterprise docs are not correctly versioned


nfuden:
### Gloo Edge Version

1.14.x (beta)

### Kubernetes Version

None

### Describe the bug

the helm docs are always that of main branch rather than the release they should be 

### Steps to reproduce the bug

look at docs site

### Expected Behavior

The docs show helm info for the version of edge selected

PR #7874, opened 2023-02-22
docs: update to consume LTS branches, rather than patches

gunnar-solo:
# Description
"simple" change to `active_versions.json`, such that docs site produces docs based on LTS branches, rather than patches

# Context
This is in support of making docs changes easier/faster to implement

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7825

PR #7891, opened 2023-02-24
docs: enterprise proto docs now generate based on minor version

gunnar-solo:
# Description
When generating docs for LTS branches copy the `changelog` dirs to allow generation to run correctly

# Context
When users consume our [enterprise helm reference](https://docs.solo.io/gloo-edge/1.11.52/reference/helm_chart_values/enterprise_helm_chart_values/), they were previously seeing _only_ the reference for the `latest` release, as opposed to the specific minor version of the site which was being viewed

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7887

### DIFF
--- a/.github/workflows/docs-gen.yaml
+++ b/.github/workflows/docs-gen.yaml
@@ -103,6 +103,9 @@ jobs:
       uses: actions/checkout@v3
       with:
         ref: master
+    - name: Override master checkout if deploying PR test
+      if: ${{ github.event_name == 'pull_request' }}
+      uses: actions/checkout@v3
     - name: Detect Community PR
       id: community-pr-check
       if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != 'solo-io/gloo' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/docs-gen.yaml
```diff
@@ -103,6 +103,9 @@ jobs:
       uses: actions/checkout@v3
       with:
         ref: master
+    - name: Override master checkout if deploying PR test
```
sam-heilbron, line 106: Can you help me understand the purpose of this and how it related to the [USE_PR_SHA_AS_MASTER](https://github.com/solo-io/gloo/blob/master/.github/workflows/docs-gen.yaml#L118) env variable?

Happy to chat more about this offline, but I had initially introduced this because mostly our docs build the master branch, but on PRs we need to have them build off of the PR (to test changes introduced in the PR).

My guess is that this change performs the same action as that env variable I linked. I'm more than happy to move forward in this direction, but if we are, can we cleanup that env variable if it's no longer used?

sam-heilbron (commented):
Looks awesome! One question

.github/workflows/docs-gen.yaml
gunnar-solo: sure!  This actually fixes a bug I introduced in https://github.com/solo-io/gloo/pull/7874.  Specifically:

those [nice test docs builds](https://github.com/solo-io/gloo/pull/7891#issuecomment-1444093939) were being hardcoded to use `master` after my change, as opposed to the originating PR

sam-heilbron (approved):

Issue #7795, opened 2023-02-06
Kube2e Suite Test Flakes (Gateway)


nfuden:
### Gloo Edge Version

1.14.x (beta)

### Kubernetes Version

None

### Describe the bug

• Failure [90.455 seconds]
Robustness tests
/home/runner/work/gloo/gloo/test/kube2e/gateway/robustness_test.go:49
  Updates Envoy endpoints, even if proxy is invalid
  /home/runner/work/gloo/gloo/test/kube2e/gateway/robustness_test.go:182
    works, even when snapshot cache is reset [It]
    /home/runner/work/gloo/gloo/test/kube2e/gateway/robustness_test.go:412

### Steps to reproduce the bug

https://github.com/solo-io/gloo/actions/runs/4100894504/jobs/7072145044
nightlies

### Expected Behavior

not flake

sam-heilbron:
Perhaps this should live as its own issue, but I noticed that for a while we had all tests passing:
![Screen Shot 2023-02-09 at 4 20 26 PM](https://user-images.githubusercontent.com/5279627/217941805-22cfcab7-5126-4355-8805-13a6596fad20.png)

And then we added the randomize flags to our regression tests, and now we get consistent failures in the gateway suite.
![Screen Shot 2023-02-09 at 4 20 33 PM](https://user-images.githubusercontent.com/5279627/217941836-6f17a316-4ff8-4c61-8276-9f600def6bd9.png)

sam-heilbron:
Not the robustness test, but in the gateway suite: https://github.com/solo-io/gloo/actions/runs/4147743292/jobs/7174979757

ben-taussig-solo:
Another faillure in the gateway suite: https://github.com/solo-io/gloo/actions/runs/4170677836/jobs/7219876688

sam-heilbron:
Gateway suite: https://github.com/solo-io/gloo/actions/runs/4177616562/jobs/7235413367

sam-heilbron:
https://github.com/solo-io/gloo/actions/runs/4188850210/jobs/7260482639

sam-heilbron:
https://github.com/solo-io/gloo/actions/runs/4188851627/jobs/7260487895

sam-heilbron:
This has been resolved on the 1.14 branch. If you see failures in this  Suite, please re-open and comment with the logs to the failure. In the meantime, I am closing.

sam-heilbron:
[...]

PR #7768, opened 2023-01-30
build(regression): added randomize flags to ginkgo runners

gunnar-solo:
# Description
Use Ginkgo's "-randomizeSuites -randomizeAllSpecs" flags when running cron builds

caveat:  [current Ginkgo docs](https://onsi.github.io/ginkgo/#running-multiple-suites) seem to indicate that the correct flags are titled `--randomize-all` and `--randomize-suites`, rather than the current usage.  This is because we are still using the (deprecated) v1

# Test
* [manual OSS run](https://github.com/solo-io/gloo/actions/runs/4047255584)
  * drilling into https://github.com/solo-io/gloo/actions/runs/4047255584/jobs/6961060079#step:3:1597, for example, shows the specified flags being passed

# Checklist:
- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #7897, opened 2023-02-27
Stability: Deflake Gateway Regression Suite Tests

sam-heilbron:
# Description
Changes to ensure the Gateway test suite passes consistently

# Context

### Cleanup
Updated our code which attempted to match that two proto messages are identical, to use a shared matcher, instead of a custom one built in our kube suite. 

### Leader Election
Fixed a bug in our implementation to allow an env variable to override the default value. Also expose other overrides for the toggles that are exposed in our leader election implementation.

### Helm Values in Tests
We previously defined these values in a string and then wrote them to a file. I decided to just migrate these definitions to a file in the suite package.

### When did this suite start failing?
We [merged code](https://github.com/solo-io/gloo/pull/7768) to randomize the order of tests within our regression (kube2e) test suites.

### How frequently does this suite fail?
It's challenging to get the exact numbers. We have some [testspace metrics](https://solo-io.testspace.com/spaces/137684/metrics?timespan=recent) which measure our test suites when they merge into the main branch. These metrics indicate that we're pass **44%** of the time. While this isn't necessarily all due to this test suite, it is the main culprit.

You can also see on recent merges to master those results. The image is included in a [comment on the attached Github issue](https://github.com/solo-io/gloo/issues/7795#issuecomment-1424856885)

#### Lease Duration
We have a set of [logs from a failure](https://github.com/solo-io/gloo/actions/runs/4188851627/jobs/7260487895) which indicate that the failure was:
```
2023-02-15T23:10:52.4073879Z       Expected
2023-02-15T23:10:52.4074155Z           <string>: Status
2023-02-15T23:10:52.4074428Z       to match fields: {
2023-02-15T23:10:52.4074669Z       .State:
2023-02-15T23:10:52.4074899Z       	Expected
2023-02-15T23:10:52.4075211Z       	    <core.Status_State>: 0
2023-02-15T23:10:52.4075465Z       	to equal
2023-02-15T23:10:52.4075762Z       	    <core.Status_State>: 2
2023-02-15T23:10:52.4075999Z       }
```

This indicates that the resource we are checking, did not have a status on the object, but expected a particular status. 

A resource would not have a status for 2 reasons:
1. The resource wasn't processed in the Sync. This may occur if it wasn't selected, or if the emitter didn't initiate a snapshot sync.
2. If a leader hadn't been elected. When running multiple replicas, only the leader will write statuses (to avoid conflict). 

#### Test Pollution
I need to investigate this further. However, there are situations where we write Kube services before a test, and that causes Gloo discovery to dynamically create an Upstream to route traffic to that Service. We then delete the Service after the test, but leave the Upstream around, which causes errors in our logs.

#### Lease Duration
I inspected the generated lease, and it always had a lease duration of 15 seconds (the default), even though we set an env variable to override this. In the end, there was a small logic error in our reading of the variable. Previously we would only overwrite if there was an error parsing the duration string. What we wanted is to overwrite if there was NOT an error parsing the string.

#### Test Pollution
I disabled leader election, and ran the tests locally a number of times, as well as got 3 consecutive passes in CI. Therefore, I don't think test pollution is at play here.

### How can we be confident these changes will fix the failures?
I ran these tests a number of times locally. And only put it up for review after 5 consecutive passes in CI.

# Checklist:


test/kube2e/helm/artifacts/helm.yaml
```diff
@@ -0,0 +1,37 @@
+global:
```
gunnar-solo, line 1: off-topic, perhaps:  I usually consider artifacts _generated_ files from a set of compute.  What does it mean in this instance?

test/kube2e/util.go
```diff
@@ -71,49 +69,6 @@ func GlooctlCheckEventuallyHealthy(offset int, testHelper *helper.SoloTestHelper
 	}, timeoutInterval, "5s").Should(BeNil())
 }
 
-func GetHelmValuesOverrideFile() (filename string, cleanup func()) {
```
gunnar-solo, line 74: very ok with having this moved to (several) yaml files

test/kube2e/gateway/gateway_suite_test.go
```diff
@@ -93,9 +93,6 @@ func installGloo() {
 	// Check that everything is OK
 	kube2e.GlooctlCheckEventuallyHealthy(1, testHelper, "90s")
 
-	// Explicitly enable strict validation,
-	kube2e.UpdateAlwaysAcceptSetting(ctx, false, testHelper.InstallNamespace)
```
gunnar-solo, line 97: help me understand this change, please?

sam-heilbron: Yeah a great callout. 

This line runs after we install Gloo, and it works by modifying the Settings resource, to set `alwaysAccept=false`. When alwaysAccept=true, our validation webhook accepts all resource updates, regardless of whether they are valid or invalid.

I kept this behavior, but changed they way we set it. Before we installed gloo, then modified the settings resource. Now I pass a helm value `alwaysAcceptResources` so that it is set at install time.

test/kube2e/util.go
sam-heilbron: Me too!

test/kube2e/helm/artifacts/helm.yaml
sam-heilbron: Yeah it's a good point. I don't have a great name for it. I wanted some generalized folder for files that we would use (be it helm files like in this case, or other manifests). I'm open to any naming suggestions.

I was following the artifacts pattern here: https://github.com/solo-io/gloo/tree/master/test/kube2e/knative

.github/workflows/docs-gen.yaml
```diff
@@ -13,7 +13,7 @@ on:
     - 'v1.10.x'
   pull_request:
     branches:
-      - 'master'
+    - 'master'
```
nfuden, line 16: Were we not running docs gen?

test/kube2e/helm/artifacts/helm.yaml
nfuden: Ive often see testdata folders used but that is often when its a series of different tests rather than the one setup for all the tests

test/kube2e/helm/helm_test.go
```diff
@@ -514,8 +514,9 @@ func makeUnstructuredFromTemplateFile(fixtureName string, values interface{}) *u
 }
 
 func installGloo(testHelper *helper.SoloTestHelper, chartUri string, fromRelease string, strictValidation bool, additionalInstallArgs []string) {
-	valueOverrideFile, cleanupFunc := kube2e.GetHelmValuesOverrideFile()
-	defer cleanupFunc()
+	cwd, err := os.Getwd()
+	Expect(err).NotTo(HaveOccurred())
```
nfuden, line 518:
```suggestion
	Expect(err).NotTo(HaveOccurred(), "working dir could not be retrieved while installing gloo")
```

nfuden:
overall seenms reasonable to me

.github/workflows/docs-gen.yaml
sam-heilbron: No each syntax works for lists. I just noticed that in the above "push branches" list we don't indent, so I tried to make this match

test/kube2e/helm/helm_test.go
sam-heilbron: Good call, pushed this to each of the places I introduced this: https://github.com/solo-io/gloo/pull/7897/commits/9fff95fdc37e3a01238761c09dfe90b157241b04

gunnar-solo (approved):

test/kube2e/helm/artifacts/helm.yaml
gunnar-solo: since we're copying existing patterns, I've no issue

ianmacclancy (approved):
LGTM

PR #394, opened 2019-02-09
Spell check

yuval-k:
Mea culpa mea culpa mea culpa!
note there are slight api changes to too spelling errors

PR #7910, opened 2023-03-02
Improve Build-Bot Execution

sam-heilbron:
# Description

Improve how long it takes our build-bot to execute the cloudbuild configuration

# Context

## Cloudbuild Changes
We have a single cloudbuild job, which runs a number of steps to execute tests, build assets and publish assets. This is slow to iterate on, so below are the results of the changes in this PR and an outline of the strategy

### Results:
Before: Exact results are challenging to find, but https://storage.googleapis.com/solo-public-build-logs/log-a2311fb1-9995-4b58-be4e-f9d974b7e57d.txt is the most recent merge to master, and I generally see it take between **35-40 minutes**

After: [Logs from this PR](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=acb842ff-22a2-4c00-b4ca-488217c873de). [Test logs take ~19 minutes](https://console.cloud.google.com/cloud-build/builds/c6a197e4-9f3f-4a0a-abb5-e7b2b4078bf1?project=907026730415). [Artifact logs take ~10.5 minutes](https://console.cloud.google.com/cloud-build/builds/38b9c8ef-5c23-404e-a4fa-bc766a83c7c1?project=solo-public). Since we run these in parallel, the completed build takes **19.5 minutes**

### Tests Sub-Build
The tests run as their own build, and are only run on PRs and not releases (like they were before).

The main difference here is that I setup our e2e tests and non e2e tests to run in separate steps, in parallel. 

[Logs from a recent run](https://console.cloud.google.com/cloud-build/builds/c6a197e4-9f3f-4a0a-abb5-e7b2b4078bf1?project=solo-public) indicate that unit tests take 11 minutes and e2e take 17.

### Artifacts Sub-Build
[Logs from a recent run](https://console.cloud.google.com/cloud-build/builds/38b9c8ef-5c23-404e-a4fa-bc766a83c7c1?project=solo-public) indicate that this build takes ~11 minutes minutes to complete. Now we can use CI to publish assets relatively quickly and not wait for tests to pass before doing so.

### Deleted Steps
Certain steps were removed the original cloudbuild:
- Proxycontroller: This executed a make target which does not exist. [Internal Slack ctx](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1677778226742009)
- Spelling check: This script was executed but alway failed quietly. I added an explanatory comment in the script, and removed the step from our cloudbuild

## Cleanup Up Files
`BUILD.md` and `artifacts.yaml` were unused in the project.

# Follow-Up
This does not resolve the linked issue. That issue can track the work required to speed up building and publishing assets or finding more creative ways to speed up test execution.

# Checklist:


sam-heilbron:
/kick staged transformation e2e failure, perhaps a flake? 

https://console.cloud.google.com/cloud-build/builds/e440e7b7-7ec9-43e9-9abb-e1073f8a347d;step=7?project=solo-public

artifacts.yaml
```diff
@@ -1,40 +0,0 @@
-# For building Gloo Edge artifacts with Valet
```
kdorosh, line 1: i can't believe this is still in the repo! what a throwback

ci/spell.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+# Update March 2, 2023
+# This is a potentially powerful script, and one that has been running in CI but doing nothing:
```
kdorosh, line 4: fwiw, I have definitely seen this work in the past. I've seen spelling errors in our docs get caught and fail ci. but perhaps this has been broken a long time

sam-heilbron: Yeah I agree. I'd be open to finding the git history and figuring out when those files were deleted and re-adding them so that spellcheck runs as part of this. What do you think?

As much as the comment may be helpful for explaining why the script doesn't run, I fear if I leave it as is, we'll lose track of it, so I'd almost rather just fix it now

artifacts.yaml
sam-heilbron: haha yeah the number of times I almost deleted this file as part of other work is too much to count

ianmacclancy (commented):
Looks good to me, As this is a similar change to what we have in solo projects we should align the two projects on folder structure and filenames for cloudbuilds. I like the structure and naming here - I think we could add a small readme to the cloudbuild folder to outline what each sub built is responsible for. 

Only other thought is that we stripped our the extended docker builds in enterprise. Can we do the same here? That will also save us some time

ci/spell.sh
kdorosh: yeah I'd prefer to fix it but timebox investigation to half hour -> hour max. Otherwise seems not worth it. Also if it's not technically too complex we can see if the docs team has bandwidth? That's the real value of this script

ci/cloudbuild/publish-artifacts.yaml
```diff
@@ -0,0 +1,119 @@
...
+  waitFor:
+  - 'gcr-auth'
+  - 'prepare-gcr-zone'
+  - 'docker-push-extended'
+
+secrets:
+- kmsKeyName: 'projects/solo-public/locations/global/keyRings/build/cryptoKeys/build-key'
+  secretEnv:
+    GITHUB_TOKEN: 'CiQABlzmSYYiveU0gTxGH2139eaBUedjV3vNCxQmJU+nRPlfQ/YSUQCCPGSGzbGp49fwDYuefAx9V94b8mivdp9AqB7zQAa07VtGJmrGdg9ZuhKGFrrgqxwABE0LLVNHyngCSHYSYMH8Vn/mRtT7wQuEHBlKVGtqPw=='
+    QUAY_IO_PASSWORD: 'CiQABlzmSRx5TcOqbldXa/d/+bkmAfpNAWa3PTS06WvuloZL+vASaQCCPGSGCogonVZVEUNx4G3YJtWi18gSuNx4PvLe08q8xAflTMFkjsyQirAOK3Y2oCvgYwiw/ITcuydjkpMjxDygFyENXS9FKFJoAXHlPQE5qidKr8xxmxF5ezhmjGB0gjyjXIIkbSEnBg=='
```
ashishb-solo, line 104: are these supposed to be publicly visible?

ci/spell.sh
kdorosh: I _suspect_ it's more likely we changed the directory ci runs in rather than deleted the files being referenced, but I have not searched the repo for them

nfuden: We did change from not specifying dir to specifying the dir at one point but going back to 1.4.x i still dont see the whitelists

nfuden: Worst case we could just add the fields or add this to the docs folder.

cloudbuild.yaml
```diff
@@ -1,170 +1,29 @@
...
-    path: '/go'
+    - '-c'
+    - |
+      if [ -z "$TAG_NAME" ]; then
+      gcloud builds submit --config=ci/cloudbuild/run-tests.yaml --substitutions COMMIT_SHA=$COMMIT_SHA,REPO_NAME=$REPO_NAME,TAG_NAME=$TAG_NAME,_PR_NUM=$_PR_NUM;
+      fi
+  waitFor: [ '-' ]
+  id: 'run-tests'
+
+timeout: '3600s'
```
nfuden, line 26: Should this be larger? If sub job 1 starts first and job 2 takes a while to start we no longer have the full 3600s for the second job

ci/cloudbuild/run-tests.yaml
```diff
@@ -0,0 +1,128 @@
...
...UGcahIWpl0Ll6QdJ2NhFkMGZgkr33yDl4FTGbkeSgfuf5Ou2kDis9QPg4GSjS7hTmSMCcmeawRK6CFC/yLHQoEuipNtK9D1n063FmEZK7ZftsGE7L+TQjn5HHjU78qJ4I7K2oH1cRle+PgWq35zG70VByhpYP9uBJD2f+GEjzCXLTBm05Y4bk1ODEhSCN6GlqGisQcNhA95M7RXXlNTM1K6ZGqUK7KUCLPbpI8j2svv5fqw3ZcZEM51Q0NSUPKS+tEb+yD+OwZ5cQQymLz2H7O5fPQtYyFTnko1sOfVfSpphjebudaL4agZKM2FWrS4SC1qAbDe2ymoQMRybNDTllJIopflit1q0fBQvmkVC3Iu6njsjCLtPSqJ5KBVNwwm3GPG59+kqN3PnogDP64JmOqkpiENmA6k+QHI2DsvYoMkkpXOct9L/+tt2oqiVHUBRkaqFyzt5Do2CSkjpziwyXRmE8e5PrXUli8U/EbohFN7fi6YjY4qQF9XptGUHI0wgMXi+uBbxwWLN/HiAEX7EqQCY7Vpu+u5E4mceTG/M69C/CMtlFGyfPWakWjR8XNBlsawr7edQKupQJ3dW1HZL/lMmAXWiPp4SCEix+IwSdhI2qPYOdbZeCH37CktYzbGRgxdqVifCZOCtQwktjq7beOPJaYGMKC8md/0u6yaUkFmsd8Uu6rw8LbZtqM6T9XSDgCHegouUlP4A/mWF1t35BDZ8IJl9lxtwmXjQkQcjWnOMKTL+2BJ87CHTlgHAOiy5jtNCgF9ZU0qA6R4l15i6avOru58XzVK92CMgEFd67pna69IjpgnYawqi3UiEtArX2Ru5S8IGFe8SSiivvkeFZWkLOQxvhWUFI+7+vpritnt4lwKJua7Yd7GU+i0RJ1Jbvz8eFWFwfLko7okM6WfLDwkTHIWwuPgqTM2Zmaj7ayjZxxFVpEiKFAScVVoUoz1WKz1mEVTh5U/YkQFClDt/cfJLmTKjFeubOiH0nYp8le8JTC58syynU5jnZ3uZ7RnjcAWF0shFCCnxhgp4S7zU0NVE/ZuscpkFhjsGda095Bdmfq/uOO88eFB6ouRop4UqsaTqxV63KZ/7ZICm2N6+w4EC1ra2Ixl76N1Ix6IycvJEMvMSPhTwRDVco0dfvsF++vvhujVeoui7mw15x7g8ZJqXi3ND7mTlWXad2D2bQoUog=='
```
elcasteel, line 107: should this key be public?

elcasteel: I see that it's actually moved from another file so never mind

sam-heilbron: These always have been public: https://github.com/solo-io/gloo/blob/master/cloudbuild.yaml#L153. I figured as part of this to leave it as is. 

It looks like there are improvements to the secrets API that we can use, but I think these are safe. They are encrypted and then decrypted by our kms key (https://cloud.google.com/build/docs/securing-builds/use-encrypted-credentials#configuring_builds_to_use_encrypted_data). However, there are ways to use a secrets manager and not expose these at all.

I figured we can follow up to handle this later as it has been this way for some time. But open to other ideas, what do you think?

cloudbuild.yaml
sam-heilbron: Both jobs are started immediately, so I would be surprised if there was a delay. The jobs themselves take 11 and 19 minutes respectively, so I don't think we're near to the ceiling. Certainly open to bumping it if you think it's valuable. What would you propose?

ci/cloudbuild/publish-artifacts.yaml
sam-heilbron: Responded to this here: https://github.com/solo-io/gloo/pull/7910#discussion_r1124812016

Interested in your thoughts. can you follow up in that thread if you have more ideas?

sam-heilbron:
> > The tests run as their own build, and are only run on PRs and not releases (like they were before).
> 
> This makes me a bit uneasy to be honest. If someone ran skip-ci on the latest pull request into the release branch, we could theoretically be pushing releases out the door that weren't fully tested on the latest commit. Do we feel that this is potentially unsafe, or is it not too serious of an issue?

I don't quite follow, though would love to better understand your concerns. The behavior hasn't changed: build-bot runs on PRs and not on releases. You can skip build-bot on PRs using the /skip-ci directive. The difference is that previously build-bot would run as a single build. Now we run it as a build that spins off 2 parallel sub-builds.

sam-heilbron:
> Looks good to me, As this is a similar change to what we have in solo projects we should align the two projects on folder structure and filenames for cloudbuilds. I like the structure and naming here - I think we could add a small readme to the cloudbuild folder to outline what each sub built is responsible for.
> 
> Only other thought is that we stripped our the extended docker builds in enterprise. Can we do the same here? That will also save us some time

Yeah I like the README idea, ill add it. 

When you say strip the extended builds, do you mean not building extended images? We certainly can do that, but I thought we wanted to build as much as possible? Plus, our publish-artifacts step takes 10 minutes and run-tests take 19 minutes, so even if we improve our artifact building, tests will still be the limiting factor

ci/spell.sh
sam-heilbron: Yeah so I searched the repo and couldn't find them using:
```
git log --all --full-history -- "**/ci/spelling_whitelist_words.*"
```

I also checked out the PR that introduced this script: https://github.com/solo-io/gloo/pull/394 and that didn't have the referenced files either.

sam-heilbron: When I added the files and gave them placeholder data, I ran the script again and hit:
```
fatal error: runtime: bsdthread_register error
```

I think given the findings, it's hard to get a quick solution for having this working. I'm leaning towards just leaving this as is (with a comment on the script). What do you think?

Makefile
```diff
@@ -245,14 +245,15 @@ test-with-coverage: test
 	go tool cover -html $(OUTPUT_DIR)/coverage.cov
 
 .PHONY: run-tests
-run-tests: install-test-tools ## Run all tests, or only run the test package at {TEST_PKG} if it is specified
-run-tests: GINKGO_FLAGS += -skip-package=kube2e
-ifneq ($(RELEASE), "true")
```
sam-heilbron, line 250: We can remove this guard in the taget itself, because the entire cloud build isn't kicked off if we're a release.

elcasteel (approved):

ci/cloudbuild/run-tests.yaml
ashishb-solo: yeah, i noticed that they were public in other places too. i think it's probably not appropriate to address those issues in this pull request because it would make the scope too large. additionally, if this is something we want to resolve, then we probably have to audit the entire repository for secrets like this, rather than just addressing it in the current one. so i agree overall: leave it for now, but preferably try to address it as soon as possible.

ci/spell.sh
nfuden: I think that the error you are getting is due to an old install as older versions of go would sometimes throw this in certain circumstances on mac

sam-heilbron: https://github.com/solo-io/gloo/issues/7923

nfuden (approved):

PR #7921, opened 2023-03-06
Stability: Fix Route Transformation Test Flake

sam-heilbron:
# Description

Update our route transformation e2e tests to fix a flake that occurred intermittently in CI.

# Context

## Fix Flake
I believe the flake occurs because we attempt to communicate with `httpbin.org`. This means that if there is a network failure during the CI processing, the test will fail.

Our other tests rely on a local echo server, so I updated this test to use that same pattern.

### Testing
There's no perfect way to validate that a fix is resolved. However, you can run this 21 times in a row, by doing the following:
1. Focus the Describe block in the route_transformation file
2. Execute:
```
ENVOY_IMAGE_TAG=1.14.0-beta8 TEST_PKG=./test/e2e WAIT_ON_FAIL=1 GINKGO_USER_FLAGS=-repeat=20 make test
```

## Code Cleanup
Most of our other e2e tests use the same [e2e test framework](https://github.com/solo-io/gloo/blob/master/test/e2e/test_context.go). This framework is intended to abstract a lot of the details of standing up resources, and allows developers to write expressive tests, only modifying the resources needed for the test. I updated this test to follow that framework.

We introduced an [http matching library](https://github.com/solo-io/gloo/blob/master/test/gomega/matchers/http.go) which allows us to easily assert http responses. This test had some of that same logic written in the test, so I updated it to use this shared library.

## Export Default Route Name
In the test context, we now export the default route name, instead of relying on the hardcoded "test" string.

I also updated the name to have the word route in it. Luckily, this exposed the tests which relied on that hardcoded value matching, so I updated those places to reference the constant.

## NothingWrittenErr Flake
While this PR was running through CI, I encountered a diffferent failure:
```
The function passed to Eventually failed at /workspace/gloo/test/e2e/staged_transformation_test.go:208 with:
  Unexpected error:
      <*url.Error | 0xc0076d4510>: {
          Op: "Post",
          URL: "http://localhost:11147/",
          Err: <http.nothingWrittenError>{
              error: <*errors.errorString | 0xc001da5ca0>{
                  s: "http: ContentLength=4 with Body length 0",
              },
          },
      }
      Post "http://localhost:11147/": http: ContentLength=4 with Body length 0
```

This occurs when we create an http.Request object, and then retry that same object within an Eventually. The fix is to create a unique object for each retry.

# Checklist:


test/e2e/route_transformation_test.go
```diff
@@ -360,60 +189,122 @@ var _ = Describe("Transformations", func() {
...
-				Transformations: transform,
-			}
 		})
-		It("should error on non-json body when ignoreErrorOnParse/parseBodyBehavior/passthrough is disabled", func() {
-			WriteVhost(vh)
 
-			// execute request -- expect a 400 response
-			ExpectUnsuccessfulResponse(GetHtmlRequest())
+		It("should error on non-json body when ignoreErrorOnParse/parseBodyBehavior/passthrough is disabled", func() {
+			transform.ResponseTransformation.GetTransformationTemplate().IgnoreErrorOnParse = false
```
sam-heilbron, line 195: I added this line, to explicitly set the IgnoreErrorOnParse. I found that explicitly defining it made it easier to read, especially since the test description talks about it being "disabled" so it through me off at first when I didn't see any configuration disabling it.

sam-heilbron:
/kick

prepare-go-workspace flaked
```
Step #0 - "prepare-workspace": go: cuelang.org/go@v0.3.2 requires
Step #0 - "prepare-workspace": 	golang.org/x/exp@v0.0.0-20210126221216-84987778548c: reading https://proxy.golang.org/golang.org/x/exp/@v/v0.0.0-20210126221216-84987778548c.mod: 502 Bad Gateway
Step #0 - "prepare-workspace": {"level":"fatal","ts":"2023-03-06T18:37:42.609Z","caller":"src/main.go:26","msg":"unable to complete cloud build preparation","error":"exit status 1","stacktrace":"main.main\n\t/src/main.go:26\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
Finished Step #0 - "prepare-workspace"
```

test/e2e/route_transformation_test.go
```diff
@@ -2,351 +2,180 @@ package e2e_test
...
-						},
-					},
-				},
-			}},
+				}
+
+				return vs
+			})
+
+			Eventually(func(g Gomega) {
```
ben-taussig-solo, line 83: How do you feel about bringing back `ExpectSuccess` or something like it (admittedly, the naming of the original method was not very descriptive). It seems like we're making the same series of checks at the end of a couple of these tests and it might be cleaner to have a helper do that

test/e2e/route_transformation_test.go
```diff
@@ -360,60 +189,122 @@ var _ = Describe("Transformations", func() {
...
-			err := json.NewDecoder(res.Body).Decode(&body)
-			Expect(err).To(HaveOccurred())
+			testContext.PatchDefaultVirtualService(func(vs *v1.VirtualService) *v1.VirtualService {
+				vs.GetVirtualHost().Options = &gloov1.VirtualHostOptions{
+					Transformations: transform,
+				}
+				return vs
+			})
+
+			Eventually(func(g Gomega) {
```
ben-taussig-solo, line 227: Same as above, I think we can extract the contents of this `Eventually` into a helper

test/e2e/route_transformation_test.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7921/commits/396a17eba20de9dce5dff55f01c048e8f1321846

test/e2e/route_transformation_test.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7921/commits/396a17eba20de9dce5dff55f01c048e8f1321846

test/e2e/route_transformation_test.go
```diff
@@ -2,351 +2,177 @@ package e2e_test
...
-		ExpectSuccess()
-	})
+		// EventuallyResponseTransformed returns an Asynchronous Assertion which
+		// validates that a request with a body will return the requested content.
+		// This will only work if the above transformation is applied to the request
+		EventuallyResponseTransformed := func() AsyncAssertion {
+			return Eventually(func(g Gomega) {
+				req, err := http.NewRequest(
+					http.MethodPost,
+					fmt.Sprintf("http://localhost:%d/1", defaults.HttpPort),
```
ianmacclancy, line 81: We can probably have a util method for this - buildDefaultLocalHostPath("1") - takes in the path after localhost+port.

test/e2e/route_transformation_test.go
```diff
@@ -2,351 +2,177 @@ package e2e_test
...
-			Name:    "virt1",
-			Domains: []string{"*"},
-			Routes: []*gloov1.Route{{
-				Options: &gloov1.RouteOptions{
+		It("should fail if no transform defined", func() {
+			testContext.PatchDefaultVirtualService(func(vs *v1.VirtualService) *v1.VirtualService {
+				vs.GetVirtualHost().Options = &gloov1.VirtualHostOptions{
+					Transformations: nil,
+				}
+
```
ianmacclancy, line 97: Nit: extra whitespace here and line 100.

test/e2e/route_transformation_test.go
sam-heilbron: Yeah that's fair. We follow this pattern pretty extensively in all other e2e tests, so if we did that, Id rather do in a standalone PR. I wonder if a utility method to abstract a 1 line string concat will, make it harder to track what url is actually being requested, and not get us much in the form of saved space/knowledge. Open to it if you think there's value

test/e2e/route_transformation_test.go
sam-heilbron: good call! I was about to say "I generally split off return statements to be their own block" but then realized i didn't follow that for other tests. Updating to remove the space.

sam-heilbron: https://github.com/solo-io/gloo/pull/7921/commits/664dc31445b99d0b8996062ecc681bff9672c05c

test/e2e/route_transformation_test.go
ianmacclancy: Are there cases where we are sending requests to something other than localhost? Its more cleanup of hardcoded values that can be declared in one place rather than many. The value that's important for the test is the "/1"

ianmacclancy: Not a blocker on this PR - just an observation

sam-heilbron: Yeah that's a good point. When we run our envoy wrapper for our in-memory tests, we port-forward the relevant ports (https://github.com/solo-io/gloo/blob/master/test/services/envoy.go#L563), which is why we can make the requests to localhost. I think it's safe to assume that for these e2e tests, requests will be to localhost. 

I like your point that when we have frequently used patterns/constants/assumptions, even if they are small in code, its good to aggregate those in one place, both to outline why we are safe with that assumption, and remove any chance that developers will not follow the pattern.

What if we built this utility specifically for our e2e tests at first since that's where we know it's used? And then if there are opportunities to use it more broadly we can think about expanding the scope.

ben-taussig-solo (approved):

ashishb-solo (dismissed):

ashishb-solo (approved):

sam-heilbron:
/kick
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4f934309-8e30-4bd5-b26e-061412cffc64
```
[38;5;9mFull Stack Trace[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests":     github.com/solo-io/gloo/test/e2e_test.glob..func6.3.6()
Step #1 - "run-tests": Step #8 - "run-e2e-tests":     	/workspace/gloo/test/e2e/consul_test.go:239 +0x99a
Step #1 - "run-tests": Step #8 - "run-e2e-tests": [38;5;243m------------------------------[0m
```

I haven't seen this much and there is no issue for it. If it happens again we should create aan issue

ben-taussig-solo (approved):

PR #7942, opened 2023-03-09
test(gateway): add test specs to better detect future flakes

gunnar-solo:
# Description
hit a build flake, here: https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1678253879426959.  The idea is that the flake will be more addressable if it happens after this merges

# Checklist:


sam-heilbron (commented):
I like the direction, though I think we can simplify the changes further.

To test this locally, you could introduce a new test case (or modify an existing one) to fail the test intentionally. Then you can determine if the output of that failure is enough metadata to help developers know which test case failed, what the true output was and what the expected output was.

When these tests fail in CI, is there enough information in CI to detail what happened? Most logic around this will be in the validation webhook logs, so I wonder if we want to set those to debug for this test suite for a week while we investigate this flake. If we have enough context already, we don't need to do that.

sam-heilbron (approved):

gunnar-solo:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=1e18717f-a02c-49e8-9a40-39d60b5d8dfe

elcasteel (approved):

PR #7943, opened 2023-03-10
upgrade retry-go

jenshu:
Upgrade to latest retry-go. We are upgrading it in solo-projects to pick up new features, and solo-projects relies on gloo for some test utils that use retry-go, and it's easier if they're both on the same version.


jenshu:
closing for now; some of our other dependencies are still on an older version

PR #7972, opened 2023-03-15
Add UUID to api keys

rinormaloku:
Required to implement the following GH issue: https://github.com/solo-io/gloo-mesh-enterprise/issues/7756

Which has the option for API Keys to be hashed and in order to query API keys we need the UUID


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -977,6 +977,8 @@ message ApiKey {
     repeated string labels = 3;
     // additional data the client needs associated with this API key
     map<string, string> metadata = 4;
+    // Optional: Unique identifier for the API key
+    string uuid = 5;
```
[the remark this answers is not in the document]
rinormaloku: > Is the intention to make the field api_key optional if only a uuid is specified?

That's not the intention. In Gloo Portal we are adding an option to hash the ApiKey and only keep the hashed value stored in the database.

And because we never store the ApiKey and we wouldn't like to show the hashed value in the UI, we need some other way to identify the keys.

For example for this key creation request:

```
{
  "cleartextapikey": {
     "metadata": {
        "environmentName": "dev",
        "apiProductName": "my-api",
        "usagePlanName":"premium",
        "email": "johndoe@email.com",
        "userName": "john.doe",
        "created-ts-unix":"1678440029",
    }
  } 
}
```

We store the following data in the database:

The hashed key
```
{
  "hashed_apikey": {
     "metadata": {
        "environmentName": "dev",
        "apiProductName": "my-api",
        "usagePlanName":"premium",
        "email": "johndoe@email.com",
        "userName": "john.doe",
        "created-ts-unix":"1678440029",
    },
    "uuid": "uuid-unique-for-this-hashed-key"
  } 
}
```

And the uuid pointing to the hashed key:
```
{
  "uuid-unique-for-this-hashed-key": "hashed_apikey" 
}
```

We additionally might extend the UUID key when stored in Redis to contain other information such as user details, etc. to help with listing the API keys of a user.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -977,6 +977,8 @@ message ApiKey {
     repeated string labels = 3;
     // additional data the client needs associated with this API key
     map<string, string> metadata = 4;
+    // Optional: Unique identifier for the API key
```
rinormaloku, line 980: @ashleywang1 can we use here an annotation to hide this from the public docs as this is an implementation detail for storing this info in the storage?

jbohanon (approved):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
rinormaloku: @jbohanon asked in gloo-platform-core and this doesn't seem to be supported

jmhbh (approved):

Issue #7932, opened 2023-03-09
E2E Suite Test Flakes (Consul)


sam-heilbron:
### Gloo Edge Version

1.14.x (beta)

### Kubernetes Version

None

### Describe the bug

The [e2e tests which integrate with consul](https://github.com/solo-io/gloo/blob/master/test/e2e/consul_test.go) are non-deterministic.

These tests interact with code that hasn't been updated in a while, so we believe that it is an issue with the test and not the underlying implementation.

Recent logs:
- https://console.cloud.google.com/cloud-build/builds;region=global/d892e067-dcff-4300-a354-e142735dad68;step=1?project=solo-public

```
Step #8 - "run-e2e-tests":   [FAILED] Failed after 0.001s.
Step #8 - "run-e2e-tests":   Expected
Step #8 - "run-e2e-tests":       <<-chan *v1helpers.ReceivedRequest | len:0, cap:100>: 0xc000e55f20
Step #8 - "run-e2e-tests":   to receive something.
Step #8 - "run-e2e-tests":   In [It] at: /workspace/gloo/test/e2e/consul_test.go:239 @ 03/08/23 20:55:42.181
Step #8 - "run-e2e-tests": 
Step #8 - "run-e2e-tests":   Full Stack Trace
Step #8 - "run-e2e-tests":     github.com/solo-io/gloo/test/e2e_test.glob..func6.3.6()
Step #8 - "run-e2e-tests":     	/workspace/gloo/test/e2e/consul_test.go:239 +0x99a
```

### Steps to reproduce the bug

See logs above from previous runs.

To reproduce, just run the test suite:
1. Focus the Describe block
2. Run `ENVOY_IMAGE_TAG=1.14.0-beta8 TEST_PKG=./test/e2e make test`

### Expected Behavior

Tests pass consistently and the reason for the flake is outlined in the PR that resolves it

elcasteel:
Without having read any of the code involved, an error like 
`Expected <<-chan *v1helpers.ReceivedRequest | len:0, cap:100>: 0xc000e55f20 to receive something.` makes me think that maybe there is a channel that should be blocking but isn't (easy mistake to make in go)

gunnar-solo:
https://console.cloud.google.com/cloud-build/builds;region=global/351eab32-5881-43f9-bd7b-0ff0ccbebf0b;step=1?project=solo-public

sam-heilbron:
another:

```
[...]

PR #7981, opened 2023-03-19
Simplify Http Request Building in E2E Tests

sam-heilbron:
# Description

# Context

# Checklist:


test/testutils/http.go
```diff
@@ -0,0 +1,143 @@
...
+
+	scheme   string
+	hostname string
+	port     uint32
+	path     string
+
+	body string
+
+	host    string
+	headers map[string]string
```
ben-taussig-solo, line 29: I think we should use `map[string][]string` here. We want to be able to send multi-value headers, and if we do this we can simplify lines 138-140 to `request.Header = h.headers`

test/e2e/buffer_test.go
```diff
@@ -79,7 +61,12 @@ var _ = Describe("buffer", func() {
 			})
 
 			It("valid buffer size should succeed", func() {
-				Eventually(testRequest(), 10*time.Second, 1*time.Second).Should(matchers.HaveExactResponseBody("{\"value\":\"test\"}"))
+				requestBuilder := testContext.GetHttpRequestBuilder().
```
ben-taussig-solo, line 64: can we create `requestBuilder` in a broader scope so we don't have to instantiate it in each of these tests?

test/e2e/test_context.go
```diff
@@ -248,8 +248,28 @@ func (c *TestContext) PatchDefaultUpstream(mutator func(us *gloov1.Upstream) *gl
...
 
-// Use a different function to create a test upstream call before testContext.BeforeEach()
+// SetUpstreamGenerator Use a different function to create a test upstream call before testContext.BeforeEach()
 // Used for example with helpers.NewTestGrpcUpstream which has the side effect of also starting a grpc service
 func (c *TestContext) SetUpstreamGenerator(generator func(ctx context.Context, addr string) *v1helpers.TestUpstream) {
 	c.testUpstreamGenerator = generator
 }
+
+// GetHttpRequestBuilder returns an HttpRequestBuilder to easily build http requests used in e2e tests
+func (c *TestContext) GetHttpRequestBuilder() *testutils.HttpRequestBuilder {
```
ben-taussig-solo, line 258: just curious, what's the rationale behind defining `GetHttpRequestBuilder` and `GetHttpsRequestBuilder` here instead of in `test/testutils/http.go`?

test/e2e/proxy_protocol_test.go
```diff
@@ -134,15 +124,15 @@ var _ = Describe("Proxy Protocol", func() {
 		Context("with TLS", func() {
 
 			BeforeEach(func() {
-				requestScheme = "https"
+				requestBuilder = testContext.GetHttpsRequestBuilder()
 				rootCACert = gloohelpers.Certificate()
 
 				secureVsToTestUpstream := gloohelpers.NewVirtualServiceBuilder().
```
nfuden, line 130: Since we are doing this to clean up things with regards to the builder perhaps its also time to consider having these actions not be on the builder interface but rather as options that you pass to the New or with an Apply? 
Also having a copy function might be worthwhile for these as we could then pass around an Anded set of options to make reuse more easily.

For example NewVirtualServiceBuilderFromOptions(WithName(e2e.DefaultVirtualServiceName),
					WithNamespace(writeNamespace))

test/testutils/http.go
```diff
@@ -0,0 +1,143 @@
...
+	h.port = port
+	return h
+}
+
+func (h *HttpRequestBuilder) WithPath(path string) *HttpRequestBuilder {
+	h.path = path
+	return h
+}
+
+func (h *HttpRequestBuilder) WithPostBody(body string) *HttpRequestBuilder {
```
ben-taussig-solo, line 77: Can we just make this `WithBody` and not set the request method here? I feel like there might be situations where we want to set a body but not change the request method (I.E., PUT request).

I could potentially be open to changing the method to POST if the current method is GET, but to be honest I think it would probably just be better to error in that case because I'm not sure if it's ideal for our test utils to have potentially misleading/unexpected behavior

test/e2e/proxy_protocol_test.go
sam-heilbron: Certainly open to it. I'm not entirely sure I follow. How does a constructor that accepts options differ from just a new builder that uses a fluent api and allows you to define the specific options you want?

My hope for a copy was that you can just re-use a builder or extend one:
```
builderEast := [initialize]

builderWest := builderEast.WithName("custom-name")
```

test/e2e/test_context.go
sam-heilbron: Yeah it's a fair question. My hope was to have:
1. A generic utility that we can use across all tests
2. Some specifics for our e2e test

I decided to export "GetHttpRequestBuilder" in the TestContext, because it initializes a request with details that are particular to our e2e tests. Specifically, it sets the port to the port we care about and the host to the host we need to make requests. Previously developers had to always set these 2 things but now they don't have to think about it becomes it comes by default.

test/e2e/buffer_test.go
sam-heilbron: Yeah it's a fair point and I'm happy to move it out in the scope. The real hope was the instead of re-using variables across tests, we can just have a test construct the resources it needs locally. Since some tests use http and others https, it seems that re-using some object across a suite might just confuse people. However, in this test we are clearly only testing certain functionality so I'm happy to inititalize it in the broader scope. What do you think?

test/testutils/http.go
sam-heilbron: Yeah my thinking of PostBody was to try to turn a common setter (set a body, set the method to post) into a single api call. I figured though we don't have it, as you say, we might need to add a `withPutBody` in the future if we need to suport it.

Again, the hope is just ease of use for developers. If you think separating this into 2 calls:
WithBody()
WithMethod()

would be easier, I'm more than happy to adjust

test/testutils/http.go
sam-heilbron: https://github.com/solo-io/gloo/pull/7981/commits/aa1a91970e4ab0e3c659d8d86831650b28f40ae7

let me know what you think about this approach. I tried to keep the syntax simple (a single string) because most times people will only set a single value. However, I made it so that you can pass a comma separate list in the case that you need to set multiple values. We don't have test cases that use this yet, but might in the future

test/e2e/proxy_protocol_test.go
sam-heilbron: To be clear, I'm not than happy to shift the API, just wanted to undersatnd what you saw as the benefits to a constructor with a set of option functions that we can pass in, versus just the existing pattern

test/testutils/http.go
ben-taussig-solo: This looks good! I feel a little strange about this because `,` is a valid character in HTTP headers. That being said, I don't expect to run into the situation where we intentionally send a header value that contains a literal `,` in our tests so I am resolving the thread

test/e2e/proxy_protocol_test.go
sam-heilbron: whoops! initially wrote that as "not that happy" but meant to write "more than happy". updated.

test/e2e/buffer_test.go
ben-taussig-solo: I see your point + I think the git diff tricked me a little bit here -- looking at the file as a whole, I think we probably lose some clarity by defining `requestBuilder` in a different scope. Given how many lines are between some of these tests, I think it's more expressive to just construct the requestBuilder in line

TL;DR on second glance this seems good as is

test/e2e/test_context.go
ben-taussig-solo: ah, makes perfect sense! thanks!

test/testutils/http.go
ben-taussig-solo: We already have `WithMethod`, so I think it makes sense to just make this `WithBody` and then every one of the `With..` methods only affects a single property on the `HttpRequestBuilder`. I think if `WithPostBody` is the only way to set the body property, it can potentially lead to confusion if someone doesn't realize that they're also modifying the method by calling it

sam-heilbron: Yeah that's fair. What do you think about https://github.com/solo-io/gloo/pull/7981/commits/f3726d8e7bcec2fabd1a965bd69f93147f5df537?

I made sure we expose both WithMethod and WithBody, and left WithPostBody around and a comment that outlines it just aggregates the two

ben-taussig-solo: Yeah that makes sense! I like `WithPostBody` as a convenience method!

test/testutils/http_request.go
```diff
@@ -0,0 +1,154 @@
...
+		headers:  make(map[string][]string),
+	}
+}
+
+func (h *HttpRequestBuilder) WithContext(ctx context.Context) *HttpRequestBuilder {
+	h.ctx = ctx
+	return h
+}
+
+func (h *HttpRequestBuilder) WithMethod(method string) *HttpRequestBuilder {
```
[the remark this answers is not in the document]
sam-heilbron: Yeah that's a great point, ill update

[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/pull/7981/commits/eb011ccccf3fa80a7006c6320b2df8a5d1af856d

sam-heilbron:
/kick https://github.com/solo-io/gloo/issues/7932#issuecomment-1477236420

test/e2e/proxy_protocol_test.go
nfuden: I like the signature more as you can build complex options as a unit with functional statements. That being said its not a big deal. The copy thing is a big deal. 
Unless I am missing something the thing I am worried about extending something is eventually we may have `sharedbuilder := NewBuilder().WithName("blah").WithSomething("else")

    Context(test1, blah){
         test1Builde := sharedBuilder.WithName("east")
    }
    Context(test2, bleegh){
      test2Builde := sharedBuilder.WithSomething("else")
      
     }

`
Because the with extensions modify the builder the name "east" leaks to test2 (or with randomize it may only sometimes leak)

nfuden: In general the option based approach means that we are more explicit about not passing  the structure just the modifications to a default.
so 
`requestBuilder := testContext.GetHttpRequestBuilder().
					WithPath("1").
					WithPostMethod()`
becomes 
`requestBuilder := testContext.GetHttpRequestBuilder(
					opts.WithPath("1"), 
					opts.WithPostMethod())`

or at times 
`
 sharedOpts := opts.And(opts.WithPath("1"), 
					opts.WithPostMethod()))
requestBuilder := testContext.GetHttpRequestBuilder(sharedOpts)

`

nfuden (dismissed):

nfuden (approved):

ben-taussig-solo (approved):

PR #7989, opened 2023-03-21
test(kube2e/gateway): ensure validation server successfully restarts

gunnar-solo:
# Description
When toggling validation server back _on_, added blocking statements to _ensure_ that it has successfully been enabled.

# Context
This patches a flake that presents itself as
```
[FAILED] Expected
     <string>: virtualservice.gateway.solo.io/method-matcher created  
to contain substring
     <string>: invalid route: routes with delegate actions must use a prefix matcher
```
which, on the face of it, is a failure in our validation server.

Looking more closely at our test suite, we can see that in several tests of the `kube2e/gateway` suite, we make use of this pattern:
```golang
BeforeEach(func() { kube2e.UpdateAlwaysAcceptSetting(ctx, true , testHelper.InstallNamespace) })
AfterEach (func() { kube2e.UpdateAlwaysAcceptSetting(ctx, false, testHelper.InstallNamespace) })
```
The _intention_ was to allow us to apply invalid resources to the test cluster, then observe us successfully creating snapshots that _exclude_ said invalid resources.  There was "a little" bit of slop, here, in that, while the validation server _was_ re-enabled after these tests, it didn't necessarily fully restart.

The thought process behind this PR is to _guarantee_ that our validation server restarts before progressing behind these "should not write bad resources to envoy" tests.

# Checklist:


sam-heilbron (commented):
Nice work! 

A few considerations:
1. Right now we manually call `verifyValidationWorks` after updating the setting. I fear that if we don't bake it into the utility method, future developers will forget to wait the appropriate time. If this only needs to be called when we set AlwaysAccept=false, perhaps we have a utilty method call: 'DisableAlwaysAccept', which calls https://github.com/solo-io/gloo/blob/master/test/kube2e/util.go#L135 under the hood, and the `waitForSettingsToPropagate` method is this `verifyValidationWorks`

2. Is there a way to verify this works? One option I had considered was that we previously have a 3 second delay for settings changes. Perhaps if we turned that down to 0 we would see the flake frequently, and then we could easily test without your change vs with your change. I'm not sure if you've considered other approaches but it would be nice if we had some manual testing steps (or just running this 7 times in CI consecutively)

test/kube2e/util.go
```diff
@@ -103,11 +103,15 @@ func UpdateDisableTransformationValidationSetting(ctx context.Context, shouldDis
 }
 
 // enable/disable strict validation
-func UpdateAlwaysAcceptSetting(ctx context.Context, alwaysAccept bool, installNamespace string) {
+func UpdateAlwaysAcceptSetting(ctx context.Context, alwaysAccept bool, installNamespace string, verifyUpdate func()) {
```
sam-heilbron, line 106: I like this! Right now `UpdateSettings` has a method that is `waitForSettingsToBePropagated` which is a 3 second sleep. It feels unncessary to both perform that AND the verifyUpdate. What if instead we just say:

```go
if verifyUpdate == nil {
  verifyUpdate := func() {
    time.Sleep(3 * time.Second)
  }
}
UpdateSettingsWithPropagationDelay(updateSettings, verifyUpdate, ctx, installNamespace)
```

test/kube2e/gateway/gateway_test.go
```diff
@@ -117,7 +117,7 @@ var _ = Describe("Kube2e: gateway", func() {
 
 			attempt += 1
 			return false
-		}, time.Second*15, time.Second*1).Should(BeTrue())
+		}, time.Second*60, time.Second*1).Should(BeTrue())
```
sam-heilbron, line 120: I'm nervous that we need to bump this up. Why would 15 seconds not be enough time? Did you run this locally and see that after 15 seconds the webhook resolved?

jackstine (commented):
Looking good, take consideration from Sam's comment.

gunnar-solo:
pausing this work to pick up https://github.com/solo-io/solo-projects/issues/4649

gunnar-solo:
closing via https://github.com/solo-io/gloo/issues/7795#issuecomment-1503399758

PR #7997, opened 2023-03-22
Stability: Fix Leader Elector Test Flake

sam-heilbron:
# Description

Fix a test flake in our leader elector test suite

# Context
## Theory on the Flake
I noticed that we would fail with 2 different errors: sometimes it was that the startupAction had been called twice when we expect it to be called once, and the other was that it was called 0 times when we expected it to be called once.

My first thought was that there was a race condition in the implementation itself. I performed some static analysis and re-ran individual tests in that suite repeatedly, but could not reproduce a flake.

This led me to think that something in the test was wrong. So in the tests, right after I started up the election watching, I applied:
```go
Consistently(func(g Gomega) {
  g.Expect(atomic.LoadUint64(&startupOps)).To(Equal(uint64(0)))
}).ShouldNot(HaveOccurred())
```

I noticed that when I could reproduce the flake, it was failing on this line. This meant that the flake was occurring before we had performed any context cancellation or channel closing, so it really felt like there was some pollution between tests.

I didn't exactly dig into what exactly is happening, but by removing the AfterEach section, I noticed that the flakes disappeared. Below I outline why this is a safe change.

## Why the fix is safe
These tests aim to verify 3 cases:
- That no action is performed without election
- That after an election a single action is performed
- That after a context is cancelled, election results are ignored

Previously I had split these into 3 cases. The impact was that I needed to cleanup the channel in an AfterEach, which was not closed by the test case for the first issue.

Really, the first test case can be validated in each of the subsequent tests. First we start an election, then assert that no action is performed and then perform some actions and validate them. These changes allowed me to delete the AfterEach cleanup because each test now closes the election channel itself.

## Testing
The following command can be used to run these tests repeatedly
```
TEST_PKG=pkg/bootstrap/leaderelector/... GINKGO_USER_FLAGS=-until-it-fails make test
```

I noticed that if I ran this repeatedly, I would see a failure within 30 or so repetitions. Sometimes it would take a few runs to reproduce. After, I have not been able to reproduce.

# Checklist:


sam-heilbron:
/kick https://github.com/solo-io/gloo/issues/7932#issuecomment-1480048417

sam-heilbron:
/kick https://github.com/solo-io/gloo/issues/7932#issuecomment-1480069393

jackstine (commented):
So I guess the `AfterEach` call to `close(electedChan)` was causing a race condition, in which the startupOps was incrementing after the `BeforeEach()` statement.

If this is not the reason, I am still unclear. I understand we were not closing this in the `It()` of the first test, but I think this would be taken care of in the `AfterEach()`. Because of this I am not sure what is the difference is.

nfuden (approved):

sam-heilbron:
> So I guess the `AfterEach` call to `close(electedChan)` was causing a race condition, in which the startupOps was incrementing after the `BeforeEach()` statement.
> 
> If this is not the reason, I am still unclear. I understand we were not closing this in the `It()` of the first test, but I think this would be taken care of in the `AfterEach()`. Because of this I am not sure what is the difference is with, or without the `AfterEach`

Yeah that is my suspicion. I didn't dig too deeply into it, but I outlined how I arrived there in the updated PR context (under "Theory on the Flake")

jackstine (approved):

PR #7999, opened 2023-03-23
Fix SslConfig Merge When Using isolateVirtualHostsBySslConfig

sam-heilbron:
# Description

Fix a bug in the logic we have to consolidate SslConfiguration which are identical except for SNI.

# Context

During the investigation into the attached issue, @bdecoste identified this bug. We attempt to merge configuration that is identical (or overlapping) and in that merge, we were dropping Virtual Services. The impact is that a VirtualService would be accepted by our webhook, and during processing, it would not be persisted in the Proxy object, and therefore not appear in the Envoy config_dump.

The fix is rather trivial: if there is a list, append, if there is not, create a new one. We had that logic inverted.

## Testing
I would have expected unit tests to catch this, but there were none, so I added them. These tests fail if you undo the fix.

We have a number of e2e tests around this space, but they didn't cover the case of 2 overlapping sslConfigs. I updated the tests to cover this case. Similarly, if you ran these tests without the fix, they would fail.

# Checklist:


test/helpers/virtual_services.go
```diff
@@ -208,6 +208,25 @@ func (b *virtualServiceBuilder) errorIfInvalid() error {
 	return nil
 }
 
+func (b *virtualServiceBuilder) Clone() *virtualServiceBuilder {
```
nfuden, line 211: Can we add a unit test to validate that this doesnt miss fields?

projects/gateway/pkg/translator/ssl_configuration_test.go
```diff
@@ -0,0 +1,113 @@
...
+				vsEast = helpers.NewVirtualServiceBuilder().
+					WithName("vs-east").
+					WithNamespace(defaults.GlooSystem).
+					WithDomain("east.com").
+					WithSslConfig(&ssl.SslConfig{
+						OneWayTls: &wrappers.BoolValue{
+							Value: true,
+						},
+						SniDomains: []string{
+							"east.com",
```
nfuden, line 74: Lets also validate with multiple snidomains

test/helpers/virtual_services.go
sam-heilbron: Yeah good call, I had added it for the other two but had neglected to do it for this builder, ill update

sam-heilbron: https://github.com/solo-io/gloo/pull/7999/commits/e97444facb22b20e190af406194c9503a9b4aee6

projects/gateway/pkg/translator/ssl_configuration_test.go
sam-heilbron: https://github.com/solo-io/gloo/commit/e7e0d824de16e9b7959e6dfd01151fc049024ec7 I updated the existing test to include multiple values, some of which were overlapping

test/helpers/helpers_suite_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"testing"
+
+	"github.com/solo-io/go-utils/testutils"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+func TestHelpers(t *testing.T) {
+	testutils.RegisterCommonFailHandlers()
+	RunSpecs(t, "Helpres Suite")
```
sam-heilbron, line 13: typo, fix!

nfuden (approved):

gunnar-solo (approved):
make it so.

PR #8011, opened 2023-03-27
API adding OIDC ciphers

jackstine:
# Description

API that enables encryption and ciphers within the userSession API. This provides the `cipher_config` attribute, which has the `keyRef` attribute to contain the type of `CipherKey` used.  

### Encryption Key Secret
A `CipherKey` can have a reference to a `EncryptionKeySecret`. The `EncryptionKeySecret` will look like. A key can be generated using `echo -n "<encryption-key>" | base64`, where `<encryption-key>` is 32 bytes.
```yaml
apiVersion: v1
kind: Secret
type: gloo.solo.io.EncryptionKeySecret
metadata:
  name: my-encryption-key
  namespace: gloo-system
  labels:
    app: my-label
    version: v1
data:
  key: "dGhpcyBpcyB0aGUgZW5jcnlwdGlvbiBzZWNyZXQga2U="
```

Adds the `EncryptionKeySecret` to the secret protos.

### OIDC Authorization Code
You can add the `cipherConfig.keyRef` as follows to the auth config API by applying it to the `session` of the `oidcAuthorizationCode`.
```yaml
spec:
  configs:
  - oauth2:
      oidcAuthorizationCode:
      # other properties
        session:
          cipherConfig:
            keyRef:
              name: my-encryption-key
              namespace: gloo-system
```

### Kube Secret converter
Any kube secret with the following information will be converted into an EncryptionKeySecret. Note the `encryptionKey` must be present in the Secret.
```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: my-encryption-key-2
  namespace: gloo-system
  labels:
    app: my-label
    version: v1
data:
  encryptionKey: "dGhpcyBpcyB0aGUgZW5jcnlwdGlvbiBzZWNyZXQga2U="
 ```

## Deprecation of UserSession in ExtAuthConfig
The UserSession `Session` is now deprecated in the `ExtAuthConfig`, please use the `UserSession` (`UserSessionConfig`). 

The reason being that the two APIs `ExtAuthConfig` and `AuthConfig` were not aligned with the addition of the `CipherConfig` added to the `AuthConfig` and the `ExtAuthConfig`.  The value obtained from the `AuthConfig`'s `CipherConfig` reference should not be visible in the `AuthConfig` API, whereas it should be visible in the `ExtAuthConfig` API.

## Glooctl encryptionkey
Create an encryptionkey secret using `glooctl create secret encryptionkey --name <name> --key <key>`.

# Context

Users will be able to supply a new key reference Encryption Key Secret to apply encryption to the Cookie Sessions.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/ext-auth-service/issues/433

### CODE REVIEW

File: projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto [hunk 1]
```diff
@@ -444,6 +444,21 @@ message UserSession {
...
         RedisSession redis = 4;
     }
+    message CipherConfig {
+        message CipherKey {
+            // to enable the cipher encryption, the key has to be present.
+            oneof key {
+                // The key reference used for the cipher. The reference must be a gloo.solo.io.EncryptionKeySecret custom resource.
+                core.solo.io.ResourceRef key_ref = 1;
+                // This is the raw key value. It is not recommended to set this value.
+                string key_value = 2;
```
sam-heilbron, line 454: Why do we expose this at all? It feels like if we are building an API around secrets, we would just want to use secure patterns of defining a secret separately and referencing it using the resource ref here

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,21 @@ message UserSession {
...
+            }
+        }
+        // This is the value used for the key
+        CipherKey key = 1;
```
sam-heilbron, line 458: Is there more components of the API that we plan to expand this to in the future? Right now we have:
```
message {
  message {
    oneof {
      key_ref = 1
    }
  }
}
```

but it feels like this could be flatter/simpler and jsut be:
```
message {
  key_ref = 1
}
```

projects/gloo/api/v1/secret.proto
```diff
@@ -146,4 +148,13 @@ message HeaderSecret {
 message AccountCredentialsSecret {
     string username = 1;
     string password = 2;
+}
+
+/*
+ Secret used for key encryption.
+ This is used for encrypting Session Values.
+ */
+message EncryptionKeySecret {
```
sam-heilbron, line 157: Do we need to support this new secret types in Edge with a corresponding secret converter: https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/api/converters/kube/secretconverter.go#L16 ?

changelog/v1.14.0-beta20/user-session-encryption.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: nit: I tend to use NON_USER_FACING for open source changes that are API updates for an enterprise feature, and then NEW_FEATURE in the enterprise changelog

changelog/v1.14.0-beta20/user-session-encryption.yaml
```diff
@@ -0,0 +1,5 @@
...
+    issueLink: https://github.com/solo-io/ext-auth-service/issues/433
```
sam-heilbron, line 3: Since this is linked in our changelog docs, users will have access to this link, but it goes to a private repository. It may be worth it to make a Gloo open source issue that outlines the feature so that we can link to it.

changelog/v1.14.0-beta20/user-session-encryption.yaml
```diff
@@ -0,0 +1,5 @@
...
+    description: >-
```
sam-heilbron, line 4: Let's add resolvesIssue=false so this doesn't auto-close

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jackstine: used in Solo-projects to pass the value from the secret to the Solo-projects UserSession.
https://github.com/solo-io/solo-projects/blob/70539faab6b1e246efc11022c82ae0924de1956e/projects/gloo/pkg/syncer/extauth/translate.go#L487

projects/gloo/api/v1/secret.proto
jackstine: What is the secret converter?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jackstine: Yeah let's make it `config.key.key_ref` instead.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,17 @@ message UserSession {
         // Use redis to store the tokens and just store a random id in the cookie.
         RedisSession redis = 4;
     }
+    message CipherConfig {
+        // to enable the cipher encryption, the key has to be present.
+        oneof key {
+            // The key reference used for the cipher. The reference must be a gloo.solo.io.EncryptionKeySecret custom resource.
+            core.solo.io.ResourceRef key_ref = 1;
+            // This is the raw key value. It is not recommended to set this value.
```
elcasteel, line 452: Explain in the comment when it might be set (I assume testing?)

projects/gloo/api/v1/secret.proto
elcasteel: I think it's for cases when the users have pre-existing secrets that they want gloo to use but I could be wrong. (If I'm right then we probably don't need this)

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,17 @@ message UserSession {
         // Use redis to store the tokens and just store a random id in the cookie.
         RedisSession redis = 4;
     }
+    message CipherConfig {
```
elcasteel, line 447: Are we currently only supporting one type of cipher?

projects/gloo/api/v1/secret.proto
sam-heilbron: https://github.com/solo-io/gloo/blob/master/projects/gloo/pkg/bootstrap/utils.go#L195 I would poke around that area to see how it's used. 

https://github.com/solo-io/solo-kit/blob/main/pkg/api/v1/clients/kubesecret/resource_client.go#L31

This is what allows us to read a secret from kubernetes that is untyped (opaque secret) and convert it into a typed secret in Gloo, where we have getters to the appropriate fields

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jackstine: currently, we are only supporting one type of cipher.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,15 @@ message UserSession {
...
+        // to enable the cipher encryption, the key has to be present.
+        oneof key {
```
sam-heilbron, line 449: Do we need a oneof here? It seems like users will always supply the key as a reference to a secret and not inline

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1301,7 +1329,7 @@ message ExtAuthConfig {
         // scopes to request in addition to the openid scope.
         repeated string scopes = 7;
 
-        UserSession session = 8;
+        UserSession session = 8 [deprecated = true];
```
sam-heilbron, line 1332: Thanks for adding the deprecated flag. I think it makes sense as well to add a comment clarifying why it's deprecated and what the preferred field is so that it's easy for future developers to understand. We will also want to add a big comment above the UserSession message calling out that its deprecated so people don't unintentionally add to it

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1515,8 +1546,7 @@ message ExtAuthConfig {
         // Scopes to request for.
         repeated string scopes = 6;
 
-        // Configuration related to the user session.
-        UserSession session = 7;
+        UserSession session = 7 [deprecated = true];
```
sam-heilbron, line 1549: Same as above. I think we should include a comment clarifying what the preferred field is

sam-heilbron (commented):
Looking good!

1. We spoke in a huddle yesterday around the approach of splitting the internal and external api (so they aren't always in sync) and there was some useful information about why we did that and why it's safe change. Can you include that context in this PR so reviewers and future devs can understand those details?

2. We spoke about how a secret converter may be necessary for this new secret type. You were going to test (manually or automated kube2e test) that we can apply a secret with this data and the controller could pick up the secret and send it to the ext-auth-service. Were you able to verify that that works without a secret converter?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jackstine: I would like to have the flexibility to make this more versatile. The reason is that could add more options in the future to supply keys.  If you think we should keep it as a single attribute, I am fine doing that, but I believe this just provides the flexibility to include other features in the future. 

As we build APIs I think we should have the mindset of building them to be as flexible as possible, so that they are backward compatible and work with the intent of change.

projects/gloo/api/v1/secret.proto
jackstine: The SecretConverter is required to enable the finding of the secret in the snapshot.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,15 @@ message UserSession {
         // Use redis to store the tokens and just store a random id in the cookie.
         RedisSession redis = 4;
     }
+    message CipherConfig {
```
elcasteel, line 447: Add a description about what this is used for

elcasteel:
API generally looks good, I requested a couple more comments.

projects/gloo/pkg/api/converters/kube/encryption_key_secret.go
```diff
@@ -0,0 +1,75 @@
...
+type EncryptionSecretConverter struct{}
+
+var _ kubesecret.SecretConverter = &EncryptionSecretConverter{}
+
+func (t *EncryptionSecretConverter) FromKubeSecret(ctx context.Context, rc *kubesecret.ResourceClient, secret *kubev1.Secret) (resources.Resource, error) {
+	if secret == nil {
+		contextutils.LoggerFrom(ctx).Warn("unexpected nil secret")
+		return nil, nil
+	}
+
```
jenshu, line 29: we should also check that `secret.type == EncryptionKeySecretType`

projects/gloo/pkg/api/converters/kube/secretconverter.go
```diff
@@ -23,6 +23,9 @@ var GlooSecretConverterChain = NewSecretConverterChain(
 	// the header converter needs to run last because it has a fall-back to convert any opaque k8s secret with
 	// non-empty data into a gloo header secret
 	new(HeaderSecretConverter),
+	// added the encryption secret primarily for usersession cookie encryption. This enables us to
+	// create and update the key used for encryption.
+	new(EncryptionSecretConverter),
```
jenshu, line 28: nit: since EncryptionSecretConverter is now last in the list, the comment above should be updated to something like `the header converter needs to run after the OpaqueSecretConverter because...`, or could also just move the encryption converter up

projects/gloo/pkg/api/converters/kube/encryption_key_secret.go
jackstine: this is fromKubeSecret -> toEncryptionKeySecret.  the secret should be at most type Opaque. I can add that in.

jackstine: Not sure why, but leaving this out `secret.type == EncryptionKeySecretType` caused it to not work in the kube e2e test, but it was working when I manually created the secret.

elcasteel: should we be setting `secret.type` when we create the secret in the kube e2e test?

jackstine: no, there is `Type` option when creating the `gloov1.Secret` only `Kind`

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1259,6 +1268,26 @@ message ExtAuthConfig {
...
+        // Set-Cookie options
+        UserSession.CookieOptions cookie_options = 2;
+        oneof session {
+            // Set the tokens in the cookie itself. No need for server side state.
+            UserSession.InternalSession cookie = 3;
+            // Use redis to store the tokens and just store a random id in the cookie.
+            UserSession.RedisSession redis = 4;
+        }
+        message CipherConfig {
+            // to enable the cipher encryption, the key has to be present. Note that the key has to be 32 bytes.
```
sam-heilbron, line 1284: Can we clarify the behavior if the key is not 32 bytes? Wil the control plane reject it? Will the data plane error out?

projects/gloo/pkg/api/converters/kube/secretconverter.go
```diff
@@ -20,6 +20,9 @@ var GlooSecretConverterChain = NewSecretConverterChain(
 	new(OAuthSecretConverter),
 	new(AccountCredentialsSecretConverter),
 	new(OpaqueSecretConverter),
+	// added the encryption secret primarily for usersession cookie encryption. This enables us to
+	// create and update the key used for encryption.
```
sam-heilbron, line 24: I know there had been some confusion around what secretconverts are/are used for. It seems we're come to the agreement that for each secret type we support, we need to support the subsequent converter as well. For learnings like this, it helps to document the behavior for others to follow, but I think it's nice when we codify it in a test as well.

https://github.com/solo-io/solo-projects/blob/master/projects/extauth/pkg/config/translator_test.go#L740 is an example where we try to make it harder to add a field and forget to add the corresponding code. Could we try to add something lke this where we compare the number of secret types to the number of secret converters in our chain? Then we can error if they differ and include a message about why they need to be in sync

projects/gloo/api/v1/secret.proto
```diff
@@ -146,4 +148,13 @@ message HeaderSecret {
 message AccountCredentialsSecret {
     string username = 1;
     string password = 2;
+}
+
+/*
+ Secret used for key encryption.
+ This is used for encrypting Session Values.
+ */
```
sam-heilbron, line 156: I think it would be valuable to add glooctl support for this secret (https://github.com/solo-io/gloo/blob/master/projects/gloo/cli/pkg/cmd/create/secret/root.go#L36). Even if we don't actually implement it, it may be valuable to add the scaffolding for it so it's clear we thought about it and intentionally didn't support it

projects/gloo/api/v1/secret.proto
```diff
@@ -43,6 +43,8 @@ message Secret {
         HeaderSecret header = 8;
         // Secrets to represent user/secret pairs. Used to authenticate to LDAP service accounts and hold shared secrets for HMAC auth.
         AccountCredentialsSecret credentials = 9;
+        // Enterprise-only: Secrets used to encrypt messages and data. Used to encrypt and decrypt session values in Ext-Auth.
```
sam-heilbron, line 46: nit: I'm not familiar with the previous reasoning for adding the Enterprise-only tag on  field comments. The secret itself will be parsed by the open source project if it is defined because we read all secrets and use the same secret converter chain in open source that we do for enterprise. We do only consume the secret in enterprise, that is true.

Up to you on how you want to approach this clarification. We alreayd use this for other secret types so it's ok to follow that pattern. I just wanted to call out what "enterprise-only" means in this instance.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,16 @@ message UserSession {
         // Use redis to store the tokens and just store a random id in the cookie.
         RedisSession redis = 4;
     }
+    // the cipher config is used to encrypt session cookie values.  This is currently only available for OIDC.
+    message CipherConfig {
+        // to enable the cipher encryption, the key has to be present.
+        oneof key {
+            // The key reference used for the cipher. The reference must be a gloo.solo.io.EncryptionKeySecret custom resource.
```
sam-heilbron, line 451: What is the behavior of the product if the secret type doesn't match or can't be found? I think it would help to clarify this so users can understand (since the implementation is private to them)

sam-heilbron (commented):
Looking great! A few nits that aren't blocking. Can we also include in the PR outline some piece about how we deprecate the UserSession and introduce a new UserSessionConfig (and why we do that)? I think it will be a useful breadcrumb for people tracking these changes

projects/gloo/pkg/api/converters/kube/secretconverter.go
elcasteel: +1 (although I would be ok with a TODO in this PR to follow up on later)

projects/gloo/api/v1/secret.proto
elcasteel: +1

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
elcasteel: I think we can check the size in the control plane and reject the authconfig. Also this whole structure won't be visible to users so if we want to document the required key size it should be in the public interface.

projects/gloo/pkg/api/converters/kube/secretconverter.go
jackstine: not sure if this is possible, given we need to collect the types from the package itself, and not count the number of fields in a type.  I can't find a reasonable way to perform this.  I have tried using the `go/importer` and `go/types` packages, but unable to get the types from this scope. Also the package that I get resolves with no elements in it, so I think there is something wrong.   When using the `go/importer` I get an error saying `can't find import:`.

```golang
		something, _ := reflect.TypeOf(v1.Secret{}).FieldByName("Kind")
		kindType := something.Type
		info := types.Info{}
		fmt.Println(kindType)
		pathOfKind := "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
		pathOfKind = "."
		v1Package := types.NewPackage(pathOfKind, "")
		checker := types.NewChecker(&types.Config{}, nil, v1Package, &info)
		for _, scope := range checker.Scopes {
			fmt.Println(scope)
		}
		pathOfKind = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
		pkg, err := importer.Default().Import(pathOfKind)
		if err != nil {
			fmt.Printf("error: %s\n", err.Error())
			return
		}
		for _, declName := range pkg.Scope().Names() {
			fmt.Println(declName)
		}
		// ifaceType := v1Package.Scope().Lookup("isSecret_Kind").Type().(*types.Named)
		names := v1Package.Scope().Names()
		for _, obj := range names {
			fmt.Println(obj)
			// if types.Implements(named, ifaceType) {
			// 	count++
			// }
		}
```

jackstine: another way would be to read the secrets.proto file for the names of the secrets and count those?

projects/gloo/api/v1/secret.proto
jackstine: Good callout, It does not have any use unless it is in enterprise.  I am fine with leaving this comment as is.  It does mention Ext-Auth, which is enterprise only.

changelog/v1.15.0-beta2/user-session-encryption.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
```
jackstine, line 2: I do not think this is still `NON_USER_FACING` with the changes with glooctl and the kube secret converters.

elcasteel (dismissed):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -444,6 +444,17 @@ message UserSession {
         // Use redis to store the tokens and just store a random id in the cookie.
         RedisSession redis = 4;
     }
+    // the cipher config is used to encrypt session cookie values.  This is currently only available for OIDC.
+    message CipherConfig {
+        // to enable the cipher encryption, the key has to be present. Note that the key has to be found and 32 bytes in 
+        // length for the authconfig to not be rejected.
+        oneof key {
+            // The key reference used for the cipher. The reference must be a gloo.solo.io.EncryptionKeySecret custom resource.
```
jenshu, line 452:
```suggestion
            // The key reference used for the cipher. The reference must be a Kubernetes Secret of type `gloo.solo.io.EncryptionKeySecret`.
```

changelog/v1.15.0-beta2/user-session-encryption.yaml
jenshu: can you include the glooctl change as another entry (NEW_FEATURE)?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jenshu: fwiw i think it would be safe to remove the oneof for now, and if we want to add more options later, it should be a [backwards-compatible](https://protobuf.dev/programming-guides/proto3/#tag-reuse-issues) change to move it into a new oneof

but i am ok either way, it shouldn't affect the yaml

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1259,6 +1269,27 @@ message ExtAuthConfig {
...
+        oneof session {
+            // Set the tokens in the cookie itself. No need for server side state.
+            UserSession.InternalSession cookie = 3;
+            // Use redis to store the tokens and just store a random id in the cookie.
+            UserSession.RedisSession redis = 4;
+        }
+        message CipherConfig {
+            // to enable the cipher encryption, the key has to be present.
+            // Note that the key has to be found and 32 bytes in length for the authconfig to not be rejected.
+            string key_value = 1 [(extproto.sensitive) = true];
```
jenshu, line 1287: naming: is this both a key and value? if it's just the key, should we rename it `key`?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1259,6 +1269,27 @@ message ExtAuthConfig {
 
     }
 
+    message UserSessionConfig {
+        // should we fail auth flow when failing to get a session from redis, or allow it to continue,
+        // potentially starting a new auth flow and setting a new session.
```
jenshu, line 1274: i think to be more clear, we should say "If true, fail the auth flow ... If false, ..."

jackstine: same description as https://github.com/solo-io/gloo/blob/master/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto#L403

projects/gloo/pkg/api/converters/kube/secretconverter.go
sam-heilbron: yeah seems like good exploration, and given that it's not a requirement, let's not leave this as a blocker. Perhaps we track this in some follow-up work to explore and flush out?

elcasteel (approved):

PR #8017, opened 2023-03-27
build: place kube dumps into saved artifacts

gunnar-solo:
# Description
In _my_ opinion, our current failure dumps are very verbose, and make it hard to read the originating failure.  While the context of a full Kube cluster dump is _sometimes_ helpful, it's better served by segmenting it into individual files.  This PR accomplishes this

[working example](https://github.com/solo-io/gloo/actions/runs/4660905817?pr=8017)

⚠️ a follow-on PR will need to be done for Solo-Projects, since this dump method is consumed in both repositories ⚠️ 

# Checklist:


test/helpers/kube_dump.go
```diff
@@ -47,58 +64,86 @@ func KubeDumpOnFail(out io.Writer, namespaces ...string) func() {
 		if err != nil {
 			fmt.Fprintf(out, "getting kube dump failed: %v", err)
 		}
-		fmt.Fprintf(out, dump)
+
+		f, err := os.OpenFile(mustOutDir()+"/kube-dump.log", os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
```
sam-heilbron, line 68: One option, would be that we have different types of Writers and that is configurable. You can either have a Stdout Writer or FileWriter.

sam-heilbron (commented):
Nice! Hopefully this makes debugging easier. We don't actually have to wait for a flake, we can just add a failure to a test so it fails every time.

test/helpers/kube_dump.go
```diff
@@ -4,13 +4,30 @@ import (
...
+	"os"
 	"os/exec"
 
 	"github.com/solo-io/gloo/pkg/cliutil/install"
 
 	errors "github.com/rotisserie/eris"
 )
 
+func mustOutDir() string {
+	out := os.Getenv("TEST_FAILURE_DUMP_DIR")
```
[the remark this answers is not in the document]
gunnar-solo: it's not so much that I wanted this to be configurable, so much as I didn't want to hardcode it twice

(in the `upload-artifact` arg _and_ in the dump script)

gunnar-solo:
/kick https://console.cloud.google.com/cloud-build/builds;region=global/277a2e8d-4b02-4bd8-a45e-a885c29dcd53;step=1?project=solo-public

.github/actions/regression-tests/action.yaml
```diff
@@ -87,3 +80,9 @@ runs:
       name: ${{matrix.kube-e2e-test-type}}@k8s${{matrix.kube-version.kubectl}}
       path: test-out.json
       if-no-files-found: warn
+  - uses: actions/upload-artifact@v3
```
sam-heilbron, line 83: Currently we have a couple of steps used to save failure results:
- Save the results to test-out.json
- Upload test-out.json
- Upload the test-failure-dump

Could we somehow consolidate these actions so it's abundantly clear what is being uploaded and to where?

ci/deploy-to-kind-cluster.sh
```diff
@@ -127,7 +127,7 @@ if [[ -z $SKIP_DOCKER || $SKIP_DOCKER == 'false' ]]; then
     # if using arm64, push to the docker registry container, instead of kind
     VERSION=$VERSION CREATE_TEST_ASSETS="true" TEST_ASSET_ID="docker-reg" IMAGE_REPO="localhost:$REGISTRY_PORT" make docker-push-local-arm -B
   else
-    VERSION=$VERSION CLUSTER_NAME=$CLUSTER_NAME make push-kind-images
+    VERSION=$VERSION CLUSTER_NAME=$CLUSTER_NAME make -s push-kind-images
```
sam-heilbron, line 130: What does `-s` provide? If we add it here, should we be adding it to other commands in this file as well?

test/helpers/kube_dump.go
sam-heilbron: Followed up offline, agreement was to not support multiple writers as it feels to be overkill.

test/helpers/kube_dump.go
```diff
@@ -47,58 +60,86 @@ func KubeDumpOnFail(out io.Writer, namespaces ...string) func() {
...
+			fmt.Println("unable to get kube dump")
+			return
+		}
+		defer f.Close()
+		f.WriteString(dump)
 	}
 }
 
 func PrintKubeState() {
+	f, err := os.OpenFile(mustOutDir()+"/kube-state.log", os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
```
sam-heilbron, line 75: This function, `PrintKubeState`, is not entirely accurate anymore. Now it seems to be that we are `AppendKubeState` to some output stream. What if we pass in the file we're writing to, and then this function appends the kube details to that file?

test/helpers/kube_dump.go
```diff
@@ -47,58 +60,86 @@ func KubeDumpOnFail(out io.Writer, namespaces ...string) func() {
...
-	fmt.Println(string(kubeEndpointsState))
-	fmt.Println("*** End Kube state ***")
+	f.WriteString("*** Kube state ***\n")
+	f.WriteString(string(kubeState) + "\n")
+	f.WriteString(string(kubeEndpointsState) + "\n")
+	f.WriteString("*** End Kube state ***\n")
 }
 
 func PrintDockerState() {
+	f, err := os.OpenFile(mustOutDir()+"/docker-state.log", os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
```
sam-heilbron, line 100: Similar to the KubeState function. I think this function should do one thing, and that take debug output about the docker state and append it to an existing (or new file). 

i think it would be powerful if we had something like:

```
func WriteDebugStateToFiles() {
  debugDir := [the directory]
  dockerStateFile := [the docker state file]
  WriteDockerStateToFile(dockerStateFile)
  
  kubeStateFile := [the kube state file]
  WriteKubeStateToFile(kubeStateFile)

}
```

test/helpers/kube_dump.go
```diff
@@ -4,13 +4,26 @@ import (
...
+	"os"
 	"os/exec"
 
 	"github.com/solo-io/gloo/pkg/cliutil/install"
 
 	errors "github.com/rotisserie/eris"
 )
 
+func mustOutDir() string {
+	out := "./test-failure-dump"
```
sam-heilbron, line 16: I like the fact that this is hardcoded, and I think we can clarify this a little more. I think we can make a constant and a comment clearly outlining what this folder is, what it should contain. 

When we run this locally where is this file? Should it be placed in the `_output` directory or some other directory that is gitignored? 
When subsequent runs occur locally, how will I distinguish which run produced which output? Will we have separate directories for runs, or each run will overwrite the previous?

ci/deploy-to-kind-cluster.sh
gunnar-solo: this made me sad

I was hoping it would suppress the ~500 lines of docker build output by having a --silent make command, but it did not

I’ll remove it, as it’s not useful

test/helpers/kube_dump.go
nfuden: how is this affected by runs from various initiating locations. IE run from top level or from a lower folder.

test/helpers/kube_dump.go
```diff
@@ -47,58 +60,86 @@ func KubeDumpOnFail(out io.Writer, namespaces ...string) func() {
...
-	fmt.Println("*** Docker state ***")
-	fmt.Println(dockerState.String())
-	fmt.Println("*** End Docker state ***")
+	f.WriteString("*** Docker state ***\n")
+	f.WriteString(dockerState.String() + "\n")
+	f.WriteString("*** End Docker state ***\n")
 }
 
 func PrintProcessState() {
+	f, err := os.OpenFile(mustOutDir()+"/process-state.log", os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
```
nfuden, line 124: Do we want to create kube dump files on local? Also whats our cleanup strategy

gunnar-solo: the strategy is not "build to _output, and overwrite old dump files"

test/helpers/kube_dump.go
gunnar-solo: current iterations of this always write to `_output` and overwrite the previous set of logs

test/helpers/kube_dump.go
```diff
@@ -4,101 +4,224 @@ import (
...
+			return err
+		}
+		if len(crs) == 0 {
+			continue
+		}
+
+		// we record each one in its own .yaml representation
+		for _, cr := range crs {
+			crdDir := filepath.Join(namespaceDir, crd)
+			if err := os.MkdirAll(crdDir, os.ModePerm); err != nil {
```
gunnar-solo, line 155: nit: this directory creation would be more efficient a level up

test/kube2e/upgrade/upgrade_suite_test.go
```diff
@@ -47,6 +47,8 @@ var _ = BeforeSuite(func() {
 	testHelper, err := kube2e.GetTestHelper(suiteCtx, namespace)
 	Expect(err).NotTo(HaveOccurred())
 
+	skhelpers.RegisterPreFailHandler(helpers.KubeDumpOnFail(GinkgoWriter, "upgrade", testHelper.InstallNamespace, "other-ns"))
```
gunnar-solo, line 50: for some reason, this test suite didn't _have_ kube dumps.  I've elected to add them, just because I noticed they were missing

🤷‍♂️

sam-heilbron, line 50: Why do we have other-ns as well? Is that used in the test?

.github/actions/regression-tests/action.yaml
```diff
@@ -63,21 +63,14 @@ runs:
...
-    run: |
-      # see what's in the cluster if we failed
-      kubectl get all -A
-      kubectl get configmaps -A
   - name: save results
     if: ${{ failure() }}
     shell: bash
-    run: echo '{"url":"https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}",
-            "name":"${{matrix.kube-e2e-test-type}}@k8s${{matrix.kube-version.kubectl}}"}' > test-out.json
+    run: mkdir -p _output/test-failure-dump && echo '{"url":"https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}",
```
sam-heilbron, line 69: Could we support the auto-generation of this folder in the code? That way it behaves the same locally and in CI

test/helpers/kube_dump.go
```diff
@@ -4,101 +4,223 @@ import (
...
+		if split := strings.Split(line, " "); len(split) > 1 {
+			toReturn = append(toReturn, split[0])
+		}
+	}
+	return toReturn, nil
+}
+
+// setupOutDir forcibly deletes/creates the output directory, then return the path to it
+func setupOutDir() string {
+	outDir := filepath.Join(util.GetModuleRoot(), "_output", "test-failure-dump")
```
sam-heilbron, line 204: could we make thiis directory "test-failure-dump" a constant at the top of the file? Also could we rename this something like `kube2e-artifacts` or something to clarify what is in there?

.github/actions/regression-tests/action.yaml
gunnar-solo: the folder _is_ autogenerated in-code _so long as the failure handler runs_.  There's a non-zero chance the failure handler doesn't run (ex:  a go panic, which halts all compute.  See:  what the upgrade suite is doing right now).

There are 2 sets of things that need the `_output/test-failure-dump` folder:
1. `test-out.json` (must always appear to notify end-user)
2. the log dumps from this PR (don't appear if go compute halts)

sam-heilbron (approved):

test/helpers/kube_dump.go
```diff
@@ -4,101 +4,221 @@ import (
...
+	outDir  = filepath.Join(util.GetModuleRoot(), "_output", "kube2e-artifacts")
 )
 
-var dumpCommands = func(namespace string) []string {
-	return []string{
-		fmt.Sprintf("echo PODS FROM %s: && kubectl get pod -n %s --no-headers -o custom-columns=:metadata.name", namespace, namespace),
-		fmt.Sprintf("for i in $(kubectl get pod -n %s --no-headers -o custom-columns=:metadata.name); do echo STATUS FOR %s.$i: $(kubectl get pod -n %s $i -o go-template=\"{{range .status.containerStatuses}}{{.state}}{{end}}\"); done", namespace, namespace, namespace),
-		fmt.Sprintf("for i in $(kubectl get pod -n %s --no-headers -o custom-columns=:metadata.name); do echo LOGS FROM %s.$i: $(kubectl logs -n %s $i --all-containers); done", namespace, namespace, namespace),
-		fmt.Sprintf("kubectl get events -n %s", namespace),
+func KubeDumpOnFail(out io.Writer, namespaces ...string) func() {
```
nfuden, line 21: nit: exported function without comment

test/helpers/kube_dump.go
```diff
@@ -4,101 +4,221 @@ import (
...
 
-func PrintDockerState() {
-	dockerCmd := exec.Command("docker", "ps")
+func recordKubeDump(namespaces ...string) {
+	// for each namespace, create a namespace directory that contains...
+	for _, ns := range namespaces {
+		// ...a pod logs subdirectoy
+		if err := recordPods(filepath.Join(outDir, ns, "_pods"), ns); err != nil {
+			fmt.Printf("error recording pod logs: %f, \n", err)
+			break
```
nfuden, line 96: Should this be a break? What if there exists the folder but not one for the crs?

test/helpers/kube_dump.go
```diff
@@ -4,101 +4,221 @@ import (
...
+	"path/filepath"
+	"strings"
 
 	"github.com/solo-io/gloo/pkg/cliutil/install"
+	"github.com/solo-io/skv2/codegen/util"
+)
 
-	errors "github.com/rotisserie/eris"
+var (
+	kubeCli = &install.CmdKubectl{}
```
nfuden, line 17: Is it safe to be reusing this kubectl cmd across all users?

test/helpers/kube_dump.go
gunnar-solo: fair point--there's no real reason not to let the CRs _try_ to be scraped if the pod logs fail

sam-heilbron (approved):

nfuden (approved):

PR #8021, opened 2023-03-28
Stability: Use TestContext Framework With Vault Tests

sam-heilbron:
# Description

Update Vault tests to use [TestContext](https://github.com/solo-io/gloo/blob/master/test/e2e/test_context.go) framework

# Context

We have some [upcoming efforts in the Vault space](https://github.com/solo-io/solo-projects/issues/4636). We will need to write thorough e2e tests, so updating these tests to use our framework will make our lives easier in the future.

## Enable Vault AWS Tests
These tests were previously skipped in our CI pipeline.

I confirmed in [slack](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1680028033362169) that we could enable these.

I opened https://github.com/solo-io/gloo/pull/8023 to validate that these tests were passing as currently written, which they were. 

## VaultFactory and VaultInstance
This is the main utility code to start and interact with a running instance of vault. I tried to clean these up using the following strategies:
- Configured which was not being used was deleted
- Tls configuration is not used, but I confirmed with @jbohanon  that we may want it soon so I added a comment to clarify the behavior and explicitly set it to false
- The Vault running would use sleep as a mechanism to wait for vault to be running. I changed this to a poll mechanism which, while not ideal, matches our Envoy service implementation, and works (plus its faster than sleeping for a few seconds since we poll every 100 milliseconds)
- The VaultInstance API doesn't contain hard-coded values. Instead, anything that a test needs to modify is injected as a function parameter. For example, now we have a `WriteSecret` method which takes an arbitrary secret, instead of a hardcoded one

## TestContextWithVault
We have an existing TestContext, the framework used to consolidate all the complexity of running Gloo controllers and Envoy in our e2e tests. These Vault tests use that framework, but also require a running instance of Vault. Instead of adding complexity to this framework, I built an extension to it, and ensured our TestContextFactory can create either a `TestContext` or `TestContextWithVault`.

# Checklist:


test/e2e/test_context.go
```diff
@@ -273,3 +285,28 @@ func (c *TestContext) GetHttpsRequestBuilder() *testutils.HttpRequestBuilder {
...
+// VaultInstance returns the wrapper for the running instance of Vault that this test is using
+func (v *TestContextWithVault) VaultInstance() *services.VaultInstance {
+	return v.vaultInstance
+}
+
+// RunVault starts running the VaultInstance and blocks until it has successfully started
+func (v *TestContextWithVault) RunVault() {
+	ginkgo.By("TestContextWithVault: Running Vault")
+
+	// The VaultInstance will be cleaned up when the provided context is cancelled
```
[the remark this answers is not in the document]
sam-heilbron: It's guaranteed  at the package level that it's tied to the lifecycle of whatever context it's run with. The hope here is that this removes another cleanup step that devs have to worry about. Now they just have to start it, and the TestContext will clean it up for them

test/e2e/vault_aws_test.go
```diff
@@ -0,0 +1,150 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	gloov1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+)
+
+const (
+	// These tests run using the following AWS ARN for the Vault Role
+	// If you want to run these tests locally, ensure that your local AWS credentials match,
+	// or adjust the configured role
```
ianmacclancy, line 20: Nit - lets say "or use another role" we dont have protective permissions around the test role and people should not be modifying it in most cases.

test/e2e/vault_aws_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html
+	vaultAwsRole   = "arn:aws:iam::802411188784:user/gloo-edge-e2e-user"
```
ianmacclancy, line 22: This could be moved higher up as a shared "awsTestRole" - its used in other aws tests as well. Not needed in this pr. Just thinking about possible shared code. If the role is deleted again it would be nice to only have to update it in one location.

test/e2e/vault_aws_test.go
sam-heilbron: goood call! Yeah that's what I meant, but I can see now how "adjust configured role" could be interpreted differently

ianmacclancy (approved):
Small comments - nothing blocking. Feel free to resolved after responding

jbohanon (dismissed):

test/e2e/vault_aws_test.go
ianmacclancy: Follow-up works, I had seen arns used before but looks like gloo does not have any other locations with it... for now. I still think with the move to shared values we could have it in a file of shared constants or move it to the test context setup.

jbohanon (approved):

ianmacclancy (approved):

PR #8049, opened 2023-04-03
4169 tagging metrics

sheidkamp:
# Description

# Context

# Checklist:


changelog/v1.14.0-beta20/tagging-metrics.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/4169
+    resolvesIssue: false
+    description: |
+      The `validation_gateway_solo_io_valid_config` does not provide a useful measure when there are multiple prxoies.
```
sam-heilbron, line 6:
```suggestion
      The `validation_gateway_solo_io_valid_config` does not provide a useful measure when there are multiple proxies.
```

projects/gateway/pkg/validation/validator.go
```diff
@@ -236,7 +245,7 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
 	}
 	ref := opts.Resource.GetMetadata().Ref()
 	ctx = contextutils.WithLogger(ctx, "gateway-validator")
-
+	contextutils.LoggerFrom(ctx).Info("in validateSnapshot")
```
sam-heilbron, line 248: Let's remove this, as I figure it was used for debugging and not necessary. Though correct me if I'm wrong

projects/gateway/pkg/validation/validator.go
```diff
@@ -454,14 +463,16 @@ func (v *validator) copySnapshotNonThreadSafe(ctx context.Context, dryRun bool)
 	if v.latestSnapshot == nil {
 		return nil, HasNotReceivedFirstSync
 	}
+
+	snapshotClone := v.latestSnapshot.Clone()
```
sam-heilbron, line 467: Is there a reason we move the snapshotClone earlier here? I figured that we only want to clone if the error was nil as a minor optimization

Makefile
```diff
@@ -790,6 +790,27 @@ endif
 
 CLUSTER_NAME ?= kind
 
+#----------------------------------------------------------------------------------
+# Kind Load Images
+#----------------------------------------------------------------------------------
+
+CLUSTER_NAME ?= kind
+
+kind-load-%:
```
sam-heilbron, line 799: 👏 🎉

sam-heilbron (commented):
I like it! Could you add a test to demonstrate the use case of multiple proxies? I'm concerned that as we currently have it, all proxy metrics will still move together, but would happy to be proven otherwise with a test

projects/gateway/pkg/validation/validator.go
sheidkamp: I was using it to get the gateways, but it looks like I can do that with 'latestSnapshot'

Makefile
```diff
@@ -790,6 +790,27 @@ endif
...
+	kind load docker-image $(IMAGE_REPO)/$*:$(VERSION) --name $(CLUSTER_NAME)
+
+kind-load-%-debug:
+	kind load docker-image $(IMAGE_REPO)/$*:$(VERSION)-debug --name $(CLUSTER_NAME)
+
+kind-load-%-extended:
+	kind load docker-image $(IMAGE_REPO)/$*:$(VERSION)-extended --name $(CLUSTER_NAME)
```
jackstine, line 806: do we even use the extended any more?  Should we raise the fact that we want to delete these from the CI workflow as well, just like we did in enterprise?
[docker extended images are generated here](https://github.com/solo-io/gloo/tree/master/ci/extended-docker)
[here is the cloudbuild job during CI](https://github.com/solo-io/gloo/blob/OIDC-cookie-encryption/ci/cloudbuild/publish-artifacts.yaml#L47)

Makefile
```diff
@@ -790,6 +790,27 @@ endif
...
+
+
+# Build and push an updated image into kind
+kind-reload-%: %-docker kind-load-%
+	kubectl rollout restart deployment/$* -n $(INSTALL_NAMESPACE)
+
+
```
jackstine, line 813:
```suggestion
# Build and push an updated image into kind
kind-reload-%: %-docker kind-load-% ## reload an image in a kubernetes cluster, flags are (IMAGE_REPO, VERSION, INSTALL_NAMESPACE , CLUSTER_NAME)
	kubectl rollout restart deployment/$* -n $(INSTALL_NAMESPACE)


```

jackstine: this way it will show up in `make help`

projects/gateway/pkg/validation/validator.go
```diff
@@ -454,13 +463,14 @@ func (v *validator) copySnapshotNonThreadSafe(ctx context.Context, dryRun bool)
 	if v.latestSnapshot == nil {
 		return nil, HasNotReceivedFirstSync
 	}
+
 	if v.latestSnapshotErr != nil {
-		if !dryRun {
-			utils2.MeasureZero(ctx, mValidConfig)
-		}
+		gatewaysByProxy := utils.GatewaysByProxyName(v.latestSnapshot.Gateways)
+		measureInvalidConfig(ctx, gatewaysByProxy)
```
jackstine, line 469: nice catch with the `!dryRun`.

Makefile
ianmacclancy: I agree - We can remove these.

Makefile
```diff
@@ -780,16 +762,27 @@ docker-push-local: $(DOCKER_IMAGES)
 	docker push $(IMAGE_REPO)/sds:$(VERSION) && \
 	docker push $(IMAGE_REPO)/access-logger:$(VERSION)
 
-# To mimic the effects of CI, CREATE_ASSETS, TAGGED_VERSION and CREATE_TEST_ASSETS need to be set
-# Extended images are the same as regular images but with curl
-.PHONY: docker-push-extended
```
sam-heilbron, line 785: This isn't safe to remove, because it's referenced by: https://github.com/solo-io/gloo/blob/master/ci/cloudbuild/publish-artifacts.yaml#L50

I see that you're trying to remove the extended images, which seems valuable, but we may want to split off into a separate PR. As part of my work to simplify our build pipeline I tried to address this in https://github.com/solo-io/gloo/pull/8060. Want to take a peek at that and let me know what you think?

I think we can remove the extended work from this PR, and even if my PR doesn't go through we can approach this work in a separate PR. How does that sound?

Makefile
```diff
@@ -780,16 +762,27 @@ docker-push-local: $(DOCKER_IMAGES)
...
-	ci/extended-docker/extended-docker.sh
-endif
 
 CLUSTER_NAME ?= kind
 
+#----------------------------------------------------------------------------------
+# Kind Load Images
+#----------------------------------------------------------------------------------
+
+CLUSTER_NAME ?= kind
```
sam-heilbron, line 772: CLUSTER_NAME is defined a few lines above so I don't think we need to redefine it here

Makefile
```diff
@@ -780,16 +762,27 @@ docker-push-local: $(DOCKER_IMAGES)
...
+kind-load-%:
+	kind load docker-image $(IMAGE_REPO)/$*:$(VERSION) --name $(CLUSTER_NAME)
+
+kind-load-%-debug:
+	kind load docker-image $(IMAGE_REPO)/$*:$(VERSION)-debug --name $(CLUSTER_NAME)
+
+
+# Build and push an updated image into kind
+kind-reload-%: %-docker kind-load-% ## reload an image in a kubernetes cluster, flags are (IMAGE_REPO, VERSION, INSTALL_NAMESPACE , CLUSTER_NAME)
+	kubectl rollout restart deployment/$* -n $(INSTALL_NAMESPACE)
```
sam-heilbron, line 783: Do we default INSTALL_NAMESPACE in our Makefile? If not, it might make sense to define it just below cluster name: `INSTALL_NAMESPACE ?= gloo-system`. That way it can still be overridden if a user defines a value, but it will inherit the default of gloo-system (which is used in all of our tests)

projects/gateway/pkg/defaults/gateway.go
```diff
@@ -17,14 +17,14 @@ const (
 	ConfigDumpServiceSuffix = "-config-dump-service"
 )
 
-func DefaultGateway(writeNamespace string) *v1.Gateway {
+func NamedGateway(writeNamespace string, proxyName string) *v1.Gateway {
```
sam-heilbron, line 20: I like it! Big fan of having generic constructors, and passing constants in as arguments

projects/gateway/pkg/validation/validator.go
```diff
@@ -298,21 +307,35 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
 			err = errors.Wrapf(err, failedResourceReports)
 			errs = multierr.Append(errs, err)
+			if !opts.DryRun {
+				measureInvalidConfigProxy(ctx, proxyName)
+			}
 			continue
+		} else {
+			if !opts.DryRun {
+				measureValidConfigProxy(ctx, proxyName)
+			}
```
sam-heilbron, line 336: We have some logic a little lower: 

```
if errs != nil {
		contextutils.LoggerFrom(ctx).Debugf("Rejected %T %v: %v", opts.Resource, ref, errs)
		if !opts.DryRun {
			utils2.MeasureZero(ctx, mValidConfig)
		}
```

Can you help me understand how the changes you're making here are different than the case defined below?

projects/gateway/pkg/validation/validator.go
```diff
@@ -454,13 +476,18 @@ func (v *validator) copySnapshotNonThreadSafe(ctx context.Context, dryRun bool)
...
 	}
+
 	if v.latestSnapshotErr != nil {
-		if !dryRun {
-			utils2.MeasureZero(ctx, mValidConfig)
+		gatewaysByProxy := utils.GatewaysByProxyName(v.latestSnapshot.Gateways)
+		// Since it is an invalid snapshot, and not just an invalid resource, the failure applies to all associated proxies (?)
+		utils2.MeasureZero(ctx, mValidConfig)
+		for proxyName := range gatewaysByProxy {
+			utils2.MeasureOne(ctx, mProxyValidConfig, tag.Insert(syncerstats.ProxyNameKey, proxyName))
```
sam-heilbron, line 485: This metric used to be guarded by the 'dryRun' option, but it no longer is. Is this an intentional change, and if so, can you help me understand why?

I see that we measure 0 for the old metric, and 1 for the new metric. I would have imagined that we meant to measure 0 for both. Am I missing something?

projects/gateway/pkg/validation/validator.go
```diff
@@ -519,3 +546,36 @@ func UnmarshalResource(kubeJson []byte, resource resources.Resource) error {
 	}
 	return nil
 }
+
+func measureInvalidConfigInit(ctx context.Context, gatewaysByProxy map[string]v1.GatewayList) {
```
sam-heilbron, line 550: It seems like all of our metric logic is guarded by the 'dryRun' behavior of the webhook (as in, if we are a dryRun we never emit a metric). I generally don't like passing in a boolean to a function, to then perform a conditional on it, but how do you feel about passing in the dryRun flag, and we emit a metric if it's false and log something if it's true? 

I noticed that it's easy to forget to guard metrics by that flag, so my hope was that if we pull it into this function, we prevent devs from making that mistake. Would love to hear your thoughts on this, and Im more than open to just keeping the current pattern

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -598,6 +594,26 @@ var _ = Describe("Validator", func() {
...
+				v.glooValidator = ValidateFailToggle
+
+				snap := samples.TwoProxyGlooSnapshot(ns)
+				err := v.Sync(context.TODO(), snap)
+				Expect(err).NotTo(HaveOccurred())
+
+				_, err = v.ValidateModifiedGvk(context.TODO(), v1.VirtualServiceGVK, snap.VirtualServices[0], false)
+				Expect(err).To(HaveOccurred())
+
+				data, err := helpers.ReadMetricByLabel("validation.gateway.solo.io/proxy_valid_config", syncerstats.ProxyNameKey.Name(), "gateway-proxy")
```
sam-heilbron, line 609: You had added the nice utility `validateMValidConfig(..)`, is there a reason we can't use that in this test? I imagine it's because that utility asserts taht both the old and new metric match a value, and this test is validating that they differ

test/samples/sample_gloo_api_snapshot.go
```diff
@@ -239,6 +239,183 @@ func SimpleGlooSnapshot(namespace string) *v1snap.ApiSnapshot {
 	}
 }
 
+func TwoProxyGlooSnapshot(namespace string) *v1snap.ApiSnapshot {
```
sam-heilbron, line 242: I really like this! 

We've been trying to ensure that exported functions have a little comment outlining the behavior, can you add one here?

There's some data in here that isn't super necessary for the purposes of your test (hybrid listener, aggregate listener, tcp listener). It totally makes sense that since you're adding a generic utility we want to keep it, but we can also remove those if you feel they enlarge the object without much value. Totally up to you, just wanted to point it out

sam-heilbron (commented):
This is great! The tests around this space are much easier to read now and it will be nice to have some validation around the multi-proxy case. A few small questions/nits

changelog/v1.14.0-beta20/tagging-metrics.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
```
sam-heilbron, line 1: This should be moved to the 1.15.0-beta1 directory

Makefile
sheidkamp: I'll put it back.

Makefile
sheidkamp: https://github.com/solo-io/gloo/blob/fad6be97325bf07ed497ea4ad40e73d0e172be9a/Makefile#L654

projects/gateway/pkg/validation/validator.go
sheidkamp: the changes lower are for the aggregate metric, and are outside of the `for proxyName, gatewayList := range gatewaysByProxy {` loop.

The changes in lines 329-336 are inside the per proxy loop and tracking at the per-proxy level. Going to leave this one open, let me know if you want to discuss.

projects/gateway/pkg/validation/validator.go
sheidkamp: Yup, this one is just an oversight.

projects/gateway/pkg/validation/validator.go
sheidkamp: That was something I thought about too... Its a lesser of 2 evils situation. I think I am going to change it to a function that passes the dryrun flag, because it does result in cleaner looking code, and it will also prevent people (like me, see above) from forgetting to gate it.

Makefile
sheidkamp: Removed

projects/gateway/pkg/validation/validator_test.go
sheidkamp: You exactly landed on the reason we can't use it. Didn't seem worth making the utility more complicated for one case.

test/samples/sample_gloo_api_snapshot.go
sheidkamp: Added a comment, and cleaned up by taking the SimpleSnapshot and adding the extra gateway and proxy. Clean code, plus a consistent snapshot/code reuse.

projects/gateway/pkg/validation/validator.go
sheidkamp: its back in... I think it needs to be.

PR #8054, opened 2023-04-03
fix go mod path for push solo-apis action

jenshu:
we do the gloo checkout at path `gloo` so the go.mod should be at `gloo/go.mod`

this should fix the push solo-apis [failure](https://github.com/solo-io/gloo/actions/runs/4599755409) for next time

already fixed the latest push manually so don't need to do another release

### DIFF
--- a/.github/workflows/push-solo-apis-branch.yaml
+++ b/.github/workflows/push-solo-apis-branch.yaml
@@ -85,7 +85,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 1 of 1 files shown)


nfuden:
I think this breaks the gloo-mesh use case

sam-heilbron:
> I think this breaks the gloo-mesh use case

What do you mean? And if so, should we not merge this?

nfuden:
Nah resolved confusion in slack

ashishb-solo (approved):

PR #8061, opened 2023-04-05
Stability: Improve and Clarify How Envoy Test Runner Chooses Image Tag

sam-heilbron:
# Description

Allow the e2e tests to not require specifying an ENVOY_IMAGE_TAG

# Context

## E2E Test Improvement
E2E tests run locally by starting a gloo-envoy-wrapper docker container, and executing requests against it. Developers supply a `ENVOY_IMAGE_TAG` variable which defines the tag of the image to run. This allows developers to specify a previously released tag, if there are no changes to the data plane, or a locally built tag if there are.

Most times we shouldn't need to build a local version of the image, and instead can just inherit the latest released version.

This change updates the behavior such that if you do not supply a tag, we now choose the latest tag for you, instead of failing.

Also, this fixes a bug where we incorrectly consolidated code for running `envoy-gloo` and `gloo-envoy-wrapper`. The images are separate and so are the tags.

## Generated Kube Types Test
During development, I hit a test flake: 

```
Step #1 - "run-tests": Step #7 - "run-tests": [0mGenerated Kube Code [38;5;9m[1m[It] can read and write a gloo resource as a typed kube object[0m
Step #1 - "run-tests": Step #7 - "run-tests": [38;5;243m/workspace/gloo/projects/gateway/pkg/api/v1/kube/generated_kube_types_test.go:94[0m
Step #1 - "run-tests": Step #7 - "run-tests":   [38;5;9m[FAILED] Unexpected error:
Step #1 - "run-tests": Step #7 - "run-tests":       <*errors.StatusError | 0xc000e5dd60>: {
Step #1 - "run-tests": Step #7 - "run-tests":           ErrStatus: {
Step #1 - "run-tests": Step #7 - "run-tests":               TypeMeta: {Kind: "", APIVersion: ""},
Step #1 - "run-tests": Step #7 - "run-tests":               ListMeta: {
Step #1 - "run-tests": Step #7 - "run-tests":                   SelfLink: "",
Step #1 - "run-tests": Step #7 - "run-tests":                   ResourceVersion: "",
Step #1 - "run-tests": Step #7 - "run-tests":                   Continue: "",
Step #1 - "run-tests": Step #7 - "run-tests":                   RemainingItemCount: nil,
Step #1 - "run-tests": Step #7 - "run-tests":               },
Step #1 - "run-tests": Step #7 - "run-tests":               Status: "Failure",
Step #1 - "run-tests": Step #7 - "run-tests":               Message: "create not allowed while custom resource definition is terminating",
Step #1 - "run-tests": Step #7 - "run-tests":               Reason: "MethodNotAllowed",
Step #1 - "run-tests": Step #7 - "run-tests":               Details: {
Step #1 - "run-tests": Step #7 - "run-tests":                   Name: "",
Step #1 - "run-tests": Step #7 - "run-tests":                   Group: "gloo.solo.io",
Step #1 - "run-tests": Step #7 - "run-tests":                   Kind: "upstreams",
Step #1 - "run-tests": Step #7 - "run-tests":                   UID: "",
Step #1 - "run-tests": Step #7 - "run-tests":                   Causes: nil,
Step #1 - "run-tests": Step #7 - "run-tests":                   RetryAfterSeconds: 0,
Step #1 - "run-tests": Step #7 - "run-tests":               },
Step #1 - "run-tests": Step #7 - "run-tests":               Code: 405,
Step #1 - "run-tests": Step #7 - "run-tests":           },
Step #1 - "run-tests": Step #7 - "run-tests":       }
Step #1 - "run-tests": Step #7 - "run-tests":       create not allowed while custom resource definition is terminating
```

I updated this test to try to resolve this behavior. You can see the approach here: https://github.com/solo-io/gloo/pull/8061/commits/84a9d566ad9dfa9b4ca3e2882b5d49587b5e17b7

# Testing
## Run Tests With Local Image
Build a local image, using the `0.0.0-sh` tag:
```
ENVOY_IMAGE_TAG=0.0.0-sh TEST_PKG=./test/e2e INVALID_TEST_REQS=skip make test
```

## Run Tests Without Local Image
```
TEST_PKG=./test/e2e INVALID_TEST_REQS=skip make test
```

## Run Generated Kube Tests
1. `kind create cluster --name test`
2. Focus the generated kube types test [link](https://github.com/solo-io/gloo/blob/master/projects/gateway/pkg/api/v1/kube/generated_kube_types_test.go#L32)
3. `TEST_PKG=projects/gateway/pkg/api/v1/kube/ INVALID_TEST_REQS=run make test`

# Checklist:


test/testutils/version/changelog.go
```diff
@@ -0,0 +1,56 @@
...
+
+		// We only care about the 2 latest versions
+		if len(latestVersions) > 2 {
+			latestVersions = latestVersions[:2]
+		}
+	}
+
+	// The latest version will be for a changelog entry that is in progress, but not yet released.
+	// Therefore, we always take the second latest
+	secondLatestChangelogVersion := latestVersions[1]
```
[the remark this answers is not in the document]
sam-heilbron: I'm not sure how that would happen. In this logic we parse all changelogs, sort them, and take the two most recent entries, and then select the second most recent entry (since the most recent one is in flight and not released).

I'm happy to add a guard to prevent an NPE, though I don't think the case should ever happen (if it did, we should probably fail loudly anyway)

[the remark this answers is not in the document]
elcasteel: ah, it's starting in `/changelog` not `/changelog/<branch>`

elcasteel (dismissed):

test/testutils/version/changelog.go
```diff
@@ -0,0 +1,56 @@
...
+			// This would happen if a changelog is poorly formatted
+			// We don't want to let that break this functionality
+			continue
+		}
+
+		latestVersions = append(latestVersions, changelogVersion)
+		sort.Slice(latestVersions, func(i, j int) bool {
+			var version1 = *latestVersions[i]
+			var version2 = *latestVersions[j]
+			return !version2.MustIsGreaterThanOrEqualTo(version1)
```
gunnar-solo, line 43: without looking into this function, I'm going to pretend that it is functional in all "difficult" cases.  Ex:
* the transition from an LTS patch version to the first release candidate
* the transition from an LTS patch version to the first changelog in a new LTS

gunnar-solo (approved):
no strong opposition, here.  A few comments, mostly observations

test/e2e/README.md
```diff
@@ -24,9 +24,9 @@ If a test fails, you can retry it using the build-bot [comment directives](https
 
 ## Local Development
 ### Setup
-For these tests to run, we require Envoy be built in a docker container.
+For these tests to run, we require that our gateway-proxy component built as a docker image. If you have made changes to the component, refer to the [Envoyinit README](https://github.com/solo-io/gloo/blob/master/projects/envoyinit) for build instructions.
```
ben-taussig-solo, line 27: Nit: Isn't this no longer true, as of this PR? 

`For these tests to run, we require that our gateway-proxy component built as a docker image.`

Can we change this to something like:

`By default, these tests will run against the latest version of the gateway-proxy/gloo-ee-envoy-wrapper-docker image. To run these tests against a different version, set the ENVOY_IMAGE_TAG environment variable to the tag of the desired image.`

test/e2e/README.md
```diff
@@ -38,9 +38,14 @@ The `run-tests` make target runs ginkgo with a set of useful flags. The followin
...
 | INVALID_TEST_REQS | fail    | The behavior for tests which depend on environment conditions that aren't satisfied. Options are `skip`, `run`, `fail`                                                                                                                             |
 
-Example:
+#### Using Locally Built Image
 ```bash
-ENVOY_IMAGE_TAG=solo-test-image TEST_PKG=./test/e2e/... make run-tests
+ENVOY_IMAGE_TAG=0.0.1-local TEST_PKG=./test/e2e/... make run-tests
+```
+
+#### Using Recently Published Image
```
ben-taussig-solo, line 46: Nit: can we
1. Put the `#### Using Recently Published Image` section before the `#### Using Locally Built Image` section
2. Identify that the recently published image workflow is the default way of doing things
   - we could potentially rename the section to `Using Recently Published Image (default)`
   - or we could add a note under the example command saying the same thing
3. Could we add a note under the `Using Locally Built Image` section indicating that this is necessary if the user has made changes to envoy or the gateway proxy/gloo-ee-envoy-docker-wrapper image?

test/e2e/README.md
sam-heilbron: I'm happy to reword it. We do in fact need the docker image to be built. Either it will be previously built (by our release pipeline) or locally built by a developer.

test/services/envoy.go
```diff
@@ -203,37 +206,10 @@ type EnvoyFactory struct {
...
-	scanner := bufio.NewScanner(inFile)
-	for scanner.Scan() {
-		line := scanner.Text()
-		if strings.HasPrefix(line, prefix) {
-			return strings.TrimSpace(strings.TrimPrefix(line, prefix))
-		}
-	}
-	return ""
+	// The tag of the image that will be used to run the Envoy instance in Docker
+	// Find valid tag names here https://quay.io/repository/solo-io/gloo-envoy-wrapper?tab=tags
```
ben-taussig-solo, line 210: Nit: can we clarify here that you can also build the `gloo-ee-envoy-wrapper-docker` image and use that image's tag?

test/services/envoy.go
```diff
@@ -266,18 +242,19 @@ func NewEnvoyFactory() (*EnvoyFactory, error) {
...
 		tmpdir, err := ioutil.TempDir(os.Getenv("HELPER_TMP"), "envoy")
 		if err != nil {
 			return nil, err
 		}
 
-		envoyImageTag := getEnvoyImageTag()
-		if envoyImageTag == "" {
-			panic("Must set the ENVOY_IMAGE_TAG env var. Find valid tag names here https://quay.io/repository/solo-io/gloo-envoy-wrapper?tab=tags")
-		}
+		envoyImageTag := mustGetEnvoyGlooTag()
```
ben-taussig-solo, line 256: So is this just for convenience? i.e., because the default workflow has less control over the envoy image used, this makes it easy to identify which tag of envoy-gloo is running? If that's the case I think that's really useful, but just want to make sure that this value isn't consumed elsewhere

I also think we should consider renaming this to `envoyGlooImageTag` -- before these change, envoy image tag referred to a gloo-ee-envoy-wrapper image, now it refers to a envoy-gloo image, if I'm not mistaken

test/services/envoy.go
ben-taussig-solo: I'm thinking could we also change the name of this field to something like `envoyWrapperDockerImageTag` to clarify _which_ docker image we're using?

test/services/envoy.go
sam-heilbron: In CI, we use the linux case. It works be specifying an ENVOY_BINARY (which we use) or falling back to the ENVOY_IAMGE_TAG value and pulling the envoy-gloo repo for that tag.

Locally, we use the darwin case. It works by specifying an ENVOY_IMAGE_TAG and using that version. Or falling back to the latest released version. Previously we didn't perform that fallback and tests would just fail locally.

test/services/envoy.go
sam-heilbron: good call!

test/services/envoy.go
```diff
@@ -313,6 +290,60 @@ docker rm $CID
...
+	eit := os.Getenv(testutils.EnvoyImageTag)
+	if eit != "" {
+		return eit
+	}
+
+	// get project base
+	gomod, err := exec.Command("go", "env", "GOMOD").CombinedOutput()
+	Expect(err).NotTo(HaveOccurred())
+	gomodfile := strings.TrimSpace(string(gomod))
+	projectbase, _ := filepath.Split(gomodfile)
```
ben-taussig-solo, line 307: can we use `util.GetModuleRoot()` here like we do in `test/testutils/version/changelog.go` to get the file path to the project base? I think it might be a little cleaner than working out the project base from the go mod location

test/e2e/README.md
sam-heilbron: Great call! I have some updates that ill run by you

test/testutils/version/changelog.go
sam-heilbron: Yeah I really am trusting the existing Version utility in our go-utils library

test/testutils/version/changelog.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/solo-io/go-utils/changelogutils"
+	"github.com/solo-io/go-utils/versionutils"
+	"github.com/solo-io/skv2/codegen/util"
+)
+
+// GetLastReleaseOfCurrentBranch returns the Version of the latest patch version for the current minor version
+// To avoid querying the Github API, we use the changelog folder to determine this version
+// Be aware, that while this has the benefit of not using an external API,
+// we may hit issues where a release failed and therefore the returned version is not actually published.
+func GetLastReleaseOfCurrentBranch() (*versionutils.Version, error) {
```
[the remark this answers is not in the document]
sam-heilbron: Good point! I think it's safe for now given that we expect everyone to be up to date. The impact would just be that we would select the wrong version locally, and behavior may differ. in the end, CI would catch that if it were the case

test/services/envoy.go
ben-taussig-solo: ah, my bad, totally missed the bash command we execute right after this

test/e2e/README.md
ben-taussig-solo: Workshopping here -- I'm thinking we can say something like:
> These tests will run against the gloo-envoy-wrapper image. By default, the latest published version of the image will be used. Alternatively, you can supply the tag of an image to be used by setting the `ENVOY_IMAGE_TAG` environment variable

I'm not in love with that suggestion, because the "latest published version" is not strictly true; there's all that logic about finding an appropriate changelog version and pulling the tag from that. 

Really I just think it's important to get the word "built" out of there, because we want to make it clear that you don't have to build the image yourself

test/services/envoy.go
ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/8061/commits/1e9491031b0dc90b536471d18cac9827fe5c91db

gunnar-solo (approved):
Make it so.

ben-taussig-solo (approved):
Looks good! Big fan of this change

Issue #8003, opened 2023-03-24
E2E Suite Test Flakes (AWS)


sam-heilbron:
### Gloo Edge Version

1.14.x (beta)

### Kubernetes Version

None

### Describe the bug

Logs: https://console.cloud.google.com/cloud-build/builds/4a1c5170-c424-4a51-89d8-9944908dc896;step=8?project=solo-public

```
  INFO	gloo.uds.kube-uds	ended
  DEBUG	gloo.setup	context cancelled
  << Timeline

  [FAILED] Timed out after 300.000s.
  Unexpected error:
      <*eris.rootError | 0xc0065b2ab0>: {
          global: false,
          msg: "404 is not OK",
          ext: nil,
          stack: [0x5daa1f3, 0x1e299aa, 0x1e28348, 0x24ea9e7, 0x24ec5cc, 0x24e7e6d, 0x5da9727, 0x24b02f1],
      }
      404 is not OK
  occurred
  In [It] at: /go/pkg/mod/github.com/onsi/ginkgo/v2@v2.8.1/internal/node.go:463 @ 03/24/23 18:01:32.812

  Full Stack Trace
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Summarizing 1 Failure:
  [FAIL] AWS Lambda Basic Auth With gateway translation [It] should be able to call lambda transformation and regular transformation
```

### Steps to reproduce the bug

These logs were identified from a run in CI

### Expected Behavior

Tests consistently pass

### Additional Context

This test file uses the old structure. We should update this test first to use the next TestContext and see if that fixes the flake.

sam-heilbron:
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=9bca17aa-c01d-4928-8dca-523e372b13b4

```
tep #1 - "run-tests": Step #8 - "run-e2e-tests":   [38;5;13m[PANICKED] Test Panicked[0m
[...]

PR #8075, opened 2023-04-06
Stability: Unit Test Custom Matchers

sam-heilbron:
# Description

We rely on custom matchers in our e2e tests, this PR adds some tests to improve confidence those matchers behave as we would expect

# Context

We've hit some response assertion panics in our aws tests and opened https://github.com/solo-io/gloo/issues/8003 as a result. One of them was the result of an update we needed to make to this custom matcher.

My hope was that if we added tests we might see what is causing the current (infrequent) panic. It didn't demonstrate this, but the side effect is now we have test to demonstrate these utilities work as expected.

# Checklist:

### DIFF
--- a/test/gomega/matchers/contain_headers.go
+++ b/test/gomega/matchers/contain_headers.go
@@ -19,11 +19,7 @@ func ContainHeaders(headers http.Header) types.GomegaMatcher {
 	}
 	headerMatchers := make([]types.GomegaMatcher, 0, len(headers))
 	for k, v := range headers {
-		vals := make([]interface{}, len(v))
-		for i := range v {
-			vals[i] = v[i]
-		}
-		headerMatchers = append(headerMatchers, gomega.WithTransform(transforms.WithHeaderValues(k), gomega.ContainElements(vals...)))
+		headerMatchers = append(headerMatchers, gomega.WithTransform(transforms.WithHeaderValues(k), gomega.ContainElements(v)))
 	}
 	return gomega.And(headerMatchers...)
 }
--- a/test/gomega/matchers/contain_substrings.go
+++ b/test/gomega/matchers/contain_substrings.go
@@ -22,8 +22,8 @@ func ContainSubstrings(substrings []string) types.GomegaMatcher {
 	}
 
 	substringMatchers := make([]types.GomegaMatcher, 0, len(substrings))
-	for i := range substrings {
-		substringMatchers = append(substringMatchers, gomega.ContainSubstring(substrings[i]))
+	for _, str := range substrings {
+		substringMatchers = append(substringMatchers, gomega.ContainSubstring(str))
 	}
 	return gomega.And(substringMatchers...)
 }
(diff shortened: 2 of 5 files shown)


test/gomega/matchers/contain_substrings.go
```diff
@@ -22,8 +22,8 @@ func ContainSubstrings(substrings []string) types.GomegaMatcher {
 	}
 
 	substringMatchers := make([]types.GomegaMatcher, 0, len(substrings))
-	for i := range substrings {
-		substringMatchers = append(substringMatchers, gomega.ContainSubstring(substrings[i]))
+	for _, str := range substrings {
+		substringMatchers = append(substringMatchers, gomega.ContainSubstring(str))
```
jbohanon, line 26: Nit: this change introduces unnecessary copies from slice `substrings` into temp val `str`. The index access is safe since we know the length of source and target. I could see an argument for readability, but I've leaned more toward index access for performance wherever safe. Not blocking though, totally up to you

test/gomega/matchers/contain_headers.go
```diff
@@ -19,11 +19,7 @@ func ContainHeaders(headers http.Header) types.GomegaMatcher {
 	}
 	headerMatchers := make([]types.GomegaMatcher, 0, len(headers))
 	for k, v := range headers {
-		vals := make([]interface{}, len(v))
-		for i := range v {
-			vals[i] = v[i]
-		}
-		headerMatchers = append(headerMatchers, gomega.WithTransform(transforms.WithHeaderValues(k), gomega.ContainElements(vals...)))
+		headerMatchers = append(headerMatchers, gomega.WithTransform(transforms.WithHeaderValues(k), gomega.ContainElements(v)))
```
jbohanon, line 22: :tada:

test/gomega/matchers/contain_substrings_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	DescribeTable("contains substrings",
+		func(expectedSubstrings []string) {
+			actualString := "this is the string"
+			Expect(actualString).To(matchers.ContainSubstrings(expectedSubstrings))
+		},
+		Entry("empty list", []string{}),
+		Entry("empty string", []string{""}),
+		Entry("single substring", []string{"this"}),
+		Entry("multiple substrings", []string{"this", "is", "the"}),
```
jbohanon, line 19: nit: could we place these out of the order they appear in the string? A naive implementation I could really never see us moving toward might accidentally care

jbohanon (commented):
Who watches the watchmen? :detective: 

Jk looks good though, happy to approve as is

test/gomega/matchers/contain_headers.go
nfuden, line 22: What does the error look like if it doesnt contain headers? Does this pop out the entire map instead of just the incorrect one?

sam-heilbron: The failure message will be subject to whatever the matcher is. In this case, the failure will be whatever ContainElements provides, and it will say something like:
expect this list to contain this element

so it won't pop out the entire map, just the failed matcher

test/gomega/matchers/contain_substrings_test.go
sam-heilbron: good call! updated

test/gomega/matchers/contain_substrings.go
sam-heilbron: works for me! updated

jbohanon (approved):

elcasteel (approved):

PR #8076, opened 2023-04-07
test/upgrade: Update version logic to still return last minors versio…

nfuden:
Update the underlying signature to not have incorrect assumptions about erroring
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7517

### DIFF
--- a/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
+++ b/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
@@ -1,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Update upgrade test to not fail on first of new version.
--- a/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
+++ b/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
@@ -1,5 +1,0 @@
-changelog:
-  - type: NON_USER_FACING
-    issueLink: https://github.com/solo-io/gloo/issues/7517
-    description: >-
-      Update upgrade test to not fail on first of new version.
--- a/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
+++ b/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
@@ -1,3 +1,4 @@
 changelog:
   - type: NON_USER_FACING
-    description: Update upgrade test to not fail on first of new version.
+    description: >|
+      Update upgrade test to not fail on first of new version.
--- a/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
+++ b/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
@@ -1,4 +1,4 @@
 changelog:
   - type: NON_USER_FACING
-    description: >|
+    description: >-
       Update upgrade test to not fail on first of new version.
--- a/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
+++ b/changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
@@ -1,4 +1,5 @@
 changelog:
   - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/7517
     description: >-
       Update upgrade test to not fail on first of new version.
--- a/test/kube2e/upgrade/upgrade_suite_test.go
+++ b/test/kube2e/upgrade/upgrade_suite_test.go
@@ -57,7 +57,7 @@ var _ = BeforeSuite(func() {
 	}
 
 	LastPatchMostRecentMinorVersion, CurrentPatchMostRecentMinorVersion, err = upgrade.GetUpgradeVersions(suiteCtx, "gloo")
-	if err != nil && errors.Is(err, upgrade.FirstReleaseError) {
+	if err != nil && (errors.Is(err, upgrade.FirstReleaseError )|| errors. ) {
 		firstReleaseOfMinor = true
 		fmt.Println("First release of minor, skipping some upgrade tests")
 	} else if err != nil {
--- a/test/kube2e/upgrade/upgrade_suite_test.go
+++ b/test/kube2e/upgrade/upgrade_suite_test.go
@@ -41,7 +41,7 @@ var (
 	CurrentPatchMostRecentMinorVersion *versionutils.Version
 	firstReleaseOfMinor                bool
 
-	skipFunc func()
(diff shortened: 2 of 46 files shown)


changelog/v1.15.0-beta1/upgrade-panic-part2.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
```
sam-heilbron, line 2: Can we link https://github.com/solo-io/gloo/issues/7517, it was reopened due to this bug. now that it is resolved, I think we can close the issue and make the upgrade tests required on master

test/kube2e/upgrade/upgrade_suite_test.go
```diff
@@ -34,11 +35,13 @@ var (
...
 	// Versions to upgrade from
 	// ex: current branch is 1.13.10 - this would be the latest patch release of 1.12
-	LastPatchMostRecentMinorVersion *versionutils.Version
+	LastPatchPreviousMinorVersion *versionutils.Version
 
 	// ex:current branch is 1.13.10 - this would be 1.13.9
 	CurrentPatchMostRecentMinorVersion *versionutils.Version
 	firstReleaseOfMinor                bool
+
+	skipFunc func()
```
sam-heilbron, line 44: This skipFunc is only executed for certain tests. Since it lives at the suite level, it may be confusing to future devs when this function should be used. Could we update that name to describe that this applies only to a subset of the tests?

test/kube2e/upgrade/upgrade_test.go
```diff
@@ -58,19 +58,15 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
 
 	Describe("Upgrading from a previous gloo version to current version", func() {
 
-		Context("When upgrading from lastPatchMostRecentMinorVersion to PR version of gloo", func() {
-			if LastPatchMostRecentMinorVersion == nil {
-				fmt.Println("Skipping upgrade tests because LastPatchMostRecentMinorVersion is not set")
-				return
-			}
+		Context("When upgrading from LastPatchPreviousMinorVersion to PR version of gloo", func() {
```
sam-heilbron, line 61: Why are we safe to no longer skip these tests if the LastPatchMostRecentMinorVersion is not set? Was the previous conditional not doing anything?

test/kube2e/upgrade/upgrade_suite_test.go
nfuden: yeah I like it. will do

test/kube2e/upgrade/upgrade_utils.go
```diff
@@ -71,21 +90,39 @@ func getLastReleaseOfCurrentMinor() (*versionutils.Version, error) {
...
 
-	versions := make([]*versionutils.Version, len(files)-1) //ignore validation file
-	for idx, f := range files {
-		if f.Name() != "validation.yaml" {
-			version, err := versionutils.ParseVersion(f.Name())
-			if err != nil {
-				return nil, errors.Errorf("Could not get version for changelog folder: %s\n", f.Name())
-			}
-			versions[idx] = version
+	return filterFilesForLatestRelease(files...)
```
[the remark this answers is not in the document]
nfuden: Alright ive consolidated them and moved the associated test. I would like to update go-utils some day to make the github client testable more easily. Today the best approach i found was rather annoying with putting a mocked http client and dealing with ALL the calls it needs XD

nfuden:
/kick gateway couldnt find its upstream. ie more timing issues

test/testutils/version/changelog.go
```diff
@@ -21,36 +22,60 @@ func GetLastReleaseOfCurrentBranch() (*versionutils.Version, error) {
...
-		})
+	if len(files) < 3 {
+		return nil, nil, errors.Errorf("Could not get sufficient versions from files: %v\n", files)
+	}
 
-		// We only care about the 2 latest versions
-		if len(latestVersions) > 2 {
-			latestVersions = latestVersions[:2]
+	versions := make([]*versionutils.Version, 0, len(files))
+	for _, f := range files {
```
sam-heilbron, line 46: nit: While a directory is technically a file, it may be misleading to readers who assuming we're traversing non-directory files

test/testutils/version/changelog.go
```diff
@@ -21,36 +22,60 @@ func GetLastReleaseOfCurrentBranch() (*versionutils.Version, error) {
...
-	// Therefore, we always take the second latest
-	secondLatestChangelogVersion := latestVersions[1]
-	return secondLatestChangelogVersion, nil
+func (a sortableVersionSlice) Len() int { return len(a) }
+func (a sortableVersionSlice) Less(i, j int) bool {
+	var version1 = *a[i]
+	var version2 = *a[j]
+	return version2.MustIsGreaterThan(version1)
 }
+func (a sortableVersionSlice) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
```
sam-heilbron, line 81: nit: https://pkg.go.dev/sort#Slice sort.Slice allows us to define a comparator inline, instead of needing to add a new type. Your call on which direction you'd prefer

sam-heilbron (approved):
LGTM! Can we port this over to solo-projects as well, which I believe is failing with the same "first entry in release" behavior?

test/testutils/version/changelog.go
nfuden: Technically at this point we are not sure if they are directories we will throw away things like validation.yaml given that they do not conform to our changelog format.

test/kube2e/upgrade/upgrade_utils.go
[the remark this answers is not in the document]
sam-heilbron: Yeah I agree that it would be nice to expand the functionality (and testability) of the github client. Want to open an issue directly in go-utils to track?

test/testutils/version/changelog.go
nfuden: After thinking on it i think the inline is cleaner for a one time use like this

test/testutils/version/changelog.go
sam-heilbron: Yeah not a huge deal. We expect all valid entries to be a directory with the name that corresponds to semantic versioning, and if it doesn't fit that (like validation.yaml) we ignore. I just know if my original PR that introduced some of this logic, @elcasteel had a question about what we were iterating over. Again, I think this is good (and accurate), more just wanted to point it out

elcasteel (approved):

PR #8085, opened 2023-04-12
Graphql per resolver timeout API

jbohanon:
# Description

Update GraphQL API to allow configurable timeouts in resolver definitions.

Additionally, the test upstream server utility has been updated to include a channel of responses. This can be useful in determining the values returned to envoy

The proto comments were only changed because the apostrophes were being viewed as single quotes by my editor and were playing nasty games with the syntax highlighting. These can be reverted if desired

# Context

# Checklist:


test/v1helpers/test_upstream.go
```diff
@@ -149,42 +178,58 @@ func newTestUpstream(addr string, ports []uint32, responses <-chan *ReceivedRequ
...
 			_, _ = rw.Write([]byte(reply))
+			rresp.Code = http.StatusOK
+			rresp.Headers = rw.Header()
+			rresp.Body = []byte(reply)
 		} else if body != nil {
 			_, _ = rw.Write(body)
+			rresp.Code = http.StatusOK
+			rresp.Headers = rw.Header()
+			rresp.Body = body
 		}
```
jbohanon, line 231:
```suggestion
			rresp = *retresp
		} else if reply != "" {
			rw.Write([]byte(reply))
			rresp.Code = http.StatusOK
			rresp.Headers = rw.Header()
			rresp.Body = []byte(reply)
		} else if body != nil {
			rw.Write(body)
			rresp.Code = http.StatusOK
			rresp.Headers = rw.Header()
			rresp.Body = body
		}
```

kdorosh (commented):
api lgtm for now, we will at least want this granularity. happy to defer the rest of the UX improvements until we have explicit customer painpoints

i think the default 1s is reasonable

projects/gloo/api/v1/enterprise/options/graphql/v1beta1/graphql.proto
```diff
@@ -147,6 +147,9 @@ message RESTResolver {
   // before being handled by the graphql server.
   ResponseTemplate response = 3;
   string span_name = 4;
+
+  // The timeout to use for this resolver.
```
kdorosh, line 151: nit; include default timeout in comment

projects/gloo/api/v1/enterprise/options/graphql/v1beta1/graphql.proto
```diff
@@ -186,7 +189,13 @@ message GrpcResolver {
   core.solo.io.ResourceRef upstream_ref = 1;
   // configuration used to compose the outgoing request to a REST API
   GrpcRequestTemplate request_transform = 2;
+  // Unclear why 3 was skipped, but reserve it in case it was previously used and removed
+  reserved 3;
   string span_name = 4;
+
+  // The timeout to use for this resolver. This timeout should be shorter than the
+  // overall request timeout set for the route // TODO(jbohanon) check on this
+  google.protobuf.Duration timeout = 5;
```
kdorosh, line 198: nit; include default timeout in comment

projects/gloo/api/v1/enterprise/options/graphql/v1beta1/graphql.proto
```diff
@@ -147,6 +147,9 @@ message RESTResolver {
   // before being handled by the graphql server.
   ResponseTemplate response = 3;
   string span_name = 4;
+
+  // The timeout to use for this resolver.
+  google.protobuf.Duration timeout = 5;
```
jackstine, line 152: Has it been discussed why applying the timeout to the Resolvers is the best solution as opposed to the `GraphQLApi` maybe under the `GraphQLApiOptions`?

jackstine: or at least considered?

kdorosh: I don't think we have a graphql api options? and even if we did.. it does not make sense for all resolver types (e.g. mock resolver) so i don't think it should live on any parent fields unless it was clearly a default that only applied to the correct resolver types

jbohanon: I originally had it under the `Resolution`, but for the reason @kdorosh listed and also due to the layout of the translation code, it made sense to move it down into the resolvers themselves. I don't think it makes sense to go any higher because of where we consume this timeout in the envoy-gloo-ee code. We [instantiate a resolver](https://github.com/solo-io/envoy-gloo-ee/blob/release-v1.24/source/extensions/graphql/resolvers/rest/rest_resolver.cc#L24) with typed config, which is the level we're configuring [here](https://github.com/solo-io/solo-projects/pull/4758/files#diff-408fa2d2ec5595c9d80f3da7e33db6a33e59380bb2b16e78f843d19bf5ed5e8a) and [here](https://github.com/solo-io/solo-projects/pull/4758/files#diff-34d7b6e30c4734abaa0ac701fd11d6355382c7839307c4e9c144bbdf5f99e48b)

kdorosh (approved):

saiskee (approved):

PR #8096, opened 2023-04-14
build: reduce log burden by suppressing docker build logs

gunnar-solo:
# Description
Add some `2> /dev/null` flags to docker builds on CI

# Context
_optimally_ this makes logs easier to use when tracking down flakes

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,7 +322,7 @@ $(OUTPUT_DIR)/.generated-code:
 .PHONY: verify-enterprise-protos
 verify-enterprise-protos:
 	@echo Verifying validity of generated enterprise files...
-	$(GO_BUILD_FLAGS) GOOS=linux go build projects/gloo/pkg/api/v1/enterprise/verify.go
+	$(GO_BUILD_FLAGS) GOOS=linux go build projects/gloo/pkg/api/v1/enterprise/verify.go $(STDOUT_SILENCE_REDIRECT)
 
 #----------------------------------------------------------------------------------
 # Generate mocks
@@ -359,9 +359,10 @@ generate-client-mocks:
 #----------------------------------------------------------------------------------
 # build-ci and glooctl along with others are in the ci makefile
 
-
-
-
+ifeq ($(USE_SILENCE_REDIRECTS), true)
+STDOUT_SILENCE_REDIRECT := 1> /dev/null
+STDERR_SILENCE_REDIRECT := 2> /dev/null
+endif
 
 #----------------------------------------------------------------------------------
 # Ingress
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -358,8 +358,9 @@ generate-client-mocks:
 # build-ci and glooctl along with others are in the ci makefile
 
 
-
-
+ifeq ($(USE_SILENCE_REDIRECT), true)
```
sam-heilbron, line 361: I like this idea to make the output location configurable! I have an idea that would allow us to:
- Decouple this from purely a silence/not-silence
- Not require 2 separate env variables to track 1 behavior

What if we define a  `DOCKER_OUTPUT_FILE`. We could initialize this to empty:
```
DOCKER_OUTPUT_FILE ?=
```

or specifically to stdout to make it clear. Then in the step to build kind images, we could pass `/dev/null` to show that is the output file?

nfuden:
is there a reasonable place to add this variable to documentation?

gunnar-solo:
/kick https://console.cloud.google.com/cloud-build/builds;region=global/c63337d7-8bdc-468b-a0e8-9a65712a9c1e?project=solo-public

sam-heilbron (approved):
Love this change! I think it's a great addition and while I may have approached it differnetly, that shouldn't prevent this work from being incporporated.

Below is how I may have approached it, feel free to ingore:
Since the context of whether something defaults to stderr or stdout is particular to the make target itself, I would even just add `2>` directly inline to the ones that use it and `1>` to the others.  Then we could just inject the file or default to stdout. This would allow each target to own the complexiy of where logs come from, and the variable to own the complexity of where the logs should go to

Makefile
```diff
@@ -359,9 +360,9 @@ generate-client-mocks:
 #----------------------------------------------------------------------------------
 # build-ci and glooctl along with others are in the ci makefile
 
-
-
-
+ifeq ($(USE_SILENCE_REDIRECTS), true)
+STDERR_SILENCE_REDIRECT := 2> /dev/null
```
jenshu, line 364: will this swallow important errors that we might want to see in case a build fails?

sam-heilbron: Yeah so this is the tradeoff here. My understanding is this will swallow the logs, if enabled, and we enable it in our regression suite. Our true build pipeline which also builds images is not swallowed.

This is why I was comfortable with this change. Should a build fail, I think we'd see the logs in build-bot. And even if we don't, we could easily enable the flag in CI and run again, or really the best is we should just run it locally as something is probably broken that we need to fix.

jenshu (approved):

PR #8120, opened 2023-04-21
Check go version before building docker images

Bslabe123:
make docker now also runs the check-go-version make target as well. Other small bug fixes and improvements

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ fmt-changed:
 
 # must be a seperate target so that make waits for it to complete before moving on
 .PHONY: mod-download
-mod-download:
+mod-download: check-go-version
 	go mod download all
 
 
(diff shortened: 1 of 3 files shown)


Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=fd5cdc62-d010-4666-acd3-1a1b562e29ae

sam-heilbron (commented):
I like this script and think it can be useful! Can you outline on the PR the uses cases that we're trying to cover here? (ie what actions would a developer take that we'd want to protect against).

I think we need to be careful about adding steps to make targets. For example, `make docker` used to have a simple contract: build all docker images. Now it does 2 things. This may be ok if that's the direction we want to go, but I prefer to have targets which do single things, and then recipes which join together targets in a flow that follow some development pattern

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e3581623-0bc4-48a3-aa91-5ab9cab2b8a4

nfuden:
make generated-code should have the check go version as well

Bslabe123:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=0b014ec6-3eec-41d7-b9ba-40fc197ab065

jbohanon (approved):

PR #8138, opened 2023-04-25
Ensure that grpc transcoding works with new US API and old VS API

elcasteel:
# Description

When processing routes with the grpc destination spec defined with a destination upstream that is using the `GrpcJsonTranscoder` serviceSpec, we should not error but also we don't need to create any of the transformations required by the old plugin structure. 
This will allow users to take the following steps during upgrade:

1.  Use discovery to generate grpc upstreams and routes with the grpc destination spec set.
2. Upgrade to 1.14 - discovery will not overwrite the old upstreams
3. Delete the old upstreams and allow them to be recreated with the new API - the old routes will continue to work
4. Swap out routes for the new API

This does assume that the exposed descriptors match the routes defined on the VS but that seems like a reasonable requirement. If anyone objects, we can look into their use case and see there is a config workaround or if we need to implement additional transformations in the plugin.

Also set `matchIncomingRequestRoute = true` on discovered upstreams to allow matchers besides `/`

# Checklist:

### DIFF
--- a/projects/discovery/pkg/fds/discoveries/grpc/grpc.go
+++ b/projects/discovery/pkg/fds/discoveries/grpc/grpc.go
@@ -194,6 +194,7 @@ func (f *UpstreamFunctionDiscovery) DetectFunctionsOnce(ctx context.Context, url
 		}
 		svcSpec.DescriptorSet = &grpc_json_plugins.GrpcJsonTranscoder_ProtoDescriptorBin{ProtoDescriptorBin: rawDescriptors}
 		svcSpec.Services = servicesDiscovered
+		svcSpec.MatchIncomingRequestRoute = true
 		return nil
 	})
 }
--- a/projects/gloo/pkg/plugins/grpc/plugin.go
+++ b/projects/gloo/pkg/plugins/grpc/plugin.go
@@ -78,7 +78,13 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 	if upstreamType.GetServiceSpec() == nil {
 		return nil
 	}
-
+	// If the upstream uses the new API we should record that it exists for use in `ProcessRoute` but not make any changes
+	_, ok = upstreamType.GetServiceSpec().GetPluginType().(*glooplugins.ServiceSpec_GrpcJsonTranscoder)
+	if ok {
+		contextutils.LoggerFrom(params.Ctx).Infof("ELC, recording grpsjson us")
+		p.recordedUpstreams[translator.UpstreamToClusterName(in.GetMetadata().Ref())] = in
+		return nil
+	}
 	grpcWrapper, ok := upstreamType.GetServiceSpec().GetPluginType().(*glooplugins.ServiceSpec_Grpc)
 	if !ok {
 		return nil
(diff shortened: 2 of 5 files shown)


projects/gloo/pkg/plugins/grpc/plugin.go
```diff
@@ -183,7 +188,13 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 			if upstream == nil {
 				return nil, errors.New(fmt.Sprintf("upstream %v was not recorded for grpc route", upstreamRef))
 			}
+			upstreamType, ok := upstream.GetUpstreamType().(v1.ServiceSpecGetter)
```
sam-heilbron, line 191: Why do we get the upstreamType and ok, and then ignore the ok value here?

projects/discovery/pkg/fds/discoveries/grpc/grpc.go
```diff
@@ -194,6 +194,7 @@ func (f *UpstreamFunctionDiscovery) DetectFunctionsOnce(ctx context.Context, url
 		}
 		svcSpec.DescriptorSet = &grpc_json_plugins.GrpcJsonTranscoder_ProtoDescriptorBin{ProtoDescriptorBin: rawDescriptors}
 		svcSpec.Services = servicesDiscovered
+		svcSpec.MatchIncomingRequestRoute = true
```
sam-heilbron, line 197: It seems like this is acritical part of the change, though I don't fully understand the behavior with this true versus false. Can you help me undersatnd?

test/e2e/test_context.go
```diff
@@ -115,7 +115,7 @@ func (c *TestContext) BeforeEach() {
 		WithName(DefaultVirtualServiceName).
 		WithNamespace(WriteNamespace).
 		WithDomain(DefaultHost).
-		WithRoutePrefixMatcher(DefaultRouteName, "/").
+		WithRoutePrefixMatcher(DefaultRouteName, "/test").
```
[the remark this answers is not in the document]
elcasteel: oh right meant to move that

projects/discovery/pkg/fds/discoveries/grpc/grpc.go
elcasteel: Without it, the transcoder changes the request path to `/packageName.MethodName`. If our route only has a matcher for `/semantic/url` then we end up with 404s without it. This makes our guides/normal way of configuring routes work correctly.
I can't actually explain why this option exists or why you would want it to be false.

gunnar-solo (commented):
a request:  could I get a simple(ish) bash script to spin up a cluster, resources, then query it with -both- APIs?

elcasteel:
[migration-demo.zip](https://github.com/solo-io/gloo/files/11401459/migration-demo.zip)
@gunnar-solo 
This has the virtual service configs and the steps to demonstrate.
The post to create a new shelf does look different between the two versions but users could avoid that by leaving out the `body:` field in the relevant proto definitions

```
// Creates a new shelf in the bookstore.
  rpc CreateShelf(CreateShelfRequest) returns (Shelf) {
    option (google.api.http) = {
      post: "/shelf"
      body: "shelf" //don't include this to match old api
    };
  }
```

demo.txt
```diff
@@ -0,0 +1,18 @@
+#create cluster and install gloo 1.13 with fds enabled
```
sam-heilbron, line 1: Should this be checked into the repo?

sam-heilbron (commented):
All looks good to me. One question about the demo.txt file. I'm going to try to run the tests locally to see how they work, before approving

demo.txt
elcasteel: https://github.com/solo-io/engineering-demos/pull/97

gunnar-solo (approved):

PR #8141, opened 2023-04-26
Stability: Ensure kind-reload-% always updates deployment

sam-heilbron:
# Description

Ensure that kind-reload always restarts the deployment

# Context

When you update the image and use the same image tag, the deployment spec doesn't change, and the new pods aren't rolled out.

This update annotates the pod template with a timestamp. This ensures that new pods will be rolled out, and it also makes developing easier because you can see on the pod when the last update was pulled in.

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -790,8 +790,15 @@ kind-build-and-load-%: %-docker kind-load-% ; ## Use to build specified image an
 
 # Update the docker image used by a deployment
 # This works for most of our deployments because the deployment name and container name both match
+# NOTE TO DEVS:
+#	I explored using a special format of the wildcard to pass deployment:image,
+# 	but ran into some challenges with that pattern, while calling this target from another one.
+#	It could be a cool extension to support, but didn't feel pressing so I stopped
 kind-set-image-%:
+	kubectl rollout pause deployment $* -n $(INSTALL_NAMESPACE) || true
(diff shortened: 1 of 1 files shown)


sam-heilbron:
/skip-ci 
/skip-changelog

small dev change

Makefile
```diff
@@ -790,8 +790,15 @@ kind-build-and-load-%: %-docker kind-load-% ; ## Use to build specified image an
 
 # Update the docker image used by a deployment
 # This works for most of our deployments because the deployment name and container name both match
+# NOTE TO DEVS:
+#	I explored using a special format of the wildcard to pass deployment:image,
+# 	but ran into some challenges with that pattern, while calling this target from another one.
+#	It could be a cool extension to support, but didn't feel pressing so I stopped
 kind-set-image-%:
+	kubectl rollout pause deployment $* -n $(INSTALL_NAMESPACE) || true
```
jbohanon, line 798: what is the case where this command needs to be guarded with `|| true` and why do the other commands not need it?

sam-heilbron: I found that if this command had failed previously (while testing) and the target would fail because you can't pause twice in a row. The hope here was to just make this target safer by trying to pause, but continuing anyway if that failed

bewebi (approved):

jbohanon (approved):

Issue #8091, opened 2023-04-13
Gloo validation leaks invalid regexRewrite to Envoy 


bdecoste:
### Gloo Edge Version

1.13.x (latest stable)

### Kubernetes Version

None

### Describe the bug

The following route regexRewrite is leaked by Gloo validation and is rejected by Envoy:

`^/proxy/v1/store/[^/]*/xyz/order/status/[^/]*)`

`[2023-04-13 19:43:03.888][8][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:126] gRPC config for [type.googleapis.com/envoy.config.route.v3.RouteConfiguration](http://type.googleapis.com/envoy.config.route.v3.RouteConfiguration) rejected: unexpected ): ^/proxy/v1/store/[^/]*/xyz/order/status/[^/]*)`

### Steps to reproduce the bug

Deploy a VirtualService with a Route with the regexRewrite above

### Expected Behavior

Invalid route regexRewrite are caught by Gloo validation and do not leak to Envoy

ianmacclancy:
Envoy docs: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-regex-rewrite

Gloo edge docs: https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/regex_rewrite/
- I figure adding regex spec to the docs will be helpful to customers - I will be updating the description of regexRewrite and will see if I can update the guide as well. 

Edge code location: https://github.com/solo-io/gloo/blob/5994ce1005a74a7e5d990ecc5abd1af5c9b3e3fb/projects/gloo/pkg/plugins/basicroute/plugin.go#L106

ianmacclancy:
CFA example link for long regex - https://solo-io-corp.slack.com/archives/C04EBE250E9/p1681996879842649?thread_ts=1681993929.222929&cid=C04EBE250E9

bdecoste:
Can we get this back-ported to 1.12.x?

ianmacclancy:
Link to the conversation that pertains to @bdecoste ask - https://solo-io.slack.com/archives/GPG5SMLH1/p1683133066309609?thread_ts=1681417024.839469&cid=GPG5SMLH1

Issue #8099, opened 2023-04-14
[Generated Kube Types Suite] 1.13-1.14 Generated Kube Code


jackstine:
### Gloo Edge Version

1.13.x (latest stable)

### Kubernetes Version

None

### Describe the bug

Generated Kube Code [It] can read and write a gloo resource as a typed kube object 
"upstreams.gloo.solo.io \"petstore-static\" already exists"

/workspace/gloo/projects/gateway/pkg/api/v1/kube/generated_kube_types_test.go:97
Failing on Line generated_kube_types_test.go:139

We may want to move this to kube e2e tests so that it is tested there and not in the regressions tests suites.

[Logs are here](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=4450e4d8-4532-49db-9625-6c67f6882084)
[Experienced in this PR](https://github.com/solo-io/gloo/pull/8098)

### Steps to reproduce the bug

N/A

### Expected Behavior

Should work

sheidkamp:
Seeing it here too: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d9c6d8c6-0c67-4fa7-9c6a-9e90a421f624 on a backport to 1.13.x

Bslabe123:
Here as well
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e3581623-0bc4-48a3-aa91-5ab9cab2b8a4

artberger:
Saw this [here](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=7826b58e-684d-423f-82c7-9e2455cc7546), from this docs 1.14 backport [PR](https://github.com/solo-io/gloo/pull/8119).

ben-taussig-solo:
Been seeing this a lot the last two days:
 - https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=1d9f9301-ed0a-4aed-92f8-ad6a8280b3a8
 - https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=76f158f5-fdc8-4bd8-ad04-e15aec98b0f7
 - https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e182c3f1-d62f-46e2-84cb-410180c0cbb7

ianmacclancy:
unable to get this pr to build because of this - https://github.com/solo-io/gloo/pull/8183

ashishb-solo:
[...]

PR #8143, opened 2023-04-26
Regex fixes

ianmacclancy:
# Description

Initial work on regex fixes
There will either be a follow-up on compiled regex program max size checks or it will be added to this PR

For these validations, I decided not to rely on envoy validation as this should not be a check that can be disabled by user choice. Having these values leak to envoy would be a bad experience due to user error or choice. 

I have tested this locally and added unit tests for the new util behavior and on the route plugin

There was a rewrite after comments that moves the validation from our gloo translation to the gateway translation. This will catch the errors earlier in our translation process and also aligns with the validation that was already present in the gateway translation. 

There could be a case where a user has written their own proxy translation and can bypass the validation that is added in this change. In this case, the regex validation will be on that user to add to their own customer translation. 

I have renamed validateRoutes to validateRoutesRegex as that name better represents what is happening in the function 

# Context

Users have been able to submit regex that will be passed to and fail the envoy configuration: https://github.com/solo-io/gloo/issues/8091

# Trying out the new validation 
You can follow the guides here - https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/regex_rewrite/

And once set up - change the regex used to an invalid regex pattern in the regexRewrite to invalid regex (ex: "["). You will now see that the resource is rejected and when you check out the VS with kubectl or glooctl you should see the regex syntax error. 

# Checklist:


projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -330,10 +330,27 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
 // and are in their final sorted form
-func validateRoutes(vs *v1.VirtualService, vh *gloov1.VirtualHost, reports reporter.ResourceReports) {
+func validateRoutesRegex(vs *v1.VirtualService, vh *gloov1.VirtualHost, reports reporter.ResourceReports) {
 	for _, rt := range vh.GetRoutes() {
+		options := rt.GetOptions()
+		if options != nil {
+			// validate HostRewrite regex
+			if options.GetHostRewritePathRegex() != nil {
+				_, err := regexp.Compile(options.GetHostRewritePathRegex().GetPattern().GetRegex())
+				if options.GetHostRewritePathRegex().GetPattern().GetRegex() != "" && err != nil {
```
bewebi, line 342:
```suggestion
			if options.GetHostRewritePathRegex().GetPattern().GetRegex() != nil {
				_, err := regexp.Compile(options.GetHostRewritePathRegex().GetPattern().GetRegex())
				if err != nil {
```

It looks like this ought to work and would avoid unnecessarily calling `regexp.Compile()` when it has no chance of succeeding

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -330,10 +330,27 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
+					reports.AddError(vs, InvalidRegexErr(vs.GetMetadata().Ref().Key(), err.Error()))
+				}
+			}
+			// validate RegexRewrite regex
+			if options.GetRegexRewrite() != nil {
+				_, err := regexp.Compile(options.GetRegexRewrite().GetPattern().GetRegex())
+				if options.GetRegexRewrite().GetPattern().GetRegex() != "" && err != nil {
```
bewebi, line 349:
```suggestion
			if options.GetRegexRewrite().GetPattern().GetRegex() != nil {
				_, err := regexp.Compile(options.GetRegexRewrite().GetPattern().GetRegex())
				if err != nil {
```

Same as above

bewebi (commented):
Looks good, just one suggestion (repeated twice) for cleaner logic

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: You can't call a nil check on a string so this would not work. I also think this might panic on trying to access a field of options.hostRewritePathRegex if it is nil

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -330,10 +330,27 @@ func VirtualHostName(vs *v1.VirtualService) string {
 	return fmt.Sprintf("%v.%v", vs.GetMetadata().GetNamespace(), vs.GetMetadata().GetName())
 }
 
-// this function is written with the assumption that the routes will not be modified afterwards,
+// this function is written with the assumption that the routes will not be modified afterward,
 // and are in their final sorted form
-func validateRoutes(vs *v1.VirtualService, vh *gloov1.VirtualHost, reports reporter.ResourceReports) {
+func validateRoutesRegex(vs *v1.VirtualService, vh *gloov1.VirtualHost, reports reporter.ResourceReports) {
```
sam-heilbron, line 335: We are expanding validation of regexes at the route level, are there any regex options available at the virtual_host level? (https://github.com/solo-io/gloo/blob/master/projects/gateway/api/v1/virtual_service.proto#L171). If not, I think it would be helpful to clarify why we are not validating that object (ie because there is no regex)

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -330,10 +330,27 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
 	for _, rt := range vh.GetRoutes() {
+		options := rt.GetOptions()
```
sam-heilbron, line 337: A route can have inline options (https://github.com/solo-io/gloo/blob/master/projects/gateway/api/v1/virtual_service.proto#L239) but it can also have options defined in an external resource (https://github.com/solo-io/gloo/blob/master/projects/gateway/api/v1/virtual_service.proto#L255). Does the validation that we're performing prevent both cases?

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -330,10 +330,27 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
+		if options != nil {
+			// validate HostRewrite regex
+			if options.GetHostRewritePathRegex() != nil {
```
sam-heilbron, line 340: idea: given that we have a few options where regex can be defined, and we really only need to validate the regex, what if we joined all defined regexes into a single expression and validated that? I'm not sure what the performance impact would be, but it's possible that compiling it in a batch is more efficient

ianmacclancy:
CI flake - https://github.com/solo-io/solo-projects/issues/4876
Kube2E istio flake - https://github.com/solo-io/solo-projects/issues/4875
/kick

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: Scope creep - I can create another issue that investigates all regex use in the vs and adds validation if we want that as my next priority

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: This is a good callout - I will see about adding validation for all vs option scenarios

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: We should not do this. You could end up in a situation where two invalid regex expressions combine to form a correct one or a valid and invalid combine to form a valid regex. ex "[" is invalid "ab]" is valid. If you concat the two you end up with "[ab]" which is valid

projects/gateway/pkg/translator/virtual_service_translator.go
sam-heilbron: Is the effort that large? It feels like it's adding other calls to the properties on the virtual service

sam-heilbron:
> solo-io/solo-projects#4876

There is already a issue tracking the generated kube types flake, we should use that. https://github.com/solo-io/gloo/issues/8099. I think issues tracking flakes should live in the same repo as the source code

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: Investigated this - the options from external_options_config merge with the options on the virtual service here - https://github.com/solo-io/gloo/blob/bdc16b3effe839564473de1a4f185950bd259c5c/projects/gateway/pkg/translator/virtual_service_translator.go#L287 

This merge takes place before I validate all regex on the virtual host so this is already covered by these changes

changelog/v1.15.0-beta6/regex-validation.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/8091
+    resolvesIssue: false
+    description: >-
+      Adds a compile check on regex to avoid leaking incorrect regex syntax to envoy
```
sam-heilbron, line 6: Can we clarify which APIs are now validated that previously weren't? This way customers who consume this changelog can underestand the user-facing impact

projects/gateway/pkg/translator/virtual_service_translator.go
sam-heilbron: Oh sorry yeah I just mean the VS. There are virtual host level options (apply to all routes on a virtual host) and route level options (apply to just the route). I looked and it didn't seem like there were any options on the virtual host where you can defined regex. It seems like the kind of thing we can comment in our validateRoutesRegex function

projects/gateway/pkg/translator/virtual_service_translator.go
sam-heilbron: Nice! Is there a test that covers this case?

sam-heilbron (commented):
This looks great! Can you include the end to end steps a user could take to install gloo and confirm that virtual services with invalid regexes are now rejected by our webhook and not send to envoy?

projects/gateway/pkg/translator/virtual_service_translator.go
ianmacclancy: The virtual host is modified by the merge and then used in ValidateRouteRegex. The tests I added cover all validation that could take place in validateRouteRegex. The changes the merge could make are fully covered.

changelog/v1.15.0-beta6/regex-validation.yaml
ianmacclancy: yup added and renamed the changelog

bewebi (approved):

gunnar-solo (approved):
LGTM

nfuden:
does this cover https://github.com/solo-io/gloo/blob/dcp-cp1/projects/gloo/pkg/plugins/cors/plugin.go#L120

nfuden:
also https://github.com/solo-io/gloo/blob/dcp-cp1/projects/ingress/pkg/translator/translate.go#L221

ianmacclancy:
@nfuden  no - this only covers the mentioned fields here - https://github.com/solo-io/gloo/pull/8143/files#diff-0d49ccf69ef98a89ae479f1ab9e026978cc150003a9a72dc9138e374983f2b48R6

There are a lot more locations where we have regex that are not validated. It would make sense to do a full audit of regex in our apis - we could have a spike for that and follow up tasks for locations that are not validated. At that point we need to really consider performance impact of validation

PR #8154, opened 2023-04-27
Filter chain matching tests

ashishb-solo:
# Description

Add some tests to more thoroughly validate our filter chain matching behaviour

# Context

As we are migrating from Envoy's old filter chain match to the new API, we want to write some tests to make sure that we clearly test our current behaviour. We only use filter chain matchers on source IP and servername, so those are the only two values that are being set and tested here.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/4812


test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+				testContext: testContext,
+			}
+		})
+
+		// Table test:
+		// Each entry contains a connection properties struct that is used to
+		// create a request.
+		// And a map of named matchers (name is arbitrary) to configure envoy with.
+		// the the last argument is the name of the matcher that should match,
+		// or `NoMatch` if nothing should match.
```
bewebi, line 234:
```suggestion
		// Each entry contains a connection properties struct that is used to
		// create a request and a map of named matchers (name is arbitrary)
		// to configure envoy with.
		// The the last argument is the name of the matcher that should match,
		// or `NoMatch` if nothing should match.
```
nit: grammar

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+							{
+								AddressPrefix: "1.2.3.4",
+								PrefixLen: &wrappers.UInt32Value{
+									Value: 32,
+								},
+							},
+						},
+					},
+				}, "more-specific"),
+			Entry("sni matcher with multiple source ip",
```
bewebi, line 382: This test has the same name as the previous

Not sure if this is the best name for it, just a suggestion, feel free to come up with a better one
```suggestion
			Entry("sni matcher with multiple source ip, less precise CIDR range match",
```

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+
+type GwTester struct {
+	secret      *gloov1.Secret
+	testContext *e2e.TestContext
+}
+
+func (gt *GwTester) configureEnvoy(matchers map[string]*v1.Matcher) {
+	// create a magic servername value to ensure that envoy is configured. we
+	// first send a request against this magic servername to make sure envoy
+	// has been fully configured
```
bewebi, line 426:
```suggestion
	// create a magic servername value to ensure that envoy is configured
	// we first send a request against this magic servername to make sure
	// envoy has been fully configured
```
nit: formatting

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+	}
+	requestBuilder := testContext.GetHttpRequestBuilder().WithScheme("https").WithPort(defaults.HybridPort)
+	proxyProtocolBytes = []byte("PROXY TCP4 " + srcip.String() + " 1.2.3.4 123 123\r\n")
+	client := testutils.DefaultClientBuilder().
+		WithTLSRootCa(gloohelpers.Certificate()).
+		WithProxyProtocolBytes(proxyProtocolBytes).
+		WithTLSServerName(sni).
+		Build()
+
+	// skip ssl verification as it wouldnot work for test.com
```
bewebi, line 495:
```suggestion
	// skip ssl verification as it would not work for test.com
```

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+				},
+			}},
+		},
+	}
+	var matchedGw []*v1.MatchedGateway
+
+	i := 0
+	for name, m := range matches {
+		i++
+		curvs := gatewaydefaults.DirectResponseVirtualService(gw.Metadata.Namespace, fmt.Sprintf("vs-%s-%d", configver, i), name)
```
bewebi, line 526: nit: `curVs` is a clearer var name

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+	// skip ssl verification as it wouldnot work for test.com
+	//       Get "https://localhost:8087/": tls: failed to verify certificate: x509: certificate is valid for gateway-proxy, knative-proxy, ingress-proxy, not test.com
+	client.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify = true
+
+	return client.Do(requestBuilder.Build())
+}
+
+func (gt *GwTester) getGwWithMatches(configver string, matches map[string]*v1.Matcher) ([]*v1.VirtualService, *v1.Gateway) {
+	gw := gatewaydefaults.DefaultHybridGateway(writeNamespace)
+	vs := []*v1.VirtualService{}
```
bewebi, line 504: nit: `vss` would be a clearer var name (ie plural of vs) so that below the loop can use `vs` as the singular

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+
+type ClientConnectionProperties struct {
+	SrcIp net.IP
+	SNI   string
+}
+
+type GwTester struct {
+	secret      *gloov1.Secret
+	testContext *e2e.TestContext
+}
```
bewebi, line 421: nit: I personally would prefer to define these before they're used in the table test, just so it's easier to read from top to bottom
just a matter of preference up to you whether it's worthwhile

bewebi: Similarly I would order the helper functions in order of first use, so:
- `GetMatchedMatcher()`
- `configureEnvoy()`
- `getGwWithMatches()`
- `makeARequest()`

Also a matter of preference, but as is it's a bit confusing when read from top to bottom

test/e2e/hybrid_test.go
```diff
@@ -163,4 +185,367 @@ var _ = Describe("Hybrid Gateway", func() {
...
+		}
+		defer resp.Body.Close()
+		return resp.StatusCode, nil
+	}, "5s", "0.1s").Should(Equal(200))
+}
+
+// Configure the gateway with the provided `matchers`, then send a request
+// against the gateway using the information in ClientConnectionProperties and
+// return the matcher that is matched.
+func (gt *GwTester) GetMatchedMatcher(cp ClientConnectionProperties, matchers map[string]*v1.Matcher) string {
```
bewebi, line 463: Is there a need to export this function?

bewebi (commented):
A handful of nits, but really looks great
I love the structure, use of a table, and highly descriptive comments

test/e2e/hybrid_test.go
ashishb-solo: 0d9a539f665620c185ee1101bd7cacba36059ded

test/e2e/hybrid_test.go
ashishb-solo: 56073ac779e2e10e24f506a9f2000116d261045b

bewebi (dismissed):
LGTM!

test/e2e/hybrid_test.go
```diff
@@ -163,4 +325,227 @@ var _ = Describe("Hybrid Gateway", func() {
...
+						},
+					},
+				}, NoMatch),
+			Entry("sni matcher with multiple source ip",
+				ClientConnectionProperties{
+					SrcIp: net.ParseIP("1.2.3.4"),
+					SNI:   "foo.test.com",
+				},
+				map[string]*v1.Matcher{
+					"more-specific": {
```
ben-taussig-solo, line 502: nit: I think `more-specific` is a little confusing here, given that there's only one matcher in this test

ben-taussig-solo:
These tests are great! Very easy to read through. 
Some outstanding questions:
 - Are wildcards supported in SourcePrefixRange/IP matchers? 
	 - I.E., could you do something like:
		```go
		SourcePrefixRanges: []*v3.CidrRange{
			{
				AddressPrefix: "1.2.3.*",
				PrefixLen: &wrappers.UInt32Value{
					Value: 32,
				},
			},
		},
		```
	- If so, are specificity rules applied to IP constraints?
		- If they are, can we test them? In order words, replicate the `most specific sni matcher` test cases for IP matchers 
 - Is it possible to configure an IP matcher without an SNI matcher?
	 - We never do this in these tests -- seems like a potential blind spot
	 - If this is possible, can you create a `ip match` test case to go with the `sni match` case?
	 - If this is possible, could you also create a `most specific ip matcher` case?
 - I think we should have a `sni non-match, ip match` test case to match the `sni match, ip non-match` case

ashishb-solo:
> * Are wildcards supported in SourcePrefixRange/IP matchers?
>     If they are, can we test them? In order words, replicate the most specific sni matcher test cases for IP matchers

There is a wildcard-like feature for `SourcePrefixRange` values - you would achieve it by just reducing the `PrefixLen`. However, it does not change the output of the filter chain selection. Whereas with `servername`, if multiple wildcard values match, the most specific is chosen, there is no such specificity on the `SourcePrefixRange`. If any of the values for it are true, then the filter chain continues, otherwise the connection is terminated.

I'm happy to add another test to verify this at the top level, but I think permuting on this would be excessive because it always results in a true/false value in the end. Do you think that's okay?

ashishb-solo:
> Is it possible to configure an IP matcher without an SNI matcher?
>     We never do this in these tests -- seems like a potential blind spot
>     If this is possible, can you create a ip match test case to go with the sni match case?
>     If this is possible, could you also create a most specific ip matcher case?
> I think we should have a sni non-match, ip match test case to match the sni match, ip non-match case

Great suggestions, I'll add all of these

ashishb-solo:
> Is it possible to configure an IP matcher without an SNI matcher?

https://github.com/solo-io/gloo/pull/8154/commits/1cbb5d9ad94c756a10e14f48efe48949d2c68d67

I was surprised to find out that we cannot configure an IP matcher without an SNI matcher. See the comment in the linked commit for details

>  I think we should have a sni non-match, ip match test case to match the sni match, ip non-match case

https://github.com/solo-io/gloo/pull/8154/commits/e376bfcfacf445a1da6c84a94f1315c4cabdf94d

ashishb-solo:
hit by [this flake](https://github.com/solo-io/gloo/issues/8099)

i have tried to add it to this pull request

ashishb-solo:
Reverted https://github.com/solo-io/gloo/pull/8154/commits/f505a2893de21bc3f3d854f7cba75732baa612c0 - it did not resolve the flake

ben-taussig-solo (approved):
Looks good to me -- Do you know if this will pass CI without the generated_kube_types_test fix?

ashishb-solo:
> Looks good to me -- Do you know if this will pass CI without the generated_kube_types_test fix?

looks like it did already!

bewebi (approved):

ben-taussig-solo (approved):

PR #8112, opened 2023-04-20
docs oidc cookie encryption

jackstine:
# Description

Adding in docs for oidc cookie encryption.

# Context

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #8170, opened 2023-04-28
1.14.x API adding OIDC ciphers (#8011)

jackstine:
# Description

backport of https://github.com/solo-io/gloo/pull/8011
backport of docs https://github.com/solo-io/gloo/pull/8112

# Context

# Checklist:


projects/gateway/pkg/api/v1/kube/generated_kube_types_test.go
```diff
@@ -84,6 +85,22 @@ var _ = Describe("Generated Kube Code", func() {
 			SharedCache: kubeCache,
 		})
 		Expect(err).NotTo(HaveOccurred())
+		err = glooV1Client.Upstreams("default").Delete(ctx, "petstore-static", v1.DeleteOptions{})
```
sam-heilbron, line 88: What are we attempting to solve here? I'm concerned that we're deleting resources in a beforeEach. It feels like an anti-pattern. 

My proposal is that either: we delete resource in an AfterEach, or directly in the test that creates them.

Also, I think we can explore the DeleteOptions which exposes an "ignoreNotExist" that I think we can use. Alternatively, I think we can write the error check a little more idiomatically, where we assert the error didn't happen or matches one of the expect errors:
```
Expect(err).To(Or(Not(HaveOccurred), MatchError("not found"), MatchError("already exists"))
```

jackstine: I am trying to solve the flake error, `"upstreams.gloo.solo.io \"petstore-static\" already exists"`. This occurs in t he original [here](https://github.com/solo-io/gloo/blob/v1.14.x/projects/gateway/pkg/api/v1/kube/generated_kube_types_test.go#L136), when we create the resource.  I do not know why this resource already exists, but this seems to fix it.

`ignoreNotExist` is supported by `github.com/solo-io/solo-kit/pkg/api/v1/clients`, this is `k8s.io/apimachinery/pkg/apis/meta/v1`.
`MatchError()` does a `==` equivalent match not a `strings.Contains()` match. So I have to know the error message to have this correct. The error message is dynamically related to the name of the resource being deleted.

sam-heilbron (approved):

elcasteel (approved):

PR #7727, opened 2023-01-24
7473 security context customization

sheidkamp:
# Description

This new feature can be used to customize security configurations in the gloo and gloo-gateway Helm deployment charts.

Still outstanding: tests and documentation. Current install tests are passing (`TEST_PKG=install/test make test`)

Manual validation steps:
In master, generate helm charts and save to a file:
```
 VERSION=$VERSION make build-test-chart
 helm template install/helm/gloo  > ../helm_templates_master.txt
```
then check out this branch and generate charts with the new code:
```
 VERSION=$VERSION make build-test-chart
 helm template install/helm/gloo  > ../helm_templates_branch.txt
```
then diff the files to see the changes. It should only be ordering of securityContext elements.
you can also update `glooMtls.enabled` to `true` and repeat the process.

Added the ability to specify the PodSecurityContext for the pod and the SecurityContext for the gateway-proxy, sds, and istio-proxy containers in the gateway-proxy-deployment chart
      and the SecurityContext in the envoy-sidecar, sds, and gloo containers in the gloo-deployment chart. You can expose access to these values with the following helm values:
        gatewayProxies.gatewayProxy.podTemplate.podSecurityContext.
        gatewayProxies.gatewayProxy.podTemplate.glooContainerSecurityContext
        gatewayProxies.gatewayProxy.podTemplate.containerSecurityContext
        gloo.deployment.glooContainerSecurityContext
        global.glooMtls.istioProxy.securityContext
        global.glooMtls.envoy.securityContext
        global.glooMtls.sds.securityContext

~~There are commented out security contexts in `values-template.yaml` that demonstrate where the custom security contexts would be places, and can be uncommented to demonstrate the new functionality.~~ removed for final PR

# Context

Users needed this feature to specify custom security configurations to conform to organization standards and allow integration with other tools

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
[...]

PR #8192, opened 2023-05-04
Add securityContext include and update templates

sheidkamp:
# Description

Similar to https://github.com/solo-io/gloo/pull/7727/

Adding a "merge" option to our security context overrides to allow the defined security context to define updates to the defaults instead of entirely replacing them.

# Context

We have a client that wants to update the `runAsUser` field without changing the rest of the default securityContext in a deployment where we don't have something like  [runAsUser](https://github.com/solo-io/gloo/blob/d36213e98289c28ebd4017d268f77afec0e6f9e5/install/helm/gloo/generate/values.go#LL251C10-L251C10) implemented.

# Checklist:


install/helm/gloo/generate/values.go
```diff
@@ -241,27 +241,40 @@ type Gloo struct {
 	DisableLeaderElection *bool                 `json:"disableLeaderElection,omitempty" desc:"Set to true to disable leader election, and ensure all running replicas are considered the leader. Do not enable this with multiple replicas of Gloo"`
 }
 
+type SecurityOpts struct {
+	Merge *bool `json:"merge,omitempty" desc:"If true, merge this security context with the default security context. If false, replace the default security context with this security context."`
```
jackstine, line 245: what does merge mean, I think it is worth it to note to users that this is [a helm merge.](https://helm.sh/docs/chart_template_guide/function_list/#merge-mustmerge)

jackstine: Please note what takes precedence as well.

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -91,6 +91,26 @@ ttlSecondsAfterFinished: {{ . }}
 {{- end -}}
 {{- end -}}
 
+
+{{- define "gloo.securityContext" }}
```
jackstine, line 95: please add some comments on this defined helm method

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -91,6 +91,26 @@ ttlSecondsAfterFinished: {{ . }}
...
+  {{- if .values.merge -}}
+    {{- $overwrite = false -}}
+  {{- end -}}
+{{- end -}}
+{{- if $overwrite -}}
+  {{- $securityContext = or .values .defaults (dict) -}}
+{{- else -}}
+  {{- $securityContext = merge .values .defaults }}
+{{- end }}
+  {{- $securityContext = omit $securityContext "merge" -}}
```
jackstine, line 108: I think it would be best to say why we are omitting the "merge" field

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -82,14 +81,10 @@ spec:
...
-        {{- if not .Values.gloo.deployment.floatingUserId -}}
+        {{- if not .Values.gloo.deployment.floatingUserId }}
           {{- $_ := set $securityDefaults "runAsUser" .Values.gloo.deployment.runAsUser }}
-        {{- end -}}
-        {{- $securityContext := or  .Values.global.glooMtls.envoy.securityContext $securityDefaults }}
-        {{- with $securityContext }}
-        securityContext:
-        {{- toYaml . | nindent 10  }}
         {{- end }}
+{{ include "gloo.securityContext" (dict "values" .Values.global.glooMtls.envoy.securityContext "defaults" $securityDefaults) | indent 8 }}
```
jackstine, line 87: `{{- $securityContext := or  .Values.global.glooMtls.envoy.securityContext $securityDefaults }}` implies that use the first or the second.  Since this is not a merge, are we ok with this change?  Is this is a breaking change?

Since this is in place, I think we should use the same logic we used before.

jackstine: There are several instances that this occurs similarly.

install/helm/gloo/generate/values.go
jackstine: also what is the default value?

install/test/helm_test.go
```diff
@@ -5899,6 +5899,84 @@ metadata:
 					Entry("1-gloo-deployment-sds", "gloo", "sds", "global.glooMtls.sds.securityContext", "global.glooMtls.enabled=true"),
 				)
 
+				DescribeTable("merges resources for container security contexts", func(resourceName string, containerName string, securityRoot string, extraArgs ...string) {
+					// Run once "plain" to get the baseline
+					prepareMakefile(namespace, helmValues{
+						valuesArgs: extraArgs,
+					})
+
+					// Get the resources
```
jackstine, line 5908: I did not find a test that uses the `floatingUserId`, can we include one table to cover that as well, so that we know we can set this value. And then one that uses has this value when it is merged as well.

sheidkamp: tests created, but only covers places where `floatingUserId` and the new include are both in use.

install/helm/gloo/generate/values.go
```diff
@@ -241,27 +241,40 @@ type Gloo struct {
 	DisableLeaderElection *bool                 `json:"disableLeaderElection,omitempty" desc:"Set to true to disable leader election, and ensure all running replicas are considered the leader. Do not enable this with multiple replicas of Gloo"`
 }
 
+type SecurityOpts struct {
+	ApplyAsHelmMerge *bool `json:"applyAsHelmMerge,omitempty" desc:"If true, merge this security context with the default security context according to the logic of the helm 'merge' function. This is intended to be used to modify a field in a security context, while using all other default values. Please note that due to how helm's 'merge' functino works, you can not override a 'true' value with a 'false' value, and for that case you will need to define the entire security context and set this values to false. If false, replace the default security context with this security context. Default value is false."`
```
sam-heilbron, line 245: Love this! Are there any other merge options that we could support in the future? Wondering if this should be a string "mergePolicy" field that supports: "", "no-merge", and "helm" for now, but gives us room to add others. If there are no other merge options I'm good with this as is!

install/helm/gloo/generate/values.go
```diff
@@ -241,27 +241,40 @@ type Gloo struct {
...
+}
+type PodSecurityContext struct {
```
sam-heilbron, line 247: I like this a lot

sam-heilbron (commented):
This is awesome work! I think this sets a good standard moving forward for defining shared properties and re-using them. One small question about the merge functionality but apart from that LGTM

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -105,17 +105,19 @@ Because of this, if a value is "true" in defaults it can not be modified with th
 {{- $securityContext := dict -}}
 {{- $overwrite := true -}}
 {{- if .values -}}
-  {{- if .values.applyAsHelmMerge -}}
+  {{- if eq .values.mergePolicy "helm-merge" -}}
     {{- $overwrite = false -}}
+  {{- else if and .values.mergePolicy (ne .values.mergePolicy "no-merge") -}}
+  {{- fail "value for mergePolicy is not as expected" }}
```
sam-heilbron, line 111: nit: let's include the allowed values in the error just so its trivial for users to know how to fix it should they hit this case

sheidkamp: done

jackstine (approved):

jackstine (approved):

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=ca2d10af-e31b-4fca-a084-5452683bf92c

This is a flake the stability team is planning on tackling next week so I'm not going to go through the steps to track the frequency of this since its on our radar

sam-heilbron (approved):

PR #8213, opened 2023-05-11
Add gRPC transcoding filter to routes

elcasteel:
When grpc transcoding is defined on upstreams, create filters on the routes to the upstreams instead of on the listener. 

# Description
Previously, when we generated gRPC transcoding filters based on config from upstreams, we would add all the filters to every listener. With this change we add specific filters to the routes and an empty filter at the listener level.  

## Performance
This adds a `ProcessRoute` method to add filters to routes. When no gRPC upstreams are configured we don't do anything and when gRPC upstreams are configured, we check the destination of every route to see whether it needs a filter added.

## Testing
I added new unit tests to demonstrate that filters appear on the route level. The existing e2e tests will show that envoy is happy with the configuration and can continue to route traffic. 
Example filter:
```
"routes": [
         {
          "match": {
           "prefix": "/",
           "headers": [
            {
             "name": ":method",
             "safe_regex_match": {
              "google_re2": {
               "max_program_size": 1024
              },
              "regex": "GET|POST"
             }
            }
           ]
          },
          "route": {
           "cluster": "default-grpcstore-demo-80_gloo-system"
          },
          "typed_per_filter_config": {
           "envoy.filters.http.grpc_json_transcoder": {
            "@type": "type.googleapis.com/envoy.extensions.filters.http.grpc_json_transcoder.v3.GrpcJsonTranscoder",
            "services": [
             "solo.examples.v1.StoreService"
            ],
            "proto_descriptor_bin": "CqwFChVnb29nbGUvYXBpL2h0dHAucHJvdG8SCmdvb2dsZS5hcGkieQoESHR0cBIqCgVydWxlcxgBIAMoCzIULmdvb2dsZS5hcGkuSHR0cFJ1bGVSBXJ1bGVzEkUKH2Z1bGx5X2RlY29kZV9yZXNlcnZlZF9leHBhbnNpb24YAiABKAhSHGZ1bGx5RGVjb2RlUmVzZXJ2ZW=",
            "match_incoming_request_route": true
           }
```

# Context
This bug has been around for a while. The fix is only in the new API.

# Checklist:


projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -55,7 +55,7 @@ const (
 var pluginStage = plugins.BeforeStage(plugins.OutAuthStage)
 
 type plugin struct {
-	upstreamFilters []plugins.StagedHttpFilter
+	upstreamFilters map[string]plugins.StagedHttpFilter
```
sam-heilbron, line 58: nit: Can we clarify what this is a mapping between? Ive found this helpful when reading over code to know what the keys are/represent

sam-heilbron (commented):
In this code, there are a number of loops that occur in some of our plugin functions (which themselves occur in a loop). Can you outline the general performance impact of these changes? (even something like: we iterate over N routes, lookup M upstreams , and iterate over L listeners so we are adding N*M*L lookups (that is made up, but an example)

Ideally we would have some performance testing around this space, and that is somethign we are going to add with the stability initiative, so I don't think its necessary for this work. But I would like a ballpark estimate so we are aware of potential impact on larger systems

projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -102,9 +107,18 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+	var usFilters []plugins.StagedHttpFilter
+	// Skip looking at all the routes on the listener if we know no upstreams have a grpcJsonTranscoderSpec
+	if len(p.upstreamFilters) > 0 {
+		//this loops over each upstream on the listener and does a fixed time check to see if it's the map of gRPC upstreams.
+		for _, listenerUs := range upstreamsFromListener(params, listener) {
+			if filter, ok := p.upstreamFilters[listenerUs.Key()]; ok {
+				usFilters = append(usFilters, filter)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
elcasteel: changed this implementation entirely

projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -102,24 +113,53 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+		return err
 	}
-
-	grpcJsonFilter, err := plugins.NewStagedFilter(wellknown.GRPCJSONTranscoder, envoyGrpcJsonConf, pluginStage)
-	if err != nil {
-		return nil, eris.Wrapf(err, "generating filter config")
+	for _, us := range routeUpstreams {
+		if filter, ok := p.upstreamFilters[us.Key()]; ok {
+			if out.GetTypedPerFilterConfig() == nil {
+				out.TypedPerFilterConfig = make(map[string]*any.Any)
```
elcasteel, line 150: Other plugins use `MarkFilterPerPluginConfig` to set the config here, but part of that marshals the config into an `any.Any` struct and when I do that with these filters Envoy complains about `Didn't find a registered implementation for 'envoy.filters.http.grpc_json_transcoder' with type URL: 'google.protobuf.Any'`

projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -102,24 +109,53 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+		envoyGrpcJsonConf, err := translateGlooToEnvoyGrpcJson(params, grpcJsonConf)
+		if err != nil {
+			return nil, err
+		}
+		grpcJsonFilter, err := plugins.NewStagedFilter(wellknown.GRPCJSONTranscoder, envoyGrpcJsonConf, pluginStage)
+		if err != nil {
+			return nil, eris.Wrapf(err, "generating filter config")
+		}
+		return []plugins.StagedHttpFilter{grpcJsonFilter}, nil
+	} else if len(p.upstreamFilters) > 0 {
```
sam-heilbron, line 123: This logic adds the "placeholder" http filter, if any upstreams require it. Should this instead be checking if any upstreams routed to by this listener require it?

For example:
listener-east has routes to us-ne, us-se
listener-west has routes to us-nw, us-sw

In this instance, shouldn't we only add the placholder filter to listener-east IFF us-ne or us-se requires it?

projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -102,24 +109,53 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+		// There needs to be a filter on the listener to be able to set filters on routes
+		// Envoy errors if this config is empty
+		emptyGlooConfig := &grpc_json.GrpcJsonTranscoder{
```
sam-heilbron, line 126: What is the behavior when this is defined, but the route does not override it?

projects/gloo/pkg/plugins/grpcjson/plugin.go
sam-heilbron: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1673022330837809?thread_ts=1673020471.840819&cid=G01EERAK3KJ

What situation led to this error ocurring? I would have expected that if we register the grpc-json-transcoder properly then Envoy wouldn't complain. Though I see it https://github.com/solo-io/gloo/blob/main/projects/envoyinit/hack/filter_types/filter_types.gen.go#L116 which confuses me about how this error occurs

sam-heilbron (commented):
Since we've changed the logic of the plugin, can we try to expand our e2e tests as well to track the various behaviors? Ideally if we can prove the cases with route specific configuration since that is what is being added in this PR

projects/gloo/pkg/plugins/grpcjson/plugin.go
elcasteel: It works normally

projects/gloo/pkg/plugins/grpcjson/plugin.go
elcasteel: My simplest set up with the grpc bookstore works with this and gets the `Didn't find a registered implementation...` error with `MarkFilterPerPluginConfig`

elcasteel:
> Since we've changed the logic of the plugin, can we try to expand our e2e tests as well to track the various behaviors? Ideally if we can prove the cases with route specific configuration since that is what is being added in this PR

I added unit tests to demonstrate that filters are on routes when we expect and not on listeners when they shouldn't be. The end to end behavior hasn't changed - we still expect to be able to route to gRPC services in the same way - so the existing e2e tests are still correct.

projects/gloo/pkg/plugins/grpcjson/plugin.go
sam-heilbron: And was this running Gloo installed in a kube cluster or locally? Are there other instances where this exists in our codebase? I'm nervous that if this is the first instance, we're misunderstanding something else that is going on, and side stepping the problem instead of addressing it.

projects/gloo/pkg/plugins/grpcjson/plugin.go
```diff
@@ -102,24 +107,59 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
...
+	} else if len(p.upstreamFilters) > 0 {
+		if _, ok := p.affectedListeners[listener]; !ok {
+			return nil, nil
+		}
+		// There needs to be a filter on the listener to be able to set filters on routes
+		// Envoy errors if this config is empty
+		emptyGlooConfig := &grpc_json.GrpcJsonTranscoder{
+			DescriptorSet: &grpc_json.GrpcJsonTranscoder_ProtoDescriptor{ProtoDescriptor: "/to/override"},
+		}
+		envoyGrpcJsonConf, err := translateGlooToEnvoyGrpcJson(params, emptyGlooConfig)
```
nfuden, line 130: What do the logs look like with grpc set on one upstream but not on all. Is there a way to avoid the envoy log in that case?

elcasteel: There's one debug level log line so not excessively verbose (and users could configure multiple listeners if they really wanted isolation I guess)

nfuden (approved):

sam-heilbron (approved):
Nice work!

PR #8228, opened 2023-05-16
Fix non-persisted proxy validation in glooctl check

ben-taussig-solo:
# Description
 - Updates `glooctl check` to use the GPRC endpoint for proxy validation, ala `glooctl get proxies`.
 - This fixes an issue where the check would report false-positives for invalid proxies that were not persisted.
   - Prior to this change, glooctl check used a k8s client to obtain proxy resources
   - Since 1.12.x, proxy resources are not persisted by default, so this k8s client failed to obtain proxies, and in turn failed to report errors on proxy resources

# Validating these changes
 - Build glooctl locally
    - `make build-cli-local`
    - for me, this is equivalent to `make glooctl-darwin-amd64`. This will differ depending on your OS and system architechture. I will use `glooctl-darwin-amd64` as a stand-in for the built glooctl binary for the rest of this PR description
  - Install Gloo and set up the failing resources
    - `helm repo update`
    - `helm install gloo gloo/gloo -n gloo-system --create-namespace --devel --version v1.15.0-beta6`
    - ```
      kubectl apply -f - <<EOF
      apiVersion: gloo.solo.io/v1
      kind: Upstream
      metadata:
        name: httpbin
        namespace: gloo-system
      spec:
        static:
          hosts:
            - addr: httpbin.org
              port: 80
      EOF
      ```
    - ```
      kubectl apply -f - <<EOF
      kind: VirtualService
      metadata:
        name: test-static
        namespace: gloo-system
      spec:
        virtualHost:
          domains:
            - '*'
          routes:
            - matchers:
              - prefix: /inja
              options:
                prefixRewrite: /
                transformations:
                  responseTransformation:
                    transformationTemplate:
                      body:
                        text: '{% if int(header("content-length")) > 2000 %} {"error" : "Too
                          Damn Big!!"} {% else %} {{ body() }} {% endif %}'
                      headers:
                        :status:
                          text: '{% if int(header("content-length")) > 2000 %}401{% else %}{{header(":status")
                            }}{% endif %}'
                      parseBodyBehavior: DontParse
              routeAction:
                single:
                  upstream:
                    name: httpbin
                    namespace: gloo-system
      EOF
      ```
  - Update the virtual service so that the body transformation is no longer valid (e.g., remove the leading `{`)
  - Use the newly built binary and confirm that an error with the proxy is reported
    - `./_output/glooctl-darwin-amd64 check`
  - Use system glooctl install and confirm that an error with the proxy is not reported
    - `glooctl check`
  - set `gateway.persistProxySpec` to `true` in the system settings resource
  - Use the newly built binary and confirm that an error with the proxy is still reported
    - `./_output/glooctl-darwin-amd64 check`
  - Use the system glooctl install, and confirm that an error with the proxy is reported, and that the output matches that of the newly built binary exactly
    - `glooctl check`


projects/gloo/cli/pkg/common/get.go
```diff
@@ -118,10 +118,7 @@ func GetUpstreamGroups(name string, opts *options.Options) (gloov1.UpstreamGroup
 
 func GetProxies(name string, opts *options.Options) (gloov1.ProxyList, error) {
 	settingsClient := helpers.MustNamespacedSettingsClient(opts.Top.Ctx, opts.Metadata.GetNamespace())
-	settings, err := settingsClient.Read(opts.Metadata.GetNamespace(), defaults.SettingsName, clients.ReadOpts{Ctx: opts.Top.Ctx})
-	if err != nil {
-		return nil, err
-	}
+	settings, _ := settingsClient.Read(opts.Metadata.GetNamespace(), defaults.SettingsName, clients.ReadOpts{Ctx: opts.Top.Ctx})
```
jenshu, line 121: i don't think the error check should be removed?

elcasteel (commented):
+1 to Jenny's comment but otherwise looks good

projects/gloo/cli/pkg/common/get.go
ben-taussig-solo: Yeah, I'm not quite sure how to handle this -- the issue here is that we look for a settings resource in every namespace we scan. If there isn't a settings resource present, then these errors pollute the `glooctl check` output. 

```
❯ ./_output/glooctl-darwin-amd64 check -n gloo-system
Warning: Could not determine gloo server versions (is Gloo running outside of kubernetes?): Tag v1.15.0-beta6-6-g7bb095026-dirty is not a valid semver version, must be of the form vX.Y.Z[-<label>#]
Checking deployments... OK
Checking pods... OK
Checking upstreams... OK
Checking auth configs... OK
Checking RouteOptions... OK
Checking secrets... OK
Checking gateways... OK
Checking proxies... 5 Errors!
Error: 5 errors occurred:
        * default.default does not exist: settings.gloo.solo.io "default" not found
        * kube-node-lease.default does not exist: settings.gloo.solo.io "default" not found
        * kube-public.default does not exist: settings.gloo.solo.io "default" not found
        * kube-system.default does not exist: settings.gloo.solo.io "default" not found
        * local-path-storage.default does not exist: settings.gloo.solo.io "default" not found
```

I think I may need to update the function that calls this one to be more careful about choosing when to call it

elcasteel: As a first pass, we can ignore that error in `checkProxies` and then later consider whether we should do something to only look at namespaces that we expect to have proxies in.

ben-taussig-solo: This commit adds a new helper, `getProxiesFromSettings`, which uses the global settings object retrieved earlier in `glooctl check`, instead of attempting to retrieve the settings object every time we look for a proxy: https://github.com/solo-io/gloo/pull/8228/commits/7010de60277bc079998847e95bcdb6ca05e4f950

elcasteel (dismissed):

ben-taussig-solo:
/kick CI build error, seems to have not started properly

elcasteel (dismissed):

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -864,13 +856,20 @@ func checkProxies(ctx context.Context, opts *options.Options, namespaces []strin
...
+	glooNamespaces := []string{}
+	// iterate over deployments, and add the namespaces of all gloo deployments to the glooNamespaces slice
+	for _, deployment := range deployments.Items {
+		if deployment.Labels["gloo"] == "gloo" {
+			glooNamespaces = append(glooNamespaces, deployment.Namespace)
 		}
-		proxies, err := proxyClient.List(ns, clients.ListOpts{})
+	}
+
+	for _, ns := range glooNamespaces {
```
sam-heilbron, line 868: Why is this change necessary? We pass in namespaces to the function (and relied on that before). Should we be using that value  (and if not, should we be removing it)

projects/gloo/cli/pkg/common/get.go
```diff
@@ -128,6 +135,15 @@ func GetProxies(name string, opts *options.Options) (gloov1.ProxyList, error) {
 	}
 	return getProxiesFromK8s(name, opts)
 }
+
+func GetProxiesFromSettings(name string, opts *options.Options, settings *gloov1.Settings) (gloov1.ProxyList, error) {
```
sam-heilbron, line 139: Love this! Since it's exported, can be add a docstring

projects/gloo/cli/pkg/common/get.go
```diff
@@ -116,9 +116,16 @@ func GetUpstreamGroups(name string, opts *options.Options) (gloov1.UpstreamGroup
 	return list, nil
 }
 
+func GetSettings(ctx context.Context, opts *options.Options) (*gloov1.Settings, error) {
```
sam-heilbron, line 119: Right now we pass in a full options object, but only seem to use the namespace from it. Is there a reason we don't just pass in the namespace?

jbohanon:
It's a little confusing to me that it is the VirtualService which we modified to be invalid, but it is the proxy which pops in `glooctl check`. Is it just prerequisite knowledge to operators of the software that a proxy is derived from a virtual service?

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -864,13 +856,20 @@ func checkProxies(ctx context.Context, opts *options.Options, namespaces []strin
...
 	}
 	var multiErr *multierror.Error
-	for _, ns := range namespaces {
-		proxyClient, err := helpers.ProxyClient(ctx, []string{ns})
-		if err != nil {
-			multiErr = multierror.Append(multiErr, err)
-			continue
+
+	glooNamespaces := []string{}
+	// iterate over deployments, and add the namespaces of all gloo deployments to the glooNamespaces slice
```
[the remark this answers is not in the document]
ben-taussig-solo: Updated this here: https://github.com/solo-io/gloo/pull/8228/commits/5aa81720762ad1f821c200bbf002f2798771cf48

projects/gloo/cli/pkg/common/get.go
ben-taussig-solo: After reviewing the rest of the helpers in this file, I think we should probably be passing in _just_ an options object, and deriving the context from that. The pattern in this file seems to be to pass around options objects to all helpers, so I think this might be more useful if we keep it that way, but it does seem like an anti-pattern to not be getting the context form that

ben-taussig-solo: Updated here: 20ec6fcd1fe1c15109868d38d1ca03ed766c9412

Let me know what you think!

projects/gloo/cli/pkg/common/get.go
ben-taussig-solo:
```go
// Retrieve proxies from the proxy debug endpoint, or from kubernetes if the proxy debug endpoint is not available
// Takes in a settings object to determine whether the proxy debug endpoint is available
func GetProxiesFromSettings(name string, opts *options.Options, settings *gloov1.Settings) (gloov1.ProxyList, error) {
```

elcasteel (dismissed):

projects/gloo/cli/pkg/common/get.go
sam-heilbron: Looking good! Can we use the recommendation from https://go.dev/blog/godoc? 

```suggestion go
// GetProxiesFromSettings retrieves proxies from the proxy debug endpoint, or from kubernetes if the proxy debug endpoint is not available
// Takes in a settings object to determine whether the proxy debug endpoint is available
func GetProxiesFromSettings(name string, opts *options.Options, settings *gloov1.Settings) (gloov1.ProxyList, error) {
```

ben-taussig-solo: Updated here: a1190b83cdc96db9bfe56e1fa132c18c128183e5

projects/gloo/cli/pkg/cmd/check/root.go
```diff
@@ -864,13 +856,22 @@ func checkProxies(ctx context.Context, opts *options.Options, namespaces []strin
...
-		if err != nil {
-			multiErr = multierror.Append(multiErr, err)
-			continue
+
+	glooNamespaces := []string{}
+	// identify namespaces where gloo is deployed (i.e., where the deployment has the label "gloo=gloo")
+	// the gloo deployment contains the endpoint we use to obtain proxy resources, so we only need to check
+	// for proxies in namespaces where gloo is deployed
+	for _, deployment := range deployments.Items {
+		if deployment.Labels["gloo"] == "gloo" {
```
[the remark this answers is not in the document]
ben-taussig-solo: Made the change we discussed on a call here: ba40c4788f4e12f5607b4fc666cb94fbb8112007

Let me know what you think!

ben-taussig-solo:
Updated the helper we use to get proxy resources to be easier to reason about here: 72fba99200b8d8ea0712d48133253d42c88975a1

projects/gloo/cli/pkg/cmd/check/root.go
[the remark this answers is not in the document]
ben-taussig-solo: This block has been removed, so this thread is no longer relevant -- closing it now

elcasteel (approved):

sam-heilbron (approved):

PR #8229, opened 2023-05-16
refactor(CI): use common base action

gunnar-solo:
# Description
created and consumed a new composite action:  `prep-go-runner.yaml`.  It...
* cancels previous actions
* frees disk space
* sets up go

Additional Changes:
* minor action renaming (don't have 3 actions called CI)
* removed `DO_NOT_SUBMIT` action
* moved `do-not-submit.sh` to `codegen.yaml` workflow
* moved composite actions from `.github/actions` to `.github/workflows/composite-actions`
* removed testspace server uploads

Considered (but reject for now) Changes:
* making each step of `prep-go-runner.yaml` optional, with a bounding boolean input variable.  Coupled with reasonable defaults, the thought was that you could call the setup with exactly what you want.  Opted for simplicity, however

# Context
As I sat down to create _yet another_ github action with the same leading 4 steps, I realized we may want a composite action to use as a helper

# Checklist:


.github/workflows/push-solo-apis-branch.yaml
```diff
@@ -55,42 +55,12 @@ jobs:
...
       - name: Install SSH Deploy key
         uses: webfactory/ssh-agent@v0.4.1
         with:
           ssh-private-key: ${{ secrets.SOLO_API_DEPLOY_KEY }}
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
-        with:
-          repository: solo-io/gloo
-          path: gloo
-          ref: ${{ env.SOURCE_COMMIT }}
```
nfuden, line 88: Do we no longer need this?

.github/workflows/codegen.yaml
```diff
@@ -1,40 +1,12 @@
...
-          sudo rm -rf /opt/ghc
-          sudo rm -rf /usr/local/share/boost
-          sudo rm -rf $AGENT_TOOLSDIRECTORY
-
-          echo "After clearing disk space:"
-          df -h
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
```
nfuden, line 32: do we no longer need this option?

.github/workflows/do-not-submit.yaml
jenshu, line 1: is this no longer needed? should we delete the `ci/do-not-submit.sh` script too?

.github/workflows/codegen.yaml
gunnar-solo: this _particular_ option is counterproductive, in my opinion.  From the [checkout action docs](https://github.com/actions/checkout):
> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer [here](https://help.github.com/en/articles/events-that-trigger-workflows) to learn which commit $GITHUB_SHA points to for different events.

.github/workflows/do-not-submit.yaml
gunnar-solo: I'm not totally sure--I don't recall ever using it, personally, and it's been spinning compute for a long time.  I figured this was a good opportunity for removing it, or at least discussing it

my opinion (given the state of the PR) is that it's unneeded, and a bit wasteful on compute

assuming I get buy-in, I'd be happy to remove the associated bash script

jenshu (dismissed):

jackstine (commented):
LGTM

.github/actions/common/action.yaml
```diff
@@ -0,0 +1,32 @@
...
+      # Clean up pre-installed tools
+      # https://github.com/actions/virtual-environments/issues/1918
+      sudo rm -rf /usr/share/dotnet
+      sudo rm -rf /opt/ghc
+      sudo rm -rf /usr/local/share/boost
+      sudo rm -rf $AGENT_TOOLSDIRECTORY
+
+      echo "After clearing disk space:"
+      df -h
+  - name: Set up Go
```
sam-heilbron, line 28: I'm not convinced this is necessary. As it is, setup-go is very compact step. In my view, moving this into a sub-action obfuscates what we're doing, more than it simplifies it. I'd be happy to leave this in each job that needs it

.github/actions/common/action.yaml
```diff
@@ -0,0 +1,32 @@
+name: Gloo Common Actions
```
sam-heilbron, line 1: nit: I'd rather name this something explicit that refresent what this does. Clean Disk Space seems like it describes what this does. This way, we can easily compose actions that we need, rather than having a big "common" action, whose responsibilities are unclear/need to be read

sam-heilbron (commented):
I like the change! I see we're removing "DO NOT SUBMIT", can you outline why we are doing this? This was intended to be a way to allow devs to commit code that would prevent a PR from merging. Is there an alternative flow you would expect devs to take?

.github/actions/common/action.yaml
gunnar-solo: I'm thinking along the lines of all of parallel edits that need to be made to this block whenever we
* update our go version
* update our used action

Having this in one spot is my preference.  The above ^ is nice, but really, I have the opposite base opinion of yourself:  that it is cleaner/simpler to have a single common step that manages this

.github/actions/common/action.yaml
gunnar-solo: totally happy to meaningfully rename, based on the results of other discussion

gunnar-solo:
> I see we're removing "DO NOT SUBMIT", can you outline why we are doing this? This was intended to be a way to allow devs to commit code that would prevent a PR from merging. Is there an alternative flow you would expect devs to take?
I've opened a [slack discussion](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1684268068865899) around this to see if the action is in active use.  My current understanding is that it's not

I *could* totally be wrong, though.  I figure it's easiest to ask the Edge Team.
---
alternate flow:  add a "Work In Progress" label

sam-heilbron (dismissed):

sam-heilbron (approved):

bewebi (approved):
LGTM

PR #8233, opened 2023-05-16
Stability: Demonstrate How To Run Subset of Tests

sam-heilbron:
# Description

Create a mechanism to skip tests on PRs but run them on nightlies

# Context
## Label Selection
We are using https://onsi.github.io/ginkgo/#spec-labels as a way to easily select a subset of tests

Both @gunnar-solo and @bewebi provided feedback on this solution and are happy with it.

Note, that I've added references to nightly tests, but we do not yet have those running. That should be introduced in the coming week so I figured I would leave the references in there so it's clear how this can be used.

## Split E2E Tests Up
We have end to end tests which store configuration in: memory, vault/consul and kubernetes. Previously we combined the in-memory and consul/vault tests together. I opted to split them into different targets. The idea here is we can run those consul/vault tests in parallel (save time).

## Rename Regression Tests
Provide some consistency to our make targets, by renaming `run-ci-regression-tests` to `run-kube-e2e-tests`

# Checklist:


Makefile
```diff
@@ -250,16 +250,22 @@ test-with-coverage: test
 
 .PHONY: run-tests
 run-tests: GINKGO_FLAGS += -skip-package=e2e ## Run all non E2E tests, or only run the test package at {TEST_PKG} if it is specified
```
gunnar-solo, line 252: is this still needed?

Makefile
```diff
@@ -250,16 +250,22 @@ test-with-coverage: test
...
+run-tests: GINKGO_FLAGS += --label-filter="!end-to-end && !nightly"
 run-tests: test
 
 .PHONY: run-e2e-tests
-run-e2e-tests: TEST_PKG = ./test/e2e/ ./test/consulvaulte2e ## Run all E2E tests
+run-e2e-tests: TEST_PKG = ./test/e2e/ ## Run all In Memory E2E tests
```
bewebi, line 257:
```suggestion
run-e2e-tests: TEST_PKG = ./test/e2e/ ## Run all in-memory E2E tests
```

bewebi (commented):
One grammar nit

Also, to be clear, are we intending to remove the intentionally failing test immediately once we confirm that it is run?

Makefile
sam-heilbron: Technically no, because the labels will filter it out. However I think it speeds up the test by not having ginkgo compile the tests in those directories

gunnar-solo (approved):

PR #8235, opened 2023-05-16
K8s tests in kind

bewebi:
# Description

- Move tests which rely on k8s to the kube2e test package, which uses KinD
  - Remove `ShouldRunKubesTests()` checks
  - Ensure all tests use actual kubes client
  - Add relevant setup/breakdown steps to kube2e's test suite file
  - Include comment referring to package each file is meant to test
  - Cleanup tests, removing code that was tied to the persistent cluster paradigm
- Remove `RUN_KUBES_TESTS` from env.go test file, `Kuberentes()` from requirements.go test file
- Move kubernetes happy_path test to kube2e
- Remove kube client from aws e2e tests which do not appear to require it 

# Context

This is the first step towards migrating off of build bot for testing

We will stop setting `RUN_KUBES_TESTS` in a followup PR, once it's clear that no tests are breaking

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/4907


test/kube2e/gloo/gloo_suite_test.go
```diff
@@ -41,11 +47,38 @@ var (
...
+
+	apiExts apiext.Interface
+	locker  *clusterlock.TestClusterLocker
 )
 
-var _ = BeforeSuite(func() {
+var _ = SynchronizedBeforeSuite(beforeSuiteOne, beforeSuiteAll)
+var _ = SynchronizedAfterSuite(afterSuiteOne, afterSuiteAll)
+
+func beforeSuiteOne() []byte {
```
sam-heilbron, line 58: I don't think we need to do this. In the previous model, when using a long-lived cluster, we registered the CRD before/after this test. In this new model, we install gloo via helm, which in term applies the CRDs that we need. I think we can just remove this new setup/teardown code and I would expect it to work

test/kube2e/gloo/kubernetes_test.go
```diff
@@ -0,0 +1,203 @@
...
+	kubev1 "k8s.io/api/core/v1"
+
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/kubernetes"
+
+	// From https://github.com/kubernetes/client-go/blob/53c7adfd0294caa142d961e1f780f74081d5b15f/examples/out-of-cluster-client-configuration/main.go#L31
+	// Uncomment the following line to load the gcp plugin (only required to authenticate against GKE clusters).
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
sam-heilbron, line 28: In most other places I think we've cleaned this up. Since these tests don't interact with GCP I think we should remove this import as well

test/kube2e/gloo/utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	. "github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// Kubernetes tests for utils from projects/gloo/pkg/bootstrap
+var _ = Describe("Utils", func() {
```
sam-heilbron, line 26: I had read https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common a while back and found it interesting. Could we update this to reflect the tests included? This is testing the ArtifactClient used in our Bootstrap code...I could see this even living in the resource_client_test file

test/kube2e/gloo/utils_test.go
```diff
@@ -0,0 +1,101 @@
...
+				&kube,
+				&kubeCoreCache,
+				&api.Client{},
+				"artifacts")
+			Expect(err).NotTo(HaveOccurred())
+			artifactClient, err = v1.NewArtifactClient(ctx, factory)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should work with artifacts", func() {
```
sam-heilbron, line 94: What this is testing is that we can define a ConfigMap in a kube cluster, and then our ArtifactClient (generic term for artifact) can read that resource. I think we likely have a test already which confirms we can apply a ConfigMap and then have a GlooResource reference that ConfigMap (implying the underlying client could query it properly). We may be able to just delete this test if we can find a test that replicates this behavior (I'm happy to look further if you want)

test/testutils/env.go
```diff
@@ -21,10 +21,6 @@ const (
 	// are not met. See ValidateRequirementsAndNotifyGinkgo for a detail of available behaviors
 	InvalidTestReqsEnvVar = "INVALID_TEST_REQS"
 
-	// RunKubeTests is used to enable any tests which depend on Kubernetes. NOTE: Kubernetes back tests should
-	// be written into the kube2e suites, and those don't require this guard.
-	RunKubeTests = "RUN_KUBE_TESTS"
```
sam-heilbron, line 26: 👏

sam-heilbron: In addition to removing this, let's remove the place where we set it when we run tests: https://github.com/solo-io/gloo/blob/main/ci/cloudbuild/run-tests.yaml#L132

sam-heilbron (commented):
Looking good! A few comments and we'll need to address:
```
Error: ../../testutils/requirements.go:171:13: undefined: RunKubeTests
```

test/kube2e/gloo/gloo_suite_test.go
sam-heilbron: In turn not in term *. I'm hoping that was auto correct but could very have just been my brain

test/kube2e/gloo/gloo_suite_test.go
```diff
@@ -41,11 +43,19 @@ var (
...
+
+	locker *clusterlock.TestClusterLocker
+
+	envoyFactory *services.EnvoyFactory
 )
 
 var _ = BeforeSuite(func() {
-	ctx, cancel = context.WithCancel(context.Background())
 	var err error
+	locker, err = clusterlock.NewTestClusterLocker(kube2e.MustKubeClient(), clusterlock.Options{})
```
sam-heilbron, line 54: I don't think we should need a clusterLocker. This is intended to synchronize access to a shared cluster, which we shouldn't have to do since our kube cluster is short lived (ie a kind cluster spun up just for this test)

test/kube2e/gloo/kubernetes_test.go
```diff
@@ -0,0 +1,199 @@
...
+		ctx, cancel = context.WithCancel(context.Background())
+		ctx = settingsutil.WithSettings(ctx, &v1.Settings{})
+
+		svcNamespace = helpers.RandString(8)
+
+		cfg, err := kubeutils.GetConfig("", "")
+		Expect(err).NotTo(HaveOccurred())
+
+		kubeClient, err = kubernetes.NewForConfig(cfg)
+		kubeCoreCache, err = kubecache.NewKubeCoreCache(context.TODO(), kubeClient)
```
sam-heilbron, line 59: Let's use the ctx instead of context.TODO

test/kube2e/gloo/kubernetes_test.go
```diff
@@ -0,0 +1,199 @@
...
+			}},
+		}, metav1.UpdateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		err := kubeClient.CoreV1().Namespaces().Delete(ctx, svcNamespace, metav1.DeleteOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		if cancel != nil {
```
sam-heilbron, line 147: I don't think cancel is ever nil, so we can just call cancel directly, without the guard

test/kube2e/gloo/artifact_client_test.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+		}, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		kubeCoreCache, err = corecache.NewKubeCoreCacheWithOptions(ctx, kube, time.Hour, []string{namespace})
+		Expect(err).NotTo(HaveOccurred())
+
+	})
+	AfterEach(func() {
+		cancel()
+		setup.TeardownKube(namespace)
```
sam-heilbron, line 58: What does this command do? I see that it's imported from solo-kit test/setup package, and I'd like to avoid that if possible. My fear is that if we need to change it, we are dependent on another library, which just means updating tests takes longer. As it is, a lot of the test utilities in solo-kit are out of date, and we've either replicated the functionality in a gloo test util, or found another way to solve the problem

test/e2e/aws_test.go
```diff
@@ -763,7 +752,7 @@ var _ = Describe("AWS Lambda", func() {
 				WhatToRun: services.What{
 					DisableGateway: justGloo,
 				},
-				KubeClient: kube2e.MustKubeClient(),
+				//KubeClient: kube2e.MustKubeClient(),
```
sam-heilbron, line 755: nit: delete, if safe (I figure this was the easiest way to test it, which makes sense...just wanted to make sure it wasn't lost)

sam-heilbron (commented):
Awesome work! Not sure if it's ready for review, but figured I'd share a few other thoughts just to get ahead of it. If everything is passing, then this LGTM

test/kube2e/gloo/artifact_client_test.go
bewebi: It's deprecated...I did replace it in another test with an actual Delete call using a Namespace client, will do the same here

test/kube2e/gloo/artifact_client_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	Context("artifacts as config maps", func() {
+
+		var (
+			artifactClient v1.ArtifactClient
+		)
+
+		BeforeEach(func() {
+			_, err := kube.CoreV1().ConfigMaps(namespace).Create(ctx,
```
sam-heilbron, line 69: This creates a configMap in the beforeEach, but doesn't delete it in the afterEach. Can we add that cleanup step?

test/kube2e/gloo/gloo_suite_test.go
```diff
@@ -62,13 +65,17 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 
 	snapshotWriter = helpers.NewSnapshotWriter(resourceClientset, []retry.Option{})
+
+	envoyFactory, err = services.NewEnvoyFactory()
```
sam-heilbron, line 69: Which envoy is this using? Does this use an envoy binary or does it run envoy in a docker container?

test/kube2e/gloo/happypath_test.go
```diff
@@ -0,0 +1,414 @@
...
+					svc = nil
+					cfg, err = kubeutils.GetConfig("", "")
+					Expect(err).NotTo(HaveOccurred())
+					kubeClient, err = kubernetes.NewForConfig(cfg)
+					Expect(err).NotTo(HaveOccurred())
+				})
+
+				AfterEach(func() {
+					if namespace != "" {
+						_ = setup.TeardownKube(namespace)
```
sam-heilbron, line 113: This is solo-kit Teardown that I'd like us to move away from. What is this doing/is it necssary? Can we re-define it in Edge if it's necessary?

test/kube2e/gloo/happypath_test.go
```diff
@@ -0,0 +1,414 @@
...
+							return u, nil
+						}
+					}
+					return nil, fmt.Errorf("not found")
+				}
+
+				Context("specific namespace", func() {
+
+					BeforeEach(func() {
+						prepNamespace()
```
sam-heilbron, line 179: Prep namespace creates resources, but we don't have an AfterEach that cleans them up. Can you add that?

sam-heilbron (commented):
Looking good! Some thoughts just around resource cleanup to ensure we don't potentially cause flakes (I dont think we will, but better to be consistent)

test/kube2e/gloo/artifact_client_test.go
bewebi: The entire namespace is deleted, so that step is not necessary

test/kube2e/gloo/gloo_suite_test.go
bewebi: In CI we are using an image specified via the `ENVOY_GLOO_IMAGE` var, set to 1.25.6-patch1 in the Makefile [here](https://github.com/solo-io/gloo/blob/20ef52e64e654f2c85412ca8cfbcca8ad881895a/Makefile#L96)

test/kube2e/gloo/happypath_test.go
bewebi: As with `artifact_client_test.go`, the entire namespace is deleted, so a step for each resource is not necessary

sam-heilbron (approved):
Awesome!

gunnar-solo (approved):

PR #8236, opened 2023-05-17
[DNM] Multiple secret sources

jbohanon:
# Description

We want to support configuration and use of multiple secrets resource clients. 

After discussing with members of the Edge team, it is deemed unnecessary complexity to include the logic for client selection at this time. This is based on the assumption (to be tested) that we do not use the client interface methods which deal with single secrets, rather we are exclusively listing and watching. This provides the benefit that customers can handle the issue of namespacing their secrets to avoid collisions in whatever way makes the most sense for their organization. Because of this, the behavior of Gloo Edge is undefined when a secret with the same name and namespace exists in multiple source locations. This will need to be clearly communicated in user guide and release/upgrade docs.

# Context

This is needed to facilitate the Vault PKI implementation.

This will also be particularly useful when connection details are required for authenticating to a secret client. We can, for example, define Vault TLS and auth configuration as Kubernetes secrets, then use these to initiate the resource client for Vault.

# Checklist:


projects/gloo/pkg/bootstrap/clients/clients_suite_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/solo-io/gloo/test/services"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestBootstrap(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Bootstrap Suite")
+}
```
jbohanon, line 16:
```suggestion
func TestClients(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Bootstrap Clients Suite")
}
```

projects/gloo/pkg/bootstrap/clients/artifact.go
```diff
@@ -0,0 +1,62 @@
...
+	kubeconverters "github.com/solo-io/gloo/projects/gloo/pkg/api/converters/kube"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/factory"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/cache"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/memory"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// sharedCach OR resourceCrd+cfg must be non-nil
```
jbohanon, line 18: This comment seems to be from before a refactor. I need to look into where `resourceCrd` ended up and what really needs to be non-nil here.

Also: `sharedCache`

projects/gloo/pkg/bootstrap/clients/config_factory.go
```diff
@@ -0,0 +1,116 @@
...
+			kubeCache: cache,
+			restCfg:   cfg,
+		},
+		consul: configFactoryParamsConsul{
+			consulClient: consulClient,
+		},
+	}
+}
+
+// sharedCache, resourceCrd+cfg OR consulClient must be non-nil
```
jbohanon, line 57: This elucidates the required non-nils. I think it's

```suggestion
// params.memory.sharedCache, resourceCrd+params.kube.restCfg OR params.consul.consulClient must be non-nil
```

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -0,0 +1,295 @@
...
+			secretOpts.SecretSources = []*v1.Settings_SecretOptions_Source{newApiSource}
+		}
+	}
+	if secretOpts != nil {
+		return NewMultiResourceClientFactory(secretOpts.GetSecretSources(), sharedCache, cfg, clientset, kubeCoreCache, vaultClient)
+	}
+	return nil, errors.Errorf("invalid config source type")
+}
+
+type MultiResourceClientFactory struct {
```
jbohanon, line 86: Since we're not in a `secret` package, this should be renamed to something like `MultiSecretResourceClientFactory`

same with all references in this file to the factory or the client itself

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -0,0 +1,295 @@
...
+func NewMultiResourceClientFactory(secretSources []*v1.Settings_SecretOptions_Source,
+	sharedCache memory.InMemoryResourceCache,
+	cfg **rest.Config,
+	clientset *kubernetes.Interface,
+	kubeCoreCache *cache.KubeCoreCache,
+	vaultClient *vaultapi.Client) (*MultiResourceClientFactory, error) {
+
+	// Guard against nil source slice
+	if secretSources == nil {
+		return nil, ErrNilSourceSlice
```
jbohanon, line 115: I need to find out what the behavior of returning this error is and if we should attempt to use a default secret source. I think if we crashloop the pod on this error, that's pretty bad. I don't know what default we could safely presume to use though or if that would hide the issue from the user

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -0,0 +1,295 @@
...
+		switch concreteClient := v.(type) {
+		case *vault.ResourceClient:
+			continue
+		case *file.ResourceClient:
+			continue
+		default:
+			return concreteClient.Register()
+		}
+	}
+	return errors.New("Register method only implemented on Kubernetes resource client")
```
jbohanon, line 231: Do we need to return an error here? Not sure from where if anywhere this interface function is called

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -0,0 +1,295 @@
...
+	return list, nil
+}
+
+func (m *MultiResourceClient) ApplyStatus(statusClient resources.StatusClient, inputResource resources.InputResource, opts clients.ApplyStatusOpts) (resources.Resource, error) {
+	errMsg := "ApplyStatus not implemented on multiSecretSourceResourceClient"
+	contextutils.LoggerFrom(opts.Ctx).DPanic(errMsg)
+	return nil, errors.New(errMsg)
+}
+
+func (m *MultiResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-chan resources.ResourceList, <-chan error, error) {
```
jbohanon, line 271: This needs thorough testing/review to make sure I didn't screw up the chan/goroutine implementation

projects/gloo/pkg/bootstrap/clients/secret_test.go
jbohanon, line 1: move this to `clients_test.go`

projects/gloo/pkg/bootstrap/consul.go
jbohanon, line 1: delete file

projects/gloo/pkg/bootstrap/utils.go
jbohanon, line 1: delete file

projects/gloo/pkg/bootstrap/utils_test.go
jbohanon, line 1: delete file

projects/gloo/pkg/bootstrap/clients/clients_test.go
```diff
@@ -0,0 +1,227 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(cfg).ToNot(BeNil())
+		Expect(kubefactory.(*factory.KubeResourceClientFactory).Cfg).To(Equal(cfg))
+
+		Expect(cfg.QPS).To(Equal(float32(100.5)))
+		Expect(cfg.Burst).To(Equal(1000))
+	})
+
+	Context("kube tests", func() {
+		BeforeEach(func() {
```
sam-heilbron, line 65: Heads up we're removing this guard in some in flight work: https://github.com/solo-io/gloo/pull/8235

PR #8243, opened 2023-05-17
[1.14.x] Fix non-persisted proxy validation in glooctl check

ben-taussig-solo:
# Description
 - Backport #8228 to 1.14.x
 - Updates `glooctl check` to use the GPRC endpoint for proxy validation, ala `glooctl get proxies`.
 - This fixes an issue where the check would report false-positives for invalid proxies that were not persisted.
   - Prior to this change, glooctl check used a k8s client to obtain proxy resources
   - Since 1.12.x, proxy resources are not persisted by default, so this k8s client failed to obtain proxies, and in turn failed to report errors on proxy resources


sam-heilbron (approved):

nfuden (approved):
interesting gap! nice find!

PR #8251, opened 2023-05-18
build(release): move to GHA from cloudbuild

gunnar-solo:
# Description

# Context

# Checklist:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,62 @@
+name: Release
```
sam-heilbron, line 1: Given that we are "publishing assets (or artifacts)", I would prefer that we name it something in line with that. The challenge I see with release, is that we publish assets on a release, but also on pull requests, and soon on nightly builds. I fear that having to use the term "release" will confuse people about the context in which we're publishing

.github/workflows/release.yaml
```diff
@@ -0,0 +1,62 @@
...
+        if: ${{ github.event_name == 'pull_request' }}
+        run: echo "PR_RELEASE=true" >> $GITHUB_ENV
+      - name: Set RELEASE env var
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: echo "RELEASE=true" >> $GITHUB_ENV
+      - run: make publish-docker                # ON RELEASE and PR_RELEASE, push docker images to quay.io/solo-io
+      - run: make fetch-package-and-save-helm   # on RELEASE, push helm chart to gs://solo-public-helm.  on PR_RELEASE, push helm chart to gs://solo-public-tagged-helm
+      - run: make upload-github-release-assets  # on RELEASE, push CLI assets
+      - run: make push-chart-to-registry        # on RELEASE, push helm chart to oci://gcr.io/solo-public/gloo-helm
+      - run: make publish-docker-retag          # on RELEASE, push docker images to gcr.io/gloo-edge
```
sam-heilbron, line 50: Perhaps this is just a nit, but I think we could split the publishing into separate jobs: docker, helm, cli

Makefile
```diff
@@ -44,85 +44,19 @@ IMAGE_REGISTRY ?= quay.io/solo-io
...
+VERSION 	?= 0.0.1-dev    # a basic version that can be overidden
 
-# CREATE_TEST_ASSETS allows us to create assets on PRs that are unique
-# This flag will set the version to be PR-unique rather than commit-unique for charts and images
-CREATE_TEST_ASSETS := "false"
-ifneq ($(TEST_ASSET_ID),)
-	CREATE_TEST_ASSETS := "true"
-endif
-
-# ensure we have a valid version from a forked repo, so community users can submit PRs
```
sam-heilbron, line 59: Why are we safe to remove this? Don't we still need to support identifying when a PR is running from a fork?

Makefile
```diff
@@ -44,85 +44,19 @@ IMAGE_REGISTRY ?= quay.io/solo-io
 z := $(shell mkdir -p $(OUTPUT_DIR))
 
 SOURCES := $(shell find . -name "*.go" | grep -v test.go)
-RELEASE := "false"
 
-# CREATE_ASSETS is used to protect certain make targets which publish assets and are used for releases
-CREATE_ASSETS := "true"
+# Only one of RELEASE or PR_RELEASE should be set to true
+RELEASE 	?= false        # if true, build all release artifacts
```
sam-heilbron, line 49: Since there is an inherent relationship between "release" and "pr_release", what do you think about consolidating them? I was imagining we could have something like "PUBLISH_CONTEXT" and it can have values of "RELEASE","PULL_REQUEST", "NIGHTLY_BUILD". Then, we can use the value to identify what locations assets should be published to (ie releases publish their helm charts to X but PRs publish them elsewhere)

Makefile
```diff
@@ -149,9 +83,6 @@ GO_BUILD_FLAGS := GO111MODULE=on CGO_ENABLED=0 GOARCH=$(GOARCH)
 GOLANG_ALPINE_IMAGE_NAME = golang:$(shell go version | egrep -o '([0-9]+\.[0-9]+)')-alpine
 
 # Passed by cloudbuild
-GCLOUD_PROJECT_ID := $(GCLOUD_PROJECT_ID)
```
sam-heilbron, line 152: Nice! Let's try to cleanup any stray comments that reference cloudbuild as well

Makefile
```diff
@@ -660,13 +588,15 @@ package-chart: generate-helm-files
 	helm repo index $(HELM_SYNC_DIR)
 
 .PHONY: push-chart-to-registry
+ifeq (true, $(RELEASE))
```
sam-heilbron, line 591: I think we can consolidate these operations even more: push-chart-to-registyr still calls: generate-helm-files and helm-package. I think we want to make it very simple for developers to have make targets which they can call, and the release/publish targets should be named to demonstrate that they are intended only to be used by releases

Makefile
```diff
@@ -679,44 +609,6 @@ ifeq ($(CREATE_ASSETS), "true")
 	done
 endif
 
-.PHONY: render-manifests
```
sam-heilbron, line 682: Let's just double check with the team about these values, but it would be great to clean these up!

docs/Makefile
```diff
@@ -7,11 +7,6 @@
 # build hugo v0.69.2:
 # https://github.com/gohugoio/hugo must be built from commit ec9dcf304849ded11cea70ab9ad46e0a0a4615b2 and installed on $PATH
 
-
-
-# remove the "v" prefix
-VERSION ?= $(shell echo $(TAGGED_VERSION) | cut -c 2-)
```
sam-heilbron, line 13: How will the version be set for our docs?

sam-heilbron (commented):
This is a great start! A few thoughts:
1. I think any targets that are run by CI and publish assets, we should put in a section of our Makefile, or name them to represent that they are run on releases
2. I'd like to avoid conditional logic around a bunch of make targets. Ideally everything can just be run locally, and the default location where assets are published is just your local file. If that's not possible, I think only the true "release" targets should have conditionals around them
3. I think we want to make it super clear what variables are available, and which are set by devs vs which are determined by other values. We have an opportunity here to improve our Makefile readability/usability

gunnar-solo:
closing in favor of https://github.com/solo-io/gloo/pull/8278

currently open conversations have been ported

PR #3171, opened 2020-06-17
API for Access Token Validation in OAuth/OIDC

kdorosh:
# Description

API for Gloo to handle other parts of the [OAuth2.0](https://tools.ietf.org/html/rfc6749), [OIDC](https://openid.net/specs/openid-connect-core-1_0.html), [token introspection](https://tools.ietf.org/html/rfc7662), and [bearer token](https://tools.ietf.org/rfc/rfc6750.txt) specs in our external auth server implementation. Currently we support an OIDC authorization code flow, but users have requested that Gloo handle other parts of the OAuth2.0 spec as well. In this iteration, we are adding access token validation via token introspection and making the userinfo response available to users in extauth go plugins.

# Context

This PR deprecates the old OAuth config as it really only supported OIDC authorization code flows and left little room to cleanly implement other parts of the specs (both OAuth2.0 and OIDC).

The new API allows us to easily add more parts of the spec incrementally (such as `OidcImplicit` and `OidcHybrid` flows, as well as other ways to validate the access token, such as `Jwt` or `Hmac`) without making awkward additions to the previous message that would require setting one field completely obsolesce large portions of the other fields.

# Checklist:

- [x] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [x] If I updated APIs (our protos) or helm values, I ran `make update-deps generated-code` to ensure there will be no code diff
- [x] I followed guidelines laid out in the Gloo [contribution guide](https://docs.solo.io/gloo/latest/contributing/)
- [x] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #8257, opened 2023-05-19
Oauth2 ClaimsToMetadata API

jmhbh:
# Description
- This API change on the `OidcAuthorizationCode` and `AccessTokenValidation` will allow us to proceed with implementation in the `ext-auth-service` to support extracting claims from either the `IdToken` or `AccessToken` for the `OidcAuthorizationCode` flow or from the `AccessToken` for the `AccessTokenValidation` flow. 

# Context:
- Currently, in Gloo Platform, we cannot pair an ext-auth policy with `AccessTokenValidation` configured with a rate limit policy, and in order to configure a rate limit and ext auth policy for the OIDC flow we require an additional JWT policy to extract the claims from the id/accesstoken to headers. 
- In order to add support for configuring a rate limit policy in Gloo Platform with an ext-auth policy that uses any OAuth2 flow. We landed on an approach in this [design doc](https://soloio.slab.com/posts/portal-support-o-auth-security-for-usage-plans-ykvns72m) to enrich the current API with a `ClaimsToMetadata` API that would allow us to implement changes in the ext-auth-service to extract the claims from a token and emit it as dynamic metadata to be consumed by the rate limit filter.
- The idea is that a user configuring an ext auth policy using OIDC with this new API and would like claims to be extracted and emitted as metadata from the idToken would do the following:
```
apiVersion: security.policy.gloo.solo.io/v2
kind: ExtAuthPolicy
metadata:
  name: oidc-auth
  namespace: gloo-mesh
spec:
  applyToRoutes:
    - route:
        labels:
          oauth: "true"
  config:
    server:
      name: ext-auth-server
      namespace: gloo-mesh-addons
      cluster: cluster-1
    glooAuth:
      configs:
        - oauth2:
            oidcAuthorizationCode:
              ...
              dynamicMetadataFromClaims:
                 sub: userId
                 custom_claim: customClaimKey
```
Similarly for a user configuring an ext auth policy that uses AccessValidation and want to extract claims from the accessToken and emit metadata that is consumed by the rate limit filter would do the following:

```             
apiVersion: security.policy.gloo.solo.io/v2
kind: ExtAuthPolicy
metadata:
  name: access-validation-auth
  namespace: gloo-mesh
spec:
  applyToRoutes:
    - route:
        labels:
          oauth: "true"
  config:
    server:
      name: ext-auth-server
      namespace: gloo-mesh-addons
      cluster: cluster-1
    glooAuth:
      configs:
        - oauth2:
             accessTokenValidationConfig: 
                ...
              dynamicMetadataFromClaims:
                  sub: userId
                  custom_claim: customClaimKey
```

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -840,6 +846,32 @@ message OauthSecret {
...
 }
 
+// ClaimsToMetadata allows claims to be extracted from the ID token and/or the access token and
+// emitted as dynamic metadata.
+message ClaimsToMetadata {
+
+    // Use this field to extract claims from the ID token. (Note that this field only works when used with the OIDC flow.)
+    IdToken id_token = 1;
+    // Use this field to extract claims from the access token.
+    AccessToken access_token = 2;
```
marcogschmidt: After thinking more about this and looking the examples you added (thanks!), I am wondering whether we need this distinction in the API. My thinking:
- `AccessTokenValidation` only concerns itself with the `access_token`. It currently ignores the `id_token` and I think it should continue doing so. Therefore we don't need any config related to the `id_token` there.
- `OidcAuthorizationCode`: the docs for that field say `can also authorize the access token if configured`, but looking at the code I don't find any logic to validate the `access_token`; all we seems to be doing is adding it to a header. The text might refer to [this commented out config](https://github.com/solo-io/gloo/blob/becaafbd0222e440172e042a506b28f6537fa305/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto#L639); this is supported by the fact that both have been added in [the same PR](https://github.com/solo-io/gloo/pull/3171). To be able to extract the claims here we would need to parse the `access_token`, but since there is not guarantee that it is a JWT, we would need to basically repeat what we are doing in the `AccessTokenValidation` API, which I think we should not do, even more so because we can chain the two auth existing steps to achieve this (*). TL;DR similar comment to the previous bullet point, i.e. that we do not need to account for the `access_token` here.

(*) If the user wants to use OIDC to authenticate and then also validate the `access_token` and extract claims they can do something like this:

```yaml
apiVersion: security.policy.gloo.solo.io/v2
kind: ExtAuthPolicy
metadata:
  name: oidc-auth
  namespace: gloo-mesh
spec:
  applyToRoutes: [] #omitted for brevity
  config:
    server: {} #omitted for brevity
    glooAuth:
      configs:
      - oauth2:
          oidcAuthorizationCode: {} #omitted for brevity
      - accessTokenValidationConfig:
          # either JWT or introspection
          claimsToMetadata:
            usagePlan: usage-plan-id
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -840,6 +846,32 @@ message OauthSecret {
...
+
+    message IdToken {
+        // The name of the claim to extract the value from
+        string claim = 1;
+
+        // The key to use for the dynamic metadata emitted
+        string metadata_key = 2;
+    }
```
marcogschmidt, line 864: Assuming that we apply the suggestion in my previous comment, the `claims_to_metadata` fields might just be simple string/string maps.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -840,6 +846,32 @@ message OauthSecret {
...
+
+    message AccessToken {
+        // The name of the claim to extract the value from
+        string claim = 1;
```
marcogschmidt, line 868: Some questions:
- what if the claim is not a simple string data type? (e.g. it's a list, an object, a comma-separated list (e.g. scopes), a bool, etc)
- can it contain special characters? Thinking specifically about `.`, which in the future we might want to use to access nested data, e.g. using XPath notation

We should document the valid format for this field here.

marcogschmidt (commented):
Thanks for documenting the PR so well! I think we can simplify the API a bit, let me know what you think.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jmhbh: This makes sense, thanks for the comment. I took a look at the code as well, and it doesn't look like we're parsing the accessToken at all so this approach makes more sense.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jmhbh: This is a good callout, we should document that the data type of the claim value can be any of a list, an object, or a bool. In terms of special characters, I don't see why the value cannot contain them. I think for the rate limit use case we will primarily be interested in the claim value being a simple string, but there potentially could be a future use case where we could leverage the claim value containing special characters? Either way since we are just passing this on as dynamic metadata we shouldn't add any logic that will prevent claim values with special characters from being passed on within the ext-auth-service.

marcogschmidt: I was referring to the value of the field that we use to look up the claim, I realize that was not clear, sorry. I think that we should validate that it is just a DNS label (same as labels, annotations, etc), so it can contain only `[a-zA-Z\-_]` chars. Example:

```yaml
claimsToMetadata:
  sub: subject # ok
  my-claim: myClaim # ok
  foo.bar: fooBar # invalid 
```

So you plan on handling claims of any type? So that means that we will copy whatever structure that claim has to the dynamic metadata right (which is possible since it's proto `Struct` field)?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,15 @@ message OidcAuthorizationCode {
     // specifications. Noted [here](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
     // in the OIDC documentation.
     EndSessionProperties end_session_properties = 19;
+
+    // Map of claims to metadata keys. Ie:
+    // claims_to_metadata:
+    //   iss: issuer
+    //   email: email
```
marcogschmidt, line 662: I think we should invert the order of claim name and MD field name here. That way:
- we can automatically guarantee that metadata keys will be unique
- we allow users to copy the same claim to more than one field (can't see why they would do that, but since we get this for free, why not?)

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,15 @@ message OidcAuthorizationCode {
...
+    // claim values can be a bool, string, or list and can contain special characters.
```
marcogschmidt, line 663: We should say that claims can be on any type and that their structure will be reflected in the dynamic metadata entry 1:1.

marcogschmidt (changes requested):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jmhbh: gotcha, that makes sense. The types of the claim value should be either a list, a bool or a string, and in the `ext-auth-service` we will determine the type of the claim value and use the approiate type `structpb.Value_StringValue`, `structpb.Value_ListValue` or `structpb.Value_BoolValue`. The dynamic metadata has the `Struct` type and the claim value could be any json value, but in order to create the correct data structure we would have to recursively figure out each type within the json object. I think we should start with supporting the claim value being a bool, string or list for now and if we get a request to support the claim value being a complex json object we could tackle it then.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jmhbh: that makes sense, will make this change.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,16 @@ message OidcAuthorizationCode {
     // specifications. Noted [here](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
     // in the OIDC documentation.
     EndSessionProperties end_session_properties = 19;
+
+    // Map of metadata key to claim. Ie:
+    // metadata_keys_to_claim:
+    //   issuer: iss
+    //   email: email
+    // When specified, the matching claims from the access token will be emitted as dynamic metadata.
```
marcogschmidt, line 663:
```suggestion
    // When specified, the matching claims from the ID token will be emitted as dynamic metadata.
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,16 @@ message OidcAuthorizationCode {
...
+    // Note that metadata keys must be unique, and the claim names must match the regular expression for DNS labels `[a-zA-Z\-_]`.
+    // The metadata will live in a namespace specified by the canonical name of the ext auth filter (in our case `envoy.filters.http.ext_authz`).
+    // claim values can be a bool, string, or list.
```
marcogschmidt, line 666: Here we should just say that the structure of the claim will be preserved in the metadata struct.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,16 @@ message OidcAuthorizationCode {
...
+    map<string, string> metadata_keys_to_claim = 20;
```
marcogschmidt, line 667: Name implies MD -> claims. Should be something like `dynamic_metadata_from_claims` or `claims_to_dynamic_metadata`.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -655,6 +655,16 @@ message OidcAuthorizationCode {
     // specifications. Noted [here](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
     // in the OIDC documentation.
     EndSessionProperties end_session_properties = 19;
+
+    // Map of metadata key to claim. Ie:
+    // metadata_keys_to_claim:
+    //   issuer: iss
+    //   email: email
+    // When specified, the matching claims from the access token will be emitted as dynamic metadata.
+    // Note that metadata keys must be unique, and the claim names must match the regular expression for DNS labels `[a-zA-Z\-_]`.
```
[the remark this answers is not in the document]
jmhbh: Thanks, I forgot to update this after the implementation yesterday. The correct regex should actually be `^[a-zA-Z\-_]*$`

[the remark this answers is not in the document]
jmhbh: well, this regex should only apply for the claim key and not the fields within the claim value. AFAIK, the claim names should only contain lower/upper case alphabets with `-` or `_` as separators so the DNS label actually makes sense to me.

rinormaloku (dismissed):
LGTM

jbohanon (approved):

marcogschmidt (approved):

jbohanon (approved):

PR #8261, opened 2023-05-19
Refactor bootstrap clients

jbohanon:
# Description

Refactor the client constructors in `bootstrap/utils.go` to their own package `bootstrap/clients` and separated into smaller, discrete files to more easily work with a single client. The code in `bootstrap/utils.go` has been modified to transparently use the `bootstrap/clients` package code. 

# Checklist:


projects/gloo/cli/pkg/flagutils/common.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/options"
 	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/options/contextoptions"
 	"github.com/solo-io/gloo/projects/gloo/cli/pkg/printers"
-	"github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+	bootstrap_clients "github.com/solo-io/gloo/projects/gloo/pkg/bootstrap/clients"
```
nfuden, line 10: https://go.dev/blog/package-names

can we rename the import as? Technically its not the package name itself but the same thing some what maintains.

In other words if clients is already used then we can just import as boostrap instead of this java naming

nfuden (commented):
I like it in general. Can I ask 3 things?
1. can we just have the deprecated functions call the new ones? The description seems to imply this but it 
2. Can we rename our imports

jbohanon:
> I like it in general. Can I ask 3 things?
> 
> 1. can we just have the deprecated functions call the new ones? The description seems to imply this but it
> 2. Can we rename our imports

1. This is what the deprecated functions do
2. I like it. The only issue I could see would be if both `clients` _and_ `bootstrap` were already being used, but I don't know that this is the case anywhere, and that would certainly be an exception that can be handled case-by-case
3. This was only two things?

nfuden:
1. Im more mentioning making the deprecated functions just call the new package.
2. bootstrap isnt used in the file in question so should be safe to import once again as bootstrap instead of client or bootstrap client
3. Ah yeah I took one thing off so 2.

projects/gloo/pkg/bootstrap/clients/consul.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+
+	consulapi "github.com/hashicorp/consul/api"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/solo-kit/pkg/utils/prototime"
+)
+
+// default query options for consul
```
jbohanon, line 12: nit: godoc format
```suggestion
// DefaultConsulQueryOptions provides default query options for consul
```

nfuden (approved):
Nice

jbohanon:
Here's a s-p PR showing no breakage https://github.com/solo-io/solo-projects/pull/4983

sam-heilbron (approved):

PR #8262, opened 2023-05-19
Multiple secret sources api

jbohanon:
# Description

This PR introduces the proto API changes needed to allow basic configuration of a `ResourceClient` for secrets which uses multiple storage sources. This is to enable Vault PKI integration, which will require the ability to read secrets from both Kubernetes and Vault. 

This is not implemented in this PR, but implementation will follow quickly in a follow-on PR.  

# Context

# Checklist:


projects/gloo/api/v1/settings.proto
```diff
@@ -50,13 +50,31 @@ message Settings {
...
 
+    // Deprecated: use secretOptions
     // Determines where Gloo will read/write secrets from/to.
     oneof secret_source {
         KubernetesSecrets kubernetes_secret_source = 6;
         VaultSecrets vault_secret_source = 7;
         Directory directory_secret_source = 8;
     };
 
+    // Settings for secrets storage
```
sam-heilbron, line 61: When this merges, this will not be impleneted. Just for safety, let's claify that in thie PR, and in the follow-up PR to implement, we can remove it

projects/gloo/api/v1/settings.proto
```diff
@@ -50,13 +50,31 @@ message Settings {
         ConsulKv consul_kv_source = 21;
     };
 
+    // Deprecated: use secretOptions
```
sam-heilbron, line 53: Let's hold off on this, until the feature is completed. When we do add it, let's clarify why you would want to use the new api

projects/gloo/api/v1/settings.proto
```diff
@@ -50,13 +50,31 @@ message Settings {
...
         VaultSecrets vault_secret_source = 7;
         Directory directory_secret_source = 8;
     };
 
+    // Settings for secrets storage
+    SecretOptions secret_options = 38;
+
+    message SecretOptions {
+        message Source {
+            // Determines where Gloo will read/write secrets from/to.
```
sam-heilbron, line 66: What is the beahvior if no sources are defined?

projects/gloo/api/v1/settings.proto
```diff
@@ -50,13 +50,31 @@ message Settings {
...
+            oneof source {
+                KubernetesSecrets kubernetes = 1;
+                VaultSecrets vault = 2;
+                Directory directory = 3;
+            };
+        }
+
+        // List of configured secret sources
```
sam-heilbron, line 74: Does order matter here?

sam-heilbron (commented):
Looking good!  A few nits

projects/gloo/api/v1/settings.proto
jbohanon: We will check for deprecated `settings.SecretSource` and return `nil, error`. I'm still investigating the fallout of returning nil, but I think we bubble the error up through the setup funcs

projects/gloo/api/v1/settings.proto
jbohanon: No, the clients will be initialized in parallel. If one client relies on another, eventually the dependency should be initialized, or else an error will be logged.

projects/gloo/api/v1/settings.proto
jbohanon: Updated comments to reflect not-yet-implemented 8d6ab107659e891d2241da6ce0ccb4b2351e520f

projects/gloo/api/v1/settings.proto
jbohanon: Removed deprecation; added note of future deprecation 8d6ab107659e891d2241da6ce0ccb4b2351e520f

projects/gloo/api/v1/settings.proto
```diff
@@ -51,12 +51,32 @@ message Settings {
...
+            oneof source {
+                KubernetesSecrets kubernetes = 1;
+                VaultSecrets vault = 2;
+                Directory directory = 3;
+            };
+        }
+
+        // List of configured secret sources. These clients will be initialized
+        // in parallel in indeterminate order.
+        repeated Source secret_sources = 1;
```
jackstine, line 77: so the max size that this list can be is 3, correct? one for each type of supported source.

jbohanon: Not necessarily. I think it would be very unlikely that a user would need multiple of a single type, but if they needed e.g. two different vault clusters, they could put two configs in and the multi-client would spin up a resource client for each

sam-heilbron (dismissed):
Great work!

elcasteel (approved):
Are we going to set a default in helm? Would it make sense to put that change here or in the implementation?

jbohanon:
> Are we going to set a default in helm? Would it make sense to put that change here or in the implementation?

That's a good point... Since this is purely opt-in, I think it makes sense to leave it nil for now

sam-heilbron (approved):

Issue #7641, opened 2023-01-06
Support for mTLS and self-signed certs for Vault integration


jbohanon:
### Version

None

### Is your feature request related to a problem? Please describe.

Currently the `settings` only allow for self-signed CAs and mTLS cert/key pairs to be mounted via filesystem path. This is not a very clean option when running Gloo Edge in Kubernetes. 

### Describe the solution you'd like

The ability to specify these values in some clean way when running on Kubernetes. This is complicated by the fact that we cannot use a `secretRef` to define sensitive information since the secrets backend is what we are trying to configure.

### Describe alternatives you've considered

It seems like it would be possible to achieve this today by creating secrets for the required values in Kubernetes and mounting these secrets into the gloo pod, then pointing to the mounted paths in the `settings` CR

### Additional Context

This is a required feature to fulfill https://github.com/solo-io/gloo/issues/7342

PR #8263, opened 2023-05-19
Multiple secret sources impl

jbohanon:
# Description

Implementation of API introduced in https://github.com/solo-io/gloo/pull/8262. This is in order to enable Vault PKI integration whereby the user will need to be able to access Kubernetes secrets _and_ PKI values in Vault. 

The multi-client will attempt to initialize a `ResourceClient` for each configured source in series and return any error that occurs.  

This is implemented fully opt-in. If the new API is not configured, the old API will be used. If neither is configured, we still return an error. If the new API is used, but only a single source is configured, we will use the discrete resource client for that source.

Some function signatures were changes to accept ctx + params, but this should not be breaking for anyone since these functions were recently refactored into a new package, and the functions which were previously exposed are still exposed with the same signature. (see https://github.com/solo-io/gloo/pull/8261)

# Testing
```sh
helm repo add gloo-test https://storage.googleapis.com/gloo-ee-test-helm
helm repo update
helm install -n gloo-system gloo-test gloo-test/gloo --create-namespace --version 1.15.0-beta2-bcheck-multiple-secrets-api-0005b24
```
Apply some resources:
```sh
# requires valid cert and key at ~/testcerts
kubectl create ns test-ns
glooctl create secret tls \
    --certchain ~/testcerts/cert.pem \
    --privatekey ~/testcerts/key.pem \
    --name tls-secret-k8s --namespace test-ns

glooctl create secret tls \
    --certchain ~/testcerts/cert.pem \
    --privatekey ~/testcerts/key.pem \
    --name tls-secret-vault --namespace test-ns \
    --use-vault \
    --vault-address "$VAULT_ADDRESS" \
    --vault-token "$VAULT_TOKEN"

k create -f - <<EOF
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: json-upstream-bad
  namespace: gloo-system
spec:
  sslConfig:
    secretRef:
      name: tls-secret-bad # we expect this not to exist
      namespace: test-ns
  static:
    hosts:
      - addr: jsonplaceholder.typicode.com
        port: 80
---
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: json-upstream-k8s
  namespace: gloo-system
spec:
  sslConfig:
    secretRef:
      name: tls-secret-k8s
      namespace: test-ns
  static:
    hosts:
      - addr: jsonplaceholder.typicode.com
        port: 80
---
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: json-upstream-vault
  namespace: gloo-system
spec:
  sslConfig:
    secretRef:
      name: tls-secret-vault
      namespace: test-ns
  static:
    hosts:
      - addr: jsonplaceholder.typicode.com
        port: 80
---
kind: VirtualService
metadata:
  name: vs
  namespace: gloo-system
spec:
  virtualHost:
    domains:
    - '*'
    routes:
    - matchers:
      - exact: /k8s
      routeAction:
        single:
          upstream:
            name: json-upstream-k8s
            namespace: gloo-system
    - matchers:
      - exact: /vault
      routeAction:
        single:
          upstream:
            name: json-upstream-vault
            namespace: gloo-system
    - matchers:
      - exact: /bad
      routeAction:
        single:
          upstream:
            name: json-upstream-bad
            namespace: gloo-system
EOF

glooctl check 
```
here you should see 2 upstream errors ("-bad" and "-vault") until you apply the Settings change below, at which point glooctl check should only return an error for the "bad" upstream

then, when up and running with that and a vault cluster, modify settings to remove `kubernetesSecretSource: {}` or whatever else was there, then add this at the same level the removed line was indented
```yaml
  secretOptions:
    sources:
    - vault:
        accessToken: <vault access token>
        address: <vault address>
    - kubernetes: {}
```

(to set up a vault cluster for testing see this PR description https://github.com/solo-io/gloo/pull/7669 or [install via helm](https://developer.hashicorp.com/vault/docs/platform/k8s/helm))

For extra credit, install a second vault instance/cluster and add another secret, upstream, route, and settings entry to see that we can now use two discrete vault sources
 
# Checklist:


projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -62,3 +104,322 @@ func SecretFactoryForSettings(ctx context.Context,
...
+	errNotImplemented = func(ctx context.Context, method string) error {
+		err := errors.Wrap(ErrNotImplemented, method)
+		contextutils.LoggerFrom(ctx).DPanic(err.Error())
+
+		return err
+	}
+
+	// ErrNotImplemented indicates a call was made to an interface method which this
+	// multi-client has not implemented
+	ErrNotImplemented = errors.New("client method not implemented on MultiSecretResourceClient")
```
jbohanon, line 408: Use more specific name since this is exported from the `clients` package

elcasteel (commented):
Do we retry initializing individual clients? I can't find that in the code and the documentation makes it sound like we potentially do based on the fact we are ok with some errors.

projects/gloo/pkg/bootstrap/clients/secret_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	. "github.com/solo-io/gloo/projects/gloo/pkg/bootstrap/clients"
+)
+
+var _ = Describe("secrets", func() {
+	Context("multi client factory", func() {
```
elcasteel, line 11: Would it be feasible to add unit tests for cases where all clients fail to initialize and some clients fail and some succeed?

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -35,6 +44,39 @@ func SecretFactoryForSettings(ctx context.Context,
...
+}
+
+func NewSecretResourceClientFactory(ctx context.Context,
+	settings *v1.Settings,
+	sharedCache memory.InMemoryResourceCache,
+	cfg **rest.Config,
+	clientset *kubernetes.Interface,
+	kubeCoreCache *cache.KubeCoreCache,
+	vaultClient *vaultapi.Client,
+	pluralName string) (factory.ResourceClientFactory, error) {
```
jackstine, line 79: nit: this function signature is too large. Can we use a structure to limit the size of the signature?

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -60,5 +102,320 @@ func SecretFactoryForSettings(ctx context.Context,
...
+)
+
+// NewMultiSecretResourceClientFactory returns a resource client factory for a client
+// supporting multiple sources
+func NewMultiSecretResourceClientFactory(secretSources []*v1.Settings_SecretOptions_Source,
+	sharedCache memory.InMemoryResourceCache,
+	cfg **rest.Config,
+	clientset *kubernetes.Interface,
+	kubeCoreCache *cache.KubeCoreCache,
+	vaultClient *vaultapi.Client) (*MultiSecretResourceClientFactory, error) {
```
jackstine, line 137: nit: same thing, signature is to large.

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -60,5 +102,320 @@ func SecretFactoryForSettings(ctx context.Context,
...
+	secretSources []*v1.Settings_SecretOptions_Source
+	sharedCache   memory.InMemoryResourceCache
+	cfg           **rest.Config
+	clientset     *kubernetes.Interface
+	kubeCoreCache *cache.KubeCoreCache
+	vaultClient   *vaultapi.Client
+
+	refreshRate     *durationpb.Duration
+	watchNamespaces []string
+}
```
jackstine, line 118:
```suggestion
var _ factory.ResourceClientFactory = new(MultiSecretResourceClientFactory)

type MultiSecretResourceClientFactory struct {
	secretSources []*v1.Settings_SecretOptions_Source
	sharedCache   memory.InMemoryResourceCache
	cfg           **rest.Config
	clientset     *kubernetes.Interface
	kubeCoreCache *cache.KubeCoreCache
	vaultClient   *vaultapi.Client

	refreshRate     *durationpb.Duration
	watchNamespaces []string
}
```

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -60,5 +102,320 @@ func SecretFactoryForSettings(ctx context.Context,
...
+//
+// Direct access to clientList is deliberately omitted to prevent changing clients
+// with an open Watch leading to inconsistent and undefined behavior
+type MultiSecretResourceClient struct {
+	*sync.RWMutex                          // because we are initializing clients asynchronously in parallel
+	clientList    []clients.ResourceClient // do not use clients.ResourceClients here as that is not for this purpose
+	hasKubernetes bool
+	hasDirectory  bool
+	hasVault      bool
+}
```
jackstine, line 304:
```suggestion

var _ clients.ResourceClient = new(MultiSecretResourceClient)

// MultiSecretResourceClient represents a client that is minimally implemented to facilitate Gloo operations.
// Specifically, only List and Watch are properly implemented.
//
// Direct access to clientList is deliberately omitted to prevent changing clients
// with an open Watch leading to inconsistent and undefined behavior
type MultiSecretResourceClient struct {
	*sync.RWMutex                          // because we are initializing clients asynchronously in parallel
	clientList    []clients.ResourceClient // do not use clients.ResourceClients here as that is not for this purpose
	hasKubernetes bool
	hasDirectory  bool
	hasVault      bool
}
```

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -60,5 +102,320 @@ func SecretFactoryForSettings(ctx context.Context,
...
+	// create a goroutine to receive on errChan so we can handle the possible case of a client
+	// failing to initialize after this function returns
+	loggedErrChan := make(chan error)
+	go func() {
+		logger := contextutils.LoggerFrom(ctx)
+		for errToLog := range errChan {
+			logger.Error(errToLog)
+			loggedErrChan <- errToLog
+		}
+	}()
```
jackstine, line 194: shouldn't we select on the `context.Done()`?

jackstine, line 194: what are we doing with `loggedErrChan`?

jbohanon: `loggedErrChan` is received on toward the bottom of the function to receive and count errors. We still want to log client initialization errors if we short-circuit because we received a single success, but we would otherwise no longer be monitoring for them on `errChan`, so this provides an intermediary to log the error and send it along without caring if anyone is receiving on `errChan`

jbohanon: Good call, added

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: Good call! I added a structure param for htis

jbohanon:
> Do we retry initializing individual clients? I can't find that in the code and the documentation makes it sound like we potentially do based on the fact we are ok with some errors.

Good call, no there was not any retry logic implemented. I put a retry on the `initClientAsync`, which should pretty trivially resolve some issues. More complicated was ensuring we are retrying the Vault API client's connection to the Vault instance since we currently initialize that connection way before constructing the `ResourceClientFactory` even. To address this, I modified the parameters that are passed through `constructOpts` to be initialization functions which can be retried asynchronously even after we return the `ResourceClientFactory`

projects/gloo/pkg/bootstrap/clients/secret_test.go
jbohanon: Good call. I will add these

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,344 @@ func SecretFactoryForSettings(ctx context.Context,
...
+		if err != nil {
+			return nil, err
+		}
+		list = append(list, clientList...)
+	}
+
+	return list, nil
+}
+
+func (m *MultiSecretResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-chan resources.ResourceList, <-chan error, error) {
```
jbohanon, line 455: This isn't going to work because the snapshot emitter is listening for discrete receives on the `ResourceList` chan, so each new send we make on that chan is going to blow out the last, and the snapshot is only going to have the secrets from the latest client to respond. Need to make sure each client responds before sending on the chan to the snapshot emitter. It would probably be a good idea to put some sort of timeout so we're not waiting on e.g. the vault client not able to connect

elcasteel: Can we add e2e tests to change secrets in multiple locations and expect `Consistently` that the snapshot has everything?

elcasteel: there might even be a way to fit it into a unit test because we don't necessarily need to run translation to see this working but I don't have experience with goroutines in unit tests so idk.

projects/gloo/pkg/bootstrap/clients/secret_test.go
jbohanon: I added these unit tests!

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: I saw that I missed changing the other select to listen on `loggedErrChan`. I got that fixed

projects/gloo/pkg/bootstrap/clients/consul_test.go
```diff
@@ -0,0 +1,100 @@
...
+var _ = Describe("consul", func() {
+
+	var (
+		ctx    context.Context
+		cancel context.CancelFunc
+	)
+
+	BeforeEach(func() {
+
+		if !testutils.IsEnvTruthy(testutils.RunConsulTests) {
```
sam-heilbron, line 29: Could we use https://github.com/solo-io/gloo/blob/main/test/testutils/requirements.go#L168 and the ValidateRequirements at the top of that function? I tried to guard our tests like this so users could opt in/out in a consistent manner

projects/gloo/pkg/bootstrap/clients/consul_test.go
```diff
@@ -0,0 +1,100 @@
...
+	Context("artifacts as consul key value", func() {
+
+		var (
+			ctx            context.Context
+			cancel         func()
+			artifactClient v1.ArtifactClient
+		)
+
+		BeforeEach(func() {
+			ctx, cancel = context.WithCancel(context.Background())
```
sam-heilbron, line 54: We have an outer ctx and an inner ctx. Do we need this second one, or can we just inherit the outer context?

projects/gloo/pkg/bootstrap/clients/consul_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+			value, err := protoutils.MarshalYAML(&v1.Artifact{
+				Data:     map[string]string{"hi": "bye"},
+				Metadata: &core.Metadata{Name: "name", Namespace: "namespace"},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = client.KV().Put(&api.KVPair{
```
sam-heilbron, line 62: Perhaps outside the scope of this work, but do we want to expose a Put on the ConsulInstance so people can use that test service and abstract the need to use the consul client directly?

projects/gloo/pkg/bootstrap/clients/kube.go
```diff
@@ -24,6 +25,9 @@ func initializeForKube(ctx context.Context,
 		if err != nil {
 			return err
 		}
+		if cfg == nil {
```
sam-heilbron, line 28: How can this happen? Can GetConfig return nil config and nil error?

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+type asyncClientInitParams struct {
+	multiClient     *MultiSecretResourceClient
+	rcFactoryFunc   func() (factory.ResourceClientFactory, error)
+	onInitialized   func(*MultiSecretResourceClient, clients.ResourceClient)
+	newClientParams factory.NewResourceClientParams
+	errChan         chan error
+	doneChan        chan struct{}
+}
+
+var initClientAsync = func(ctx context.Context, params *asyncClientInitParams) {
```
sam-heilbron, line 232: I like the idea behind this, though I wonder if now/here is the right place to add this complexity. A few questions:
- What would lead a resource client to fail to be constructed, that a retry would fix?
- We could add this to this custom multi secret factory, but should we instead have all of our clients be consistent?

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+			case <-ctx.Done():
+				return
+			case errToLog := <-errChan:
+				logger.Error(errToLog)
+				loggedErrChan <- errToLog
+			}
+		}
+	}()
+
+	// Create a channel to receive the event that at least one client is initialized. Because
```
sam-heilbron, line 285: I like the idea here, though I think this may be a nice iteration we can add more holistically in the product. It feels like the MultiSecret factory, when creating a new resource client, should just delegate to the underlying clients themselves. If we build in retry logic more holistically, it allows our simple abstraction here (multiple clients under the hood) to remain simple

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+	return m.hasDirectory
+}
+
+func (m *MultiSecretResourceClient) HasVault() bool {
+	m.Lock()
+	defer m.Unlock()
+	return m.hasVault
+}
+
+func (m *MultiSecretResourceClient) Kind() string {
```
sam-heilbron, line 432: If we hit this, aren't we assured that:
1. There exists >0 clients
2. All clients interact with the same kind

This feels like a function that could be simplified

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+		return nil
+	}
+
+	// Any of the clients should be able to handle this identically
+	return m.clientList[0].NewResource()
+}
+
+// Deprecated: implemented only by the kubernetes resource client. Will be removed from the interface.
+func (m *MultiSecretResourceClient) Register() error {
+	for _, v := range m.clientList {
```
sam-heilbron, line 456: Since this is deprecated, and not needed for this client to fucntion, I think we should just make this funtion a no-op (or even call an error if called)

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+		}
+	}
+
+	// return no error since the EC2 plugin calls the Register() function
+	return nil
+}
+
+func (m *MultiSecretResourceClient) List(namespace string, opts clients.ListOpts) (resources.ResourceList, error) {
+	list := resources.ResourceList{}
+	for i := range m.clientList {
```
sam-heilbron, line 473: Is the clientList guaranteed to be a consistent order? It could be useful to call out why we believe the list we return will be a consistent order (ie because the underlying client is reponsible for sorting its own list)

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+}
+
+func (m *MultiSecretResourceClient) Watch(namespace string, opts clients.WatchOpts) (<-chan resources.ResourceList, <-chan error, error) {
+	listChan := make(chan resources.ResourceList)
+	errChan := make(chan error)
+
+	// create a new aggregator so we can keep the last known state of individual clients.
+	// this allows us to send a single ResourceList to the api snapshot emitter, which
+	// expects values received on the returned channel to be atomically complete
+	resourceListPerClient, err := newResourceListAggregator(m, namespace, opts)
```
sam-heilbron, line 528: Have you had a chance to read how we solve a similar problem in our emitter(https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/api/v1/gloosnapshot/api_snapshot_emitter.sk.go#L432) ? I like your generic approach, though if it's complex we may want to use the more transparent approach that our emitter does

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -28,15 +29,50 @@ type vaultSecretClientSettings struct {
...
+	}
+}
+
+var (
+	ErrNilVaultClient = errors.New("vault API client failed to initialize")
+)
+
 // NewVaultSecretClientFactory consumes a vault client along with a set of basic configurations for retrieving info with the client
-func NewVaultSecretClientFactory(client *api.Client, pathPrefix, rootKey string) factory.ResourceClientFactory {
+func NewVaultSecretClientFactoryWithRetry(clientInit VaultClientInitFunc, pathPrefix, rootKey string) factory.ResourceClientFactory {
```
sam-heilbron, line 45: For this, would we want to pass retryOptions to the function, so the caller can determine how to retry? We do something like that https://github.com/solo-io/gloo/blob/main/test/helpers/snapshot_writer.go#L24 where we know some retry cases we want to handle, but expose more options

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+					return NewVaultSecretClientFactoryWithRetry(f, pathPrefix, rootKey), nil
+				}
+				initParams.onInitialized = clientCallback(multiClient, &multiClient.hasVault)
+			}
+		}
+
+		go initClientAsync(ctx, initParams)
+	}
+
+	// We wait for at least one client to be initialized, then return our multiClient.
```
sam-heilbron, line 361: Should we error loudly if any of the clients failed to initialize? Part of the reason we attempt to init clients at the beginning of Gloo and error loudly is so that a restart could take place if one needs to. Is there a danger in allowing "partial" config to be accepted...meaning 1 client (to kubernetes for example) works and 1 client (to vault) doesn't?

projects/gloo/pkg/bootstrap/clients/secret.go
```diff
@@ -54,11 +161,423 @@ func SecretFactoryForSettings(ctx context.Context,
...
+	SharedCache        memory.InMemoryResourceCache
+	Cfg                **rest.Config
+	Clientset          *kubernetes.Interface
+	KubeCoreCache      *cache.KubeCoreCache
+	VaultClientInitMap map[int]VaultClientInitFunc
+}
+
+// NewMultiSecretResourceClientFactory returns a resource client factory for a client
+// supporting multiple sources
+func NewMultiSecretResourceClientFactory(params MultiSecretFactoryParams) (factory.ResourceClientFactory, error) {
```
sam-heilbron, line 206: I had kind of imagined this would just wrap a call to the "sub" resource client factory (the original), and then we would use that factory whenever we needed to build the client, but we would do it for each of the clients. That way all complexity would still live in the original factory. Is that not possible given that go types are different?

sam-heilbron (commented):
This is looking great! I had a few questions about the initialization complexity and retry logic. Happy to be convinced otherwise, but I'm leaning towards saying we can do a first pass without it and then add it in more widespread across more of our init clients

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -339,7 +339,17 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 	}
 
 	var vaultClient *vaultapi.Client
-	if vaultSettings := settings.GetVaultSecretSource(); vaultSettings != nil {
+	vaultSettings := settings.GetVaultSecretSource()
```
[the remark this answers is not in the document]
jbohanon: I'm not sure. My thinking was that since the `secretSource` API will be deprecated, and the `secretOptions` is purely opt-in, if `secretOptions` are defined, then the user explicitly opted in and we shouldn't muddy with the old API if they forgot to remove it. I'm happy to invert these if that's the precedent

projects/gloo/pkg/bootstrap/clients/kube.go
jbohanon: Hmm I was experiencing an NPE when passing nil `cfg` in for testing, it would panic on the deref on L31. This check is definitely redundant though and I'll take a closer look at this

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: It is still called by e.g. the EC2 plugin and I believe a couple other places. I'm not certain at all that it's necessary and would be more than happy to return nil, but we can't return an error

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: I was attempting to implicitly provide for the case where one secret client relies upon another, but I found (to my extreme annoyance) that we don't actually check that we have connectivity to vault before returning a new rc from the factory. So we have no idea if we're hosed until a request is made. If we leave that as it is, we can probably safely block until all clients init or err. If we fix that (I see it as bad behavior but I could be missing something) then I think this still makes sense

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: I'm not following you here. Let's touch base on this tomorrow

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: So it turns out that we don't actually fail to initialize any of the RCs (except k8s) if they're not healthy. Given that, it seems pretty rare we would fail to init a rc since there are basically no guardrails until we try to use it. I'm not sure this is the best pattern of behavior, but it might have been put into place for a reason. 

That said, I think abstracting multiclient and asyncclient interfaces into sk would be a cleaner move for sure

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: This is a good point; I simplified the function

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: I added a stable sort and explained the sorting in a comment near the `clientList` definition and in the proto

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: Let's chat on this. I don't know what you mean by transparent approach

projects/gloo/pkg/syncer/setup/setup_syncer.go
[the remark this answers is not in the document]
sam-heilbron: Yeah I agree with the opt-in behavior. As I think about it more, if we want to truly deprecate the single secretSource option, then I'd like us to have a plan (timeline) for when we'll stop supporting that. If we don't yet have that, perhaps we don't mark it as deprecated, and instead clarify that we just have 2 separate ways to configure secrets. I'd rather mark something as deprecated after the new feature has been hardened/validated.

As for the behavior between the two, I more wanted to make sure that everytime we process these settings that we are consistent. I like the idea that if they've defined the multi secret options, we read that, and if not, we fallback to the original.

projects/gloo/pkg/bootstrap/clients/kube.go
sam-heilbron: Yeah I'd be curious how we hit that case, so if there are logs to share I'm happy to take a look

projects/gloo/pkg/bootstrap/clients/secret.go
sam-heilbron: Sure thing! My comment was off the cuff and perhaps not super helpful. I was more pointing out that since we have a pattern for establishing multiple watches and aggregating them, we may just want to follow that pattern (or we can find a new pattern, test it and make it more widely availabe)

projects/gloo/pkg/bootstrap/clients/secret.go
sam-heilbron: Yeah I think we need to determine/better clarify the existing contract of ResourceClients (not your fault, this is legacy code that isn't well defined). It seems like either a RC has a connection upon initialization, or were not assured that it will. I'd prefer to follow your pattern and establish a connection, retry a few times, and if unsucessful error loudly.

however, if the existing clients don't have that contract built-in (guaranteed successful connection), perhaps we work with existing assumptions, and build in proper connectivity handling in a follow up

projects/gloo/pkg/bootstrap/clients/secret.go
sam-heilbron: Sure thing!

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: It's ostensibly the same pattern as in that code:
1. loop over discrete elements (namespaces/clients)
2. initialize with a `List` on that element (subtly different; I initialized before the loop in a separate function)
3. launch a goroutine for each element
4. receive resources in goroutine
5. replace in map, convert to list, send to consumer (in this case, the consumer is the emitter or anything else watching, for the emitter, the consumer is the snapshot struct

We could possibly create a useful abstraction a la [solo-kit aggregate watcher](https://github.com/solo-io/solo-kit/blob/main/pkg/api/v1/clients/wrapper/multi_watcher.go), then refactor out the code here and in emitter, but that feels outside the scope of this. Also, I didn't use that solo-kit struct because it drops data when errors are received, and I don't think we want clients affected like that by intermittent client failures

projects/gloo/pkg/bootstrap/clients/secret.go
jbohanon: https://github.com/solo-io/gloo/issues/7641

jbohanon: > however, if the existing clients don't have that contract built-in (guaranteed successful connection), perhaps we work with existing assumptions, and build in proper connectivity handling in a follow up

They do not. I agree this is a follow-up move. I removed the asyc and retry logic from this PR.

Now the ResourceClients are initialized sequentially (kube > directory > vault) and if anything fails we error out

projects/gloo/pkg/bootstrap/clients/kube.go
jbohanon: I'm not sure that code path was ever executed properly. This is ALWAYS an NPE if `GetConfig` doesn't return an error. 
https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/clients/kube.go#L22-L28
https://goplay.tools/snippet/nyrwlIZ2-nD

projects/gloo/pkg/bootstrap/clients/consul_test.go
jbohanon: I brought the client into ConsulInstance and exposed a Put wrapper

projects/gloo/pkg/bootstrap/clients/consul_test.go
jbohanon: I removed the inner ctx

projects/gloo/pkg/bootstrap/clients/consul_test.go
jbohanon: Good call! Fixed

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -338,23 +338,42 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
...
-	var vaultClient *vaultapi.Client
-	if vaultSettings := settings.GetVaultSecretSource(); vaultSettings != nil {
-		vaultClient, err = bootstrap_clients.VaultClientForSettings(vaultSettings)
-		if err != nil {
-			return err
+	getVaultInit := func(vaultSettings *v1.Settings_VaultSecrets) bootstrap_clients.VaultClientInitFunc {
+		return func() *vaultapi.Client {
+			c, err := bootstrap_clients.VaultClientForSettings(vaultSettings)
+			if err != nil {
+				contextutils.LoggerFrom(ctx).Error(err)
```
sam-heilbron, line 345: Perhaps I'm missing something: previously we returned an error if the client construction failed. Now we log an error and continue. Am I understanding that correctly?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -140,4 +143,153 @@ var _ = Describe("Bootstrap Clients", func() {
 			})
 		})
 	})
+
+	Context("Secret Client", func() {
+		Skip("WIP")
```
sam-heilbron, line 148: I think this is causing a panic in the tests

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -338,23 +338,42 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
...
+	vaultSettings := settings.GetVaultSecretSource()
+	if vaultSettings != nil {
+		vaultInitMap[bootstrap_clients.SecretSourceAPIVaultClientInitIndex] = getVaultInit(vaultSettings)
+	}
+	if secretSources := settings.GetSecretOptions().GetSources(); secretSources != nil {
+		for i := range secretSources {
+			switch src := secretSources[i].GetSource().(type) {
+			case *v1.Settings_SecretOptions_Source_Vault:
+				vaultInitMap[i] = getVaultInit(src.Vault)
+			default:
```
sam-heilbron, line 360: nit: It may be valuable to call out why we don't handle the other types of secret sources here, as it's an intentional decision

sam-heilbron (commented):
Looking great!

projects/gloo/pkg/syncer/setup/setup_syncer.go
jbohanon: Yeah, that is correct. I'm not totally sure how I feel about this, but it was done because we now call the client initialization in vault.NewResourceClientFactory, which does not return an error. We could return nil and handle in the caller, but that seems to circumvent the purpose for that function not returning an error (I assume factories are supposed to be safe in that way). The other option would be to go into solo-kit and add a field to the vault client factory so we can accept a vault api client init func and/or check vault api client for nil on NewResourceClient and return an error.

OR... we could construct a client in the setup_syncer and error if an error occurs or throw it away if not ¯\_(ツ)_/¯

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -338,23 +339,53 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
...
+				// We are not concerned with source types that are not Vault since
+				// we are only getting the indices of Vault sources to accurately
+				// map the vault source to its API client setup func
+				continue
+			}
 		}
+		settings.GetSecretOptions().Sources = sourceList
```
sam-heilbron, line 375: I'm good with this since we don't persist the updated settings, and only gloo consumes this API. A potential challenge I can see is that discovery may consume this setting, and would need to duplicate this logic.

sam-heilbron (commented):
https://github.com/solo-io/gloo/blob/main/projects/discovery/pkg/uds/syncer/setup_syncer.go#L44 uses a secret client.Does this discovery code use the new solution (does it even need to)? I more want to unerstand the scope of this change, and if other components that use secret clients need (or definitely don't need it)

projects/gloo/pkg/syncer/setup/setup_syncer.go
jbohanon: The discovery packages call the same Setup function which runs this code and calls `constructOpts` which is the `bootstrap.Opts` that ends up getting passed around everywhere.

FDS: https://github.com/solo-io/gloo/blob/main/projects/discovery/pkg/fds/syncer/setup_syncer.go#L28
UDS: https://github.com/solo-io/gloo/blob/main/projects/discovery/pkg/uds/setup/setup.go#L16
Shared code where modifications in this PR reside: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/syncer/setup/setup_syncer.go

sam-heilbron (approved):
🚀 LGTM

elcasteel (approved):

PR #8270, opened 2023-05-22
Stability: Move Tests Which Adjust Validation Configuration to Same Context

sam-heilbron:
# Description

Place all tests which adjust the validation configuration of our Settings resource into the same context

# Context
**This just copies existing tests and moves them into a single context, and makes them Ordered**

When we change validation settings (or any setting), we must wait for the servers to start back up with the new configuration ([code](https://github.com/solo-io/gloo/blob/main/test/kube2e/util.go#L126)). There isn't an easy way to do this generically, so we add a sleep (3 seconds). 

For a lot of our tests in this space, we turn a setting on before the test, and off afterwards. This leads to a lot of flip-flopping of settings, when in reality, we're trying to test a set of behaviors when a setting is configured.

This work moves tests that are similar (ie validate the behavior of our validation config) into the same context, which gives us 2 wins:
1. We can easily see the test coverage in the space
2. We can run the context as an OrderedContainer (https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall) and only change the configuration once at the beginning of a group of tests. This will reduce test time (fewer sleeps) and hopefully reduce flakes that occur when the sleep isn't long enough.

# Checklist:

### DIFF
--- a/changelog/v1.15.0-beta8/gateway-kube2e-reorder.yaml
+++ b/changelog/v1.15.0-beta8/gateway-kube2e-reorder.yaml
@@ -1,0 +1,11 @@
+changelog:
+  - type: NON_USER_FACING
+    description: >-
+      Group tests which modify validation settings into the same context. This makes it easier to understand
+      what our existing test coverage is, as well as ensure we can run these tests in order.
+    issueLink: https://github.com/solo-io/gloo/issues/7795
+    resolvesIssue: false
+  - type: NON_USER_FACING
+    description: Begin the work to deflake our most painful kube2e tests
+    issueLink: https://github.com/solo-io/solo-projects/issues/4913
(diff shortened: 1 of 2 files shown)


sam-heilbron:
I encountered a failure (https://github.com/solo-io/gloo/actions/runs/5048970300/jobs/9057841084) of a test that was recently added. I will aim to deflake that in a separate PR

bewebi (approved):
Hard to follow the diff, and I didn't attempt to exhaustively verify each line was properly moved, but did try to find each block that was deleted in the new location and it all checked out
As promised, we minimize the number of times we need to update settings
Between that and CI passing, I approve

elcasteel (approved):

PR #8273, opened 2023-05-22
Stability: Deflake Gloo Kube2e Tests

sam-heilbron:
# Description

Update the Gloo Kube2e tests to use a similar pattern to what exists there

# Context

## De-Flake
https://github.com/solo-io/gloo/pull/8270#issuecomment-1557696904 I encountered a failure in a previous PR. The failure was related to a test which used to run in our long-lived build-bot cluster and was recently moved to our short-lived KinD clusters (https://github.com/solo-io/gloo/pull/8235)

The changes attempt to do 2 things:
1. Provide a consistent usage of ctx, cancel, which are long-lived in this test suite (managed by the suite itself)
2. Ensure the test which flaked follows a similar pattern to the test co-located with it that didn't flake (longer timeout, explicit refresh rate)

## Migrate Tests
We had some existing tests which require kube configuration (not necessary a live cluster) to work: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/clients/clients_test.go#L25. I migrated these tests to our kube2e package since we already have similar tests in there.

In doing so, I also noticed we had 2 test packages which ran the same tests:
- https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/clients/clients_test.go
- https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/utils_test.go

This was introduced when we improved the structure of this package (https://github.com/solo-io/gloo/pull/8261). The duplicate code was an effort to avoid breaking changes in our exported functions. However, we don't need to duplicate the tests because the same underlying code is used (deprecated functions just called preferred functions directly)

# Checklist:


test/kube2e/gloo/artifact_client_test.go
```diff
@@ -25,38 +24,35 @@ import (
 var _ = Describe("Artifact Client", func() {
 
 	var (
-		namespace     string
+		testNamespace string
```
sam-heilbron, line 27: The suite has a namespace variable at the suite level, so I just renamed these to make it clear

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	corecache "github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/cache"
+	"github.com/solo-io/solo-kit/test/helpers"
+	kubev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	. "github.com/solo-io/gloo/projects/gloo/pkg/bootstrap/clients"
```
jbohanon, line 20: nit: I know this is a holdover with the old file, but I think we lose context here by namespacing into a package that's not co-located with the test

test/kube2e/gloo/kubernetes_test.go
```diff
@@ -185,13 +174,27 @@ var _ = Describe("Plugins", func() {
...
+		case endpointList := <-endpoints:
+			Expect(endpointList).To(HaveLen(6))
+			break
+		case err, ok := <-errs:
+			if !ok {
+				return
+			}
+			Expect(err).NotTo(HaveOccurred())
+		}
 		Consistently(errs).Should(Not(Receive()))
```
jbohanon, line 198: Does the `Consistently` still provide value here? If we timeout, we failed; if we received on `<-endpoints` we succeeded, and if we receive on `<-errs` we have already checked for nil?

test/kube2e/gloo/resource_client_test.go
```diff
@@ -1,66 +1,56 @@
...
-		ctx       context.Context
-		cancel    context.CancelFunc
+		testNamespace string
+
+		kubeClient kubernetes.Interface
 	)
 
 	BeforeEach(func() {
-		namespace = helpers.RandString(8)
 		var err error
```
jbohanon, line 27: nit: moving this closer to where it's used

jbohanon (commented):
Just a few nits

test/kube2e/gloo/bootstrap_clients_test.go
sam-heilbron: yeah good call, I'll update

test/kube2e/gloo/kubernetes_test.go
sam-heilbron: Good point Ill double check. My understanding is that errs is a channel which returns any errors encountered during endpoint discovery. So theoretically if we had a failure on the first try (and then on a second try it worked), we would both get endpoints and errors. I think the extra assertion is valuable given that it demonstrates the behaviors of each of the returned variables. What do you think?

jbohanon: Could wrap the err `chan`s in buffered channels for the testing and check those

sam-heilbron: You're right, the switch statement should catch this case, so I removed it

sam-heilbron: Looks like my window hadn't refreshed with your updates when I posted that. What makes you think the switch statement isn't validating the errors channel properly?

bewebi (approved):
LGTM

jbohanon (approved):

PR #8274, opened 2023-05-22
Stability: Deflake Glooctl Kube2e Tests

sam-heilbron:
# Description

Move tests which depend on k8s cluster to the suites which use them

# Context
## glooctl debug [...]
Slack context: https://solo-io-corp.slack.com/archives/C056AFWPB0X/p1684787070973559

## glooctl istio [...]
These tests were the ones that flaked. They use resources that other tests in this suite don't need, so to simplify things, I did the following:
- I moved these tests to their own file
- The setup that is required onetime is done before all the tests, and undone after all the tests (previously is was before/after each test)
- Instead of trying to reuse setup for positive and negative test cases, I split out the failure to their own tests

# Checklist:


test/kube2e/glooctl/glooctl_test.go
```diff
@@ -535,21 +343,10 @@ var _ = Describe("Kube2e: glooctl", func() {
 			Expect(err.Error()).To(ContainSubstring("One or more CRDs are out of date"))
 		})
 	})
+
 })
 
 // runGlooctlCommand take a set of arguments for glooctl and then executes local glooctl with these arguments
 func runGlooctlCommand(args ...string) (string, error) {
```
Bslabe123, line 350: Why not just remove `runGlooctlCommand` altogether and just replace all `runGlooctlCommand` calls with `GlooctlOut` calls?

sam-heilbron: I was thinking about that. I was trying to keep the scope of this PR limited to the commands I was working on (glooctl debug, glooctl istio). Given that I was trying to set a standard of putting all tests about a command in a file that matches that command, I figured I could either do all the tests this way, or just the tests I'm changing.

I'm more than happy, after this PR merges, to update the other tests to follow this pattern (and use the new GlooctlOut method).

The reason I moved the function too, was so that it was clear that function could be used throughout the entire suite. Since it is now used by multiple files, I figured having it defined in a central place (either the suite file, or some utils file) would be clearer for others to find

bewebi (approved):
LGTM

Bslabe123 (approved):
Like the additional cleanup done by the EventuallyIstioUninjected() function, LGTM!

PR #6619, opened 2022-06-23
CI Improvements

sam-heilbron:
# Description

A set of minor optimizations to our CI pipeline, and other related bug fixes
# Context

## Envoy Runner
Gloo defines all of the code for running an instance of Envoy. Solo-Projects re-defines this logic verbatim. To enable this code to be imported, we move the logic into a Runner package, and out of the main package.

This is a non-functional change, but follows a good practice of keeping main as slim as possible.

## CI: Docs Gen
Do not run the docs generation GitHub action on Draft PRs. This is a fairly expensive operation (20 minutes) that only needs to run once a PR is ready for review.

**As part of this, I also added support to trigger and skip this action based on a comment. Since the `on comment` trigger only applies from the default branch, I can't validate this trigger. However, I can validate (by modifying the script) that we succeed the job if configured to auto-pass and skip the job if configured to not run.**

## CI: Regression Tests
Do not run the regresssion test GitHub action on Draft PRs. We should rely on local development to run tests, and once a PR is ready for review, these tests will run.

**As part of this, I also added support to trigger and skip these tests based on a comment. Since the `on comment` trigger only applies from the default branch, I can't validate this trigger. However, I can validate (by modifying the script) that we succeed the job if configured to auto-pass and skip the job if configured to not run.**

## CI: Notification on Docs Deploy Failure
When docs gen fails, we attempt to notify the team in a channel. Based on a conditional, we either push to a testing channel or a production channel. That conditional needed to be fixed, since docs build failures (on merges to master) were being published to the testing repo.

## CI: Readme
Updated the readme to call out explicitly that certain jobs will not run on Draft PRs.

## CI: Regression Failure User List
We maintain a list of slack handles for users who want to be notified directly on Github action failures. This list needed to be updated to reflect the current team.

# Warning
After a pull request is converted from a draft to 'ready for review', the CI will not run the docs gen or regression tests. Those must be signaled using `/sig-ci` and `/sig-docs`. That signal functionality has not beeen fully tested (and cannot be until the code is on the default branch). In the case that those signals do not work, developers will be required to push up an empty commit after the pull request is ready for review, to re-trigger all actions.

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
[...]

PR #8278, opened 2023-05-23
build(release): move to GHA from cloudbuild

gunnar-solo:
# Description
A reasonably expansive set of changes which moves this repository's release process away from google cloud compute and fully onto Github Actions

Notable Changes:
`Makefile`
- removed several layers of indirection.  Previously used variables: (`VERSION`, `TAGGED_VERSION`, `CREATE_ASSETS`, `CREATE_TEST_ASSETS`, `TEST_ASSET_ID`, `ASSETS_ONLY_RELEASE`, `ON_DEFAULT_BRANCH`, `CHECKED_OUT_SHA`, `DEFAULT_BRANCH_NAME`, `EMPTY_IF_NOT_DEFAULT`, `ORIGIN_URL`, `UPSTREAM_ORIGIN_URL`, `UPSTREAM_ORIGIN_URL_HTTPS`, `UPSTREAM_ORIGIN_URL_SSH`) have all been consolidated into (`PUBLISH_CONTEXT`, `VERSION`)
- removed targets which I considered unneeded (`print-git-info`, `push-chart-to-registry`, manifest test targets)
- only define release targets when `PUBLISH_CONTEXT` is set--optimally, preventing devs from accidentally running publish targets from their local machine

`ci/cloudbuild/publish-artifacts.yaml`
- removed this file, as was the original intention of the PR

`hack/*`
- added several `go` helpers centered around `VERSION` and `CHANGELOG` management

`.github/workflows/release.yaml`
- added 5 stages of compute: (`compute-env-vars`) --> (`images`, `charts`, `cli`) --> (`tag`)

# Testing
Extensive testing will be required.  Even so, I'm anticipating needing to make some ongoing hotfixes to `gloo` and `solo-projects` once this merges.  To be tested (prior to merge):
- [ ] validate a formal release
- [ ] validate PR-generated release assets
- [ ] validate a community user PR (PR from a fork)

In addition, I'm aware of a testing suite _explicitly_ designed around the previous complex shape of variable computation in the `Makefile`.  Clearly, this is going to fail every time, until I remove it.  I'm leaving it alone for the moment, as I want that to be a kind of safety-net preventing early merge

# Checklist:


Makefile
```diff
@@ -725,30 +578,54 @@ endif
...
-publish-docker: docker
-publish-docker: docker-push
-endif
+.PHONY: fetch-package-and-save-helm
+
+# don't define Publish Artifacts Targets if we don't have a release context
+ifneq (,$(filter $(PUBLISH_CONTEXT),RELEASE PULL_REQUEST))
+
+HELM_BUCKET     := gs://solo-public-helm
+GCR_HELM_BUCKET := oci://gcr.io/solo-public/gloo-helm
```
bewebi, line 594: This does not appear to be used
We have removed the `push-chart-to-registry` target which previously pushed to this bucket, so I would guess that we need some other mechanism to do so...or if we never need to push to this bucket, we don't need an env var for it

Makefile
```diff
@@ -725,30 +578,53 @@ endif
...
-ifeq ($(CREATE_ASSETS), "true")
-publish-docker: docker
-publish-docker: docker-push
-endif
+.PHONY: fetch-package-and-save-helm
+
+# don't define Publish Artifacts Targets if we don't have a release context
+ifneq (,$(filter $(PUBLISH_CONTEXT),RELEASE PULL_REQUEST))
+
+HELM_BUCKET     := gs://solo-public-helm
```
sam-heilbron, line 593: I've found it's useful for clarity, even at the expense of brevity, to list all the variables which affect publishing assets: IMAGE_REGISTRY, QUAY_EXPIRATION_LABEL...etc. And if any are empty we can show them to be. It's then a single place for us to look and see what the values are

Makefile
```diff
@@ -725,30 +578,53 @@ endif
...
 # and push them to a secondary repository, defined at IMAGE_REGISTRY
-.PHONY: publish-docker-retag
-ifeq ($(RELEASE), "true")
-publish-docker-retag: docker-retag
-publish-docker-retag: docker-push
-endif
+publish-docker-retag: docker-retag docker-push
+
+# publish glooctl
+upload-github-release-assets: build-cli
```
sam-heilbron, line 607: nit: can we rename this `publish-glooctl` since that's what it does?

Makefile
```diff
@@ -725,30 +578,53 @@ endif
...
+	GO111MODULE=on go run ci/upload_github_release_assets.go true
+endif # RELEASE exclusive make targets
+
+
+# Build and push docker images to the defined $(IMAGE_REGISTRY)
+publish-docker: docker docker-push
+
+# create a new helm chart and publish it to $(HELM_BUCKET)
+fetch-package-and-save-helm: generate-helm-files
```
sam-heilbron, line 616: nit: Can we call this `publish-helm-chart`

Makefile
```diff
@@ -44,85 +44,8 @@ IMAGE_REGISTRY ?= quay.io/solo-io
...
-# CREATE_TEST_ASSETS allows us to create assets on PRs that are unique
-# This flag will set the version to be PR-unique rather than commit-unique for charts and images
-CREATE_TEST_ASSETS := "false"
-ifneq ($(TEST_ASSET_ID),)
-	CREATE_TEST_ASSETS := "true"
-endif
-
-# ensure we have a valid version from a forked repo, so community users can submit PRs
-ORIGIN_URL ?= $(shell git remote get-url origin)
-UPSTREAM_ORIGIN_URL ?= git@github.com:solo-io/gloo.git
```
sam-heilbron, line 61: Why are these deletions safe? We have some code (and tests) that limit behavior for forked repos. How do we manage running code from a fork?

.github/workflows/docs-gen.yaml
```diff
@@ -63,10 +64,14 @@ jobs:
           echo "Should auto-pass docs? $should_pass"
           echo "build_value=$should_build" >> $GITHUB_ENV
           echo "pass_value=$should_pass" >> $GITHUB_ENV
+      - id: version
+        run: echo "value=GO111MODULE=on go run hack/compute_version/main.go" >> $GITHUB_OUTPUT
```
sam-heilbron, line 68: Are version and publish context tightly coupled? I had imagined that compute_version, would return us a major.minor.patch-label to represent the version. We then pass this to our PUblish_context which I thought would be NONE/PR/RELEASE. Am I missing something?

Makefile
```diff
@@ -44,85 +44,8 @@ IMAGE_REGISTRY ?= quay.io/solo-io
...
-    QUAY_EXPIRATION_LABEL := --label "quay.expires-after=3w"
-  endif
-else
-  ifeq ($(RELEASE), "true")
-  else
-    CREATE_ASSETS := "false"
-  endif
-endif
-
+VERSION 	?= 0.0.1-dev    # a basic version that can be overidden
```
sam-heilbron, line 47: nit: I think I've seen semver struggle when comparing versions <1. I had named our CI default version 1.0.0-ci to try to avoid this. if you look in the github commit history I may have included in the PR where I introduced this, the context behind it

hack/compute_version/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/solo-io/go-utils/changelogutils"
+)
+
+const (
+	repoRootPath = "."
+	repo         = "gloo"
+	owner        = "solo-io"
+)
+
+func main() {
```
sam-heilbron, line 18: Will we need to use this in enterprise? In general, I've found it's best to limit the size of the main funciton, and have it import code. I had done something like this in https://github.com/solo-io/gloo/pull/6619

hack/compute_version/main.go
```diff
@@ -0,0 +1,35 @@
...
+	ctx := context.Background()
+
+	// compute largest local changelog version
+	version, err := helpers.GetLargestLocalChangelogVersion(ctx, repoRootPath, owner, repo, changelogutils.ChangelogDirectory)
+	if err != nil {
+		panic(err)
+	}
+
+	PR_NUMBER := os.Getenv("PR_NUMBER")
```
sam-heilbron, line 27: I found that it's hard to track env variables and where they are used. As a result, I defined any variables that our tests use in a single file: https://github.com/solo-io/gloo/blob/main/test/testutils/env.go#L8. Perhaps we do another one for our CI pipeline?

hack/find_latest_enterprise_version/main.go
```diff
@@ -0,0 +1,45 @@
...
+	"math"
+	"os"
+
+	"github.com/solo-io/gloo/hack/helpers"
+	"github.com/solo-io/gloo/pkg/version"
+	"github.com/solo-io/go-utils/changelogutils"
+	"github.com/solo-io/go-utils/versionutils"
+)
+
+func main() {
```
sam-heilbron, line 15: Same comment as above, limiting scope of main function. I also think this function warrants a comment outlining what we're doing and why. At first glance, it's not clear why open source is even aware of enterprise. since it's a bit of an anti-pattern, I think we should document the rationale

.github/workflows/release.yaml
```diff
@@ -0,0 +1,123 @@
+name: Publish Resources
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
gunnar-solo: my intention surrounding this action is that we _don't_ ever create releases on our own.  Rather, we would manually trigger the release action (this) which would create a release for us.

This has the benefit of being able to compute/set release notes at runtime, rather than as an async operation

Makefile
gunnar-solo: the [current state-of-affairs](https://github.com/solo-io/gloo/blob/20ef52e64e654f2c85412ca8cfbcca8ad881895a/Makefile#L59-L67) has this set of variables being used to compute

```
	VERSION ?= 0.0.1-fork
	CREATE_TEST_ASSETS := "false"
```

which, later on, sets `CREATE_ASSETS := "true"`

as far as I can tell, these variables aren't referenced in any (relevant) tests

gunnar-solo: with how I've defined `Makefile` target guards, currently, community-user PRs would call the subset of release targets designated for "regular PR builds"

not a large issue, in my mind, given the frequency of said runs

.github/workflows/release.yaml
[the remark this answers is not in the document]
sam-heilbron: I'm not sure I follow. Currently a user manually creates a github release, and that event is picked up by our built-bot to publish assets. I'm proposing that we keep that paradigm: user creates a release, and this job picks up that even (on release) and publishes the assets. You can see https://github.com/solo-io/ext-auth-service/blob/master/.github/workflows/release.yaml#L3 for an example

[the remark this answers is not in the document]
gunnar-solo: probably best to talk this out in a huddle &&/|| at our review in an hour -- we definitely _can_ do this, but that means editing release notes of an existing release, rather than writing them correctly the first time

not insurmountable, but it makes it so we can't use the [nice release action](https://github.com/ncipollo/release-action)

.github/workflows/docs-gen.yaml
gunnar-solo: all of the things that have been said in this 🧵 are correct, save for this bit, which I'm fuzzy on:

> We then pass this to our PUblish_context

probably better to handle in a huddle as well

.github/workflows/release.yaml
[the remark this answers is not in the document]
gunnar-solo: to be addressed by polling the team/soliciting feedback

PR #8280, opened 2023-05-23
Stability: Deflake Gateway Kube2e Tests

sam-heilbron:
# Description
Ensure that Gateway Kube2e tests are less flakey

# Context

## Gateway tests: AlwaysAccept > with a mix of valid and invalid virtual services 
I noticed that the second test is the one that always fails. And it fails, not because we reject resources or don't process them, but because the resources haven't been processed by our controller and thus there is no status.

First I explored if this was due to leader election, but the logs indicated that there was no issue with lease handling.

My theory is that if we delete a resource (after the previous test finishes) and apply that same resource (for the new test) really quickly, then the API snapshot according to Gloo is no different. When this happens the Gloo controllers do not run a sync (and therefore don't update statuses)

To test this, I added a unique value to the resources we apply, and increment that for each test. I then ran these tests in parallel on a bunch of test clusters. Many tests passed in a row (see test section below), so I'm fairly confident in the fix.

## Robustness tests: Updates Envoy endpoints, even if proxy is invalid
There was a failure where we expected a list of IPs to be empty. This was because we first gather "previous" IPs and then compare IPs after we scale down and up a service. Our initial IPs should never be empty, so I added the guard to prevent that.

# Testing
I adjusted the regression tests to run this suite only, in all clusters ([commit-ref](https://github.com/solo-io/gloo/pull/8280/commits/f3ea4346f48dbf3c5272776cdefa0396a8398d13)). This way I could sees this test pass 8 times in parallel. 

I ran that a few times just to be safe:
[First iteration](https://github.com/solo-io/gloo/actions/runs/5063551620)
[Second iteration](https://github.com/solo-io/gloo/actions/runs/5063686839)

# Checklist:


jackstine (dismissed):
LGTM

test/kube2e/gateway/gateway_test.go
```diff
@@ -2146,7 +2146,10 @@ spec:
 		Context("AlwaysAccept", Ordered, func() {
 			// We want to test behaviors when Gloo allows invalid resouces to be persisted
 
+			var uniqueSuffix int
+
 			BeforeAll(func() {
+				uniqueSuffix = 1
```
inFocus7, line 2152: nit/thought: the only issue i could _possibly_ see is should we also take into account the possibility of parallizing these tests in the future? In a parallel context, i'm wonder if this could lead to issues.

test/kube2e/gateway/gateway_test.go
```diff
@@ -2194,7 +2203,9 @@ spec:
 				JustBeforeEach(func() {
 					// ensure that we have successfully gotten into an invalid state
 					helpers.EventuallyResourceRejected(func() (resources.InputResource, error) {
-						return resourceClientset.VirtualServiceClient().Read(testHelper.InstallNamespace, invalidVsName, clients.ReadOpts{})
+						return resourceClientset.VirtualServiceClient().Read(testHelper.InstallNamespace, invalidVsName, clients.ReadOpts{
+							Ctx: ctx,
```
ben-taussig-solo, line 2207: I'm assuming that this is unrelated to the fix, just something we missed when the test was initially put up?

test/kube2e/gateway/robustness_test.go
```diff
@@ -235,9 +235,13 @@ var _ = Describe("Robustness tests", func() {
 			By("force proxy into warning state")
 			forceProxyIntoWarningState(appVs)
 
-			By("force an update of the service endpoints")
-			initialEndpointIPs := endpointIPsForKubeService(resourceClientset.KubeClients(), appService)
+			var initialEndpointIPs []string
+			Eventually(func() []string {
+				initialEndpointIPs = endpointIPsForKubeService(resourceClientset.KubeClients(), appService)
+				return initialEndpointIPs
+			}, 20*time.Second, 1*time.Second).Should(Not(HaveLen(0)))
```
inFocus7, line 242: wicked smart eventually

test/kube2e/gateway/gateway_test.go
sam-heilbron: Yeah great point, I was hoping to ensure the ID was something usable (if a test failed), but since the tests run in order, we can infer what tests ran before. The GinkgoParallelProcess method is intended to ensure safe parallelization, how about I add that to this suffix to ensure we have uniqueness?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2160,11 +2165,16 @@ spec:
...
-					validVsName   = "i-am-valid"
-					invalidVsName = "i-am-invalid"
+					validVsName, invalidVsName string
 				)
 
 				BeforeEach(func() {
+					// For each test, uniquely identify the resources using a suffix that is unique per test
+					// We quickly apply and delete the same values, so we want to ensure a new snapshot (according to Gloo) is created
+					uniqueSuffix += 1
+					validVsName = fmt.Sprintf("i-am-valid-%d-%s", uniqueSuffix, parallel.GetParallelProcessCount())
```
ben-taussig-solo, line 2175: nit: why not `%d` for the parallel process count format string, given that `GetParallelProcessCount` returns an int?

ben-taussig-solo: Same question applies to the line below

sam-heilbron: ha good catch, the test failed. I missed that. Updated

inFocus7 (approved):

ben-taussig-solo (approved):

PR #8298, opened 2023-05-25
Performance tests job in nightly action

bewebi:
# Description

- Add a `run-performance-tests` Make target
- Add a performance-test composite GitHub action which invokes the `run-performance-tests` target
- Add `performance_tests_latest_main` job to `nightly-tests` action, which invokes the performance-tests composite action on the same schedule as the latest nightly main regression tests
- Other naming/formatting cleanup
  - Remove all labels from "example" test, and instead skip it, with a message clarifying that the test is meant as a useful example for developers writing new tests, but does not need to be run regularly 

# Context

The new action compiles _all_ packages and runs only tests with the `performance` label

Currently there are no actual tests with this label, however there is one test which intentionally fails, present as a smoke test to ensure it is being run in the nightly context and not otherwise
I think we can safely remove this test now that manual testing has been completed, with the action behaving as expected and running only the intentionally-failing test (see run [here](https://github.com/solo-io/gloo/actions/runs/5085107291/jobs/9138218895))

The action takes significant time to run (>20m) which is largely due to drawing tests from all packages rather than specifying a particular package
This approach is taken because moving forward we would like to be able to write performance/nightly tests in any package and designate them for nightly runs via label
The likely path to running this action within a reasonable timeframe will be to use a 4 or 8 core runner instead of the default, which has 2
This is not considered a blocker as it can be done without a code change and because this is a nightly action with does not slow down any developer's processes
See [this doc](https://docs.google.com/document/d/1PCKzZ1w8vfdawGM1XbuFKCLw1Fot_e47h2WkuPZnBsw/edit) for some additional context on how runtimes might be improved

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5003


.github/workflows/nightly-tests.yaml
```diff
@@ -28,6 +28,14 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/workflows/composite-actions/regression-tests
+  nightly_tests_latest_dev:
```
sam-heilbron, line 31: nit: Can we rename all these instances of "dev" to "main"? They run what's on the main branch and I think that's just a clearer name

Makefile
```diff
@@ -253,6 +253,10 @@ run-tests: GINKGO_FLAGS += -skip-package=e2e ## Run all non E2E tests, or only r
 run-tests: GINKGO_FLAGS += --label-filter="!end-to-end && !nightly"
 run-tests: test
 
+.PHONY: run-nightly-tests
```
sam-heilbron, line 256: I had introduced these specific targets as a way to easily group targets that are frequently run. What do you think about not defining this in the makefile, and just running the `test` target direcetly, with the proper label filter and package filter

.github/workflows/composite-actions/nightly-tests/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Gloo Nightly Tests
+description: Runs tests with the nightly label
+  
+runs:
+  using: "composite"
+  steps:
+  - name: Testing - nightly tests
```
sam-heilbron, line 7: Is this name something that affects our output logs/file names? Or is this just an internal name?

.github/workflows/composite-actions/nightly-tests/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Gloo Nightly Tests
```
sam-heilbron, line 1: Since we're in the GLoo repo, I think we can remove the "gloo". Also, since we run other tests at night (regression tests), we  might want to figure out a better name to differentiate

sam-heilbron (commented):
I like it!  A few nits

.github/workflows/composite-actions/nightly-tests/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+  using: "composite"
+  steps:
+  - name: Testing - nightly tests
+    shell: bash
+    run: make install-test-tools run-nightly-tests
+  - name: save results
+    if: ${{ failure() }}
+    shell: bash
+    run: mkdir -p _output/nightly-artifacts && echo '{"url":"https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}",
+            "name":"nightly"}' > _output/nightly-artifacts/test-out.json
```
gunnar-solo, line 14: similar to other comments:  `"name":"nightly"` is going to give reasonably obtuse output when appearing in a slack bot post

.github/workflows/composite-actions/nightly-tests/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+runs:
+  using: "composite"
+  steps:
+  - name: Testing - nightly tests
+    shell: bash
+    run: make install-test-tools run-nightly-tests
+  - name: save results
+    if: ${{ failure() }}
+    shell: bash
+    run: mkdir -p _output/nightly-artifacts && echo '{"url":"https://github.com/solo-io/gloo/actions/runs/${{github.run_id}}",
```
gunnar-solo, line 13: similar to other comments:  `"name":"nightly"` is going to give reasonably obtuse output when appearing in a slack bot post

.github/workflows/nightly-tests.yaml
```diff
@@ -28,6 +28,14 @@ jobs:
...
+    name: latest nightly tests
+    if: github.event_name == 'workflow_dispatch' || github.event.schedule == '0 5 * * 1-5'
```
gunnar-solo, line 33: not an _error_:  just callling out that this run condition makes this job _always_ run at the same time as the regression suite on `main`

.github/workflows/nightly-tests.yaml
gunnar-solo, line 31: nit: newline between different jobs for legibility

.github/workflows/composite-actions/nightly-tests/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Gloo Nightly Tests
+description: Runs tests with the nightly label
+  
+runs:
+  using: "composite"
+  steps:
+  - name: Testing - nightly tests
+    shell: bash
+    run: make install-test-tools run-nightly-tests
```
gunnar-solo, line 9: help me understand what runs when this is called?  I'm trying to understand the implications on our notification system.  Specifically:  if it plays nicely with running at the same time as all of the matrix targets from our regression suite

Makefile
bewebi: I prefer the target, since it allows us to easily change the way we run nightly tests from a single source of truth (ie if we want to specify particular packages, modify labels, etc.)

.github/workflows/composite-actions/nightly-tests/action.yaml
bewebi: Just the name of the step as seen here
IMG_0773

.github/workflows/composite-actions/nightly-tests/action.yaml
bewebi: See the Make target introduced in this PR

Since we output results to a directory under `_output` with the name `test-out.json` it is picked up pleanly

See [this notification](https://solo-io-corp.slack.com/archives/C0314KESVNV/p1685054766747199) for example

.github/workflows/nightly-tests.yaml
bewebi: I think that’s a reasonable approach, but open to hearing suggestions for how it might be better

.github/workflows/composite-actions/nightly-tests/action.yaml
bewebi: Agreed, I was going off the label name…perhaps that should (also) be changed to “performance” if that’s really what we’re hoping to encompass with this job

gunnar-solo (dismissed):

.github/workflows/nightly-tests.yaml
```diff
@@ -79,7 +89,7 @@ jobs:
           elif [[ ${{github.event.schedule == '10 10 * * 1'}} = true ]]; then
               preamble="Gloo OSS weeklies (v1.14.x)"
           elif [[ ${{github.event.schedule == '0 5 * * 1-5'}} = true ]]; then
-              preamble="Gloo OSS nightlies (dev)"
+              preamble="Gloo OSS nightlies (main)"
```
sam-heilbron, line 92: in our published slack channel, won't this mean that we have a notification titled "Gloo OSS nightlies (main) twice: once for our regression tests and one for our performance tests?

test/e2e/example_test.go
```diff
@@ -27,6 +25,8 @@ var _ = Describe("Example E2E Test For Developers", Label(labels.Nightly), func(
 	)
 
 	BeforeEach(func() {
+		Skip("this test is intended as a template for developers and does not need to be run in CI")
```
sam-heilbron, line 28: Opened a thread in slack. I think we want this running, but not sure where the appropriate place is. Happy to handle this in a follow-up

sam-heilbron (approved):

.github/workflows/nightly-tests.yaml
bewebi: Nope, it's one notification per run of the action
See [this notification](https://solo-io-corp.slack.com/archives/C0314KESVNV/p1685054766747199) for example, with both k8s and performance (then called nightly) failing

test/e2e/example_test.go
bewebi: Slack conversation [here](https://solo-io-corp.slack.com/archives/C056AFWPB0X/p1685473045130899?thread_ts=1685464746.879489&cid=C056AFWPB0X)

sam-heilbron (approved):

gunnar-solo (approved):

PR #8299, opened 2023-05-26
build(Makefile): removed several layers of configuration vars

gunnar-solo:
# Description
supporting PR for https://github.com/solo-io/gloo/pull/8278.  It is meant to consist of all of the changes that _can_ be done, prior to credential access.  So, in the spirit of that, here is a ported description:

`Makefile`
- removed several layers of indirection.  Previously used variables: (`VERSION`, `TAGGED_VERSION`, `CREATE_ASSETS`, `CREATE_TEST_ASSETS`, `TEST_ASSET_ID`, `ASSETS_ONLY_RELEASE`, `ON_DEFAULT_BRANCH`, `CHECKED_OUT_SHA`, `DEFAULT_BRANCH_NAME`, `EMPTY_IF_NOT_DEFAULT`, `ORIGIN_URL`, `UPSTREAM_ORIGIN_URL`, `UPSTREAM_ORIGIN_URL_HTTPS`, `UPSTREAM_ORIGIN_URL_SSH`) have all been consolidated into (`PUBLISH_CONTEXT`, `VERSION`)
- removed targets which I considered unneeded (`print-git-info`, `push-chart-to-registry`, manifest test targets)
- only define release targets when `PUBLISH_CONTEXT` is set--optimally, preventing devs from accidentally running publish targets from their local machine

`hack/*`
- added several `go` helpers centered around `VERSION` and `CHANGELOG` management

# Testing:
[PR from fork](https://github.com/solo-io/gloo/pull/8304)
[release from gloo-private](https://solo-io-corp.slack.com/archives/CFK0DQADU/p1685126195023509)

# Checklist:


Makefile
```diff
@@ -725,30 +578,70 @@ endif
...
+VERSION := $(shell echo $(TAGGED_VERSION) | cut -c 2-)
+LDFLAGS := "-X github.com/solo-io/gloo/pkg/version.Version=$(VERSION)"
+endif
 
-.PHONY: upload-github-release-assets
-upload-github-release-assets: print-git-info build-cli render-manifests
-	GO111MODULE=on go run ci/upload_github_release_assets.go $(ASSETS_ONLY_RELEASE)
+# Possible Values: NONE, RELEASE, PULL_REQUEST
+PUBLISH_CONTEXT       ?= NONE                   # controller variable for the "Publish Artifacts" section.  Defines which targets exist.
+VERSION               ?=                        # a semver resembling 1.0.1-dev.  Most calling jobs customize this.  Ex:  v1.15.0-pr8278
```
sam-heilbron, line 597: nit: Any reason we don't default this to a value, like we do in the docs Makefile? I like how you default the context to None (ie local), I wonder if we just default the version to 1.0.0-local

Makefile
```diff
@@ -725,30 +578,70 @@ endif
...
+HELM_BUCKET           := gs://solo-public-helm	# specify which bucket to upload helm chart to
```
sam-heilbron, line 598: Should this be `?=` since we're defaulting a value and want it to be overridden

sam-heilbron: Also, I think it would be nice if our defaults were consistent. For publish-context we default to as if you're a local dev. But the helm_bucket points to our prod instance. Should we just default all these to be the same "context"?

Makefile
```diff
@@ -725,30 +578,70 @@ endif
...
+QUAY_EXPIRATION_LABEL :=                        # modifier to docker builds which can auto-delete docker images after a set time
 
-# Intended only to be run by CI
-# Build and push docker images to the defined IMAGE_REGISTRY
+# define empty publish targets so calls won't fail
```
sam-heilbron, line 601: Should we follow the same pattern as below, and have these targets exist if publish_context = None?

Makefile
```diff
@@ -725,30 +578,70 @@ endif
...
-endif
+.PHONY: publish-helm-chart
+
+# don't define Publish Artifacts Targets if we don't have a release context
+ifneq (,$(filter $(PUBLISH_CONTEXT),RELEASE PULL_REQUEST))
 
-# Intended only to be run by CI
+GO111MODULE=on go run hack/compute_version/main.go
+
+ifeq (PULL_REQUEST, $(PUBLISH_CONTEXT)) # PULL_REQUEST contexts have different configs for publishing
```
sam-heilbron, line 612: Do we publish images, helm chart and glooctl on PRs? We used to do that, and perhaps I'm just missing the logic to do that

ci/cloudbuild/publish-artifacts.yaml
```diff
@@ -69,10 +69,8 @@ steps:
   id: 'release-chart'
   dir: *dir
   args:
-  - 'fetch-package-and-save-helm'
-  - 'render-manifests'
-  - 'upload-github-release-assets'
-  - 'push-chart-to-registry'
+  - 'publish-helm-chart'
```
sam-heilbron, line 72: nice!

hack/compute_version/main.go
```diff
@@ -0,0 +1,15 @@
...
+
+	github_action_utils "github.com/solo-io/gloo/pkg/github-action-utils"
+)
+
+func main() {
+	version, err := github_action_utils.GetVersion(".", "gloo", "solo-io")
+	if err != nil {
+		panic(err)
+	}
+	fmt.Print(version)
```
sam-heilbron, line 14: Is the print here important? I notice in find_latest_enteprirse_version we don't have the same logic. So i would propose that either we remove it if it's not necessary, or if it is, we add a comment about why it's important

pkg/github-action-utils/env.go
```diff
@@ -0,0 +1,8 @@
+package github_action_utils
+
+const (
+	// PR_NUMBER is a unique identifier of a given pull request for a given github repository.
+	// Consider: https://github.com/solo-io/gloo/pull/8278
+	// yields the PR_NUMBER: 8278
+	PR_NUMBER = "PR_NUMBER"
```
sam-heilbron, line 7: I like it!

pkg/github-action-utils/version.go
```diff
@@ -1,28 +1,45 @@
...
 	"github.com/solo-io/gloo/pkg/version"
 	"github.com/solo-io/go-utils/changelogutils"
 	"github.com/solo-io/go-utils/versionutils"
 	"github.com/solo-io/go-utils/vfsutils"
 )
 
-func main() {
+// GetVersion computes an appropriate semver string based on the contents of
+// the ./changelog directory fount at repoRootPath
+func GetVersion(repoRootPath string, repo string, owner string) (string, error) {
```
sam-heilbron, line 17: Why do we need to compute the version? In the past, we used:

```
 VERSION ?= $(shell git describe --tags --abbrev=0 | cut -c 2-)-$(TEST_ASSET_ID)
	else
	  VERSION ?= $(shell git describe --tags --dirty | cut -c 2-)
```

to compure the version. It feels like that will be less error prone (since we just rely on github api)

Makefile
gunnar-solo: VERSION is defaulted at the top of the `Makefile` with `VERSION 	?= 1.0.1-dev    # a basic version that can be overidden`

I mentioned it again here not to be computationally useful, but so a future observer can see all the relevant release variables listed in the same place

is the preference to define it here?  At the top, only?  Both?

Makefile
gunnar-solo: There's a disconnect here which I don't _quite_ get?  Currently, these stubs are _always_ defined, even if `PUBLISH_CONTEXT=None`

Makefile
gunnar-solo: sure!  The logic is fiddly/a bit annoying as a workaround to `Makefile`'s not supporting `or` operations:

```
ifneq (,$(filter $(PUBLISH_CONTEXT),RELEASE PULL_REQUEST))
ifeq (RELEASE, $(PUBLISH_CONTEXT))
    # targets/variables for just RELEASE goes here.  Ex:
    # publish-docker-retag, publish-glooctl, and redefining HELM_BUCKET
endif
    # targets/variabls for both RELEASE and PULL_REQUEST goes here.  Ex:
    # publish-docker and publish-helm-chart
endif
```

hack/compute_version/main.go
gunnar-solo: this is currently the bridge between go and the bash process calling it in `docs-gen.yaml`

In particular: `run: echo "value=$(go run hack/compute_version/main.go)" >> $GITHUB_OUTPUT`

(added comment)

pkg/github-action-utils/version.go
gunnar-solo: the practice I've tried to adopt when writing this code:

to set a version, pass it with your make call.  Ex:  `VERSION=v1.15.0-beta8 make publish-docker`.  Otherwise, you hit a default.  I came to this decision based on the complexity we had around nesting variables and using computed variables to compute other computed variables

---

in contrast with this pattern:  we _totally can_ have our Makefile manage this logic--I've chosen to not, however

Makefile
```diff
@@ -44,85 +44,8 @@ IMAGE_REGISTRY ?= quay.io/solo-io
...
-    QUAY_EXPIRATION_LABEL := --label "quay.expires-after=3w"
-  endif
-else
-  ifeq ($(RELEASE), "true")
-  else
-    CREATE_ASSETS := "false"
-  endif
-endif
-
+VERSION 	?= 1.0.1-dev    # a basic version that can be overidden
```
bewebi, line 47: Redefined/reset on 598, seemingly making this assignment redundant

pkg/github-action-utils/version.go
bewebi: I'm not sure I get the argument for going with the one-liner(s) over the Go script(s)

It seems like two different ways to achieve the same result and one is simpler...if that's the case (ie I'm not missing something) I'd advocate for the one-liner(s)

bewebi (commented):
Couple nits, and weighed in on the script vs one-liner piece

hack/compute_version/main.go
sam-heilbron: nice! Yeah I figured that it was valuable/necessary, but it wasn't immediately clear, I think the comment makes it easy to understand for new devs

bewebi (dismissed):
LGTM

test/make/make_suite_test.go
```diff
@@ -1,44 +0,0 @@
-package make_test
```
sam-heilbron, line 1: I think we actually don't want to delete this test. I find it as a valuable way to ensure future updates to our makefile don't break necessary behaviors. I think what we should do here is just update the tests in this suite to track the most important behaviors.

In a way, these tests will be really helpful to prove that the proper cloudbuild env variables still update values correctly. And also that the new values (PUBLSH_CONTEXT) properly set variables as we would expect

sam-heilbron (commented):
Looking great! Some small comments and one ask for expanded tests, but apart from that LGTM

gunnar-solo:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=6c7aeb18-8f68-48e8-a938-63a07f9f7fbf

bewebi (approved):

sam-heilbron (approved):

PR #8300, opened 2023-05-26
refactor: re-adds namedExtauth to settings resource

Todai88:
# Description

Previous releases of Gloo/EE have allowed us to configure our `customAuth` using the `namedExtauth` property in `Helm Values` (for example: Gloo: 1.10.x). This feature seems to have been removed, possibly unintentionally.

Adding back this for consistency.

# Context

As outlined in description.

# Checklist:

BOT NOTES: 
resolves TBA

### DIFF
--- a/changelog/v1.15.0-beta8/re-add-named-extauth-to-settings.yaml
+++ b/changelog/v1.15.0-beta8/re-add-named-extauth-to-settings.yaml
@@ -1,0 +1,9 @@
+changelog:
+  - type: HELM
+    issueLink: TBA
+    resolvesIssue: true
+    description: >-
+      In previous versions of Gloo, it was possible to enable custom authentication (in Enterprise) through 
+      Helm values - even though this was undocumented.
+      For backwards compatibility, this is re-added.
+      
--- a/install/helm/gloo/templates/18-settings.yaml
+++ b/install/helm/gloo/templates/18-settings.yaml
@@ -158,6 +158,10 @@ spec:
 {{- end }}
 {{- end }}
 {{- end -}}
+{{- if .Values.global.extensions.namedExtauth }}
+  namedExtauth:
+{{- toYaml .Values.global.extensions.namedExtauth | nindent 4 }}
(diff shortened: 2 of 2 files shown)


sam-heilbron:
@Todai88 thanks for the contribution! We have created an internal issue to track this work: supporting this in our main branch and backporting the work to earlier LTS branches. This will be picked up by an engineer later this week and hopefully delivered sometime next week. Does that timing work for you?

Could you also link any docs or helm values that previously worked in 1.10? I'm not aware that this was previously exposed, but if we did that somewhere it would help me to understand where

Todai88:
> @Todai88 thanks for the contribution! We have created an internal issue to track this work: supporting this in our main branch and backporting the work to earlier LTS branches. This will be picked up by an engineer later this week and hopefully delivered sometime next week. Does that timing work for you?
> 
> Could you also link any docs or helm values that previously worked in 1.10? I'm not aware that this was previously exposed, but if we did that somewhere it would help me to understand where

As far as I can tell, we haven't patched our version of 1.10.17 (now 1.10.38), and the value has always been there (on the `Settings` Resource) - just as in the PR. 

Without trawling through the history of the template I can't say for sure exactly when this was removed.
But given I can't see any patches from ourselves (on our local chart), I'm assuming this somehow just got removed. 

Keeping in mind that the value was set from `Values.global.extensions` it's not entirely unlikely this was an oversight.
Regardless of the reason, it seems reasonable to expect there to be a safer way of persistently configuring this through deployments (`helm upgrade --install`) rather than through manual tinkering as suggested by [the docs](https://docs.solo.io/gloo-edge/latest/guides/security/auth/multi_authz/).

sam-heilbron:
This work was delivered as part of https://github.com/solo-io/solo-projects/issues/5030 (an Enterprise-only issue).

This is (or will be) available in the following Gloo Enterprise versions:
- 1.15.0-beta2
- 1.14.5
- 1.13.19

Given that this is supported, can you close out this PR?

PR #8303, opened 2023-05-26
Audit use of gomega.Consistently to check receives on unbuffered channels

jbohanon:
# Description

Audit use of gomega.Consistently with channel receives. This is because the default duration and interval are 100ms and 10ms respectively, which are very short considering we're using unbuffered channels. The consequence of this could be developers trusting that behavior is safe because the test did not fail when in actuality, an error could have come in before the Consistently call and was missed due to the unbuffered channel, or the error happened after the default 100ms duration. The goal is to encourage developers to consider and use explicit durations and intervals in Consistently calls.

# Checklist:

~~- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff~~
~~- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review~~
~~- [ ] I have made corresponding changes to the documentation~~
~~- [ ] I have added tests that prove my fix is effective or that my feature works~~


nfuden:
Are we able to write a programatic checker for this?

sam-heilbron (commented):
I really like this! And I think that these messages are a useful reminder about the implications of the default poll intervals.

Do you know if we can statically override those intervals? If so, do you think it would be valuable to do so (and just set the defaults explictly that exist) in https://github.com/solo-io/gloo/tree/main/test/gomega, and perhaps even define the warning message as a string constant in there.

Im happy with this as is, just proposing this as a way to define the information in a single place in case people aren't looking at other existing tests

jbohanon:
> Are we able to write a programatic checker for this?

Not trivially (e.g. grep) because we'd need knowledge of the type of the channel. That may be available via reflection, but I'm not sure, and I think we'd have to reassemble the construction of the spec to access it to get that concrete type. Not totally familiar with this concept though. The gettercheck tool might be helpful example

docs/cmd/test/securityscanutils/generate_securityscan_doc_suite_test.go
```diff
@@ -4,12 +4,16 @@ import (
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	. "github.com/solo-io/gloo/test/gomega"
```
sam-heilbron, line 7: I had kind of thought, we define a function ,wit h an init that did:
```
g.SetDefaultConsistentlyDuration(consistentlyDuration)
	g.SetDefaultConsistentlyPollingInterval(consistentlyPollingInterval)
	g.SetDefaultEventuallyTimeout(eventuallyTimeout)
	g.SetDefaultEventuallyPollingInterval(eventuallyPollingInterval)
```

is that possible? 

My only fear with this paradigm, is that new devs wil import the standard gomega, and we'll have a split.

jbohanon: Hmm. Yeah and add that function call to all the test suites right above the `RegisterFailHandler(Fail)`? 

There's not a way to have a centralized ginkgo entrypoint I assume since you can run any combination of suites?

sam-heilbron (approved):

gunnar-solo (approved):

PR #8307, opened 2023-05-26
Support Skip CI Directives

sam-heilbron:
# Description

Support new ways to skip jobs that run in our CI pipeline

# Context

## Kube Tests
These test are required (branch protection rules), but for some PRs (docs changes for example), we want a mechanism to not run the tests.

Since branch protection rules require that a successful event it received, we need to run some placeholder code, instead of the true tests. 

I decided to keep this logic isolated to the github action logic, and only invoke the true tests if the tests needed it, otherwise we just print out a simple statement outlining that the tests auto-succeeded.

## Docs Builds
Docs builds are a required job in CI, so it's not as critical that we allow them to be skipped. However, their main purpose is to validate docs changes before they are merged. If the PR does not contain docs changes, it is nice to be able to skip them.

We use the same pattern of identifying the behavior in the changelog entry.

## Implementation Tradeoff
I decided to implement this in a bash script, though ideally more of our pipeline code is in go since it can be more easily tested. Since this is a small script I felt comfortable with it, and introduced the placeholder go code framework to demonstrate that we'd like to migrate this some day.

## Testing
We need to ensure that without these directives (or with them set to false), that the jobs run as expected.

I use a [commit](https://github.com/solo-io/gloo/pull/8307/commits/6ec1e27c72ed4c058c314c6e94b7a899c0056764) to show this behavior.

You can see the [Docs/Prepare Env job](https://github.com/solo-io/gloo/actions/runs/5125016378/jobs/9217626140?pr=8307) decided the build strategy:
```
Build strategy: Should build? true
```

You can see the [CI/Prep Env job](https://github.com/solo-io/gloo/actions/runs/5125016379/jobs/9217626097?pr=8307) decided the run strategy:
```
Run should_run=true
Run strategy: Should run? true. Should auto-succeed? false
```

# Checklist:


.github/workflows/docs-gen.yaml
```diff
@@ -23,46 +23,37 @@ jobs:
...
-        name: Process comments on Pull Request to skip Docs
-        if: ${{ github.event.issue.pull_request }}
-        run: echo "skip_docs_comment=${{ contains(github.event.comment.body, '/skip-docs') }}" >> $GITHUB_ENV
-      - id: build-docs
-        name: Determine whether or not to build docs
+        run: echo "value=true" >> $GITHUB_OUTPUT
+      - id: checkout-code
+        uses: actions/checkout@v3
+        with:
+          # We require gathering the branch and tag history since we compare
```
gunnar-solo, line 35: nit: unfinished comment?

.github/workflows/codegen.yaml
```diff
@@ -1,5 +1,7 @@
 name: Codegen
+
```
gunnar-solo, line 2: was the objective to make this workflow optional as well?  Why are there changes?

.github/workflows/docs-gen.yaml
```diff
@@ -23,46 +23,37 @@ jobs:
...
+          fetch-depth: 0
+      - uses: ./.github/workflows/composite-actions/prep-go-runner
```
gunnar-solo, line 37: I advocate against setting up a go runner, here.  It's a lot of compute for (what I can tell?) no gain

.github/workflows/regression-tests.yaml
```diff
@@ -21,36 +21,33 @@ jobs:
...
+    - id: process-skip-directives
+      uses: ./.github/workflows/composite-actions/process-skip-directives
+    - id: run-strategy
+      name: Determine Test Run Strategy
       run: |
         should_run=true
+        should_auto_succeed=false
 
         is_draft_pr=${{ steps.is-draft-pr.outputs.value }}
         if [[ ! -z $is_draft_pr && $is_draft_pr = true ]] ; then
```
gunnar-solo, line 39: nit: it'd make me happier to remove all of the setting of `draft_pr` variables entirely, and instead directly use `${{ github.event.pull_request.draft }}`

.github/workflows/regression-tests.yaml
```diff
@@ -21,36 +21,33 @@ jobs:
...
         fi
 
-        pr_comment_skip=${{ steps.skip-ci-comment.outputs.value }}
-        if [[ ! -z $pr_comment_skip && $pr_comment_skip = true ]] ; then
-          should_run=false
-        fi
-
-        echo "Should run regression tests? $should_run"
+        echo "Run strategy: Should run? $should_run. Should auto-succeed? $should_auto_succeed"
         echo "run_value=$should_run" >> $GITHUB_OUTPUT
```
gunnar-solo, line 49: nit: move this to the if block where `should_run` is set; remove `should_run` as a variable

.github/workflows/regression-tests.yaml
```diff
@@ -21,36 +21,33 @@ jobs:
...
+        echo "auto_succeed=$should_auto_succeed" >> $GITHUB_OUTPUT
```
gunnar-solo, line 50: nit: move this to the if block where `should_auto_succeed ` is set; remove `should_auto_succeed ` as a variable

.github/workflows/composite-actions/regression-tests/action.yaml
```diff
@@ -55,14 +56,6 @@ runs:
...
     shell: bash
     run: make install-test-tools run-kube-e2e-tests
-  - uses: testspace-com/setup-testspace@v1
-    with:
-      domain: solo-io.testspace.com
-    if: ${{ always() && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-  - name: Push result to Testspace server
-    shell: bash
-    run: testspace push --verbose "**/junit.xml"
-    if: ${{ always() && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
bewebi, line 65: Why are we removing these steps?

.github/workflows/docs-gen.yaml
bewebi, line 35: As written, it doesn't look like we're comparing? Though I think it would be good to

.github/workflows/docs-gen.yaml
```diff
@@ -23,46 +23,37 @@ jobs:
...
         if: ${{ github.event.pull_request.draft }}
-        run: echo "is_draft_pr=true" >> $GITHUB_ENV
-      - id: signal-docs-comment
-        name: Process comments on Pull Request to signal Docs
-        if: ${{ github.event.issue.pull_request }}
-        run: echo "signal_docs_comment=${{ contains(github.event.comment.body, '/sig-docs') }}" >> $GITHUB_ENV
-      - id: skip-docs-comment
-        name: Process comments on Pull Request to skip Docs
-        if: ${{ github.event.issue.pull_request }}
-        run: echo "skip_docs_comment=${{ contains(github.event.comment.body, '/skip-docs') }}" >> $GITHUB_ENV
```
bewebi, line 40: Why remove the option to skip via comment? This may be useful when trying to skip for a particular commit when we know further changes are coming

ci/github-actions/main.go
```diff
@@ -0,0 +1,21 @@
+package main
```
gunnar-solo, line 1: nit: I'd rather not have an unused script sitting around, potentially causing confusion to later maintenance

.github/workflows/regression-tests.yaml
```diff
@@ -21,36 +21,33 @@ jobs:
...
+          should_auto_succeed=true
         fi
 
-        pr_comment_skip=${{ steps.skip-ci-comment.outputs.value }}
-        if [[ ! -z $pr_comment_skip && $pr_comment_skip = true ]] ; then
-          should_run=false
-        fi
-
-        echo "Should run regression tests? $should_run"
+        echo "Run strategy: Should run? $should_run. Should auto-succeed? $should_auto_succeed"
```
bewebi, line 48: I'm a bit confused about what this logic is doing and, separately, under what circumstances we'd want to skip tests but not auto-succeed, or skip tests and also auto-succeed, etc.

.github/workflows/regression-tests.yaml
```diff
@@ -21,36 +21,33 @@ jobs:
...
-      if: ${{ github.event.issue.pull_request }}
-      run: echo "value=${{ contains(github.event.comment.body, '/skip-ci') }}" >> $GITHUB_OUTPUT
-    - id: regression-tests
-      name: Determine how to run regression tests
+    - id: checkout-code
+      uses: actions/checkout@v3
+      with:
+        # We require gathering the branch and tag history since we compare
+        fetch-depth: 0
+    - uses: ./.github/workflows/composite-actions/prep-go-runner
```
gunnar-solo, line 29: I advocate against using a go runner here, as it adds unnecessary time to our overall build

.github/workflows/docs-gen.yaml
```diff
@@ -23,46 +23,37 @@ jobs:
...
+        run: echo "value=true" >> $GITHUB_OUTPUT
+      - id: checkout-code
+        uses: actions/checkout@v3
+        with:
+          # We require gathering the branch and tag history since we compare
+          fetch-depth: 0
+      - uses: ./.github/workflows/composite-actions/prep-go-runner
+      - id: process-skip-directives
+        uses: ./.github/workflows/composite-actions/process-skip-directives
+      - id: build-strategy
```
gunnar-solo, line 40: nit: see comments in `regression-tests.yaml`

bewebi (commented):
Overall looks good

I like the ability to add the directive via changelog and/but:
- I don't see a reason to remove the ability to include directives via comment
- It seemed from the PR description, and comments in the actions, that we would try to detect changes in relevant code in order to determine whether we'd need to run tests (ie only run docs-gen if a docs file has been modified)

Note that my comments on the docs-gen action apply to regression-tests as well

ci/README.md
```diff
@@ -18,6 +18,12 @@
 **Keep PR Updated**: When applied, bulldozer will keep the PR up to date with the base branch, by merging any updates into it (Applied by default)
 
 **Work In Progress**: When applied, will prevent bulldozer from merging a PR, even if it has passed all checks
+### Special Directives to Skip CI
+**Skip Build-Bot**: Following the [special comment directives](https://github.com/solo-io/build-bot#issue-comment-directives), comment `skip-ci` on the PR.
```
gunnar-solo, line 22: I'm missing the part of the PR that manages this logic--can you point me in the right direction?

ci/github-actions/process-skip-directives/script.sh
```diff
@@ -0,0 +1,39 @@
...
+
+set -ex
+
+skipKubeTestsDirective="skipCI-kube-tests:true"
+shouldSkipKubeTests=false
+
+skipDocsBuildDirective="skipCI-docs-build:true"
+shouldSkipDocsBuild=false
+
+if [[ $(git diff origin/main HEAD --name-only | grep "changelog/" | wc -l) = "1" ]]; then
```
gunnar-solo, line 24: heads up:  `git diff origin/main HEAD` will not hold up once this backports

ci/github-actions/process-skip-directives/skip_ci_tests.go
```diff
@@ -0,0 +1,7 @@
+package process_skip_directives
```
gunnar-solo, line 1: I'm arguing we should *not* migrate to go code, and, instead, remove this stub

gunnar-solo (changes requested):
looking good!  A few requests/questions

.github/workflows/docs-gen.yaml
gunnar-solo: the comparison happens in `ci/github-actions/process-skip-directives/script.sh`, via `git diff origin/main HEAD`

.github/workflows/docs-gen.yaml
sam-heilbron: ah yeah good call, this was from when I tried to make the utility use go code to solve the problem. i reverted to bash and forgot to clean this up, good catch

.github/workflows/codegen.yaml
sam-heilbron: I had initially tested my changes in codegen and these spacing changes are leftover. I'm happy to revert if you'd like. I kind of like the spacing between each of the sections, but happy to defer to you

.github/workflows/regression-tests.yaml
gunnar-solo: two cases considered, here:
* **draft PRs**: a PR which we want to open, but don't want to run our tests.  Said tests, having not run, prevent a merge to `main`
* **"Docs PRs"**: a PR which we want to open, don't want to run our tests, _but_ want to be able to immediately merge to `main`.  This PR is centered around this use-case by virtue of running some empty compute to make the world happy.  Ex:
```
    - id: auto-succeed-tests
      if: needs.prepare_env.outputs.should-auto-succeed-regression-tests == 'true'
      run: |
        echo "Regression tests auto-succeeded"
```

.github/workflows/docs-gen.yaml
bewebi: True, but we only do so to find the added changelog, not to actually find changes relevant to docs-gen
As is I guess we do still need to compare (unless there's a better way to identify the new changelog) but I meant more to highlight that we're not comparing to find docs changes to determine whether to skip the action

.github/workflows/regression-tests.yaml
bewebi: Draft PRs already won't merge, but I suppose if a Draft PR was promoted to "ready" it could then merge without changes if all tests were reported as passing?

ci/README.md
sam-heilbron: This PR doesn't affect this code, but it's been a common question, and since I was editing the REAME around how to skip aspects of CI, I figured I would include it

ci/github-actions/process-skip-directives/script.sh
sam-heilbron: good call, I was initally leaning towards adding this for now and handling the other branches in a follow up, but I think I see a way to make this more future proof

.github/workflows/docs-gen.yaml
sam-heilbron: https://github.com/solo-io/gloo/tree/main/.github/workflows#future-work these actualy never worked

.github/workflows/docs-gen.yaml
sam-heilbron: I'll update this comment to clarify how it works, we use a `git diff` and need to have both branches/tags available to succeed

.github/workflows/composite-actions/regression-tests/action.yaml
sam-heilbron: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1684333291668589

This was leftover from a change previously made

.github/workflows/regression-tests.yaml
sam-heilbron: yeah the challenge is that if we "auto-succeed" tests on a draft PR, we run a no-op for each of the regression tests and report it as "passed". Then the user marks the PR as ready for review, and CI doesn't re-run then. So devs might miss that their tests didn't run when going from draft -> reviewable PR

bewebi (approved):
Discussed some possible next steps [in slack](https://solo-io-corp.slack.com/archives/C03MFATU265/p1685474010459879?thread_ts=1685470863.441859&cid=C03MFATU265) but this PR looks good and we can open follow-up issues/PRs as needed

PR #8323, opened 2023-05-30
API for disabling client secrets in OAuth

elcasteel:
# Description
The Okta PKCE flow does not use a fixed client secret. In order to support this we need to make providing the client secret optional. However, since the secret is required (and would be dangerous to omit) for other use cases and we don't have any other API to indicate which flow is being used, I'm adding a field to indicate that the client secret is disabled.

# Context

PKCE with okta docs:  https://developer.okta.com/blog/2019/08/22/okta-authjs-pkce#replace-implicit-flow-with-pkce
This was requested by customers. Additional changes will happen in ext-auth-service and solo-projects to use this API update. 
# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -706,6 +713,10 @@ message PlainOAuth2 {
     // The URL of the provider token revocation endpoint.
     // For more information, refer to https://www.rfc-editor.org/rfc/rfc7009.
     string revocation_endpoint = 13;
+    // If true, do not check for or use the client secret.
+    // Generally the client secret is required and AuthConfigs will be rejected if it isn't set.
+    // However certain implementations of the PKCE flow do not use a client secret (including Okta) so this setting allows configuring Oauth2 without a client secret.
+    bool disable_client_secret = 14;
```
sam-heilbron, line 719: Any reason we don't use a nullable boolean here?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1398,6 +1410,11 @@ message ExtAuthConfig {
 
         // Configuration related to the user session.
         UserSessionConfig user_session = 20;
+
+        // If true, do not check for or use the client secret.
+        // Generally the client secret is required and AuthConfigs will be rejected if it isn't set.
+        // However certain implementations of the PKCE flow do not use a client secret (including Okta) so this setting allows configuring Oidc without a client secret.
+        bool disable_client_secret = 21;
```
sam-heilbron, line 1417: Right now we take the user defined boolean value, and pass it right through to the auth service. An alternative would be that we don't have this new field in the internal api, and just don't pass the secret (use empty string) if the secret is disabled. Is there a reason we can't or shouldn't do that?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
elcasteel: Codegen failed when I tried to make fields optional with the message `Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.`

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -793,6 +804,10 @@ message IntrospectionValidation {
     // server is configured); if the field is set and the attribute cannot be found, the request will be denied.
     // This field is optional and by default the server will not try to derive the user ID.
     string user_id_attribute_name = 4;
+
+    // Allows setting a client id but not a client secret.
+    google.protobuf.BoolValue disable_client_secret = 5;
```
sam-heilbron, line 809: In the client_secret_ref API for IntrospectionValidation, I see:
```
// Optional: Use if the token introspection url requires client authentication.
```

Does this mean that this new boolean field isn't necessary because we already supported an empty secret ref? Or does it mean that the comment on that field was misleading, and with this new field it will work as originally documented?

elcasteel: Previously (in solo-projects) we didn't support a clientId without a secret.

sam-heilbron (approved):

sheidkamp (approved):

PR #8326, opened 2023-06-01
8225 HeaderManipulation Validation

sheidkamp:
# Description

We were passing through invalid values of `responseHeadersToAdd` to Envoy,  resulting in this type of error:
```
[2023-05-19 14:01:29.817][7][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:126] gRPC config for type.googleapis.com/envoy.config.route.v3.RouteConfiguration rejected: :-prefixed or host headers may not be modified
```

Steps to reproduce the bug
Create a VS with the following

    options:
      headerManipulation:
        responseHeadersToAdd:
          - header:
              key: ":scheme"
              value: "https"
            append: false

Additionally, we should not be allowing "Host" headers to be set (replace `:scheme` with `Host`)

# Context
Users want Gloo validation to catch this error instead passing invalid config to the gateway.

# Checklist:


projects/gloo/pkg/plugins/headers/plugin_test.go
```diff
@@ -111,6 +111,36 @@ var _ = Describe("Plugin", func() {
 		Expect(out.ResponseHeadersToAdd).To(Equal(expectedHeadersWithSecrets.ResponseHeadersToAdd))
 		Expect(out.ResponseHeadersToRemove).To(Equal(expectedHeadersWithSecrets.ResponseHeadersToRemove))
 	})
+
+	DescribeTable("Invalid headers", func(key string, value string, expectedErr error) {
```
elcasteel, line 115: 🎊

elcasteel (dismissed):
LGTM. I think it makes sense to not attempt to do any type of "best effort" on the header manipulation if any of them are bad. 
Are the limitations documented?

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -150,13 +157,23 @@ func convertResponseHeaderValueOption(
 ) ([]*envoy_config_core_v3.HeaderValueOption, error) {
 	var out []*envoy_config_core_v3.HeaderValueOption
 	for _, h := range in {
-		if h.GetHeader() == nil {
+		header := h.GetHeader()
+		if header == nil {
 			return nil, MissingHeaderValueError
 		}
+
+		if header.GetKey()[0] == byte(':') {
```
sam-heilbron, line 165: Any reason for not using the standard [HasPrefix](https://pkg.go.dev/strings#HasPrefix)?

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -150,13 +157,23 @@ func convertResponseHeaderValueOption(
...
+			return nil, CantSetPseudoHeaderError(header.GetKey())
+		}
+
+		if strings.EqualFold(header.GetKey(), "Host") {
+			return nil, CantSetHostHeaderError
+		}
+
```
sam-heilbron, line 172: The logic in this space looks very similar to what we have in our api_conversion utility (https://github.com/solo-io/gloo/blob/main/pkg/utils/api_conversion/type.go#L54). Do these guards need to be added to those headers as well?

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -22,7 +24,12 @@ const (
 )
 
 var (
-	MissingHeaderValueError = errors.Errorf("header section of header value option cannot be nil")
+	MissingHeaderValueError = eris.Errorf("header section of header value option cannot be nil")
```
sam-heilbron, line 27: We aren't passing in a value here, should we instead just be using eris.New?

sam-heilbron (commented):
Looking good!

projects/gloo/pkg/plugins/headers/plugin.go
sheidkamp: Nope, updating.

projects/gloo/pkg/plugins/headers/plugin.go
sheidkamp: yup

sam-heilbron (dismissed):

sheidkamp:
> LGTM. I think it makes sense to not attempt to do any type of "best effort" on the header manipulation if any of them are bad. Are the limitations documented?

We point to the envoy documentation in the `headers.proto` file, but it was a dead link. I have updated that link, so users can get to the documentation

sam-heilbron (approved):

elcasteel (approved):

PR #8332, opened 2023-06-01
build(publish-artifacts): restore/call render-manifests target

gunnar-solo:


sam-heilbron:
I think we can solve this differently. https://github.com/solo-io/gloo/pull/8336 outlines what I think we should do

PR #8333, opened 2023-06-02
feat: expose stats tags in gateway-proxy stats config

jlourenc:
# Description

This new feature can be used to configure gateway proxies stats configuration with stats tags.

# Checklist:


sam-heilbron (commented):
Thanks for the contribution @jlourenc!

I like the idea here, there are just a few changes that we'll need to make:
1. https://github.com/solo-io/gloo/blob/main/install/helm/gloo/generate/values.go is the source of truth for our helm docs. If you update the configuration you want to modify, and then run code-generation, the helm reference docs will be dynamically updated
2. For any helm changes, we use https://github.com/solo-io/gloo/blob/main/install/test/helm_test.go to write a unit test to assert that the generated manifests match what we would expect. https://github.com/solo-io/gloo/blob/main/install/test/helm_test.go#L3607 is an example of a test that will be similar to yours (ie we apply some values, render the manifests, and assert the existence of the injected configuration)
3. Have you explored https://docs.solo.io/gloo-edge/latest/reference/helm_chart_values/open_source_helm_chart_values/#helm-chart-kuberesourceoverrides? We expose a way to inject custom values to a lot of our templates, and perhaps you can use this to define the stats configuration generically (without being blocked by this helm change)

jlourenc:
Hey @sam-heilbron , thanks for taking a look.

1. I believe I have done the appropriate changes.
2. I have added a unit-test.
3. Unfortunately, I believe this mechanism is not available in this scenario as `envoy.yaml` is a string.

changelog/v1.15.0-beta14/gwp-stats-config.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+- type: NEW_FEATURE
```
sam-heilbron, line 2: This will cause the changelog-bot to fail with 
```
changelog-bot — Changelog entries must have an issue link
```

To solve this, you can create a Github issue that tracks this work, and then reference it with an issueLink, as identified here: https://github.com/solo-io/go-utils/tree/main/changelogutils#changelog-files

Alternatively, you can make this `type: NON_USER_FACING`

sam-heilbron (commented):
Looking great! Sorry for the delay in getting back to you

install/helm/gloo/generate/values.go
```diff
@@ -446,6 +446,7 @@ type GatewayProxy struct {
 	IstioMetaClusterId             *string                      `json:"istioMetaClusterId,omitempty" desc:"ISTIO_META_CLUSTER_ID Environment Variable. Defaults to \"Kubernetes\""`
 	IstioDiscoveryAddress          *string                      `json:"istioDiscoveryAddress,omitempty" desc:"discoveryAddress field of the PROXY_CONFIG environment variable. Defaults to \"istiod.istio-system.svc:15012\""`
 	EnvoyLogLevel                  *string                      `json:"envoyLogLevel,omitempty" desc:"Level at which the pod should log. Options include \"trace\", \"info\", \"debug\", \"warn\", \"error\", \"critical\" and \"off\". Default level is info"`
+	EnvoyStatsConfig               map[string]interface{}       `json:"envoyStatsConfig,omitempty" desc:"Envoy statistics configuration such as tagging. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#config-metrics-v3-statsconfig"`
```
Rachael-Graham, line 449: docs suggestion, if you dont mind!
```suggestion
	EnvoyStatsConfig               map[string]interface{}       `json:"envoyStatsConfig,omitempty" desc:"Envoy statistics configuration, such as tagging. For more info, see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#config-metrics-v3-statsconfig"`
```

jlourenc: Thank you for updating the PR 🙏🏼

install/test/helm_test.go
```diff
@@ -802,6 +802,26 @@ var _ = Describe("Helm Test", func() {
 					})
 				})
 
+				It("should set stats_config in gateway-proxy-envoy-config from gatewayProxies.gatewayProxy.envoyStatsConfig", func() {
+					prepareMakefileFromValuesFile("values/val_gwp_envoy_stats_config.yaml")
```
sheidkamp, line 806: We haven't really been using values files for new tests, can we rewrite this to look like the test above using `prepareMakefile(namespace, helmValues{ ... `

install/test/helm_test.go
```diff
@@ -802,6 +802,26 @@ var _ = Describe("Helm Test", func() {
...
+
+					testManifest.SelectResources(func(resource *unstructured.Unstructured) bool {
+						return resource.GetKind() == "ConfigMap"
+					}).ExpectAll(func(configMap *unstructured.Unstructured) {
+						configMapObject, err := kuberesource.ConvertUnstructured(configMap)
+						Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("ConfigMap %+v should be able to convert from unstructured", configMap))
+						structuredConfigMap, ok := configMapObject.(*v1.ConfigMap)
+						Expect(ok).To(BeTrue(), fmt.Sprintf("ConfigMap %+v should be able to cast to a structured config map", configMap))
+
+						if structuredConfigMap.GetName() == "gateway-proxy-envoy-config" {
```
sheidkamp, line 816: We should create a flag like `foundResource` that is initialized as false and gets set to true when the resource is found in this block. We should then check at the end of the test if the flag is true, as this test will pass if `gateway-proxy-envoy-config` isn't present.

install/test/values/val_gwp_envoy_stats_config.yaml
```diff
@@ -0,0 +1,6 @@
+gatewayProxies:
+  gatewayProxy:
+    envoyStatsConfig:
+      stats_tags:
+      - fixed_value: zero
+        tag_name: one
```
sheidkamp, line 6: We can remove this once the test is rewritten

sheidkamp (commented):
Looking good! I have a few suggestions to tighten up the test.

install/test/helm_test.go
jlourenc: Addressed. Replaced by using `prepareMakefile`.

install/test/helm_test.go
jlourenc: This would not work as the function was executed for all config maps. It would then fail for all but the one we're concerned about.

I've updated the test to select resources based on type AND name, then ensured only the required config map is being asserted by the test, and finally performed the assertions appropriately.

install/test/values/val_gwp_envoy_stats_config.yaml
jlourenc: Removed

jlourenc:
I have taken your suggestions onboard to update the test. Please take another look 🙏🏼

sheidkamp (approved):
LGTM, but please don't squash between reviews in the future! It takes away the ability to see just the changes since the review. Thanks for the feature!

elcasteel (dismissed):

install/helm/gloo/generate/values.go
inFocus7: Heyyo! Wondering if you were planning on committing the suggestions or not. If not, that's alright and I'll approve the PR, but if so I'll approve afterward (note: if you do commit this, i think you'll need to run code generation once more) 👀

inFocus7: Oop, looks like it got 2 approvals already so you can ignore me lol. Great work/changes!

sheidkamp: We needed a rebase, and I did want to get that change in, and I thought I had committed it (resulting in the comment "Thank you for updating the PR"), so throwing this in.

sheidkamp (approved):

elcasteel (approved):

sheidkamp (approved):

elcasteel (approved):

PR #8342, opened 2023-06-02
1.14 api/external: Cipher detection, new field for terminating ciphers

nfuden:
Pretty much this entire pr comes down to enabling and then plumbing this line https://github.com/solo-io/gloo/pull/8342/files#diff-3fc9c8b1139f1242e49ff7e5ccca3e48791180b5f90134688c610f3e21813f7dR188

Namely from where we know about sslconfigs on httpgateways we store them in our special wrapper for filterchains (extended filterchains) and then propagate that out to any plugins that mutate filterchains. 

In particular the one plugin at time of writing is the Passthrough plugin. It uses the terminating cipher list to have more granular control over what is accepted as "supported" versus not when making decisions in the presence of a set of passthrough ciphers.

https://github.com/solo-io/solo-projects/pull/5061/ for sp work


projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -79,19 +79,19 @@ func (t *tcpFilterChainTranslator) ComputeFilterChains(params plugins.Params) []
...
-		}
-	}
-
 	extFilterChains := make([]*plugins.ExtendedFilterChain, 0, len(filterChains))
 	for _, fc := range filterChains {
 		fc := fc
 		extFilterChains = append(extFilterChains, &plugins.ExtendedFilterChain{FilterChain: fc, PassthroughCipherSuites: t.passthroughCipherSuites})
 	}
 
+	// 2. Apply SourcePrefixRange to FilterChainMatch, if defined
```
jenshu, line 88: why does this part need to be moved later?

jenshu: oh because it uses ExtendedFilterChain now

nfuden: Yeah, its actually kinda nice as imo this makes the logic a bit cleaner once more as the extended filter chain wrapping loop is a little silly.

jenshu (approved):

PR #8352, opened 2023-06-05
mod: pin docker dist until we can bump. see https://github.com/kubern…

nfuden:
like https://github.com/kubernetes/kubernetes/commit/3680a5230c386602ebc82700fe3031960d0479fd

for more details see https://github.com/distribution/distribution/security/advisories/GHSA-hqxw-f8mx-cpmw

basically this docker cve doesnt really hit us but until a newer k8s comes out with the newer version we might as well pin to reduce cve noise in scanners


nfuden:
closed in favor of a top level ignore as this doesnt really affect us

PR #8359, opened 2023-06-06
Control plane support for new transformation logging functionality

ben-taussig-solo:
# Description
 - Bumps envoy-gloo to pull in new transformation filter functionality introduced in https://github.com/solo-io/envoy-gloo/pull/240
 - Adds two new settings to the user-facing transformation plugin API, each called `log_request_response_info`
   - When enabled, these settings will cause the transformation filter to log request/response (depending on which the transformation is applied to) body and headers before and after a given transformation is processed, for debug purposes
   - One of these settings is at the level of an individual transformation. The other is at route/vhost level, depending on where the transformation is configured
     - the route/vhost level setting takes precedence over the individual transformaiton level setting.
 - A solo-projects PR to support this functionality for enterprise transformers can be found here: https://github.com/solo-io/solo-projects/pull/5078


projects/gloo/pkg/plugins/transformation/plugin_test.go
```diff
@@ -109,6 +109,86 @@ var _ = Describe("Plugin", func() {
...
+
+				output, err := p.(TransformationPlugin).ConvertTransformation(
+					ctx,
+					&transformation.Transformations{},
+					inputTransformationStages,
+				)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(output).To(Equal(expectedOutput))
+			})
```
jackstine, line 160: can we have a test to ensure that when `LogRequestResponseInfo` is set to `false`, that we validate that it does not set `LogRequestResponseInfo` on the `RequestTransformation`.

ben-taussig-solo: Added this here: https://github.com/solo-io/gloo/pull/8359/commits/87cf9f84d9a89ab592b7f644ea550f9e9e6c9674

projects/gloo/api/v1/options/transformation/transformation.proto
```diff
@@ -66,6 +66,10 @@ message TransformationStages {
     // only one transformation will run per stage.
     // Defaults to false.
     bool inherit_transformation = 3;
+    
+    // When enabled, log request/response body and headers before and after all transformations defined here are applied.\
+    // This overrides the log_request_response_info field in the Transformation message.
+    google.protobuf.BoolValue log_request_response_info = 4;
```
jackstine, line 72: can we show a test to confirm the behavior of `This overrides the log_request_response_info field in the Transformation message.`

jackstine (approved):
LGTM

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -26,6 +27,15 @@ import (
 	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/pluginutils"
 )
 
+type TransformationPlugin interface {
```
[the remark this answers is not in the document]
ben-taussig-solo: This was only necessary for testing, so I just moved the definition of this interface to the plugin test here: https://github.com/solo-io/gloo/pull/8359/commits/26c0099b7621b9ab2d91c77e63f176c3c23a4ea7

nfuden (dismissed):

projects/gloo/api/external/envoy/extensions/transformation/transformation.proto
```diff
@@ -25,6 +26,12 @@ message FilterTransformations {
   // Only RouteTransformations.RouteTransformation with matching stage will be
   // used with this filter.
   uint32 stage = 2 [ (validate.rules).uint32 = {lte : 10} ];
+
+  // Logs request/response sensitive information
+  // By default, this is false so no request or response sensitive information is logged.
+  // If set to true, the filter will log the request/response body and headers before and
+  // after any transformation is applied.
+  bool log_request_response_info = 3;
```
sam-heilbron, line 34: Why is this a type bool while the other one is a nullable BoolValue?

projects/gloo/api/v1/settings.proto
```diff
@@ -692,6 +692,10 @@ message GlooOptions {
 
     // Where the `gloo` proxy debug server should bind. Defaults to `gloo:9966`
     string proxy_debug_bind_addr = 15;
+
+    // When enabled, log the request/response body and headers before and after
```
sam-heilbron, line 696: nit: Perhaps call out that it defaults to false. Do we think that it makes sense to clarify why someone would want this enabled?

test/e2e/staged_transformation_test.go
```diff
@@ -25,7 +26,7 @@ import (
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/transformation"
 )
 
-var _ = Describe("Staged Transformation", func() {
+var _ = FDescribe("Staged Transformation", func() {
```
sam-heilbron, line 29: Focused test

test/e2e/staged_transformation_test.go
```diff
@@ -197,7 +198,7 @@ var _ = Describe("Staged Transformation", func() {
 			}, "15s", ".5s").Should(Succeed())
 		})
 
-		It("Should be able to base64 decode the body", func() {
+		FIt("Should be able to base64 decode the body", func() {
```
sam-heilbron, line 201: Focused test

test/e2e/staged_transformation_test.go
```diff
@@ -307,6 +308,154 @@ var _ = Describe("Staged Transformation", func() {
...
+			requestBuilder := testContext.GetHttpRequestBuilder().WithPostBody("test")
+			Eventually(func(g Gomega) {
+				g.Expect(testutils.DefaultHttpClient.Do(requestBuilder.Build())).Should(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
+					StatusCode: http.StatusOK,
+					Body:       "regular-transformed-early-transformed",
+				}))
+			}, "15s", ".5s").Should(Succeed())
+
+			// get the logs from the gateway-proxy container
+			logs, err := testContext.EnvoyInstance().Logs()
```
sam-heilbron, line 377: Can you try running this a few times (10) in a row, and confirm there are no flakes due to delays in retrieving the logs?

test/e2e/staged_transformation_test.go
```diff
@@ -307,6 +308,154 @@ var _ = Describe("Staged Transformation", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(logs).To(ContainSubstring(`body before transformation: test`))
```
sam-heilbron, line 379: nit: We can re-write this as:
```
Expect(logs).To(And(
ContainSubstring("before"),
ContainSubstring("after"),
...
```

test/e2e/staged_transformation_test.go
```diff
@@ -307,6 +308,154 @@ var _ = Describe("Staged Transformation", func() {
...
+			Expect(logs).To(ContainSubstring(`body after transformation: regular-transformed`))
+			Expect(logs).To(ContainSubstring(`body before transformation: regular-transformed`))
+			Expect(logs).To(ContainSubstring(`body after transformation: regular-transformed-early-transformed`))
+		})
+
+		It("can enable enhanced logging for an individual transformation", func() {
```
sam-heilbron, line 385: I like both of these cases! Can we also add in a negative test case, which shows that we do not log the transformation details when the setting is disabled?

sam-heilbron (commented):
Looking good!  A few thoughts:
1. Can you help me understand why we want to expose this at the global Settings level? It seems like this is a useful debugging tool, but not one that we necessarily want to encourage widespread usage of. It feels like exposing this in Settings improves the UX of this feature, but I don't fully believe we want that
2. Can we add updates to our public transformation docs (https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/transformations/, perhaps there are others as well)? This seems like a useful feature that we're exposing and one that we should make the public aware of

test/e2e/staged_transformation_test.go
ben-taussig-solo: ugh thought I got all of these, thank you

sam-heilbron:
Can you investigate how the tests passed even while they were focused?

ben-taussig-solo:
> Looking good! A few thoughts:
> 
> 1. Can you help me understand why we want to expose this at the global Settings level? It seems like this is a useful debugging tool, but not one that we necessarily want to encourage widespread usage of. It feels like exposing this in Settings improves the UX of this feature, but I don't fully believe we want that
> 2. Can we add updates to our public transformation docs (https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/transformations/, perhaps there are others as well)? This seems like a useful feature that we're exposing and one that we should make the public aware of

1. I think the request is partially on behalf of organizations which have large numbers of transformations controlled by different teams. So requests fail for them and they don't necessarily understand which transformations are failing. The global setting makes it easier to pinpoint where things are going wrong when many transformations are potentially responsible for the issue. I hear you though, all of the new settings introduced here have the potential to log sensitive info and the global one is definitely the greatest risk. @nfuden interested in your input on this.
2. Yes, definitely -- how would you feel about me including those docs updates in a follow-up PR? There's still a decent amount of work that needs to be done with merging EE changes and backporting and I feel like it might be more efficient to me to write these docs in-between CI runs

ben-taussig-solo:
> Can you investigate how the tests passed even while they were focused?

Oh I actually ran into this while writing the e2e tests for this PR. The `run-e2e-tests` Makefile target ignores focused tests now. So I believe that all e2e tests were run, and the job exited with status 1 (not 197, which we would expect if only a focused subset of e2e tests were run). I'm not sure why this change in behavior has occurred, but it seems to be a consequence of recent updates to developer testing workflows. Locally I had to run the `test` target and specify `TEST_PKG=./test/e2e` in order to run e2e tests while respecting focused tests

You can see from the build-bot run that all e2e tests did actually run:
```
Step #1 - "run-tests": Step #8 - "run-e2e-tests": [38;5;10m[1mRan 247 of 247 Specs in 676.289 seconds[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests": [38;5;10m[1mSUCCESS![0m -- [38;5;10m[1m247 Passed[0m | [38;5;9m[1m0 Failed[0m | [38;5;11m[1m0 Pending[0m | [38;5;14m[1m0 Skipped[0m
```

It's a little difficult to read because of the formatting directives, but there is a `247 Passed` in there

sam-heilbron:
> actually ran into this while writing the e2e tests for this PR. The `run-e2e-tests` Makefile target ignores focused tests now. So I believe that all e2e tests were run, and the job exited with status 1 (not 197, which we would expect if only

What's confusing is the `run-e2e-tests` make target (https://github.com/solo-io/gloo/blob/main/Makefile#L184) just calls `test` under the hood.

"Makefile target ignores focused tests now." - Were you able to confirm this by running this target locally, with your focused tests and getting that behavior?

test/e2e/staged_transformation_test.go
ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/8359/commits/4d24794609cb7a8b74f8e635ea661929a5e4d646

test/e2e/staged_transformation_test.go
ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/8359/commits/4d24794609cb7a8b74f8e635ea661929a5e4d646

test/e2e/staged_transformation_test.go
ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/8359/commits/4d24794609cb7a8b74f8e635ea661929a5e4d646

Also did the same for the subsequent tests

projects/gloo/api/external/envoy/extensions/transformation/transformation.proto
ben-taussig-solo: There's a whole discussion about this here: https://github.com/solo-io/envoy-gloo/pull/240#discussion_r1207081565

Long story short, making this a `BoolValue` enables slightly better UX

projects/gloo/api/v1/settings.proto
ben-taussig-solo: Updated this commentary here: https://github.com/solo-io/gloo/pull/8359/commits/67584f7634a1932626209f7052316c237d9c9d55

projects/gloo/api/external/envoy/extensions/transformation/transformation.proto
ashishb-solo, line 34: nit: it may be helpful to indicate in the docstring that this value can be overridden by `TransformationRule.log_request_response_info`

ashishb-solo: disregard

projects/gloo/pkg/plugins/transformation/plugin_test.go
```diff
@@ -109,6 +110,146 @@ var _ = Describe("Plugin", func() {
...
+										HeaderBodyTransform: &envoytransformation.HeaderBodyTransform{},
+									},
+								},
+							},
+						},
+					}},
+				}
+			})
+
+			It("can set log_request_response_info on transformation-stages level", func() {
```
ashishb-solo, line 158: it would be a bit easier to see these tests in a table because it's hard to track the differences between each one. is this too much of an ask? happy to ignore if it is

test/e2e/staged_transformation_test.go
```diff
@@ -307,6 +308,206 @@ var _ = Describe("Staged Transformation", func() {
...
+				g.Expect(testutils.DefaultHttpClient.Do(requestBuilder.Build())).Should(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
+					StatusCode: http.StatusOK,
+					Body:       body,
+				}))
+			}, "15s", ".5s").Should(Succeed())
+
+			// get the logs from the gateway-proxy container
+			logs, err := testContext.EnvoyInstance().Logs()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(logs).NotTo(And(
```
ashishb-solo, line 505: should this be `Or` instead of `And`?

test/e2e/staged_transformation_test.go
```diff
@@ -307,6 +308,206 @@ var _ = Describe("Staged Transformation", func() {
...
+				g.Expect(testutils.DefaultHttpClient.Do(requestBuilder.Build())).Should(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
+					StatusCode: http.StatusOK,
+					Body:       "regular-transformed-early-transformed",
+				}))
+			}, "15s", ".5s").Should(Succeed())
+
+			// get the logs from the gateway-proxy container
+			logs, err := testContext.EnvoyInstance().Logs()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(logs).NotTo(And(
```
ashishb-solo, line 455: same comment as https://github.com/solo-io/gloo/pull/8359/files#r1223448201

test/e2e/staged_transformation_test.go
ben-taussig-solo: Ah yes, the `NotTo`s should contain an `Or`

projects/gloo/pkg/plugins/transformation/plugin_test.go
ben-taussig-solo: Refactored this here: https://github.com/solo-io/gloo/pull/8359/commits/5c2f5505342778a5c5f9d9c91da2fe672f624334

The table is much easier to reason about IMO, thanks for the suggestion!

test/e2e/staged_transformation_test.go
ben-taussig-solo: Same as above -- resolved in the comment that changes the e2e tests to a table format: https://github.com/solo-io/gloo/commit/5c2f5505342778a5c5f9d9c91da2fe672f624334

projects/gloo/pkg/plugins/transformation/plugin_test.go
ashishb-solo: that came out great, awesome job! 🔥

ashishb-solo (approved):

PR #8364, opened 2023-06-06
Initial translation benchmark

bewebi:
# Description

- Remove placeholder performance test files from syncer directory
- Add benchmarking suite to translator tests, using a table to execute translation benchmarks
  - Add `benchmarkConfig` struct used to configure criteria for benchmark testing per-table entry
  - Include some sample test cases and comments laying out intended usage and structure
- Add functions to `test/gomega/matchers` package for getting `gomega.Matcher`s to be used for benchmarking
- Add `resources.go` to `test/helpers` package with some exported basic resources and a `ScaledSnapshot()` function to allow for simple generation of snapshots with specific numbers of each resource

# Context

The intent of this PR is to provide a jumping-off point for developers to add benchmark tests, both for translation and elsewhere
The included tests can be taken as examples and are not reflective of realistic use cases or SLAs

## Example test output

Passing, all iterations completed
```
• [0.846 seconds]
Translation - Benchmarking Tests Benchmark table (upstream scale) 1 endpoint(s), 10 upstream(s) [performance]
/Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:119

  Report Entries >>
  Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s) - /Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:98 @ 06/06/23 13:30:58.606
    Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s)
    Name                                                      | N    | Min   | Median | Mean  | StdDev | Max  
    ==========================================================================================================
    (upstream scale) 1 endpoint(s), 10 upstream(s) [duration] | 1000 | 700µs | 700µs  | 800µs | 200µs  | 2.1ms
  << Report Entries
```

Passing, iterations (bumped to 10000) cut short by `maxDur` timeout
```
------------------------------
• [1.000 seconds]
Translation - Benchmarking Tests Benchmark table (upstream scale) 1 endpoint(s), 10 upstream(s) [performance]
/Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:119

  Report Entries >>
  Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s) - /Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:98 @ 06/06/23 15:26:31.648
    Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s)
    Name                                                      | N    | Min   | Median | Mean  | StdDev | Max  
    ==========================================================================================================
    (upstream scale) 1 endpoint(s), 10 upstream(s) [duration] | 1190 | 700µs | 700µs  | 800µs | 200µs  | 6.2ms
  << Report Entries
------------------------------
```

Failing (90th percentile threshold reduced to 1ms)
```
------------------------------
• [FAILED] [0.834 seconds]
Translation - Benchmarking Tests Benchmark table [It] (upstream scale) 1 endpoint(s), 10 upstream(s) [performance]
/Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:119

  Timeline >>
  Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s) - /Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:98 @ 06/06/23 15:29:06.958
    Experiment - (upstream scale) 1 endpoint(s), 10 upstream(s)
    Name                                                      | N    | Min   | Median | Mean  | StdDev | Max  
    ==========================================================================================================
    (upstream scale) 1 endpoint(s), 10 upstream(s) [duration] | 1000 | 700µs | 700µs  | 800µs | 300µs  | 5.5ms
  [FAILED] in [It] - /Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:115 @ 06/06/23 15:29:07.792
  << Timeline

  [FAILED] Expected
      <time.Duration>: 1039522
  to be <
      <time.Duration>: 1000000
  In [It] at: /Users/bernie/go/src/github.com/solo-io/gloo/projects/gloo/pkg/translator/performance_test.go:115 @ 06/06/23 15:29:07.792
```

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/4920


test/helpers/resources.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+var upName = &core.Metadata{
+	Name:      "test",
+	Namespace: "gloo-system",
+}
+
+// Upstream is a generic upstream included in snapshots generated from ScaledSnapshot
+var Upstream = &v1.Upstream{
+	Metadata: upName,
```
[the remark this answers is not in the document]
bewebi: I haven't really put thought or research into whether this is relevant for translation benchmarking, but if it is we should

test/helpers/resources.go
```diff
@@ -0,0 +1,265 @@
...
+
+var routes = []*v1.Route{{
+	Name:     "testRouteName",
+	Matchers: []*matchers.Matcher{matcher},
+	Action: &v1.Route_RouteAction{
+		RouteAction: &v1.RouteAction{
+			Destination: &v1.RouteAction_Single{
+				Single: &v1.Destination{
+					DestinationType: &v1.Destination_Upstream{
+						Upstream: upName.Ref(),
```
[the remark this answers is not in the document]
bewebi: I haven't really put thought or research into whether this is relevant for translation benchmarking, but if it is we should adjust accordingly

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+			experiment := gmeasure.NewExperiment(fmt.Sprintf("Experiment - %s", desc))
+
+			AddReportEntry(experiment.Name, experiment)
+
+			experiment.Sample(func(idx int) {
+
+				// Time translation
+				experiment.MeasureDuration(desc, func() {
+					snap, errs, report = translator.Translate(params, gloohelpers.Proxy())
```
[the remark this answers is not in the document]
bewebi: Yeah this can/should be added...I'll take a pass on this PR and if it balloons in scope/time investment we can punt to a follow-up

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,187 @@
...
+				Expect(report).To(Equal(validationutils.MakeReport(gloohelpers.Proxy())))
+			}, gmeasure.SamplingConfig{N: config.iterations, Duration: config.maxDur})
+
+			durations := experiment.Get(desc).Durations
+
+			Expect(durations).Should(And(config.benchmarkMatchers...))
+		},
+		generateDesc, // generate descriptions for table entries with nil descriptions
+		Entry("basic", basicSnap, basicConfig),
+		Entry(nil, gloohelpers.ScaledSnapshot(gloohelpers.ScaleConfig{
```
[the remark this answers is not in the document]
bewebi: I think there is value in having examples that demonstrate different aspects of the framework, but there can probably be fewer

test/helpers/resources.go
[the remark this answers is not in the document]
sam-heilbron: All good! i think this is a good first pass, and we're better off introducing the changes and then we can regroup in the coming days if we want to expand these now (or wait til later)

projects/gloo/pkg/translator/performance_test.go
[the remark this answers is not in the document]
bewebi: Looking closer, this is in a tricky spot between time investment and necessity
I think ideally I'd like feedback from the team as to whether:
- This is useful
- It's worthwhile for me to do now/as a follow-up rather than have a developer attempt it later

I see all of that logic being encompassed in `ScaledSnapshot()` and I think it's relatively doable too, but probably a ~S size issue on its own, since it starts to warrant its own unit testing

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+// benchmarkConfig allows configuration for benchmarking tests to be reused for similar cases
+// This struct can be factored out to an accessible location should additional benchmarking suites be added
+type benchmarkConfig struct {
+	iterations        int                   // the number of iterations to attempt for a particular entry
+	maxDur            time.Duration         // the maximum time to spend on a particular entry even if not all iterations are complete
+	benchmarkMatchers []types.GomegaMatcher // matchers representing the assertions we wish to make for a particular entry
+}
+
+var _ = FDescribe("Translation - Benchmarking Tests", Serial, Label(labels.Performance), func() {
```
sam-heilbron, line 45: This test is focused. It didnt fail on the PR, because the peformance tests don't run on the PR. But it will cause our nightly tests to fail once this merges

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,194 @@
...
+				errs   reporter.ResourceReports
+				report *validation.ProxyReport
+			)
+
+			params := plugins.Params{
+				Ctx:      context.Background(),
+				Snapshot: apiSnap,
+			}
+
+			if len(apiSnap.Proxies) > 0 {
```
sam-heilbron, line 95: nit: I think having default logic (if proxies not provided, fallback to default proxy) can have the effect of introducing behavior that devs don't mean. I'd prefer that we require test authors to provide a proxy in the snapshot and we can error loudly if they dont'

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,194 @@
...
+				// Time translation
+				experiment.MeasureDuration(desc, func() {
+					snap, errs, report = translator.Translate(params, proxy)
+				})
+
+				// Assert expected results
+				Expect(errs.Validate()).NotTo(HaveOccurred())
+				Expect(snap).NotTo(BeNil())
+				Expect(report).To(Equal(validationutils.MakeReport(gloohelpers.Proxy())))
+			}, gmeasure.SamplingConfig{N: config.iterations, Duration: config.maxDur})
```
sam-heilbron, line 118: is this report assertion true if someone passes a custom proxy object (ie not the gloohelpers.Proxy)? One option would be that we don't assert on reports, since our goal isn't as much accuracy-focused. I like the thoroughness of the test, so if there's a way to assert on the reports, but still allow people to pass custom proxiers for each test I like it

projects/gloo/pkg/translator/performance_test.go
bewebi: That was intentional while it was still a draft PR, but an oversight at this point

projects/gloo/pkg/translator/performance_test.go
bewebi: Good point that the helper proxy is not always what we're using here...I think it would be sufficient to pass the actual proxy used in the test

I do think it's important that the test fail if the translation doesn't baseline behave as expected and I _think_ making the assertion about reports is reasonable in that regard, but no problem removing it if the `errs` and `snap` assertions are sufficient there
I'm not actually sure what would lead the reports to have an unexpected value which is why it seemed reasonable to leave it

Separately, perhaps it would make sense to only do these assertions on the first pass? Translation is deterministic so it's overkill to do it 1000 times

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,194 @@
...
+	iterations: 1000,
+	maxDur:     time.Second,
+	benchmarkMatchers: []types.GomegaMatcher{
+		matchers.HaveMedianLessThan(5 * time.Millisecond),
+		matchers.HavePercentileLessThan(90, 10*time.Millisecond),
+	},
+}
+
+/* 1k scale Tests */
+var oneKConfig = benchmarkConfig{
```
sam-heilbron, line 187: I thought I had commented on this before, but not sure if github is acting up or I never submitted, so apologies if it's the second time.

The naming here is a bit confusing. We have bsicConfig which runs 1000 iterations and 1kConfig which runs 500 iterations, and has a higher expected median. 

I imagine the names are based off of the number of resources that we're supplying to the test and not the number of iterations for each test?

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	benchmarkMatchers []types.GomegaMatcher // matchers representing the assertions we wish to make for a particular entry
+}
+
+var _ = Describe("Translation - Benchmarking Tests", Serial, Label(labels.Performance), func() {
+	var (
+		ctrl       *gomock.Controller
+		settings   *v1.Settings
+		translator Translator
+
+		registeredPlugins []plugins.Plugin
```
sam-heilbron, line 51: It seems like the tests only use a translator, and these variables are used in the BeforeEach to help initialize the translator. Can we remove them from these variables since they might confuse devs?

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+			experiment.Sample(func(idx int) {
+
+				// Time translation
+				experiment.MeasureDuration(desc, func() {
+					snap, errs, report = translator.Translate(params, proxy)
+				})
+
+				if idx == 0 {
+					fmt.Println("ASSERTING ONCE FOR", desc)
```
sam-heilbron, line 115: Let's remove the print, or just turn it into a logger from the context

projects/gloo/pkg/translator/performance_test.go
sam-heilbron: I still see that this is included, are you planning on removing this else conditional?

sam-heilbron (approved):
Looking good! A few questions that aren't necessarily blocking

projects/gloo/pkg/translator/performance_test.go
bewebi: leftover from development, not meant to be a part of the PR in any form

test/helpers/resources.go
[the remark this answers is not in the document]
jenshu: i agree that we should use unique upstream names (same for Endpoint names and any other resource we generate multiples of), to generate a more realistic snapshot.
if we use the same name for every upstream then some operations like [Find](https://github.com/solo-io/gloo/blob/416ca543b7853522249e01b4eac93aabdfa748c3/projects/gloo/pkg/api/v1/upstream.sk.go#L71) will always return quickly which would not necessarily reflect reality

test/helpers/resources.go
```diff
@@ -0,0 +1,267 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// hybridListener invokes functions that contain assertions and therefore can only be invoked from within a test block
```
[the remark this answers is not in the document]
bewebi: Yeah, it's a bit buried
The `Certificate()` function, invoked as part of the tcpListener [here](https://github.com/solo-io/gloo/blob/4dfea0d8eacd7d4741405846582ef80f6bffcb73/test/helpers/resources.go#L116), invokes [`GetCerts()`](https://github.com/solo-io/gloo/blob/764c2832964984294edbb4fee5d036cd63992279/test/helpers/certs.go#L74) a couple layers down, which has `Fail()` and `Expect()` calls
As a result, if those functions are assigned as variables outside of a Ginkgo test block, compilation fails

test/helpers/resources.go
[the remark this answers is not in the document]
bewebi: I've taken a pass at this and did observe that processing time for 1k Upstreams was dramatically higher when names were unique and each was referenced by a route

test/helpers/resources.go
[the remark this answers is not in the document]
nfuden: Perhaps we also add some notes about how this has multiple disjoint filterchains?

test/helpers/resources.go
```diff
@@ -0,0 +1,255 @@
...
+						Addr: "Test",
+						Port: 124,
+					},
+				},
+			},
+		},
+	}
+}
+
+// Endpoint is a generic endpoint included in snapshots generated from ScaledSnapshot
```
nfuden, line 48:
```suggestion
// Endpoint is a generic endpoint included in snapshots generated from ScaledSnapshot
// These endpoints differ by ref and name based on their passed in int
```

test/gomega/transforms/benchmark.go
```diff
@@ -0,0 +1,32 @@
...
+	return func(durations []time.Duration) time.Duration {
+		sort.Slice(durations, func(i, j int) bool {
+			return durations[i] < durations[j]
+		})
+		return durations[int(float64(len(durations))*(float64(percentile)/float64(100)))]
+	}
+}
+
+// WithMedian returns a function that extracts the value at the median from a slice of durations
+func WithMedian() func([]time.Duration) time.Duration {
```
jenshu, line 19: i suppose you've already looked into this, but is there anything built into gmeasure that already can do something similar (for both the functions here, and in `matchers/benchmark.go` above)? e.g. there's some mention here of being able to extract StatMedian from the stats https://onsi.github.io/gomega/#stats-and-rankings-comparing-measurements

bewebi: We can derive the median like so:
```
median := experiment.GetStats(desc).ValueFor(gmeasure.StatMedian)
```
and then make an assertion about that value

This might be more performant than using the matcher which sorts the slice of durations and determines the median from there, however the ability to have all of the benchmark matchers take a `[]time.Duration` is worth that tradeoff IMO

There isn't anything provided in terms of percentiles
At this time we are not comparing algorithms in order to find a winner, so those utilities aren't useful to us

jenshu (approved):

sam-heilbron (approved):

test/helpers/resources.go
```diff
@@ -0,0 +1,255 @@
...
+// Additional fields should be added as needed
+type ScaleConfig struct {
+	Endpoints int
+	Upstreams int
+}
+
+func upName(i int) *core.Metadata {
+	return &core.Metadata{
+		Name:      fmt.Sprintf("test-%06d", i),
+		Namespace: "gloo-system",
```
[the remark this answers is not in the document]
bewebi: I think you were reviewing an out-of-date state of the branch, I did [switch to a better name](https://github.com/solo-io/gloo/pull/8364/files#diff-5b48759bddf50dddafbe90032303a7709f8cb01b185c650cb713704c030dd261R24)

I did intend to forward-pad the upstream name with 0's to make for cleaner sorting (though I don't expect anyone to be expecting these resources in the normal course of things)

Will switch to the default package for namespace in the follow-up PR

PR #8373, opened 2023-06-07
Stability: Cleanup Envoy/Services Package References

sam-heilbron:
# Description

Cleanup `envoy.go` and other test services

# Context

Remove `envoy.go` which contained aliases to functions/types in the envoy package. Migrate the tests to call that directly

## EnvoyInstance
A simple find/replace to convert a `services.EnvoyInstance` into an `envoy.Instance`

## EnvoyFactory
A simple find/replace to convert a `services.EnvoyFactory` into an `envoy.Factory`

## Ports
This was slightly more tricky, and I tried to document the challenge in the code. 

There are 2 sides of ports that must be kept in sync:
- The resource (gateway) must be configured with a port
- The proxy (envoy) must be run port-forwarding that port for local requests

Historically, both of these definitions were pulled from the defaults package. However, each test had to update the port dynamically `services.NextBindPort` to ensure that subsequent tests wouldn't try to use ports that were in use. 

I tried to move port management of the envoy service to be internal to the envoy package. This ensure that all tests which needed the envoy ports, would reference the envoy instance itself, instead of the defaults package. Also, the envoy factory is responsible for using different ports, so the tests are unaware. Thanks to @jbohanon for the idea!

A challenge is that our tests still use some resource fixtures, that are created from the defaults package. As a result, I needed to keep the defaults definitions for ports, and adjust them whenver we adjusted the envoy ports. To keep this complexity hidden from the users, I do this inside an internal function.

## Enterprise Impact
This code, and previous gloo PRs are all incorporated in https://github.com/solo-io/solo-projects/pull/5070. It is up to date and passing build-bot.

# Checklist:

- [X] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/main/changelogutils) which references the issue that is resolved.


ben-taussig-solo (commented):
Looks fine, although I'm not sure I fully understand the changes made to how we handle ports. I think part of that is I didn't really have a strong understanding of how things were handled before. Additionally, how do we expect this to affect solo-projects? Will we need to make similar changes there after bumping the gloo dependency?

test/services/envoy/factory.go
```diff
@@ -207,15 +206,24 @@ func (f *factoryImpl) NewEnvoyInstance() (*Instance, error) {
 		}
 	}
 
+	// Ensure that each Instance has a unique set of ports
```
sam-heilbron, line 209: todo: clarify this better. what does it do? what does it not yet do?

test/services/envoy/ports.go
```diff
@@ -3,22 +3,54 @@ package envoy
...
+
 	"github.com/solo-io/gloo/test/ginkgo/parallel"
 )
 
 var (
-	adminPort = uint32(20000)
-	bindPort  = uint32(10080)
+	bindPort = uint32(10080)
+
+	adminPort  = defaults.GlooAdminPort
```
sam-heilbron, line 14: todo: remove duplicate variables

test/e2e/grpcweb_test.go
```diff
@@ -87,7 +87,7 @@ var _ = Describe("Grpc Web", func() {
 		)
 
 		BeforeEach(func() {
-			accessLogPort := services.NextBindPort()
+			accessLogPort := envoy.NextBindPort()
```
ben-taussig-solo, line 90: Is this an oversight or an exception to the rule that individual tests shouldn't increment ports? If it is an exception, can you leave a comment explaining why this is necessary?

sam-heilbron: Good call. It was an exception to the rule, but it didn't need to be. I just pushed a commit to try to account for this.

test/services/envoy/ports.go
```diff
@@ -9,19 +9,24 @@ import (
...
+	baseHttpPort      = defaults.HttpPort
+	baseHttpsPort     = defaults.HttpsPort
+	baseTcpPort       = defaults.TcpPort
+	baseHybridPort    = defaults.HybridPort
+	baseAccessLogPort = uint32(10080)
 )
 
-func NextBindPort() uint32 {
-	return advancePort(&bindPort)
+func NextAccessLogPort() uint32 {
```
ben-taussig-solo, line 20: Does this ever get called?

sam-heilbron: Yeah this is called inside the factory implementation, when a NewInstance is created

test/services/envoy/ports.go
```diff
@@ -3,22 +3,32 @@ package envoy
...
 
+	"github.com/solo-io/gloo/projects/gloo/pkg/defaults"
+
 	"github.com/solo-io/gloo/test/ginkgo/parallel"
 )
 
 var (
-	adminPort = uint32(20000)
-	bindPort  = uint32(10080)
+	baseAdminPort     = defaults.EnvoyAdminPort
```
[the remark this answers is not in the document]
ben-taussig-solo: This is fine for me -- I know that Jacob had raised this concern initially, so I'm going to leave this thread open so he has space to push back. In the meantime, I'm happy to approve the PR

ben-taussig-solo (approved):

test/services/envoy/ports.go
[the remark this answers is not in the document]
jbohanon: I'm good with this. I think moving it here and calling it `base*Port` is more expressive.

test/services/nats.go
jbohanon, line 1: :tada:

jbohanon (approved):

PR #8376, opened 2023-06-08
Correct Envoy-Gloo image name in e2e tests

ben-taussig-solo:
# Description
 - Resolves an error introduced in https://github.com/solo-io/gloo/pull/8365 that caused e2e tests to attempt to use `quay.io/solo-io/envoy-gloo-wrapper` instead of `quay.io/solo-io/gloo-envoy-wrapper` when running e2e tests
 - I encountered this error when trying to run e2e tests using a locally built gloo-envoy-wrapper image
   - The command I used was `ENVOY_IMAGE_TAG=<TAG> TEST_PKG=./test/e2e make test`
 - see here: https://github.com/solo-io/gloo/blob/686422e2a33688ddd3cf95cb74f87a388f184f2c/Makefile#L515 the image should be called `gloo-envoy-wrapper`

### DIFF
--- a/changelog/v1.15.0-beta11/correct-envoy-gloo-image-name.yaml
+++ b/changelog/v1.15.0-beta11/correct-envoy-gloo-image-name.yaml
@@ -1,0 +1,7 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/7577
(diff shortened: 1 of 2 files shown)


sam-heilbron (approved):
Thanks for taking care of this!

PR #8377, opened 2023-06-08
Refactor test services download logic

jbohanon:
# Description

Factor out the code for acquiring a test service binary. The behavior is unchanged, simply refactored. We first check for a specific environment variable specifying where to find the binary, then check the test executor's `$PATH`, then resort to running a docker container and extracting the binary from the running container. By using this docker method, we can leverage docker's local image cache to avoid downloading the dependency every time the tests are run without having to specify any additional environment variables when running locally and rapidly iterating.

# Checklist:

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -69,6 +69,7 @@ var _ = Describe("AWS Lambda", func() {
 	)
 
 	BeforeEach(func() {
+		testutils.ValidateRequirementsAndNotifyGinkgo(testutils.AwsCredentials())
 		httpClient = http.DefaultClient
 		httpClient.Timeout = 10 * time.Second
 		runOptions = &services.RunOptions{
--- a/changelog/v1.15.0-beta12/refactor-docker-service-download.yaml
+++ b/changelog/v1.15.0-beta12/refactor-docker-service-download.yaml
@@ -1,0 +1,5 @@
+changelog:
+- type: NON_USER_FACING
+  description: >-
+    Extract logic for downloading test service binaries into a helper to de-duplicate
+    and increase the readability of the factory code.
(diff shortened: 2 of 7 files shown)


sam-heilbron (approved):

test/services/download.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/onsi/ginkgo/v2"
+)
+
+type downloadParams struct {
+	fileUrl        string
+	fileName       string
+	destinationDir string
+}
+
+func downloadZippedFile(params downloadParams) (string, error) {
```
jenshu, line 24: nit: can you add some comments in this func so it's easier to follow along?

jbohanon:
I have tried to standardize the process of downloading these service binaries by centralizing the download code and using a direct-download. This works for consul and vault because their releases are published for public download, but this won't work for our envoy binary because it's behind an ACL in GCP. Another consideration I just thought of while testing this is that by getting it from Docker, we offload the ability to cache the download to Docker. By making this change, one would be _required_ to specify the envoy path as an env var on the command line to avoid downloading the file every time. 

This leaves us with a handful of options:
1. Close this PR and leave it exactly how it is
2. Review the ACLs on the binary artifacts in GCP and potentially use a direct download for Envoy as well as the others
3. Direct download the HashiCorp services and keep the docker method for Envoy
4. Keep the Docker method for all services, but centralize the logic to de-duplicate and increase the readability of the services' factory code.

Especially considering the fact that Docker will cache the download, I am leaning toward option 4.

test/e2e/aws_test.go
jbohanon, line 1: Not sure if there's a reason this requirement wasn't in here, but I was failing locally on the AWS tests because on no AWS creds, and adding this validation fixed it

sam-heilbron (approved):
🚀

jenshu (approved):

PR #8383, opened 2023-06-12
OPA object support

tjons:
# Description

This PR adds a new field to the OpaAuthOptions API that we plan to use in `ext-auth-service` to allow users to return an optional decision reason from evaluating OPA policies. https://soloio.slab.com/posts/opa-response-passthru-design-doc-3lldb91f has more context on how we plan to use this in platform. 

Corresponding `ext-auth-service` PR here: https://github.com/solo-io/ext-auth-service/pull/566

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1042,6 +1042,9 @@ message OpaAuthOptions {
     // are included in the request input. All other fields are dropped. Dropped fields will not be evaluated by the OPA engine.
     // By default, this is set to false and all fields are evaluated by OPA.
     bool fast_input_conversion = 1;
+
+    // Return the reason given from the OPA engine after a decision made on this policy
```
marcogschmidt, line 1046: It's not clear from the comment how the decision will be returned. We should be specific in these user facing docs. From the ext auth PR I see that it will be returned in the dynamic metadata. We should mention that and also specify:
- under which key it will be available in the metadata
- what the structure of the entry will be

Related to these two points, I will leave a couple of comment in the ext auth PR.

marcogschmidt (changes requested):

sam-heilbron:
/kick 
Build-bot flaked with a new error, so I created https://github.com/solo-io/gloo/issues/8463 to track it.

sam-heilbron (dismissed):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1078,6 +1078,14 @@ message OpaAuthOptions {
     // are included in the request input. All other fields are dropped. Dropped fields will not be evaluated by the OPA engine.
     // By default, this is set to false and all fields are evaluated by OPA.
     bool fast_input_conversion = 1;
+
+    // Return the reason given from the OPA engine after a decision made on this policy. Reason must be the second
+    // parameter of the query and will be a protobuf struct if the reason is an object, and otherwise will be a json
+    // string. The entry will be in the returned DynamicMetadata in the CheckResponse and the structure will be
+    // DynamicMetadata
+    //   -> Name of the AuthConfig
+    //       -> reason
```
marcogschmidt, line 1087: This might be a bit misleading:
- First because the name of the AuthConfig might actually interpreted as the name of the CR itself; we should probably call it something like ` name of the auth step, i.e. spec.configs[].name`. Unfortunately we don't have a clear naming convention for these "auth steps" and we seem to use "auth configs" ambiguously in the proto docs.
- Second because the metadata will be prefixed with the name of the ext auth filter when it's emitted by the `ext_authz` filter, so when a user actually gets to consume it the structure of the metadata will be:
```yaml
envoy.filters.http.ext_authz:
  myOpaStepName:
    reason: because
```

marcogschmidt (approved):

sam-heilbron (approved):

PR #8393, opened 2023-06-15
Fix Transformation Logging Global Setting

ben-taussig-solo:
# Description
 - There was a bug with https://github.com/solo-io/gloo/pull/8359 that prevented the global setting (`gloo.logTransformationRequestResponseInfo`) for transformation logging from being respected.
   - See API reference [here](https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/settings.proto.sk/#gloooptions)
 - This PR updates the transformation plugin to now respect that setting
 - Additionally, the PR adds an e2e test to validate this behavior


projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -274,7 +274,8 @@ func (p *Plugin) ConvertTransformation(
 	// it will override the transformation-level settings
 	logRequestResponseInfo := stagedTransformations.GetLogRequestResponseInfo().GetValue()
 
-	if logRequestResponseInfo {
+	// if the global settings or the route/vhost settings are set to true, set all transformation-level settings to true
+	if p.logRequestResponseInfo || logRequestResponseInfo {
```
nfuden, line 278: I think the behavior we want with the child overrides (per the data plane pr) do we think its better to do the following?

```suggestion
	if stagedTransformations.GetLogRequestResponseInfo() == true ||  (stagedTransformations.GetLogRequestResponseInfo() == nil && p.logRequestResponseInfo)
```

ben-taussig-solo: good call, thanks!

jenshu (approved):

ashishb-solo (approved):

Issue #8404, opened 2023-06-20
Creating several GatewayProxies with Helm doesn't generate SSL Gateways


mtenrero:
### Gloo Edge Version

1.14.x (latest stable)

### Kubernetes Version

1.24.x

### Describe the bug

Due an error in the Helm template, when defining GatewayProxies with SSL support, those are not properly generated in some scenarios:

```yaml
---
    apiVersion: gateway.solo.io/v1
    kind: Gateway
    metadata:
      name: default
      namespace: default
      labels:
        app: gloo
    spec:
      bindAddress: "::"
      bindPort: 8080
      httpGateway:
        virtualServiceSelector:
          gateway-type: default
      useProxyProto: false
      ssl: false
      proxyNames:
      - default
    ---
    apiVersion: gateway.solo.io/v1
    kind: Gateway
    metadata:
      name: default-ssl
      namespace: default
      labels:
        app: gloo
    spec:
      bindAddress: "::"
      bindPort: 8443
      httpGateway:
        virtualServiceSelector:
          gateway-type: default
      useProxyProto: false
      ssl: true
      proxyNames:
      - default---
```

### Steps to reproduce the bug

1. Define several Gateway Proxies in Helm values
2. Install/Template Chart
3. SSL Gateways are not generated due to a missing newline char

### Expected Behavior

SSL Gateways should be correctly generated

jenshu:
fix will be available in Gloo OSS v1.15.0-beta23, v1.14.13, and v1.13.23

PR #8406, opened 2023-06-20
Fix yaml separation for ssl default gw

mtenrero:
# Description

Fixes YAML separation characters for SSL default gateways
https://github.com/solo-io/gloo/issues/8404

# Context

When setting SSL Default Gateways with Helm bad line break placements broke the ssl gateway creation


sam-heilbron (commented):
Thanks for the contribution @mtenrero! I see that this is the only instance of this case in our codebase. Can you add a test case that would fail without this change? It would help us to have this test to avoid people adding regressions to this.

Also, we tend to introduce code first to the main branch, and then backport those changes to our long term support branches after it has been merged into our stable (main) branch. Could you direct this PR to main?

ben-taussig-solo:
@mtenrero I am going to close this PR now due to inactivity. If you have time to add a test case, please feel free to reopen it and I will be happy to review it again. Thanks!

Issue #8405, opened 2023-06-20
Aggregate listener crashes when errors occur in tcp gateways


ashishb-solo:
### Gloo Edge Version

1.14.x (latest stable)

### Kubernetes Version

None

### Describe the bug

While working on https://github.com/solo-io/solo-projects/pull/5105, I noticed that pods were restarting and found a panic in the logs. The `report` field instantiated [here](https://github.com/solo-io/gloo/blob/aad361be4cca53df47830c4c7e31dca650831510/projects/gloo/pkg/translator/listener_subsystem.go#L375) is nil and a panic ensues whenn the error reporter tries to dereference it

### Steps to reproduce the bug

1. Run the tests in https://github.com/solo-io/solo-projects/pull/5105
2. 2. watch the logs for the gloo pods (`kubectl -n gloo-system logs deployments/gloo -f`)
3. 3. Panics will occur and be reported in the log (somewhat intermittent behaviour, only occurs when an error is reported

### Expected Behavior

Should not crash

ashishb-solo:
blocked by https://github.com/solo-io/solo-projects/issues/5156

PR #8407, opened 2023-06-20
8405 aggregate listener crash

ashishb-solo:
# Description

Fix a crash in aggregate filter translator - the panic was occurring because the report objects for the TCP listeners in the aggregate filter translators were never being instantiated. The issue link has a bit more more information on the cause.

# Context

Ran into this issue while working on the pull request linked in [the issue link](https://github.com/solo-io/gloo/issues/8405)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8405


projects/gloo/pkg/utils/validation/proxy_validation.go
```diff
@@ -69,23 +69,32 @@ func MakeReport(proxy *v1.Proxy) *validation.ProxyReport {
...
 				}
-				httpListenerReports[utils.MatchedRouteConfigName(listener, httpFilterChain.GetMatcher())] = httListenerReport
+				httpListenerReports[utils.MatchedRouteConfigName(listener, httpFilterChain.GetMatcher())] = httpListenerReport
+			}
+
+			for _, tcpListener := range listenerType.AggregateListener.GetTcpListeners() {
+				tcpListenerReport := &validation.TcpListenerReport{
+					TcpHostReports: makeTcpHostReports(tcpListener.GetTcpListener().GetTcpHosts()),
+				}
+				tcpListenerReports[utils.MatchedRouteConfigName(listener, tcpListener.GetMatcher())] = tcpListenerReport
```
nfuden, line 90: Are we guaranteed to have a matcher here?

ashishb-solo: I think `GetMatcher()` will return `nil` if not, and if that's the case, then `MatchedRouteConfigName` seems to handle that case: https://github.com/solo-io/gloo/blob/9202582cbbdffc6ff055365fc2eb272c52536fe1/projects/gloo/pkg/utils/names.go#L22

nfuden: Good yeah the matchedrouteconfig on nil part is what i was looking for

nfuden (approved):

projects/gloo/pkg/utils/validation/proxy_validation.go
```diff
@@ -69,23 +69,32 @@ func MakeReport(proxy *v1.Proxy) *validation.ProxyReport {
...
 
-				httListenerReport := &validation.HttpListenerReport{
+				httpListenerReport := &validation.HttpListenerReport{
 					VirtualHostReports: makeVhostReports(virtualHosts),
 				}
-				httpListenerReports[utils.MatchedRouteConfigName(listener, httpFilterChain.GetMatcher())] = httListenerReport
+				httpListenerReports[utils.MatchedRouteConfigName(listener, httpFilterChain.GetMatcher())] = httpListenerReport
+			}
+
+			for _, tcpListener := range listenerType.AggregateListener.GetTcpListeners() {
```
jenshu, line 86: do we also need to add the tcp listener reports to the AggregateListener case in `GetProxyError` below?

ashishb-solo: i believe we do, great catch

ashishb-solo: Okay, this is kinda interesting. It turns out that it doesn't seem to be necessary to catch this error, because `GetProxyError` calls `GetListenerError`[source](https://github.com/solo-io/gloo/blob/aad361be4cca53df47830c4c7e31dca650831510/projects/gloo/pkg/utils/validation/proxy_validation.go#L195) which pulls out the listener errors that occur on the TCP filters of the aggregate listener. I have tested this fairly extensively and can see that it already has working tests.

It is quite unfortunate that `proxy_validation_test.go` does not have very good test coverage for aggregate filter chain translators, but I feel that's outside the scope of this bugfix.... But I will definitely do due diligence if we feel there are any other potential errors that we're not catching here.

ashishb-solo: That said, I am happy to add this if we feel it's necessary:

```
ashish@localhost:~/go/src/github.com/solo-io/gloo/projects/gloo/pkg/utils/validation$ git diff proxy_validation.go 
diff --git a/projects/gloo/pkg/utils/validation/proxy_validation.go b/projects/gloo/pkg/utils/validation/proxy_validation.go
index 4de0554d3..18320a48b 100644
--- a/projects/gloo/pkg/utils/validation/proxy_validation.go
+++ b/projects/gloo/pkg/utils/validation/proxy_validation.go
@@ -225,6 +225,9 @@ func GetProxyError(proxyRpt *validation.ProxyReport) error {
                        for _, httpListenerReport := range listenerType.AggregateListenerReport.GetHttpListenerReports() {
                                errs = append(errs, getHttpListenerReportErrs(httpListenerReport)...)
                        }
+                       for _, tcpListenerReport := range listenerType.AggregateListenerReport.GetTcpListenerReports() {
+                               errs = append(errs, getTcpListenerReportErrs(tcpListenerReport)...)
+                       }
                }
        }
```

All tests continue to pass as intended. But I'm inclined to leave it as is

ashishb-solo: That said, I think it's a good idea to add a test for catching errors in aggregate listeners since we don't have any already in `proxy_validation_test.go`. I have added it here: https://github.com/solo-io/gloo/pull/8407/commits/bd83416c57c4f691053e722f8b5d0ae44dd7d249

jenshu: if GetListenerErr is already extracting errors, then is some (or all) of the [switch statement](https://github.com/solo-io/gloo/blob/aad361be4cca53df47830c4c7e31dca650831510/projects/gloo/pkg/utils/validation/proxy_validation.go#L198-L219) unnecessary?

ashishb-solo: Yeah, it is. It looks like some errors get added multiple times, but somehow they get deduplicated. I have to investigate it a little deeperly

ashishb-solo: Okay, I looked into things a lot more closely, and have made a few findings:

1. Aggregate listeners were not properly checking tcp errors - specifically, [this function](https://github.com/solo-io/gloo/blob/9202582cbbdffc6ff055365fc2eb272c52536fe1/projects/gloo/pkg/utils/validation/proxy_validation.go#L230) was not being called for tcp listeners on aggregate filter chains. I've added some tests to hit those cases in TCP listeners on aggregate filter chains in c6b5f7213c7310c6efcea132d5ddb8f1ff850981

2. in order to make those tests pass, it was necessary to add [this change](https://github.com/solo-io/gloo/pull/8407/commits/c6b5f7213c7310c6efcea132d5ddb8f1ff850981#diff-d8eb0189e46d1e52cd986c9ca9442663af1ea2287d7757d6483fd4926ef6f16dR228-R230) - so your call-out here was on point, we were not properly catching these errors before

3. "if GetListenerErr is already extracting errors, then is some (or all) of the [switch statement](https://github.com/solo-io/gloo/blob/aad361be4cca53df47830c4c7e31dca650831510/projects/gloo/pkg/utils/validation/proxy_validation.go#L198-L219) unnecessary?" It's possible, but I think not. Those tests are returning slices that will be empty if no errors are found.

4. Lamentably, we still don't seem to have tests for catching HTTP errors in the aggregate listeners. In the interest of time, I would prefer if we didn't add those in this pull request :-P but i think it would be good to get that tightened up. Happy to do so now if we feel that it's appropriate. It is all about trying to be as correct and thorough as we can after all!

projects/gloo/pkg/utils/validation/proxy_validation_test.go
```diff
@@ -115,6 +117,23 @@ var _ = Describe("validation utils", func() {
 		}
 		return proxy
 	}
+	makeAggregateListenerProxy := func() *v1.Proxy {
```
ben-taussig-solo, line 120: Nit: Based on the test below it looks like the point of this helper is to create an invalid aggregate listener that will produce a TCP Listener Report. Could you consider renaming this helper (maybe `makeInvalidAggregateListenerProxy`) or leaving a comment clarifying this point?

projects/gloo/pkg/utils/validation/proxy_validation_test.go
```diff
@@ -225,6 +254,53 @@ var _ = Describe("validation utils", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("VirtualHost Error: DomainsNotUniqueError. Reason: domains not unique; Listener Error: BindPortNotUniqueError. Reason: bind port not unique; HttpListener Error: ProcessingError. Reason: bad http plugin"))
 		})
+		It("aggregates the errors at every level for aggregate hybrid listener", func() {
```
jenshu, line 257:
```suggestion
		It("aggregates the errors at every level for aggregate listener", func() {
```

projects/gloo/pkg/utils/validation/proxy_validation_test.go
```diff
@@ -225,6 +254,53 @@ var _ = Describe("validation utils", func() {
...
+			rpt := MakeReport(proxy)
+
+			rpt.ListenerReports[0].Errors = append(rpt.ListenerReports[0].Errors,
+				&validation.ListenerReport_Error{
+					Type:   validation.ListenerReport_Error_BindPortNotUniqueError,
+					Reason: "bind port not unique",
+				},
+			)
+
+			// aggregate listener reports are a map of tcp listener reports -
```
jenshu, line 268: for both this "aggregate listener" test and the "aggregates the errors at every level for hybrid listener" test, can we test with a combination of http and tcp errors?

also should we have an additional test that's tcp only? (to mirror the "aggregates the errors at every level for http listener" test)

projects/gloo/pkg/utils/validation/proxy_validation_test.go
ashishb-solo: changed the name here 3bdbf75d42748419b46f958c962190fe8eed1196 to `makeInvalidAggregateListenerProxyTcp`

projects/gloo/pkg/utils/validation/proxy_validation_test.go
ashishb-solo: [This test](https://github.com/solo-io/gloo/blob/6ac54bc4f9c13037770537dec86be0741d51b90b/projects/gloo/pkg/utils/validation/proxy_validation_test.go#L257) already is a tcp-only test for aggregate filter chains. I'll try to look into adding HTTP errors as well

ashishb-solo: i've added tcp tests for hybrid listeners in here: https://github.com/solo-io/gloo/blob/6ac54bc4f9c13037770537dec86be0741d51b90b/projects/gloo/pkg/utils/validation/proxy_validation_test.go#L257

and http tests here: 76222dd4d9f1ba864c0289a683be343f83213bc5

ashishb-solo: note that the http tests aren't 100% complete, but they should still demonstrate that we hit that codepath so hopefully that's sufficient as a test

jenshu (approved):

ben-taussig-solo (approved):

PR #8415, opened 2023-06-22
Performance test snapshot generation

bewebi:
# Description

- Restructure scaled snapshot generation to use Builder paradigm
  - Unit test 
- Add tests for consistent and unique Upstream SNI cases

# Context

The Builders paradigm is likely to be easier to use and more extensible than using a struct to define scaled Snapshot parameters

As an initial example we add Upstream SNI tests

A manual run of the Performance tests from this branch can be found [here](https://github.com/solo-io/gloo/actions/runs/5350781520/jobs/9703829376)

Note that the test did not demonstrate a meaningful difference in performance for the two cases:
Consistent SNI:
```
  Experiment - (consistent SNI, upstream scale) 1 endpoint(s), 10 upstream(s) - /home/runner/work/gloo/gloo/projects/gloo/pkg/translator/performance_linux_test.go:114 @ 06/22/23 22:37:24.977
    Experiment - (consistent SNI, upstream scale) 1 endpoint(s), 10 upstream(s)
    Name                                     | N   | Min    | Median | Mean   | StdDev | Max 
    =========================================================================================
    (consistent SNI, upstream scale) 1       | 598 | 13.4ms | 15.9ms | 16.4ms | 1.4ms  | 20ms
    endpoint(s), 10 upstream(s) [duration]   |     |        |        |        |        |     
```
Unique SNI:
```
  Experiment - (unique SNI, upstream scale) 1 endpoint(s), 10 upstream(s) - /home/runner/work/gloo/gloo/projects/gloo/pkg/translator/performance_linux_test.go:114 @ 06/22/23 22:37:04.565
    Experiment - (unique SNI, upstream scale) 1 endpoint(s), 10 upstream(s)
    Name                                                                  | N   | Min    | Median | Mean   | StdDev | Max   
    ========================================================================================================================
    (unique SNI, upstream scale) 1 endpoint(s), 10 upstream(s) [duration] | 586 | 13.8ms | 16ms   | 16.8ms | 1.7ms  | 20.3ms
```

We can skip k8s tests and docs build as this PR does not touch production code or documentation

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5116


test/helpers/resources.go
```diff
@@ -3,24 +3,133 @@ package helpers
...
+		// The proxy should contain a route for each upstream
+		Proxies: []*v1.Proxy{Proxy(b.usCount)},
+
+		Endpoints: endpointList,
+		Upstreams: upstreamList,
+	}
+}
+
+// upstreamBuilder contains options for building Upstreams to be included in scaled Snapshots
+type upstreamBuilder struct {
```
sam-heilbron, line 80: I like this concept of having a builder per resource to make it easier to construct these objects. I found this useful for e2e tests, and created https://github.com/solo-io/gloo/blob/main/test/helpers/virtual_services.go. Could we add a new file for each of the builders? I figure they each may grow in complexiy, and I think having a file named for the resource might make it easier to find (even for people not working on performance tests)

test/helpers/resources.go
```diff
@@ -3,24 +3,133 @@ package helpers
...
+	b.epBuilder = eb
+	return b
+}
+
+// Build generates a snapshot populated with the specified number of each resource for the builder, using the
+// sub-resource builders to build each sub-resource
+func (b *scaledSnapshotBuilder) Build() *gloosnapshot.ApiSnapshot {
+	endpointList := make(v1.EndpointList, b.epCount)
+	for i := 0; i < b.epCount; i++ {
+		endpointList[i] = b.epBuilder.Build(i + 1) // names are 1-indexed
```
sam-heilbron, line 62: nit: I like the callout here, but I kind of think this is added complexity that could be avoided, and it might lead devs to think that the name here needs to match. Couldn't we either just loop starting at i=1 or just have names be 0-indexed based?
Any complexity in a loop or a build function makes me think that complexity is solving a particular problem

test/helpers/resources.go
```diff
@@ -3,24 +3,133 @@ package helpers
...
+// contains a builder for each sub-resource type which is responsible for building instances of that resource
 // Additional fields should be added as needed
-type ScaleConfig struct {
-	Endpoints int
-	Upstreams int
+type scaledSnapshotBuilder struct {
+	epCount int
+	usCount int
+
+	epBuilder *endpointBuilder
```
sam-heilbron, line 26: I like this idea here, of having the snapshot just delegate to resource specific builders, and the snapshot builder takes resource builders and multiplies them.

test/helpers/resources.go
```diff
@@ -3,24 +3,133 @@ package helpers
...
+		usBuilder: NewUpstreamBuilder(),
+	}
+}
+
+func (b *scaledSnapshotBuilder) WithUpstreamCount(n int) *scaledSnapshotBuilder {
+	b.usCount = n
+	return b
+}
+
+func (b *scaledSnapshotBuilder) WithUpstreamBuilder(ub *upstreamBuilder) *scaledSnapshotBuilder {
```
sam-heilbron, line 42: At the moment I don't quite see what this provides. We initialize the builder with the implementation for this, so calling WithUpstreamBuilder doesn't do anything. 

I think the idea here is to allow devs to write their own builder and inject that. I like that idea, but for it work I think we need to define an UptreamBuilder interface that has a Build() method which produces an Upstream.

test/helpers/resources.go
```diff
@@ -3,24 +3,133 @@ package helpers
...
+	switch b.sniPattern {
+	case uniqueSni:
+		up.SslConfig = &ssl.UpstreamSslConfig{
+			Sni: fmt.Sprintf("unique-domain-%d", i),
+		}
+	case consistentSni:
+		up.SslConfig = &ssl.UpstreamSslConfig{
+			Sni: "consistent-domain",
+		}
+	}
```
sam-heilbron, line 118: Is this a case that impacts performance or accuracy? I don't quite follow why, if I were writing a benchmark test, I would use each of these scenarios. Can you help me understand?

nfuden: While not a thing in Upstreams in VS the iteration time over the number of same sni when merging ssl config can have translation performance implications. Since we do not have virtual services here we used this as a way to talk about how resource updates may or may not be consistent across the whole generated set of a resource.

For example with isolatebyvirtualservice turned on having an incrementing versus static sni will have deduplication timing implications.

sam-heilbron: Given that, shouldn't we expose the settings object as well, so that users can control which settings are used?

sam-heilbron: Also does isolateVirutalHostsBySslConfig affect the Upstream resource (I dind't think it did)

test/helpers/resources.go
bewebi: See [here](https://github.com/solo-io/gloo/pull/8415/files#diff-ca62298fdc60983f462ff4959a8e087531362893244d4f502f05e1bdb743ec13R158-R159) for an example usage

>  We initialize the builder with the implementation for this, so calling WithUpstreamBuilder doesn't do anything
Presumably the user would pass a non-default `upstreamBuilder`, but you're right that if they wanted to use the default `upstreamBuilder` there would be no reason to invoke `WithUpstreamBuilder()`

It might be preferable to pass an exported type, allowing users to create sub-resource builders separately and re-use them
Nathan had raised something along these lines in [slack](https://solo-io-corp.slack.com/archives/C056AFWPB0X/p1687447789731879?thread_ts=1687369205.252839&cid=C056AFWPB0X)

I'm happy to take a pass at that given that multiple folks have made a similar suggestion

test/helpers/resources.go
bewebi: The case is mainly an example given to demonstrate the capabilities of the framework (as described in [the issue](https://github.com/solo-io/solo-projects/issues/5116))

I'm not sure which Settings object you're referring to, but there's every expectation that additional resource types and fields will be exposed as additional test cases call for it

sam-heilbron (approved):
awesome work!

test/helpers/resources.go
nfuden: yeah we expect upstreams to have no difference in usage between the two but benchmarking proves it out and was easier than implementing vs into translation benchmarking and setting the gateway settings. IE what is said here https://github.com/solo-io/gloo/pull/8415#discussion_r1240000161

sam-heilbron (approved):

nfuden (approved):
Awesome stuff I like this paradigm of testing both correctness (at least to some degree) as well as benchmark. That way we feel relatively good about the configuration being generated as expected.

A great step imo!

PR #8416, opened 2023-06-23
Stability: Support service-specific log level in tests

sam-heilbron:
# Description

Expose "SERVICE_LOG_LEVEL" an environment variable that can be used for developers to configure log levels for each tests service. This is intended to make debugging easier by being able to fine tune the level per service.

# Context

See the README additions if how this can be used is clear. I'd like the feature to be self-documenting, so I'm intentionally not adding more context here. If there are any questions about this, please raise them on the PR so I can update the dev documentation (readme) to clarify.

## Enterprise changes
This work is extended for enterprise services in https://github.com/solo-io/solo-projects/pull/5127

# Testing
Focus an e2e test (I used example_test), and pass in some options for the SERVICE_LOG_LEVEL and observe the difference in logs produced

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5006


test/e2e/example_test.go
```diff
@@ -72,6 +78,8 @@ var _ = Describe("Example E2E Test For Developers", Label(), func() {
 			Eventually(func(g Gomega) {
 				g.Expect(testutils.DefaultHttpClient.Do(requestBuilder.Build())).Should(matchers.HaveExactResponseBody(requestBody)) // The default server that we route to is an echo server
 			}, "5s", ".5s").Should(Succeed(), "POST with request body should return same body in response")
+
+			Expect(1).To(Equal(2))
```
nfuden, line 82: lol

test/services/logging.go
```diff
@@ -0,0 +1,93 @@
...
+	errors "github.com/rotisserie/eris"
+	"github.com/solo-io/gloo/test/testutils"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// _loggingConfigRegex is the format of the string that can be passed to configure the log level of services
+	// It is currently unused, but is here for reference
+	_loggingConfigRegex = "service=logLevel;service=logLevel"
```
ashishb-solo, line 18: would be great to document valid values for service (gloo, discovery, gateway-proxy, etc.) and logLevel (info, warn, etc.) here

test/services/logging.go
```diff
@@ -0,0 +1,93 @@
...
+		name := strings.Split(serviceLogPair, nameValueSeparator)[0]
+		logLevelStr := strings.Split(serviceLogPair, nameValueSeparator)[1]
+		logLevel, err := zapcore.ParseLevel(logLevelStr)
+		// We intentionally error loudly here
+		// This will occur if the user passes an invalid log level string
+		if err != nil {
+			panic(errors.Wrapf(err, "invalid log level string: %s", logLevelStr))
+		}
+
+		serviceLogLevel[name] = logLevel
```
ashishb-solo, line 44: nit: is there a strict list of allowed services here, or do we keep it dynamic? it would be nice if we can protect against typos and the like, but not a major issue if it's too hard to implement

test/services/logging.go
ashishb-solo: said documentation doesn't have to be here, by the way, it could also be in the makefile, or possibly linked from it

sam-heilbron: Yeah so I was torn here. On one hand, I thought about defining the list of services as constants, and keeping it strongly typed. On the flip side, this is a simple logging utility, and I figured the value it provides is more as a bridge betweeen some config mechanism (the env variable) and some log level for a service. By keeping it flexible, I figured it would mean that when new services are added (or modified) we wouldn't need to change this. Since there are also enterprise use cases (extauth, rate limit), I didn't want to have to define those in open source.

I agree the downside to this is there isn't a strict list of used services. What do you think about hte following as a compromise:
- Keep this utility generic and not define any specific services
- Document the currently used services (gatewway-proxy, gloo, discovery in oss, extauth, ratelimit, in enterprise) as part of a description of running e2e tests locally (in the README of each repo)

test/services/logging.go
sam-heilbron: I kept it dynamic (https://github.com/solo-io/gloo/pull/8416#discussion_r1240238403) with the hope of allowing new services or enterprise services to use this, without needing to change the implementation

test/services/logging.go
ashishb-solo: Yeah, I had the same thought process and pretty much come to the same conclusion. I think we're on the same page here, but glad that we both took it into consideration

test/e2e/README.md
```diff
@@ -39,26 +39,58 @@ The `run-tests` make target runs ginkgo with a set of useful flags. The followin
 | TEST_PKG          | ""      | The path to the package of the test suite you want to run                                                                                                                                                                                          |
 | WAIT_ON_FAIL      | 0       | Set to 1 to prevent Ginkgo from cleaning up the Gloo Edge installation in case of failure. Useful to exec into inspect resources created by the test. A command to resume the test run (and thus clean up resources) will be logged to the output. |
 | INVALID_TEST_REQS | fail    | The behavior for tests which depend on environment conditions that aren't satisfied. Options are `skip`, `run`, `fail`                                                                                                                             |
+| SERVICE_LOG_LEVEL | ""      | The log levels used for services. See "Controlling Log Verbosity of Services" below.                                                                                                                                                               |    
+
+#### Controlling Log Verbosity of Services
```
ashishb-solo, line 44: nice!

nit: can you comment on what happens if multiple values are provided? ie. if it's set to `gloo:info,gloo:debug` which one takes precedence?

test/e2e/staged_transformation_test.go
```diff
@@ -30,14 +33,26 @@ var _ = Describe("Staged Transformation", func() {
 
 	var (
 		testContext *e2e.TestContext
+
+		// This test relies on running the gateway-proxy with debug logging enabled
+		// This variable allows us to reset the original log level after the test
+		customAfterEach func()
```
ashishb-solo, line 39: nit: could we rename this to something more descriptive, say, `resetLogLevel`?

sam-heilbron: Yeah I'm happy to. I actually had originally done this, and then made it more generic in https://github.com/solo-io/gloo/pull/8416/commits/0d05638fe5cbfda5d4f2e4eb9a9e7e9111c1e71e. I'll update that function to demonstrate its the log level that is being reset. My original intention was to avoid pulling specifics of the tests into the variable naming, but I agree it's more explicit

test/e2e/README.md
sam-heilbron: I'm happy to. Do you think that's a case we need to support? I feel like documenting it will encourage people to use it that way, when I'd almost rather say its unsupported and have people explicitly set the value once. Is there a use case you see for allowing people to set it twice in the same env variable?

test/services/gloo.go
```diff
@@ -75,6 +75,12 @@ import (
 
 var glooPortBase = int32(30400)
 
+const (
+	GlooServiceName = "gloo"
```
nfuden, line 79: Question: Are we exporting these just so it mimicks the envoy service which needs to be referenced in enterprise? Or is there some other rationale for it? If its just to mimick the envoy servicenaming do you think it would be better to un-export these?

test/e2e/README.md
ashishb-solo: It's not a common use case, but this happens sometimes in scripts where users have a default value and then override it. Or when a user has it in the shell environment and then overrides it at the command line.

If we don't document it, then whatever behaviour it currently has becomes the undocumented default, so if the logic changes, then other environments (like our CI for example) could also hypothetically change. So in my opinion, it's a good idea to remove the ambiguity.

If we don't think those use cases are likely to happen, then I'm happy to overlook this

test/services/gloo.go
sam-heilbron: These are exported because tests (in open source and enterprise) need to set the log level to debug on the envoy insurance for them to work. I figured we're better off being consistent and exporting all service names, and making it easy for people to use the name or find it because it's exported. Do you have a concern with them being exported?

test/services/logging.go
```diff
@@ -0,0 +1,134 @@
...
+// This logger is configured with the appropriate log level
+func MustGetSugaredLogger(serviceName string) *zap.SugaredLogger {
+	logLevel := GetLogLevel(serviceName)
+
+	config := zap.NewProductionConfig()
+	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	config.Level.SetLevel(logLevel)
+
+	logger, err := config.Build()
+	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to build logger")
```
nfuden, line 76:
```suggestion
	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to build logger: " + serviceName)
```

test/services/logging.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return logger.Sugar()
+}
+
+// logProvider is a helper for managing the log level of multiple services
+type logProvider struct {
+	mutex sync.RWMutex
```
nfuden, line 83: nit: I personally find it cleaner to compose the mutex in cases like this.

It means that the calls of l.mutex.RLock() become l.RLock()

test/services/gloo.go
nfuden: Nah I just wanted to make sure we considered that these are not technically used at this time in their exported format and the usage pattern if we did choose to manipulate them in enterprise might look a little different from the envoy setup. I am content that we did this as a considered approach.

test/services/logging.go
sam-heilbron: I like it! YOu had had a similar comment ~1 year ago on another piece of work and I completely agree

nfuden (approved):

ashishb-solo (approved):

PR #8420, opened 2023-06-26
fix type mismatch for descriptor func

bewebi:
# Description

- Add `WithInjectedSnapshot()` to `ScaledSnapshotBuilder`, allowing the user to specify a snapshot for the builder to return directly
  - Unit test 
  - Use `WithInjectedSnapshot()` in raw snapshot table entry 
- Add getter functions to allow description generator to take builder instead of snapshot
  - Update `generateDesc()` accordingly
- Generated routes are 0-indexed, matching generation for other resources
- Add Darwin implementation of `MeasureIgnore0ns()` in order to allow for running performance tests locally

# Context

When working on #8415 we did not run the Nightly job performance tests for the final commit and therefore did not catch bugs that had been introduced, namely:
- A type mismatch was introduced because the `generateDesc()` func signature no longer matched the Entry signature for the "Benchmark table"
- Route generation was still based on 1-indexed names while Endpoints and Upstreams were switched to 0-indexing, leading to an off-by-one error and a translation warning due to routes referencing non-existent Upstreams

This PR resolves these bugs

We opt to add getter functions to `ScaledSnapshotBuilder` to allow `generateDesc()` to consume builders instead of snapshots and to add `NewInjectedSnapshotBuilder()` to allow arbitrary snapshots to be passed
The alternative would be to pass built Snapshots to each entry, like:
```
...
		Entry(nil, gloohelpers.NewScaledSnapshotBuilder().WithUpstreamCount(10).WithEndpointCount(1).Build(), basicConfig, "upstream scale"),
		Entry(nil, gloohelpers.NewScaledSnapshotBuilder().WithUpstreamCount(1000).WithEndpointCount(1).Build(), oneKUpstreamsConfig, "upstream scale"),
...
```
I am ambivalent between these choices, but calling `Build()` from within the test function seemed cleaner and, as `ScaledSnapshotBuilder` is an internal, test-only struct, there does not seem to be a downside to exposing getter functions on it
If reviewers feel the alternative is preferable we can go that route

In addition, we add a Darwin implementation for measurement, allowing tests to be run locally on Macs
Though the timing measurements are not actually meaningful as far as benchmarking, this will allow users to catch compilation and runtime errors without having to manually run the Nightly action
Users should continue to run the Nightly action when adding new tests in order to determine the benchmark targets

## Nightly build run(s)
Add additional links if additional commits are made
- [e146d3b](https://github.com/solo-io/gloo/actions/runs/5380603650/jobs/9763553477)
- [8d99208](https://github.com/solo-io/gloo/actions/runs/5381578006/jobs/9765827117)
- [445b381](https://github.com/solo-io/gloo/actions/runs/5383173639/jobs/9769589358)

# Checklist:


test/helpers/scaled_snapshots.go
```diff
@@ -20,6 +20,8 @@ import (
 // contains a builder for each sub-resource type which is responsible for building instances of that resource
 // Additional fields should be added as needed
 type ScaledSnapshotBuilder struct {
+	injectedSnap *gloosnapshot.ApiSnapshot
```
sam-heilbron, line 23: nit: I find this change now makes the snapshot builder do more than it needs to. Previously the builder was a nice mechnaism to product (build) snapshots. Now it also supports passing in a snapshot. To me, this blends the concepts of storing a snapshot with building a snapshot.

A few option I see here:
1. Remove the snap from this builder, and introduce a new Builder that just returns a snapshot that you construct it with. This seems like overkill since the new implementation would do nothing
2. Support this snapshot, but integrate it with the rest of the builder, and so if you use the injectSnap AND other builder functions, they all merge together
3. Remove the snaps from this builder, and update the test to not require a builder,but instead a snapshot

projects/gloo/pkg/translator/performance_linux_test.go
```diff
@@ -165,14 +165,18 @@ var _ = Describe("Translation - Benchmarking Tests", Serial, Label(labels.Perfor
 	)
 })
 
-func generateDesc(apiSnap *v1snap.ApiSnapshot, _ benchmarkConfig, labels ...string) string {
+func generateDesc(b *gloohelpers.ScaledSnapshotBuilder, _ benchmarkConfig, labels ...string) string {
```
sam-heilbron, line 168: Why can't we use the snapshot still? Shouldn't the description depend on the snapshot, and not how the snapshot was build (aka the builder)?

projects/gloo/pkg/translator/performance_linux_test.go
```diff
@@ -152,7 +152,7 @@ var _ = Describe("Translation - Benchmarking Tests", Serial, Label(labels.Perfor
 			Expect(durations).Should(And(config.benchmarkMatchers...))
 		},
 		generateDesc, // generate descriptions for table entries with nil descriptions
-		Entry("basic", basicSnap, basicConfig),
+		Entry("basic", gloohelpers.NewScaledSnapshotBuilder().WithInjectedSnapshot(basicSnap), basicConfig),
```
sam-heilbron, line 155: I mention this in another comment, but it seems like if we want to support this basicSnap, we are left with a few options:
1. Have all tests use a snapshot, and each entry calls builder.Build directly
2. Create a builder that just returns that snapshot, and keep relying on a builder
3. Create a test table that uses snapshots, and anotehr that uses snapshot builders

test/helpers/scaled_snapshots.go
```diff
@@ -54,9 +64,27 @@ func (b *ScaledSnapshotBuilder) WithEndpointBuilder(eb *EndpointBuilder) *Scaled
 	return b
 }
 
+/* Getter funcs to be used by the description generator */
+
+func (b *ScaledSnapshotBuilder) HasInjectedSnapshot() bool {
```
sam-heilbron, line 69: nit: It feels like this exposes too much information about a builder. Ideally a builder should just build something, and whoever consumes that shouldn't care how it happened (aka if there was an injected snapshot). What do you see as the value of some developer knowing if a builder is using an injected snapshot?

projects/gloo/pkg/translator/performance_linux_test.go
bewebi: We use it as the [Table's description generator](https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/translator/performance_linux_test.go#L154), so it has to match the signature of the Table's Entry

sam-heilbron (approved):

test/helpers/scaled_snapshots.go
bewebi: This is for use by `generateDesc()` since the `UpstreamCount()` and `EndpointCount()` getters will not be relevant or accurate for generating a description for a builder with an injected snapshot

changelog/v1.15.0-beta16/performance-tests-type-error.yaml
```diff
@@ -0,0 +1,22 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/solo-projects/issues/5116
+    resolvesIssue: false
+    description: >-
+      Resolve a bug introduced in #8415 due to type mismatch in the translation performance test table's descriptor func
+      skipCI-kube-tests:true
```
sam-heilbron, line 7: The code to process skip ci directives (https://github.com/solo-io/gloo/tree/main/ci/github-actions/process-skip-directives) I believe parses the entire changelog file, so you only need to define the entry once, not per entry.

sam-heilbron (commented):
There are a few other pieces of cleanup that can be done since the tests no longer have the linux requirement:
1. The `BeforeEach` still has 
```
testutils.LinuxOnly("uses go-utils benchmarking.Measure() which only compiles on Linux")
```
which can be removed
2. The comment above the tests: 
```
// These tests only compile and run on Linux machines due to the use of the go-utils benchmarking package which is only
// compatible with Linux
```
Can also be either deleted or updated

sam-heilbron (approved):

test/helpers/benchmark_darwin.go
```diff
@@ -0,0 +1,17 @@
+package helpers
```
nfuden, line 1: Have we considered having local benchmarks moved here? 

Namely the benchmarkMatchers for different scales probably could live here and that way both ci and local dev have a meaningful set of numbers to run against. 
Obviously it does depend on local machine to some degree so I am also fine not adding these but want to at least consider this

nfuden: Per slack too much effort https://solo-io-corp.slack.com/archives/C056AFWPB0X/p1687879605109009?thread_ts=1687798089.840059&cid=C056AFWPB0X

nfuden (approved):

PR #8429, opened 2023-06-29
Stability: Deflake Kube2e Tests

sam-heilbron:
# Description

Deflake kube2e tests, by ensuring that dynamically generated proxies are cleaned up after each test

# Context

## Kube Flakes
We have had recent failures in CI ([logs](https://github.com/solo-io/gloo/actions/runs/5408667683/jobs/9827972747)), where the test failed due to our webhook rejecting a request to update a resource. The failure was:
```
 Message: "Error Validating *v1.Upstream failed: 1 error occurred:\n\t* Validating *v1.Upstream failed: validating *v1.Upstream name:\"gloo-system-testrunner-0-1234\"  namespace:\"gloo-system\": failed to validate Proxy [namespace: gloo-system, name: gateway-proxy] with gloo validation: Listener Error: SSLConfigError. Reason: SSL secret not found: list did not find secret gloo-system.secret-native-ssl\n\n",
```

When I inspected the logs and state of the cluster at failure, I noticed that the `secret-native-ssl` secret did not exist (it was cleaned up by another test), and shouldn't have been in the cluster for the test that did fail. I believe this behavior is due to a race condition in how resources are cleaned up:

1. Child resources are deleted (VS, RouteTable..etc)
2. Gloo picks up those changes and deletes the Proxy if there are no more routes for it
3. A new test applies the new set of resources to work with

There is latency associated with Step 2, so my changes are to perform this action directly. We update the SnapshotWriter to properly cleanup the default proxy generated by each test. I expect this will ensure that each test is guaranteed to start with no Proxies.

## Kube2e Tests Using RELEASED_VERSION
I updated the README to show how kube tests can be run without rebuilding all images, as is the case with this PR, where only test code is modified.

# Checklist:


ashishb-solo:
Thanks for working on this Sam!

So if I'm understanding this correctly, do you think the cause of the flake is that the proxy was supposed to get deleted in step 2, but it wasn't? And when the new test is set up, Gloo is still trying to validate the old proxy but it fails because the certificate was deleted?

sam-heilbron:
> Thanks for working on this Sam!
> 
> So if I'm understanding this correctly, do you think the cause of the flake is that the proxy was supposed to get deleted in step 2, but it wasn't? And when the new test is set up, Gloo is still trying to validate the old proxy but it fails because the certificate was deleted?

That is my current suspicion. As I callout in the PR, this should happen dynamically by Gloo, but because there is a potential delay, I think we can hit a situation where a new test starts up, and the Proxy from the old test is around. This ultimately may not fix the flakes entirely, but I think is better practice to make sure we (test authors) are wiping the resources entirely

test/kube2e/README.md
```diff
@@ -26,7 +26,13 @@ The `GetTestHelper` util method handles installing gloo from either a local or r
 The list of tests to run during CI and nightly builds is provided in `kube-e2e-test-type` matrices in the github workflows. A new test can be added to one or both lists of tests.  
 ## Local Development
 
-### Setup
+### Setup (Previously Released Assets)
+It is possible to run these tests against a previously released version of Gloo Edge. This is useful for testing a release candidate, or a nightly build.
+
+There is no setup required for this option, as the test suite will download the helm chart archive and `glooctl` binary from the specified release. You will use the `RELEASED_VERSION` environment variable when running the tests.
```
ashishb-solo, line 32: could you add a sample value for `RELEASED_VERSION` here?

test/helpers/snapshot_writer.go
```diff
@@ -225,7 +230,15 @@ func (s snapshotWriterImpl) DeleteSnapshot(snapshot *gloosnapshot.ApiSnapshot, d
 
 	// Proxies are auto generated by Gateway resources
 	// Therefore we delete Proxies after we have deleted the resources that may regenerate a Proxy
-	for _, proxy := range snapshot.Proxies {
+	// Proxies are written to the settings.WriteNamespace, which defaults to "gloo-system"
```
ashishb-solo, line 233: what would happen here if someone (possibly by accident) were to add a proxy outside the `gloo-system` namespace?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2232,7 +2232,7 @@ spec:
 				kube2e.UpdateAlwaysAcceptSetting(ctx, false, testHelper.InstallNamespace)
 			})
 
-			Context("with a mix of valid and invalid virtual services", func() {
+			Context("with a mix of valid and invalid virtual services", FlakeAttempts(3), func() {
```
ashishb-solo, line 2235: what is the reasoning behind this change? perhaps a comment to explain why it's necessary?

test/kube2e/README.md
sam-heilbron: I include an example in the testutils.env where I define this value. Would you like me to also include an example in the README? (Happy to, just wanted to clarify)

ashishb-solo: ah, i see. yeah, i think it would still be helpful to add it here, as i didn't see the example in testutils.env on my first pass through

test/helpers/snapshot_writer.go
sam-heilbron: It kind of depends "how" they do it, because there are 2 ways:
1. YOu could write the proxy directly to a custom namespace
2. YOu could configure settings.WriteNamespace to a custom namespace, and then all dynamically generated objects by the Gloo control plane will be written to that namespace

In the first case, you would have needed to add the Proxy to the Snapshot in the first place and SnapshotWriter.Delete would have a reference to it, and we could delete it. This isn't recommended and we warn users about writing proxies directly (preferred to write the objects that customers write)

In the second case, we would need to somehow read the settings.WriteNamespace to identify where to delete from. This is an edge case that none of our tests actually use today, so I opted to keep the impelemtnation simple, and callout in a comment the asssumption that is being made.

In either case, the proxies should be deleted dynamically by the control plane. THis change is an attempt to de-flake a race in tests. If we started using a custom namespace (and this cleanup stopped working), I expect either we would see an increase in flakes (meaning this cleanup is helping) or no change (meaning this cleanup doesn't do what I think it does)

test/kube2e/gateway/gateway_test.go
sam-heilbron: Good call! I had just seen it flake during my PR, so I didn't want to ignore it, ill add a comment

test/kube2e/README.md
sam-heilbron: I ended up linking the PR to where the variable is defined. I'm more than happy to document this better/differently if you have any alternatives

test/helpers/snapshot_writer.go
```diff
@@ -3,25 +3,34 @@ package helpers
...
-var _ SnapshotWriter = new(snapshotWriterImpl)
+var _ SnapshotWriter = new(SnapshotWriterImpl)
 
 type SnapshotWriter interface {
 	WriteSnapshot(snapshot *gloosnapshot.ApiSnapshot, writeOptions clients.WriteOpts) error
 	DeleteSnapshot(snapshot *gloosnapshot.ApiSnapshot, deleteOptions clients.DeleteOpts) error
 }
 
-type snapshotWriterImpl struct {
+type SnapshotWriterImpl struct {
```
jenshu, line 23: any reason we need to export this? looks like it's not used outside of this file

test/helpers/snapshot_writer.go
```diff
@@ -120,16 +139,17 @@ func (s snapshotWriterImpl) doWriteSnapshot(snapshot *gloosnapshot.ApiSnapshot,
 			return writeErr
 		}
 	}
-	for _, proxy := range snapshot.Proxies {
-		if _, writeErr := s.ProxyClient().Write(proxy, writeOptions); !s.isContinuableWriteError(writeErr) {
+	for _, gw := range snapshot.Proxies {
+		// It is recommended to configure Gateway resources (GW, VS, RT, etc) instead of Proxy resources
+		ginkgo.GinkgoWriter.Printf("Proxies are intended to be an opaque resources to users and are not recommended to be written directly in tests")
```
jenshu, line 144: are there any tests are are trying to set snapshot.Proxies currently? should we just disallow it completely, perhaps by returning an error if Proxies is set?

test/helpers/snapshot_writer.go
```diff
@@ -120,16 +139,17 @@ func (s snapshotWriterImpl) doWriteSnapshot(snapshot *gloosnapshot.ApiSnapshot,
 			return writeErr
 		}
 	}
-	for _, proxy := range snapshot.Proxies {
-		if _, writeErr := s.ProxyClient().Write(proxy, writeOptions); !s.isContinuableWriteError(writeErr) {
+	for _, gw := range snapshot.Proxies {
```
jenshu, line 142:
```suggestion
	for _, proxy := range snapshot.Proxies {
```

test/kube2e/util.go
```diff
@@ -174,15 +176,20 @@ func GetSimpleTestRunnerHttpResponse() string {
...
-	if useVersion = os.Getenv("RELEASED_VERSION"); useVersion != "" {
-		if useVersion == "LATEST" {
-			_, current, err := upgrade.GetUpgradeVersions(ctx, repoName)
-			Expect(err).NotTo(HaveOccurred())
-			useVersion = current.String()
-		}
+	releasedVersion := os.Getenv(testutils.ReleasedVersion)
+
+	if releasedVersion == "" {
+		return releasedVersion
```
jenshu, line 182: can you add a comment here re: what will happen if we return "" for `releasedVersion` (i.e. it will use local version)

test/testutils/env.go
```diff
@@ -53,6 +53,11 @@ const (
 
 	// ServiceLogLevel is used to set the log level for the test services. See services/logging.go for more details
 	ServiceLogLevel = "SERVICE_LOG_LEVEL"
+
+	// ReleasedVersion can be used when running KubeE2E tests to have the test suite use a previously released version of Gloo Edge
+	// If set to 'LATEST', the most recently released version will be used
+	// If set to another value, the test suite will use that version (ie '1.15.0-beta1')
+	ReleasedVersion = "RELEASED_VERSION"
```
jenshu, line 60: maybe add a comment here too, for what will happen if the value is empty

test/kube2e/gateway/gateway_test.go
elcasteel: I like the comment you added

test/testutils/env.go
elcasteel, line 60: Is this used anywhere?

sam-heilbron: This is documented in the kube2e test readme. our tests support passing this value in to avoid re-buiding images/charts

test/helpers/snapshot_writer.go
sam-heilbron: The code editor was warning me that I had an exported function, returning an unexported type. So I figured I would just have them be consistent.

test/helpers/snapshot_writer.go
sam-heilbron: I like it! I think in an earlier commit I had done this. I'll revert that

test/testutils/env.go
sam-heilbron: It's also used in https://github.com/solo-io/gloo/blob/main/test/kube2e/util.go

elcasteel: I meant the constant, not the environment variable. It looks like we use a string literal whenever we look it up

elcasteel (approved):

ashishb-solo:
Happy to approve this whenever it's ready, but I can see that there are still a few comments outstanding, so I'll hold off until they get resolved

jenshu (commented):
LGTM

ashishb-solo (approved):

PR #8453, opened 2023-07-07
Performance Test changes for EE

bewebi:
# Description

- Factor out and export `BenchmarkConfig` and `GenerateBenchmarkDesc()` for use in EE
- Replace `benchmarkMatchers` field in `BenchmarkConfig` with `GhaMatchers` and `LocalMatchers`
  - Add member `GetMatchers()` function to `BenchmarkConfig` which returns appropriate matchers depending on run context
- Reduce log level in performance tests 
- Add `WithHealthChecks()` function to `UpstreamBuilder`

# Context

These changes are designed to allow EE to reuse performance test infrastructure developed for OSS

The use of different targets for benchmarking depending on test running environment is inspired by [this implementation](https://github.com/solo-io/solo-projects/blob/main/projects/gloo/pkg/syncer/extauth/benchmark_test.go#L104) in EE
Note that if an entry is missing matchers for the machine tests are being run on, the experiment will still take place and results logged, however no assertions will be made and the entry will trivially pass

See solo-io/solo-projects#5157 for proposed use of these changes in EE

Log level is reduced in order to eliminate noise from each translation iteration logging
[Here](https://github.com/solo-io/gloo/actions/runs/5490524969/jobs/10006077084#step:4:331) is a run of the performance tests without all of the noisy logs
[Here](https://github.com/solo-io/gloo/actions/runs/5521887664/jobs/10070661098) a run of the performance tests as of [50a22e4](https://github.com/solo-io/gloo/pull/8453/commits/50a22e4727ba038ca0b21c62ff87ecdf7e9e788e)

My local environment is much faster than GHA, hence the targets that are 25x faster than the existing targets for GHA
Additionally the existing GHA targets do not account for the lowered log level which also improves performance
These targets are (still) mainly intended for demonstration purposes and are not tied too customer requirements or other external expectations 

# Checklist:


test/helpers/benchmark.go
```diff
@@ -11,3 +18,33 @@ type Result struct {
 	// Time spent in user mode + kernel mode
 	Total time.Duration
 }
+
+// BenchmarkConfig allows configuration for benchmarking tests to be reused for similar cases
+// This struct can be factored out to an accessible location should additional benchmarking suites be added
+type BenchmarkConfig struct {
+	Iterations    int                   // the number of iterations to attempt for a particular entry
+	MaxDur        time.Duration         // the maximum time to spend on a particular entry even if not all iterations are complete
+	LocalMatchers []types.GomegaMatcher // matchers representing the assertions we wish to make for a particular entry when running locally
```
inFocus7, line 27: nit: My only gripe with is having a `Local` matchers, which feels like it leans towards "it works (ran this fast) on my machine". Personally I think having the Github matchers should be enough if it's generally slower, with a comment like the one in `translator/performance_test.go` re. the runner we use for performance tests. Not sure _if_ there is a better practice so I'd also be okay with this, waiting to see others opinion.

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -39,25 +41,28 @@ import (
...
-	maxDur            time.Duration         // the maximum time to spend on a particular entry even if not all iterations are complete
-	benchmarkMatchers []types.GomegaMatcher // matchers representing the assertions we wish to make for a particular entry
-}
-
 // Tests are run as part of the "Nightly" action in a GHA using the default Linux runner
 // More info on that machine can be found here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
 // When developing new tests, users should manually run that action in order to test performance under the same parameters
 // Results can then be found in the logs for that instance of the action
-var _ = Describe("Translation - Benchmarking Tests", decorators.Performance, Label(labels.Performance), func() {
+var _ = Describe("Translation - Benchmarking Tests", Ordered, decorators.Performance, Label(labels.Performance), func() {
```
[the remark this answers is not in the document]
sam-heilbron: Do all performance tests need to be Ordered, or just this one? If it's all, could we bake this into the Peformance decorator, so other devs don't have to remember it?

test/helpers/benchmark.go
```diff
@@ -11,3 +18,33 @@ type Result struct {
...
+	GhaMatchers   []types.GomegaMatcher // matchers representing the assertions we wish to make for a particular entry when running in a GHA
+}
+
+func (bc *BenchmarkConfig) GetMatchers() []types.GomegaMatcher {
+	if os.Getenv("GITHUB_ACTION") != "" {
```
sam-heilbron, line 32: When is this environment label set? Can we follow the existing pattern and define this in the shared file https://github.com/solo-io/gloo/blob/main/test/testutils/env.go, so it's easy to identify?

projects/gloo/pkg/translator/performance_test.go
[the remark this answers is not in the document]
bewebi: The requirement to be Ordered is purely because we're using a `BeforeAll` and `AfterAll`

test/helpers/benchmark.go
bewebi: If anything my intention is sort of the opposite, which is to say that _because_ GHA targets are so much looser they will tend to trivially pass on local machines
Therefore we add the option to set more stringent targets for local use so that running locally can actually serve as a useful indicator, or otherwise don't have any targets at all so there's at the very least no impression that it "passed" locally and is therefore OK

>it might even make sense to have a MatcherProvider or something, which is just a function that returns the matchers you want

The `BenchmarkConfig.GetMatchers()` function does this

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -39,25 +41,28 @@ import (
...
 	var (
 		ctrl       *gomock.Controller
 		settings   *v1.Settings
 		translator Translator
+
+		originalLogLevel zapcore.Level
 	)
 
+	BeforeAll(func() {
```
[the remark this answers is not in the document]
bewebi: This would make more sense than `BeforeEach`/`AfterEach`, I don't really have a problem with this
Happy to go with that if it's considered cleaner than having to add the `Ordered` decorator

test/helpers/benchmark.go
bewebi: This is a [default variable set for all GHAs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

projects/gloo/pkg/translator/performance_test.go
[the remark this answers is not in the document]
sam-heilbron: Just wanted to offer the alternative. The trade-off as I see it is:

BeforeAll/AfterAll: We get to perform the change once, but the cost is extra test-suite level variables
Inside the function that performs translation: The change happens once per test, but the benefit is we don't have extra variables

No strong feelings either way, just wanted to offer the options and am happy which whichever you decide. Leaving this comment open so that you can see the response, but you can resolve this once you're happy

test/helpers/benchmark.go
sam-heilbron: Sounds good! All the more reason, can we define that in a shareed place (env.go), outline when it will be set and then more of our tests can rely on it, should they need github action specific logic

sam-heilbron (approved):

projects/gloo/pkg/translator/performance_test.go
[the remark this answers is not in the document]
bewebi: After (literally) sleeping on it, I decided to set logging within the table body
It feels cleaner since, in theory, additional tests could be added to the Describe which do not require logging to be adjusted and because it locates the logic more close to the piece that requires less verbose logging, which also lends itself more naturally to a descriptive comment

sam-heilbron (dismissed):
Nice work! 🚀

projects/gloo/pkg/translator/performance_test.go
```diff
@@ -56,6 +50,8 @@ var _ = Describe("Translation - Benchmarking Tests", decorators.Performance, Lab
 		ctrl       *gomock.Controller
 		settings   *v1.Settings
 		translator Translator
+
+		originalLogLevel zapcore.Level
```
inFocus7, line 54: nit: do we need this variable after setting it in the describe table? (since we're setting it there and `defer`-ing it)

inFocus7 (commented):
small nit, happy to approve if we don't want to go through CI again

inFocus7 (approved):
🚀

sam-heilbron (approved):

PR #8458, opened 2023-07-11
Gloohelpers measure

bewebi:
# Description

- Add `Measure()` function to gloohelpers with darwin and linux implementations
  - Undo dot-import of ginkgo in `helpers` package which causes conflict on re-declaration of `Measure()`

# Context

We don't always care about the `ignore` value returned by `MeasureIgnore0ns()`
That function returns `true` if the linux-compatible `benchmark.Measure()` returns an error due to recording 0ns as the duration for the given function's runtime
This is relatively common for cases where the function being measured runs very quickly (~100µs) and the measurement may round down
In cases where our expected runtimes are orders of magnitude higher we probably do want to error if we get a 0ns measurement

See [this conversation](https://github.com/solo-io/solo-projects/pull/5142#discussion_r1258931031) about a case where we'd prefer to use `Measure()`

# Checklist:

### DIFF
--- a/changelog/v1.15.0-beta18/gloohelpers-measure.yaml
+++ b/changelog/v1.15.0-beta18/gloohelpers-measure.yaml
@@ -1,0 +1,9 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/solo-projects/issues/5038
+    resolvesIssue: false
+    description: >-
+      Add Measure function to gloohelpers for cases where we do not care about the 0ns error
+
+      skipCI-kube-tests:true
+      skipCI-docs-build:true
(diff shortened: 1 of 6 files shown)


sam-heilbron (commented):
I like it! What do you think about having the "Ignore0" variant, call the original function? That way, those two won't drift overtime. For example:

```
func MeasureIgnore0ns(f func()) (Result, bool, error) {
	result, err := Measure(f)
	return result, false, err
}

func Measure(f func()) (Result, error) {
	before := time.Now()
	f()
	elapsed := time.Since(before)
	return Result{
		Total: elapsed,
	}, nil
}

```

nfuden (dismissed):
Was just in the middle of writing that we shouldhave measureIgnore use measure... and then saw there was a button to refresh and the changes were magically already there!

sam-heilbron (dismissed):

sam-heilbron (approved):

nfuden (approved):

PR #8459, opened 2023-07-12
Expose statsconfig tags 1.14

jenshu:
backport https://github.com/solo-io/gloo/pull/8333


changelog/v1.14.11/gwp-stats-config.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+- type: NON_USER_FACING
```
sam-heilbron, line 2: Should this be Helm since we're adding a new value?

sam-heilbron (approved):
I nit on the changelog, but I see it mirrors the original PR. Thanks for taking care of this!

nfuden (approved):

PR #8468, opened 2023-07-14
Stability: Resolve Envoy Test Service Data Race

sam-heilbron:
# Description

Resolve data race in test service used in e2e tests

# Context

The `envoy.Factory` is responsible for producing an `envoy.Instance` with necessary configuration. Historically, it also tracked each instance that was created, and when the Factory was cleaned up, it would go to each instance it was aware of and call instance.Clean  on that instance. Cleaning an Envoy instance amounts to tearing down the process that is running the service.

However, when an Envoy instance is Started, it is provided a context, and when that context is cancelled the service is auto-cleaned up (https://github.com/solo-io/gloo/blob/main/test/services/envoy/instance.go#L136). This way, we only have to worry about the test context, and all other resource lifecycles are tied to it.

The race occurred because the instances cleans itself up (when the test context is cancelled) AND the factory tries to call clean on the instance when the test suite is complete. Having both of these is actually unnecessary, so I just removed the implementation of the factory cleaning up the instances, since we expect instances to clean themselves up.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5177

### DIFF
--- a/test/services/envoy/factory.go
+++ b/test/services/envoy/factory.go
@@ -51,8 +51,6 @@ type factoryImpl struct {
 	// This can either be a previously released tag or the tag of a locally built image
 	// See the Setup section of the ./test/e2e/README for details about building a local image
 	dockerImage string
-
-	instances []*Instance
 }
 
 // NewFactory returns a new envoy service Factory
@@ -171,11 +169,6 @@ func (f *factoryImpl) Clean() {
 	if f.tmpdir != "" {
 		_ = os.RemoveAll(f.tmpdir)
 	}
-	instances := f.instances
-	f.instances = nil
-	for _, ei := range instances {
-		ei.Clean()
-	}
 }
 
 func (f *factoryImpl) NewInstance() *Instance {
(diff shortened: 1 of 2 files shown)


test/services/envoy/factory.go
```diff
@@ -224,7 +217,6 @@ func (f *factoryImpl) newInstanceOrError() (*Instance, error) {
 	}
 
 	log.Printf("Envoy Instance (%s) Ports: %+v", ei.DockerContainerName, ei.RequestPorts)
-	f.instances = append(f.instances, ei)
```
jbohanon, line 227: Is there still a need for the factory to have an `instances` field if we're not registering them when we create a new instance?

jbohanon (approved):

nfuden:
While I like the general idea it seems correct that the manager would have some way to clean up its instances.
I guess this comes down to whether this is a simple factory that fires and forgets its envoy instances or a manager of them. 

Is that something we are fine with?
if so cool, if not perhaps we just add another signal or some way for it to validate that it's subinstances are completely gone.

nfuden (approved):
Currently our test factories implicitly only manage a building pattern. Ie the factories have no lifecycle management and their responsibility ends once they create an instance that the test can use. Given this it seems this change is good and this also makes instance creation safe to run in parallel and for clean to be called multiple times

Issue #8475, opened 2023-07-18
Helm Chart: Invalid YAML document separation in `gloo-custom-resource-config` when configuring multiple gatewayProxies


jrauschenbusch:
### Gloo Edge Version

1.14.x (latest stable)

### Kubernetes Version

1.25.x

### Describe the bug

When one is configuring multiple gateway proxies in the values.yaml of the Helm Chart then it seems that the document separation is not rendered correctly (at least with Flux HelmController).

Content of gloo-custom-resource-config:

```yaml
apiVersion: v1
data:
  custom-resources: |
    ---
    apiVersion: gateway.solo.io/v1
    kind: Gateway
    metadata:
      name: gateway-proxy
      namespace: gloo
      labels:
        app: gloo
    spec:
      bindAddress: "::"
      bindPort: 8080
      httpGateway:
        options:
          healthCheck:
            path: /healthz
      useProxyProto: false
      ssl: false
      proxyNames:
      - gateway-proxy---
    apiVersion: gateway.solo.io/v1
    kind: Gateway
    metadata:
      name: gateway-proxy-test
      namespace: gloo
      labels:
        app: gloo
    spec:
      bindAddress: "::"
      bindPort: 8080
      httpGateway:
        options:
          healthCheck:
            path: /healthz
      useProxyProto: false
      ssl: false
      proxyNames:
      - gateway-proxy-external
  has-custom-resources: "true"
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: gloo
    meta.helm.sh/release-namespace: gloo
  creationTimestamp: "XXX"
  labels:
    app: gloo
    app.kubernetes.io/managed-by: Helm
    gloo: custom-resources
    helm.toolkit.fluxcd.io/name: gloo
    helm.toolkit.fluxcd.io/namespace: gloo
  name: gloo-custom-resource-config
  namespace: gloo
  resourceVersion: "XXX"
  uid: XXX
```

One can see that the document delimiter is not rendered correctly in the ConfigMap.
```yaml
...
proxyNames:
      - gateway-proxy---
...
```

There was already an issue regarding SSL Gateways: https://github.com/solo-io/gloo/issues/8404

I think the solution for this bug will to replace the following line `{{- "---" }}` with `---`.
https://github.com/solo-io/gloo/blob/2fe0de3841c99a0b6a6cf6793e1f07ad24935518/install/helm/gloo/templates/_8-default-gateways.tpl#L157

[...]

PR #8476, opened 2023-07-18
[1.14] Fix yaml separation for mult. default gw

jrauschenbusch:
# Description

Fixes YAML separation characters for multiple defined default gateways
#8475

Basically the same issue as #8404 but for multiple GW

# Context

When defining multiple Default Gateways with Helm bad line break rendering breaks the gateway CRD creation


jrauschenbusch:
It seems this "bug" was introduced here for some reason:
https://github.com/solo-io/gloo/commit/b7e77c22a5aa547830e4e1114600e1eb8636d1c0#diff-192e16fa5775783d86dcc47cc40061b4913a7efe72286f53f65584128aeaf4a5

jenshu:
@jrauschenbusch thanks for opening this. similar to https://github.com/solo-io/gloo/pull/8406, we should add some testing. i'm planning to rewrite the helm tests so they can catch this sort of bug. would you be open to waiting a few days for that, (and i will also include this fix along it)?

jenshu:
closing in favor of #8493

Issue #7428, opened 2022-11-15
Support TCP Connection Limits


guydc:
### Is your feature request related to a problem? Please describe.

Limiting overall TCP connections and the rate of new TCP connections can protect proxy resources. 

### Describe the solution you'd like

Expose the [Connection Limit Filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/connection_limit_filter) and Network [Local Rate Limit Filter ](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/local_rate_limit_filter).

nfuden:
Pulling from good first issue as we had actually made changes to not translate network filters which makes this a bit harder.

davidjumani:
Adding [HTTP local rate limiting](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter) as part of this

nfuden:
reopening for enterprise  merges and comments indicating the enterprise versions this is merged for

davidjumani:
Added as part of OSS gloo [v1.16.0-beta8](https://github.com/solo-io/gloo/releases/tag/v1.16.0-beta8)
Merged into enterprise and will be released as part of v1.16.0

SantoDE:
Reopened for backort

davidjumani:
Backported to OSS [v1.15.12](https://github.com/solo-io/gloo/releases/tag/v1.15.12)
Will be part of Enterprise v1.16.0-beta2 when released

PR #8487, opened 2023-07-19
feat: Support envoy connection_limit filter

davidjumani:
# Description

Adds the ability for gloo to expose the [Connection Limit](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/connection_limit_filter#config-network-filters-connection-limit) filter

# Context

Part of https://github.com/solo-io/gloo/issues/7428

# Checklist:


projects/gloo/pkg/plugins/connection_limit/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+	_ plugins.NetworkFilterPlugin = new(plugin)
+)
+
+const (
+	ExtensionName = "envoy.extensions.filters.network.connection_limit.v3.ConnectionLimit"
+	StatPrefix    = "connection_limit"
+)
+
+var (
+	pluginStage = plugins.BeforeStage(plugins.AuthNStage)
```
davidjumani, line 25: Setting it as pre auth for now since the aim is to prevent resource drain. Can be updated later

projects/gloo/pkg/plugins/connection_limit/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (p *plugin) Init(params plugins.InitParams) {
+	p.removeUnused = params.Settings.GetGloo().GetRemoveUnusedFilters().GetValue()
+}
+
+func GenerateFilter(connectionLimit *connection_limit.ConnectionLimit) (*envoy_config_listener_v3.Filter, error) {
+	if connectionLimit == nil {
+		return nil, nil
+	}
+	if connectionLimit.GetMaxActiveConnections() == nil {
```
davidjumani, line 48: Prefer having explicit checks for each condition instead of grouping them all together with `||`

projects/gloo/pkg/plugins/plugins.go
```diff
@@ -155,6 +155,11 @@ type HttpFilterPlugin interface {
 	HttpFilters(params Params, listener *v1.HttpListener) ([]StagedHttpFilter, error)
 }
 
+type NetworkFilterPlugin interface {
```
davidjumani, line 158: Have created a `NetworkFilterPlugin` since these plugins should be handled on their own, and not grouped with the HTTP plugins

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -88,6 +92,22 @@ func (n *httpNetworkFilterTranslator) ComputeNetworkFilters(params plugins.Param
 		}
 	}
 
+	// Process the network filters.
```
davidjumani, line 95: Can move this into its own `NetworkFilterTranslator` if needed

projects/gloo/pkg/plugins/tcp/plugin.go
```diff
@@ -80,6 +81,13 @@ func (p *plugin) CreateTcpFilterChains(params plugins.Params, parentListener *v1
 			statPrefix = DefaultTcpStatPrefix
 		}
 
+		// Handle all network filters before proceeding to anything else
+		networkFilters, err := p.generateNetworkFilters(tcpListenerOptions)
```
davidjumani, line 85: Since the entire filter chain is generated in the plugin itself, have to compute the network filters here

changelog/v1.15.0-beta21/connection-limit-filter.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/7428
+  resolvesIssue: false
+  description: >-
+    Adds the ability for gloo to now expose the envoy connection limit filter.
```
sam-heilbron, line 6: These changelogs are published to our user facing docs, and I think having a little more clarity around how/wehere this is configured would help. Perhaps include a reference to the resource(s) that can configure this and what value this filter would provide?

projects/gloo/api/v1/options.proto
```diff
@@ -181,6 +182,8 @@ message HttpListenerOptions {
 
     dfp.options.gloo.solo.io.FilterConfig dynamic_forward_proxy = 28;
 
+    connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 29;
```
sam-heilbron, line 185: I like how you included a description of this message on the definition itself. The full set of options are published here: https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk/#httplisteneroptions and the comment above this line is associated with the "description" section. I think its more likely people will find this options page rather than the new file wit hthe Connectionlimit definition itself. I think we can just use the same comment you had, but could we add something to describe this field?

projects/gloo/api/v1/options.proto
```diff
@@ -189,6 +192,8 @@ message HttpListenerOptions {
 // Optional, feature-specific configuration that lives on tcp listeners
 message TcpListenerOptions {
     tcp.options.gloo.solo.io.TcpProxySettings tcp_proxy_settings = 3;
+
+    connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 4;
```
sam-heilbron, line 196: nit: same as above, let's include a description of this field

projects/gloo/pkg/plugins/connection_limit/plugin.go
sam-heilbron, line 48: Theoretically these two could be grouped together, since the second check would catch the first. The 'Getter' methods have nil checks built into them

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
```diff
@@ -0,0 +1,55 @@
...
+					Name: ExtensionName,
+					ConfigType: &envoy_config_listener_v3.Filter_TypedConfig{
+						TypedConfig: typedConfig,
+					},
+				},
+				Stage: pluginStage,
+			},
+		}))
+
+		By("Ensures that max connections must be greater than or equal to 1")
```
sam-heilbron, line 45: nit: I think this could be a separate `It` block for readability

projects/gloo/pkg/plugins/plugins.go
```diff
@@ -155,6 +155,11 @@ type HttpFilterPlugin interface {
...
+	Plugin
+	NetworkFilters(params Params, listener *v1.HttpListener) ([]StagedNetworkFilter, error)
```
sam-heilbron, line 160: This is an interesting one. The NetworkFilter interface is intentionally generic, but we configure it with an HttpListener, meaning this wil only be used by Http filter chains. Is that intentional?

projects/gloo/pkg/plugins/registry/registry.go
```diff
@@ -78,6 +79,7 @@ func Plugins(opts bootstrap.Opts) []plugins.Plugin {
 		tls_inspector.NewPlugin(),
 		pipe.NewPlugin(),
 		tcp.NewPlugin(utils.NewSslConfigTranslator()),
+		connection_limit.NewPlugin(),
```
sam-heilbron, line 82: Perhaps worth pulling in @SantoDE , is this something we want to configure in Open Source or Enterprise?

projects/gloo/pkg/translator/listener_subsystem.go
```diff
@@ -77,6 +77,7 @@ func (l *ListenerSubsystemTranslatorFactory) GetHttpListenerTranslators(ctx cont
 		listener,
 		listener.GetHttpListener(),
 		httpListenerReport,
+		l.pluginRegistry.GetNetworkFilterPlugins(),
```
sam-heilbron, line 80: I like this. The alternative, would be to just pass the pluginRegistry, instead of each set of plugins, and then call the appropriate method when we need certain plugins. The registry is pretty simple and just is a way to supply plugins

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -42,24 +44,46 @@ func NewHttpListenerNetworkFilterTranslator(
...
 			report:          report,
-			plugins:         plugins,
+			httpPlugins:     httpPlugins,
 			hcmPlugins:      hcmPlugins,
 			routeConfigName: routeConfigName,
 		},
 	}
 }
 
+func (n *httpNetworkFilterTranslator) computePreHTTPNetworkFilters(params plugins.Params) []plugins.StagedNetworkFilter {
```
sam-heilbron, line 67: nit: I find this name a little misleading. This function will compute staged network filters. The fact that they are "pre-hcm" more has to do with the stage of the plugins themselves

sam-heilbron (commented):
Looking great! A few small thoughts/nits

projects/gloo/api/v1/options/connection_limit/connection_limit.proto
```diff
@@ -0,0 +1,24 @@
...
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+
+
+// This plugin provides configuration options to limit active connections to envoy
+// ConnectionLimit can be specified on HTTPGateways and TCPGateways
+message ConnectionLimit {
+    // The maximum number of active connections for this gateway. Other connection attemps will be dropped.
+    google.protobuf.UInt64Value max_active_connections = 1;
+    // The time to wait before a connection is dropped.
```
nfuden, line 22: https://github.com/envoyproxy/envoy/blob/9a72a532fea0d99d654bf4e0216d88c479bd5cca/api/envoy/extensions/filters/network/connection_limit/v3/connection_limit.proto#L33

perhaps we mimick this and note that if unset then the default is no delay?

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
```diff
@@ -0,0 +1,55 @@
...
+		_, err = NewPlugin().NetworkFilters(plugins.Params{}, &v1.HttpListener{
```
nfuden, line 46: For this it block perhaps we calso check for an empty struct  as another case
ConnectionLimit: &connection_limit.ConnectionLimit{},

projects/gloo/pkg/plugins/plugins.go
nfuden: probably should be on the generic *v1.Listener, right?

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -42,24 +44,46 @@ func NewHttpListenerNetworkFilterTranslator(
...
+	// Process the network filters.
+	for _, plug := range n.networkPlugins {
+		stagedFilters, err := plug.NetworkFilters(params, n.listener)
+		if err != nil {
+			validation.AppendHTTPListenerError(n.report, validationapi.HttpListenerReport_Error_ProcessingError, err.Error())
+		}
+
+		for _, nf := range stagedFilters {
+			if nf.NetworkFilter == nil {
+				log.Warnf("plugin implements NetworkFilters() but returned nil")
```
nfuden, line 78: Can we call out that this is an implementation error not operator error and perhaps also add the name of the plugin?

plug.Name()

changelog/v1.15.0-beta21/connection-limit-filter.yaml
davidjumani: Done!

projects/gloo/api/v1/options.proto
davidjumani: Done!

projects/gloo/api/v1/options.proto
davidjumani: Done!

projects/gloo/api/v1/options/connection_limit/connection_limit.proto
```diff
@@ -0,0 +1,24 @@
...
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+
+
+// This plugin provides configuration options to limit active connections to envoy
+// ConnectionLimit can be specified on HTTPGateways and TCPGateways
+message ConnectionLimit {
+    // The maximum number of active connections for this gateway. Other connection attemps will be dropped.
+    google.protobuf.UInt64Value max_active_connections = 1;
```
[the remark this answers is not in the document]
davidjumani: Done!

projects/gloo/pkg/plugins/connection_limit/plugin.go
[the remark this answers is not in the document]
davidjumani: Done! However this might change so leaving this open for now

projects/gloo/pkg/plugins/connection_limit/plugin.go
```diff
@@ -0,0 +1,83 @@
...
+	}, nil
+
+}
+
+func (p *plugin) NetworkFilters(params plugins.Params, listener *v1.HttpListener) ([]plugins.StagedNetworkFilter, error) {
+	connectionLimitFilter, err := GenerateFilter(listener.GetOptions().GetConnectionLimit())
+	if err != nil {
+		return nil, err
+	}
+	return []plugins.StagedNetworkFilter{
```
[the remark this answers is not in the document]
davidjumani: Looking at the other plugins, it is used to determine if the HTTP plugin should be used based on whether any VirtualService or Route has it configured. Since this is only based off the gateway, it is pretty much useless and have removed it.

projects/gloo/pkg/plugins/plugins.go
davidjumani: Moving it to the generic `ListenerOptions` will force this to be used on all children gateways specified as part of the gateway. 
The `tcpFilterChainTranslator` has each plugin create its own chain. As the connection limit filter is on a per-filter chain level, it must be processed within the `tcp` plugin.
Since the `httpNetworkFilterTranslator` is the only one to process the filters individually to create a chain, I've written it this way to utilize the HttpListener.

projects/gloo/pkg/translator/listener_subsystem.go
davidjumani: I prefer it this way as well. This way we can restrict what can be passed to the translators

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
davidjumani: Makes sense

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -42,24 +44,46 @@ func NewHttpListenerNetworkFilterTranslator(
...
 		},
 	}
 }
 
+func (n *httpNetworkFilterTranslator) computePreHTTPNetworkFilters(params plugins.Params) []plugins.StagedNetworkFilter {
+	var networkFilters []plugins.StagedNetworkFilter
+	// Process the network filters.
+	for _, plug := range n.networkPlugins {
+		stagedFilters, err := plug.NetworkFilters(params, n.listener)
+		if err != nil {
```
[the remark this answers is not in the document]
davidjumani: Thanks for the heads up!

projects/gloo/pkg/translator/network_filters.go
davidjumani: True but I'd like to make it obvious that they have to be processed before the HCM

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
davidjumani: Added another test case

projects/gloo/pkg/translator/network_filters.go
davidjumani: Could you please explain what you mean? I've followed what is done for the HTTPFilters
https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/translator/network_filters.go#L194

nfuden: fair maybe we should change that as well. Our plugin interface asserts that the plugin has a function called Name and minimally using that would make the error more descriptive. 
We can leave the rest of message updates out but I was just thinking how we can indicate to customers that its something that means they need to triage an exact plugin or contact our support as we have a programming error

projects/gloo/pkg/translator/network_filters.go
sam-heilbron: ah I didn't reallize that we want to process them before the HCM. There are 2 orders that are important: the order we process the config, the order we send them to envoy (to process the requests). Are you saying this code is intentionally translated in the control plane before the HCM or that the code should be ordered before the HCM in the envoy config?

sam-heilbron (commented):
This is awesome! I can go through the convos shortly but this is looking really good. I do think it would be valuable for us to add some e2e tests to smoke test this functionality. I could see benefits from the following tests:
- Assert that requests are rejected by connection limit filter before hitting HCM 

We have https://github.com/solo-io/gloo/blob/main/test/e2e/example_test.go which is intended to provide an example for how to write e2e tests. We could add a new file for this feature in particular, and add any other cases that you think would be good to test against

davidjumani:
Thanks @sam-heilbron I was waiting on the pre / post auth to be finalized before writing the e2e tests. I'll get to writing them now

projects/gloo/pkg/plugins/plugins.go
davidjumani: Update: To make things simpler, I've added two methods, `NetworkFiltersHTTP` and `NetworkFiltersTCP` to be processed by the HTTP and TCP gateways

projects/gloo/pkg/translator/network_filters.go
davidjumani: The order in envoy should be before the HCM. For consistency, I'm processing them before the HCM filters as well

davidjumani: If my understanding of envoy is correct, the network filters need to be ordered before the HCM as shown [here](https://www.envoyproxy.io/docs/envoy/latest/intro/life_of_a_request#network-filter-chain-processing)
Hence I'm processing them before the HCM as well as ordering them in the network filter list before the HCM filter

test/e2e/connection_limit_test.go
```diff
@@ -0,0 +1,138 @@
...
+				},
+			}
+
+			testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+				gw,
+			}
+		})
+
+		It("Should drop connections after limit is reached", func() {
+			injectFaultDelay(testContext)
```
ben-taussig-solo, line 87: Is there any reason we can't initialize the default VirtualService with the fault delay in the`BeforeEach` block, similar to how we initialize the gateway? e.g. something similar to what we do here: https://github.com/solo-io/gloo/blob/eadaa632e13a2ce0b75a7d486d8bf30ec46b90d6/test/e2e/cors_test.go#L59-L75

ben-taussig-solo: The same question applies for the "Filter not defined" context above as well

test/e2e/connection_limit_test.go
```diff
@@ -0,0 +1,138 @@
...
+				defer GinkgoRecover()
+				defer wg.Done()
+				_, err := httpClient.Do(requestBuilder.Build())
+				Expect(err).Should(MatchError(ContainSubstring("EOF")), "The connection should close")
+			}
+
+			wg.Add(2)
+
+			go expectSuccess()
+			time.Sleep(1 * time.Second)
```
ben-taussig-solo, line 110: Why do we need to sleep here?

test/e2e/connection_limit_test.go
davidjumani: It should work. I just went with https://github.com/solo-io/gloo/blob/main/test/e2e/fault_injection_test.go#L72
Let me know if it is better in the `BeforeEach` block

test/e2e/connection_limit_test.go
davidjumani: Since we're sending requests concurrently to test the limits on active connections, it is sometimes flaky and the second request gets served first. That's why I'm adding a delay between the first and second one

ben-taussig-solo: makes sense! could you add a comment clarifying this?

test/e2e/connection_limit_test.go
ben-taussig-solo: I don't think it really matters for the purposes of stability, but I am under the impression that we're trying to use the `ResourcesToCreate()` paradigm in new tests moving forward.

davidjumani: Changed it to match the current paradigm. Do you think it would be good to make the changes in other tests as well?

test/e2e/connection_limit_test.go
davidjumani: Added the comment, thanks!

test/e2e/connection_limit_test.go
ben-taussig-solo: I think modifying other test suites is probably outside the scope of this PR

projects/gloo/pkg/plugins/registry/registry.go
SantoDE: Well, definitely Enterprise since it's for SAP. I don't have strong opinion on OSS. I guess typically we add "simple exposes" to OSS as well? @sam-heilbron

nfuden (approved):

test/e2e/connection_limit_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/solo-io/gloo/test/e2e"
+	"github.com/solo-io/gloo/test/testutils"
+	"github.com/solo-io/solo-kit/pkg/utils/prototime"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = FDescribe("Connection Limit", func() {
```
sam-heilbron, line 21: This test is focused. Can we unfocus it? Since we run our CI with command line filters (labels) this is ignored (https://onsi.github.io/ginkgo/#combining-filters) and doesn't cause an issue, but could lead to confusion

test/e2e/connection_limit_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+		It("Should not drop any connections", func() {
+			var wg sync.WaitGroup
+			httpClient := testutils.DefaultClientBuilder().WithTimeout(time.Second * 10).Build()
+			requestBuilder := testContext.GetHttpRequestBuilder()
+
+			expectSuccess := func() {
+				defer GinkgoRecover()
+				defer wg.Done()
+				_, err := httpClient.Do(requestBuilder.Build())
```
sam-heilbron, line 58: nit: I think having explicit tests when possible are better. Any reason not to assert a 200 response here?

sam-heilbron (commented):
Nice work, looks great! I had one small comment about a focused test and then a nit that can be ignored.

One other thing to consider: This feature (new network filter) overlaps with the work for DCP in that they both touch areas of translation and affect filter chain logic. Do you think it's worth it to expand the DCP tests (enterprise) that we have to support a simple use case to confirm that still works as expected when this is configured?

test/e2e/connection_limit_test.go
davidjumani: Good catch, fixed!

test/e2e/connection_limit_test.go
davidjumani: It doesn't matter since we're only checking to see if the connection is dropped, but I like the idea of them being explicit

sam-heilbron (approved):

davidjumani:
> One other thing to consider: This feature (new network filter) overlaps with the work for DCP in that they both touch areas of translation and affect filter chain logic. Do you think it's worth it to expand the DCP tests (enterprise) that we have to support a simple use case to confirm that still works as expected when this is configured?

Good call! I'll manually check it first to see if there are any issues and if so, fix them and write a test.

sam-heilbron (dismissed):
🚀

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
```diff
@@ -0,0 +1,66 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/connection_limit"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins"
+	"github.com/solo-io/gloo/projects/gloo/pkg/utils"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+var _ = Describe("Plugin", func() {
```
inFocus7, line 16: nit: More descriptive test description (ex. "Connection Limit Plugin")

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	It("Ensures that max connections must be greater than or equal to 1", func() {
+		_, err := NewPlugin().NetworkFiltersTCP(plugins.Params{}, &v1.TcpListener{
+			Options: &v1.TcpListenerOptions{
+				ConnectionLimit: &connection_limit.ConnectionLimit{
+					MaxActiveConnections: &wrappers.UInt64Value{Value: 0},
+				},
+			},
+		})
+		Expect(err).To(HaveOccurred())
```
inFocus7, line 54: nit: Explicit matching would be safest/true-r
```
Expect(err).To(MatchError("MaxActiveConnections must be greater than or equal to 1. Current value : %v"))
```
something like the above

test/e2e/connection_limit_test.go
```diff
@@ -0,0 +1,146 @@
...
+		WithRouteActionToUpstream("route", testContext.TestUpstream().Upstream).
+		WithRouteOptions("route", &gloov1.RouteOptions{
+			Faults: &fault.RouteFaults{
+				Delay: &fault.RouteDelay{
+					FixedDelay: prototime.DurationToProto(time.Second * 1),
+					Percentage: float32(100),
+				},
+			},
+		}).
+		Build()
```
inFocus7, line 142: Nit: I think we can simplify this to something like

```
defaultVs := testContext.ResourcesToCreate().VirtualServices[0]
vs := <iforgetwhichlibrary>.NewBuilderFromVirtualService(defaultVs).
    WithRouteOptions(e2e.DefaultRouteName, &gloov1.RouteOptions{
		Faults: &fault.RouteFaults{
			Delay: &fault.RouteDelay{
				FixedDelay: prototime.DurationToProto(time.Second * 1),
				Percentage: float32(100),
			},
		},
	}).Build()
```

Since our default vs has most of that setup done already.

inFocus7 (commented):
A few nits, but can approve if we want to get this out sooner 👍🏼

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
davidjumani: From the code, all the plugins are this way. If needed I can raise a PR to fix this

projects/gloo/pkg/plugins/connection_limit/plugin_test.go
davidjumani: Makes sense. Fixed

test/e2e/connection_limit_test.go
davidjumani: I tried this but the default service is a direct response and route faults will not work with it. Also while creating it from the default service, the default domain is `*` and a builder will not let this happen as it can potentially cause flakes (https://github.com/solo-io/gloo/blob/main/test/helpers/virtual_services.go#L219)

nfuden (approved):

sam-heilbron (approved):

inFocus7 (approved):
🚀

jbohanon:
/kick merge commit should not have caused failure seen https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=182afbcd-b3f0-451f-8661-8687fc417729

PR #8493, opened 2023-07-21
helm test refactor and gw yaml fix

jenshu:
Refactored the helm unit tests so that they actually parse and validate the Gateway yaml instead of just looking for snippets.

Also includes the multi-gateway typo fix from https://github.com/solo-io/gloo/pull/8406 and https://github.com/solo-io/gloo/pull/8476


install/test/helm_test.go
```diff
@@ -2223,23 +2212,21 @@ spec:
   useProxyProto: true
   ssl: true
   proxyNames:
-  - test-name`}
-						for _, gwYaml := range gwYamls {
-							expectCustomResourceSubstring(gwYaml, true)
+  - test-name
+`),
 						}
+						customResources := getCustomResourceManifest(4)
```
jenshu, line 2218: before the helm template fix, this test was previously failing as it was only finding 3 gateways (due to the `---` not being on its own line). after the template fix, this and other similar tests pass.

install/test/helm_test.go
```diff
@@ -2386,7 +2409,7 @@ spec:
 					BeforeEach(func() {
 						prepareMakefile(namespace, helmValues{
 							valuesArgs: []string{
-								"gatewayProxies.gatewayProxy.disabled=true",
+								"gatewayProxies.gatewayProxy.disabled=false",
```
sam-heilbron, line 2412: Why this chage? The test name seems to indicate that the default proxy is disabled

install/test/helm_test.go
```diff
@@ -98,15 +97,19 @@ var _ = Describe("Helm Test", func() {
...
+		getCustomResourceManifest := func(expectedResources int) TestManifest {
 			configMap := getConfigMap(testManifest, namespace, "gloo-custom-resource-config")
-			Expect(configMap.Data).NotTo(BeNil())
-			if contains {
-				Expect(configMap.Data["custom-resources"]).To(ContainSubstring(substr))
-			} else {
-				Expect(configMap.Data["custom-resources"]).NotTo(ContainSubstring(substr))
+			ExpectWithOffset(1, configMap.Data).NotTo(BeNil())
+			customResourceYaml := configMap.Data["custom-resources"]
+			if expectedResources > 0 {
```
sam-heilbron, line 105: nit: What does this assertion provide us? We then assert the number of resources match the expectedResources, and we always create a TestManifest with the yaml, even when the expectedResources is 0

install/test/helm_test.go
jenshu, line 2412: not sure why i changed this to false, let me double check

install/test/helm_test.go
```diff
@@ -98,15 +97,19 @@ var _ = Describe("Helm Test", func() {
 			statsAnnotations map[string]string
 		)
 
-		expectCustomResourceSubstring := func(substr string, contains bool) {
+		// create a TestManifest out of the custom resource yaml from the configmap
+		getCustomResourceManifest := func(expectedResources int) TestManifest {
```
sam-heilbron, line 101: I like this, and I'm happy to keep this as is. It seems like this is used to create the TestManifest and assert the number of resources, and then each test performs assertions on the state of that test manifest. One alternate way we could accomplish this is with:

``` go
func assertCustomResourceManifest(matchers map[string]GomegaMatcher) {
  configMap := getConfigMap(testManifest, namespace, "gloo-custom-resource-config")
  ExpectWithOffset(1, configMap.Data).NotTo(BeNil())
  customResourceYaml := configMap.Data["custom-resources"]
  customResourceManifest := NewTestManifestFromYaml(customResourceYaml)
  ExpectWithOffset(1, manifest.NumResources()).To(HaveLen(matchers))
  
  for matcherKey, matcher := range matchers {
     customResources.ExpectUnstructured("Gateway", namespace, matcherKey).To(matcher)
  }
}

```

In my view this consolidates all assertion logic to a single place instead of spliiting it between the test and the custom utility

install/test/helm_test.go
jenshu: yeah i accidentally changed it when testing. reverted it in 38216cb

install/test/helm_test.go
jenshu: i had added that there so we could exit early before creating the TestManifest (which involves parsing the yaml and creating the Unstructured resources). although i suppose it would only exit early if the manifest was empty in which case there isn't much to parse. i can remove that line

install/test/helm_test.go
jenshu: i like that, will try it

jenshu: updated in 6a4d441

sam-heilbron (dismissed):
Awesome!

install/test/helm_test.go
```diff
@@ -6662,3 +6682,30 @@ func getStructuredDeployment(t TestManifest, resourceName string) *appsv1.Deploy
 
 	return structuredDeployment
 }
+
+func makeUnstructuredGateway(namespace string, name string, ssl bool) *unstructured.Unstructured {
+	port := "8080"
+	gwName := name
+	if ssl {
+		port = "8443"
+		gwName = name + "-ssl"
```
bewebi, line 6691: I think it would be cleaner to either always use the passed name (and expect the caller to give a unique name if it's SSL) or use a function to generate an "ssl name" from a passed name, so that can be used throughout the suite

My preference is the former, we can add a `GatewayProxyNameSsl` const [here](https://github.com/solo-io/gloo/blob/f9c8c422f96e7e20b02e6d8a342df2b2b877d259/projects/gateway/pkg/defaults/gateway.go#L15) and reference that wherever it's needed

install/test/helm_test.go
```diff
@@ -98,15 +96,21 @@ var _ = Describe("Helm Test", func() {
 			statsAnnotations map[string]string
 		)
 
-		expectCustomResourceSubstring := func(substr string, contains bool) {
+		// Create a TestManifest out of the custom resource yaml from the configmap,
+		// and then perform assertions on the resources (which are assumed to all be Gateways).
+		// `matchers` is a map of Gateway name to the assertion to be made on that Gateway.
+		// There must be a matcher for every Gateway that's expected to be in the manifest.
+		assertCustomResourceManifest := func(matchers map[string]types.GomegaMatcher) {
 			configMap := getConfigMap(testManifest, namespace, "gloo-custom-resource-config")
```
bewebi, line 104: Might be worth a comment as to [where this hard-coded string comes from](https://github.com/solo-io/gloo/blob/b7e77c22a5aa547830e4e1114600e1eb8636d1c0/install/helm/gloo/templates/5-resource-configmap.yaml#L4), or even factoring it out (as it's used a few times in the file)

install/test/helm_test.go
```diff
@@ -98,15 +96,21 @@ var _ = Describe("Helm Test", func() {
 			statsAnnotations map[string]string
 		)
 
-		expectCustomResourceSubstring := func(substr string, contains bool) {
+		// Create a TestManifest out of the custom resource yaml from the configmap,
+		// and then perform assertions on the resources (which are assumed to all be Gateways).
+		// `matchers` is a map of Gateway name to the assertion to be made on that Gateway.
+		// There must be a matcher for every Gateway that's expected to be in the manifest.
```
bewebi, line 102: From reading the comment it seems like the test should fail if there is an unexpected gateway in the manifest, but it looks like the code will simply not make an assertion on any gateway for which there is not matcher

I'm having issues running the tests locally which are unrelated, so I'm unable to explore this independently, but assuming this is the case we should either clarify the comment or modify the code to reflect the intended behavior

install/test/helm_test.go
```diff
@@ -1745,7 +1715,8 @@ metadata:
 						name := defaults.GatewayProxyName
 						bindPort := "8080"
 						ssl := "false"
```
bewebi, line 1717: Since we set these variables up top in order to insert them into an otherwise standard hybrid gateway definition, can we definite a function that takes these arguments and outputs the expected hybrid gateway YAML?

This would prevent repetition in `It("sets https hybrid gateway"...`

Otherwise we can refrain from instantiating these vars and pass the hybrid gateway YAML as a single string without concatenation

bewebi: This sentiment is similar to the ALS and Tracing tests below, among others, where we already follow the second paradigm as of this PR

~If we choose to go the first route for hybrid gateways, we probably should for the ALS and Tracing gateways as well~

Given that, we should have the hybrid tests match

Adding (or modifying) functions to generate gateway YAMLs is too large an undertaking for something that's out of scope for this PR

bewebi (commented):
Some comments, mostly nitty and around use of hard-coded/concatenated strings

install/test/helm_test.go
jenshu: below it checks that the number of resources found in the manifest matches the number of matchers you passed in:
`ExpectWithOffset(1, customResourceManifest.NumResources()).To(Equal(len(matchers)))`
so we shouldn't have a case where there is an extra gateway without an associated matcher.
i can add a comment on that line so it's more apparent

install/test/helm_test.go
jenshu: i removed a few instances of these variables (`bindPort := "8080"`...) in favor of putting everything hardcoded into the yaml string, but it looks like i missed a few spots. will update these as well

install/test/helm_test.go
bewebi: Ah, yup, I missed that
As long as it does what the comment says I think it's good as-is, I was just unable to verify myself

install/test/helm_test.go
jenshu: when you create an ssl gateway, our helm template just [appends -ssl](https://github.com/solo-io/gloo/blob/b1d66340f871762fc34c4b1693f63e6c77f563c7/install/helm/gloo/templates/_8-default-gateways.tpl#L66) to the given name. For the proxy name it uses the [original name without ssl](https://github.com/solo-io/gloo/blob/b1d66340f871762fc34c4b1693f63e6c77f563c7/install/helm/gloo/templates/_8-default-gateways.tpl#L107) so we do need the non-ssl name to be passed in. i can move the -ssl append into a function though

jenshu: made a function that appends the "-ssl" 9367bc0

there are still a lot of places where we're using "gateway-proxy" or "gateway-proxy-ssl" hardcoded (instead of the defined const) but i'm inclined not to refactor those since there are ~250 instances of those strings in this file

install/test/helm_test.go
jenshu: added const with comment 0d333c8

install/test/helm_test.go
jenshu: updated in 0d333c8

install/test/helm_test.go
bewebi: Fair enough
That said, if we ever were going to make those changes, now seems like a reasonable time...since this is a refactor PR doing a bulk cleanup is not too significant a stretch of the scope (compared to if we were making actual helm changes, say)

sam-heilbron (approved):
🚀 nice work!

PR #8505, opened 2023-07-24
Enforce matching secret and upstream namespaces when secrets are added to headers

elcasteel:
# Description
Optionally check whether headerSecretRef and the upstream that the secret is sent to have matching namespaces. 

This is configured with an environment variable called `HEADER_SECRET_REF_NS_MATCHES_US` and can be set at install time with the value `gloo.headerSecretRefNsMatchesUs`

We use this feature both for configuring health checks and in the headers plugin so the check is added both places.

# Context
We send the plain text of arbitrary secrets to upstreams via HeaderManipulation. In a multi-tenant scenario a malicious tenant can steal a the text of a secret from another tenant by configuring an upstream/route/header manipulation to send a secret from a namespace they should not be able to access to an upstream they own. 

# Testing
This is tested via unit tests that expect either errors or no errors when configuring health checks and running the headers plugin. 

# Checklist:

### DIFF
--- a/changelog/v1.15.0-beta23/match-secret-refs.yaml
+++ b/changelog/v1.15.0-beta23/match-secret-refs.yaml
@@ -1,7 +1,8 @@
 changelog:
   - type: NEW_FEATURE
-    issueLink: https://github.com/solo-io/solo-projects/issues/5007#event-9581124373
+    issueLink: https://github.com/solo-io/solo-projects/issues/5007
     resolvesIssue: false
     description: > 
          Add an environment variable "HEADER_SECRET_REF_NS_MATCHES_US" that when set in the gloo pod requires that secrets
-         sent in headers to an upstream are in the same namespace as that upstream.
+         sent in headers to an upstream are in the same namespace as that upstream. This defaults to false and when it is not set,
+         resources can be in any namespaces watched by Gloo.
--- a/changelog/v1.15.0-beta23/match-secret-refs.yaml
+++ b/changelog/v1.15.0-beta23/match-secret-refs.yaml
@@ -1,0 +1,7 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/5007#event-9581124373
+    resolvesIssue: false
+    description: > 
+         Add an environment variable "HEADER_SECRET_REF_NS_MATCHES_US" that when set in the gloo pod requires that secrets
+         sent in headers to an upstream are in the same namespace as that upstream.
--- a/pkg/utils/api_conversion/type.go
+++ b/pkg/utils/api_conversion/type.go
@@ -9,6 +9,13 @@ import (
 	"github.com/solo-io/solo-kit/pkg/errors"
 )
 
+type HeaderSecretOptions struct {
+	EnforceNamespaceMatch bool
+	UpstreamNamespace     string
+}
+
+const MatchingNamespaceEnv = "HEADER_SECRET_REF_NS_MATCHES_US"
+
 // Converts between Envoy and Gloo/solokit versions of envoy protos
 // This is required because go-control-plane dropped gogoproto in favor of goproto
 // in v0.9.0, but solokit depends on gogoproto (and the generated deep equals it creates).
@@ -46,12 +53,12 @@ func ToEnvoyInt64Range(int64Range *envoytype_gloo.Int64Range) *envoy_type_v3.Int
 	}
 }
 
-func ToEnvoyHeaderValueOptionList(option []*envoycore_sk.HeaderValueOption, secrets *v1.SecretList) ([]*envoy_config_core_v3.HeaderValueOption, error) {
+func ToEnvoyHeaderValueOptionList(option []*envoycore_sk.HeaderValueOption, secrets *v1.SecretList, secretOptions HeaderSecretOptions) ([]*envoy_config_core_v3.HeaderValueOption, error) {
 	result := make([]*envoy_config_core_v3.HeaderValueOption, 0)
 	var err error
 	var opts []*envoy_config_core_v3.HeaderValueOption
 	for _, v := range option {
-		opts, err = ToEnvoyHeaderValueOptions(v, secrets)
+		opts, err = ToEnvoyHeaderValueOptions(v, secrets, secretOptions)
 		if err != nil {
 			return nil, err
 		}
@@ -60,7 +67,7 @@ func ToEnvoyHeaderValueOptionList(option []*envoycore_sk.HeaderValueOption, secr
 	return result, nil
 }
 
-func ToEnvoyHeaderValueOptions(option *envoycore_sk.HeaderValueOption, secrets *v1.SecretList) ([]*envoy_config_core_v3.HeaderValueOption, error) {
+func ToEnvoyHeaderValueOptions(option *envoycore_sk.HeaderValueOption, secrets *v1.SecretList, secretOptions HeaderSecretOptions) ([]*envoy_config_core_v3.HeaderValueOption, error) {
 	switch typedOption := option.GetHeaderOption().(type) {
 	case *envoycore_sk.HeaderValueOption_Header:
 		return []*envoy_config_core_v3.HeaderValueOption{
@@ -73,6 +80,10 @@ func ToEnvoyHeaderValueOptions(option *envoycore_sk.HeaderValueOption, secrets *
 			},
 		}, nil
 	case *envoycore_sk.HeaderValueOption_HeaderSecretRef:
+		if secretOptions.EnforceNamespaceMatch && secretOptions.UpstreamNamespace != typedOption.HeaderSecretRef.GetNamespace() {
+			// The secrets sent to the upstream must come from the same namespace. The error message is copied from secretList.Find() to avoid exposing new information about this setting.
+			return nil, errors.Errorf("list did not find secret %v.%v", typedOption.HeaderSecretRef.GetNamespace(), typedOption.HeaderSecretRef.GetName())
+		}
 		secret, err := secrets.Find(typedOption.HeaderSecretRef.GetNamespace(), typedOption.HeaderSecretRef.GetName())
 		if err != nil {
 			return nil, err
--- a/pkg/utils/api_conversion/type.go
+++ b/pkg/utils/api_conversion/type.go
@@ -9,9 +9,11 @@ import (
 	"github.com/solo-io/solo-kit/pkg/errors"
 )
 
+// HeaderSecretOptions is used to pass around information about whether/how to enforce that secrets and upstream namespace must match
 type HeaderSecretOptions struct {
 	EnforceNamespaceMatch bool
-	UpstreamNamespace     string
+	// this will be ignored unless EnforceNamespaceMatch is true
+	UpstreamNamespace string
 }
 
 const MatchingNamespaceEnv = "HEADER_SECRET_REF_NS_MATCHES_US"
--- a/projects/gloo/pkg/plugins/headers/plugin.go
+++ b/projects/gloo/pkg/plugins/headers/plugin.go
@@ -1,7 +1,11 @@
 package headers
 
 import (
+	"os"
 	"strings"
+
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/pluginutils"
+	"github.com/solo-io/gloo/projects/gloo/pkg/upstreams"
 
 	envoy_config_core_v3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	envoy_config_route_v3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
@@ -33,18 +37,21 @@ var (
 )
 
 // Puts Header Manipulation config on Routes, VirtualHosts, and Weighted Clusters
-type plugin struct{}
+type plugin struct {
+	enforceMatchingNamespaces bool
+}
 
 func NewPlugin() *plugin {
-	return &plugin{}
+	return &plugin{
+		enforceMatchingNamespaces: strings.ToLower(os.Getenv(api_conversion.MatchingNamespaceEnv)) == "true",
+	}
 }
 
 func (p *plugin) Name() string {
 	return ExtensionName
 }
 
-func (p *plugin) Init(_ plugins.InitParams) {
-}
+func (p *plugin) Init(_ plugins.InitParams) {}
 
 func (p *plugin) ProcessWeightedDestination(
 	params plugins.RouteParams,
@@ -55,8 +62,15 @@ func (p *plugin) ProcessWeightedDestination(
 	if headerManipulation == nil {
 		return nil
 	}
-
-	envoyHeader, err := convertHeaderConfig(headerManipulation, getSecretsFromSnapshot(params.Snapshot))
+	upstreamNamespace := "sometenant"
+	if p.enforceMatchingNamespaces {
+		us, err := upstreams.DestinationToUpstreamRef(in.GetDestination())
+		if err == nil {
+			upstreamNamespace = us.GetNamespace()
+		}
+	}
+	headerSecretOptions := api_conversion.HeaderSecretOptions{EnforceNamespaceMatch: p.enforceMatchingNamespaces, UpstreamNamespace: upstreamNamespace}
+	envoyHeader, err := convertHeaderConfig(headerManipulation, getSecretsFromSnapshot(params.Snapshot), headerSecretOptions)
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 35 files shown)


projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -79,8 +93,23 @@ func (p *plugin) ProcessVirtualHost(
 	if headerManipulation == nil {
 		return nil
 	}
-
-	envoyHeader, err := convertHeaderConfig(headerManipulation, getSecretsFromSnapshot(params.Snapshot))
+	headerSecretOptions := api_conversion.HeaderSecretOptions{
+		EnforceNamespaceMatch: p.enforceMatchingNamespaces,
+	}
+	if p.enforceMatchingNamespaces && len(in.GetRoutes()) > 0 {
```
bewebi, line 99: Similar to the comment above, I might remove the `p.enforceMatchingNamespaces` clause here so that we have a usable namespace in all case if possible

That said, there is a more significant performance implication here, since `getUpstreamNamespaceForRouteAction()` appears to be linear, and we execute that per route below, so I end up ambivalent

I do think that the behavior should be consistent between `ProcessWeightedDestination()`, `ProcessVirtualHost()`, and `ProcessRoute()` and if we choose to only populate `Upstream` when `EnforceNamespaceMatch` is `true`, we should document that in an in-line comment so that its discoverable in the future

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -100,7 +129,12 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 		return nil
 	}
 
-	envoyHeader, err := convertHeaderConfig(headerManipulation, getSecretsFromSnapshot(params.Snapshot))
+	headerSecretOptions := api_conversion.HeaderSecretOptions{EnforceNamespaceMatch: p.enforceMatchingNamespaces}
+	// If we need to enforce that the namespace of the secret and upstream match, then find the upstream namespace.
+	if p.enforceMatchingNamespaces {
```
bewebi, line 134: See above comments on `ProcessWeightedDestination()`, and `ProcessVirtualHost()`

projects/gloo/pkg/plugins/headers/plugin_test.go
```diff
@@ -141,6 +142,104 @@ var _ = Describe("Plugin", func() {
...
+							Namespace: "bar",
+						},
+					},
+				},
+			},
+		}
+		singleRouteToBadUpstream := &v1.Route{Action: &v1.Route_RouteAction{RouteAction: &v1.RouteAction{
+			Destination: &v1.RouteAction_Single{Single: &v1.Destination{DestinationType: &v1.Destination_Upstream{Upstream: &coreV1.ResourceRef{Name: "some-us", Namespace: "bad-tenant"}}}}}}}
+		singleRouteToGoodUpstream := &v1.Route{Action: &v1.Route_RouteAction{RouteAction: &v1.RouteAction{
+			Destination: &v1.RouteAction_Single{Single: &v1.Destination{DestinationType: &v1.Destination_Upstream{Upstream: &coreV1.ResourceRef{Name: "some-us", Namespace: "bar"}}}}}}}
```
bewebi, line 172: Might be cleaner to declare these variables within the `Context()` and set them in the `BeforeEach()` so that test cases can manipulate them as needed (ie in the two Route tests on 225 and 232)

projects/gloo/pkg/plugins/headers/plugin_test.go
```diff
@@ -141,6 +142,104 @@ var _ = Describe("Plugin", func() {
...
+		It("Errors with a WeightedDestination where the upstream namespace doesn't match the secret", func() {
```
bewebi, line 173: Can we add a success case for WeightedDestination?

projects/gloo/pkg/plugins/headers/plugin_test.go
```diff
@@ -141,6 +142,104 @@ var _ = Describe("Plugin", func() {
...
+			routeParamsWithSecret := plugins.RouteParams{VirtualHostParams: plugins.VirtualHostParams{Params: paramsWithSecret}}
+			out := &envoy_config_route_v3.WeightedCluster_ClusterWeight{}
+			in := &v1.WeightedDestination{
```
bewebi, line 176: nit: Assuming we add a success case for WeightedDestination, and perhaps even if not, we should pull out the WeightedDestination as a Context-scoped var as we do with `singleRouteToBadUpstream` and `singleRouteToGoodUpstream`

projects/gloo/pkg/translator/clusters.go
```diff
@@ -177,7 +179,10 @@ func createHealthCheckConfig(upstream *v1.Upstream, secrets *v1.SecretList) ([]*
 		if hc.GetHealthChecker() == nil {
 			return nil, NilFieldError(fmt.Sprintf("HealthCheck[%d].HealthChecker", i))
 		}
-		converted, err := api_conversion.ToEnvoyHealthCheck(hc, secrets)
+
+		shouldEnforceNamespaceMatch := strings.ToLower(os.Getenv(api_conversion.MatchingNamespaceEnv)) == "true"
```
bewebi, line 183: See previous comment about accepting truthy values

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -1172,91 +1173,145 @@ var _ = Describe("Translator", func() {
...
+				Expect(errs.Validate()).To(MatchError(ContainSubstring("list did not find secret bad-tenant.foo")))
 
-			expectedResult[0].GetHttpHealthCheck().RequestHeadersToAdd = []*envoy_config_core_v3.HeaderValueOption{
-				{
-					Header: &envoy_config_core_v3.HeaderValue{
-						Key:   "Authorization",
-						Value: "basic dXNlcjpwYXNzd29yZA==",
+			})
+			DescribeTable("can translate health check with secret header", func(enforceMatch, secretNamespace string) {
+				os.Setenv(api_conversion.MatchingNamespaceEnv, enforceMatch)
```
bewebi, line 1226: nit: we can check the swallowed error

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -1172,91 +1173,145 @@ var _ = Describe("Translator", func() {
...
+				_, errs, _ := translator.Translate(params, proxy)
+				Expect(errs.Validate()).To(MatchError(ContainSubstring("list did not find secret bad-tenant.foo")))
 
-			expectedResult[0].GetHttpHealthCheck().RequestHeadersToAdd = []*envoy_config_core_v3.HeaderValueOption{
-				{
-					Header: &envoy_config_core_v3.HeaderValue{
-						Key:   "Authorization",
-						Value: "basic dXNlcjpwYXNzd29yZA==",
+			})
+			DescribeTable("can translate health check with secret header", func(enforceMatch, secretNamespace string) {
```
bewebi, line 1225: I think it would be cleaner to include the failure case in the table, with Entries accepting an additional boolean argument like `shouldSucceed` which determines whether we make the success assertions or the error assertions

bewebi (commented):
Looks good! Some suggestions, mostly nitty/subjective

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -55,8 +62,15 @@ func (p *plugin) ProcessWeightedDestination(
...
 	}
-
-	envoyHeader, err := convertHeaderConfig(headerManipulation, getSecretsFromSnapshot(params.Snapshot))
+	upstreamNamespace := "sometenant"
+	if p.enforceMatchingNamespaces {
+		us, err := upstreams.DestinationToUpstreamRef(in.GetDestination())
+		if err == nil {
+			upstreamNamespace = us.GetNamespace()
+		}
+	}
```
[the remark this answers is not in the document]
elcasteel: I kept the conditional because the performance impacts for routes and virtualhosts is non trivial and I can't think of any reason why we'd need the namespace otherwise that is remotely related to this current feature.

projects/gloo/pkg/plugins/headers/plugin_test.go
elcasteel: updated

projects/gloo/pkg/plugins/headers/plugin_test.go
elcasteel: added

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -33,18 +37,21 @@ var (
...
 // Puts Header Manipulation config on Routes, VirtualHosts, and Weighted Clusters
-type plugin struct{}
+type plugin struct {
+	enforceMatchingNamespaces bool
+}
 
 func NewPlugin() *plugin {
-	return &plugin{}
+	return &plugin{
+		enforceMatchingNamespaces: strings.ToLower(os.Getenv(api_conversion.MatchingNamespaceEnv)) == "true",
```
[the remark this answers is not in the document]
elcasteel: I think we should mostly err on the side of not doing this check. I added a check for "1" and this does convert to all lower case so any variation of "true" will be ok.

projects/gloo/pkg/plugins/headers/plugin.go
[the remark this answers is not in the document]
bewebi: makes sense, the comments look good!

bewebi (dismissed):

projects/gloo/pkg/translator/clusters.go
```diff
@@ -177,7 +179,10 @@ func createHealthCheckConfig(upstream *v1.Upstream, secrets *v1.SecretList) ([]*
 		if hc.GetHealthChecker() == nil {
 			return nil, NilFieldError(fmt.Sprintf("HealthCheck[%d].HealthChecker", i))
 		}
-		converted, err := api_conversion.ToEnvoyHealthCheck(hc, secrets)
+		envVar := strings.ToLower(os.Getenv(api_conversion.MatchingNamespaceEnv))
```
sam-heilbron, line 182: I think we could re-use the IsEnvTruthy utility. Also, since this will run per upstream, should we try to pull the env variable reading to an init function so it happens once?

sam-heilbron (commented):
Nice work! 

I think the unit tests cover all of the functionality, but I still think its wortwhile to add an e2e test (kubernetes ideally) to demonstrate the funcationality and help prevent regressions.

pkg/utils/api_conversion/type.go
```diff
@@ -46,12 +55,12 @@ func ToEnvoyInt64Range(int64Range *envoytype_gloo.Int64Range) *envoy_type_v3.Int
...
 }
 
-func ToEnvoyHeaderValueOptionList(option []*envoycore_sk.HeaderValueOption, secrets *v1.SecretList) ([]*envoy_config_core_v3.HeaderValueOption, error) {
+func ToEnvoyHeaderValueOptionList(option []*envoycore_sk.HeaderValueOption, secrets *v1.SecretList, secretOptions HeaderSecretOptions) ([]*envoy_config_core_v3.HeaderValueOption, error) {
 	result := make([]*envoy_config_core_v3.HeaderValueOption, 0)
 	var err error
 	var opts []*envoy_config_core_v3.HeaderValueOption
 	for _, v := range option {
-		opts, err = ToEnvoyHeaderValueOptions(v, secrets)
+		opts, err = ToEnvoyHeaderValueOptions(v, secrets, secretOptions)
```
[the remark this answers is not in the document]
elcasteel: This avoids filtering the secrets when we don't need to read a secret - there's a switch in `ToEnvoyHeaderValueOptions` that conditionally looks in the SecretList

changelog/v1.15.0-beta23/match-secret-refs.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/5007#event-9581124373
+    resolvesIssue: false
+    description: > 
+         Add an environment variable "HEADER_SECRET_REF_NS_MATCHES_US" that when set in the gloo pod requires that secrets
+         sent in headers to an upstream are in the same namespace as that upstream.
```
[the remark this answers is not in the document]
elcasteel: Updated here and in values.go to include that it defaults to false.

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -33,18 +37,22 @@ var (
...
-type plugin struct{}
+type plugin struct {
+	enforceMatchingNamespaces bool
+}
 
 func NewPlugin() *plugin {
-	return &plugin{}
+	envVar := strings.ToLower(os.Getenv(api_conversion.MatchingNamespaceEnv))
+	return &plugin{
+		enforceMatchingNamespaces: envVar == "true" || envVar == "1",
```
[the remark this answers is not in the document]
elcasteel: I used the same `strconv.ParseBool` instead of importing the test utils

bewebi (approved):

changelog/v1.15.0-beta23/match-secret-refs.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/5007#event-9581124373
```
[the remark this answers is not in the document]
elcasteel: good catch, removed

bewebi (dismissed):

test/e2e/headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+	coreV1 "github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+
+	"github.com/solo-io/gloo/test/testutils"
+
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/solo-io/gloo/test/e2e"
+	"github.com/solo-io/gloo/test/helpers"
+)
+
+var _ = Describe("Test", Label(), func() {
```
sam-heilbron, line 24: nit: Can we clarify what this is testing? Also do we need the label?

test/e2e/headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+			Kind: &gloov1.Secret_Header{
+				Header: &gloov1.HeaderSecret{
+					Headers: map[string]string{
+						"Authorization": "basic dXNlcjpwYXNzd29yZA==",
+					},
+				},
+			},
+			Metadata: &coreV1.Metadata{
+				Name:      "foo",
+				Namespace: writeNamespace,
```
sam-heilbron, line 44: nit: It's confusing that the namespace would be the writeNamespace ( I would imagine upstream are also put in the write namespace). It might be clearer if we make the namespace explicitly something else

test/e2e/headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+var _ = Describe("Test", Label(), func() {
+
+	var (
+		testContext *e2e.TestContext
+	)
+
+	BeforeEach(func() {
+		var testRequirements []testutils.Requirement
+		testContext = testContextFactory.NewTestContext(testRequirements...)
```
sam-heilbron, line 32: nit: We can just call NewTestContext() without passing in anything

test/e2e/headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+	Context("With matching not enforced", func() {
+
+		BeforeEach(func() {
+			os.Setenv(api_conversion.MatchingNamespaceEnv, "false")
+		})
+
+		It("Accepts all virtual services", func() {
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				vs, err := testContext.TestClients().VirtualServiceClient.Read(writeNamespace, "bad", clients.ReadOpts{})
+				fmt.Printf("status: %v", vs.GetNamespacedStatuses())
```
sam-heilbron, line 112: Can we remove this?

test/e2e/headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	JustBeforeEach(func() {
+		testContext.JustBeforeEach()
+	})
+
+	JustAfterEach(func() {
+		testContext.JustAfterEach()
+	})
+
+	Context("With matching not enforced", func() {
```
sam-heilbron, line 103: I like how these tests explicitly show the difference in behavior!

sam-heilbron (commented):
Looking good! Mostly nits, so im happy to approve as is. Since it would be the second approval just wanted to run them by you before doing so.

Do we have recommendations around upgrading to use this feature in our docs (ie what to be aware of if you turn this on and you have resources that become invalid)?

test/e2e/headers_test.go
elcasteel: We only watch `default` and `gloo-system` (writeNamespace). The upstream is written to `default`

test/e2e/headers_test.go
elcasteel: I only caught the above thing with watched namespaces by having this.

test/e2e/headers_test.go
```diff
@@ -0,0 +1,134 @@
...
+				Append: &wrappers.BoolValue{Value: true}}},
+			RequestHeadersToRemove: []string{"a"},
+		}}
+
+		testContext.ResourcesToCreate().VirtualServices = v1.VirtualServiceList{goodVS, badVS}
+		testContext.ResourcesToCreate().Secrets = gloov1.SecretList{forbiddenSecret, allowedSecret}
+	})
+
+	AfterEach(func() {
+		os.Setenv(api_conversion.MatchingNamespaceEnv, "")
```
sam-heilbron, line 89: Let's use UnsetEnv instead

sam-heilbron (approved):
🚀

bewebi (approved):

PR #8535, opened 2023-08-03
[1.14] Introduce GHA Go Cache on LTS

sam-heilbron:
# Description

Run a job on push to branch, to ensure a go cache entry exists for the LTS branch

## CI changes
- Create a new cache job to run on pushes to LTS branches

# Context
https://github.com/solo-io/solo-apis/pull/847#issuecomment-1552915687

We need to introduce this "on push" trigger on the branch that is performing the action, not just on the main branch

I specified the branch regex, because otherwise, if you just use "on push", all pushes to all branches will re-run this. We only want to run this on a push to an LTS branch

## Interesting decisions
 
There are 2 ways (as I see it) that we can have a job run on push to the branch:
1. Introduce a new job, that only runs on push
2. Re-use an existing job, and have it run on push

When deciding, I wanted to really focus on:
- Efficiency (let's avoid running jobs unnecessarily when we can)
- Maintainability (let's avoid introducing a maintenance burden every time we cut an LTS branch)

At first I went with Option 2. I modified the codegen job to run on pushes to main. This has the benefit of re-using an existing job, and a low maintenance burden, because we can add this to the main branch, and leave it as is, when we cut LTS branches. The downside is it runs a 5 minute job that is unnecessary. Also, if we need other caches reset/seeded, the codegen job might not provide that. As a result, I opted for Option 1. 

The value I see is:
- We control what runs, so we can run the minimal code (just setup-go for now)
- If we need to expand what we cache, we can easily update the new job, since it's sole responsibility is maintaining the caches
- We can add this to the main branch and don't need to modify it when we cut new branches so it's maintainable

## Testing steps

It's not really possible to test this without merging this code.

I did however, introduce this without a filter on the branches it should run on, and you can see that it correctly created the cache: 
Screen Shot 2023-08-03 at 9 26 42 AM

## Follow-Up
After this merges, we can then test this, and ensure PRs against this LTS branch benefit from caching. For a list of next steps, see https://github.com/solo-io/solo-projects/issues/4903#issuecomment-1664194821

# Checklist:

### DIFF
--- a/.github/workflows/cache.yaml
+++ b/.github/workflows/cache.yaml
@@ -2,7 +2,10 @@ name: Cache
 
 on:
   # We utilize this job to seed the GitHub action cache(s) for the LTS branch
-  push: {}
+  push:
+    branches:
+      - 'main'
+      - 'v1.**.x'
 
 jobs:
   go-modules:
--- a/.github/workflows/cache.yaml
+++ b/.github/workflows/cache.yaml
@@ -1,0 +1,27 @@
+name: Cache
+
+on:
+  # We utilize this job to seed the GitHub action cache(s) for the LTS branch
+  push: {}
+
+jobs:
+  go-modules:
+    name: Cache Go Modules
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Cancel Previous Actions
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      # This is the crucial step
+      # The setup-go action has caching built-in,
+      # so when this runs, a cache entry will be created, if it does not already exist
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
+        id: go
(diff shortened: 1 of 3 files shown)


nfuden (approved):
Approved as long as we take the next steps of validating as alluded to in description

.github/workflows/cache.yaml
```diff
@@ -0,0 +1,30 @@
+name: Cache
+
+on:
+  # We utilize this job to seed the GitHub action cache(s) for the LTS branch
+  push:
```
elcasteel, line 5: if you want to test this before merging you could temporarily trigger it on pull requests

sam-heilbron: In the PR body, I point out that I didn't include the filter for the branches at first, so I could prove that this job does run and produce a cache. What I can't prove/test is that when this merges, the cache will be associated with the 1.14 branch and useful for future PRs

elcasteel (approved):

.github/workflows/cache.yaml
sam-heilbron: Its a good point though, so from it I opened a PR against this branch: https://github.com/solo-io/gloo/pull/8537

and the job that ran https://github.com/solo-io/gloo/actions/runs/5753433524/job/15596620617?pr=8537 was able to use the cache built from this branch. While not a perfect test, it means that I'm pretty confident this will work

PR #8538, opened 2023-08-03
[1.14] Use Native Cache Action (instead of setup-go)

sam-heilbron:
# Description

Continuation of work from https://github.com/solo-io/gloo/pull/8535

## CI changes
- Use a shared action to setup the go code
- Use the base cache action, instead of the caching provided by the setup-go action
- Remove build cache from entries that we store

# Context

I noticed inconsistent cache behaviors: https://github.com/solo-io/solo-projects/issues/4903#issuecomment-1664341061 

The hash used in the cache key was identical for a PR that was merged into 1.14 as a cache entry for the main (1.15) branch. This shouldn't be the case because the go.sums are very different

The build cache (understandably) grows between the time we:
- Before the job (`188M	/home/runner/.cache/go-build`, https://github.com/solo-io/gloo/actions/runs/5765012033/job/15630127216?pr=8538)
- After codegen runs (`2.1G	/home/runner/.cache/go-build`, https://github.com/solo-io/gloo/actions/runs/5765012033/job/15630127216?pr=8538)
- After regression test run (`3.3G	/home/runner/.cache/go-build`, https://github.com/solo-io/gloo/actions/runs/5765012028/job/15630131257)

It's not clear to me what elements from that cache are useful for other jobs, and under what circumstances that cache should be invalidated (ie what do we use to build a cache key). As a result, I don't feel comfortable including it in our CI. I would rather understand exactly what is cached, and if we decide we need more caching, add it. 

It should be noted [this mirrors GME](https://github.com/solo-io/gloo-mesh-enterprise/blob/a68c23c21e09274a76caac47fa218a3505a176b9/.github/workflows/pull_request.yaml#L57)

## Interesting decisions
 
I opted to just rely on the base cache action since it provides more functionality. To avoid drift in how we use it, I wrapped this setup in a shared github action.

## Testing steps

Again, I can prove that the cache is built, but there isn't a perfect way to prove out the cache will be re-used for future PRs against this branch.

## Notes for reviewers

None

# Checklist:


elcasteel (approved):
LGTM - do we need to put some of these changes on main as well?

PR #8550, opened 2023-08-08
[1.13] Introduce GHA Cache on LTS

sam-heilbron:
# Description

Run a job on push to branch, to ensure a go cache entry exists for the LTS branch

**This is a backport of work introduced into 1.14**

## CI changes
- Create a new cache job to run on pushes to LTS branches
- Update existing jobs to use a composite-action

# Context
https://github.com/solo-io/solo-apis/pull/847#issuecomment-1552915687

## 1.14 Changes
The work that merged into 1.14 was introduced in the following segments:
- https://github.com/solo-io/gloo/pull/8535
- https://github.com/solo-io/gloo/pull/8538
- https://github.com/solo-io/gloo/pull/8545

## Interesting decisions
I followed the same patterns/decisions as the changes in the 1.14 branch
 
## Testing steps
Since this was tested for 1.14, we assume this will work, but it is good to verify. This can be done after this PR merges.

## Follow-Up
After this merges, we can then test this, and ensure PRs against this LTS branch benefit from caching.

# Checklist:


.github/workflows/cache-setup.yaml
```diff
@@ -0,0 +1,18 @@
+name: Cache
```
nfuden, line 1: So ive been wondering if we should start consiering having discrete naming such as on the lts branch being merged.
That being said I think this is the correct approach here but its a decision that im trying to think about harder in general and socializea

nfuden (approved):

bewebi (approved):

PR #8553, opened 2023-08-08
Update Nightly tests

sam-heilbron:
# Work In Progress
I will run these tests manually, and link the logs that are produced, to confirm they work as expected

# Description

Update nightly tests to ensure they run the proper branches, and complete successfully

## CI changes
- Adjust the branches used in our nightly tests (removed 1.13, added 1.15)
- Adjust the cron used in nightly tests (simplified to run on the hour)
- Adjust the syntax used in 1.14 nightly tests

# Context

## Cron Update
The cron was a bit confusing to reason through so I added comments to clarify the syntax. Also, I updated the timing to run at the top of the hour so it was easier to reason through.

## Composite Actions
In https://github.com/solo-io/gloo/pull/8538 we moved the location of the regression-tests action on the 1.14 branch. As a result, we need to update the nightly job to reference that code. What is nice is that this brings parity with our other branches, so all of the recent branches use this folder structure.

## Branch Updates
Now that 1.15 is an LTS branch we should run nightly tests against it.

1.13 was a noisy branch, whose flakes are accounted for on later versions.

## Testing steps

TODO: I plan on running this job manually, and linking the logs/slack to verify the success

## Notes for reviewers

# Checklist:

### DIFF
--- a/.github/workflows/nightly-tests.yaml
+++ b/.github/workflows/nightly-tests.yaml
@@ -4,10 +4,16 @@ env:
   GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }} # necessary to pass upgrade tests
 
 on:
+  # https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07
+  # Minute [0,59]
+  # Hour [0,23]
+  # Day of the month [1,31]
+  # Month of the year [1,12]
+  # Day of the week ([0,6] with 0=Sunday)
   schedule:
-    - cron: "0 5 * * 1-5"   # weekdays @ 00:00 EST, run tests against latest main
-    - cron: "10 10 * * 1"   # monday   @ 05:10 EST, run expanded tests against v1.14.x
-    - cron: "6 6 * * 1"     # monday   @ 01:06 EST, run expanded tests against v1.13.x
+    - cron: "0 5 * * 1-5" # weekdays @ 00:00 EST, run tests against latest main
+    - cron: "0 6 * * 1"   # monday   @ 01:00 EST, run expanded tests against v1.15.x
+    - cron: "0 7 * * 1"   # monday   @ 02:00 EST, run expanded tests against v1.14.x
   workflow_dispatch:
     inputs:
       run-regression:
@@ -46,9 +52,25 @@ jobs:
       - uses: ./.github/workflows/composite-actions/prep-go-runner
       - uses: ./.github/workflows/composite-actions/performance-tests
 
+  regression_tests_15:
+    name: v1.15.x regression tests
+    if: github.event.schedule == '0 6 * * 1'
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        kube-e2e-test-type: ['gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'glooctl', 'upgrade']
+        kube-version: [ { node: 'v1.23.13@sha256:ef453bb7c79f0e3caba88d2067d4196f427794086a7d0df8df4f019d5e336b61', kubectl: 'v1.23.17', kind: 'v0.17.0', helm: 'v3.11.2' },
+                        { node: 'v1.27.3@sha256:3966ac761ae0136263ffdb6cfd4db23ef8a83cba8a463690e98317add2c9ba72', kubectl: 'v1.27.3', kind: 'v0.20.0', helm: 'v3.12.2' } ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: v1.15.x
+      - uses: ./.github/workflows/composite-actions/regression-tests
+
   regression_tests_14:
     name: v1.14.x regression tests
-    if: github.event.schedule == '10 10 * * 1'
+    if: github.event.schedule == '0 7 * * 1'
     runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-tests.yaml
```diff
@@ -4,23 +4,29 @@ env:
...
+  # Day of the month [1,31]
+  # Month of the year [1,12]
+  # Day of the week ([0,6] with 0=Sunday)
   schedule:
-    - cron: "0 5 * * 1-5"   # weekdays @ 00:00 EST, run tests against latest main
-    - cron: "10 10 * * 1"   # monday   @ 05:10 EST, run expanded tests against v1.14.x
-    - cron: "6 6 * * 1"     # monday   @ 01:06 EST, run expanded tests against v1.13.x
+    - cron: "0 5 * * 1-5" # weekdays @ 01:00 EST, run tests against latest main
+    - cron: "0 6 * * 1"   # monday   @ 02:00 EST, run expanded tests against v1.15.x
+    - cron: "0 7 * * 1"   # monday   @ 03:00 EST, run expanded tests against v1.14.x
```
bewebi, line 16: I know I get confused about EST and EDT...perhaps just using UTC in the comment would help avoid confusion as to what time each job is running locally (though also maybe it wouldn't)

Another way to help avoid confusion might be to have a non-zero minute, perhaps corresponding to the minor version

I'm not sure either is strictly necessary or even an improvement, as the names of the jobs themselves are clear, but worth considering

.github/workflows/nightly-tests.yaml
```diff
@@ -4,23 +4,29 @@ env:
...
   workflow_dispatch:
     inputs:
-      run-regression:
-        description: "Run regression tests"
-        type: boolean
+      run-regression-branch:
+        description: "Run regression tests for a branch (main, v1.15.x, ..etc)"
+        type: string
```
bewebi, line 21: Instead of overloading one input, let's add a new input to allow specifying the branch, and this can apply to regression and/or performance tests

.github/workflows/nightly-tests.yaml
```diff
@@ -4,23 +4,29 @@ env:
...
       run-performance:
         description: "Run performance tests"
         type: boolean
 
 jobs:
   regression_tests_latest_main:
```
bewebi, line 27: While we're here, any reason this shouldn't be `regression_tests_main`? I don't think anyone would think it was some non-HEAD commit from the `main` branch

.github/workflows/nightly-tests.yaml
sam-heilbron: I like it!

.github/workflows/nightly-tests.yaml
sam-heilbron: Yeah good call, Ill just update to UTC and call it a day. At the end of the day this comment is just trying to help a bit, but really the slack notifications are what we (I) tend to rely on

.github/workflows/nightly-tests.yaml
```diff
@@ -11,22 +11,26 @@ on:
...
+    - cron: "0 5 * * 1-5" # weekdays @ 05:00 UTC, run tests against latest main
+    - cron: "0 6 * * 1"   # monday   @ 06:00 UTC, run expanded tests against v1.15.x
+    - cron: "0 7 * * 1"   # monday   @ 07:00 UTC, run expanded tests against v1.14.x
   workflow_dispatch:
     inputs:
-      run-regression-branch:
-        description: "Run regression tests for a branch (main, v1.15.x, ..etc)"
+      branch:
+        description: "The branch to run tests against (main, v1.15.x, ..etc)"
         type: string
```
bewebi, line 21: Since there are three specific options we support, can we make this a dropdown?
If we supported arbitrary branches a string would make more sense, but that's not currently the case and I don't think we want it to be

.github/workflows/nightly-tests.yaml
```diff
@@ -84,6 +79,15 @@ jobs:
         ref: v1.14.x
     - uses: ./.github/workflows/composite-actions/regression-tests
 
+  performance_tests_main:
+    name: main performance tests
+    if: ${{ (github.event_name == 'workflow_dispatch' && inputs.run-performance && inputs.branch == 'main') || github.event.schedule == '0 5 * * 1-5' }}
```
bewebi, line 84: Can we also add a job for performance for 1.15?
If we don't want to run it nightly that's fine, but can at least be done on `workflow_dispatch` if desired

1.14 isn't set up for performance tests using the composite action, so we _could_ add a 1.14 performance job that fails with a "not implemented" message so that a user who manually triggers performance tests for that branch gets a loud failure with an explanation

.github/workflows/nightly-tests.yaml
sam-heilbron: I figured what is nice about being able to specify this manually, is you could introduce tests or make changes to tests, and then run them through this workflow. With a dropdown, we limit devs in what they can run (only code that has been merged on an LTS branch)

.github/workflows/nightly-tests.yaml
sam-heilbron: Yeah good call, ill update

.github/workflows/nightly-tests.yaml
bewebi: But all of the jobs require one of three specific strings or nothing will run
We don't support the specifying of arbitrary branches so there's no reason to allow it

If we ever modify the action such that we can run tests for an arbitrarily-specified branch then it would make sense to have string input

sam-heilbron: Oh yeah good point, I forget we have jobs per branch because we use different kube versions, it's not just a generic job that takes inputs. I can update to a dropdown

bewebi (dismissed):

bewebi (approved):

Issue #8568, opened 2023-08-10
`google.protobuf.UInt64Value` does not marshal correctly in kubernetes CRDs 


davidjumani:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.15.x

### Kubernetes Version

v1.26

### Describe the bug

If a proto field is `google.protobuf.UInt64Value`, it does not marshal correctly
Eg: 
```
httpGateway:
   options:
     caching:
        maxPayloadSize: 3

The Gateway "tcp-limit" is invalid: spec.httpGateway.options.caching.maxPayloadSize: Invalid value: "integer": spec.httpGateway.options.caching.maxPayloadSize in body must be of type object: "integer"



  httpGateway:
   options:
     caching:
        maxPayloadSize: 
          value: 3


Error from server: error when creating "gw.yaml": admission webhook "gloo.gloo-system.svc" denied the request: resource incompatible with current Gloo snapshot: [1 error occurred:
	* could not unmarshal raw object: parsing resource from crd spec tcp-limit in namespace gloo-system into *v1.Gateway: json: cannot unmarshal object into Go value of type uint64
]
```

AFAIK this affects :
- https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/enterprise/options/caching/caching.proto#L28
- https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/options/connection_limit/connection_limit.proto#L23

### Expected Behavior

```
httpGateway:
   options:
     caching:
        maxPayloadSize: 3
```
should succeed

### Steps to reproduce the bug

Apply the above schema

davidjumani:
The bug is fixed in https://github.com/solo-io/protoc-gen-openapi/pull/9 and imported with tests in https://github.com/solo-io/solo-kit/pull/547
Will bump the deps in gloo and enterprise with the fix

davidjumani:
Merged into main. Will raise PRs to backport as well

PR #8570, opened 2023-08-10
fix: google.protobuf.UInt64Value does not marshal correctly in CRDs

davidjumani:
# Description

Fixes incorrect marshalling of `google.protobuf.UInt64Value` in CRDs

# Context

https://github.com/solo-io/gloo/issues/8568

## Testing steps
Additional tests in https://github.com/solo-io/solo-kit/pull/547

```
  kind: Gateway
  metadata:
    name: tcp-limit
    namespace: gloo-system
  spec:
    bindAddress: '::'
     bindPort: 7080
     httpGateway:
     options:
       caching:
          maxPayloadSize: 3.0
```
Before :
```
The Gateway "tcp-limit" is invalid: spec.httpGateway.options.caching.maxPayloadSize: Invalid value: "integer": spec.httpGateway.options.caching.maxPayloadSize in body must be of type object: "integer"
```

After :
```
gateway.gateway.solo.io/tcp-limit created
```

# Checklist:


install/helm/gloo/crds/gateway.solo.io_v1_Gateway.yaml
```diff
@@ -76,12 +76,10 @@ spec:
                                 type: string
                             type: object
                           maxPayloadSize:
-                            properties:
-                              value:
-                                format: int64
-                                type: integer
-                                x-kubernetes-int-or-string: true
-                            type: object
+                            maximum: 1.8446744073709552e+19
```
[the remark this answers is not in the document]
davidjumani: It is added [here](https://github.com/solo-io/gloo/blob/main/test/e2e/connection_limit_test.go#L80) however since this doesn't involve parsing raw yaml it didn't catch this issue. Perhaps adding a test in solo-kit to parse it into a CR would be a good test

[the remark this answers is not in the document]
jenshu: we have some helm e2e tests that do install/upgrades in k8s. perhaps a good place to add this int64 field is under the options [here](https://github.com/solo-io/gloo/blob/f10e498d6fc6524336babba399e764cac5807c0d/test/kube2e/helm/helm_test.go#L623). those helm values are used to test upgrades from other gloo edge versions to the current version

sam-heilbron (dismissed):
LGTM

jenshu (dismissed):

sam-heilbron (dismissed):

projects/gloo/api/v1/enterprise/options/caching/caching.proto
```diff
@@ -25,7 +25,7 @@ message Settings {
    // Max payload size to cache. If unset defaults to a reasonable value.
    // If explicitly set to 0 will prevent anything with a body from
    // being cached.
-   google.protobuf.UInt64Value max_payload_size = 4;
+   google.protobuf.UInt32Value max_payload_size = 4;
```
davidjumani, line 28: Unmarshalling int64 leads to an overflow issue. Ref: 
- https://github.com/OpenAPITools/openapi-generator/issues/11940
- https://github.com/go-swagger/go-swagger/issues/581
- https://github.com/go-swagger/go-swagger/issues/2755

The proto docs also mention that uin32 and uint64 are compatible and can be swapped out
https://protobuf.dev/programming-guides/proto2/#updating
```
int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn’t fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (for example, if a 64-bit number is read as an int32, it will be truncated to 32 bits).
```

projects/gloo/pkg/plugins/connection_limit/plugin.go
```diff
@@ -49,9 +50,11 @@ func GenerateFilter(connectionLimit *connection_limit.ConnectionLimit) ([]plugin
 	}
 
 	config := &envoy_config_connection_limit_v3.ConnectionLimit{
-		StatPrefix:     StatPrefix,
-		MaxConnections: connectionLimit.GetMaxActiveConnections(),
-		Delay:          connectionLimit.GetDelayBeforeClose(),
+		StatPrefix: StatPrefix,
+		MaxConnections: &wrapperspb.UInt64Value{
+			Value: uint64(connectionLimit.GetMaxActiveConnections().GetValue()),
```
jenshu, line 55: should this be UInt32Value?

test/kube2e/helm/helm_test.go
```diff
@@ -628,6 +628,9 @@ gatewayProxies:
                 keyValueAction:
                   keyToMask: test
                   name: test
+          # This checks the proper parsing of wrappers.UInt64Value
```
jenshu, line 631:
```suggestion
          # This checks the proper parsing of wrappers.UInt32Value
```

changelog/v1.16.0-beta3/fix-protobuf-uint64.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
+    description: Fix incorrect CRD generation of proto fileds of type `google.protobuf.UInt64Value`. Previously fields of this type had an incorrect schema which led to them failing being unmarshalled, which is now resolved. Affected protos include gateway specs `httpGateway.options.caching.maxPayloadSize` and `httpGateway.options.connectionLimit.maxActiveConnections`
```
sam-heilbron, line 3: I'm good with this update, as it reflects the user facing sentiment: we fixed the schemas for these types. Given the updates to our protos, do we want to clarify the actual changes now, or are we good with this as is?

projects/gloo/pkg/plugins/connection_limit/plugin.go
davidjumani: The envoy type is uint64 while the solo API type is uint32, so converting between the two over here

sam-heilbron (commented):
Looks good! Before we backport this, let's also pull this into enterprise (in a PR with passing CI), to prove out that the update to the caching API is also fixed properly

changelog/v1.16.0-beta3/fix-protobuf-uint64.yaml
davidjumani: Makes sense, I'll update the changelog

projects/gloo/pkg/plugins/connection_limit/plugin.go
jenshu: ohh right

jenshu (approved):

sam-heilbron (approved):

PR #8574, opened 2023-08-11
Doc fixes

Rachael-Graham:
Weekly doc fixes for main.


docs/content/_index.md
```diff
@@ -33,7 +33,7 @@ When we say Gloo Edge is a "next-generation" gateway, we mean that it was purpos
 ## Blogs & Demos
 
 * [Blog Articles](https://www.solo.io/blog/announcing-gloo-the-function-gateway/)
-* [Watch Gloo Edge Videos and Demos](https://www.youtube.com/playlist?list=PLBOtlFtGznBiN5dZmaYsP-VxoVxOdxsVq)
+* [Watch Gloo Edge Videos and Demos](https://www.youtube.com/watch?v=HEN8IHCOqSo&list=PLBOtlFtGznBgrbLwyrPNIsdXLuq-mqa6U)
```
jenshu, line 36: it seems this goes directly to a video. do we want to link to the entire playlist instead?
https://www.youtube.com/playlist?list=PLBOtlFtGznBgrbLwyrPNIsdXLuq-mqa6U
(also, there seem to only be 2 videos in that playlist, is that the right one?)

elcasteel (approved):

docs/content/_index.md
artberger: @jenshu yeah, I think I changed it to the video link in part because there were only 2 videos in the playlist, so it might make it seem fuller if they open to a video? and then there is the playlist from the video if they want. But happy to change the link to the playlist!

In terms of the right playlist, it is the Gloo Edge one featured on our YouTube landing page [here](https://www.youtube.com/@soloio_inc/playlists).

PR #8575, opened 2023-08-11
Reject ratelimit configs that are missing required values.

elcasteel:
# Description
Add checks for the presence of the required fields for ratelimit actions (spelled out here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action) in the rate limit plugins. Reject the config if they are empty instead of letting Envoy fail. 
This change will also require updates in solo-projects and should not merge until those changes are tested

## API changes
The proto is already annotated to require these fields but we don't pass that through to our CRDs. 

## Docs changes
Docs show that these fields are required based on proto, and this change is enforcing the requirements.

# Context
I ran across this reproducing a different ratelimit translation issue and decided it was worth fixing. 

## Testing steps
Added unit test.
Manual testing and e2e test will happen on solo-projects PR.
## Notes for reviewers

You can use the envoy docs for ratelimit actions to verify that I got everything: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action

# Checklist:


davidjumani (dismissed):

changelog/v1.16.0-beta1/rl-nil-checks.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/8573
+    resolvesIssue: false
+    description: Check for empty values in fields that are required by Envoy for RateLimitActions.
```
sam-heilbron, line 5: nit: Since this is a fix, and the changelogs can be consumed by customers, can we clarify before this fix what field/value combination would produce what behavior, and after this fix what hte new behavior is?

projects/gloo/pkg/plugins/ratelimit/raw_util.go
```diff
@@ -16,30 +19,44 @@ import (
 // to treat those descriptors differently (for the set-style rate-limit API)
 const SetDescriptorValue = "solo.setDescriptor.uniqueValue"
 
+// TODO: this error and handling around it can be removed if we eventually generate CRDs with required values from the ratelimit proto.
```
sam-heilbron, line 22: I like this callout. Could we create an open sourece github issue to track the feature request of making proto field labels interpreted by our schema generation tool? I think things like: required and gte,lte would be nice additions. We could even link that issue in the comment here

projects/gloo/pkg/plugins/ratelimit/raw_util.go
```diff
@@ -16,30 +19,44 @@ import (
...
+			}
 			ret = append(ret, rl)
 		}
 	}
-	return ret
+	return ret, allErrors
 }
 
-func ConvertActions(ctx context.Context, actions []*solo_rl.Action) []*envoy_config_route_v3.RateLimit_Action {
+func ConvertActions(ctx context.Context, actions []*solo_rl.Action) ([]*envoy_config_route_v3.RateLimit_Action, error) {
```
sam-heilbron, line 50: nit: Since this is exported, could we add a docstring?

sam-heilbron (commented):
LGTM! Small nits but I'm happy to approve

sam-heilbron (approved):
🚀

jbohanon (approved):

PR #8589, opened 2023-08-16
ci: Fix check skip directives script

davidjumani:
# Description

This fixes a bug in [process-skip-directives script](https://github.com/solo-io/gloo/blob/main/ci/github-actions/process-skip-directives/script.sh#L28) that checks the changelog from main instead of the base branch

# Context

This causes issues when raising PRs against older branches as this can result in including changelogs from main which may contain the skip directives that can lead to kube tests being skipped

# Checklist:


sam-heilbron (dismissed):
🚀 

LGTM! If you want, you can also push a commit that uses the skip directive to prove that they are skipped properly still (and it will allow this to merge sooner)

sam-heilbron (approved):

PR #8597, opened 2023-08-18
ci: Fix skip directives script failing on merge

davidjumani:
# Description

Fixes the failing https://github.com/solo-io/gloo/blob/main/ci/github-actions/process-skip-directives/script.sh when it runs on merge - caused when githubBaseRef` is not defined
```
Prepare all required actions
Run ./.github/workflows/composite-actions/process-skip-directives
Run ./ci/github-actions/process-skip-directives/script.sh 
+ skipKubeTestsDirective=skipCI-kube-tests:true
+ shouldSkipKubeTests=false
+ skipDocsBuildDirective=skipCI-docs-build:true
+ shouldSkipDocsBuild=false
+ githubBaseRef=
++ git diff origin/ HEAD --name-only
++ grep changelog/
fatal: ambiguous argument 'origin/': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
+ changelog=
Error: Process completed with exit code 1.
```

## CI changes

Fixed the script to not error out when `githubBaseRef` is not defined

# Tests

When the base branch is passed
```
./ci/github-actions/process-skip-directives/script.sh main

+ skipKubeTestsDirective=skipCI-kube-tests:true
+ shouldSkipKubeTests=false
+ skipDocsBuildDirective=skipCI-docs-build:true
+ shouldSkipDocsBuild=false
+ githubBaseRef=main
+ '[' '!' -z main ']'
++ git diff origin/main HEAD --name-only
++ grep changelog/
+ changelog=changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
+ '[' '!' -z changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml ']'
++ echo changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
++ wc -l
++ tr -d ' '
+ [[ 1 = \1 ]]
+ echo 'exactly one changelog added since main'
exactly one changelog added since main
+ echo 'changelog file name == changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml'
changelog file name == changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
++ cat changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
++ grep skipCI-kube-tests:true
+ [[ -n       skipCI-kube-tests:true ]]
+ shouldSkipKubeTests=true
++ cat changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
++ grep skipCI-docs-build:true
+ [[ -n       skipCI-docs-build:true ]]
+ shouldSkipDocsBuild=true
+ echo skip-kube-tests=true
```

When no branch is passed
```
./ci/github-actions/process-skip-directives/script.sh 

+ skipKubeTestsDirective=skipCI-kube-tests:true
+ shouldSkipKubeTests=false
+ skipDocsBuildDirective=skipCI-docs-build:true
+ shouldSkipDocsBuild=false
+ githubBaseRef=
+ '[' '!' -z '' ']'
+ echo skip-kube-tests=false
```

# Checklist:

### DIFF
--- a/ci/github-actions/process-skip-directives/script.sh
+++ b/ci/github-actions/process-skip-directives/script.sh
@@ -13,7 +13,7 @@
 # by other steps in our CI pipeline
 ################################################################
 
-set -ex
+set -x
 
 skipKubeTestsDirective="skipCI-kube-tests:true"
 shouldSkipKubeTests=false
@@ -22,19 +22,21 @@ skipDocsBuildDirective="skipCI-docs-build:true"
 shouldSkipDocsBuild=false
 
 githubBaseRef=$1
-
-changelog=$(git diff origin/$githubBaseRef HEAD --name-only | grep "changelog/")
-if [[ $(echo $changelog | wc -l | tr -d ' ') = "1" ]]; then
-    echo "exactly one changelog added since main"
-    echo "changelog file name == $changelog"
-    if [[ $(cat $changelog | grep $skipKubeTestsDirective) ]]; then
-        shouldSkipKubeTests=true
+if [ ! -z "$githubBaseRef" ]; then
+    changelog=$(git diff origin/$githubBaseRef HEAD --name-only | grep "changelog/")
+    # An empty string is also one line in bash. Hence adding the first check
+    if [ ! -z "$changelog" ] && [[ $(echo $changelog | wc -l | tr -d ' ') = "1" ]]; then
+        echo "exactly one changelog added since main"
+        echo "changelog file name == $changelog"
+        if [[ $(cat $changelog | grep $skipKubeTestsDirective) ]]; then
+            shouldSkipKubeTests=true
+        fi
+        if [[ $(cat $changelog | grep $skipDocsBuildDirective) ]]; then
+            shouldSkipDocsBuild=true
+        fi
+    else
+        echo "no changelog found (or more than one changelog found) - not skipping CI"
     fi
-    if [[ $(cat $changelog | grep $skipDocsBuildDirective) ]]; then
-        shouldSkipDocsBuild=true
-    fi
-else
-    echo "no changelog found (or more than one changelog found) - not skipping CI"
 fi
 
 echo "skip-kube-tests=${shouldSkipKubeTests}" >> $GITHUB_OUTPUT
--- a/changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
+++ b/changelog/v1.16.0-beta3/fix-check-skip-directives-again.yaml
@@ -1,0 +1,7 @@
+changelog:
+  - type: NON_USER_FACING
+    description: >-
+      Fix process-skip-directive script that breaks when run after  merge.
+      skipCI-kube-tests:true
+      skipCI-docs-build:true
+


inFocus7 (commented):
Overall looks good! 1 suggestion + 1 question + below

Also do you happen to have a link to an example were this happened? I'm wondering why haven't noticed this before, or maybe it doesn't break anything major so that's why we haven't noticed. 🤔 

---

We have this sort of skip directive logic in solo-projects (and idk what other repos). It might be worth looking into if it's a possible issue there, probably by looking at if the actions which call skip-directive are ran outside of pull_requests. (Not saying you need to look into it, just something for us to keep in mind in case something like this pops up in other repos.)

ci/github-actions/process-skip-directives/script.sh
```diff
@@ -13,7 +13,7 @@
 # by other steps in our CI pipeline
 ################################################################
 
-set -ex
+set -x
```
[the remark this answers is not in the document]
davidjumani: When grepping for changelog, if it finds none, the grep command fails and this causes the entire script to fail. I'll try and skip fast failing around that bit of code instead of the entire script

bewebi (approved):
LGTM pending Fabian’s comment

davidjumani:
Thanks @inFocus7 I'll look into the script in other repos once this has been running for a while
I've updated the description with links to the prior failures
The doc build action runs this script on push

ci/github-actions/process-skip-directives/script.sh
```diff
@@ -22,19 +22,25 @@ skipDocsBuildDirective="skipCI-docs-build:true"
...
-        shouldSkipKubeTests=true
-    fi
-    if [[ $(cat $changelog | grep $skipDocsBuildDirective) ]]; then
-        shouldSkipDocsBuild=true
+if [ ! -z "$githubBaseRef" ]; then
+    # If there is no changelog found, the grep command fails and in turn the entire script exits since the error on exit flag has been set
+    # To avoid that, unsetting it before this command and setting it after
+    set +e
+    changelog=$(git diff origin/$githubBaseRef HEAD --name-only | grep "changelog/")
+    set -e
```
inFocus7, line 30:
```suggestion
    changelog=$(git diff origin/$githubBaseRef HEAD --name-only | grep "changelog/" || true)
```

inFocus7: Tested locally by doing
```
set -ex
changelog=$(git diff origin/stability/generalize-eks-scale-tests HEAD --name-only | grep "changelog/s")
```

exited/failed.

Then
```
set -ex
changelog=$(git diff origin/stability/generalize-eks-scale-tests HEAD --name-only | grep "changelog/s" || true)
```

and it did not exit, and returned
```
++ git diff origin/stability/generalize-eks-scale-tests HEAD --name-only
++ grep changelog/s
++ true
+ changelog=
```

which i believe means that changelog is empty, which i think your added `[ ! -z "$changelog" ]` handles.

ci/github-actions/process-skip-directives/script.sh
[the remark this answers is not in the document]
inFocus7: Gotcha thanks for clarity! Closing this, since I added another suggestion/comment [here](https://github.com/solo-io/gloo/pull/8597#discussion_r1298644168)!

ci/github-actions/process-skip-directives/script.sh
davidjumani: I did try using || true but couldn't get it to work! Thanks!

sam-heilbron (commented):
Perhaps a simpler approach to this is to guard against an empty githubBaseRef, and return early if one is found. I think a comment indicating _why_ an emptry baseRef would occur and why we would not have skip directives in that case, would be clearer to future devs. 

Also, it would allow us to not mess with error handling, since our expectation could then be:
- if the ref is provided, we're running on a PR, and want to error loudly if the script fails

davidjumani:
> Perhaps a simpler approach to this is to guard against an empty githubBaseRef, and return early if one is found. I think a comment indicating _why_ an emptry baseRef would occur and why we would not have skip directives in that case, would be clearer to future devs.

The check for an empty githubBaseRef is added over at https://github.com/solo-io/gloo/pull/8597/files#diff-8b602e8c4f1b1eba2ceddf17430086ca289b08041cc218308fd870a4d29579daR25
Just didn't comment on why it is there as there might be more than once reason why it is not provided (perhaps some we haven't come across yet)

sam-heilbron (approved):

inFocus7 (approved):
🚀

Issue #8445, opened 2023-07-07
Pod disruption budget is not exposed in enterprise helm chart values for Gloo Pod 


AkshayAdsul:
### Version

None

### Is your feature request related to a problem? Please describe.

Pod disruption budget is not exposed in enterprise helm chart values for Gloo Pod .
Need to add this.

### Describe the solution you'd like

Currently we do not have a way to set the PDB for the Gloo Pod from the helm charts. We would like a way to expose this.

bewebi:
The issue was closed prematurely

The PR has merged for this feature and will be released in v1.16.0-beta5, however backports to 1.15 and 1.14 have not been completed

bewebi:
Backports have now been merged

The new values will be available in v1.16.0-beta5, v1.15.2, and v1.14.19

PR #8600, opened 2023-08-18
Gloo PDB

bewebi:
# Description

Add `PodDisruptionBudget` to `gloo` helm values

## Helm changes
- Update PodDisruptionBudget fields in values.go to have more relevant descriptions which reference k8s documentation and to have string type, which better represents the k8s IntOrString type
  - Update existing Helm tests for GatewayProxy PDB accordingly
- Add support for `gloo.podDisruptionBudget.minAvailable` and `gloo.podDisruptionBudget.maxUnavailable` values
  - If `gloo.podDisruptionBudget` is specified, a PDB is created with those settings which selects on the `gloo` deployment
  - If both are specified, rendering fails, as those values are mutually exclusive in PDBs
  - Add helm tests, including for the mutually exclusive failure case

# Context

It is not currently possible to configure a PDB for the Gloo controller deployment via Helm

## Interesting decisions

The template is based on that which we already have for the Gateway Proxy [here](https://github.com/solo-io/gloo/blob/a3e5552d9164bb2ca84f0ca6e8b721e72632cbb9/install/helm/gloo/templates/8-gateway-proxy-pod-disruption-budget.yaml)
The template for the Gloo deployment is much simpler because there is only ever one deployment and we don't need to specialize logic to handle PDBs associated with different dynamically-named deployments

As with that template, there is an edge case whereby a user could specify `gloo.podDisruptionBudget` without specifying one of the two accepted values (ie `minAvailable` and `maxUnavailable`)
In this case a PDB will be created for the Gloo deployment without any settings, which appears to always have a `False` status with `InsufficientPods` as the reason
I wasn't able to determine what ramifications this might have for the normal operation of the Gloo deployment (see below under Testing steps)
If we'd prefer to be more strict about requiring a valid setting to present in order to create the PDB I'll fix it in both locations

Though the `PodDisruptionBudget` object in `values.go` includes `KubeResourceOverride`, this functionality is not relevant for the Gloo PDB
A line is [generated in docs](https://github.com/solo-io/gloo/pull/8600/files#diff-2baf5c822ecb28b315e303c9d28c5c3a6486b975c91e7ba2d19d7eb05966c882R235) describing this value
I do not expect users to be too confused by this, however if reviewers would prefer, I can create a new `PodDisruptionBudget` struct that omits `KubeResourceOverride` 

It may be overly verbose/redundant to refer to values by their full reference (eg `.Values.gloo.podDisruptionBudget.minAvailable`) every time, however I find this more readable and there isn't much benefit to assigning local variables in such a small and straightforward template

Some improvements have been made to the paradigm set by the Gateway Proxy PDB however they have not been propagated to that existing PDB template in order to keep the scope narrow
I am open-minded if reviewers feel that it is worthwhile to propagate these changes despite scope creep

## Testing steps

I used `helm template` commands to confirm that the output looked right, for example:
```
$ helm template install/helm/gloo --set gloo.podDisruptionBudget.maxUnavailable=3 > template.yaml
```

I subsequently did a local `helm install` and confirmed that the generated PDB correctly referenced the Gloo Deployment
```
$ helm install gloo install/helm/gloo --set gloo.podDisruptionBudget.maxUnavailable=3
```

This results in the following PDB:
```
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
...
  name: gloo-pdb
  namespace: default
...
spec:
  maxUnavailable: 3
  selector:
    matchLabels:
      gloo: gloo
status:
  conditions:
  - lastTransitionTime: "2023-08-21T18:14:04Z"
    message: ""
    observedGeneration: 1
    reason: SufficientPods
    status: "True"
    type: DisruptionAllowed
  currentHealthy: 1
  desiredHealthy: 0
  disruptionsAllowed: 1
  expectedPods: 1
  observedGeneration: 1
```
Note the `currentHealthy: 1` field in the status, indicating that the deployment's pods are being watched correctly
Even with a PDB, manual deletes of pods are always allowed, and I wasn't able to determine another method for testing a PDB, however it is sufficient to verify correctness of the Helm values to observe that a PDB with the correct settings is correctly associated with the Gloo deployment
 
## Notes for reviewers

Feel free to pull the branch and verify locally with either `helm template` or `helm install` as I have done

I welcome feedback on the design decisions described above

Note that though the Helm tests required updating to explicitly pass values as strings, in practice it still works to pass the values as ints (ie without quotes)

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8445


docs/content/reference/values.txt
```diff
@@ -230,6 +230,9 @@
 |gloo.logLevel|string||Level at which the pod should log. Options include "info", "debug", "warn", "error", "panic" and "fatal". Default level is info|
 |gloo.disableLeaderElection|bool||Set to true to disable leader election, and ensure all running replicas are considered the leader. Do not enable this with multiple replicas of Gloo|
 |gloo.headerSecretRefNsMatchesUs|bool||Set to true to require that secrets sent in headers via headerSecretRefs come from the same namespace as the destination upstream. Default: false|
+|gloo.podDisruptionBudget.minAvailable|int32||An eviction is allowed if at least "minAvailable" pods selected by "selector" will still be available after the eviction, i.e. even in the absence of the evicted pod. So for example you can prevent all voluntary evictions by specifying "100%".|
```
inFocus7, line 233: Q:
> by specifying "100%"
Technically the user would just set this to `100` without the percentage right? 👀

install/helm/gloo/templates/1-gloo-pod-disruption-budget.yaml
inFocus7, line 1: Q: How do we know what ordering/number prefix to use for our helm templates? Like what makes this a `1-` instead of a different number?

install/test/helm_test.go
```diff
@@ -205,6 +205,60 @@ var _ = Describe("Helm Test", func() {
 				}
 			})
 
+			It("gloo pdb disabled by default", func() {
```
inFocus7, line 208: nit: This reads weird "It gloo pdb disabled by default", although not sure a better structure.

Maybe something like this?
```
Context("gloo pdb") {
   It("is disabled by default")
   It("can create gloo pdb with ...")
}
```

 Or just a different name like `It("does not create gloo pdb by default") {}`

install/test/helm_test.go
```diff
@@ -205,6 +205,60 @@ var _ = Describe("Helm Test", func() {
...
+  minAvailable: 2
+  selector:
+    matchLabels:
+      gloo: gloo
+`)
+
+				testManifest.ExpectUnstructured("PodDisruptionBudget", namespace, "gloo-pdb").To(BeEquivalentTo(pdb))
+			})
+
+			It("can create gloo pdb with maxUnavailable", func() {
```
inFocus7, line 238: What happens when both are set? Is that testable? Or is the order of
```
  minAvailable: X
  maxUnavailable: Y
```

different each render?

docs/content/reference/values.txt
bewebi: Without `%` the value will be interpreted as an integer (ie 100 pods must be available)

install/helm/gloo/templates/1-gloo-pod-disruption-budget.yaml
bewebi: I thought the answer was as simple as "they are logical groupings of CRs that handle the same piece of functionality"

That's the intent here, where `1` is the grouping for the Gloo controller deployment and the PDB that exclusively selects on the Deployment's pods, however when looking into it in response to your question, I'm not sure how much consistency there is there

Regardless, it seems to be an internal convention rather than a best practice Helm itself defines, though if we have one written down for this I haven't found it

Perhaps @sam-heilbron or @nfuden would have more insight?

install/test/helm_test.go
bewebi: Good callout, one I frequently make on other PRs
I copied the tests from the Gateway Proxy PDB tests with minimal modification, hence the naming that doesn't meet my usual standard...will fix!

install/test/helm_test.go
bewebi: This would create a PDB CR like
```
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: gloo-pdb
  namespace: default
spec:
  minAvailable: 2
  maxUnavailable: 3
  selector:
    matchLabels:
      gloo: gloo
```
however this is invalid
The user will get an error like:
```
Error: PodDisruptionBudget.policy "gloo-pdb" is invalid: spec: Invalid value: policy.PodDisruptionBudgetSpec{MinAvailable:(*intstr.IntOrString)(0xc007b82060), Selector:(*v1.LabelSelector)(0xc007b82080), MaxUnavailable:(*intstr.IntOrString)(0xc007b82040), UnhealthyPodEvictionPolicy:(*policy.UnhealthyPodEvictionPolicyType)(nil)}: minAvailable and maxUnavailable cannot be both set
```
and the Helm install will fail

This is not a case I'd experimented with

I'm pushing a logic check within the template to enforce their mutual exclusivity prior to attempting to apply the CR, which is hopefully a better UX

bewebi: I've added a test for this as well, slightly clunkily adapting the existing structure for tests that expect there not to be errors on render failure

docs/content/reference/values.txt
inFocus7: 👍🏼 In that case, shouldn't this be an IntOrString(?) type to account for "XYZ%"? Can't remember if that's a type

bewebi: Though that's the type we map to in k8s, it's not one we support in our generated Helm docs

The documentation we have is directly copied from k8s reference, but isn't actually the most descriptive for our case

I think the path forward is to expose it as a `string`, which is the most accurate for the Helm value, and update the description to better explain the usage

install/helm/gloo/generate/values.go
```diff
@@ -240,6 +240,7 @@ type Gloo struct {
 	LogLevel                   *string               `json:"logLevel,omitempty" desc:"Level at which the pod should log. Options include \"info\", \"debug\", \"warn\", \"error\", \"panic\" and \"fatal\". Default level is info"`
 	DisableLeaderElection      *bool                 `json:"disableLeaderElection,omitempty" desc:"Set to true to disable leader election, and ensure all running replicas are considered the leader. Do not enable this with multiple replicas of Gloo"`
 	HeaderSecretRefNsMatchesUs *bool                 `json:"headerSecretRefNsMatchesUs,omitempty" desc:"Set to true to require that secrets sent in headers via headerSecretRefs come from the same namespace as the destination upstream. Default: false"`
+	PodDisruptionBudget        *PodDisruptionBudget  `json:"podDisruptionBudget,omitempty"`
```
ashishb-solo, line 243: pardon my ignorance, but why is there no description here?

docs/content/reference/values.txt
ashishb-solo, line 233: So if I'm understanding the description here correctly, the allowed values are [0-100] (inclusive), correct? What happens if the value is outside this range?

install/test/helm_test.go
```diff
@@ -205,6 +213,73 @@ var _ = Describe("Helm Test", func() {
...
+					valuesArgs: []string{
+						"gloo.podDisruptionBudget.maxUnavailable=2",
+						"gloo.podDisruptionBudget.minAvailable=2",
+					},
+				})
+
+				Expect(renderErr).NotTo(BeNil())
+				Expect(renderErr.Error()).To(ContainSubstring("gloo PDB values minAvailable and maxUnavailable are mutually exclusive"))
+			})
+
```
ashishb-solo, line 282: Based on [this question](https://github.com/solo-io/gloo/pull/8600/files#r1301976011), we may need to add more tests here to assert that the behaviour is correct

install/helm/gloo/generate/values.go
bewebi: Initially because I was just getting it working and didn't want to worry about it on the first pass

But actually the description isn't propagated to docs for types that are structs with fields of their own, so there is no user-facing benefit, and it seems self explanatory for any developer viewing `values.go`

docs/content/reference/values.txt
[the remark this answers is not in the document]
bewebi: I've now updated the description for these values to link to the k8s documentation directly
Beyond that I suppose we _could_ try and do a better job explaining what these fields do than k8s docs, but it's not really in scope IMO

Of course if neither field is set, no PDB will be created

bewebi: any positive integer is a valid input, as is `[0-100]%`

If a bad value is set, the helm install will fail when the PDB is applied, with something like:
```
Error: PodDisruptionBudget.policy "gloo-pdb" is invalid: spec.maxUnavailable: Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:"300%"}: must not be greater than 100%
```

install/test/helm_test.go
bewebi: Because the error stems from k8s validation and not helm, it's not something for us to add a test for

The manifest will correctly render with any value and only error once applied

I don't think we need to be responsible for enforcing k8s validation, especially since those requirements could theoretically change
By that token we may not even want to enforce the mutual exclusivity, but that's simple enough to do without resorting to [schema validation](https://helm.sh/docs/topics/charts/#schema-files) which is not something we currently do for our values

ashishb-solo: Agreed, I definitely don't think we need to revalidate existing kubernetes behaviour, I just wanted to make sure this angle is covered

install/helm/gloo/generate/values.go
```diff
@@ -493,8 +494,8 @@ type HorizontalPodAutoscaler struct {
 }
 
 type PodDisruptionBudget struct {
-	MinAvailable   *int32 `json:"minAvailable,omitempty" desc:"An eviction is allowed if at least \"minAvailable\" pods selected by \"selector\" will still be available after the eviction, i.e. even in the absence of the evicted pod. So for example you can prevent all voluntary evictions by specifying \"100%\"."`
-	MaxUnavailable *int32 `json:"maxUnavailable,omitempty" desc:"An eviction is allowed if at most \"maxUnavailable\" pods selected by \"selector\" are unavailable after the eviction, i.e. even in absence of the evicted pod. For example, one can prevent all voluntary evictions by specifying 0. This is a mutually exclusive setting with \"minAvailable\"."`
+	MinAvailable   *string `json:"minAvailable,omitempty" desc:"Corresponds directly with the _minAvailable_ field in the [PodDisruptionBudgetSpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-disruption-budget-v1/#PodDisruptionBudgetSpec). This value is mutually exclusive with _maxUnavailable_."`
```
ashishb-solo, line 497: The linked documentation specifies that these fields are `IntOrString` - is it possible for us to use the same type here?

bewebi: I had dismissed the possibility because it's not a valid type for Helm values, however we _can_ use the k8s `intstr` package type `IntOrString` in values.go

The generated docs continue to list the values as `string` type, however the Helm tests then allow either type

I've pushed a commit that makes the change and adds a Helm test to demonstrate that either type is accepted, though this is basically just internal signaling about what we expect, and not user-facing

ashishb-solo: I don't think it's a major issue though as long as the documentation is visible to the user somewhere. As long as it's visible somewhere to the user, I can ignore this and approve

bewebi: It is what's expected based on the same principle mentioned [here](https://github.com/solo-io/gloo/pull/8600#discussion_r1301987189) whereby our generated docs contain only "leaf" fields and their descriptions
If any type is not a base Go type it will be ignored in favor of its children

Therefore using the `IntOrString` type results in that type's fields being exposed instead of the field using `IntOrString`

This is actually what I expected, but my locally running instance of the Docs site didn't refresh promptly (and I didn't check the diff in values.txt directly) so I didn't notice the change and assumed there must be some magic that made this a special case

The commit will be reverted

ashishb-solo (dismissed):

install/helm/gloo/generate/values.go
ashishb-solo: sounds good to me, thanks for taking the extra time to look into this!

ben-taussig-solo, line 497: Is converting `MinAvailable` and `MaxUnavailable` from int32 to string a breaking change? I saw some discussion about this in the PR description, so I'll go back there

ben-taussig-solo: `Note that though the Helm tests required updating to explicitly pass values as strings, in practice it still works to pass the values as ints (ie without quotes)`

Is this still the case if helm values are defined in a file? Just trying to be careful that we don't break existing PDB config

ben-taussig-solo:
> Though the PodDisruptionBudget object in values.go includes KubeResourceOverride, this functionality is not relevant for the Gloo PDB
> A line is generated in docs describing this value

Could you point me to this line?

install/helm/gloo/generate/values.go
bewebi: Appreciate the call out

I had only tested it with in-line `--set` flags previously
I've not confirmed that it also works as expected when using a `values.yaml` file

Ultimately Helm is not governed by our `values.go` file, which really only serves as a source of truth for our own documentation and tests, and will accept any values passed regardless of what we've described/documented there

bewebi:
@ben-taussig-solo [this](https://github.com/solo-io/gloo/pull/8600/files#diff-2baf5c822ecb28b315e303c9d28c5c3a6486b975c91e7ba2d19d7eb05966c882R235) is the line in question
I've also updated the PR description note to contain a link

ben-taussig-solo:
> @ben-taussig-solo [this](https://github.com/solo-io/gloo/pull/8600/files#diff-2baf5c822ecb28b315e303c9d28c5c3a6486b975c91e7ba2d19d7eb05966c882R235) is the line in question I've also updated the PR description note to contain a link

Ah, I see -- I had misinterpreted the line in the description.
Could you create a new PodDisruptionBudget struct that omits KubeResourceOverride? I was confused by the presence and lack of support for the setting when I originally read through the PR

install/helm/gloo/generate/values.go
```diff
@@ -493,8 +494,8 @@ type HorizontalPodAutoscaler struct {
...
+	MaxUnavailable *string `json:"maxUnavailable,omitempty" desc:"Corresponds directly with the _maxUnavailable_ field in the [PodDisruptionBudgetSpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-disruption-budget-v1/#PodDisruptionBudgetSpec). This value is mutually exclusive with _minAvailable_."`
 	*KubeResourceOverride
 }
```
nfuden, line 500: Should we also be adding the last field from upstream? https://github.com/kubernetes/kubernetes/commit/a429797f2e84adf5582d3d30d23c9fcfce2b66d8

bewebi:
@ben-taussig-solo I think [this commit](https://github.com/solo-io/gloo/pull/8600/commits/32e60b8b0fd908aba1174d8523123261f77865ae) takes care of that fairly cleanly, despite the aggressive diff

ben-taussig-solo (approved):

install/helm/gloo/generate/values.go
bewebi: As the original issue suggests that we'd like to backport this to 1.14, where we don't even support k8s 1.26, so I don't think it's strictly necessary

On the other hand, AFAICT, we'd only get a warning when applying a PDB with an unknown field
We could call out the k8s version limitation in the description/docs, and even omit it from the 1.14 backport

I ultimately end up ambivalent about whether to put in the effort at this point, happy to defer to your opinion and/or that of Product if it's worth checking in with them

nfuden: Yeah fair enough lets not worry about it

nfuden (approved):

PR #8605, opened 2023-08-22
Handle STS Credentials region

ben-taussig-solo:
# Description
 - Along with the changes in https://github.com/solo-io/envoy-gloo/pull/267, fixes a bug where assuming a secondary role using an EKS ServiceAccount would fail when the service account was not in us-east-1 in some instances


projects/gloo/api/external/envoy/extensions/aws/filter.proto
```diff
@@ -116,6 +116,8 @@ message AWSLambdaConfig {
     string uri = 2 [ (validate.rules).string.min_bytes = 1 ];
     // timeout for the request
     google.protobuf.Duration timeout = 3;
+    // Region for the sts endpoint
+    string region = 4 [ (validate.rules).string.min_bytes = 1 ];
```
ben-taussig-solo, line 120: I am not sure at the moment if `region` be required or not.

I think it's safe to assume that anyone who upgrades to a new version and has `serviceAccountCredentials` configured [via helm](https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/18-settings.yaml#L50-L58) will have a valid configuration after updating to a new version. However, making `region` required would break users who have manually configured `serviceAccountCredentials` after an upgrade. For that reason, I'm leaning towards making this optional, but looking for input before I commit to anything

nfuden: I think this should be optional as aws itself has a default us-east-1 region

ben-taussig-solo: Resolved here: https://github.com/solo-io/gloo/pull/8605/commits/f52cad4dbb7cab17b8ceb43ae8be89fe3c047269

test/e2e/aws_test.go
```diff
@@ -783,29 +793,61 @@ var _ = Describe("AWS Lambda", func() {
...
+					setupEnvoySts(true, secondaryRegion)
+					addCredentialsSts()
+					addUpstreamSts(secondaryRegion)
+				})
 
-			It("should be able to call lambda with response transform", testProxyWithResponseTransform)
+				It("should be able to call lambda", testProxy)
 
-			It("should be able to call lambda with request transform", testProxyWithRequestTransform)
+				It("should be able to call lambda with response transform", testProxyWithResponseTransform)
```
nfuden, line 824: does it make sense for us to do response transform again in secondary?

ben-taussig-solo: Fair point, really the only interesting tests in the secondary and default regions are `should be able to call lambda`. I think it's probably clearer what is going on (plus marginally better for e2e test performance) if we only include those

ben-taussig-solo: Removed those here: https://github.com/solo-io/gloo/pull/8605/commits/776da72fe472e02b4d4f9c3c0d531f38d856c69c

ashishb-solo (approved):

ben-taussig-solo:
/kick failure in consul plugin tests, presumably a flake: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=aaa1a4e9-cf13-457f-8478-17ec2c9dd0b1

nfuden (approved):
Looks good!

Makefile
```diff
@@ -595,8 +595,7 @@ package-chart: generate-helm-files
 # https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC59
 git_tag = $(shell git describe --abbrev=0 --tags)
 # Semantic versioning format https://semver.org/
-# Regex copied from: https://github.com/solo-io/go-utils/blob/16d4d94e4e5f182ca8c10c5823df303087879dea/versionutils/version.go#L338
-tag_regex := v[0-9]+[.][0-9]+[.][0-9]+(-[a-z]+)*(-[a-z]+[0-9]*)?$
+tag_regex := ^v([0-9]{1,}\.){2}[0-9]{1,}$
```
jenshu, line 598: looks like you accidentally reverted this change from https://github.com/solo-io/gloo/pull/8653

PR #8606, opened 2023-08-22
Docs: version bumps for 1.15 release

Rachael-Graham:
Bumps active versions and version conrefs in the docs for 1.15.0 release.


docs/active_versions.json
```diff
@@ -1,12 +1,12 @@
 {
-  "latest": "v1.14.x",
+  "latest": "v1.15.x",
```
sam-heilbron, line 2: Let's keep 1.14 as the default "latest" for now, and we will likely switch this in a week/two

sam-heilbron (dismissed):

sam-heilbron (approved):

docs/active_versions.json
```diff
@@ -2,11 +2,11 @@
   "latest": "v1.14.x",
```
artberger, line 2: Should this be 1.15.x now?

docs/active_versions.json
```diff
@@ -2,11 +2,11 @@
...
   "versions": [
     "main",
+    "v1.15.x",
     "v1.14.x"
```
artberger, line 6: Should 1.14.x be moved to "oldVersions"?

Rachael-Graham: Threw an error when I did, so I have to keep it in here as long as it is latest still

docs/active_versions.json
Rachael-Graham: Request from Sam to keep it as 1.14 for now

artberger (approved):

Issue #6076, opened 2022-03-11
Expose local rate limit filter in gloo


jmunozro:
### Version

1.10.x (latest stable)

### Is your feature request related to a problem? Please describe.

A found a case with a customer that wants to implement RL based on capacity, not time.

Having a token approach is preferable on some use cases because it allows us to handle bursts regardless of how fast they show up, while still preventing resource exhaustion. We can then tune how much we let through, and how much we fail, unlike the normal rate limit which is a set amount per time unit.

### Describe the solution you'd like

I would like to have [Local rate limit](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#local-rate-limit) exposed in gloo.

bcollard:
got a new demand for this capability, where the client is doing some sort of reverse exponential back-off and overflowing the gateway. Having a local rate-limiter would help protect the proxy and upstream services.
+1 for exposing the Envoy local rate-limiter.

guydc:
Local rate limit is a resource-efficient and performant mechanism for protection against client misbehavior (as opposed to global rate limiting with all of its requirements). 
I'm also interested in this capability.

davidjumani:
Added as part of OSS gloo [v1.16.0-beta8](https://github.com/solo-io/gloo/releases/tag/v1.16.0-beta8)
Merged into enterprise and will be released as part of v1.16.0

PR #8618, opened 2023-08-24
feat: Add support for envoy local rate limit filters 

davidjumani:
# Description

This adds support for envoy's [local rate limit filter](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/local_rate_limiting#arch-overview-local-rate-limit)
This includes the :
- [Network Filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/local_rate_limit_filter#config-network-filters-local-rate-limit)
- [HTTP Filter
](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/local_rate_limit_filter#config-http-filters-local-rate-limit)

This allows users to limit requests without the need for an external RL service to save resources
This is initially course grained RL without the support for actions and descriptors and is executed pre-auth

## API changes
- Added `httpLocalRatelimit` and `l4LocalRatelimit` to the httpGateway and TCP gateway.options. This defines the optional default token bucket. It also specifies the enableXRatelimitHeaders and localRateLimitPerDownstreamConnection that is inherited by all vhost and route specific local RL
```
httpGateway :
.............
     httpLocalRatelimit:
       defaults:
         fillInterval: 15s
         maxTokens: 5
         tokensPerFill: 5
        enableXRatelimitHeaders: true
    l4LocalRatelimit:
       fillInterval: 15s
       maxTokens: 5
       tokensPerFill: 5
```
- Added `local_ratelimit` to the vhost and route `options.ratelimit` fields that specifies a token bucket for that specific vhost and route and overrides the existing default
```
virtualHost :
............
  options:
    ratelimit:
      localRatelimit:
        fillInterval: 15s
        maxTokens: 3
        tokensPerFill: 3

```

# Context

Feature request in https://github.com/solo-io/gloo/issues/7428 && https://github.com/solo-io/gloo/issues/6076

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7428


test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+			ExpectWithOffset(1, response).Should(matchers.HaveOkResponse())
+			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "The connection should not be rate limited")
+		}
+
+		expectRateLimitedWithXRateLimitHeader = func(limit int) {
+			defer GinkgoRecover()
+			response, _ := httpClient.Do(requestBuilder.Build())
+			// fmt.Println(response)
+			ExpectWithOffset(1, response).To(matchers.ContainHeaders(http.Header{
+				"X-Ratelimit-Limit":     []string{fmt.Sprint(limit)},
```
nfuden, line 61: should these be normalized to lower case?

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+			defer GinkgoRecover()
+			response, err := httpClient.Do(requestBuilder.Build())
+			// fmt.Println(response)
+			ExpectWithOffset(1, response).Should(matchers.HaveOkResponse())
+			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "The connection should not be rate limited")
+		}
+
+		expectRateLimitedWithXRateLimitHeader = func(limit int) {
+			defer GinkgoRecover()
+			response, _ := httpClient.Do(requestBuilder.Build())
```
nfuden, line 58: nit: check the error

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+	JustBeforeEach(func() {
+		testContext.JustBeforeEach()
+	})
+
+	JustAfterEach(func() {
+		testContext.JustAfterEach()
+	})
+
+	Context("Filter not defined", func() {
```
nfuden, line 84: filter not configured?

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+		It("Should not rate limit", func() {
```
nfuden, line 85: This seems to imply we actually check for rate limit, can we update this to show that we just check that weve configured it?

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+	// 			// 	},
+	// 			// },
+	// 		}
+
+	// 		testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+	// 			gw,
+	// 		}
+	// 	})
+
+	// 	// TODO : Investigate this failure - either the test or the filter itself
```
nfuden, line 128: what failure was this? Is this still needed?

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+	// 		expectSuccess()
+	// 		expectSuccess()
+	// 		expectRateLimited()
+	// 	})
+
+	// })
+
+	Context("HTTP Local Rate Limit", func() {
+		Context("Overrides the default", func() {
```
nfuden, line 151: which default is being overriden?

Is this more of: ?
```suggestion
		Context("sets local rate limit on default gateway", func() {
```

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,344 @@
...
+	)
+
+	BeforeEach(func() {
+		testContext = testContextFactory.NewTestContext()
+		testContext.BeforeEach()
+
+		httpClient = testutils.DefaultHttpClient
+		requestBuilder = testContext.GetHttpRequestBuilder()
+
+		expectSuccess = func() {
```
nfuden, line 48: can we rename this?  when i read this function i expect that we are succeeeding on whats tested but what we are really doing here is expected an ok non-ratelimited response

Perhaps 
```suggestion
		expectNotLimited = func() {
```

test/e2e/local_ratelimit_test.go
davidjumani: We are checking that the filter is not present in the envoy config, as well as no x-ratelimit headers are present in the response

test/e2e/local_ratelimit_test.go
davidjumani: Yes, kept this as a TODO to fix before merge as I'm not able to test this consistently

test/e2e/local_ratelimit_test.go
davidjumani: Makes sense, fixed

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+	}, nil
+}
+
+func (p *plugin) NetworkFiltersHTTP(params plugins.Params, listener *v1.HttpListener) ([]plugins.StagedNetworkFilter, error) {
+	return generateNetworkFilter(listener.GetOptions().GetL4LocalRatelimit())
+}
+
+func (p *plugin) NetworkFiltersTCP(params plugins.Params, listener *v1.TcpListener) ([]plugins.StagedNetworkFilter, error) {
+	return generateNetworkFilter(listener.GetOptions().GetLocalRatelimit())
+}
```
jenshu, line 128: are these reversed? should the NetworkFiltersHTTP be returning localRateLimit and NetworkFiltersTCP be returning  L4 ratelimit?

projects/gloo/api/v1/options.proto
```diff
@@ -197,6 +198,15 @@ message HttpListenerOptions {
     // ConnectionLimit can be used to limit the number of active connections per gateway. Useful for resource protection as well as DoS prevention.
     connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 29;
 
+    // L4LocalRatelimit can be used to rate limit the connections per gateway at the L4 layer.
+    // It uses envoy's own local rate limit filter to do so, wihtout the need for an external rate limit server to be set up
+    local_ratelimit.options.gloo.solo.io.TokenBucket l4_local_ratelimit = 32;
```
jenshu, line 203: nit: i know the envoy docs refer to it as such, but the name `l4_local_ratelimit` seems a little awkward. would either network_local_ratelimit or tcp_local_ratelimit be a better name?

projects/gloo/api/v1/options.proto
```diff
@@ -197,6 +198,15 @@ message HttpListenerOptions {
...
+
+    // HttpLocalRatelimit can be used to rate limit the number of requests per gateway.
+    // Unlike the L4LocalRatelimit, this works as part of the HCM and limits requests pre-auth.
+    // It uses envoy's own local rate limit filter to do so, wihtout the need for an external rate limit server to be set up
```
jenshu, line 207:
```suggestion
    // It uses envoy's own local rate limit filter to do so, without the need for an external rate limit server to be set up.
```

projects/gloo/api/v1/options.proto
```diff
@@ -197,6 +198,15 @@ message HttpListenerOptions {
     // ConnectionLimit can be used to limit the number of active connections per gateway. Useful for resource protection as well as DoS prevention.
     connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 29;
 
+    // L4LocalRatelimit can be used to rate limit the connections per gateway at the L4 layer.
+    // It uses envoy's own local rate limit filter to do so, wihtout the need for an external rate limit server to be set up
```
jenshu, line 202:
```suggestion
    // It uses envoy's own local rate limit filter to do so, without the need for an external rate limit server to be set up.
```

projects/gloo/api/v1/options.proto
```diff
@@ -208,6 +218,10 @@ message TcpListenerOptions {
 
     // ConnectionLimit can be used to limit the number of active connections per gateway. Useful for resource protection as well as DoS prevention.
     connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 4;
+
+    // LocalRatelimit can be used to rate limit the connections per gateway at the L4 layer.
+    // It uses envoy's own local rate limit filter to do so, wihtout the need for an external rate limit server to be set up
```
jenshu, line 223:
```suggestion
    // It uses envoy's own local rate limit filter to do so, without the need for an external rate limit server to be set up.
```

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
```diff
@@ -0,0 +1,41 @@
...
+    // Must be greater than zero.
+    google.protobuf.UInt32Value tokens_per_fill = 2;
+
+    // The fill interval that tokens are added to the bucket. During each fill interval tokens_per_fill are added to the bucket.
+    // The bucket will never contain more than max_tokens tokens.
+    // Defaults to 1 second.
+    google.protobuf.Duration  fill_interval = 3;
+}
+
+message Settings {
```
jenshu, line 34: can you add comments for this message and fields? 
where would `Settings` vs. `TokenBucket` be used?

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+
+	fillInterval := localRatelimit.GetFillInterval()
+	if fillInterval == nil {
+		fillInterval = &durationpb.Duration{
+			Seconds: 1,
+		}
+	}
+
+	maxTokens := localRatelimit.GetMaxTokens()
+	if maxTokens < 1 {
```
jenshu, line 81: does this need to be > 1 or >= 1?
the proto docs above say "Must be greater than one." which is different than the check here

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: `NetworkFiltersTCP` applies the network filter for a TCP Gateway and `NetworkFiltersHTTP` does the same for the HTTPGateway

projects/gloo/api/v1/options.proto
davidjumani: I'm hesitant to use tcp_local_ratelimit in the options for the HTTPGateway as it might be misleading. I'll change it to network_local_ratelimit instead

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: Mistake in the docs. It should be greater than or equal to one, thanks

projects/gloo/api/v1/options.proto
```diff
@@ -197,6 +198,15 @@ message HttpListenerOptions {
     // ConnectionLimit can be used to limit the number of active connections per gateway. Useful for resource protection as well as DoS prevention.
     connection_limit.options.gloo.solo.io.ConnectionLimit connection_limit = 29;
 
+    // NetworkLocalRatelimit can be used to rate limit the connections per gateway at the L4 layer.
+    // It uses envoy's own local rate limit filter to do so, without the need for an external rate limit server to be set up.
+    local_ratelimit.options.gloo.solo.io.TokenBucket network_local_ratelimit = 32;
+
+    // HttpLocalRatelimit can be used to rate limit the number of requests per gateway.
+    // Unlike the NetworkLocalRatelimit, this works as part of the HCM and limits requests pre-auth.
```
jbohanon, line 206: Isn't the network local rate limit also pre-auth/pre-filterchain? This line kinda makes it sound like it's not

davidjumani: It is part of the http filter chain ie: L7 while the network filter chain works at L4, but good point, I'll update it to make it clear

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+
+	BeforeEach(func() {
+		testContext = testContextFactory.NewTestContext()
+		testContext.BeforeEach()
+
+		httpClient = testutils.DefaultHttpClient
+		requestBuilder = testContext.GetHttpRequestBuilder()
+
+		expectNotRateLimited := func() *http.Response {
+			defer GinkgoRecover()
```
sam-heilbron, line 50: Why is this required? I had thought this was only used when invoking code that could panic inside of a separate goroutine?

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+			ExpectWithOffset(1, response).To(matchers.HaveOkResponse())
+			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "The connection should not be rate limited")
+			return response
+		}
+
+		expectNotRateLimitedWithOutXRateLimitHeader = func() {
+			defer GinkgoRecover()
+			response := expectNotRateLimited()
+			// Since the values of the x-rate-limit headers change with time, we only check the presence of these header keys and not match their value
+			ExpectWithOffset(2, response).ToNot(matchers.ContainHeaderKeys([]string{"x-ratelimit-reset", "x-ratelimit-limit", "x-ratelimit-remaining"}),
```
sam-heilbron, line 61: Is the offset of 2 correct? This is invoked from a test directly, so I would have expected an offset of 1

sam-heilbron, line 61: nit: I think we can combine these assertions into a single one:
```go
Expect(response).To(matchers.HaveHttpResponse(&matchers.HttpResponse{
StatusCode: 200,
Custom:  NotTo(ConatinerHeaderKeys)...)
}
```

https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_http_response.go#L79

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+		testContext.JustBeforeEach()
+	})
+
+	JustAfterEach(func() {
+		testContext.JustAfterEach()
+	})
+
+	Context("Filter not configured", func() {
+		It("Should not add the filter to the list of HCM filters", func() {
+			// Since the filter is not configured in gloo, the filter should not be present in the envoy config, and requests should not be rate limited
```
sam-heilbron, line 107: nit: Since this is based on the functionality not of default gloo edge, but on the `removeUnusedFilters` setting in our API, it may be worthwhile to include that in the comment

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+			testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+				gw,
+			}
+		})
+
+		It("Should rate limit at the l4 layer", func() {
+			expectNotRateLimited := func() {
+				defer GinkgoRecover()
+
+				// We use a new client every time as TCP connections are cached and this needs to be avoided in order to test L4 rate limiting
```
sam-heilbron, line 142: nit: It feels like this is a great find during investigation, and I like this comment here. What is the value that you see have using the default client for the other tests? I imagine at the top of the file, in the beforeEach we could setup a new client each time, with a comment outlining why we do that for these tests

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+					},
+				}
+
+				testContext.ResourcesToCreate().Gateways = v1.GatewayList{
+					gw,
+				}
+			})
+
+			It("Should rate limit the default value config when nothing else overrides it", func() {
+				// The gateway level rate limit is 3
```
sam-heilbron, line 189: nit: Since this is based on the value of defaultLimit, we could loop from 1..defaultLimit and perform this assertion

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,356 @@
...
+	JustBeforeEach(func() {
+		testContext.JustBeforeEach()
+	})
+
+	JustAfterEach(func() {
+		testContext.JustAfterEach()
+	})
+
+	Context("Filter not configured", func() {
+		It("Should not add the filter to the list of HCM filters", func() {
```
sam-heilbron, line 106: I like this test a lot. In a later test, we wrap the assertion in an eventually to ensure that the config eventually makes it to envoy. Why do we believe that this test doesn't need that?

test/gomega/matchers/contain_headers.go
```diff
@@ -23,3 +25,24 @@ func ContainHeaders(headers http.Header) types.GomegaMatcher {
 	}
 	return gomega.And(headerMatchers...)
 }
+
+// getHeaderKeys returns a Gomega Transform that returns the headers in a request
+func getHeaderKeys() func(response *http.Response) []string {
```
sam-heilbron, line 30: nit: Since this is a transform, what do you think about putting this in https://github.com/solo-io/gloo/blob/main/test/gomega/transforms/transforms.go?

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
```diff
@@ -0,0 +1,49 @@
...
+    uint32 max_tokens = 1;
+
+    // The number of tokens added to the bucket during each fill interval. If not specified, defaults to a single token.
+    // Must be greater than zero.
+    google.protobuf.UInt32Value tokens_per_fill = 2;
+
+    // The fill interval that tokens are added to the bucket. During each fill interval tokens_per_fill are added to the bucket.
+    // The bucket will never contain more than max_tokens tokens.
+    // Defaults to 1 second.
+    google.protobuf.Duration  fill_interval = 3;
```
sam-heilbron, line 31: Perhaps a low-level detail that isn't super relevant, though I'm curious: how is the re-fill implemented by Envoy and does it introduce any latency to have a frequent (low internal) interval?

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
```diff
@@ -0,0 +1,49 @@
...
+// Configures the token bucket, used for rate limiting.
+// Ref. https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/local_rate_limit_filter
+message TokenBucket {
+    // The maximum tokens that the bucket can hold. This is also the number of tokens that the bucket initially contains.
+    // Must be greater than or equal to one.
+    uint32 max_tokens = 1;
+
+    // The number of tokens added to the bucket during each fill interval. If not specified, defaults to a single token.
+    // Must be greater than zero.
+    google.protobuf.UInt32Value tokens_per_fill = 2;
```
sam-heilbron, line 26: Tokens are extracted for each request, which theoretically could happen in different listener threads, correct? Or is the state shared across threads?

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return []plugins.StagedNetworkFilter{
+		{
+			NetworkFilter: &envoy_config_listener_v3.Filter{
+				Name: NetworkFilterName,
+				ConfigType: &envoy_config_listener_v3.Filter_TypedConfig{
+					TypedConfig: marshalledConf,
+				},
+			},
+			Stage: pluginStage,
```
sam-heilbron, line 121: The stages that have existed, are (I believe) more suited for http filters, as the final stage is the router stage. We can of course re-use these for the network filters as well, I just wanted to make sure that was intentional. Since the only previous network filter that we supported was the HCM, we didn't need to have stages (stages are really just used to sort filters before storing them in the xds cache). Do you think it's worth it to introduce different stages for our network filters?

sam-heilbron (commented):
Looking great! 

I imagine user facing docs/examples will follow this implementation, who will be responsible for delivering those?

A few minor nits (that can be ignored) and questions

test/e2e/local_ratelimit_test.go
davidjumani: I wrapped this in another function to make the tests more concise. Should be good now

davidjumani: Since we are already checking for status 200, I just reused that bit of code rather

test/e2e/local_ratelimit_test.go
davidjumani: The other tests validate the http local rate limit filter where TCP connection caching should not cause any issues.
At the network level, the client needs to be created before every request rather than every test - this is to ensure that connections are not cached and the behaviour can be validated

test/e2e/local_ratelimit_test.go
davidjumani: Wrapped it in a function, thanks for this!

test/e2e/local_ratelimit_test.go
davidjumani: In this test, since the filter is not configured, it should not make its way to envoy and we validate that via an assertion. We do not need to wrap it in an eventually since there's no patching of resources going on.
In other tests, we want to ensure that the RL config makes it to envoy when configured on the vHost and route via the `PatchDefaultVirtualService` method. Since this takes time to trickle down to the envoy config, we wrap an assertion that it should do so in an eventually and then begin the RL tests after this config makes its way to envoy

test/gomega/matchers/contain_headers.go
davidjumani: Moved, thanks!

test/e2e/local_ratelimit_test.go
davidjumani: Added comments outlining it

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
davidjumani: I've pretty much copied the envoy docs on these fields. I have to look into the re-fill implementation, but this is an [envoy warning](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/local_ratelimit/v3/local_rate_limit.proto.html#extensions-filters-http-local-ratelimit-v3-localratelimit)
```
In the current implementation the token bucket’s fill_interval must be >= 50ms to avoid too aggressive refills.
```

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
davidjumani: That depends on the value of [local_rate_limit_per_downstream_connection](https://github.com/solo-io/gloo/pull/8618/files#diff-ffc03b7ba1b505c76b3893bfa61aa649b7021387797351f3b35590c897f96275R44)

davidjumani:
> I imagine user facing docs/examples will follow this implementation, who will be responsible for delivering those?

Thanks @sam-heilbron I'll work with @Nadine2016 to ensure the docs are delivered as well

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
sam-heilbron: Given that warning, should we just be even more aggressive and error if that duration is supplied?

test/e2e/local_ratelimit_test.go
sam-heilbron: Yeah that makes sense, I was more proposing consistency across the entire test, even though some don't need the rebuilt client. Happy with what you have though

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+		if err != nil {
+			return err
+		}
+		p.filterRequiredForListener[params.HttpListener] = struct{}{}
+		return nil
+	}
+	return nil
+}
+
+func ConfigureRouteHostFilter(settings *local_ratelimit.Settings, localRatelimit *local_ratelimit.TokenBucket, stage uint32, out *envoy_config_route_v3.Route) error {
```
jenshu, line 202:
```suggestion
func ConfigureRouteFilter(settings *local_ratelimit.Settings, localRatelimit *local_ratelimit.TokenBucket, stage uint32, out *envoy_config_route_v3.Route) error {
```
can you add a comment that this func is exported for use in the enterprise plugin? (so that we don't accidentally un-export it later)

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+func modIfNoExisting(protoext proto.Message) pluginutils.ModifyFunc {
+	return func(existing *any.Any) (proto.Message, error) {
+		if existing == nil {
+			return protoext, nil
+		}
+		return nil, ErrConfigurationExists
+	}
+}
+
+func ConfigureVirtualHostFilter(settings *local_ratelimit.Settings, localRatelimit *local_ratelimit.TokenBucket, stage uint32, out *envoy_config_route_v3.VirtualHost) error {
```
jenshu, line 177: can you add a comment that this func is exported for use in the enterprise plugin? (so that we don't accidentally un-export it later)

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (p *plugin) NetworkFiltersHTTP(params plugins.Params, listener *v1.HttpListener) ([]plugins.StagedNetworkFilter, error) {
+	return generateNetworkFilter(listener.GetOptions().GetNetworkLocalRatelimit())
+}
+
+func (p *plugin) NetworkFiltersTCP(params plugins.Params, listener *v1.TcpListener) ([]plugins.StagedNetworkFilter, error) {
+	return generateNetworkFilter(listener.GetOptions().GetLocalRatelimit())
+}
+
+func GenerateHTTPFilter(settings *local_ratelimit.Settings, localRatelimit *local_ratelimit.TokenBucket, stage uint32) (*envoy_extensions_filters_http_local_ratelimit_v3.LocalRateLimit, error) {
```
jenshu, line 134: can you add a comment that this func is exported for use in the enterprise plugin? (so that we don't accidentally un-export it later)

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: Good catch. I've moved the helper methods to utils.go and added the comments there

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		p.filterRequiredForListener[params.HttpListener] = struct{}{}
+		return nil
+	}
+	return nil
+}
+
+func (p *plugin) HttpFilters(params plugins.Params, listener *v1.HttpListener) ([]plugins.StagedHttpFilter, error) {
+	settings := listener.GetOptions().GetHttpLocalRatelimit()
+	filter, err := GenerateHTTPFilter(settings, settings.GetDefaults(), CustomStageBeforeAuth)
```
jenshu, line 92: is the 2nd arg always going to come from settings? if so, can we just pass in `settings`?

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
```diff
@@ -0,0 +1,49 @@
...
+    TokenBucket defaults = 1;
+
+    // Specifies the scope of the rate limiter’s token bucket. If set to false, the token bucket is shared across all worker threads, thus the rate limits are applied per Envoy process.
+    // If set to true, a token bucket is allocated for each connection, thus the rate limits are applied per connection thereby allowing one to rate limit requests on a per connection basis.
+    // Defaults to false
+    bool local_rate_limit_per_downstream_connection = 2;
+
+    // Set this to true to return Envoy's X-RateLimit headers to the downstream.
+    // reference docs here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/common/ratelimit/v3/ratelimit.proto#envoy-v3-api-enum-extensions-common-ratelimit-v3-xratelimitheadersrfcversion
+    bool enable_x_ratelimit_headers = 3;
```
jenshu, line 48: can we use `google.protobuf.BoolValue` for these? that way we can differentiate between explicitly set vs. not set (if not set, we don't need to set the envoy field either), and also in case we ever want to change the default in the future

projects/gloo/api/v1/options/local_ratelimit/local_ratelimit.proto
```diff
@@ -0,0 +1,49 @@
...
+    // Defaults to 1 second.
+    google.protobuf.Duration  fill_interval = 3;
+}
+
+// The Local Rate Limit settings define the default local rate limit token bucket to apply as well as other configurations
+message Settings {
+    // The token bucket configuration to use for rate limiting requests.
+    // These options provide the ability to locally rate limit the connections in envoy. Each request processed by the filter consumes a single token.
+    // If the token is available, the request will be allowed. If no tokens are available, the request will receive the configured rate limit status.
+    TokenBucket defaults = 1;
```
jenshu, line 39: nit: can this have a more descriptive name, token_bucket or default_token_bucket?

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: Nope, the token bucket can be passed either from the route, vhost or gateway. In this case the default token bucket from the gateway is being passed

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *envoy_config_route_v3.Route) error {
+	if limits := in.GetOptions().GetRatelimit().GetLocalRatelimit(); limits != nil {
+		err := ConfigureRouteFilter(params.HttpListener.GetOptions().GetHttpLocalRatelimit(), limits, CustomStageBeforeAuth, out)
+		if err != nil {
+			return err
+		}
+		p.filterRequiredForListener[params.HttpListener] = struct{}{}
+		return nil
```
nfuden, line 85: nit: not needed
```suggestion
```

install/helm/gloo/crds/gateway.solo.io_v1_MatchableHttpGateway.yaml
```diff
@@ -1148,9 +1148,44 @@ spec:
...
+                            nullable: true
+                            type: boolean
+                          localRateLimitPerDownstreamConnection:
+                            nullable: true
+                            type: boolean
+                        type: object
                       leftmostXffAddress:
                         nullable: true
                         type: boolean
+                      networkLocalRatelimit:
```
nfuden, line 1176: So I think its kind of interesting that we have a commingling of concerns on the listener versus gateway tiers. 
Gateways originally mapped roughly to a listener but now are composable filterchains (sorta). Given this what is the expected behavior of setting this when there are multiple gateways being consumed matched?

davidjumani: Since each gateway creates its own filter chain on a listener, if there is a listener with multiple filter chains, the gateway configured with the network filter will contain the filter in its filter chain. 
Eg.: 
```
hybridGateway:
  matchedGateways:
  - httpGateway: {}
    matcher:
      sourcePrefixRanges:
      - addressPrefix: 0.0.0.0
        prefixLen: 1
  - httpGateway:
      options:
        networkLocalRatelimit:
          fillInterval: 100s
          maxTokens: 10
          tokensPerFill: 10
    matcher: {}
```
will result in two filter chains for the listener
```
{
"listener": {
  "name": "listener-::-8080",
  "filter_chains": [
  {
    "filter_chain_match": {
    "source_prefix_ranges": [
      {
      "address_prefix": "0.0.0.0",
      "prefix_len": 1
      }
    ]
    },
    "filters": [            <- the chains for the first gateway
    .........                   //  filters defined on the first gateway
    ]
  },
  {
    "filters": [            <- the chains for the second gateway with the network local rate limit configured
    {
      "name": "envoy.filters.network.local_ratelimit",  // the network local rate limit filter configured on the second gateway
     }
     .........  // other filters defined on the second gateway
    ]
  }
 ]
}
```

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/solo-io/gloo/test/gomega/matchers"
+	"github.com/solo-io/gloo/test/testutils"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = FDescribe("Local Rate Limit", func() {
```
sam-heilbron, line 24: Focused test

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+const (
+	ExtensionName           = "local_ratelimit"
+	NetworkFilterStatPrefix = "network_local_ratelimit"
+	HTTPFilterStatPrefix    = "http_local_ratelimit"
+	NetworkFilterName       = "envoy.filters.network.local_ratelimit"
+	HTTPFilterName          = "envoy.filters.http.local_ratelimit"
+	CustomStageBeforeAuth   = uint32(3)
+	CustomDomain            = "custom"
```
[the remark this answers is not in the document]
davidjumani: Good catch, removed

test/e2e/local_ratelimit_test.go
```diff
@@ -0,0 +1,352 @@
...
+
+		expectNotRateLimitedWithOutXRateLimitHeader = func() {
+			response := expectNotRateLimited()
+			// Since the values of the x-rate-limit headers change with time, we only check the presence of these header keys and not match their value
+			ExpectWithOffset(1, response).ToNot(matchers.ContainHeaderKeys([]string{"x-ratelimit-reset", "x-ratelimit-limit", "x-ratelimit-remaining"}),
+				"x-ratelimit headers should not be present for non rate limited requests")
+		}
+
+		expectNotRateLimitedWithXRateLimitHeader = func() {
+			defer GinkgoRecover()
```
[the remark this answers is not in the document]
davidjumani: Done!

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+var (
+	// Since this is an L4 filter, it would kick in before any HTTP auth can take place.
+	// This also bolsters its main use case which is protect resources.
+	pluginStage            = plugins.BeforeStage(plugins.AuthNStage)
```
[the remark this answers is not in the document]
davidjumani: I've cleared this up by splitting the HTTP and network filter stages. Also since pre / post auth doesn't make sense wrt TCP, I've changed the stage to make it more intuitive

sam-heilbron (approved):
🎉

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+	_ plugins.RoutePlugin         = new(plugin)
+)
+
+const (
+	ExtensionName           = "local_ratelimit"
+	NetworkFilterStatPrefix = "network_local_ratelimit"
+	HTTPFilterStatPrefix    = "http_local_ratelimit"
+	NetworkFilterName       = "envoy.filters.network.local_ratelimit"
+	HTTPFilterName          = "envoy.filters.http.local_ratelimit"
+	CustomStageBeforeAuth   = uint32(3)
```
jenshu, line 25: where does the value 3 come from? should this be using a constant defined somewhere?

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+var (
+	// For the network filter, it would kick in after the TCP connection limit filter to rate limit.
+	networkFilterPluginStage = plugins.DuringStage(plugins.RateLimitStage)
+	// For the HTTP filter, this is designed to rate limit early, and consequently kick in before auth
+	httpFilterPluginStage = plugins.BeforeStage(plugins.AuthNStage)
+
+	ErrConfigurationExists = errors.New("configuration already exists")
```
jenshu, line 34: what configuration is this referring to? can we make the error message more useful to help user debug config errors?

jenshu:
LGTM just a couple small questions

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: This indicates the stage of the filter's typed config in the vhost or route to map to the filter in the filter chain. If we decide to add local rate limit post auth, it will have its own stage mapping to the post auth filter

projects/gloo/pkg/plugins/local_ratelimit/plugin.go
davidjumani: This error indicates that the typed config for the filter (in this case local rate limit) already exists. This should never occur in OSS, since it is the first plugin to run. In enterprise, this error is returned if local rate limit is already configured in OSS and handled appropriately there

jenshu (approved):

PR #8622, opened 2023-08-24
Expand Credential Support for Vault's AWS Integration

inFocus7:
# Description

This change allows users to omit credentials when setting up Vault + AWS authentication, allowing for more ways of users to handle credentials (such as AWS Security Token Service (STS) or IAM Roles for Service Accounts (IRSA) which gives users temporary credentials).

## API changes
- Exposed `secretOptions` in the settings, so users don't need to go through `kubeResourceOverride.spec` to set it.

## Code changes
- Made `Settings_VaultAwsAuth.AccessKeyId` and `Settings_VaultAwsAuth.SecretAccessKey` optional by not erroring when omitted.
- Added Helm test to verify that setting `secretOptions` in settings works as expected.
- Reformatted the `vault_aws_test.go` code to allow for different VaultSecretSource + Credentials (explicit, and sts) setup.
    - At the moment this is essentially doing nothing new while the STS tests are skipped.

## Docs changes
- Added sections in out e2e README for AWS test setup.
- Updated the e2e README to refer to the new make command for running e2e tests.

# Context

> Currently for the Vault Integration in Gloo edge when using AWS IAM it is mandatory to configure the sensitive secret key and access key id in settings object. This does not augur well with Security teams. Also In cases for inferred EC2 environments using STS API which doesn't need the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY it causes a panic.
> Currently we do not have a way when Customer uses[ IRSA ](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)to assume the role and get the credentials programmatically.

Previously, by forcing the credentials either the users would need to use long-term credentials (not always advised by security teams), or they'd need to manually re-update every time their temporary credentials update (bad user experience).

### Helm values test
I opted out (commented) of testing that the `kubernetes` source gets set as it was causing failures to run the tests. The kubernetes source is enabled by setting it as an empty object `{}`, but trying to create a flattened version of it in our test causes issues when unmarshalling.

_how we set kubernetes source in helm override_
```
settings:
  secretOptions:
    sources:
    - kubernetes: {}
```

_the (flattened) helm value that i tried in helm tests_
```
  valuesArgs: []string{
    "settings.secretOptions.sources[0].kubernetes={}"
  }
}
```

_the error caused by the above_
```
json: cannot unmarshal array into Go struct field SecretOptionsSource.settings.secretOptions.sources.kubernetes of type generate.KubernetesSecrets.
```

### Skipping/Postponing STS E2E Tests
Decided on merging the main work (exposing helm values and making credentials optional) in this PR. Will be fixing the e2e tests and documentation afterward. Asked Kasun for help and waiting on a reply (time zone differences).

I kept the WIP tests with `TODO` and `Skip()` so they don't affect the PR, and since I will be following up on the tests as soon as (and while) this merges.

One difference I noticed from the e2e test setup and the documentation was that the documentation had extraAnnotations set in Helm which may be necessary here as well. I couldn't find a quick way to set Helm settings through our current test setup.

```
gloo:
  gloo:
    serviceAccount:
      extraAnnotations:
        eks.amazonaws.com/role-arn: <role-arn>
```

## Testing steps

** This wasn't manually tested, but Kasun & Akshay verified that making credentials optional fixes their STS/IRSA integration issues the documents included as part of the issue.

# Checklist:


nfuden:
This removes errors that might still have informative value in some cases consider this https://github.com/solo-io/gloo/commit/b5eb93fa93b1c43f86161613a3f0fa4c90cb4c0a

inFocus7:
> This removes errors that might still have informative value in some cases consider this https://github.com/solo-io/gloo/commit/b5eb93fa93b1c43f86161613a3f0fa4c90cb4c0a

Wouldn't this lead to confusion and/or annoyance on the user side? If someone has a valid reason for not setting those credentials they would still be getting errors logged? 
I'll look into if there's any way to handle it being a warning instead (after completing the DoD tasks, although that would probably require a third return type (since we still need to return an error here) and could get tricky as this is propagated through a few functions.

inFocus7:
/kick (just stopped midrun)
```
Step #0 - "publish-artifacts": Already have image (with digest): gcr.io/cloud-builders/gcloud
Step #1 - "run-tests": Already have image (with digest): gcr.io/cloud-builders/gcloud
Step #0 - "publish-artifacts": Creating temporary tarball archive of 0 file(s) totalling 0 before compression.
Step #0 - "publish-artifacts": Uploading tarball of [.] to [gs://solo-public_cloudbuild/source/1693330383.115358-dfdf4bffe12f4304be8d59cfbd3b376f.tgz]
Step #1 - "run-tests": Creating temporary tarball archive of 0 file(s) totalling 0 before compression.
Step #1 - "run-tests": Uploading tarball of [.] to [gs://solo-public_cloudbuild/source/1693330383.11255-8ca7b632c8fc4dcf9f3613b572d01bcf.tgz]
Step #0 - "publish-artifacts": Created [https://cloudbuild.googleapis.com/v1/projects/solo-public/locations/global/builds/4dee81cc-e3c7-4ad3-966c-6b67d132d5ba].
Step #0 - "publish-artifacts": Logs are available at [ https://console.cloud.google.com/cloud-build/builds/4dee81cc-e3c7-4ad3-966c-6b67d132d5ba?project=907026730415 ].
Step #1 - "run-tests": Created [https://cloudbuild.googleapis.com/v1/projects/solo-public/locations/global/builds/6ab02df0-9f1d-44b6-ab57-5c169771ff31].
Step #1 - "run-tests": Logs are available at [ https://console.cloud.google.com/cloud-build/builds/6ab02df0-9f1d-44b6-ab57-5c169771ff31?project=907026730415 ].
```

Running CI to test that the changes still work for aws+vault tests. Tried locally but getting permission denied when running tests using updated tmp credentials.

inFocus7:
Marking ready for review so all tests pass. Unable to run tests locally as my credentials keep getting rejected, so running tests through CI in the meantime while I look into it.

inFocus7:
/kick updated role trust relationships in aws

changelog/v1.16.0-beta6/make-vault-aws-aim-creds-optional.yaml
```diff
@@ -0,0 +1,14 @@
+changelog:
+  - type: NEW_FEATURE
+    description: >
+      Make Vault's AWS credentials optional. 
+      This allows users to integrate AWS Security Token Service (STS) or IAM Roles for Service Accounts (IRSA) for temporary credentials.
+      https://developer.hashicorp.com/vault/docs/auth/aws#iam-authentication-inferences
+
+      skipCI-kube-tests:true
```
sam-heilbron, line 8: This will show up in our changelog, so I would encourage putting this in a new entry in the list, under NON_USER_FACING

changelog/v1.16.0-beta6/make-vault-aws-aim-creds-optional.yaml
```diff
@@ -0,0 +1,14 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: nit: I would consider this a fix. The feature hasn't changed, we have just fixed an incorrect assumption about env variables being required in the settings, that don't need to be. I like the references to the docs in here!

install/helm/gloo/generate/values.go
```diff
@@ -219,6 +220,58 @@ type AwsSettings struct {
...
+type SecretOptionsSource struct {
+	Kubernetes KubernetesSecrets `json:"kubernetes,omitempty" desc:"Only one of kubernetes, vault, or directory may be set"`
+	Vault      VaultSecrets      `json:"vault,omitempty" desc:"Only one of kubernetes, vault, or directory may be set"`
+	Directory  Directory         `json:"directory,omitempty" desc:"Only one of kubernetes, vault, or directory may be set"`
+}
+
+type KubernetesSecrets struct {
+}
+
+type VaultSecrets struct {
```
sam-heilbron, line 236: Given that these are all deprecated in the Settings API, should we expose them in our Helm API? I figure it eases the user experience, if we only expose the fields that we are encouraging people to configure

install/helm/gloo/templates/18-settings.yaml
```diff
@@ -13,6 +13,13 @@ spec:
 {{- if .Values.settings.devMode }}
   devMode:
     {{ .Values.settings.devMode}}
+{{- end }}
+{{- if .Values.settings.secretOptions }}
```
sam-heilbron, line 17: Does this mean that if you set secretOptions, but not sources, we will configure an empty secretOptions object? What will the behavior of Edge be then?

install/test/helm_test.go
```diff
@@ -4008,6 +4008,23 @@ spec:
...
+					It("sets secretOptions in settings", func() {
+						settings := makeUnstructureFromTemplateFile("fixtures/settings/set_secretSettings_in_settings.yaml", namespace)
+						prepareMakefile(namespace, helmValues{
+							valuesArgs: []string{
+								"settings.secretOptions.sources[0].vault.address=http://vault-internal.vault:8200",
+								"settings.secretOptions.sources[0].vault.aws.iamServerIdHeader=vault.gloo.example.com",
+								"settings.secretOptions.sources[0].vault.aws.mountPath=aws",
+								"settings.secretOptions.sources[0].vault.aws.region=us-east-1",
+								"settings.secretOptions.sources[0].vault.pathPrefix=dev",
+								// The kubernetes source is enabled through `kubernetes: {}` in our settings, but JSON fails to unmarshal the empty object below.
```
sam-heilbron, line 4020: I like this callout here in the comment, though as a developer experiencing this (if it merges), it's unclear to me what I should do. Can you help me understand this behavior a bit? Is this a case that we want to test, but due to the error, the tests fail?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -155,15 +155,14 @@ func configureAwsAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*api.C
 }
 
 func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*api.Client, error) {
-	if accessKeyId := aws.GetAccessKeyId(); accessKeyId == "" {
-		return nil, errors.New("access key id must be defined for AWS IAM auth")
```
sam-heilbron, line 159: This error did help indicate that the env variables weren't defined. Without this, we don't have feedback to cover this use case. Since there are 2 possiblities:
- Misconfig that user wants to be aware of
- Intentionally not including since using STS/IRSA

Do you think it would be valuable to include a warning log that those env variables aren't defined? That way we at least alert users about the potential

test/e2e/README.md
```diff
@@ -108,4 +108,25 @@ Once halted, use `docker ps` to determine the admin port for the Envoy instance,
 #### Use INVALID_TEST_REQS
 Certain tests require environmental conditions to be true for them to succeed. For example, there are tests that only run on Linux machines.
 
-By setting `INVALID_TEST_REQS=skip`, you can run all tests locally, and any tests which will not run in your local environment will be skipped. The default behavior is that they fail.
\ No newline at end of file
+By setting `INVALID_TEST_REQS=skip`, you can run all tests locally, and any tests which will not run in your local environment will be skipped. The default behavior is that they fail.
+
+#### Focusing on tests
```
sam-heilbron, line 113: I like this addition! Part of the reason I didn't use the `run-e2e-tests` is because it include the command line filter, and thus excludes the in-code "F"-prefix focus. I'm happy to have this as is, just wanted to call out, that an alternative to ensure focused tests are stil respected, is to use the underlying `make test` and just passing the e2e directory

test/e2e/vault_aws_test.go
```diff
@@ -24,86 +23,164 @@ const (
...
+						Ctx: testContext.Ctx(),
+					})
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(secret.GetOauth().GetClientSecret()).To(Equal("new-secret"))
+			}, "5s", ".5s").Should(Succeed())
+		})
+
+	})
+
+	// TODO(fabiangonz98): Fix these tests
```
sam-heilbron, line 155: Is this TODO still relevant?

sam-heilbron (commented):
I really like the approach! A few small comments/questions

install/test/helm_test.go
inFocus7: Nothing we need to do, it was just an addition to be closer to the initial request's setup in the issue. 

All adding it would do is make sure that the following is rendered:
```
settings:
  secretOptions:
    sources:
    - vault:
      <values>
    - kubernetes: {}
```

But due to some unmarshalling done in the tests, we can't have `value={}` since we hit the issue:
```
json: cannot unmarshal array into Go struct field SecretOptionsSource.settings.secretOptions.sources.kubernetes of type generate.KubernetesSecrets.
```

I'll remove it since it would just cause confusion.

test/e2e/vault_aws_test.go
inFocus7: Yeah, it's relevant to the comment I left in the description. The tests don't work so need to fix them in followup.

projects/gloo/pkg/bootstrap/clients/vault.go
inFocus7: Yeah Nathan brought this up as well. Unsure how to log only as a warning, since returning as an `err` is wrong. 

A workaround I can think of is returning a third type for warning message(s), and finding the right place to log a warning/info, although this function is called in multiple places and didn't want to risk breaking anything.

install/helm/gloo/templates/18-settings.yaml
inFocus7: I'm not sure what our default for that case is. Technically it is possible to do this through the `spec` setup people have to do for now, so I kept the possibility here just in case. 
I could change the logic to `if secretOptions & secretOptions.sources`, I don't have any strong opinions either way.

projects/gloo/pkg/bootstrap/clients/vault.go
inFocus7: Two approaches I can think of for this:
1. Add an extra return to `VaultClientForSettings` of type `string`. This string will be the warning message(s) regarding missing aws id and/or secret.
2. Create a specific `error` to return when missing credentials. 

The in the [following code](https://github.com/solo-io/gloo/blob/f87182430382b64c503a53c93d17b3c42ec79447/projects/gloo/pkg/syncer/setup/setup_syncer.go#L345-L352) we could: `if warning / if err = MissingCredentials: log.Warn/Info(warning)`.

install/helm/gloo/templates/18-settings.yaml
ben-taussig-solo: I prefer the suggested change (`if secretOptions & secretOptions.sources`) to what is here currently

test/services/vault.go
```diff
@@ -188,32 +188,71 @@ func (i *VaultInstance) EnableSecretEngine(secretEngine string) error {
...
-	tmpFileName := filepath.Join(i.tmpdir, "policy.json")
-	err = os.WriteFile(tmpFileName, []byte(`{"path":{"*":{"capabilities":["create","read","update","delete","list","patch","sudo"]}}}`), 0666)
+	err = i.addAdminPolicy()
 	if err != nil {
 		return err
 	}
-	_, err = i.Exec("policy", "write", "admin", tmpFileName)
+
+	// Configure the AWS auth method with the creds provided
+	_, err = i.Exec("write", "auth/aws/config/client", fmt.Sprintf("secret_key=%s", settings.GetAws().GetSecretAccessKey()), fmt.Sprintf("access_key=%s", settings.GetAws().GetAccessKeyId()))
```
ben-taussig-solo, line 220: nit: does it make sense to extract this into a helper like `addAuthRole`?

test/services/vault.go
```diff
@@ -188,32 +188,71 @@ func (i *VaultInstance) EnableSecretEngine(secretEngine string) error {
...
+	_, err = i.Exec("policy", "write", "admin", tmpFileName)
+	return err
+}
+
+func (i *VaultInstance) addAuthRole(awsAuthRole string) error {
+	_, err := i.Exec("write", "auth/aws/role/vault-role", "auth_type=iam", fmt.Sprintf("bound_iam_principal_arn=%s", awsAuthRole), "policies=admin")
+	return err
+}
+
+func (i *VaultInstance) EnableAWSCredentialsAuthMethod(settings *v1.Settings_VaultSecrets, awsAuthRole string) error {
```
ben-taussig-solo, line 206: Where does `EnableAwsCredentialsAuthMethod` get called from?

install/helm/gloo/templates/18-settings.yaml
inFocus7: Yeah that sounds good, just pushed changes 👍🏼

test/services/vault.go
inFocus7: I extracted the other pieces of code since they're relevant for different "auth methods" (sts, credentials) although didn't extract this since it's specific for this setup. Similar to [here](https://github.com/solo-io/gloo/blob/98b553ebe107b597a7c3c6e0314b152355479cc4/test/services/vault.go#L248-L249) where I kept the execution code as-is for the sts auth method, since it's only used there.

test/e2e/README.md
```diff
@@ -108,4 +108,19 @@ Once halted, use `docker ps` to determine the admin port for the Envoy instance,
 #### Use INVALID_TEST_REQS
 Certain tests require environmental conditions to be true for them to succeed. For example, there are tests that only run on Linux machines.
 
-By setting `INVALID_TEST_REQS=skip`, you can run all tests locally, and any tests which will not run in your local environment will be skipped. The default behavior is that they fail.
\ No newline at end of file
+By setting `INVALID_TEST_REQS=skip`, you can run all tests locally, and any tests which will not run in your local environment will be skipped. The default behavior is that they fail.
+
+### Notes
+### AWS Tests
```
sam-heilbron, line 114: There is an aws-tests file which I tried to separate, since it includes details that are unique: https://github.com/solo-io/gloo/blob/main/devel/testing/aws-tests.md

inFocus7: Removed this section from here, but added one in the `aws-tests.md` doc for temporary credentials. From personal experience and from what I can tell, "newer" employees (at least me) can't generate credentials as said in our docs, but we can use temporary credentials.

test/services/vault.go
ben-taussig-solo: makes sense!

projects/gloo/pkg/bootstrap/clients/vault.go
ben-taussig-solo: It's a little frustrating that `enableServiceAccountCredentials` (which is typically set to true on install when using the IRSA/STS workflow for authenticating with AWS) is only supported as a helm setting. If it existed in global settings, then we could check whether it is set, and throw the now-removed errors if it isn't.

I guess we could do something similar by checking for the presence of `settings.gloo.awsOptions.serviceAccountCredentials`, which [gets created](https://github.com/solo-io/gloo/blob/5d6e4d1fbfcd4f5298ad285ca88e194903c1aa03/install/helm/gloo/templates/18-settings.yaml#L46-L48) when `enableServiceAccountCredentials` is set to true. It feels a little indirect or potentially unclear to future maintainers though, so I don't love it

inFocus7: Is that for our aws lambda work? 👀

Haven't looked into the code much, but iirc on the lambda side of things we have our own code to get credentials based on the arn+token(?). 

In the case for Vault+AWS integration, Vault/AWS handles the credential fetching not us, so not sure if `enableServiceAccountCredentials` would be as relevant since we aren't necessarily enabling anything. (Although I'm probably reading into the naming a bit too much)

ben-taussig-solo: Yeah I don't really know enough about the vault workflow to be confident in my suggestion

nfuden: Setting one but not the other is a DEFINITE configuration error though.
Consider something like https://github.com/solo-io/gloo/compare/main...vault-moreaws-temp

nfuden (changes requested):
We cant fully remove gaurdrails here we should call out what is truly configuration error

projects/gloo/pkg/bootstrap/clients/vault.go
inFocus7: Love the idea/implementation! Need to decide between adding the final error message through `awsauth.NewAWSAuth`'s error, or `client.Auth().Login`'s.
https://github.com/solo-io/gloo/blob/15f0c2c39c1d57049f340f693a2587003d1e3a2a/projects/gloo/pkg/bootstrap/clients/vault.go#L191-L200

I'm thinking on login. The `NewAWSAuth` just creates an object with parameters set (through WithRole, WithRegion, etc.), and it would make sense to have the error message on the failure to login.

nfuden: Yeah I think that makes sense I just want to make sure we are capturing errors as best as we can and the login failure would make sense

test/e2e/vault_aws_test.go
```diff
@@ -24,86 +23,164 @@ const (
...
+					},
+				},
+				PathPrefix: bootstrap.DefaultPathPrefix,
+				RootKey:    bootstrap.DefaultRootKey,
 			}
 		})
 
+		// add the extra annotation
+		JustBeforeEach(func() {
+			// can't find helm settings on test start...
```
sam-heilbron, line 177: We spoke in slack about e2e v kube2e tests, so perhaps this just needs to be updated. I don't think that a reference to helm in these tests makes sense, given that these don't interact with helm or kubernetes

test/services/vault.go
```diff
@@ -188,32 +188,71 @@ func (i *VaultInstance) EnableSecretEngine(secretEngine string) error {
...
+	// Configure the Vault role to align with the provided AWS role
+	err = i.addAuthRole(awsAuthRole)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (i *VaultInstance) EnableAWSSTSAuthMethod(awsAuthRole, serverIdHeader, stsRegion string) error {
```
sam-heilbron, line 234: I like how you exposed these methods on the service itself

sam-heilbron (approved):
🚀

nfuden (approved):

Issue #5944, opened 2022-02-17
certgen waits for cert to be expired for renewing it


dguihal:
### Gloo Edge Version

1.10.x (latest stable)

### Describe the bug

We are facing some issues about gloo internal cert being expired ...

Looking at code, looks like certgen is waiting the cert to be effectively expired before renewing it 

https://github.com/solo-io/gloo/blob/a6e02607c5d6c9a458388f707c881345a0554eab/jobs/pkg/run/run.go#L72

https://github.com/solo-io/gloo/blob/a6e02607c5d6c9a458388f707c881345a0554eab/jobs/pkg/kube/secret.go#L59

As it is run as a cronjob, it means that during a timeframe (depending on cronjob frequency) cert is effectively expired .....

### Steps to reproduce the bug

1. Instanciate a cert using certgen as a cronjob
2. Wait for expiration date
3. Enjoy having an expired cert while waiting for the cronjob to run

### Expected Behavior

I would expect such tool to renew cert before expiration (at least one day before but preferably a configurable date as it will also depend on cronjob run frequency)

DiegoLelis:
+1

Issue #6105, opened 2022-03-17
Admission control certificates not re-generated after upgrades


antonioberben:
### Gloo Edge Version

1.10.x (latest stable)

### Kubernetes Version

1.21.x

### Describe the bug

Gloo Edge deploys some `ValidatingWebhookConfiguration` which requires also to deploy a certificate so that k8s api server can call the defined admission controller.

This certificates are creates through a job which is deployed as a helm hook.

The problem is that this certificates never got renew after running a helm upgrade which might lead to an outdated certificate. Currently, we give 1 year expiration time. However, since we do not renew ever the certificate we can get into the situation where the certificate got expired and implies a manual actual.

This is an inconvenient for people who rely on automation or gitops tools.

### Steps to reproduce the bug

1. Install gloo edge with helm chart:
```
helm upgrade --install gloo  glooe/gloo-ee --version 1.10.12 --devel -n gloo-system --create-namespace
```
2. A job is created and the secret is generated. Check the value of the certificate:
```
date; k get secret gateway-validation-certs -n gloo-system -o jsonpath={'.data'} | jq -r '."tls.crt"' | base64 -d | openssl x509 -noout -text | head
```
3. Upgrade the installation:
```
helm upgrade --install gloo  glooe/gloo-ee --version 1.10.13 --devel -f $SOLO_EDGE_DEFAULT_VALUES -n gloo-system --create-namespace
```
4. Check the certificate:
```
date; k get secret gateway-validation-certs -n gloo-system -o jsonpath={'.data'} | jq -r '."tls.crt"' | base64 -d | openssl x509 -noout -text | head
```

Th certificate hasn't been re-created with a new expiration date

### Expected Behavior

The certificate should be re-new in every upgrade so that it is quite impossible to meet the expiration date. It will be only happening if the software is not upgraded in a whole year.

The affected Jobs:
https://github.com/solo-io/gloo/blob/34c21d74729a887b0c9e473c15865999e481af7a/install/helm/gloo/templates/6.5-gateway-certgen-job.yaml#L17

The hook should be: 
```
"helm.sh/hook": pre-install,pre-upgrade,pre-rollback
```

Also, there is this other job: https://github.com/solo-io/gloo/blob/28c1ce44f00d7ef97974e9ad053d5b7e4be8b85f/install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml#L19
But the logic is better. Not tested this scenario.

[...]

PR #8643, opened 2023-09-04
Renew gloo certificates before expiration

Megum1n:
# Description

This change will make certgen job renew existing certificate if it is expiring in two months.

Issues: https://github.com/solo-io/gloo/issues/5944 https://github.com/solo-io/gloo/issues/6105

## Code changes
- Update `GetExistingValidTlsSecret()` function to return `nil` in cert is expiring soon
- Add `renew-before` flag to be able to configure certificate rotation
- Add ability to set `force-rotation` flag via helm chart
- Create certgen cronjob if validation webhook is enabled 

# Context

New resources cannot be deployed in k8s cluster if validation webhook does not work. If gloo certificate is expired and gloo validation webhook is deployed to the cluster - new resources cannot be applied.
```plaintext
Upstream/test-ns/test-service dry-run failed, reason: InternalError, error: Internal error occurred: failed calling webhook "gloo.gloo-system.svc": failed to call webhook: Post "https://gloo.gloo-system.svc:443/validation?timeout=10s": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2023-09-04T12:45:55Z is after 2023-08-31T11:40:47Z
```
To avoid this issue, certificate should be renewed before expiration date.

## Testing steps

I manually created certgen job in my cluster and old certificate (Oct 2023 expiration date) was renewed.

## Notes for reviewers

~~We can go further and add configuration option for renewal period. Also, we can create cronjob for gateway certgen which will perform cert validation every X days/months (cron schedule). Current solution requires someone to perform helm upgrade, which executes pre-upgrade hook with certgen job in it.~~
Addressed it in the latest changes.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5944


jobs/pkg/kube/secret.go
```diff
@@ -60,7 +60,8 @@ func GetExistingValidTlsSecret(ctx context.Context, kube kubernetes.Interface, s
 			return nil, errors.Wrapf(err, "failed to decode pem encoded ca cert")
 		}
 
-		if now.Before(cert.NotBefore) || now.After(cert.NotAfter) {
+		// Create new certificate if old one is expiring soon (two months)
+		if now.Before(cert.NotBefore) || now.After(cert.NotAfter.AddDate(0, -2, 0)) {
```
nfuden, line 64: This is pretty opinionated, i think its a reasonable default but it pretty much has to be configurable imo.

nfuden:
Thanks for the contribution! May I ask where the 2 months number is coming from?

install/helm/gloo/values-template.yaml
```diff
@@ -90,6 +90,7 @@ gateway:
     setTtlAfterFinished: true
     ttlSecondsAfterFinished: 60
     runAsUser: 10101
+    forceRotation: false
```
elcasteel, line 93: nit: make this default to true to match the current behavior

elcasteel (commented):
Overall looks great, thanks! One tiny comment.

elcasteel (dismissed):

install/helm/gloo/templates/19-gloo-certgen-cronjob.yaml
```diff
@@ -60,13 +60,15 @@ spec:
               args:
                 - "--secret-name=gloo-mtls-certs"
                 - "--svc-name=gloo"
+                {{- if .Values.gateway.certGenJob.forceRotation }}
                 - "--force-rotation=true"
-{{- end }} {{/* if and .Values.global.glooMtls.enabled .Values.gateway.certGenJob.cron.enabled */}}
+                {{- end }} {{/* if .Values.gateway.certGenJob.forceRotation */}}
```
jenshu, line 65: i think we would want the cronjob to always force rotation, rather than be guarded by a flag

jobs/pkg/kube/secret.go
```diff
@@ -60,7 +60,8 @@ func GetExistingValidTlsSecret(ctx context.Context, kube kubernetes.Interface, s
 			return nil, errors.Wrapf(err, "failed to decode pem encoded ca cert")
 		}
 
-		if now.Before(cert.NotBefore) || now.After(cert.NotAfter) {
+		// Create new certificate if old one is expiring soon (two months)
```
jenshu, line 63: can remove the "two months" since it's now configurable

install/helm/gloo/templates/19-gloo-certgen-cronjob.yaml
Megum1n: @jenshu I'm not sure about this one

- Force rotation is enabled by default
- There is a new `renew-before` flag now which is set to `2160h` (90 days). Any certificate will be renewed if its expiration time is due soon

I think that more configuration options for users is a better solution than forcing renewal for everyone.

jenshu (dismissed):
just one more comment, but approving. thanks for the PR!

install/helm/gloo/templates/19-gloo-certgen-cronjob.yaml
```diff
@@ -1,9 +1,9 @@
-{{ define "glooMtls.certGenCronJobSpec" }}
+{{ define "gloo.certGenCronJobSpec" }}
 {{- $image := .Values.gateway.certGenJob.image }}
 {{- if .Values.global  }}
 {{- $image = merge .Values.gateway.certGenJob.image .Values.global.image }}
 {{- end }}
-{{- if and .Values.global.glooMtls.enabled .Values.gateway.certGenJob.cron.enabled }}
+{{- if and (or .Values.global.glooMtls.enabled .Values.gateway.validation.webhook.enabled) .Values.gateway.certGenJob.cron.enabled }}
```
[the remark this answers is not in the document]
Megum1n: I just found out that this cronjob is for mtls certificates only because of `--secret-name=gloo-mtls-certs` flag.
Certgen is needed to be executed  with `--secret-name={{ .Values.gateway.validation.secretName }}` flag as well. Seems like small refactoring is required :)

Options:
- Add second container in the same cronjob which will be added if `.Values.gateway.validation.enabled`
- Add second separate cronjob for "validation secret"

First option is better imo. What do you think?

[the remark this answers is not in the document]
jenshu: ahh good catch. there are also some other things like the CronJob name, that are hardcoded with "mtls".  my preference would be to add a completely separate CronJob for the validation webhook certgen. even though there might be some duplication, they are used for different purposes, and it would make the templates more readable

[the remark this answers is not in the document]
Megum1n: @jenshu is it okay for two cronjobs to use common `.Values.gateway.certGenJob.cron` options or should I separate them?

install/helm/gloo/templates/6.5-gateway-certgen-cronjob.yaml
```diff
@@ -0,0 +1,75 @@
...
+                {{- if .Values.gateway.certGenJob.forceRotation }}
+                - "--force-rotation=true"
+                {{- end }} {{/* if .Values.gateway.certGenJob.forceRotation */}}
+{{- end }} {{/* if and .Values.gateway.enabled .Values.gateway.validation.enabled .Values.gateway.validation.webhook.enabled .Values.gateway.certGenJob.cron.enabled */}}
+{{- end }} {{/* define gateway.certGenJob.JobSpec*/}}
+
+{{/* Render template with yaml overrides */}}
+{{- $kubeResourceOverride := dict -}}
+{{- if .Values.gateway.certGenJob -}}
+{{- $kubeResourceOverride = .Values.gateway.certGenJob.cron.KubeResourceOverride -}}
```
jenshu, line 73: typo probably from find-and-replace - `KubeResourceOverride`
we should define a new value `kubeResourceOverride` for the validation webhook cronjob

install/helm/gloo/templates/19-gloo-certgen-cronjob.yaml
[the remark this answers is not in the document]
jenshu: yeah it's unfortunate that the mtls and vwh certgen jobs share the same `certGenJob` helm values, they really should probably be separate. for now it's probably ok to keep them both using the same `cron` settings since if we want to separate them, there's a bigger refactoring we would have to do

install/helm/gloo/templates/6.5-gateway-certgen-cronjob.yaml
Megum1n: No a typo, I just used same resources as defined in `6.5-gateway-certgen-job.yaml`

Megum1n: Added new variable

install/helm/gloo/generate/values.go
```diff
@@ -451,9 +452,10 @@ Scheduling:
 * * * * *
 */
 type CertGenCron struct {
-	Enabled                  *bool                  `json:"enabled,omitempty" desc:"enable the cronjob"`
-	Schedule                 *string                `json:"schedule,omitempty" desc:"Cron job scheduling"`
-	MtlsKubeResourceOverride map[string]interface{} `json:"mtlsKubeResourceOverride,omitempty" desc:"override fields in the gloo-mtls-certgen cronjob."`
+	Enabled                               *bool                  `json:"enabled,omitempty" desc:"enable the cronjob"`
+	Schedule                              *string                `json:"schedule,omitempty" desc:"Cron job scheduling"`
+	MtlsKubeResourceOverride              map[string]interface{} `json:"mtlsKubeResourceOverride,omitempty" desc:"override fields in the gloo-mtls-certgen cronjob."`
+	ValidationWebhookKubeResourceOverride map[string]interface{} `json:"ValidationWebhookKubeResourceOverride,omitempty" desc:"override fields in the gateway-certgen cronjob."`
```
jenshu, line 458:
```suggestion
	ValidationWebhookKubeResourceOverride map[string]interface{} `json:"validationWebhookKubeResourceOverride,omitempty" desc:"override fields in the gateway-certgen cronjob."`
```
(should be lowercase)

install/helm/gloo/templates/6.5-gateway-certgen-cronjob.yaml
```diff
@@ -0,0 +1,75 @@
...
+{{- $image = merge .Values.gateway.certGenJob.image .Values.global.image }}
+{{- end }}
+{{- if and .Values.gateway.enabled .Values.gateway.validation.enabled .Values.gateway.validation.webhook.enabled .Values.gateway.certGenJob.cron.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  labels:
+    app: gloo
+    gloo: gloo-certgen-cronjob
+  name: gloo-certgen-cronjob
```
jenshu, line 13: nit: to match the name of the Job `gateway-certgen`, this could be renamed to `gateway-certgen-cronjob`

install/helm/gloo/templates/6.5-gateway-certgen-cronjob.yaml
```diff
@@ -0,0 +1,75 @@
...
+                {{- if .Values.gateway.certGenJob.forceRotation }}
+                - "--force-rotation=true"
+                {{- end }} {{/* if .Values.gateway.certGenJob.forceRotation */}}
+{{- end }} {{/* if and .Values.gateway.enabled .Values.gateway.validation.enabled .Values.gateway.validation.webhook.enabled .Values.gateway.certGenJob.cron.enabled */}}
+{{- end }} {{/* define gateway.certGenJob.JobSpec*/}}
+
+{{/* Render template with yaml overrides */}}
+{{- $kubeResourceOverride := dict -}}
+{{- if .Values.gateway.certGenJob -}}
+{{- $kubeResourceOverride = .Values.gateway.certGenJob.cron.ValidationWebhookKubeResourceOverride -}}
```
jenshu, line 73:
```suggestion
{{- $kubeResourceOverride = .Values.gateway.certGenJob.cron.validationWebhookKubeResourceOverride -}}
```

install/helm/gloo/templates/6.5-gateway-certgen-job.yaml
```diff
@@ -69,6 +72,6 @@ spec:
 {{/* Render template with yaml overrides */}}
 {{- $kubeResourceOverride := dict -}}
 {{- if .Values.gateway.certGenJob -}}
-{{- $kubeResourceOverride = .Values.gateway.certGenJob.kubeResourceOverride -}}
+{{- $kubeResourceOverride = .Values.gateway.certGenJob.ValidationWebhookKubeResourceOverride -}}
```
jenshu, line 75: this change should be reverted
```suggestion
{{- $kubeResourceOverride = .Values.gateway.certGenJob.kubeResourceOverride -}}
```

jenshu (commented):
the changes look good to me! 2 questions:
- do you think there might be a need for the jobs and cronjobs to have different `forceRotation` values? (e.g. `.Values.gateway.certGenJob.forceRotation` and `.Values.gateway.certGenJob.cron.forceRotation`)
- can you add tests for the new gateway-certgen-cronjob in `install/test/helm_test.go`?  (can probably just insert a new entry in all the places the mtls cronjob is tested)

Megum1n:
@jenshu 

- `forceRotation` option will be used not only for cronjobs, but also for helm-hook jobs. I don't think that there is a need for separate configuration for each job. I can create a separate issue to consider configuration split for each job if you want. For now, I suggest we leave it as is.

- Thank you, I added new tests. Also added forceRotation option to the test. 
Sorry, my Goland formatted the whole file a bit, I hope you do not mind.

jenshu:
> * `forceRotation` option will be used not only for cronjobs, but also for helm-hook jobs. I don't think that there is a need for separate configuration for each job. I can create a separate issue to consider configuration split for each job if you want. For now, I suggest we leave it as is.

I agree we can let them share the same values for now, and if the need arises we can split them later

elcasteel (approved):

jenshu:
@Megum1n sorry about that. to get around this, can you please remove the offending `echo "WARNING"...` from the Makefile?

elcasteel (approved):

install/test/helm_test.go
```diff
@@ -6357,6 +6364,7 @@ metadata:
 					Entry("6-access-logger-deployment", "accessLogger.deployment.kubeResourceOverride", "accessLogger.enabled=true"),
 					Entry("6-access-logger-service", "accessLogger.service.kubeResourceOverride", "accessLogger.enabled=true"),
 					Entry("6.5-gateway-certgen-job", "gateway.certGenJob.kubeResourceOverride"),
+					Entry("6.5-gateway-certgen-cronjob", "gateway.certGenJob.kubeResourceOverride"),
```
jenshu, line 6367:
```suggestion
					Entry("6.5-gateway-certgen-cronjob", "gateway.certGenJob. cron.validationWebhookKubeResourceOverride"),
```

jenshu: i'm not sure but might also need to add the other values like `gateway.validation.enabled=true`, etc

jenshu (approved):

elcasteel (dismissed):

install/test/helm_test.go
```diff
@@ -6357,6 +6364,7 @@ metadata:
 					Entry("6-access-logger-deployment", "accessLogger.deployment.kubeResourceOverride", "accessLogger.enabled=true"),
 					Entry("6-access-logger-service", "accessLogger.service.kubeResourceOverride", "accessLogger.enabled=true"),
 					Entry("6.5-gateway-certgen-job", "gateway.certGenJob.kubeResourceOverride"),
+					Entry("6.5-gateway-certgen-cronjob", "gateway.certGenJob.cron.validationWebhookKubeResourceOverride"),
```
jenshu, line 6367: probably need to add the other values `"gateway.enabled=true", "gateway.validation.enabled=true", "gateway.validation.webhook.enabled=true", "gateway.certGenJob.cron.enabled=true"`

elcasteel (approved):

jenshu (approved):

PR #8594, opened 2023-08-17
OIDC add claims to headers

jjamroga:
# Description

Please include a high level summary of the changes.

This bug fixes ... \ This new feature can be used to ...

_Fill out any of the following sections that are relevant and remove the others_
## API changes
- Added x field to y resource
- ...

## Code changes
- Fix error in `Foo()` function
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- Added guide about feature x to public docs
- Updated README to account for y behavior
- ...

# Context

Users ran into this bug doing ... \ Users needed this feature to ...

See slack conversation [here](https://solo-io-corp.slack.com/archives/some/post)

## Interesting decisions
 
We chose to do things this way because ...

## Testing steps

I manually verified behavior by ...

## Notes for reviewers

Be sure to verify intended behavior by ...

Please proofread comments on ...

This is a complex PR and may require a huddle to discuss ...

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #8651, opened 2023-09-06
[v1.15.x] - Backport OIDC add claims to headers

jjamroga:
Backports https://github.com/solo-io/gloo/pull/8594 to v1.15.x

# Description

Introduced two new fields on the OIDC Authorization Code plugin to enable mapping claims to headers for the OAuth2 access token and the OIDC identity token.

## API changes
- Added nested messages for both access and identity tokens, each containing repeated claims to headers fields that are equivalent to what is present in JWT policy.

# Context
Some users are requesting ExtAuthPolicy used in conjunction with a JWT policy, and when these are configured in an `||` evaluation via ExtAuthPolicy, we cannot map claims from just-issued tokens.


timflannagan (approved):

timflannagan:
Can we add some commentary on why these API changes need to be backported as well?

sheidkamp (approved):

PR #8631, opened 2023-08-30
Version Assets Properly from Forked Repositories

sam-heilbron:
# Description

Ensure that a PR from a forked repository, can correctly version resources generated in the PR

## CI changes
- Provide a default VERSION, in case `git describe` returns nothing

# Context

## Problem
The `VERSION` variable in the Makefile is used as the source of truth for identifying asset (docker images, helm charts) tags. On releases, this tag is supplied by the manual version that is supplied, but on PRs, we must determine the version dynamically. To do so, and to ensure that each PR has a unique version, we determine the value by the following:
- Use git describe to gather the latest tag
- Append a unique suffix that is provided by cloudbuild

And this ensures that versions on PRs tend to look like: `1.16.0-beta4-12345`, where 12345 is the `TEST_ASSET_ID`.

_This doesn't work however, when `git describe` returns no tags, because the resulting version is `-12345` which is not a valid tag for docker images_

Example logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=1abb81c6-0985-4012-856d-d338c1936ccb
```
docker buildx build --load  /workspace/gloo/_output/projects/gloo -f /workspace/gloo/_output/projects/gloo/Dockerfile.gloo \
Step #0 - "publish-artifacts": Step #4 - "publish-docker": 	--build-arg GOARCH=amd64 \
Step #0 - "publish-artifacts": Step #4 - "publish-docker": 	--build-arg ENVOY_IMAGE=quay.io/solo-io/envoy-gloo:1.26.4-patch1 \
Step #0 - "publish-artifacts": Step #4 - "publish-docker": 	-t quay.io/solo-io/gloo:-8617 --label quay.expires-after=3w
Step #0 - "publish-artifacts": Step #4 - "publish-docker": ERROR: invalid tag "quay.io/solo-io/gloo:-8617": invalid reference format
```

## Manual Workaround
You just need to pull down the tags from the Gloo project:
```
git remote add gloo https://github.com/solo-io/gloo
git fetch gloo --tags
```

## Solution
Add logic to evalute the tags, and if the resulting value does not match semver regex, default to 1.0.0

## Interesting decisions
-

### From a Repo Without Tags
I created this PR from a newly forked repo, that does not contain any tags. So by nature of this passing CI, this proves the change works

Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=54f85a03-7942-48e5-aab2-72707ac84f0b
```
[...]

PR #8653, opened 2023-09-06
Fix Version Tagging on PRs

sam-heilbron:
# Description

Ensure that the VERSION variable, which corresponds to the tag for docker images that are built on the PR, includes the latest git tag, if one exists

# Context
This is follow up to https://github.com/solo-io/gloo/pull/8631. In that PR, we identified that forked repos, which do not contain git tags, were failing, because we did not properly identify when `git describe` returned an empty string. The testing steps included there worked on a local machine (mac) but did not verify the tag that our CI uses (linux). 

In a slack conversation (https://solo-io-corp.slack.com/archives/C03MFATU265/p1693594894391679?thread_ts=1693492511.156679&cid=C03MFATU265) the behavior of tagging in CI was identified. It did not break the behavior but was a confusing change.

This PR updates the logic to ensure that we abide by the following cases:
- If there are no tags available, we fallback to our `1.0.0-{TEST_ID}` tag. This will be unique per PR since the TEST_ID is unique to a PR, but should only occur on PRs from forked repos without all the tags pulled down
- If there are tags available (regardless of OS), the tag will fit the format `{GIT_TAG}-{TEST_ID}`

## Interesting decisions
I updated the semver we use in the regex (which I believe was failing on linux machines) to match what we use in our go-utils (leveraged by changelog-bot).

## Mac
```
TEST_ASSET_ID=123 make print-VERSION
```
returns
```
1.16.0-beta6-123
```

## Linux
In CI, you can see that the logs indicate what version is being used to tag assets:
```
-ldflags="-X github.com/solo-io/gloo/pkg/version.Version=1.16.0-beta7-8653"
```

## Notes for reviewers
-

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -595,7 +595,8 @@ package-chart: generate-helm-files
 # https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC59
 git_tag = $(shell git describe --abbrev=0 --tags)
 # Semantic versioning format https://semver.org/
-tag_regex := ^v([0-9]{1,}\.){2}[0-9]{1,}$
+# Regex copied from: https://github.com/solo-io/go-utils/blob/16d4d94e4e5f182ca8c10c5823df303087879dea/versionutils/version.go#L338
+tag_regex := v[0-9]+[.][0-9]+[.][0-9]+(-[a-z]+)*(-[a-z]+[0-9]*)?$
 
 ifneq (,$(TEST_ASSET_ID))
 PUBLISH_CONTEXT := PULL_REQUEST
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -595,14 +595,15 @@ package-chart: generate-helm-files
 # https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC59
 git_tag = $(shell git describe --abbrev=0 --tags)
 # Semantic versioning format https://semver.org/
-tag_regex := ^v([0-9]{1,}\.){2}[0-9]{1,}$
+# Regex copied from: https://github.com/solo-io/go-utils/blob/16d4d94e4e5f182ca8c10c5823df303087879dea/versionutils/version.go#L338
+tag_regex := v[0-9]+[.][0-9]+[.][0-9]+(-[a-z]+)*(-[a-z]+[0-9]*)?$
```
davidjumani, line 599: Assuming this is to handle beta tags as well

ashishb-solo (approved):

davidjumani (approved):
Verified in CI as well

PR #8658, opened 2023-09-07
Fix gateway yaml when using kube overrides

jenshu:
# Description

Fix a typo in the gateway yaml template.

# Checklist:


install/test/helm_test.go
```diff
@@ -6437,19 +6437,52 @@ metadata:
 			})
 
 			Context("Kube GatewayKube overrides", func() {
-				It("httpsGatewayKubeOverride allows to override ssl true value with false", func() {
+				It("can override values on default Gateways", func() {
```
jenshu, line 6440: confirmed that with the k8s-utils update (which enforces newlines after the `---` divider) and no other changes, these tests fail.
after fixing the default-gateways template, the tests succeed

ben-taussig-solo (approved):

install/helm/gloo/templates/_8-default-gateways.tpl
```diff
@@ -155,12 +155,12 @@ spec:
...
 {{- $defaultGatewayOverride := $spec.gatewaySettings.httpGatewayKubeOverride }}
 ---
-{{- include "gloo.util.merge" (list $ctx $defaultGatewayOverride "defaultGateway.gateway") -}}
+{{ include "gloo.util.merge" (list $ctx $defaultGatewayOverride "defaultGateway.gateway") -}}
 {{- end }}{{/* if not $gatewaySettings.disableHttpGateway */}}
 {{- if not $gatewaySettings.disableHttpsGateway }}
 {{- $sslGatewayOverride := $spec.gatewaySettings.httpsGatewayKubeOverride }}
 ---
-{{- include "gloo.util.merge" (list $ctx $sslGatewayOverride "defaultGateway.sslGateway") -}}
+{{ include "gloo.util.merge" (list $ctx $sslGatewayOverride "defaultGateway.sslGateway") -}}
```
inFocus7, line 163: QQ: why do we remove the leading `-` here for gateway + ssl gateway but not for others? 👀 (ex. failoverGateway right below)

jenshu: nice catch, this should have it as well

nfuden (approved):

Issue #8092, opened 2023-04-14
Gloo Edge 1.13.x with IstioSDS 1.14 doesn't mount spiffe volumes


edubonifs:
### Gloo Edge Version

1.13.x (beta)

### Kubernetes Version

None

### Describe the bug

Hi, 

We have a customer that wants to install Gloo Edge 1.13.8 with Istio integration, Istio version is 1.14.6. The problem is that they don't want all the pods from Gloo-system to have a sidecar injected, they only want the gateway-proxy pod to have a sidecar injected.

If you label Gloo-system namespace and install GE with istioSDS enabled everything works fine, but for installing GE with the sidecar injected on Gateway Proxy pod only, we found the following problem:

### Steps to reproduce the bug

If you install with GE with the following command:
```
helm install gloo glooe/gloo-ee --namespace gloo-system \
  --create-namespace --set-string license_key=LICENSE_KEY --version v1.13.8 --values values.yaml
```
And this values.yaml file:

```
global:
  istioIntegration:
    whitelistDiscovery: true
    enableIstioSidecarOnGateway: true
  istioSDS:
    enabled: true
    customSidecars:
      - name: istio-proxy
        image: 'docker.io/istio/proxyv2:1.14.6'
```

The gateway fails getting the certificates mounted:

```
{"level":"fatal","ts":"2023-04-14T13:18:38.721Z","logger":"sds_server","caller":"cmd/main.go:84","msg":"could not find file '/etc/istio-certs/key.pem'","version":"1.13.7","stacktrace":"main.main\n\t/workspace/gloo/projects/sds/cmd/main.go:84\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:250"}
```

The only possible way I have seen of injecting a sidecar proxy only to the gateway proxy without labeling the full namespace is using this values.yaml file, in which clients would need to specify spiffe volumes to be added:

```
global:
  istioIntegration:
    labelInstallNamespace: true
    whitelistDiscovery: true
    enableIstioSidecarOnGateway: true
  istioSDS:
    enabled: true
    customSidecars:
      - name: istio-proxy
        image: 'docker.io/istio/proxyv2:1.14.6'
        args:
          - proxy
          - sidecar
          - '--domain'
          - $(POD_NAMESPACE).svc.cluster.local
          - '--configPath'
          - /etc/istio/proxy
          - '--binaryPath'
          - /usr/local/bin/envoy
          - '--serviceCluster'
          - istio-proxy-prometheus
[...]

PR #8666, opened 2023-09-08
Istio spiffe volumes

bewebi:
# Description

Add volumes, mounted volumes, and env vars to GatewayProxy template as required to support newer versions of Istio

Also update the default Helm value for Istio image to be one we claim to support

## Helm changes
- Add `CANONICAL_SERVICE` and `CANONICAL_REVISION` env vars to GatewayProxy IstioProxy container, from corresponding labels
- Add `workload-socket` and `workload-secret` volumes to GatewayProxy and mount them at `/var/run/secrets/workload-spiffe-uds` and `/var/run/secrets/workload-spiffe-credentials` respectively in the IstioProxy container
- Update default `global.glooMtls.istioProxy.image.tag` value to `1.14.6`, as the the previous was no longer supported

# Context

These changes match exactly what is requested in the issue and what is required by newer versions of Istio

I was not able to find explicit documentation of what versions introduced the requirement to have the spiffe volumes, or even what they are...the best I found was [this page](https://istio.io/latest/docs/ops/integrations/spire/) which references one of the volumes
Instead I went off the example config given in the issue, which seems to be accurate

I do not believe additional documentation is necessary, as this is effectively a fix for supporting Istio versions we already claim to support and does not require any change in user behavior

This will be backported to 1.15, 1.14, and 1.13

## Interesting decisions
 
As previously stated, it's unclear exactly what Istio version starts to require the spiffe volumes, however the presence of the (initially empty) volumes should not break usage of Istio versions that did not require them

The earliest version of Istio we claim to support on any LTS version of Gloo is 1.11, so it made sense to bump the default value to a version we support
I chose 1.14.6 as it's the version specifically referenced in the issue and supported by all LTS versions, however I'm happy to change to a different value if we prefer to go with the minimum or maximum per LTS version

I added to Helm unit tests to test for newly added values, without adding tests for previously supported env vars, volumes, or mounts
It's preferable to have the strongest tests possible for new functionality, but seems out of scope to exhaustively test every aspect of existing functionality

## Testing steps

- Create a fresh KinD cluster and install Istio 1.14.6
- Installing Gloo from `main` with the following values set (in addition to the default values for everything else):
```
global:
  istioIntegration:
    whitelistDiscovery: true
    enableIstioSidecarOnGateway: true
  istioSDS:
    enabled: true
```
- Observe the expected error in the GatewayProxy `istio-proxy` container
- Helm uninstall `gloo` deployment, modify the template/checkout the `istio-spiffe-volumes` branch and Helm install from the local chart
- Observe the GatewayProxy pod comes up healthily

# Checklist:


install/helm/gloo/values-intended.yaml
```diff
@@ -0,0 +1,227 @@
+accessLogger:
```
jenshu, line 1: what is this file for?

install/helm/gloo/values-template.yaml
```diff
@@ -224,7 +224,7 @@ global:
       image:
         repository: proxyv2
         registry: docker.io/istio
-        tag: 1.9.5
+        tag: 1.14.6
```
sam-heilbron, line 227: Should this update to the default version of the istio proxy be exposed somewhere in our changelog entry?

sam-heilbron (commented):
Changes look good! Can you help me understand why the existing kubernetes e2e tests that we have (https://github.com/solo-io/gloo/tree/main/test/kube2e/gloomtls, https://github.com/solo-io/gloo/tree/main/test/kube2e/istio) around the istio integration  did not catch this error? I like the inclusion of the helm tests, but I think we'll also want some simple smoke tests to prove that all the components start up correctly when installed with a given version of Istio

install/helm/gloo/values-template.yaml
bewebi: I'm not sure...given that the existing value was no longer supported, it seemed unlikely that this is a value customers rely on heavily but I do not actually have context on the usage pattern
Happy to add an entry if it's deemed worthwhile

bewebi:
Having not looked _too_ closely at the tests themselves, the [helm values they use](https://github.com/solo-io/gloo/blob/main/test/kube2e/istio/artifacts/helm.yaml) do not override the default value
Therefore they were using an Istio version which did not require the spiffe volumes and my guess is that this issue was not caught for that reason
That said, I'm not sure why they were able to pass when 1.9.5 is seemingly unsupported itself

Happy to investigate that further, but not sure whether that would be considered in scope for this issue or worthy of a separate new issue.

sam-heilbron:
I confirmed the behavior using the following steps:

1. Create a kind cluster, and setup istio on it
```
VERSION=1.15.0 SKIP_DOCKER=true KUBE2E_TESTS=glooctl ISTIO_VERSION=1.14.6 ci/kind/setup-kind.sh
```
2. Build a Helm chart with the local changes. I cheat by specifying a version that exists, so that it installs images that are already published
```
VERSION=1.15.0 make -B build-test-chart
```
3. Create a values.yaml with the values provided
```
global:
  istioIntegration:
    whitelistDiscovery: true
    enableIstioSidecarOnGateway: true
  istioSDS:
    enabled: true
```
4. Install Gloo using the newest chart
```
helm install gloo _test/gloo-1.15.0.tgz -n gloo-system --create-namespace -f values.yaml
```
5. Confirm that the gateway-proxy is healthy and the underlying istio-proxy sidecar is as well
6. Upgrade the install to point to a recently published chart, and confirm it errors
```
helm upgrade gloo gloo/gloo --version 1.15.0 -n gloo-system -f values.yaml
```
When I perform this upgrade, I do get an error in the gateway-proxy, but it is not the same as the spiffe volume mount error. It seems more tied to the version of the istio-proxy that we run.
7. I updated the values.yaml to see if we just update the istio-proxy sidecar
```
global:
  istioIntegration:
    whitelistDiscovery: true
    enableIstioSidecarOnGateway: true
  istioSDS:
    enabled: true
  glooMtls:
    istioProxy:
      image:
        repository: proxyv2
        registry: docker.io/istio
        tag: 1.14.6
```
8. Upgrade the install to point to a recently published chart, but using the new values, and confirm it works
```
helm upgrade gloo gloo/gloo --version 1.15.0 -n gloo-system -f values.yaml
```

To me this means that the steps in the PR prove that upgrading the istio-proxy default tag fixes the issue. It may be that the spiffe volume mount is also "fixed", but I dont think we've ironed out the steps to reproduce that error.

bewebi:
Closing per [this comment](https://github.com/solo-io/gloo/issues/8092#issuecomment-1720244799)

Issue #8525, opened 2023-07-28
append: false not working for headerManipulation


edubonifs:
### Gloo Edge Version

1.15.x (beta)

### Kubernetes Version

None

### Describe the bug

It seems that [append](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-field-config-core-v3-headervalueoption-append) field is deprecated in envoy, I have a headerManipulation in my VirtualService and the same headerManipulation in my child RouteTable with append false, but they are getting appended any way, the header never gets overwritten, so it seems with the deprecation of this field this stopped working.

### Steps to reproduce the bug

Steps to reproduce:
```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: default
  namespace: gloo-system
spec:
  virtualHost:
    domains:
    - '*'
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
            key: X-Frame-Options
            value: SAMEORIGIN
    routes:
    - delegateAction:
        selector:
          labels:
            delegate: true
      matchers:
      - prefix: /
 --------
apiVersion: gateway.solo.io/v1
kind: RouteTable
metadata:
  labels:
    delegate: true
  name: child
  namespace: gloo-system
spec:
  routes:
  - matchers:
    - prefix: /get
    options:
      headerManipulation:
        responseHeadersToAdd:
        - append: false
          header:
            key: X-Frame-Options
            value: NEWORIGIN
    routeAction:
      single:
        upstream:
          name: default-httpbin-8000
          namespace: gloo-system
```

### Expected Behavior

Regarding the above VS and RT, X-Frame-Options header should be overwritten with value NEWORIGIN as I have append: true in the RouteTable, however it gets added to the previous header so in the response I am able to see both NEWORIGIN and SAMEORIGIN.

edubonifs:
Might be related to https://github.com/solo-io/gloo/issues/8477

SantoDE:
It's also happening on 1.14.9. Its based on potentially https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-field-config-core-v3-headervalueoption-append

Jimmmz:
From Crealogix 
```
We have upgraded to Gloo Edge 1.15.1 and were running into the issue https://github.com/solo-io/gloo/issues/8525
[...]

Issue #8634, opened 2023-09-01
Ability to append a value to a set of headers


edubonifs:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

1.14.x

### Kubernetes Version

1.25

### Describe the bug

According to the [docs](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/request_processing/append_remove_headers/), if I set append: true in a HeaderManipulation:
```
if the header HEADER_NAME is already present, append the value
```

### Expected Behavior

What I understand with this, is that if I have the following headerManipulation in place:
```
      options:
        headerManipulation:
          responseHeadersToAdd:
          - header:
              key: x-version
              value: fake version
            append: true
```

And I receive a request with the responseHeader x-version: 5, the expected outcome would be that the value get happened to the existing header:
```
x-version: 5 fake version
```

However, what is happening, is that a new header with the same name gets created:
```
x-version: 5
x-version: fake version
```

We have a client that wants to append a string to a set of "set-cookie" headers, so if I have two response headers set-cookie, and I want to append a string, for example "; Partitioned", I should be able to do that:
Response headers before HeaderManipulation:
```
set-cookie: TOKEN=eyJhbGciO...8; Version=1; Path=/blah/blah/; Secure; HttpOnly
set-cookie: AUTH_SESSION_ID=ec90fdc7-6f16-4ab1-bdfa-6a2478a1f25b; Version=1
```
Response headers after headerManipulation:
```
set-cookie: TOKEN=eyJhbGciO...8; Version=1; Path=/blah/blah/; Secure; HttpOnly; Partitioned
set-cookie: AUTH_SESSION_ID=ec90fdc7-6f16-4ab1-bdfa-6a2478a1f25b; Version=1; Partitioned
```
However, what I am getting is the following:
```
set-cookie: TOKEN=eyJhbGciO...8; Version=1; Path=/blah/blah/; Secure; HttpOnly
set-cookie: AUTH_SESSION_ID=ec90fdc7-6f16-4ab1-bdfa-6a2478a1f25b; Version=1; Path=/blah/blah/; SameSite=None; Secure; HttpOnly
set-cookie: ; Partitioned
```

### Steps to reproduce the bug

Create a VS with the following headerManipulation:
```
      options:
        headerManipulation:
          responseHeadersToAdd:
          - header:
              key: x-version
              value: fake version
            append: true
```

Check that a new response header is created instead of being appended

### Additional Context

[...]

PR #8678, opened 2023-09-12
Use appendAction for translated HeaderValueOptions

inFocus7:
# Description

Envoy has deprecated the `append` field: [ref](https://github.com/solo-io/gloo/issues/8525#issuecomment-1702610478). So we are now translating to the newer field `append_action` so users can keep the same functionality as before.

There are four `append_action` fields:
- APPEND_IF_EXISTS_OR_ADD
    - Appends value if the header exists. Else, adds the header with the key and value.
- ADD_IF_ABSENT
    - Adds the header if it doesn’t already exist. Else, no-op.
- OVERWRITE_IF_EXISTS_OR_ADD
    - Overwrites the value if the header already exists. Else, adds the header with the key and value.
- OVERWRITE_IF_EXISTS
    - ⁣Overwrite the header's value, if it exists. Else, no-op.

As of now, since we're just translating our existing `append` which is a `boolean` onto the `append_action`, we are mapping to two fields as follows:
- `append: true` => `append_action: APPEND_IF_EXISTS_OR_ADD`
- `append: false` => `append_action: OVERWRITE_IF_EXISTS_OR_ADD`

## API/Code changes
- `append` translates to Envoy's `append_action` 

# Context

~Users ran into issues where the `append` field was essentially being ignored.~
`append` has been deprecated in favor of `append_action`

## Testing steps

```bash
./ci/kind/setup-kind.sh
helm install gloo _test/gloo-1.0.0-ci.tgz
```
```bash
kubectl apply -f - << 'EOF' 
apiVersion: v1
kind: ServiceAccount
metadata:
  name: httpbin
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
spec:
  type: ClusterIP
  ports:
  - name: http
    port: 80
    targetPort: 80
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
    spec:
      serviceAccountName: httpbin
      containers:
      - image: docker.io/kennethreitz/httpbin
        imagePullPolicy: IfNotPresent
        name: httpbin
        ports:
        - containerPort: 80
EOF
```
```bash
kubectl apply -f - << 'EOF'
kind: VirtualService
metadata:
  name: default
  namespace: default
spec:
  virtualHost:
    domains:
    - '*'
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
           key: X-Frame-Options
           value: SAMEORIGIN
          append: false
    routes:
    - delegateAction:
        selector:
          labels:
            delegate: 'true'
      matchers:
      - prefix: /
EOF
```
```bash
kubectl apply -f - << 'EOF'
kind: RouteTable
metadata:
  labels:
    delegate: 'true'
  name: child
  namespace: default
spec:
  routes:
  - matchers:
    - prefix: /get
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
            key: X-Frame-Options
            value: NEWORIGIN
    routeAction:
      single:
        upstream:
          name: default-httpbin-80
          namespace: default
EOF
```
```
kubectl port-forward -n gloo-system deploy/gateway-proxy 8080:8080 & proxynormalPortFwdPid=$!
```

```
curl http://localhost:8080/get -v
```
Notice only the VS header (`SAMEORIGIN`) is in the header.

```
kubectl port-forward -n gloo-system deploy/gateway-proxy 19000:19000 & proxyAdminPortFwdPid=$!
```

Go to: http://localhost:19000/config_dump
Notice the `append_action: OVERWRITE_IF_EXISTS_OR_ADD ` in the Vhost. 

Note: If `append: true`, we'll notice both headers in the curl and we may not see `append_action` as. the `true` value is the default value.
Note: Setting `append: false` on the Route adds the `append_action: OVERWRITE_IF_EXISTS_OR_ADD` in Envoy, but due to the order that header manipulation gets processed, both headers are returned.
> Headers are appended to requests/responses in the following order: weighted cluster level headers, route level headers, virtual host level headers and finally global level headers.
> - [Envoy Docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers)

## Notes for reviewers

# Checklist:


sam-heilbron (commented):
Looking great! Some small thoughts for consolidation that I think we can do. As part of this, should we update our docs about this feature to clarify what the current behavior is?

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -247,12 +247,20 @@ func convertResponseHeaderValueOption(
 			return nil, CantSetHostHeaderError
 		}
 
+		// By default, or if `append` is set to true, we'll append the header if it exists.
```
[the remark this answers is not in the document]
inFocus7: I don't believe so, they're both (header + request response) different types which is why they have different conversion functions

projects/gloo/pkg/plugins/headers/plugin.go
```diff
@@ -247,12 +247,19 @@ func convertResponseHeaderValueOption(
 			return nil, CantSetHostHeaderError
 		}
 
+		appendAction := envoy_config_core_v3.HeaderValueOption_APPEND_IF_EXISTS_OR_ADD
```
sam-heilbron, line 250: nit: We might be able to remove:
```go
if header == nil {
  return nil, MissingHeaderValueError
}
if strings.HasPrefix(header.GetKey(), ":") {
  return nil, CantSetPseudoHeaderError(header.GetKey())
}
if strings.EqualFold(header.GetKey(), "Host") {
  return nil, CantSetHostHeaderError
}
```

and replace it with: https://github.com/solo-io/gloo/blob/main/pkg/utils/api_conversion/type.go#L76 (`CheckForbiddenCustomHeaders`). 
Not sure if the types line up but they seem to be performing the same validation

sam-heilbron (dismissed):

projects/gloo/pkg/plugins/headers/plugin.go
inFocus7: Yeah, I think that function wasn't used since the types don't line up. The header here is our internal `HeaderValueOption`, but the parameter expects the `HeaderValueOption` from Envoy

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -282,4 +283,13 @@ Headers can be inherited by children options, such as shown in the following exa
 4. Now, requests that match the route `/a/1` get the following headers:
    * The `x-gateway-start-time` request header is inherited from the parent VirtualHost option.
    * The `x-route-table` request header is set in the child Route option.
-   * The `x-route-table` response header in the child overwrites the parent object's value of `alphabet` instead to `a`, because child objects take precedence in case of conflict.
+   * The `x-route-table` response header in the parent overwrites the child object's value of `a` instead to `alphabet`, because child objects take precedence in case of conflict.
```
artberger, line 286: > because child objects take precedence in case of conflict.

Should this now say the parent object takes precedence?

artberger: or maybe we just remove that last part and let the description you added after explain it?
```suggestion
   * The `x-route-table` response header in the parent overwrites the child object's value of `a` instead to `alphabet`.
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -282,4 +283,13 @@ Headers can be inherited by children options, such as shown in the following exa
...
+
+Due to how header manipulations are processed, less specific headers overwrite more specific headers.
+
+From the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers):
+> Headers are appended to requests/responses in the following order: weighted cluster level headers, route level headers, virtual host level headers and finally global level headers.
+
+In our above example with `x-route-table`, since the virtual host level header is evaluated after the route level header, the virtual host level header overwrites the route level header.
+If we had set `append: true` on the virtual host, or removed it, then the route level header would have been appended to the virtual host level header.
```
artberger, line 294: By "removed it" do you mean remove the `append` field altogether? like not have it at all as in the original example?

artberger: Could we also provide an example of what appending would look like?
```suggestion
If you set `append: true` or omit this field on the virtual host, then the route level response header (`a`) would get appended to the virtual host level header (`alphabet`), such as `x-route-table: alphabeta`.
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -282,4 +283,13 @@ Headers can be inherited by children options, such as shown in the following exa
...
+If we had set `append: false` on the route, since the route is evaluated before the virtual host, it would not affect the virtual host.
```
artberger, line 295: similar comment
```suggestion
If you set `append: false` on the route, the route is evaluated before the virtual host and does not affect the virtual host. In the example, the response header would stay `x-route-table: alphabet`.
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -282,4 +283,13 @@ Headers can be inherited by children options, such as shown in the following exa
...
    * The `x-route-table` request header is set in the child Route option.
-   * The `x-route-table` response header in the child overwrites the parent object's value of `alphabet` instead to `a`, because child objects take precedence in case of conflict.
+   * The `x-route-table` response header in the parent overwrites the child object's value of `a` instead to `alphabet`, because child objects take precedence in case of conflict.
+
+Due to how header manipulations are processed, less specific headers overwrite more specific headers.
+
+From the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers):
+> Headers are appended to requests/responses in the following order: weighted cluster level headers, route level headers, virtual host level headers and finally global level headers.
+
+In our above example with `x-route-table`, since the virtual host level header is evaluated after the route level header, the virtual host level header overwrites the route level header.
```
artberger, line 293:
```suggestion
In the previous example of the `x-route-table` response header, the virtual host level header overwrites the route level header because the virtual host level header is evaluated after the route level header.
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
inFocus7: Ah yeah good callout, good to say `omit`!

As far as what the expected response, I'm not quite sure. I know using `curl`, appending leads to two headers being returned, each with the individual value.

_But_ I also saw [this issue](https://github.com/solo-io/gloo/issues/8634) that is pointing to the likelyhood that the above is broken. and instead of multiple of the same header, it should be `x-route-table: alphabet a` (for our example).

I left it a bit vague just referencing an append occuring, since I'm not sure if we have a bug causing this to append wonkily.

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
inFocus7: Do you think we could we keep the `since` or add a `because`? 

The reason this behaves the way it does is _because_ of the route is evaluated before virtual host. Without the "since", at least to me, it sounds like this evaluation is due to `append: false`, which would be wrong.

artberger: Yeah, that makes sense! I would just say to use because instead of since (since is usually for time, like since five minutes ago).

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
artberger: ok, yeah we can leave out the example if we have bugs, and just add the expected outcome after that is fixed.

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
inFocus7: Ah makes sense, thank you!

artberger (approved):

Issue #8690, opened 2023-09-15
Expose Envoy's most_specific_header_mutation_wins field


inFocus7:
### Gloo Edge Product

Open Source

### Gloo Edge Version

1.16

### Is your feature request related to a problem? Please describe.

The order of header mutations being evaluated goes from most specific to least specific, meaning setting mutations (such as `header_responses_to_add`, and `append` options) on least specific fields take priority over the most specific fields.

> Headers are appended to requests/responses in the following order: weighted cluster level headers, route level headers, virtual host level headers and finally global level headers.
> - [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers)

Relevant Issue: #8525 

In the case of
```
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: default
  namespace: gloo-system
spec:
  virtualHost:
    domains:
    - '*'
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
            key: X-Frame-Options
            value: SAMEORIGIN
    routes:
    - delegateAction:
        selector:
          labels:
            delegate: true
      matchers:
      - prefix: /
 --------
apiVersion: gateway.solo.io/v1
kind: RouteTable
metadata:
  labels:
    delegate: true
  name: child
  namespace: gloo-system
spec:
  routes:
  - matchers:
    - prefix: /get
    options:
      headerManipulation:
        responseHeadersToAdd:
        - append: false
          header:
            key: X-Frame-Options
            value: NEWORIGIN
    routeAction:
      single:
        upstream:
          name: default-httpbin-8000
          namespace: gloo-system
```

Since the VirtualHost's header manipulation gets evaluated **after** the Route's, the Route's `append` does not affect the VirtualHost.

### Describe the solution you'd like

We should expose Envoy's [most_specific_mutation_wins](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto#config-route-v3-routeconfiguration) in RouteConfiguration. This would reverse the order of evaluation, meaning that a more specific header would be able to overwrite a less specific one.

inFocus7:
Barebones Setup instructions (w/o routeConfiguration): https://github.com/solo-io/gloo/pull/8678

inFocus7:
# Release Information

Updating as code gets merged and released. Check-marking when released.

## OSS
- [ ] v1.16.0-beta10
- [ ] v1.15.6
[...]

PR #8692, opened 2023-09-15
Expose most_specific_header_mutations_wins

inFocus7:
# Description

## API/Code changes
- Exposed the `most_specific_header_mutation_wins` in **RouteConfigurations**.
- Add RouteTableOptions to TestContext's TestClients
- Add tests for most_specific_header_mutations_wins

# Docs changes
- Add section in header manipulation docs bringing up this flag and its behavior

# Context
- Users want a way to prioritize the mutation done by a route, instead of one done by a vhost. (more information on the issue #8690)

## Testing steps

```bash
./ci/kind/setup-kind.sh
```
```bash
helm install gloo _test/gloo-1.0.0-ci.tgz
```
```bash
kubectl apply -f - << 'EOF' 
apiVersion: v1
kind: ServiceAccount
metadata:
  name: httpbin
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
spec:
  type: ClusterIP
  ports:
  - name: http
    port: 80
    targetPort: 80
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
      version: v1
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
    spec:
      serviceAccountName: httpbin
      containers:
      - image: docker.io/kennethreitz/httpbin
        imagePullPolicy: IfNotPresent
        name: httpbin
        ports:
        - containerPort: 80
EOF
```
```bash
kubectl apply -f - << 'EOF'
kind: VirtualService
metadata:
  name: default
  namespace: default
spec:
  virtualHost:
    domains:
    - '*'
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
           key: X-Frame-Options
           value: VSORIGIN
          append: false
    routes:
    - delegateAction:
        selector:
          labels:
            delegate: 'true'
      matchers:
      - prefix: /
EOF
```
```bash
kubectl apply -f - << 'EOF'
kind: RouteTable
metadata:
  labels:
    delegate: 'true'
  name: child
  namespace: default
spec:
  routes:
  - matchers:
    - prefix: /get
    options:
      headerManipulation:
        responseHeadersToAdd:
        - header:
            key: X-Frame-Options
            value: RTORIGIN
          append: false
    routeAction:
      single:
        upstream:
          name: default-httpbin-80
          namespace: default
EOF
```
```
kubectl port-forward -n gloo-system deploy/gateway-proxy 8080:8080 & proxynormalPortFwdPid=$!
```

```
curl http://localhost:8080/get -v
```
Notice only the VS header (`VSORIGIN`) is in the header `x-frame-options`. This is because we have `append: false` on the header, therefore it is set to overwrite any headers that came before.

Edit the gateway resource to include `spec.routeOptions.mostSpecificHeaderMutationsWins: true` field.
```
curl http://localhost:8080/get -v
```
Notice that _now_, only the Route header (`RTORIGIN`) is in the header `x-frame-options`. This is because our route config is now set to prioritize the most specific header mutation.

## Notes for reviewers

# Checklist:


sam-heilbron (commented):
I think this is a great use case for a set of e2e tests which show how configuration at the route/vs level bheavior differently based on this setting (ie when are heads actually appended vs overwritten). Since this functionality depends on multiple APIs in the data plane, unit tests alone won't be able to verify this.
 The test could then be used as the source of truth for our docs.

projects/gloo/api/v1/options.proto
```diff
@@ -110,6 +110,12 @@ message RouteConfigurationOptions {
     // Please refer to the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto#envoy-v3-api-field-config-route-v3-routeconfiguration-max-direct-response-body-size-bytes)
     // for more details about the `max_direct_response_body_size_bytes` attribute.
     google.protobuf.UInt32Value max_direct_response_body_size_bytes = 1;
+
+    // By default, headers that should be added/removed are evaluated from most to least specific:
+    // To allow setting overrides at the route or virtual host level, this order can be reversed by setting this option to true.
+    // Refer to the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto#envoy-v3-api-field-config-route-v3-routeconfiguration-most-specific-header-mutations-wins) for more details.
+    // Defaults to false.
+    bool most_specific_header_mutations_wins = 2;
```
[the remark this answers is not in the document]
inFocus7: Hm we could set it to a boolValue, but I used bool since by default it should be `false` (which a bool does) and to match Envoy's field as well https://github.com/envoyproxy/envoy/blob/2cdc11065126a8cbd0d258acd5f6e4f7da4205b1/api/envoy/config/route/v3/route.proto#L94

changelog/v1.16.0-beta10/expose-most-specific-header-mutations-wins.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: NEW_FEATURE
+    description: |
+      Add support for `most_specific_header_mutations_wins` to allow the more specific header mutation to win. Toggling it on would reverse the evaluation order of the header mutations, allowing more specific header mutations to overwrite less specific.
+      https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto#envoy-v3-api-field-config-route-v3-routeconfiguration-most-specific-header-mutations-wins
+    issueLink: https://github.com/solo-io/gloo/issues/8690
+    resolvesIssue: false
+  - type: NON_USER_FACING
+    description: |
+      skipCI-kube-tests:true
```
sam-heilbron, line 10: I think we should be cautious about overusing this directive. Even though we are minimally modifying our translation logic, I think we should still run our kube tests just to be safe.

projects/gloo/api/v1/options.proto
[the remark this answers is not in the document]
sam-heilbron: I agree with @inFocus7 that I'm not sure we need to differentiate false and unset, since the default value is false, though I could see this working either way. The benefit to having the primitive value, is the minimal encoding/decoding benefit that exists from not having to distinguish nil.

test/e2e/headers_test.go
```diff
@@ -188,4 +198,165 @@ var _ = Describe("HeaderManipulation", func() {
...
+				g.Expect(resp.StatusCode).To(Equal(http.StatusOK))
+				respHeaders := resp.Header.Values(responseHeader)
+				g.Expect(respHeaders).To(ContainElements(expectedHeaders))
+				select {
+				case received := <-testContext.TestUpstream().C:
+					reqHeaders := received.Headers
+					g.Expect(reqHeaders).To(HaveKeyWithValue(textproto.CanonicalMIMEHeaderKey(requestHeader), ContainElements(expectedHeaders)))
+				case <-time.After(time.Second * 5):
+					Fail("request didn't make it upstream")
+				}
```
sam-heilbron, line 312: Could we consolidate this to follow what we do in another test that asserts header behavior: https://github.com/solo-io/gloo/blob/main/test/e2e/aws_test.go#L147

test/e2e/headers_test.go
```diff
@@ -188,4 +198,165 @@ var _ = Describe("HeaderManipulation", func() {
...
+			testContext.ResourcesToCreate().RouteTables = v1.RouteTableList{overwriteRT, appendRT}
+		})
+
+		JustBeforeEach(func() {
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				return testContext.TestClients().VirtualServiceClient.Read(writeNamespace, "append-vs", clients.ReadOpts{})
+			})
+			helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+				return testContext.TestClients().VirtualServiceClient.Read(writeNamespace, "overwrite-vs", clients.ReadOpts{})
+			})
```
sam-heilbron, line 333: Does this provide us something extra? The TestContext is intended to create all resources and wait for a Proxy to exist. Did you notice that tests were behaving differently without this logic?

test/services/gloo.go
```diff
@@ -441,6 +441,8 @@ func getTestClients(ctx context.Context, bootstrapOpts bootstrap.Opts) TestClien
 	Expect(err).NotTo(HaveOccurred())
 	proxyClient, err := gloov1.NewProxyClient(ctx, bootstrapOpts.Proxies)
 	Expect(err).NotTo(HaveOccurred())
+	routeTableClient, err := gatewayv1.NewRouteTableClient(ctx, bootstrapOpts.RouteTables)
```
sam-heilbron, line 444: oh nice!

test/e2e/headers_test.go
```diff
@@ -188,4 +198,165 @@ var _ = Describe("HeaderManipulation", func() {
...
 		})
 	})
+
+	Context("Header mutation order", func() {
+		const (
+			responseHeader = "response-header"
+			requestHeader  = "request-header"
+		)
+
+		rtWithOptions := func(append bool) *v1.RouteTable {
```
sam-heilbron, line 208: nit: If you wanted, we could create a very simply RouteTableBuilder, which mirrors the existing VirtualServiceBuilder, and just exposes methods for the settings we currently need. That way others could expand the usage as they see fit

sam-heilbron (commented):
Looking good!

projects/gloo/api/v1/options.proto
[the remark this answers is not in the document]
jenshu: another vote here for BoolValue, this allows us to inherit the envoy default (whether it's true or false) if it's unset in the gloo config.
also to avoid potential future issues like https://github.com/solo-io/gloo/issues/6476 that need to distinguish between false and unset for overriding purposes

test/e2e/headers_test.go
inFocus7: It's possible to do so for the respose (which I did in latest commit), but not request as we don't have any matchers and transformations set for requests.

Spent about 30 minutes creating a request matcher, but was annoying as some of our other matchers (ex. `ContainHeaders`) are currently using transforms specifically for _responses_. They would fail to be used for the requests unless more refactoring was done, which was out-of-scope for this work.

test/e2e/headers_test.go
inFocus7: Removed 👍🏼

test/e2e/headers_test.go
inFocus7: Created a `RouteTableBuilder` + `RouteBuilder` with what's needed for this / for now

projects/gloo/api/v1/options.proto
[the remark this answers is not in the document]
inFocus7: Pushed updates to use BoolValue, defaulting to `false`

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -293,3 +293,52 @@ From the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/
...
 If you set `append: true` or omit this field on the virtual host, then the route level response header (`a`) would get appended to the virtual host level header (`alphabet`).
 If you set `append: false` on the route, the route does not affect the virtual host because the route is evaluated before the virtual host. In the example, the response header would stay `x-route-table: alphabet`.
+
+### Reversing the order of header manipulation evaluation
+
+You can reverse the order in which header manipulations are evaluated so that order of evaluation becomes: global level headers, virtual host level headers, route level headers, and finally weighted cluster level headers.
+With the order of evaluation being reversed, more specific header manipulations can override less specific ones.
+This can be done by setting the `mostSpecificHeaderMutationsWins` field to `true` in the [routeOptions]({{< versioned_link_path fromRoot="/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk/#routeconfigurationoptions" >}}) settings for a [Gateway]({{< versioned_link_path fromRoot="/reference/api/github.com/solo-io/gloo/projects/gateway/api/v1/gateway.proto.sk/" >}}).
```
artberger, line 301: Break up paragraphs:
```suggestion
With the order of evaluation being reversed, more specific header manipulations can override less specific ones.

To reverse the order of evaluation, set the `mostSpecificHeaderMutationsWins` field to `true` in the [routeOptions]({{< versioned_link_path fromRoot="/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk/#routeconfigurationoptions" >}}) settings for a [Gateway]({{< versioned_link_path fromRoot="/reference/api/github.com/solo-io/gloo/projects/gateway/api/v1/gateway.proto.sk/" >}}).
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -293,3 +293,52 @@ From the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/
...
+apiVersion: gateway.solo.io/v1
+kind: Gateway
+metadata: # collapsed for brevity
+spec:
+  routeOptions:
+    mostSpecificHeaderMutationsWins: true
+status: # collapsed for brevity
+```
+
+Using our VirtualService and RouteTable setup above, we can set the `append: false` on the RouteTable:
```
artberger, line 315:
```suggestion
Following the VirtualService and RouteTable setup in the previous section, add the `append` setting to `false` on the RouteTable. This way, the header values added by the RouteTable override less specific headers instead of append to the values.
```

docs/content/guides/traffic_management/request_processing/append_remove_headers/_index.md
```diff
@@ -293,3 +293,52 @@ From the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/configuration/
...
+            key: x-route-table
+            value: a
+      responseHeadersToAdd:
+        - header:
+            key: x-route-table
+            value: a
+          append: false
+{{< /highlight >}}
+
+Now, because we have `mostSpecificHeaderMutationsWins` set to `true`, the `x-route-table` response header in the child overwrites the parents object's value of `alphabet` instead to `a`.
```
artberger, line 344:
```suggestion
With `mostSpecificHeaderMutationsWins` set to `true` and `append` set to `false`, now the `x-route-table` response header in the child RouteTable overwrites the parent object's value of `alphabet` instead to `a`.
```

test/gomega/matchers/contain_headers.go
```diff
@@ -25,6 +25,21 @@ func ContainHeaders(headers http.Header) types.GomegaMatcher {
 	return gomega.And(headerMatchers...)
 }
 
+// ConsistOfHeaders produces a matcher that will only match if all provided headers are completely accounted for, including multi-value headers.
+// This matcher will fail if there are any extra headers that are not specified in the headers passed in.
+func ConsistOfHeaders(headers http.Header) types.GomegaMatcher {
```
sam-heilbron, line 30: Excellent! I really like the expansion of our shared matchers here. Do you think it's worth it to add a unit test to demonstrate that this utility works as expected?

sam-heilbron (commented):
🚀

test/e2e/headers_test.go
sam-heilbron: Ohh interesting, I wasn't aware of this. It makes sense that often we validate that responses match, but this is a good use case where we need to validate request behavior. One option is that we have our server that handles the requests actually echo the headers in the response. Alternatively, I like the idea of asserting on the request itself. Do you want to discuss how to do this or do you have a  good sense / feel that it's covered now

test/e2e/headers_test.go
```diff
@@ -188,4 +194,163 @@ var _ = Describe("HeaderManipulation", func() {
...
+					Custom:     testmatchers.ConsistOfHeaders(http.Header{responseHeader: expectedResHeaders}),
+				}))
+
+				select {
+				case req := <-testContext.TestUpstream().C:
+					g.Expect(req.Headers).To(HaveKeyWithValue(requestHeader, ConsistOf(expectedReqHeaders)))
+				case <-time.After(time.Second * 5):
+					Fail("request didn't make it upstream")
+				}
+			}, "5s", "0.5s").Should(Succeed())
```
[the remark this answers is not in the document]
inFocus7: This tests the the request itself has the correct headers from RequestHeadersToAdd, which is gotten through the upstream. The other check is for the response headers (to add).

projects/gloo/pkg/translator/route_config.go
```diff
@@ -85,11 +85,16 @@ type httpRouteConfigurationTranslator struct {
 
 func (h *httpRouteConfigurationTranslator) ComputeRouteConfiguration(params plugins.Params) []*envoy_config_route_v3.RouteConfiguration {
 	params.Ctx = contextutils.WithLogger(params.Ctx, "compute_route_config."+h.routeConfigName)
+	mostSpecificHeaderMutationsWins := false
+	if mostSpecificVal := h.parentListener.GetRouteOptions().GetMostSpecificHeaderMutationsWins(); mostSpecificVal != nil {
+		mostSpecificHeaderMutationsWins = mostSpecificVal.GetValue()
+	}
```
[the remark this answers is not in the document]
sam-heilbron: We spoke during standup and @nfuden agrees that if the user hasn't set the value, then the gloo api won't configure the underlying data plane api. So the logic will be:
- Unset -> Unset
- True -> True
- False -> False

test/e2e/headers_test.go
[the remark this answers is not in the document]
sam-heilbron: ohhhh I see! Thanks for clarifying, and good catch

test/gomega/matchers/contain_headers.go
inFocus7: Added some unit tests!

projects/gloo/pkg/translator/route_config.go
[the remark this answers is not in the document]
inFocus7: Updated the logic so we only set if when the field is not unset 👍🏼

sam-heilbron (approved):
Great work! I confirmed that the tests fail when I remove the translation code, and also when I adjust the value of this field

artberger (approved):

PR #8695, opened 2023-09-15
Rotate certs without downtime

elcasteel:
# Description

The certgen job now rotates certs in stages to minimize downtime when mTLS is used.

The mtls cert secret is updated in the following steps:
1. Set current secret's ca bundle to the current + next ca bundle (so both CAs are accepted temporarily)
2. Wait for a configurable time for the change to propagate
3. Set the current secret's server cert and private key to those of the newly generated certs
4. Wait for a configurable time for the change to propagate
5. Set the current secret's ca bundle to the next ca bundle. Now it contains only the next server cert and next ca bundle and the old ones are no longer supported.

## Code changes
The certgen job was updated to do cert rotation in stages as described above.

## Helm changes
A new helm value `gateway.certGenJob.rotationDuration` (default `65s`) was added to configure the duration to wait for each stage of cert rotation to complete (as described above in steps 2 and 4). This value must be tuned according to the performance of the specific environment in which the jobs are run.

# Context

A customer using mTLS and ExtAuth observed brief outages every time the mtls certgen job ran. It was likely because not every pod (gateway-proxy, extauth, etc) received secret updates at the same time, so when some pods had the new ca/cert, they couldn't communicate with pods with the old ca/cert. Updating the certs in stages so that for a brief period both ca's are accepted should minimize the downtime as long as the rotation duration is tuned appropriately to give all pods time to pick up the update.

## Testing steps

Here are steps that can be used to verify the fix locally, from this (unmerged) branch:

```
# create a cluster with kubelet config that makes changes sync faster (makes testing faster…otherwise secret changes can take 60seconds+ to propagate)
cat <<EOF | kind create cluster --name test --image kindest/node:v1.27.3 --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
  - |
    apiVersion: kubelet.config.k8s.io/v1beta1
    kind: KubeletConfiguration
    syncFrequency: 2s
    fileCheckFrequency: 2s
EOF

# build the sds and certgen images (which are the 2 modules that have been modified in this PR)
MY_TAG=0.0.0-test1
make VERSION=$MY_TAG sds-docker -B
make VERSION=$MY_TAG certgen-docker -B
kind load docker-image quay.io/solo-io/sds:$MY_TAG --name test
kind load docker-image quay.io/solo-io/certgen:$MY_TAG --name test

# install a release version of Gloo Edge but with the locally-built certgen and sds images (can also build the whole chart locally but this is a bit faster):
helm install -n gloo-system gloo-ee gloo-ee/gloo-ee --create-namespace --set-string license_key=$GLOO_LICENSE_KEY --version v1.16.0-beta1 -f - <<EOF
gloo:
  gateway:
    certGenJob:
      image:
        tag: $MY_TAG
    persistProxySpec: true
    logLevel: info
    validation:
      allowWarnings: false
      alwaysAcceptResources: false
gloo-fed:
  enabled: false
  glooFedApiserver:
    enable: false
global:
  glooMtls:
    enabled: true
    sds:
      image:
        tag: $MY_TAG
EOF

# set up extauth with basic auth (mostly copied from the guide https://docs.solo.io/gloo-edge/latest/guides/security/auth/extauth/basic_auth/)
kubectl apply -f - <<EOF
apiVersion: enterprise.gloo.solo.io/v1
kind: AuthConfig
metadata:
  name: basic-auth
  namespace: gloo-system
spec:
  configs:
  - basicAuth:
      apr:
        users:
          user:
            salt: "TYiryv0/"
            hashedPassword: "8BvzLUO9IfGPGGsPnAgSu1"
EOF

k apply -f - <<EOF
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: json-upstream
  namespace: gloo-system
spec:
  static:
    hosts:
      - addr: jsonplaceholder.typicode.com
        port: 80
EOF

# wait a couple seconds for the upstream to be picked up before applying this VS
kubectl apply -f - <<EOF
kind: VirtualService
metadata:
  name: auth-tutorial
  namespace: gloo-system
spec:
  virtualHost:
    domains:
      - 'foo'
    routes:
      - matchers:
        - prefix: /
        routeAction:
          single:
            upstream:
              name: json-upstream
              namespace: gloo-system
        options:
          autoHostRewrite: true      
    options:
      extauth:
        configRef:
          name: basic-auth
          namespace: gloo-system
EOF

# (in a new tab) port-forward
k port-forward -n gloo-system deploy/gateway-proxy 8080

# (in another new tab) continuously send requests (they should all return 200s)
while true
  do curl -v -H "Authorization: basic dXNlcjpwYXNzd29yZA==" -H "Host: foo" http://localhost:8080/posts/1
  sleep 1
done

# if desired, save a copy of the current mtls cert so we can compare it to after
k get secret -n gloo-system gloo-mtls-certs -oyaml > cert-before.yaml

# (in first tab) run the mtls certgen job
# job can either be run manually from the command line (this way allows us to more easily observe the logging output as the job runs)
go run jobs/certgen/cmd/main.go --svc-namespace=gloo-system --svc-name=gloo --force-rotation=true --secret-name=gloo-mtls-certs --secret-namespace=gloo-system --rotation-duration=10s

# or can be applied as yaml
k apply -f - <<EOF
apiVersion: batch/v1
kind: Job
metadata:
  labels:
    app: gloo
    gloo: gloo-mtls-certgen
  name: gloo-mtls-certgen-test
  namespace: gloo-system
spec:
  ttlSecondsAfterFinished: 60
  template:
    metadata:
      labels:
        gloo: gloo-mtls-certs
        sidecar.istio.io/inject: "false"
    spec:
      serviceAccountName: certgen
      restartPolicy: OnFailure
      containers:
        - image: quay.io/solo-io/certgen:$MY_TAG
          imagePullPolicy: IfNotPresent
          name: certgen
          securityContext:
            runAsNonRoot: true
            runAsUser: 10101
          env:
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
          args:
            - "--secret-name=gloo-mtls-certs"
            - "--svc-name=gloo"
            - "--rotation-duration=10s"
            - "--force-rotation=true"
EOF

# verify that we continue seeing only 200 responses from the curl (keep it running for awhile even after the certgen job finishes)

# check the secret and compare to cert-before.yaml to verify it has been updated
k get secret -n gloo-system gloo-mtls-certs -oyaml > cert-after.yaml
```

# Checklist:


jobs/pkg/kube/secret.go
```diff
@@ -25,7 +25,7 @@ type TlsSecret struct {
 // If there is a currently valid TLS secret with the given name and namespace, that is valid for the given
 // service name/namespace, then return it. Otherwise return nil.
 func GetExistingValidTlsSecret(ctx context.Context, kube kubernetes.Interface, secretName string, secretNamespace string,
-	svcName string, svcNamespace string, renewBeforeDuration time.Duration) (*v1.Secret, error) {
+	svcName string, svcNamespace string, renewBeforeDuration time.Duration) (*v1.Secret, bool, error) {
```
jenshu, line 28: can you document what the new `bool` return value indicates?

jobs/pkg/run/run.go
```diff
@@ -53,6 +55,9 @@ func Run(ctx context.Context, opts Options) error {
 	if opts.ServerKeySecretFileName == "" {
 		return eris.Errorf("must provide name for the server key entry in the secret data")
 	}
+	if opts.NextSecretName == "" {
+		return eris.Errorf("must provide secret-name")
```
jenshu, line 59:
```suggestion
		return eris.Errorf("must provide next-secret-name")
```

jobs/pkg/kube/secret.go
```diff
@@ -111,7 +112,36 @@ func CreateTlsSecret(ctx context.Context, kube kubernetes.Interface, secretCfg T
...
+	secret1.Cert = secret2.Cert
+	secret1.PrivateKey = secret2.PrivateKey
+	secretToWrite := makeTlsSecret(secret1)
+	_, err := secretClient.Update(ctx, secretToWrite, metav1.UpdateOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed updating current private key")
+	}
+	// wait for SDS
+	contextutils.LoggerFrom(ctx).Infow("Wrote new cert, waiting to rotate CaBundles")
+	time.Sleep(time.Second * 20)
```
elcasteel, line 127: TODO: there's a production appropriate way to sleep.
Consider configuring time spent sleeping

projects/sds/pkg/server/server_test.go
```diff
@@ -27,12 +27,23 @@ var _ = Describe("SDS Server", func() {
...
-
+	const (
+		CertData = `
+-----BEGIN CERTIFICATE----- 
+testdata
+-----END CERTIFICATE-----
+`
+		ExtraCaData = `
+-----BEGIN CERTIFICATE----- 
+testdata
```
elcasteel, line 38: make these different

jobs/pkg/run/run.go
```diff
@@ -61,30 +66,64 @@ func Run(ctx context.Context, opts Options) error {
...
-		// check if there is an existing valid TLS secret
-		secret, err = kube.GetExistingValidTlsSecret(ctx, kubeClient, opts.SecretName, opts.SecretNamespace,
-			opts.SvcName, opts.SvcNamespace, renewBeforeDuration)
+	// check if there is an existing valid TLS secret
+	secret, renewCurrent, err := kube.GetExistingValidTlsSecret(ctx, kubeClient, opts.SecretName, opts.SecretNamespace,
+		opts.SvcName, opts.SvcNamespace, renewBeforeDuration)
+	if err != nil {
+		return eris.Wrapf(err, "failed validating existing secret")
+	}
+	nextSecret, renewNext, err := kube.GetExistingValidTlsSecret(ctx, kubeClient, opts.NextSecretName, opts.SecretNamespace,
```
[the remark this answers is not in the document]
nfuden: yeah definitely makes sense. We dont want a small delay to affect the intermediate cert and we can specify the time period as to which it will survive for so that users know what the total exposure they have is

jobs/certgen/cmd/main.go
```diff
@@ -57,6 +59,8 @@ func cmd(ctx context.Context) *cobra.Command {
 		"name of the ValidatingWebhookConfiguration to patch with the generated CA bundle. leave empty to skip this step.")
 	pFlags.BoolVar(&opts.ForceRotation, "force-rotation", false,
 		"if set, will create new certs even if the old one are still valid")
+	pFlags.StringVar(&opts.RotationDuration, "rotation-duration ", "10s",
+		"time duration string value which specifies the total delay expected while swapping certs, if too short will cause requests to potentially be dropped")
```
nfuden, line 63:
```suggestion
	pFlags.StringVar(&opts.RotationDuration, "rotation-duration", "10s",
		"time duration string value which specifies the total delay expected while swapping certs, if too short will cause requests to potentially be dropped. If activedeadlineseconds is set make sure this is less")
```

projects/sds/pkg/server/server.go
```diff
@@ -169,9 +168,41 @@ func readAndVerifyCert(certFilePath string) ([]byte, error) {
...
+		return fileBytes, err
+	}
+
 	return fileBytes, nil
 }
 
+// checkCert uses pem.Decode to verify that the given
+// bytes are not malformed, as could be caused by a
+// write-in-progress.
+// func checkCert(ctx context.Context, certs []byte) bool {
```
ben-taussig-solo, line 182: nit: we need to uncomment this before proceeding

jobs/pkg/kube/secret.go
```diff
@@ -112,6 +128,97 @@ func CreateTlsSecret(ctx context.Context, kube kubernetes.Interface, secretCfg T
 	return createdSecret, nil
 }
 
+// RotateSecrets rotates secrets in a few steps.
```
ben-taussig-solo, line 131: thanks for this comment, it was very helpful for understanding the rotation strategy

nfuden:
For build bot we need a slight update otherwise it looks good

projects/sds/pkg/server/server.go
jenshu: was hoping to add more strict checking but it made tests fail so backed out that change

nfuden (approved):

ben-taussig-solo (approved):
Works as described, with no downtime

Issue #8288, opened 2023-05-25
[Helm Chart] custom readiness probe with custom healthCheck path not working


jrauschenbusch:
### Gloo Edge Version

1.14.x (latest stable)

### Kubernetes Version

1.24.x

### Describe the bug

Deploying the Gloo Community via the Helm chart and Flux 2 fails when using a custom readiness probe.

Reason is that the rollout job is triggered via a Helm post-install hook which is creating the `Gateway` resources where the changed healthCheck path is configured. If one changes the healthCheck path and also wants to apply a customReadinessProbe as mentioned in https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/#configuring-the-kubernetes-probes this will fail. Reason is that first the Flux HelmRelease must get healthy before the hooks are executed. But when the rollout job is not executed, the re-configuration of the healthCheck path is not applied.

Workaround is either to use the default /healthz endpoint and not modifying the healthCheck path at all, or to disable the customReadinessProbe.

Maybe this issue is heavily related to Flux, but it is widely adopted and hence maybe others will have the problem too.

Are there any other options? Changing the HelmRelease health check may also work.

### Steps to reproduce the bug

```yaml
gatewayProxies:
  gatewayProxy:
    gatewaySettings:
      customHttpGateway:
        options:
          healthCheck:
            path: /envoy-hc # custom path that is available when the Gateway (Envoy listener) is actually listening
    service:
      externalTrafficPolicy: Local
      extraAnnotations:
        service.beta.kubernetes.io/azure-load-balancer-internal: "false"
        service.beta.kubernetes.io/azure-load-balancer-resource-group: "foobar"
        service.beta.kubernetes.io/azure-load-balancer-health-probe-protocol: "http"
        service.beta.kubernetes.io/azure-load-balancer-health-probe-request-path: "/envoy-hc"
        service.beta.kubernetes.io/azure-load-balancer-health-probe-interval: "10"
        service.beta.kubernetes.io/azure-load-balancer-health-probe-num-of-probe: "2"
    podTemplate:
      terminationGracePeriodSeconds: 7
      gracefulShutdown:
        enabled: true
        sleepTimeSeconds: 5
      probes: true
      customReadinessProbe: # fails as rollout job is not triggered when HelmRelease gets not healthy, but to get healthy readiness probe must succeed
        httpGet:
          scheme: HTTP
          port: 8080
          path: /envoy-hc
        failureThreshold: 2
[...]

PR #8698, opened 2023-09-19
fix: custom readiness probe with custom healthCheck path not working

davidjumani:
# Description

Fixes https://github.com/solo-io/gloo/issues/8288

When following the instructions [here](https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/), a user defines a custom health check on the default gateway (which is deployed in the post-install hook) and defines a readiness probe on the gateway proxy.
During a helm install with `--wait` specified, the gateway proxy must come up and be ready before moving to execute the post-install hooks that deploy the default gateway. As a result, the gateway proxy is never ready since the config to create the health check endpoint is defined in the default gateway. 

This attempts to fix that issue by moving the resource rollout job out of the post install hook. Since the resource rollout is a job, it is one of the last resources to be applied via helm (https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/)
Hence when the jobs runs, it waits for the gloo deployment to complete and then applies the default resources
It also adds a default virtual service not having any VS leads to no listener created (https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/translator/http_translator.go#L28) - in cases of a clean install

## Helm changes
- Removed the post install / upgrade hooks from the resource rollout job
- Added a default VS

# Context

https://github.com/solo-io/gloo/issues/8288

## Testing steps :

This works on both upgrades && clean installs

values.yaml
```
virtualService:
  enabled: true
gateway:
  validation:
    allowWarnings: false
    alwaysAcceptResources: false
    failurePolicy: Fail
gatewayProxies:
  gatewayProxy:
    service:
      type: ClusterIP    # Since the test is running in kind
    gatewaySettings:
      customHttpGateway:
        options:
          healthCheck:
            path: /envoy-hc
    podTemplate:
      terminationGracePeriodSeconds: 7
      gracefulShutdown:
        enabled: true
        sleepTimeSeconds: 5
      probes: true
      customReadinessProbe:
        httpGet:
          scheme: HTTP
          port: 8080
          path: /envoy-hc
        failureThreshold: 2
        initialDelaySeconds: 5
        periodSeconds: 5
```

Before :
```
helm install gloo gloo/gloo --version v1.16.0-beta9 --create-namespace  --namespace gloo-system  --set gloo-fed.enabled=false --devel --values values.yaml --wait

Error: INSTALLATION FAILED: context deadline exceeded

# The gateway proxy pod is not ready
gloo-system          pod/gateway-proxy-67c67c84cd-7522j               0/1     Running   0          21m

```

After :
```
VERSION=1.16.0-beta9 make build-test-chart
helm install gloo _test/gloo-1.16.0-beta9.tgz --create-namespace  --namespace gloo-system  --set gloo-fed.enabled=false --devel --values values.yaml --wait

NAME: gloo
LAST DEPLOYED: Tue Sep 19 15:04:19 2023
NAMESPACE: gloo-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8288


install/helm/gloo/templates/5-resource-rollout-job-clusterrole.yaml
```diff
@@ -8,10 +8,6 @@ metadata:
   labels:
     app: gloo
     gloo: rbac
-  annotations:
-    "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "0" # must be created before rollout job
```
nfuden, line 13: There previously was ordering to avoid churn. Are we sure that this wont cause the rollout job to sometimes fail?

davidjumani: It should not because :
- It is one of the last resources to be created via helm
- The resource rollout job explicitly waits for the gloo deployment to be rolled out 
- We can run helm with `--wait-for-jobs` to ensure helm waits for the job to complete successfully or mark the release as failed

install/helm/gloo/templates/5-resource-configmap.yaml
```diff
@@ -11,4 +11,13 @@ data:
...
   custom-resources: |
 {{ $cr }}
+    ---
+    apiVersion: gateway.solo.io/v1
+    kind: VirtualService
+    metadata:
+      name: default
+      namespace: gloo-system
+    spec:
+      virtualHost: {}
```
davidjumani, line 21: Added for clean installs since not having any VS leads to no listener created (https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/translator/http_translator.go#L28)

install/helm/gloo/templates/5-resource-rollout-job-clusterrole.yaml
davidjumani: In case of Argo, we can add this job in a latter wave to ensure that it is one of the last to run

install/helm/gloo/templates/5-resource-configmap.yaml
jenshu: is there any way we can gate this so it's only created if readiness probe is enabled?
also i think the vs should go in a separate template similar to `_8-default-gateways.tpl`

davidjumani: Yes, will move it there along with the checks. Just kept it here for quick testing

install/helm/gloo/values-template.yaml
```diff
@@ -101,7 +104,7 @@ gateway:
       repository: kubectl
     restartPolicy: OnFailure
     runAsUser: 10101
-    ttlSecondsAfterFinished: 60
+    ttlSecondsAfterFinished: 300
```
davidjumani, line 107: When adding the `--wait` && `--wait-for-jobs` flag in helm, sometimes the job can complete and be deleted before helm can check if it has finished (for instance, while helm is waiting on a deployment). So increasing this to a higher value where it should still be around after all the other resources are ready

install/helm/gloo/templates/5-resource-rollout-job-check.yaml
```diff
@@ -0,0 +1,106 @@
...
+              then
+                echo "The following resources failed to be applied in the resource rollout job $FAILED_RESOURCES"
+                exit 1
+              else
+                echo "The resource rollout job successfully applied all the resources"
+              fi
+            else
+              echo "no custom resources to check"
+            fi
+            # TODO: Cleanup RBAC created for the rollout job
```
davidjumani, line 105: With this PR, the post-hook jobs & RBAC for the resource-rollout are removed from the post-install hook, so there is no automatic deletion policy. Looking for ideas on how to remove the obsolete RBAC resources

install/helm/gloo/templates/5-resource-rollout-job-check.yaml
```diff
@@ -0,0 +1,106 @@
...
+              if [ ! -z "$FAILED_RESOURCES" ]
+              then
+                echo "The following resources failed to be applied in the resource rollout job $FAILED_RESOURCES"
+                exit 1
+              else
+                echo "The resource rollout job successfully applied all the resources"
+              fi
+            else
+              echo "no custom resources to check"
+            fi
```
davidjumani, line 104: This might be unnecessary as the resource rollout job should fail if it did not apply the CRs successfully

install/helm/gloo/values-template.yaml
kdorosh: can you add this comment to the code too? next reader will probably not understand the high value

install/helm/gloo/templates/_8-default-virtualservice.tpl
```diff
@@ -0,0 +1,13 @@
+{{- define "gloo.customResources.defaultVirtualService" -}}
+{{- if .Values.virtualService.enabled }}
+---
+apiVersion: gateway.solo.io/v1
+kind: VirtualService
+metadata:
+  name: {{ .Values.virtualService.name | kebabcase }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  virtualHost: {}
```
kdorosh, line 10: IIUC, the issue is that envoy doesn't consider itself "ready" unless it can handle traffic / there is a listener + routes

by default, translation does not run if a gateway selects no virtual services, so this is a workaround to ensure we translate and provide envoy valid xds config?

if so, could we instead work around this by updating translation itself rather than writing a nonsense VS? im not really sure looking at this if we should even accept it (VWH should probably reject) / what does it mean?

kdorosh: I'm wondering if instead this should be a Gloo `Settings` field akin to: "add default route" that controls translation.

I think the default no listener unless you have routes is reasonable and good for backwards compat, but for users who want to use `--wait` or be able to remove all listeners from envoy then having that behavior toggle-able seems reasonable

davidjumani: The main issue here is that although there is a readiness probe configured on the envoy pod, the config to generate the health check endpoint on envoy is in the default gateway(which is created in a post hook job), and is not created unless envoy is ready(in case of using `--wait`). This creates a chicken and egg problem - which is why I've suggested moving that out of a post-hook.

Another issue is that a gateway is not translated into a listener / filter chain unless there is a VS defined that maps to it (which adding this nonsense VS tries to fix). I think that the approach of adding a new setting you suggested can fix that

install/helm/gloo/templates/5-resource-rollout-job-check.yaml
davidjumani: Discussed with @kdorosh The RBAC resources can be left as it doesn't cause any issues

install/helm/gloo/templates/_8-default-virtualservice.tpl
davidjumani: Have added a default gateway with a default as well as customizable domains and routes. Thanks @kdorosh for the suggestion

install/helm/gloo/generate/values.go
```diff
@@ -20,6 +20,7 @@ type Config struct {
 	IngressProxy   *IngressProxy           `json:"ingressProxy,omitempty"`
 	K8s            *K8s                    `json:"k8s,omitempty"`
 	AccessLogger   *AccessLogger           `json:"accessLogger,omitempty"`
+	VirtualService *VirtualService         `json:"virtualService,omitempty"`
```
inFocus7, line 23: nit/Q: Is the root config the right/best place for this? 

Thinking of UX + as if I'm someone new to Gloo: I'd be a bit confused on this field, especially since `VirtualService.enable: true` is what enables the default virtualService we want for the readiness probe. I'd think this would need to be enabled to use VirtualServices.

I personally would prefer this to be somewhere else, like inside `Settings` at least, as far as naming to make it clearer: not sure... 🤔

kdorosh: agree that this UX is weird. maybe something like:

```yaml
defaultVirtualService:
  enabled: false
  path: /
  directResponse: 'hello world!'
```

with those fields all overrideable

davidjumani: @inFocus7 Thanks for the feedback. I agree that it can be misleading, and changing it to defaultVirtualService, as @kdorosh had mentioned would be better UX.
However, since a default VS is needed just for the sake of the custom readiness probe (at this point), would it be better to add a dummy one with a default route only if the custom readiness probe is defined and not allow the user to specify a default one themselves (as later on it might cause conflicts with domains if there are already existing VS and the user decides to enable the default VS)

kdorosh: If the default is off (I think it should be) then I don't see a huge issue requiring people to opt in when they use a custom readiness probe

We might even be able to reuse the custom readiness probe path and such to signal instead of a new helm value

install/helm/gloo/templates/_8-default-virtualservice.tpl
kdorosh: If they already have a VS then they don't need our custom one? That's a good argument for making it a separate helm value in the chart like I originally proposed

davidjumani: Yes, if a VS already exists that matches the gateway on which the health check endpoint is defined, then there's no need for a default one.
So then I'll update this with a gloo setting to add a default route on a gateway if there is no VS that matches it

install/test/helm_test.go
```diff
@@ -6896,6 +6926,26 @@ func getStructuredDeployment(t TestManifest, resourceName string) *appsv1.Deploy
 	return structuredDeployment
 }
 
+func makeUnstructuredVirtualService(namespace, name, domain, path, response string) *unstructured.Unstructured {
+	return makeUnstructured(`
+apiVersion: gateway.solo.io/v1
```
sam-heilbron, line 6931: I understand what this VS solves (it ensures if you have a default gw enabled, and a default vs enabled, we will translate the configuration, and the xds snapshot that is generated will contain a route that we can use as the readinessProbe for our gw-proxy). However, what confuses me a bit is the use case:
If I'm a customer who wants Gloo to be installable, with my own configuration already generated, in "one step", this readinessProbe will indicate that Gloo is "ready" once we've processed the default configuration (ie the rollout jobs run), but how will a customer know that Gloo is "ready" with the custom resources they define?

install/helm/gloo/values-template.yaml
```diff
@@ -64,6 +64,14 @@ discovery:
...
     enablePodSecurityContext: true
     fsGroup: 10101
+virtualService:
+  enabled: false
+  name: default
+  domains:
+  - "*"
+  routes:
+  - path: "/"
+    response: "Welcome to Gloo Edge!"
```
sam-heilbron, line 74: Do we want to default all the values? My only fear is that if a user just enables this, and inherits the default, they will inherit a wildcard domain and root path which may cause issues with their other VirtualServices. What do you see as a downside to just not defaulting any of these values and requiring the user, should they opt into this, to configure the domain and routes themselves?

install/test/helm_test.go
```diff
@@ -110,10 +110,20 @@ var _ = Describe("Helm Test", func() {
 			// matchers passed in, so we can ensure that every resource has an associated matcher
 			ExpectWithOffset(1, customResourceManifest.NumResources()).To(Equal(len(matchers)))
 			for gwName, matcher := range matchers {
-				customResourceManifest.ExpectUnstructured("Gateway", namespace, gwName).To(matcher)
+				customResourceManifest.ExpectUnstructured(kind, namespace, gwName).To(matcher)
 			}
 		}
 
+		// There must be a matcher for every Gateway that's expected to be in the manifest.
+		assertCustomResourceManifestMatchesGateway := func(matchers map[string]types.GomegaMatcher) {
```
sam-heilbron, line 118: Since we use these utilties, I think we need to adjust the Offset value used in `assertCustomResourceManifest`, since previously it was an offset of 1, but now it will always be an offset of 2

test/kube2e/helm/helm_test.go
```diff
@@ -183,6 +183,68 @@ var _ = Describe("Kube2e: helm", func() {
...
+					"--wait-for-jobs",
+				}
+			})
+
+			It("succeeds adding a custom readiness probe and the gateway-proxy deployment is ready", func() {
+				expectGatewayProxyIsReady()
+			})
+		})
+
+		Context("On upgrade", func() {
```
sam-heilbron, line 234: We have a suite of "upgrade" tests, which are run nightly instead of on PRs, aimed at covering cases like this. Do you think it makes sense to incorporate this into that suite or keep it colocated with the other cases here?

test/kube2e/helm/helm_test.go
```diff
@@ -595,6 +662,50 @@ func uninstallGloo(testHelper *helper.SoloTestHelper, ctx context.Context, cance
 	cancel()
 }
 
+func getHelmUpgradeValuesOverrideFileForCustomReadinessProbe() (filename string, cleanup func()) {
+	values, err := os.CreateTemp("", "values-*.yaml")
```
sam-heilbron, line 666: nit: I've found that creating tmp files, and then deleting them can be more confusing than just checking this yaml into it's own file in the `artifacts` directory and referencing it

changelog/v1.16.0-beta12/fix-readiness-probe.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/8288
+    description: >- 
+      Fixes the issue of defining a custom readiness probe for the gateway proxy for zero downtime upgrades via helm with the `--wait` flag.
```
sam-heilbron, line 5: Should we clarify that this works only when the defaultHttpGateway is enabled? For customers using just the defaultHttpsGateway, or disabling default gateways, this will not work, right?

install/helm/gloo/templates/_8-default-virtualservice.tpl
kdorosh: isn't there a catch-22 here though? how can a virtual service exist before we install the CRD during initial install?

davidjumani: That implies that it must be a multi step install to configure the health check. I'm okay (but a bit hesitant) to add a default VS but it might lead to more complicated installs

nfuden: The CRD should exist at the point we spin up the gatewayproxy. It seems like the requirement is that a VS MUST be added at a stage prior or equal to when the gateway-proxy exists. This means CRDs must be applied and the custom resources must be applied.
if this is in the same stage of helm with appropriate retries then it hypothetically can work for other deployment patterns such as argo right? Or am I missing something here?

davidjumani: So looking at the docs for the health check and zero downtime, this is my understanding :
1. Gloo is installed
2. The user creates a VS
3. The user changes the helm values to add the health check plugin and the readiness probe on the gateway proxy as an upgrade

Now if a user tries to install / upgrade gloo with the health check enabled and zero downtime, it fails since :
1. The resource rollout runs only once the gateway-proxy is ready which never happens since it is a post-install hook
2. There is no VS - specific to clean installs

This PR fixes both the issues by :
1. Running the resource rollout job as part of the chart rather than a post hook
2. Adding a default VS - specific fix for clean installs

So now my question is :
Since the docs explicitly mention that a VS must exist prior to applying the health check and zero downtime configs, does it make sense to add the fix for the missing VS caused during clean installs (either by adding default VS or the setting to allow gloo to translate an gateway to a listener without any VS defined which will lead to more complicated installs) OR just stick with the docs and force the user into a multi step install to get the zero downtime config working (Fix 1 will still go in to resolve the issue when using Argo or helm with --wait)

kdorosh: I'm partial to doing this in translation via a new gloo setting since we don't have to worry about domain conflicts on the default virtual service (by not having one); for this special placeholder we can remove it during translation if there are other defined virtual services

sam-heilbron: Yeah I'm good with that plan, to have the translation engine support this, and not need more default resources to exist.

I think the broader question of: "does our zero downtime upgrade config need to work on out of the box installs" is one we can engage with the product team on. I like that default resources provide simple ways for new users of Edge to demonstrate the functionality. As it stands today, some of our production recommended settings (like the zero downtime config) haven't worked with these default resources, and I'm not sure we have to make them work. I'd rather slim down our install to be minimal, and let customers manage their own configuration. I like having optional defaults which are not intended to be feature complete, but provide an easy way to be introduced to the product. Either way, this is a larger conversation that I don't think blocks this current work.

tldr. yeah let's move forward with solving this in translation with an optional flag

install/test/helm_test.go
davidjumani: Thanks for calling out the niche case. Have updated the logic to use a gloo setting instead

install/helm/gloo/values-template.yaml
davidjumani: Removed the default VS idea itself!

install/test/helm_test.go
davidjumani: Reverted the changes

changelog/v1.16.0-beta12/fix-readiness-probe.yaml
davidjumani: From the docs - it should work on HTTP && HTTPS gateway based on how it is configured and enabling them is a prerequisite

test/kube2e/helm/helm_test.go
davidjumani: I like the idea. I've moved generation of all temp files out of this to use static files only

changelog/v1.16.0-beta12/fix-readiness-probe.yaml
sam-heilbron: Correct, though my fear is that the docs are inaccurate. As we've noted, a listener will only be created if a GW and VS both exist. HttpGateways select VirtualServices without any SslConfig defined on them (https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/translator/virtual_service_translator.go#L177) and an HttpsGateway selects the inverse (only VirtualServices with sslConfig). In your use case of using a default VS, without any VS, I don't think an HttpsGateway will work (yet)

projects/gloo/api/v1/settings.proto
```diff
@@ -826,6 +826,9 @@ message GatewayOptions {
     // Individual Gateways can override this behavior by configuring the "gateway.solo.io/isolate_vhost" annotation
     // to be a truthy ("true", "false") value
     google.protobuf.BoolValue isolate_virtual_hosts_by_ssl_config = 10;
+
+    // If set, the gateways wil be translated into Envoy listeners even if no VirtualServices exist
```
sam-heilbron, line 830: I think we can flush this out a bit more to clarify the impact. When we say "into Listener even if not VirtualServices exist", what does that mean for the routing config on the listener? What is the default value and what is the behavior if there are no routes for a given gateway and this is false?

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -85,9 +86,11 @@ func (n *httpNetworkFilterTranslator) computePreHCMFilters(params plugins.Params
 }
 
 func (n *httpNetworkFilterTranslator) ComputeNetworkFilters(params plugins.Params) ([]*envoy_config_listener_v3.Filter, error) {
-	// return if listener has no virtual hosts
-	if len(n.listener.GetVirtualHosts()) == 0 {
-		return nil, nil
+	if !settingsutil.MaybeFromContext(params.Ctx).GetGateway().GetTranslateEmptyGateways().GetValue() {
```
sam-heilbron, line 89: I'm a bit confused about when this flag plays a role in translation, what it's responsibility is, and how it changes behavior. The are Gateway resources and VirtualServices, and we're saying that if you have this flag enabled, you can have a Gateway, with no VirtualServices and we'll still product a Listener (a Gloo listener) that lives on the Proxy object. 
The translation here, takes a Gloo Proxy (and all the listeners on it) and convert it into Envoy resources. In the case where this setting is enabled, what "routes" will live on the Gloo listener?

projects/gateway/pkg/translator/aggregate_translator.go
```diff
@@ -37,7 +37,11 @@ func (a *AggregateTranslator) ComputeListener(params Params, proxyName string, g
 		if len(snap.VirtualServices) == 0 {
 			snapHash := hashutils.MustHash(snap)
 			contextutils.LoggerFrom(params.ctx).Debugf("%v had no virtual services", snapHash)
-			return nil
+			if settingsutil.MaybeFromContext(params.ctx).GetGateway().GetTranslateEmptyGateways().GetValue() {
+				contextutils.LoggerFrom(params.ctx).Debugf("but continuing since translateEmptyGateways is set", snapHash)
+			} else {
+				return nil
+			}
```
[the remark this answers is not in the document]
davidjumani: I ran a test which created a default VS along with an update to the gateway-proxy with a selector that did not match the default VS. This was all done as part of the resource rollout, to simulate an existing VS but a GW that did not select it. The GW came up along with the health check config, even though it did not match any VS. The changes in https://github.com/solo-io/gloo/pull/8698/files#diff-28c6070543dbbd1dd464a1b5903147133dc3815cc74aeb6f49e749d333b9afa0R89-R93 handle that scenario

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -94,12 +95,16 @@ func (t *HybridTranslator) computeHybridListenerFromMatchedGateways(
...
 		case *v1.MatchedGateway_HttpGateway:
 			if len(snap.VirtualServices) == 0 {
+				snapHash := hashutils.MustHash(snap)
 				if !loggedError {
-					snapHash := hashutils.MustHash(snap)
 					contextutils.LoggerFrom(params.ctx).Debugf("%v had no virtual services", snapHash)
 					loggedError = true // only log no virtual service error once
 				}
-				continue
+				if settingsutil.MaybeFromContext(params.ctx).GetGateway().GetTranslateEmptyGateways().GetValue() {
```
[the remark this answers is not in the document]
davidjumani: It still does optimize error logging, however in case there were not VS, the loop would `continue` on to the next iteration. The changes made here ensure that it proceeds with the loop even if there are no VS present in the snap

test/kube2e/helm/helm_test.go
davidjumani: I think that right here is fine since https://github.com/solo-io/gloo/pull/8738 should take care of most upgrades

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -94,12 +95,16 @@ func (t *HybridTranslator) computeHybridListenerFromMatchedGateways(
 		switch gt := matchedGateway.GetGatewayType().(type) {
 		case *v1.MatchedGateway_HttpGateway:
 			if len(snap.VirtualServices) == 0 {
+				snapHash := hashutils.MustHash(snap)
```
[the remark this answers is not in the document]
davidjumani: Thanks! I've updated it

projects/gloo/api/v1/settings.proto
davidjumani: Updated the docs!

projects/gloo/pkg/translator/network_filters.go
davidjumani: I've added an explanation of this flag in the commit to update the proto docs.
The listener will be created without routes, so it will return a 404 for every request. This is to ensure that plugins which do not require a VS or routes (like the health check plugin) can still work

changelog/v1.16.0-beta12/fix-readiness-probe.yaml
davidjumani: You're right. It does create the health check endpoint on the HTTPS listener but since there is no virtual service, which in turn means no SSL config, the health check endpoint returns an HTTP response.
I think this will be a good callout in the docs that unless a VS with sslconfigs is provided, the health check plugin will return only HTTP responses

install/helm/gloo/generate/values.go
davidjumani: Decided to go with the new gloo setting as discussed in https://github.com/solo-io/gloo/pull/8698#discussion_r1341564189

install/helm/gloo/templates/_8-default-virtualservice.tpl
davidjumani: Updated to solve this in translation using the new setting

install/helm/gloo/generate/values.go
```diff
@@ -370,6 +370,7 @@ type Gateway struct {
 	IsolateVirtualHostsBySslConfig *bool             `json:"isolateVirtualHostsBySslConfig,omitempty" desc:"if true, Added support for the envoy.filters.listener.tls_inspector listener_filter when using the gateway.isolateVirtualHostsBySslConfig=true global setting."`
 	CompressedProxySpec            *bool             `json:"compressedProxySpec,omitempty" desc:"if true, enables compression for the Proxy CRD spec"`
 	PersistProxySpec               *bool             `json:"persistProxySpec,omitempty" desc:"Enable writing Proxy CRD to etcd. Disabled by default for performance."`
+	TranslateEmptyGateways         *bool             `json:"translateEmptyGateways,omitempty" desc:"If true, the gateways wil be translated into Envoy listeners even if no VirtualServices exist"`
```
sam-heilbron, line 373: This is introduced to support "default" Gateway's (ones that we ship with an install via the Helm chart), and therefore this is only really relevant to HttpGateway's. In cases where users have already defined resources (VirtualServices, MatchableHttpGateway) do we want to encourage users to not rely on this functionality and instead define their own routes? I wonder if we just want to scope this functionality to our default resources only to limit the scope

sam-heilbron (commented):
This looks great! I ran through the zero-downtime rollout guide (https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/) and it uses an HttpsGateway, which we agreed will not work. I confirmed that following the guide as is, produces the following event on the pod:
```
 Warning  Unhealthy  3s (x12 over 53s)  kubelet            Readiness probe failed: Get "https │
│ ://10.244.0.7:8443/envoy-hc": http: server gave HTTP response to HTTPS client    
```

I think we'll need to update the guide to reflect what works.

I was able to confirm that using an HttpGateway this works.

I asked one question about the scope of this work and whether we should apply this to all Gateway's or just the default resources that we generate.

One other thought I had was whether we want to define this globally, or on a per-gateway level. Is that something you considered?

install/helm/gloo/generate/values.go
davidjumani: If users have defined their own gateways and resources, then there should be no need to enable this to get zero downtime rollouts. This should be enabled only in the case of clean installs, or if the user has a gateway with plugins but no routes (eg. the health check plugin - to ensure the gateway is healthy before adding anything to it.
I'll add some notes regarding using this and pass it on to the docs team to ensure that the use of this field is only for the reasons stated

davidjumani:
> I asked one question about the scope of this work and whether we should apply this to all Gateway's or just the default resources that we generate.
> 
> One other thought I had was whether we want to define this globally, or on a per-gateway level. Is that something you considered?

Thanks for the review and tests @sam-heilbron 
I did consider this on a per gateway basis - however these were the reasons I chose to make this a global setting instead :
-  This affects translation and is not an add-on to a specific gateway such as a plugin or feature (which almost all per-gateway-level options are)
- Users would expect translation behaviour to be similar across gateways and features to be per-gateway

sam-heilbron (approved):

nfuden (approved):

test/kube2e/helm/helm_test.go
davidjumani: I'd prefer to have it here as this is a potential blocker for any CD deployments. This way if it breaks we can address it in the PR instead of allowing it in and fixing it when detected

davidjumani: I'll create an issue to keep the helm upgrade tests in sync with the nightly ones and pick it up if it is priority

PR #8713, opened 2023-09-22
OIDC internal api refactor

sheidkamp:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1403,12 +1403,11 @@ message ExtAuthConfig {
     }
 
     message OidcAuthorizationCodeConfig {
-
         // your client id as registered with the issuer
         string client_id = 1;
 
-        // your client secret as registered with the issuer
-        string client_secret = 2 [(extproto.sensitive) = true];
+        // your client secret as registered with the issuer. Fully deprecated and will be ignored.
```
sam-heilbron, line 1409: I'm a little concerned by this. To me deprecated means that we will support it for some window, but not recommend it. This provides a buffer for customers to use the old API before forcing them to migrate to the new one. Can we do this more softly and just deprecate for a a few releases, and perhaps note that in 1.18 we will remove this fuctionality?

Also in the comment for this field (any deprecated field), I think we should include a reference to the preferred field

sam-heilbron: On second thought, I realize this is our internal API, so we control what is being set and can stop configuring it. I still think it would be helpful to outline in this API what the preferred fields are so future devs can follow the change

sam-heilbron: The user facing API for this field has:
```
// your client secret as registered with the issuer.
// This is required unless `disable_client_secret` is true
core.solo.io.ResourceRef client_secret_ref = 2;
```

Given that, would it help if we added a comment in the internal API to lay out the situations where this field will be set, and which ones where it will not?

sheidkamp: Added a comment to the internal API by the new nested client_secret field

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1510,6 +1510,18 @@ message ExtAuthConfig {
...
+
+        message ClientSecretExchangeConfig {
+            // your client secret as registered with the issuer. This field may be nil/empty. Any validation of the client secret happens when translating the public interface
+            string client_secret = 1 [(extproto.sensitive) = true];
+        }
+
+        // Configure the specifics of how the client secret is exchanged for an access token. Currently only supporting client secret.
+        oneof exchange_config {
+            // Use the client secret to exchange for an access token.
+            ClientSecretExchangeConfig client_secret_exchange_config = 21;
```
kdorosh, line 1522: why wrap in the new `ClientSecretExchangeConfig` type? we can just put the string in the oneof

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1510,6 +1510,18 @@ message ExtAuthConfig {
 
         // Configuration related to the user session.
         UserSessionConfig user_session = 20;
+
+        message ClientSecretExchangeConfig {
+            // your client secret as registered with the issuer. This field may be nil/empty. Any validation of the client secret happens when translating the public interface
```
sam-heilbron, line 1515: nit: I think we could be more precise with "may be nil/empty" and outline the cases that we expect. More importantly from an API perspective, what does the behavior for a empty client secret look like? I think we can handle this clarification in the follow-up PRs that will come in this space

sam-heilbron (approved):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1403,12 +1403,12 @@ message ExtAuthConfig {
...
     message OidcAuthorizationCodeConfig {
-
         // your client id as registered with the issuer
         string client_id = 1;
 
-        // your client secret as registered with the issuer
-        string client_secret = 2 [(extproto.sensitive) = true];
+        // your client secret as registered with the issuer. Fully deprecated and will be ignored.
+        // Please use the client_secret_exchange_config option of the exchange_config field to configure the client secret.
+        string client_secret = 2 [(extproto.sensitive) = true, deprecated = true];
```
kdorosh, line 1411: we can deprecate, but gloo control plane and extauth dataplane need to look at both for a release or two or else there's a risk of downtime on upgrade due to mismatched control and data planes. an unfortunate side effect of managing data plane and control plane in a single helm release.. we don't know which may happen first (or if the other one can even be scheduled or complete)

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: The public facing API has comments that address this

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: This is the internal API. We will keep the public facing API backwards compatible for a while... does that address these concerns?

kdorosh: No; the internal API can still result in downtime if there isn't a version that supports both. Let's consider the case on upgrade where the internal API changes and we're updating gloo.

Let's say that the control plane updates first and the dataplane updates second (timing issue, fails to schedule etc).

Now the internal API served by gloo control plane is unintelligbile to the older extauth and we have downtime for the extauth service.

We can also flip the order here and have similar issues (dataplane updates before control plane)

sheidkamp: Thanks for the feedback, after discussing it with Sam, we're going to just leave the client secret where it is on the internal API to make these kind of compatibility issues moot. I'm going to close this PR because we can do the first-pass extauth refactor without API changes.

Issue #8724, opened 2023-09-27
Add docs to install Edge via ArgoCD 


davidjumani:
### Version

main (1.16.x beta)

### Describe the requested changes

There's already a blog on [From Zero to Gloo Edge in 15 Minutes: ArgoCD GitOps Edition](https://www.solo.io/blog/from-zero-to-gloo-edge-in-15-minutes-gitops-edition/)
As of now, this is just a blog, but there are customers who install Edge via Argo. Having this in the docs will be a good reference, and grow to include best practices specific to this topic.

### Link to any relevant existing docs

https://www.solo.io/blog/from-zero-to-gloo-edge-in-15-minutes-gitops-edition/

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1207635280052738/1207994530449008) by [Unito](https://www.unito.io)

davidjumani:
Worth mentioning https://github.com/solo-io/gloo/pull/8738/files#r1341595379 as well

Nadine2016:
there is another blog about this https://github.com/ably77/gloo-edge-argo-rollouts-blog

davidjumani:
Thanks @Nadine2016, that blog is on argo rollouts to trigger changes. The docs that need to ba added here should detail how to install edge via argo

Nadine2016:
this was presented in the enablement session https://drive.google.com/drive/folders/1hq0_AQq7DnJNh1PgMVDo1jwYQijmcf2V .

Nadine2016:
We should also show an example for how to use the Argo Rollouts plugin

Nadine2016:
done for Gloo Gateway https://docs.solo.io/gateway/latest/installation/argocd/

Nadine2016:
edge PR https://github.com/solo-io/gloo/pull/10173

Nadine2016:
in prod https://docs.solo.io/gloo-edge/main/installation/gateway/argo/ closing

PR #8738, opened 2023-09-29
Add --wait to helm tests and other improvements to support CD tools

davidjumani:
# Description

Most CI tools wait until resources are ready to mark a release as successful. Eg. [ArgoCD](https://argo-cd.readthedocs.io/en/stable/operator-manual/health/) and [Flux](https://fluxcd.io/flux/components/helm/helmreleases/#disabling-resource-waiting)
We can emulate this by passing the `--wait` flag to helm rather than individually adding support for every CD tool out there.
This PR adds this flag to the helm tests as well as some tweaks to support ArgoCD

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml
+++ b/install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml
@@ -4,12 +4,6 @@
 {{- $image = merge .Values.gateway.certGenJob.image .Values.global.image }}
 {{- end }}
 {{- if .Values.global.glooMtls.enabled }}
-{{- /* With ArgoCD, there is no concept or installs or upgrades. Everything is a sync. */}}
-{{- /* This causes an issue when a hook must run on an install but not an upgrade based on a flag. */}}
-{{- /* We address this by creating the job definition for an upgrade only if the flag says so */}}
-{{- /* This is done in the next two lines */}}
-{{- $shouldJobRunOnUpgrade := and .Release.IsUpgrade .Values.gateway.certGenJob.runOnUpdate }}
-{{- if or .Release.IsInstall $shouldJobRunOnUpgrade }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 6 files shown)


sam-heilbron (commented):
This is awesome! 

We also have an upgrade suite, which doesn't run on PRs, but will run nightly and is intended to identify regressions with upgrading from an old version of Gloo to a newer one. Do you think it would be valuable to add this functionality to that suite as well? (example: https://github.com/solo-io/gloo/blob/main/test/kube2e/upgrade/upgrade_test.go#L302)

install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml
```diff
@@ -4,6 +4,12 @@
 {{- $image = merge .Values.gateway.certGenJob.image .Values.global.image }}
 {{- end }}
 {{- if .Values.global.glooMtls.enabled }}
+{{- /* With ArgoCD, there is no concept or installs or upgrades. Everything is a sync. */}}
+{{- /* This causes an issue when a hook must run on an install but not an upgrade based on a flag. */}}
+{{- /* We address this by creating the job definition for an upgrade only if the flag says so */}}
+{{- /* This is done in the next two lines */}}
+{{- $shouldJobRunOnUpgrade := and .Release.IsUpgrade .Values.gateway.certGenJob.runOnUpdate }}
+{{- if or .Release.IsInstall $shouldJobRunOnUpgrade }}
```
davidjumani, line 12: This would not work as argo doesn't template with --is-install or --is-upgrade. This would be a good callout in the docs that if a user is using ArgoCD and global.glooMtls.enabled=true, it will run on every sync irrespective of the value of `gateway.certGenJob.runOnUpdate`

install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml
```diff
@@ -13,6 +13,8 @@ metadata:
   name: gloo-mtls-certgen
   namespace: {{ .Release.Namespace }}
   annotations:
+    {{- /* With ArgoCD, there is no concept or installs or upgrades. Everything is a sync. */}}
+    {{- /* Due to this behavour of ArgoCD, the value of `gateway.certGenJob.runOnUpdate` can not be respected and will run on every sync. */}}
```
davidjumani, line 17: This would need to be called out in the docs

sam-heilbron: Is that something you plan on doing as part of this work?

sam-heilbron (dismissed):
Love the comments!

install/helm/gloo/templates/19-gloo-mtls-certgen-job.yaml
davidjumani: I've added a reference to this in https://github.com/solo-io/gloo/issues/8724
There might be other instances which might need to be called out because of the way Argo works
Once I've addressed the issues mentioned in https://github.com/solo-io/solo-projects/issues/5293 we'll have confirmed what works and what doesn't in Argo and can start with the docs

jenshu (approved):

sam-heilbron (approved):

PR #8740, opened 2023-10-02
fix: Validate on `Secret` deletions

inFocus7:
# Description

Fixes the issue where the Validation Webhook did not validate on `Secret` deletions. 
This was due to our `ValidatingWebhookConfiguration` configured to validate on secrets part of the `gloo.solo.io` groups, which does not exist (we don't have a secret CRD). Now we correctly validate against Kubernetes `Secret` deletions.

_proof of no secret crd, all crds from 1.16.0-beta1_
Screen Shot 2023-09-28 at 12 28 56 PM

## A bit of history

Secret validation was added in [this PR](https://github.com/solo-io/gloo/pull/5470/files). 
The tests added in those are valid as they **do** test that our secret validation logic works, **BUT** don't include an end-to-end test _beginning_ from an actual deletion of a secret.

If the (simplified) flow of validation is:
`[delete secret request] > [validation webhook activates] > [request gets parsed] > [internal validation] > [answer]`
The tests covered from 'internal validation' onward, so they didn't catch the case/fact that a secret is not in our CRDs. I manually tested that secrets are not being validated on `k delete secret` on the version that code was included.

The end-to-end tests added here cover from the initial flow of validation by actually deleting a secret, which are of GVK `<kube api, aka. "">/v1/Secrets`.

## API changes

- Updated our `ValidatingWebhookConfiguration` to correctly validate on Kubernetes' core secrets.
  - Those are referenced by the API Group `""`.

## Code changes

- Allow the API Group during validation to be empty `""`, allowing Kubernetes APIs to be validated, by not enforcing a match to exist in our `HashableResource`.
  - `Secrets` are part of Kubernetes core API `""`.
- Mapped the Kubernetes `Secret` to our `HashableResource` on deletion.
  - `HashableResources` (and `resource.Resource`) are maintained for **our** resources, but a lot of our validating logic requires one, so we needed a way to still validate against the Kube secret resource. Since we only care about the Secret **deletions**, we can safely do this,as deletions only care about the name + namespace, which **all** resources (including Kube's) have.
- Added tests to verify a validation webhook occurs (and doesn't) depending on conditions.
- Added secrets to our translation performance tests. Used the basicConfig, as its constraints worked for `Secrets` as well.

## Performance Implications

Performance will likely face a hit for users **with** `webhook validation` enabled **when** deleting a `Secret`, as we're now correctly validating their deletion. The performance hit during Secret deletion will be similar to any other validation we have on resources, but may be more noticable in environments rotating and deleting secrets more consistently.

To opt-out, you can add `'secrets'` to our following api `gateway.validation.webhook.skipDeleteValidationResources`.

# Context

When deleting secrets when the validating webhook was enabled, users noticed that their deletion were not being validated. This meant that they would always successfully delete a secret, even if it causes a proxy error when another resource (ex. VirtualService) referenced it.

## Interesting decisions
 
- Mapped the Kubernetes core secret API `""/v1/Secret` (where `""` represents the Kube core APIs) to our `Secret` `HashableResource`.
  - This had to be done since our validation functions require an internal `resource.Resource`.
  - I did not create a custom `HashableResource` type for the Kubernetes secret, as (from the looks of it) they are meant for **our** APIs, whereas Kubernetes is not.
  - This works in _this_ scenario as the deletion of a resource _only_ requires the `name` and `namespace` for the resource, and not any other internal APIs.

### 1. Install Gloo
**validation-webhook.yaml**
```yaml
gateway:
  enabled: true
  validation:
    enabled: true
    alwaysAcceptResources: false
    webhook:
      enabled: true
```

_To see the issue, install a released (or non-updated) version_
```bash
helm install gloo gloo/gloo --namespace default --create-namespace --version 1.16.0-beta1 --values validation-webhook.yaml
```

_To notice the fix install this branch locally_
```bash
./ci/kind/setup-kind.sh
```
```bash
helm install gloo _test/gloo-1.0.0-ci.tgz --values validation-webhook.yaml
```

#### Create a certificate + key
We'll need a certificate and key for TLS secrets. Since we don't need to test the functionality of SSL, and just need to reference the cert + key, can simply create & use a key created from the following - with whatever certificate data you want (or default):
```bash
openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out tls.crt -keyout tls.key
```

#### Create Secret(s) using created cert + key
_We are creating 3 secrets, in order to manually verify the different possibilities_
```bash
k create secret tls upstream-tls --key tls.key --cert tls.crt
```
```bash
k create secret tls upstream-tls2 --key tls.key --cert tls.crt
```
```bash
k create secret tls upstream-tls3 --key tls.key --cert tls.crt
```

### 3. Create VirtualService
```bash
cat << EOF | kubectl apply -f -
kind: VirtualService
metadata:
  name: direct-response
  namespace: default
spec:
  sslConfig:
    secretRef:
      name: upstream-tls
      namespace: default
  virtualHost:
    domains:
    - '*'
    routes:
    - matchers:
      - prefix: /
      directResponseAction:
        status: 200
        body: "Hello, world!"
EOF
```

### 4. Delete the Secrets (manual verification)
**Without branch fix**
You can delete all resources without issue
```bash
k delete secret upstream-tls
```
```bash
k delete secret upstream-tls2
```
```bash
k delete secret upstream-tls3
```

**With branch fix**
1. **`upstream-tls2`** should delete successfully, as it is not referenced elsewhere.
```bash
k delete secret upstream-tls2
```
2. **`upstream-tls`** should fail to delete due to it being used in the **`direct-response`** VS.
```bash
k delete secret upstream-tls
```
3. Edit the **`direct-response`** VS to point to **`upstream-tls3`**.
```bash
k edit vs direct-response
# Point the secret ref to upstream-tls3
```
4. **`upstream-tls`** should now delete, as it's not referenced anywhere
```bash
k delete secret upstream-tls
```
5. **`upstream-tls3`** should fail to delete, due to it being used in the `direct-response` VS.
```bash
k delete secret upstream-tls3
```
6. Delete the **`direct-response`** VS, so nothing references the **`upsteram-tls3`**.
```
k delete vs direct-response
```
7. **`upstream-tls3`** should now delete, as it's not reference anywhere.
```
k delete secret upstream-tls3
```

## Notes for reviewers

If we fear the implications on performance on this fix, we _could_ add a helm field for validation to enable/disable secret validation and gate `webhookvalidationconfiguration`'s secret ops. (unless something like this exists already)

# Checklist:

- ~I have made corresponding changes to the documentation~


projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -396,9 +398,12 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
 		return wh.validateList(ctx, admissionRequest.Object.Raw, dryRun)
 	}
 
-	if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk]; !hit {
-		contextutils.LoggerFrom(ctx).Debugf("unsupported validation for resource namespace [%s] name [%s] group [%s] kind [%s]", ref.GetNamespace(), ref.GetName(), gvk.Group, gvk.Kind)
-		return &validation.Reports{}, nil
+	// Only check resources that are part of our internal API
+	if gvk.Group != kubernetesCoreApiGroup {
```
bewebi, line 402: If the logic is meant to be "only Gloo API resources", it should be
```
if gvk.Group == glooApiGroup {
```

There may be another group we care about at some point, so the logic should be as precise as possible

test/kube2e/gateway/gateway_test.go
```diff
@@ -2522,6 +2529,71 @@ spec:
...
+							SecretRef: &core.ResourceRef{
+								Name:      tlsSecret.GetName(),
+								Namespace: tlsSecret.GetNamespace(),
+							},
+						},
+					}
+				})
+
+				AfterEach(func() {
+					// The secrets kube client does not have an IgnoreNotFound field, so not removing the created secret here, as the removal of the secret part of the test
```
bewebi, line 2555:
```suggestion
					// The secrets kube client does not have an IgnoreNotFound field, and deleting the secret is part of the test
					// so there is no need to remove the created secret here
```

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
inFocus7: I initially stuck with this because there are multiple groups we allow 
https://github.com/solo-io/gloo/blob/20ef52e64e654f2c85412ca8cfbcca8ad881895a/projects/gloo/pkg/api/v1/gloosnapshot/api_snapshot.sk.go#L820-L837

so the logic would be 
```
if gvk.Group == glooApiGroup || gvk.Group == glooEnterpriseApiGroup || gvk.Group == graphqlApiGroup || gvk.Group == rateLimitApiGroup {}
```
or
```
groups := [glooApiGroup, glooEnterpriseApiGroup, graphqlApiGroup, rateLimitApiGroup]
if groups.includes(gvk.Group) {}
```
which seemed a bit cumbersome to me.

That, and the `_, hit := gloosnapshot.ApiGvkToHashableResource[gvk]` does the check in a map which is fast. I could also update the comment itself.

inFocus7:
pushed noop, since changelog bot acted up. seeing the same in another pr, so checking if re-pushing fixes it. it does 🥳

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
bewebi: Yeah, that makes sense
I might replace or add to the comment with
```
// kubernetesCoreApiGroup is the only group we get validation requests for that is not a part of a Solo API
```

bewebi (commented):
Looking good, appreciate the thorough PR description
It may be worth adding context/info about the translation performance tests that have been added/modified to the description as well
In the testing steps it would be great to include steps for generating the key and crt files as well

Will approve once the open conversations are resolved and I've verified using the provided steps

test/kube2e/gateway/gateway_test.go
```diff
@@ -2522,6 +2529,72 @@ spec:
...
+							SecretRef: &core.ResourceRef{
+								Name:      tlsSecret.GetName(),
+								Namespace: tlsSecret.GetNamespace(),
+							},
+						},
+					}
+				})
+
+				AfterEach(func() {
+					// The secrets kube client does not have an IgnoreNotFound field, and deleting the secret is part of the test
```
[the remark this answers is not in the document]
inFocus7: Made some updates. Didn't add an eventually, but used an `Or` to either expect no error (deleted successfully) or the secret not to be found (if it gets deleted as part of a test) [8ce18ff](https://github.com/solo-io/gloo/pull/8740/commits/8ce18fff4c6c41ba372f2ce4b03eeff6498830c4)

bewebi (approved):
LGTM and I've manually verified intended behavior using the provided steps

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -396,9 +398,12 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
 		return wh.validateList(ctx, admissionRequest.Object.Raw, dryRun)
 	}
 
-	if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk]; !hit {
-		contextutils.LoggerFrom(ctx).Debugf("unsupported validation for resource namespace [%s] name [%s] group [%s] kind [%s]", ref.GetNamespace(), ref.GetName(), gvk.Group, gvk.Kind)
-		return &validation.Reports{}, nil
+	// Kubernetes' core API group is the only group we get validation requests for that is not a part of a Solo API.
+	if gvk.Group != kubernetesCoreApiGroup {
+		if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk]; !hit {
```
jenshu, line 403: a couple questions here:
1. why do we need this check? is it in case users call the validation endpoint directly (and try to pass through a resource that we don't validate) instead of configuring a ValidatingWebhookConfiguration (which will filter by gvk for us)?
2. if we are doing this check for gloo resources, should we also add a check for kubernetesCoreApiGroup and make sure the kind == secret?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -410,7 +415,11 @@ func (wh *gatewayValidationWebhook) validateAdmissionRequest(
 
 func (wh *gatewayValidationWebhook) deleteRef(ctx context.Context, gvk schema.GroupVersionKind, ref *core.ResourceRef, admissionRequest *v1beta1.AdmissionRequest) (*validation.Reports, *multierror.Error) {
 	newResourceFunc := gloosnapshot.ApiGvkToHashableResource[gvk]
-
+	// Special case for Kubernetes secrets, since they are not handled by our hashable resource.
+	// We can reuse the NewSecretHashableResource resource.Resource, since all that matters is the metadata for deletion.
+	if gvk.Group == kubernetesCoreApiGroup && gvk.Kind == "Secret" {
```
jenshu, line 420: do we need to do something similar in `validateGvk` below? Currently the ValidatingWebhookConfiguration only validates secret deletion, but if a customer modifies their ValidatingWebhookConfiguration to validate secrets on create/update, and/or calls the validation endpoint directly, i think it might error.

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
inFocus7: We initially just did a check like so
```
if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk]; !hit {}
```

which checks that the resource we're validating (against?) is a Gloo API/GVK. 
When deleting Secrets, their resource/GVK is core Kubernetes API, whose group is `''`. Without the check I added, we'd be trying to find the kubernetes secret in our `ApiGvkToHashableResource` table, which would fail and lead to not validating due to "unsupported validation for resource namespace".

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
inFocus7: Hmm... that's a good question. I haven't thought about customers configuring non-Delete validations for Secrets. Without this code, if they added more operation checks on Secrets, nothing would happen as the CRD didn't exist. But you do have a point that now, it could lead to an error.

inFocus7: I'll test this out locally, but adding support _may_ be very very tricky. Our HashableResources, which our validation uses, are set up for Gloo APIs, but the Secret would be Kubernetes. 

It was easy for deletion, as the only field that matter for deletion are the name + namespace. For creating/updating, this would be tricky since we can't unmarshal(?) a Kubernetes secret into a Gloo secret due to our added fields (kind, and maybe others.) We'd likely need to create/reuse some translation logic (kubeSecret->glooSecret) so it could work 🤔

inFocus7: The easy / cop out way would be to just _not_ allow non-deletion on secrets by doing something like but would likely require discussion with leads/field
```go
if isKubernetesGroup && isSecret && !isDelete {
  contextutils.LoggerFrom(ctx).Infof("unsupported validation for resource namespace [%s] name [%s] group [%s] kind [%s]", ref.GetNamespace(), ref.GetName(), gvk.Group, gvk.Kind)
  return &validation.Reports{}, nil
}
```

inFocus7: Update: So adding the below at least doesn't break anything. I haven't tested without it, but it makes sense to at least have this so that `newResourceFunc` is not nil and error/panic later when called with `newResourceFun()`.
```go
// Special case for Kubernetes secrets, since they are not handled by our hashable resource.
if gvk.Group == kubernetesCoreApiGroup && gvk.Kind == "Secret" {
	newResourceFunc = gloov1.NewSecretHashableResource
}
```

With the above, I am able to CRUD the Secret without any issues. My only concern is that it's _possibly_ not actually validating Secrets on creation/update since I don't believe we can unmarshal kube secret data onto our gloo secret api. 

The logs were as follow when creating a secret w/ create validation set:
```
{"level":"info","ts":"2023-10-10T20:44:41.834Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:198","msg":"received validation request on webhook","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:41.836Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.264Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.TranslatorSyncer","caller":"syncer/translator_syncer.go:92","msg":"begin sync 5392730014985031803 (0 virtual services, 2 gateways, 0 route tables)","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.271Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.TranslatorSyncer","caller":"syncer/translator_syncer.go:104","msg":"end sync 5392730014985031803","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.273Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.envoyTranslatorSyncer","caller":"syncer/envoy_translator_syncer.go:84","msg":"begin sync 5392730014985031803 (0 proxies, 23 upstreams, 26 endpoints, 4 secrets, 15 artifacts, 0 auth configs, 0 rate limit configs, 0 graphql apis)","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.273Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.envoyTranslatorSyncer","caller":"syncer/envoy_translator_syncer.go:179","msg":"end sync 5392730014985031803","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.276Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:198","msg":"received validation request on webhook","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.276Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.280Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:198","msg":"received validation request on webhook","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.281Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.284Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:198","msg":"received validation request on webhook","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T20:44:42.289Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
```

Technically the behavior will be the same as before, but I'll need to keep digging into making sure it validates. 

In addition, tbh I'm not sure _when_ a Secret's creation/update would lead to an error _not_ caught by Kubernetes 🤔 

From my understanding, we're checking that _Gloo's_ snapshot/setup is valid through our validation server, so it makes sense to watch Secret deletion as our resources may call it. For Secret creation/update, I'm not sure what we'd want validated, as Kubernetes already has checks, such as the metadata is not allowed to get changed, the data (ex. TLS) is valid (i think base64 or something). Since Kubernetes does those checks it shouldn't be possible to modify the resource's reference meaning our Proxy will be okay since the secret ref can't change under our toes, but I could be wrong...

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
jenshu: sorry i meant why this check `if _, hit := gloosnapshot.ApiGvkToHashableResource[gvk];` - i think that if we are making sure that it is one of the resources for which we support validation, we should also add a check for secrets, i.e. something like
```
if gvk.Group == kubernetesCoreApiGroup && gvk.Kind != "Secret" {
 // log a similar "unsupported" message
}
```

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
inFocus7: resolving in preference of the other discussion which will kill 2 birds with one stone

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
inFocus7: Made an update somewhat close to this 👍🏼 [bf010bb](https://github.com/solo-io/gloo/pull/8740/commits/bf010bba09a847f66875db85ee13474aa5b88e07)

inFocus7: Some logs I got when manually verifying though creating + updating a secret. I validation on + having, and create + update operations for secret validation. (note this was before i changed the message string to flip the words "validation" and "operation")
```
{"level":"info","ts":"2023-10-10T21:37:31.451Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:198","msg":"received validation request on webhook","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T21:37:31.452Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:406","msg":"unsupported validation operation [CREATE] for resource namespace [default] name [upstream-tls2] group [] kind [Secret]","version":"1.0.0-ci","Kind":"/v1, Kind=Secret","Namespace":"default","Name":"upstream-tls2","UID":"27b11e62-5f07-4821-a738-138ce5d87f4c","PatchOperation":"CREATE","UserInfo":{"username":"kubernetes-admin","groups":["system:masters","system:authenticated"]}}
{"level":"info","ts":"2023-10-10T21:37:31.452Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
[...]
{"level":"info","ts":"2023-10-10T21:38:24.511Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:406","msg":"unsupported validation operation [UPDATE] for resource namespace [default] name [upstream-tls] group [] kind [Secret]","version":"1.0.0-ci","Kind":"/v1, Kind=Secret","Namespace":"default","Name":"upstream-tls","UID":"1ff639b4-4dd0-4d4d-bce3-63d0d938af8a","PatchOperation":"UPDATE","UserInfo":{"username":"kubernetes-admin","groups":["system:masters","system:authenticated"]}}
{"level":"info","ts":"2023-10-10T21:38:24.511Z","logger":"gloo.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:266","msg":"ready to write response ...","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T21:38:25.212Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.TranslatorSyncer","caller":"syncer/translator_syncer.go:92","msg":"begin sync 17330435888742097365 (1 virtual services, 2 gateways, 0 route tables)","version":"1.0.0-ci"}
{"level":"info","ts":"2023-10-10T21:38:25.212Z","logger":"gloo.v1.event_loop.setup.gloosnapshot.event_loop.TranslatorSyncer","caller":"syncer/translator_syncer.go:132","msg":"desired proxy name:\"gateway-proxy\"  namespace:\"default\"","version":"1.0.0-ci"}
[...]
```

jenshu (approved):

bewebi (approved):

inFocus7:
/kick

bulldozer didn't automatically merge, kicking tests to retry them and see if bulldozer merges afterward

PR #8750, opened 2023-10-05
1.15: Add tests for non custom probes and add wget back

nfuden:
# Description

Add a failing test for noncustom readiness probes.
Add wget back to ourgloo and gateway-proxy pods

## Code changes
- New Test option
- Docker files get wget

# Context

We do not currently test our default probes. We should.
Ubuntu update broke non-custom probes. This is a quick fix.

Not needed for any alpine setups as it comes with wget by default https://pkgs.alpinelinux.org/package/edge/main/x86/wget 
We probably should consider if we want to update our probes to use curl in the future though.

## Interesting decisions
 
We chose to do things this way because we dont test all of our helm install options fully.

## Testing steps

I manually verified behavior by  checking for wget + sleep in our new docker files.
Added failing test as first commit  and then had it run here with changelog https://github.com/solo-io/gloo/actions/runs/6419883299/job/17430709129

### DIFF
--- a/test/kube2e/helm/artifacts/helm.yaml
+++ b/test/kube2e/helm/artifacts/helm.yaml
@@ -35,3 +35,5 @@ gloo:
 gatewayProxies:
   gatewayProxy:
     healthyPanicThreshold: 0
+    podTemplate:
+      probes: true
--- a/changelog/v1.15.9/reenable-probes.yaml
+++ b/changelog/v1.15.9/reenable-probes.yaml
@@ -1,0 +1,8 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/8749
+    resolvesIssue: false
+    description: > 
+      Re-enable wget in our gloo and gateway-proxy pods.
+      Wget is currently used in our non-custom probes which are currently not well covered in tests.
+      Caused by https://github.com/solo-io/solo-projects/issues/5344 shift to ubuntu.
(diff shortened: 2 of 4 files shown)


jenshu (approved):

test/kube2e/helm/artifacts/helm.yaml
```diff
@@ -34,4 +34,6 @@ gloo:
     livenessProbeEnabled: true
 gatewayProxies:
   gatewayProxy:
-    healthyPanicThreshold: 0
\ No newline at end of file
+    healthyPanicThreshold: 0
+    podTemplate:
```
sam-heilbron, line 38: nit: When we introduce this to our main branch, do you think it would be worthwhile to add this to all of our suites since it's a recommended default? Or is that something @davidjumani will be doing in his work

test/kube2e/helm/artifacts/helm.yaml
```diff
@@ -34,4 +34,6 @@ gloo:
...
+      probes: true
```
sam-heilbron, line 39: This proves that our default probes (using wget) will work. Since users can define custom probes, are there any other types of probes that might not work with these new alpine images?

sam-heilbron (approved):

test/kube2e/helm/artifacts/helm.yaml
nfuden: its possible

nfuden:
Updated with comments

sam-heilbron (approved):

PR #8758, opened 2023-10-06
Istio Spiffe fix 1.13

bewebi:
# Description

This PR backports the fix for our Istio sidecar definition as introduced in #8676

The Helm fix has been backported
The new kube2e tests and docs diverged to much to make cherry-picking worthwhile

The [support matrix](https://docs.solo.io/gloo-edge/latest/reference/support/) claims Gloo 1.13 supports up to Istio 1.15, so we update our tests and values to use the latest (last) Istio 1.15 release


inFocus7 (commented):
> The [support matrix](https://docs.solo.io/gloo-edge/latest/reference/support/) claims Gloo 1.13 supports up to Istio 1.15, so we update our tests and values to use the latest (last) Istio 1.15 release

Sounds reasonable, I also saw that the new mount paths are added/expected in Istio 1.14+1.15, so makes sense to have them here as well 👍🏼

inFocus7 (approved):

sam-heilbron (approved):

PR #8768, opened 2023-10-10
Adding helm tests for production recommendations

davidjumani:
# Description

Adds helm kube e2e tests to mimic the recommendations for production deployments mentioned [here](https://docs.solo.io/gloo-edge/latest/operations/production_deployment/)

# Context

This is to ensure that we test gloo as users would deploy it in production and catch potential issues early

## Interesting decisions
 
N/A

## Testing steps

Run the helm kube2e tests

## Notes for reviewers

The issue mentioned [here](https://github.com/solo-io/gloo/pull/8768/files#diff-0a35ec0dc12ec375e603ca73e843938aaecb0a71f5390ed3f6abe21d05d74bd8R9-R13) needs to be fixed

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8804


test/kube2e/helm/helm_test.go
```diff
@@ -245,15 +248,12 @@ var _ = Describe("Kube2e: helm", func() {
 		})
 	})
 
-	// The aim of the test is to ensure that the readiness probe is configured on the gateway proxy
-	// and the gateway-proxy deployment is ready before helm marks the upgrade as successful.
-	// Ref: https://github.com/solo-io/gloo/issues/8288
-	Context("Custom readiness probe", func() {
```
davidjumani, line 251: Replacing the custom readiness probe test with the Production recommendations tests as the latter tests the former along with other prod settings

test/helpers/kube_dump.go
```diff
@@ -85,13 +85,19 @@ func recordKubeState(f *os.File) {
 		f.WriteString("*** Unable to get kube state ***\n")
 		return
 	}
+	kubeDescribe, err := kubeCli.KubectlOut(nil, "describe", "all", "-A")
```
sam-heilbron, line 88: Can you help me understand (or better yet, include in an inline comment) what information we are looking to get out of this command?

sam-heilbron (commented):
I like this and I think it's a nice step towards codifying our production recommendations! For each of these tests, we assert that the components are health (which is good) but not much beyond that. I had kind of envisioned we could also update the other suites (https://github.com/solo-io/gloo/blob/main/test/kube2e/gloo/artifacts/helm.yaml for example) so that when we ran the tests we configured as many production recommendations as well. That way we assert both that everything is healthy AND that the behavior matches what we would expect. What do you think about that?

davidjumani:
Thanks @sam-heilbron ! I would like to use these same prod recommendations for all the tests but setting `settings.disableKubernetesDestinations` && `global.glooRbac.namespaced` (which is the standard in all our tests) leads to panic in gloo (https://github.com/solo-io/gloo/issues/8801), which is why this was added as a separate test.
Most of the other recommendations such as pod disruption budgets, gzip, etc. have their own tests (albeit not e2e or helm tests). I can add those as well if you think it will be better to test them all together

sam-heilbron (approved):

davidjumani:
Will create a new PR along with the custom readiness probe fix

PR #8770, opened 2023-10-11
[1.14] fix: Validate on Secret deletions

inFocus7:
_backport of #8740_

Seeing flaky behavior, especially on the newly added test. For some reason the VS/Proxy does not point out the error due to deleting the used Secret. When I `WAIT_ON_FAIL`ed, it took a while longer (~1 minute, although I also added an annotation on the VS so not sure which did it) before the VS/Proxy had an error due to the deleted Secret. Seems to flake ~1/2 of the time.

I've ran the test on the `1.15` branch 5 times in a row with no fail; in this one I get the test fail ~1/2 runs, so something differs between 1.14.x and 1.15.x 🤔

Manual installation + verification of this seems okay ([re-]installed 5 times without issue), so this seems to be an issue specific to the tests.

For now, I added `FlakeAttempts(N)`. I chose `N` because, from the test runs I've done locally, (did enough to flake 5 times) `N-1` was the max, and I'm playing it safe by doing N+1.

# Description

Fixes the issue where the Validation Webhook did not validate on `Secret` deletions. 
This was due to our `ValidatingWebhookConfiguration` configured to validate on secrets part of the `gloo.solo.io` groups, which does not exist (we don't have a secret CRD). Now we correctly validate against Kubernetes `Secret` deletions.

_proof of no secret crd, all crds from 1.16.0-beta1_
Screen Shot 2023-09-28 at 12 28 56 PM

## Backport Changes
- No performance tests, as this gloo version does not have them, and backporting is out of scope
- Modified gateway_test, as this gloo version has a different setup

## A bit of history

Secret validation was added in [this PR](https://github.com/solo-io/gloo/pull/5470/files). 
The tests added in those are valid as they **do** test that our secret validation logic works, **BUT** don't include an end-to-end test _beginning_ from an actual deletion of a secret.

If the (simplified) flow of validation is:
`[delete secret request] > [validation webhook activates] > [request gets parsed] > [internal validation] > [answer]`
The tests covered from 'internal validation' onward, so they didn't catch the case/fact that a secret is not in our CRDs. I manually tested that secrets are not being validated on `k delete secret` on the version that code was included.

The end-to-end tests added here cover from the initial flow of validation by actually deleting a secret, which are of GVK `<kube api, aka. "">/v1/Secrets`.

## API changes

- Updated our `ValidatingWebhookConfiguration` to correctly validate on Kubernetes' core secrets.
  - Those are referenced by the API Group `""`.

## Code changes

- Allow the API Group during validation to be empty `""`, allowing Kubernetes APIs to be validated, by not enforcing a match to exist in our `HashableResource`.
  - `Secrets` are part of Kubernetes core API `""`.
- Mapped the Kubernetes `Secret` to our `HashableResource` on deletion.
  - `HashableResources` (and `resource.Resource`) are maintained for **our** resources, but a lot of our validating logic requires one, so we needed a way to still validate against the Kube secret resource. Since we only care about the Secret **deletions**, we can safely do this,as deletions only care about the name + namespace, which **all** resources (including Kube's) have.
- Added tests to verify a validation webhook occurs (and doesn't) depending on conditions.
- Added secrets to our translation performance tests. Used the basicConfig, as its constraints worked for `Secrets` as well.

## Performance Implications

Performance will likely face a hit for users **with** `webhook validation` enabled **when** deleting a `Secret`, as we're now correctly validating their deletion. The performance hit during Secret deletion will be similar to any other validation we have on resources, but may be more noticable in environments rotating and deleting secrets more consistently.

To opt-out, you can add `'secrets'` to our following api `gateway.validation.webhook.skipDeleteValidationResources`.

# Context

When deleting secrets when the validating webhook was enabled, users noticed that their deletion were not being validated. This meant that they would always successfully delete a secret, even if it causes a proxy error when another resource (ex. VirtualService) referenced it.

## Interesting decisions
 
- Mapped the Kubernetes core secret API `""/v1/Secret` (where `""` represents the Kube core APIs) to our `Secret` `HashableResource`.
  - This had to be done since our validation functions require an internal `resource.Resource`.
  - I did not create a custom `HashableResource` type for the Kubernetes secret, as (from the looks of it) they are meant for **our** APIs, whereas Kubernetes is not.
  - This works in _this_ scenario as the deletion of a resource _only_ requires the `name` and `namespace` for the resource, and not any other internal APIs.

### 1. Install Gloo
**validation-webhook.yaml**
```yaml
gateway:
  enabled: true
  validation:
    enabled: true
    alwaysAcceptResources: false
    webhook:
      enabled: true
```

_To see the issue, install a released (or non-updated) version_
```bash
helm install gloo gloo/gloo --namespace default --create-namespace --version 1.16.0-beta1 --values validation-webhook.yaml
```

_To notice the fix install this branch locally_
```bash
./ci/kind/setup-kind.sh
```
```bash
helm install gloo _test/gloo-1.0.0-ci.tgz --values validation-webhook.yaml
```

#### Create a certificate + key
We'll need a certificate and key for TLS secrets. Since we don't need to test the functionality of SSL, and just need to reference the cert + key, can simply create & use a key created from the following - with whatever certificate data you want (or default):
```bash
openssl req -new -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -out tls.crt -keyout tls.key
```

#### Create Secret(s) using created cert + key
_We are creating 3 secrets, in order to manually verify the different possibilities_
```bash
k create secret tls upstream-tls --key tls.key --cert tls.crt
```
```bash
k create secret tls upstream-tls2 --key tls.key --cert tls.crt
```
```bash
k create secret tls upstream-tls3 --key tls.key --cert tls.crt
```

### 3. Create VirtualService
```bash
cat << EOF | kubectl apply -f -
kind: VirtualService
metadata:
  name: direct-response
  namespace: default
spec:
  sslConfig:
    secretRef:
      name: upstream-tls
      namespace: default
  virtualHost:
    domains:
    - '*'
    routes:
    - matchers:
      - prefix: /
      directResponseAction:
        status: 200
        body: "Hello, world!"
EOF
```

### 4. Delete the Secrets (manual verification)
**Without branch fix**
You can delete all resources without issue
```bash
k delete secret upstream-tls
```
```bash
k delete secret upstream-tls2
```
```bash
k delete secret upstream-tls3
```

**With branch fix**
1. **`upstream-tls2`** should delete successfully, as it is not referenced elsewhere.
```bash
k delete secret upstream-tls2
```
6. **`upstream-tls`** should fail to delete due to it being used in the **`direct-response`** VS.
```bash
k delete secret upstream-tls
```
7. Edit the **`direct-response`** VS to point to **`upstream-tls3`**.
```bash
k edit vs direct-response
# Point the secret ref to upstream-tls3
```
8. **`upstream-tls`** should now delete, as it's not referenced anywhere
```bash
k delete secret upstream-tls
```
9. **`upstream-tls3`** should fail to delete, due to it being used in the `direct-response` VS.
```bash
k delete secret upstream-tls3
```
10. Delete the **`direct-response`** VS, so nothing references the **`upsteram-tls3`**.
```
k delete vs direct-response
```
11. **`upstream-tls3`** should now delete, as it's not reference anywhere.
```
k delete secret upstream-tls3
```

## Notes for reviewers

If we fear the implications on performance on this fix, we _could_ add a helm field for validation to enable/disable secret validation and gate `webhookvalidationconfiguration`'s secret ops. (unless something like this exists already)

# Checklist:

- ~I have made corresponding changes to the documentation~


inFocus7:
not backporting

PR #8791, opened 2023-10-12
Preconnectpolicy

nfuden:
# Description

This feature allows preconnect policy for an envoy cluster to be configured on a gloo upstream.
It is not configurable at any higher level at this time and likely never should be.
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-msg-config-cluster-v3-cluster-preconnectpolicy

Preconnect policy should almost never be used unless a misbehaving upstream is prone to initial issues on connection that resolve fairly quickly 

## API changes
- Added  Preconnect policy to upstreams. This mimicks closely the envoy cluster variant

## Code changes
Cluster translator now supports the new field and will return errors if the configured values fall outside of the acceptable proto based constraints in upstream https://github.com/envoyproxy/envoy/blob/353c0a439ef8bc8eb63bf08c2db4fd3fc3778dec/api/envoy/config/cluster/v3/cluster.proto#L724

## Docs changes
- Did not add anything other than proto docs but called out loudly there that users in general should not configure this new option

# Context
There exists some instances when working with another party who owns an upstream that either drags their feet on fixing their noncompliant behavior and their failure case is well known such that configuring this option makes sense.

## Testing steps

In addition to our automated "can it set" type tests to test this manually I set preconnect perupstream to a go service I was running and checked for number of connections without sending a request. Then I sent a request and saw that no new connections were created

# Checklist:


projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3651,6 +3651,45 @@ var _ = Describe("Translator", func() {
 			Entry("When value=nil", nil, false))
 	})
 
+	FContext("PreconnectPolicy", func() {
```
davidjumani, line 3654: Focused Test

davidjumani (approved):

projects/gloo/pkg/translator/clusters.go
```diff
@@ -291,6 +298,45 @@ func getCircuitBreakers(cfgs ...*v1.CircuitBreakerConfig) *envoy_config_cluster_
...
+		eString := ""
+		if perUpstream != nil && (perUpstream.GetValue() < 1 || perUpstream.GetValue() > 3) {
+			eString = fmt.Sprintf("perupstream must be between 1 and 3 if set, it was: %v", perUpstream.GetValue())
+		}
+		if predictive != nil && (predictive.GetValue() < 1 || predictive.GetValue() > 3) {
+			eString += fmt.Sprintf("predictive must be between 1 and 3 if set, it was: %v", perUpstream.GetValue())
+		}
+		if eString != "" {
+			return nil, errors.New("invalid preconnect policy: " + eString)
+		}
```
bewebi, line 330: As is, if both errors are present there will be no space
Perhaps something like this:
```suggestion
		eStrings := []string{}
		if perUpstream != nil && (perUpstream.GetValue() < 1 || perUpstream.GetValue() > 3) {
			eStrings = append(eStrings, fmt.Sprintf("perupstream must be between 1 and 3 if set, it was: %v", perUpstream.GetValue()))
		}
		if predictive != nil && (predictive.GetValue() < 1 || predictive.GetValue() > 3) {
			eStrings = append(eStrings, fmt.Sprintf("predictive must be between 1 and 3 if set, it was: %v", perUpstream.GetValue()))
		}
		if len(eStrings) > 0 {
			return nil, fmt.Errorf("invalid preconnect policy: %s", strings.Join(eStrings, "; ")
		}
```

(hand written in browser not guaranteed to compile)

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3651,6 +3651,45 @@ var _ = Describe("Translator", func() {
...
+				Expect(actualPrecon.PerUpstreamPreconnectRatio).To(Equal(upstreamValue.PerUpstreamPreconnectRatio))
+
+				Expect(actualPrecon.PredictivePreconnectRatio).To(Equal(upstreamValue.PredictivePreconnectRatio))
+			},
+			Entry("When unset", nil, false),
+			Entry("When valid perupstream", &v1.PreconnectPolicy{PerUpstreamPreconnectRatio: asDouble(1)}, false),
+			Entry("When valid predictive", &v1.PreconnectPolicy{PredictivePreconnectRatio: asDouble(2)}, false),
+			Entry("When valid", &v1.PreconnectPolicy{PerUpstreamPreconnectRatio: asDouble(1), PredictivePreconnectRatio: asDouble(2)}, false),
+			Entry("When invalid perupstream", &v1.PreconnectPolicy{PerUpstreamPreconnectRatio: asDouble(0.5)}, true),
+			Entry("When valid perupstream but invalid predictive", &v1.PreconnectPolicy{PerUpstreamPreconnectRatio: asDouble(1), PredictivePreconnectRatio: asDouble(1000)}, true),
```
bewebi, line 3689: Add an entry for both invalid

projects/gloo/pkg/translator/clusters.go
```diff
@@ -291,6 +298,45 @@ func getCircuitBreakers(cfgs ...*v1.CircuitBreakerConfig) *envoy_config_cluster_
 	return nil
 }
 
+// getPreconnectPolicy follows the naming of the rest of functions here
+// it MAY someday deal with inheritance like the other functions
+// it consumes an ordered list of preconnect policies
+// it returns the first non-nil policy
+func getPreconnectPolicy(cfgs ...*v1.PreconnectPolicy) (*envoy_config_cluster_v3.Cluster_PreconnectPolicy, error) {
+
+	// since we dont want strict reliance on envoys current api
```
bewebi, line 307:
```suggestion
	// since we dont want strict reliance on envoy's current api
```

bewebi (commented):
These are nits, and can be ignored if time pressure requires

ashishb-solo (approved):

projects/gloo/pkg/translator/translator_test.go
ashishb-solo, line 3689: I've already approved, but in the future, it would be helpful to be a bit more specific with explanations of why we expect errors (ie. `PredictivePreconnectRatio` should be `< 3`

projects/gloo/pkg/translator/clusters.go
nfuden: I like it, have added it

bewebi (approved):

projects/gloo/pkg/translator/translator_test.go
nfuden: Good point I put in the pr description not the pr itself. updating

ashishb-solo (approved):

nfuden:
Thanks all!

PR #8795, opened 2023-10-12
Fix kube2e gateway_test flakes due to recent secret deletion validation

inFocus7:
# Description

Fixing flakes seen by adding retries.

### "secret validation [It] should act as expected with secret validation" flake
The `FlakeAttempt` is a general (hot)fix.

There are times during our CI tests when we _do not_ get a validation error when deleting a `Secret` that is referenced elsewhere. 
This was an issue I previously hit constantly in 1.14- for 1/2 of runs, but after ~20 runs, I haven't reproduced it locally. In the 1.14 tests, I saw that the VS/proxy did not have an error status until way after (~1 minute, plus I added an annotation to try and force an update). Something similar could be happening here?

My only hypothesis is that resources (specifically the Secret) are created and deleted _too fast_... maybe even _before_ the VS updates to reference it, which is why it deletes with no issue. If not, then more thorough investigation needs to be done either in our tests and/or translation + proxy code.

### "native ssl [It] works with ssl" flake
Secrets are now handled through our SnapshotWriter. This avoids the previous possibility of the Secret failing to be deleted because the VirtualService referencing it still existed. The SnapshotWriter handles creating + deleting resources in order of dependencies, so it correctly creates + deletes secrets in relation to resources that reference it.

## Code changes
- Create a 'Secret creator' helper that returns an internal Secret - as opposed to the existing one that returns a kube secret.
- Passing Secrets into glooResources so the SnapshotWriter handles creating + deleting them.
- Add `FlakeAttempts` to flaky test that needs further investigation.

# Context

We saw these gateway_test flakes in recent nightly tests, after introducing delete validation through #8740.
- secret failed to delete flake: [1](https://github.com/solo-io/gloo/actions/runs/6491470077), [2](https://github.com/solo-io/gloo/actions/runs/6504230547)
- no error occurred when deleting secret flake: [1](https://github.com/solo-io/gloo/actions/runs/6478192275)

## Interesting decisions
 
Decided on using `FlakeAttempts` for one of the flakes. I have not been able to reproduce locally as of now, so this is just a patch in the meantime. This was something I previously ran in 1.14-, the difference being that in the older version it happened in ~1/2 of runs, whereas I haven't hit this yet (after 20 runs).

# Checklist:


inFocus7:
note-to-self: backport to 1.15, as tests exist there as well

test/kube2e/gateway/gateway_test.go
```diff
@@ -2532,38 +2521,39 @@ spec:
...
-						Not(HaveOccurred()),
-						MatchError(errors.NewNotFound(corev1.Resource("secrets"), secretName).Error()),
-					))
-				})
-
-				It("should act as expected with secret validation", func() {
+				// There are times when the VirtualService + Proxy do not update Status with the error when deleting the referenced Secret, therefore the validation error doesn't occur.
+				// It isn't until later - either a few minutes and/or after forcing an update by updating the VS - that the error status appears.
+				// The reason is still unknown, so we retry on flakes in the meantime.
+				It("should act as expected with secret validation", FlakeAttempts(3), func() {
```
bewebi, line 2556: If this test flakes half the time, three attempts will only be successful 7/8 of the time
If it's truly random and 50/50 we should have five retries in order to succeed >95% of the time or seven for >99% of the time

test/kube2e/gateway/gateway_test.go
```diff
@@ -1368,9 +1361,12 @@ var _ = Describe("Kube2e: gateway", func() {
 			)
 
 			BeforeEach(func() {
+				// get the certificate so it is generated in the background
+				go helpers.Certificate()
```
bewebi, line 1365: I see that we're already doing this elsewhere in the suite, but I'm not clear why we want to do this async
I get that we don't actually need the return value, but it seems like it would be safer to wait for it to return before proceeding

test/kube2e/gateway/gateway_test.go
inFocus7: Oh wait, my description might be confusing
**Failed half the time on 1.14 and less, when I was previously backporting (before we decided not to). As of on `main`, I haven't gotten it to fail locally after 20+ runs, and we've only seen it fail once in CI/nightlies so far, so hard to gague how often here which is why I did 3 (for now at least)

bewebi (commented):
Overall LGTM, just a couple comments

test/kube2e/gateway/gateway_test.go
bewebi: Seems like a judgement call...tricky to find the balance between ensuring the suite passes consistently while also having the test still be meaningful
If we haven't seen failures often on branches where this test actually is merged then three is probably fine

test/kube2e/gateway/gateway_test.go
inFocus7: I'm actually wondering if these are even necessary... They are called before a `GetKubeSecret` and `GetTlsSecret`, who call the `Certificate()` function themselves, so looks like either way we're going to generate/get a certificate whether or not we do the `go helpers.Certificate()`.

bewebi (approved):

test/kube2e/gateway/gateway_test.go
```diff
@@ -2532,38 +2518,36 @@ spec:
 		})
 
 		Context("resource CRUD validation", func() {
+			BeforeEach(func() {
```
sam-heilbron, line 2521: We have a batch of other tests around validation inside of a Context labeled "Validation Configuration". We configure settings up front, and then run them in an ordered container to avoid changing settings (which cause there to be a delay in propagation). Is there a reason we don't have the secret validation exist within this context? We have a list of test cases that I imagine we could add to

inFocus7: Do you mean moving the context above a level? The `resource CRUD validation` context is within the "validation configuration" one, although I can move the specific secret validation up if wanted. This isn't a part of the other Context/When's as:
1. Warnings do not matter, so placing it in the `allowWarnings=` true & false doesn't accomplish anything imo.
2. `DisableTransformationValidation` doesn't relate to the secret deletion validation
3. The `When(alwaysAccept=true)`, also doesn't make sense since we want it to be false.

I could rename this context to `When(alwaysAccept=false)` and make it ordered to match the other blocks though

inFocus7: Talked with Sam in Slack. Moving to `allowWarnings=false` as its a strict check which still accomplishes what we're testing for with secret deletion

bewebi (approved):

sam-heilbron (approved):

PR #8807, opened 2023-10-18
Sha1 for basic auth

sheidkamp:
# Description

Adding sha1 support to basic auth

## API changes
2 New fields were added to support the changes:
- `encryption` - defines which hashing algorithm to use
- `user_source` - defines where the list (map) of valid users and credentials comes from
  - We currently support only `user_list`, an inline list of users, but now have the ability to define other sources, for example k8s secrets.
 
These new fields can not be used in combination with the existing `apr` field.

example:
```
spec:
  configs:
  - basicAuth:
      encryption: 
        sha1: {}
      userList:
          users:
            testuser:
              salt: "ef6dba8b-57ff-448c-819f-9a9b3a169769"
              hashedPassword: "8BvzLUO9IfGPGGsPnAgSu1"
```
## Code changes
In this repository, only the protobuf file and its generated artifacts were changed.

## Docs changes
Inline protobuf documentation

# Context
Client wants to use the insecure SHA1 hashing algorithm to support migration to gloo from a solution already using SHA1.

## Interesting decisions
discussed in the design doc.
https://soloio.slab.com/posts/sha-1-for-ext-auth-basic-auth-api-design-doc-r7bz7cy8

## Testing steps
Follow the steps in the [Basic Auth Guide](https://docs.solo.io/gloo-edge/latest/guides/security/auth/extauth/basic_auth/), and use an authconfig that looks like:

```
apiVersion: enterprise.gloo.solo.io/v1
kind: AuthConfig
metadata:
  name: basicauth
  namespace: gloo-system
spec:
  configs:
  - basicAuth:
      encryption:
        sha1: {}
      userList:
        users:
          sha1user:
            hashedPassword: 002112fbb9e58a85ec2bfd220c581d60896f4507
            salt: ef6dba8b-57ff-448c-819f-9a9b3a169769
```

and use the following curl:
`curl -vk https://glootest-basic.com:32500/get  -H "Authorization: basic c2hhMXVzZXI6b1YkOmJEMVBh"`
which is equivalent to:
`curl -vk https://glootest-basic.com:32500/get -u sha1user:oV$:bD1Pa`

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -257,16 +257,62 @@ message AuthPlugin {
 message BasicAuth {
     string realm = 1;
 
+    // This is the legacy/simple basic auth config. It supports the APR hashing algorithm and an inline userlist.
+    // If apr is defined, encryption and user_source must not be defined.
     message Apr {
```
inFocus7, line 262: Q: were we going to deprecate this field in favor of the encryption+userList field? If so, should/could we add `[deprecated = true]`?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -257,16 +257,62 @@ message AuthPlugin {
...
             string salt = 1;
+            // Salted and hashed password for the user
             string hashed_password = 2;
         }
+        // Map of authorized usernames to stored credentials
         map<string, SaltedHashedPassword> users = 2;
     }
     Apr apr = 2;
+
+    // Below here is the "extended" basic auth config. Hashing algorithm and user source are independent and configurable.AccessToken
```
inFocus7, line 275: Is the `AccessToken` word misplaced/a type?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -257,16 +257,62 @@ message AuthPlugin {
...
+    // It is required to define exactly one of "apr" and ("encryption" + "user_source")
```
inFocus7, line 276:
```suggestion
    // It is required to define exactly one of "apr" or ("encryption" + "user_source")
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -257,16 +257,62 @@ message AuthPlugin {
...
+        // Apache specific iterated MD5 hashing: (https://httpd.apache.org/docs/2.4/misc/password_encryptions.html)
+        message Apr {}
+        oneof algorithm {
+            Apr apr = 1;
+            Sha1 sha1 = 2;
+        }
+    }
+
+    // The encryption type to use to store the password on the server
+    // If encryption is defined user_source must be defined and the top level apr field must not be defined
```
inFocus7, line 292:
```suggestion
    // If encryption is defined, 'user_source' must be defined, and the top-level 'apr' field must not be defined
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -257,16 +257,62 @@ message AuthPlugin {
...
+        string hashed_password = 2;
+    }
+
+     // Map of valid usernames to stored credentials
+    message UserList {
+        map<string, User> users= 1;
+    }
+
+     // Source of user credential data
+     // If user_source is defined encryption must be defined and the top level apr field must not be defined
```
inFocus7, line 309:
```suggestion
     // If user_source is defined, encryption must be defined, and the top-level apr field must not be defined
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1908,6 +1993,7 @@ message ExtAuthConfig {
             PassThroughAuth pass_through_auth = 13;
             HmacAuthConfig hmac_auth = 15;
             OpaServerAuthConfig opa_server_auth = 16;
+            BasicAuthExtended basic_auth_extended = 17;
```
inFocus7, line 1996: q: should we deprecate the existing `BasicAuth basic_auth = 4;`?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1392,6 +1438,45 @@ message ExtAuthConfig {
     // Any request to the external auth server includes an identifier that is matched against this field to determine
     // which AuthConfig should be applied to it.
     string auth_config_ref_name = 1;
+
+    // Message to store Basic Auth Configuration.
+    // "Extended" refers to this format allowing for selection of the hashing algorithm and user source.
+    // If only the legacy "apr" field is defined, the existing public BasicAuth configuration will continue be used.
+    message BasicAuthExtended {
```
inFocus7, line 1445: I may have forgotten if explained in our meetings, but we do we create a separate message for this extended one instead of extending the existing `BasicAuth` type? 👀

inFocus7 (commented):
a few question, generally lgtm (although have not tried locally*)

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sam-heilbron, line 1996: nit: Following the pattern of Ldap, we may call this basic_auth_internal. Also, what do you think about defining this message (even though it will stil be field number 17) just after the basic auth so it's easier for devs to find both references

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1392,6 +1438,45 @@ message ExtAuthConfig {
...
+        }
+
+        // Map of valid usernames to stored credentials
+        message UserList {
+            map<string, User> users= 1;
+        }
+
+        // Source of user credential data.
+        oneof user_source {
+            UserList user_list = 4;
```
sam-heilbron, line 1475: nit: Should/could this just be field number 3?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: I thought about this and talked to Sam a bit as well, and I think we want to keep this supported, especially since its in so much documentation and has literally worked for everyone else for years.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sam-heilbron, line 292: When you say "must not be defined", can you clarify what the behavior will be? Is it that the confgiuration will be considered invalid and an error will be returned? Is it that one field will be respected over the other?

changelog/v1.16.0-beta17/sha1-for-basicauth.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink:  https://github.com/solo-io/solo-projects/issues/5234
+    description: >
+      Add support to use sha1 for basic auth. This requires extended the basic auth API to allow selection of an excyption algorithm.
+      While extending the API, we had to create a new method of defining the users, and we took this opportunity to add support for
```
sam-heilbron, line 6: I like this callout in the PR "this opportunity to add support for sources of data other than an inline list in anticipation of being asked to support user credentials from a secret." though I'm not sure this needs to be in the changelog since that is user facing. Is there a reason you think we should includd this?

changelog/v1.16.0-beta17/sha1-for-basicauth.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: nit: I've tended to make API changes non-user facing in the oss repo, and then the code that introduces the feature in gloo enterprise actually a new_feature

sam-heilbron (commented):
Looking good! This matches the general structure of the API we had discussed. Can you put up the other PRs so that we can see this working end to end before merging?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: I probably didn't go over this, but the internal interface was just using the same object as the public interface. So we needed a new object on the internal interface. It is basically the same, (but doesn't have to be), and I think its best practice to have interfaces separate.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: probably accidentally accepted an errant codepilot suggestion, will update

changelog/v1.16.0-beta17/sha1-for-basicauth.yaml
sheidkamp: 👍

changelog/v1.16.0-beta17/sha1-for-basicauth.yaml
sheidkamp: Trying to include context. Will remove.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: Behavior currently is "will fail validation", have updated messages.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: yup

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: I think this should match the behavior of the public interface (see comment there)

sheidkamp: Done and done

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: I like the first comma but not the second one. Made some other changes to this line as well.

sam-heilbron (commented):
Changes LGTM! Holding off on a formal approval until the full feature is in review and we can test it end to end

sam-heilbron (dismissed):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1396,6 +1442,45 @@ message ExtAuthConfig {
     // Any request to the external auth server includes an identifier that is matched against this field to determine
     // which AuthConfig should be applied to it.
     string auth_config_ref_name = 1;
+
+    // Message to store Basic Auth Configuration.
+    // "Extended" refers to this format allowing for selection of the hashing algorithm and user source.
```
sam-heilbron, line 1447: I think this "extended" reference is no longer accurate

sheidkamp: updated

inFocus7 (approved):

sam-heilbron (approved):

PR #8836, opened 2023-10-27
Renew certs 1.14

jenshu:
backport #8643
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5944


ben-taussig-solo:
Do we want to backport #8695 as well, like we did here? https://github.com/solo-io/gloo/pull/8821

nfuden (approved):
lgtm seems safer to force these rotations and shouldnt impact too much given the standard length of our default certs

PR #8850, opened 2023-10-30
Tap filter v1.16.x

ashishb-solo:
# Description

Implement the tap filter API changes. Incorporates our discussion from the API design review from 2023-10-24

Corresponding `solo-projects` updates can be found in this pull request: https://github.com/solo-io/solo-projects/pull/5442

## API changes
- Add tap object to Settings and Gateway

## Code changes
- None

## CI changes
- None

## Docs changes
- See .protos
- Guide on Tap filter implementation outside the scope of this pull request

# Context

Users would like to have a tap filter

## Interesting decisions
 
See meeting notes from tap filter API discussion

## Testing steps

I manually verified behavior by testing the tap filter

## Notes for reviewers

# Checklist:


changelog/v1.16.0-beta19/tap-filter-proto.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
```
jenshu, line 1: can you add the go-control-plane-fork dep bump too?

go.mod
```diff
@@ -11,9 +11,9 @@ require (
 	github.com/avast/retry-go v2.4.3+incompatible
 	github.com/aws/aws-sdk-go v1.34.9
 	github.com/census-instrumentation/opencensus-proto v0.4.1
-	github.com/cncf/xds/go v0.0.0-20230428030218-4003588d1b74
+	github.com/cncf/xds/go v0.0.0-20230607035331-e9ce68804cb4
 	github.com/cratonica/2goarray v0.0.0-20190331194516-514510793eaa
-	github.com/envoyproxy/go-control-plane v0.11.1
+	github.com/envoyproxy/go-control-plane v0.11.2-0.20230819171109-107d4056d823
```
jenshu, line 16: is this needed? since are pinning to the fork

go.mod
```diff
@@ -70,8 +70,8 @@ require (
 	golang.org/x/sync v0.3.0
 	golang.org/x/tools v0.9.1
 	google.golang.org/genproto v0.0.0-20230526203410-71b5a4ffd15e // indirect
-	google.golang.org/grpc v1.55.0
-	google.golang.org/protobuf v1.30.0
+	google.golang.org/grpc v1.57.0
+	google.golang.org/protobuf v1.31.0
```
jenshu, line 74: are these needed in this PR? (they seem to have caused a lot of files to change)

projects/gloo/api/v1/enterprise/options/tap/tap.proto
```diff
@@ -0,0 +1,43 @@
...
+package tap.options.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/tap";
+
+// import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/config/core/v3/grpc_service.proto";
+import "github.com/solo-io/solo-kit/api/v1/ref.proto";
+import "validate/validate.proto";
+import "google/protobuf/duration.proto";
+
+message Tap {
```
jenshu, line 12: can you add a top level summary of what Tap is used for?

changelog/v1.16.0-beta19/tap-filter-proto.yaml
```diff
@@ -0,0 +1,6 @@
...
+- type: NEW_FEATURE
```
sam-heilbron, line 2: For Enterprise features, I tend to mark the changelog type as NON_USER_FACING in the oss repo, and then NEW_FEATURE when we introduce the actual implementation.

go.mod
```diff
@@ -70,8 +70,8 @@ require (
 	golang.org/x/sync v0.3.0
 	golang.org/x/tools v0.9.1
 	google.golang.org/genproto v0.0.0-20230526203410-71b5a4ffd15e // indirect
-	google.golang.org/grpc v1.55.0
-	google.golang.org/protobuf v1.30.0
+	google.golang.org/grpc v1.57.0
```
sam-heilbron, line 73: There are a lot of dependency updates, and as a developer/reviewer it's challenging to understand which are required. Can you ensure that relevant libraries are reflected in hte changelog update, and that any notable upgrades are called out in the PR body so reviewers understand why they are required?

As an aside, when there are many depdency updates, which impact code generation, it can be hepful to split the PR into 2: one two upgrade deps, the other to introduce the API.

projects/gloo/api/external/envoy/config/tap/output_sink/v3/grpc_output_sink.proto
```diff
@@ -0,0 +1,22 @@
...
+
+import "validate/validate.proto";
+import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/config/core/v3/grpc_service.proto";
+
+// [#protodoc-title: GRPC Output Sink]
+// An output sink for the TAP filter that writes to a GRPC socket
+// [#extension: envoy.tap.sinks.solo.grpc_output_sink]
+
+// GRPC output sink definition
+message GrpcOutputSink {
```
sam-heilbron, line 18: Why do we have an output sink wrap a grpc service? Can you help me understand what other types we expect to live on this output sink object?

projects/gloo/api/v1/options.proto
```diff
@@ -216,6 +217,9 @@ message HttpListenerOptions {
     // Router is an extension of the envoy http filters
     // Maps to https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto
     Router router = 18;
+
+    // Tap filter settings
```
jenshu, line 221: is it possible to have tap globally enabled (via the field in settings.proto) but disable it for a certain listener?
e.g. see [extproc listener settings](https://github.com/solo-io/gloo/blob/196ba6c24b48cacb413b07483a5db4d92d4aba02/projects/gloo/api/v1/options.proto#L148-L157)

sam-heilbron, line 221: I imagine this is an Enterprise-only piece of configuration, correct? If so, I think we should reflect that this comment.

Is this API subject to change? Is it worth marking this as an experimental or alpha API so that if it has to change in the near future we can break it?

projects/gloo/api/v1/settings.proto
```diff
@@ -551,6 +552,9 @@ message Settings {
     // defaults globally, and can be overridden by HttpListenerOptions, VirtualHostOptions,
     // or RouteOptions.
     extproc.options.gloo.solo.io.Settings ext_proc = 39;
+
+    // Enterprise-only: Tap filter settings.
```
sam-heilbron, line 556: We have an enterprise_warning plugin (https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/enterprise_warning/plugin.go) whose job is to perform minimal warnings when open source users attempt to configure enterrpsie only APIs. If we're adding an enterprise-only API, can we add the subsequent warning to this plugin?

sam-heilbron (commented):
Looking good! Mostly minor comments. Can you link the Enterprise equivalent code so that we can validate this end to end before merging this API?

projects/gloo/api/v1/options.proto
```diff
@@ -216,6 +217,9 @@ message HttpListenerOptions {
...
+    tap.options.gloo.solo.io.Tap tap = 34;
```
jenshu, line 222: would be good to add a sentence that this overrides the global tap settings

changelog/v1.16.0-beta19/tap-filter-proto.yaml
ashishb-solo: added that in this commit: 2cea1929568668a5ae03aa2c42f1f18ac6d978eb

go.mod
ashishb-solo: reverted this change here: 1fbe1b85aa642012dfe9bab92f925fab95ccfa1b

projects/gloo/api/v1/enterprise/options/tap/tap.proto
ashishb-solo: added this here: 8fda510658827f9ef792bfdfe2b7d971c62e3660

changelog/v1.16.0-beta19/tap-filter-proto.yaml
ashishb-solo: changed the changelog type to NON_USER_FACING in this commit: 2cea1929568668a5ae03aa2c42f1f18ac6d978eb

projects/gloo/api/v1/options.proto
ashishb-solo: this is very tricky and going to be a hard one to figure out. it has already been suggested that we may want to support multiple tap filters at some point in the future. figuring out the interplay between that in addition to settings/gateway overrides will be something we might need to have a larger discussion over.

since this is an experimental api for now, i would suggest that we put it into the milestone as one to come back to later

projects/gloo/api/v1/settings.proto
ashishb-solo: added support for that here: cad5a52098bcd3cec324720dae03aa58b36982ac

projects/gloo/api/v1/options.proto
ashishb-solo: added that here: 0dd33ff0a74ba9b715a7a71b95044607dcaf703f

projects/gloo/api/v1/options.proto
ashishb-solo: This commit should cover that: 0dd33ff0a74ba9b715a7a71b95044607dcaf703f

Let me know if you had something else in mind regarding the labeling of experimental/alpha status

projects/gloo/api/v1/enterprise/options/tap/tap.proto
```diff
@@ -0,0 +1,43 @@
...
+
+    // Write tap data out to a HTTP service
+    HttpService http_service = 2;
+  }
+}
+
+// A tap sink over a GRPC service
+message GrpcService {
+  // Upstream reference to the tap server
+  core.solo.io.ResourceRef tap_server = 1 [(validate.rules).message = {required: true}];
```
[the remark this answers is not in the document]
ashishb-solo: we're not using any of them right now in the enterprise plugin, so i would be inclined not to do so for now, especially considering that the supported featureset is not yet finalised and the api is still experimental.

are there any particular fields from here that you think are necessary to have implemented up front? happy to consider adding them if you think so... if you could drop a comment on [this issue](https://github.com/solo-io/solo-projects/issues/5314) for fields that you think should be added, i think that might be the best way to handle this for now. does this sound reasonable?

go.mod
ashishb-solo: my apologies for this. i think the extproc bump pulled in some dependency changes that might have triggered the massive list of changes. i tried reverting them (https://github.com/solo-io/gloo/pull/8850/commits/1a8ff9372bdd901092ff9839a1117cf1ccbfdeee) but codegen re-reverted them! so they appear to be necessary.

> As an aside, when there are many depdency updates, which impact code generation, it can be hepful to split the PR into 2: one two upgrade deps, the other to introduce the API.

i'll keep this in mind in the future though. when i have some time, i'll try to write a comprehensive list of files that require review and which ones can be ignored.

projects/gloo/api/v1/options.proto
ashishb-solo: this is the rudimentary approach we're using for now: https://github.com/solo-io/solo-projects/pull/5442/files#diff-ae6a076fec6047261050c2d715da624fe914e210aea6dc70e422d88feefcd766R117

would love if you have feedback on this, since this will hopefully be up for review tomorrow :-P

go.mod
ashishb-solo: as mentioned [here](https://github.com/solo-io/gloo/pull/8850/files/1a5d7e8dba215a97ae5737b6114190dd4849e8e8#r1376769532), i tried to avoid these changes, but they seem necessary. I think they might be transient dependencies from go-control-plane that were pulled in when we updated our definitions to incorporate tap filter support from upstream envoy

projects/gloo/api/external/envoy/config/tap/output_sink/v3/grpc_output_sink.proto
ashishb-solo: i think i might be misunderstanding what you're asking here... our objective is to achieve the following envoy configuration:

```
                          sinks:
                            - format: PROTO_TEXT
                              # file_per_tap:
                              #   path_prefix: /dev/shm/taps/taps
                              custom_sink:
                                name: custom_sink
                                typed_config:
                                  "@type": type.googleapis.com/envoy.config.tap.output_sink.v3.GrpcOutputSink
                                  grpc_service:
                                    envoy_grpc:
                                      cluster_name: tap_server
```

in order to do so, we need to instantiate an object in envoy that implements [this interface](https://github.com/envoyproxy/envoy/pull/28808/files#diff-eba1e4dfe999e97fde97408f49d81f77a83abed97c3d59d46263a177ec4ab131R86-R97). for this reason, we can't use a grpc_service directly and have to use an object that wraps it.

does this answer your question? happy to talk further, because i'm not 100% sure i understand where you're coming from here

ashishb-solo: so i think tl;dr: we need to wrap it in order to instantiate an object that satisfies the linked envoy interface. i think that might be the answer to your question....

go.mod
ashishb-solo: After pulling in master, the only updates now are this:

```
ashish@localhost ~/go/src/github.com/solo-io/gloo $ git diff origin/main... go.mod 
diff --git a/go.mod b/go.mod
index 01115b5ce..2913bc5e0 100644
--- a/go.mod
+++ b/go.mod
@@ -275,7 +275,7 @@ require (
        github.com/xlab/treeprint v1.1.0 // indirect
        github.com/yuin/goldmark v1.4.13 // indirect
        go.mongodb.org/mongo-driver v1.1.2 // indirect
-       go.opentelemetry.io/proto/otlp v0.19.0 // indirect
+       go.opentelemetry.io/proto/otlp v1.0.0 // indirect
        go.starlark.net v0.0.0-20200306205701-8dd3e2ee1dd5 // indirect
        go.uber.org/atomic v1.9.0 // indirect
        golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
@@ -326,7 +326,7 @@ replace (
        github.com/dgrijalva/jwt-go => github.com/form3tech-oss/jwt-go v0.0.0-20210511163231-5b2d2b5f6c34
        github.com/docker/docker => github.com/moby/moby v24.0.4+incompatible
        // Remove once we get to a version with extproc changes in upstream envoy 1.28
-       github.com/envoyproxy/go-control-plane => github.com/solo-io/go-control-plane-fork-v2 v0.0.0-20230902211618-ada8201b381c
+       github.com/envoyproxy/go-control-plane => github.com/solo-io/go-control-plane-fork-v2 v0.0.0-20231026174848-be784d54c527
 
        // Newer versions result in changes to the names of generated metrics
        // remove this when we fix https://github.com/solo-io/solo-kit/issues/516
```

so i think this issue is resolved. thanks for pointing this out, there were a lot of dependency updates that shouldn't have been in there

projects/gloo/pkg/defaults/port.go
```diff
@@ -17,6 +17,8 @@ var GlooValidationPort = 9988
 var GlooMtlsModeRestXdsPort = 9998
 var GlooMtlsModeXdsPort = 9999
 var DefaultRefreshRate = time.Minute
+var TapServerHttpPort uint16 = 9001
+var TapServerGrpcPort uint16 = 9002
```
jenshu, line 21: if these are only used for testing, i don't think we need to expose them here

projects/gloo/api/v1/enterprise/options/tap/tap.proto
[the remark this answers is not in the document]
jenshu: if we're not using any other fields then this is fine for now. it's easier to add new fields later than to remove them

projects/gloo/api/v1/enterprise/options/tap/tap.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+package tap.options.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/tap";
+
+// import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/config/core/v3/grpc_service.proto";
```
jenshu, line 7: can remove if unused

projects/gloo/pkg/defaults/port.go
ashishb-solo: my understanding of how we expose ports is that we want to have them all defined in one place so that we don't have conflicts. is that not the case? i did notice that we don't have ports for other services like redis, extauth, etc. so perhaps i was mistaken....

projects/gloo/api/external/envoy/config/tap/output_sink/v3/grpc_output_sink.proto
sam-heilbron: I see, so this API is not actually user-facing. This is the custom data-plane API. Can you help me understand why we need to define it in this repo and not import the proto from the repo where the API is defined?

projects/gloo/api/v1/options.proto
```diff
@@ -216,6 +217,9 @@ message HttpListenerOptions {
     // Router is an extension of the envoy http filters
     // Maps to https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto
     Router router = 18;
+
+    // Enterprise only: Tap filter settings (experimental). Overrides global tap options configured via settings
```
sam-heilbron, line 221: nit: Overriding configuration can sometimes be confusing to customers. Does override mean we perform a shallow merge and we override each top level field? Or does overriding mean that if any config is defined we respect that entirely?

projects/gloo/api/external/envoy/config/tap/output_sink/v3/http_output_sink.proto
```diff
@@ -0,0 +1,22 @@
...
+
+option java_package = "io.envoyproxy.envoy.config.output_sink.tap.v3";
+option java_outer_classname = "HttpOutputSinkProto";
+option java_multiple_files = true;
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/config/tap/output_sink/v3";
+
+import "validate/validate.proto";
+import "github.com/solo-io/gloo/projects/gloo/api/external/envoy/config/core/v3/http_uri.proto";
+
+// [#protodoc-title: HTTP Output Sink]
```
sam-heilbron, line 13: Similar to the other comment about the grpc output sync, I think it's always been confusing (to me and other devs) why this API is required in our open source control plane api definitions. Could we add a simple readme to the `/api/external/envoy` package that outlines how to structure these APIs and why they're required?

nfuden: Id rather if we updated the existing one but yes it should be discoverable

projects/gloo/api/v1/options.proto
nfuden: Its a great point. I almost lean to the statement that its user error to configure both settings level and gateway level but would suspect that if we dont error out on it then it likely should be a merge?

projects/gloo/pkg/defaults/port.go
nfuden: Yeah I think this should be user defined and we shouldnt be in the business of managing their tap server ports as a variable here.

projects/gloo/pkg/plugins/enterprise_warning/plugin.go
```diff
@@ -236,6 +241,11 @@ func isSanitizeClusterHeaderConfiguredOnListener(in *v1.HttpListener) bool {
 	return in.GetOptions().GetSanitizeClusterHeader() != nil
 }
 
+// tap
+func isTapConfiguredOnListener(in *v1.HttpListener) bool {
+	return in.GetOptions().GetTap() != nil
```
nfuden, line 246: Does this also catch settings level setting of tap?

projects/gloo/api/v1/options.proto
jenshu: currently the Tap settings consist of just one top-level array of size 1, so likely overwrite is the only thing that makes sense. but if we add more top-level fields in the future then we should define what kind of merge (or overwrite) it does

projects/gloo/pkg/defaults/port.go
ashishb-solo: Ah okay. I have no idea why, but I thought this file was only being used for tests.

Removing the ports here in 4c766058446348336573ccb4e7c4976d07872a6f

projects/gloo/pkg/plugins/enterprise_warning/plugin.go
ashishb-solo: no, it does not. but i don't actually see any settings-level checks in that file... i see virtual services, routes, upstreams and http filters (ie. listeners), i don't see settings in there

projects/gloo/api/v1/enterprise/options/tap/tap.proto
ashishb-solo: i thought i had already addressed this. will push a fix for this soon

ashishb-solo: Removed in this commit: https://github.com/solo-io/gloo/pull/8850/commits/aec6d35b40e4140f73a8236582055b2ab32432c3

projects/gloo/api/external/envoy/config/tap/output_sink/v3/http_output_sink.proto
ashishb-solo: It seems that we already have a README for this? https://github.com/solo-io/gloo/tree/main/projects/gloo/api/external/envoy

projects/gloo/pkg/plugins/enterprise_warning/plugin.go
ashishb-solo: since we removed the option from settings, we don't need to do this any longer

projects/gloo/api/v1/options.proto
ashishb-solo: we are holding off on exposing tap via settings for now and only exposing it via the gateway/listener. we will make a decision later as to how we override behaviours from settings to gateway

projects/gloo/api/v1/options.proto
jenshu, line 221: `Overrides global tap options configured via settings` this sentence should be removed now that we removed the global settings

ashishb-solo: nice catch! i removed that here: https://github.com/solo-io/gloo/pull/8850/commits/e10c4febb5d48f2c4e4992cd52ee547eac3ad652

ashishb-solo: we removed tap from global settings for now until we can make a decision on this: https://github.com/solo-io/gloo/pull/8850/commits/c0da9347c2590c00db44294b02a9e78f3cdea934

projects/gloo/api/external/envoy/config/tap/output_sink/v3/http_output_sink.proto
ashishb-solo: and more here: https://github.com/solo-io/gloo/pull/8850/commits/117a1960fe81c8c4734c11aa6ed4df9319c01f6a

nfuden (approved):

jenshu (approved):

nfuden (approved):

PR #8856, opened 2023-11-03
cli: Reject higher majors from this main line update, tbd a new command flag

nfuden:
Avoid allowing glooctl to install v2 variants without an explicit call for it.
      This may be reverted later once we have a better understanding of upgrade path.

Testing by looking the test.sh file and modifying it to prove out the filtering works


sam-heilbron (commented):
lgtm

projects/gloo/cli/pkg/cmd/upgrade/cmd.go
```diff
@@ -138,6 +138,11 @@ func getReleaseWithAsset(ctx context.Context, httpClient *http.Client, tag strin
 			if err != nil {
 				continue
 			}
+
+			if v.Major != 1{
```
sam-heilbron, line 142: nit: Will this fail codegen?

sam-heilbron: nit: A comment here would be really helpful I think, so people know when we can remove this

projects/gloo/cli/test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
```
sam-heilbron, line 1: Is this run in CI?

sam-heilbron (approved):

projects/gloo/cli/test.sh
nfuden: This should be used to validate our changes

sam-heilbron (approved):

projects/gloo/cli/pkg/cmd/upgrade/cmd.go
nfuden: Unsure at this time. We should work with others to understand if we can ever remove it from v1 branches

EItanya (approved):

PR #8859, opened 2023-11-05
API for Azure OIDC distributed claims

tjons:
# Description

This PR introduces some new Azure-specific configuration options for the `OidcAuthorizationCodeConfig` API Oauth2 type. 

## API changes
- Added `Provider` `oneof` field to the `OidcAuthorizationCodeConfig` API
- Added a standard implementation option, `Default`
- Added appropriate Azure-specific configuration to the `Azure` option.

# Context

To support Azure distributed claims, users need to be able to specify a `tenant_id`, a reference to a `client_secret` and a `client_id` to enable the connection to the MS Entera graph API to resolve distributed claims. Also supported is redis caching of claims, which I've also added as an additional `RedisOptions` field on this message.

# Checklist:

- [ * ] I have performed a self-review of my own code
- [ * ] I have commented my code, particularly in hard-to-understand areas
- [ * ] I have made corresponding changes to the documentation


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -785,6 +785,34 @@ message OidcAuthorizationCode {
...
+    oneof Provider {
+        DefaultOIDCProvider default = 26;
+        Azure azure = 27;
+    }
+
+    // No-op, represents default OIDC behavior
+    message DefaultOIDCProvider {}
+
+    // Represents config specific to Azure Distributed Claims OIDC implementation
+    message Azure {
```
Sodman, line 798: If we're using these messages twice in the same proto would it make sense to define the message once at a higher level and use the same message in both places?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -785,6 +785,34 @@ message OidcAuthorizationCode {
...
 
     ClientAuthentication client_authentication = 25;
+
+    oneof Provider {
+        DefaultOIDCProvider default = 26;
+        Azure azure = 27;
+    }
+
+    // No-op, represents default OIDC behavior
+    message DefaultOIDCProvider {}
```
EItanya, line 795: Since this is already nested should we just call it default?

tjons: yes good idea

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
tjons: yes good thinking

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
tjons: actually I'm going to take @Sodman's idea and move these types to the top level (see latest change)... I think they should stay suffixed with `-OidcProvider` to make it clear what they're for. Otherwise, it would be `extauthv1.Default` which is really confusing (what default?)... wdyt?

EItanya: Agreed since they are no longer nested

EItanya:
This API makes sense to me. I think it would be a good idea if @marcogschmidt would take a look at this. He's one of the most familiar with this API.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1684,6 +1712,11 @@ message ExtAuthConfig {
             // A list of claims to be mapped from the JWT token received by ext-auth-service to an upstream destination
             repeated ClaimToHeader claims_to_headers = 1;
         }
+
+        oneof Provider {
+            DefaultOidcProvider default = 26;
+            AzureOidcProvider azure = 27;
```
marcogschmidt, line 1718: We cannot use the same type we have in the `AuthConfigSpec` here. The `ExtAuthConfig` message is what is sent via xDS to ext auth from the control plane in Gloo Edge Enterprise. The assumption is that it contains the _fully resolved_ , self-contained, configuration for the server. This means that e.g. this message cannot contain references to resources, but must contain the information that we need from the referenced objects. 

Just to explain this a little better in case the above is not clear:
- to configure the server we need to do two things:
  - read all the resources from storage that are needed to configure it
  - configure the server state based these resources
- when in **k8s controller mode** (how we use it in GP), the server does both of these things itself
- when in **xDS mode** the server expects the control plane to do the first part, and then just send over the `ExtAuthConfig`, which it uses to configure itself

When deployed in xDS mode, the server has no way of e.g. reading secrets or config maps. That's why if you check the `ExtAuthConfig` you will see that secrets appear as strings, e.g.

```protobuf
string client_secret = 2 [(extproto.sensitive) = true];
```

**TL;DR** we need two different messages for `AzureOidcProvider`:
- one with the secret ref that is what we use on the `AuthConfigSpec`
- one with the resolved secret value that we need for the internal xDS API

marcogschmidt (changes requested):

marcogschmidt (approved):

Sodman (approved):
LGTM!

PR #8868, opened 2023-11-07
Native translation

yuval-k:
Part 1: remove proxy IR in favor of direct translation of listeners, routes.
In future episodes we will remove clusters/endpoints


projects/gateway2/translator/listener/util.go
```diff
@@ -0,0 +1,184 @@
...
+
+func toMatch(info *FilterChainInfo) *listenerv3.FilterChainMatch {
+	if info == nil {
+		return nil
+	}
+	return &listenerv3.FilterChainMatch{
+		ServerNames: info.SslConfig.SniDomains,
+	}
+}
+func toTransportSocket(info *FilterChainInfo) *corev3.TransportSocket {
```
EItanya, line 97: I think these 2 functions should instead be methods on the `FilterChainInfo` struct given that the only arg is a pointer to the object itself.

yuval-k: fixed

lgadban (approved):
LGTM!

PR #8831, opened 2023-10-26
fix: custom readiness probe with custom healthCheck path not working

davidjumani:
# Description

Fixes https://github.com/solo-io/gloo/issues/8288

When following the instructions [here](https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/), a user defines a custom health check on the default gateway (which is deployed in the post-install hook) and defines a readiness probe on the gateway proxy.
During a helm install with `--wait` specified, the gateway proxy must come up and be ready before moving to execute the post-install hooks that deploy the default gateway. As a result, the gateway proxy is never ready since the config to create the health check endpoint is defined in the default gateway. 

This attempts to fix that issue by moving the resource rollout job out of the post install hook. Since the resource rollout is a job, it is one of the last resources to be applied via helm (https://docs.solo.io/gloo-edge/latest/operations/advanced/zero-downtime-gateway-rollout/)
Hence when the jobs runs, it waits for the gloo deployment to complete and then applies the default resources
It also adds a default virtual service not having any VS leads to no listener created (https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/translator/http_translator.go#L28) - in cases of a clean install

There were issues caused when it was first added in https://github.com/solo-io/gloo/pull/8698
These have been resolved and a kube2e test for installation via argo has been added as well, although not as part of CI but for a developer to test

## Helm changes
- Removed the post install / upgrade hooks from the resource rollout job
- Created a resource rollout check job to ensure the resources have been applied
- Created a resource rollout cleanup job to delete the existing resource rollout job prior to upgrades

# Context

https://github.com/solo-io/gloo/issues/8288

### Via Helm
This works on both upgrades && clean installs

values.yaml
```
gateway:
  translateEmptyGateways: true
  validation:
    allowWarnings: false
    alwaysAcceptResources: false
    failurePolicy: Fail
gatewayProxies:
  gatewayProxy:
    service:
      type: ClusterIP    # Since the test is running in kind
    gatewaySettings:
      customHttpGateway:
        options:
          healthCheck:
            path: /envoy-hc
    podTemplate:
      terminationGracePeriodSeconds: 7
      gracefulShutdown:
        enabled: true
        sleepTimeSeconds: 5
      probes: true
      customReadinessProbe:
        httpGet:
          scheme: HTTP
          port: 8080
          path: /envoy-hc
        failureThreshold: 2
        initialDelaySeconds: 5
        periodSeconds: 5
```

Before :
```
[...]

PR #8877, opened 2023-11-08
helm: Refactor resource-rollout-job to deploy extra custom resources

davidjumani:
# Description

This PR allows the deployment of extra CRs as helm partials.
The same [logic](https://github.com/solo-io/gloo/blob/7da4d2fcf8beeeb41822a2572c334dde6284b49e/install/helm/gloo/templates/18-settings.yaml#L172-L174) is used to deploy gloo extra specs used in [enterprise](https://github.com/solo-io/gloo/blob/7da4d2fcf8beeeb41822a2572c334dde6284b49e/install/helm/gloo/templates/18-settings.yaml#L172-L174).
The end goal is to deploy all CRs in the resource rollout job to negate the need for another job to do the same in enterprise.

## Helm changes
- Added `global.extraCustomResources` to add extra custom resources to create, as defined by a helm partial. Defaults to false in open source, and true in enterprise
- Deploying extra CRs in the resource rollout job itself
- Removing the resource rollout cleanup job as this can be accomplished in the rollout check job

# Context

A refactor recommendation to simplify resource rollout mentioned [here](https://github.com/solo-io/gloo/pull/8831#discussion_r1376691065)

## Interesting decisions
 
N/A

## Testing steps

CI passes as well as argocd tests
```
Running Suite: ArgoCD Suite - /Users/djumani/lab/gloo/test/kube2e/argocd
========================================================================
Random Seed: 1699462406 - will randomize all specs

Ran 1 of 1 Specs in 329.556 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

## Notes for reviewers

N/A

# Checklist:


install/helm/gloo/templates/5-resource-configmap.yaml
```diff
@@ -12,3 +12,13 @@ data:
   custom-resources: |
 {{ $cr }}
   has-custom-resources: "{{- $hasCr -}}"
```
jenshu, line 14: as an alternative, we can also stuff all the EE and OSS resources into the same `custom-resources` field, and then the rollout job / rollout check job just needs to read from the one field.

install/helm/gloo/templates/5-resource-rollout-check-job.yaml
```diff
@@ -96,6 +110,18 @@ spec:
               # gateway.gateway.solo.io/default configured
               # gateway.gateway.solo.io/default created
               # This indicates that the resource was not applied in the resource rollout job
+              if [ "$HAS_EXTRA_CUSTOM_RESOURCES" == "true" ]
+              then
+                FAILED_RESOURCES=$(kubectl apply -f /etc/gloo-custom-resources/extra-custom-resources | grep -iv ' unchanged')
```
jenshu, line 115: if we fail to apply some EE CRs, it will only print those and then exit. it would be useful to print both the EE and OSS resources that failed to apply (might be made easier if we combine into one configmap field per previous comment)

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
jenshu, line 1: just to double check, are you sure we can safely delete this job? this one used to run pre-upgrade (so before rollout job ran), but the rollout-check-job runs post-install / post-upgrade which might be too late to clean up the existing rollout job?

install/helm/gloo/templates/5-resource-configmap.yaml
davidjumani: The enterprise upstreams should first be applied successfully since they need to exist before they are referenced in settings (eg. extauth). If all the resources are applied in a single go, the upstream might fail to be created and the setting which refers to that upstream will also fail.
This might be trivial since the job will retry but I'd like to preserve the ordering

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: Yes. Sometimes the chart would fail since the rollout job is present. To be on the safe side, deleting the job after upgrade / install
```
UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: Job "gloo-resource-rollout" in namespace "gloo-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "gloo-ee"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "gloo-system"

```

jenshu: oh gotcha, we are now cleaning it up after _this_ install, instead of right before the _next_ install?

davidjumani: That's right!

install/helm/gloo/templates/5-resource-configmap.yaml
jenshu: agreed that we need to preserve the right ordering. but if we put them all in the same yaml in the correct order, kubectl should apply them in that order, right? i'm not sure though what happens if one of the resources fails to apply - does kubectl keep trying to apply the remaining resources, or does it fail as soon as one resource fails?

davidjumani: Running a quick test, kubectl apply continues to apply resources in the file even some of them are invalid but still returns an exit value of 1

```
apiVersion: gloo.solo.io/v1
kind: Upstream
...
---
kind: Gateway
...
```

```
gateway.gateway.solo.io/abc-proxy created
The Upstream "rateLimit" is invalid: metadata.name: Invalid value: "rateLimit": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

```

jenshu: ok in that case, if we want to fail early, i think it's fine to split custom-resources and extra-custom-resources as you have.
but can we indicate in the error message (see other comment below) something like "there may also be other resources (e.g. gateways) that were not applied" so they know it's not just the upstreams that failed to apply?

davidjumani: Good idea. Added the changes

install/helm/gloo/templates/5-resource-rollout-check-job.yaml
davidjumani: Added https://github.com/solo-io/gloo/pull/8877/commits/dbcb16047876eb77cbbf90c06801e0e4a0e7983f as discussed in the previous comment

elcasteel (approved):

jenshu (approved):

jenshu (approved):

elcasteel (approved):

PR #8879, opened 2023-11-09
fix no match

yuval-k:


projects/gateway2/translator/testutils/inputs/https-routing/gateway.yaml
```diff
@@ -0,0 +1,53 @@
+#$ Used in:
+#$ - site-src/guides/http-routing.md
```
lgadban, line 2: is this the actual example from the gwAPI docs?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -33,6 +33,14 @@ func TranslateGatewayHTTPRouteRules(
 ) []*routeutils.SortableRoute {
 	var outputSortableRoutes []*routeutils.SortableRoute
 	for i, rule := range route.Spec.Rules {
+
+		rule := rule
+		if rule.Matches == nil {
+			// from the spec:
+			// If no matches are specified, the default is a prefix path match on “/”, which has the effect of matching every HTTP request.
```
EItanya, line 40: Breaks `DeepCopy` 😆

projects/gateway2/translator/testutils/inputs/https-routing/gateway.yaml
yuval-k: i forget where i got this from :/ why?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
yuval-k: it doesn't as the rule is copied just before the `if`

EItanya (approved):

projects/gateway2/translator/testutils/inputs/https-routing/gateway.yaml
lgadban: don't remember why I asked either; not important !

PR #8886, opened 2023-11-10
fix xds gc, add logs

yuval-k:


projects/gateway2/xds/xds_syncer.go
```diff
@@ -222,24 +221,31 @@ func (s *XdsSyncer) syncEnvoy(ctx context.Context, listenersAndRoutesForGateway
...
-	ctx = contextutils.WithLogger(ctx, "envoyTranslatorSyncer")
-	logger := contextutils.LoggerFrom(ctx)
+	logger := log.FromContext(ctx, "pkg", "envoyTranslatorSyncer")
 	snapHash := hashutils.MustHash(snap)
-	logger.Infof("begin sync %v (%v proxies, %v upstreams, %v endpoints, %v secrets, %v artifacts, %v auth configs, %v rate limit configs, %v graphql apis)", snapHash,
-		len(snap.Proxies), len(snap.Upstreams), len(snap.Endpoints), len(snap.Secrets), len(snap.Artifacts), len(snap.AuthConfigs), len(snap.Ratelimitconfigs), len(snap.GraphqlApis))
-	defer logger.Infof("end sync %v", snapHash)
+	logger.Info("begin sync", "snapHash", snapHash,
+		"len(proxies)", len(snap.Proxies), "len(upstreams)", len(snap.Upstreams), "len(endpoints)", len(snap.Endpoints), "len(secrets)", len(snap.Secrets), "len(artifacts)", len(snap.Artifacts), "len(authconfigs)", len(snap.AuthConfigs), "len(ratelimits)", len(snap.Ratelimitconfigs), "len(graphqls)", len(snap.GraphqlApis))
+	debugLogger := logger.V(1)
```
EItanya, line 228: Why a new logger vs. `logger.Debug_`

yuval-k: currently the logger that we set it's log level is the controller runtime logger. so i change the usage of this package to that logger. This logger has a different philosophy, where log levels are integers (the higher the level, the more vebose it is). So basically V(1) is the "equivalent" to Debug

EItanya (approved):

PR #8890, opened 2023-11-11
Add Request Mirror Gateway API Filter

EItanya:
This PR adds the remaining Gateway API Filter support. It adds the Mirror filter as well as fixing conformance for URL Rewrite.

This PR also switches to the now maintained Uber fork of `mockgen`, rather than the deprecated google version.


projects/gateway2/query/utils.go
```diff
@@ -0,0 +1,70 @@
...
+		switch cli := cli.(type) {
+		case *corev1.Service:
+			if port == 0 {
+				reporter.SetCondition(reports.HTTPRouteCondition{
+					Type:   gwv1.RouteConditionResolvedRefs,
+					Status: metav1.ConditionFalse,
+					Reason: gwv1.RouteReasonUnsupportedValue,
+				})
+			} else {
+				name := kubernetes.UpstreamName(cli.Namespace, cli.Name, int32(port))
```
yuval-k, line 58: if my PR merges before yours, you will need to update this

projects/gateway2/translator/httproute/filterplugins/mirror/mirror_plugin.go
```diff
@@ -0,0 +1,49 @@
...
+
+	cli, err := ctx.Queries.GetBackendForRef(ctx.Ctx, ctx.Queries.ObjToFrom(ctx.Route), &config.BackendRef)
+	clusterName := query.ResolveBackendRef(
+		cli,
+		err,
+		ctx.Reporter,
+		config.BackendRef,
+	)
+	if clusterName == nil {
+		return nil
```
yuval-k, line 41: maybe return `err`? or is it ok because we don't want to drop the route in this case?

projects/gateway2/query/utils.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/solo-io/gloo/projects/gateway2/reports"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/kubernetes"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+func ResolveBackendRef(cli client.Object, err error, reporter reports.ParentRefReporter, backendRef gwv1.BackendObjectReference) *string {
```
yuval-k, line 15: The backend ref is already resolved here, right? maybe a better name is ProcessBackendRef?

projects/gateway2/translator/httproute/filterplugins/mirror/mirror_plugin_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/solo-io/gloo/projects/gateway2/translator/httproute/filterplugins"
+	"github.com/solo-io/gloo/projects/gateway2/translator/httproute/filterplugins/mirror"
+	"github.com/solo-io/gloo/projects/gateway2/translator/httproute/filterplugins/mirror/mocks"
+	"go.uber.org/mock/gomock"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+//go:generate mockgen -destination mocks/mock_queries.go -package mocks github.com/solo-io/gloo/projects/gateway2/query GatewayQueries
```
yuval-k, line 18: instead of mocking, we have been using queries with a fake client. This has the advantage that if an index is ever introduced, it will still work (as the fake client implements it).

projects/gateway2/translator/httproute/filterplugins/mirror/mirror_plugin.go
EItanya: Exactly, I didn't want to drop the route because of a misconfigured filter, seemed like a weird reason to drop traffic

projects/gateway2/translator/httproute/filterplugins/mirror/mirror_plugin_test.go
EItanya: Spoke offline, not relevant here

projects/gateway2/translator/httproute/filterplugins/mirror/mirror_plugin.go
lgadban: is the behavior for a misconfigured filter not defined per spec?

projects/gateway2/query/utils.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/solo-io/gloo/projects/gateway2/reports"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/kubernetes"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+func ProcessBackendRef(cli client.Object, err error, reporter reports.ParentRefReporter, backendRef gwv1.BackendObjectReference) *string {
```
lgadban, line 15: 2 nits:
* add godoc for function? most important to explain the return being either the correct cluster name or nil in case of error (and that reporter will be used to report whatever the error is encountered)
* `cli` is confusing name for the first parameter (especially since we use it for the k8s client elsewhere in the codebase); maybe it can be `obj` or something more descriptive?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -240,58 +251,10 @@ func translateRouteAction(
...
-					clusterName = kubernetes.UpstreamName(cli.Namespace, cli.Name, int32(port))
-				}
-			default:
-				reporter.SetCondition(reports.HTTPRouteCondition{
-					Type:   gwv1.RouteConditionResolvedRefs,
-					Status: metav1.ConditionFalse,
-					Reason: gwv1.RouteReasonInvalidKind,
-				})
-			}
+		cli, err := queries.GetBackendForRef(context.TODO(), queries.ObjToFrom(gwroute), &backendRef.BackendObjectReference)
```
lgadban, line 254: ditto on the `cli` name here

lgadban (approved):

Issue #8430, opened 2023-06-29
Ability to specify which gateway to sidecar the extauth in multi-gateway deployment


pszeto:
### Version

1.14.x (latest stable)

### Is your feature request related to a problem? Please describe.

Currently you are able to sidecar the extauth into the gateway proxy with the `global.extensions.extAuth.envoySidecar` option.  For a multi gateway deployment https://docs.solo.io/gloo-edge/latest/installation/advanced_configuration/multi-gw-deployment/ it will side car all gateways with the extauth container.

### Describe the solution you'd like

It would be nice to have the ability to specify which gateway to sidecar the extauth container in multi-gateway deployment.

davidjumani:
Added in https://github.com/solo-io/solo-projects/pull/5503 and will be part of v1.16.0-beta3 when released

PR #8898, opened 2023-11-13
feat: Allow users to disable extauth sidecar on a per proxy basis

davidjumani:
# Description

When multiple gateway proxies are defined as well as extauth enabled, a user might want to enable the sidecar for a subset of gateway proxies (since some of them do not require extauth) to reduce resource usage and unnecessary network calls to an unused sidecar.
This feature adds the new helm value `gatewayproxy.proxyName.disableExtauthSidecar` to disable the extauth sidecar on it
It also adds additional helper methods to merge resources while omitting keys in those resources which should not be overwritten if empty

## Helm changes
- Added a new `gatewayproxy.proxyName.disableExtauthSidecar` value

## Code changes
N/A

## CI changes
N/A

## Docs changes
N/A

# Context

https://github.com/solo-io/gloo/issues/8430

## Interesting decisions
 
Had initially thought of the idea to omit fields in an overwrite based on keys as in ddc6774fe70399bdbbd37a870b3f777d9a63f9ab but decided against it as future changes would require modifying those keys if the same field is modified in different locations. Eg: if one part of the code did not want the field modified it would add the key to the list of omitKeys but another part of the code when changing the field would need to remove it from the list

## Testing steps

Unit tests in enterprise to validate the behaviour of the new helm fields and helpers

## Notes for reviewers

N/A

N/A

# Checklist:

### DIFF
--- a/install/helm/gloo/generate/values.go
+++ b/install/helm/gloo/generate/values.go
@@ -505,7 +505,7 @@ type GatewayProxy struct {
 	XdsServicePort                 *uint32                          `json:"xdsServicePort,omitempty" desc:"The k8s service port for the xds server. Defaults to the value from .Values.gloo.deployment.xdsPort, but can be overridden to use, for example, xds-relay."`
 	TcpKeepaliveTimeSeconds        *uint32                          `json:"tcpKeepaliveTimeSeconds,omitempty" desc:"The amount of time in seconds for connections to be idle before sending keep-alive probes. Defaults to 60. See here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-tcpkeepalive"`
 	DisableCoreDumps               *bool                            `json:"disableCoreDumps,omitempty" desc:"If set to true, Envoy will not generate core dumps in the event of a crash. Defaults to false"`
-	DisableExtauthSidecar          *bool                            `json:"disableExtauthSidecar,omitempty" desc:"When extAuth.envoySidecar is set to true, every gateway proxy pod will come up with an extauth container. If this setting is set to true, the extauth container will not be added to the gatway proxy pod. Defaults to false"`
+	DisableExtauthSidecar          *bool                            `json:"disableExtauthSidecar,omitempty" desc:"If set to true, this gateway proxy will not come up with an extauth sidecar container when global.extAuth.envoySidecar is enabled. This setting has no effect otherwise. Defaults to false"`
 	*KubeResourceOverride
 }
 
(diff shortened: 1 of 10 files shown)


install/helm/gloo/generate/values.go
```diff
@@ -505,6 +505,7 @@ type GatewayProxy struct {
 	XdsServicePort                 *uint32                          `json:"xdsServicePort,omitempty" desc:"The k8s service port for the xds server. Defaults to the value from .Values.gloo.deployment.xdsPort, but can be overridden to use, for example, xds-relay."`
 	TcpKeepaliveTimeSeconds        *uint32                          `json:"tcpKeepaliveTimeSeconds,omitempty" desc:"The amount of time in seconds for connections to be idle before sending keep-alive probes. Defaults to 60. See here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-tcpkeepalive"`
 	DisableCoreDumps               *bool                            `json:"disableCoreDumps,omitempty" desc:"If set to true, Envoy will not generate core dumps in the event of a crash. Defaults to false"`
+	DisableExtauthSidecar          *bool                            `json:"disableExtauthSidecar,omitempty" desc:"When extAuth.envoySidecar is set to true, every gateway proxy pod will come up with an extauth container. If this setting is set to true, the extauth container will not be added to the gatway proxy pod. Defaults to false"`
```
sheidkamp, line 508: `When extAuth.envoySidecar is set to true` ... wrong field name?

sheidkamp (changes requested):
One nit, but LGTM

sheidkamp (approved):

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -1,5 +1,14 @@
 {{/* vim: set filetype=mustache: */}}
 
+{{- /*
+There can be cases when we do not want to overwrite an empty value on a resource when merged.
+Eg. To generate a proxy config, we mergeOverwrite it with the default gateway-proxy config.
+If we want to preserve the empty value of the gateway and not have them overwritten, we set it to `gloo.omitOverwrite`
+and call `gloo.util.mergeOverwriteWithOmit` when merging. This sets all fields with values equal to this back to empty after the overwrite
+*/ -}}
+{{- define "gloo.omitOverwrite" }}
+{{ printf "\n" }}{{/* This template is set to a new line. There may be scenarios where a field is initailly set to this value and the same field is appended to later on. Since this is just a new line, it won't cause rendering issues */}}
```
jenshu, line 10: > There may be scenarios where a field is initailly set to this value and the same field is appended to later on. 

just curious, in which cases would the field get modified or appended to multiple times?
did the hardcoded magic string not work in that scenario?

davidjumani: As of now, certain fields are set based on flags. Eg: [extauth sidecar](https://github.com/solo-io/solo-projects/blob/main/install/helm/gloo-ee/templates/_helpers.tpl#L12)
If in the future another feature is added and requires a helper container, it would append to the `extraContainersHelper` field. If extauth sidecar is disabled, it will append to a new line and if not, to the extauth sidecar container. So rendering the `extraContainersHelper` field will not cause any issues

jenshu (approved):

PR #8903, opened 2023-11-14
remove stuff

yuval-k:


Makefile
```diff
@@ -249,10 +248,6 @@ generated-code: fmt
 go-generate-all: clean-vendor-any ## Run all go generate directives in the repo, including codegen for protos, mockgen, and more
 	GO111MODULE=on go generate ./...
 
-.PHONY: go-generate-apis
```
lgadban, line 252: general question for all deletions in this `Makefile`;
how did you determine what to delete and what to keep?

yuval-k: i ran make generated and deleted everything that was erroring :)

lgadban: So ideally we will clean it up further right? I.e. all of the build functionality currently needed by gateway2 is still in `projects/gateway2/Makefile`

lgadban (approved):

PR #8907, opened 2023-11-15
make go test ./... work again

yuval-k:


projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -61,7 +59,7 @@ func (k *kube) Get(ctx context.Context) (*ServerVersionInfo, error) {
 
 	deployments, err := client.AppsV1().Deployments(k.namespace).List(ctx, metav1.ListOptions{
 		// search only for gloo deployments based on labels
-		LabelSelector: "app=gloo",
+		LabelSelector: "gloo=gateway-v2",
```
lgadban, line 62: super nit: if we're going this far to make it constant in helm, should we also make it a named constant in the code as well (preferably with a comment linking to the `const_labels` helm helper)?

projects/gloo/cli/pkg/cmd/version/version_test.go
```diff
@@ -192,24 +143,26 @@ Server: {"type":"Gateway","enterprise":true,"kubernetes":{"containers":[{"Tag":"
 							Output: test.outputType,
 						},
 					}
-					sv.Enterprise = test.enterprise
-					client.EXPECT().Get(nil).Times(1).Return([]*version.ServerVersion{sv}, nil)
-					err := printVersion(client, buf, opts)
+					//					sv.Enterprise = test.enterprise
```
lgadban, line 146: ultra nit: keeping this around for potential enterprise versioning? also why the huge amount whitespace?

yuval-k: i prefer to add this back when we'll have the enterprise, as i have no idea what it will look like

projects/gloo/cli/pkg/cmd/version/clients.go
yuval-k: makes sense!

lgadban (approved):

EItanya (approved):

PR #8908, opened 2023-11-15
restructure

yuval-k:
- move files
- fix imports


main.go
EItanya, line 1: Should this go inside of `cmd`? That's what we typically do

EItanya (approved):

EItanya (approved):

PR #8912, opened 2023-11-16
refactor route status

lgadban:
* update Route report & reporter to use variant types
* use Getter for route report
* reorder reporter.go for readability
* extract & unit-test building of route status


pkg/reports/status.go
```diff
@@ -59,6 +60,47 @@ func (r *ReportMap) BuildGWStatus(ctx context.Context, gw gwv1.Gateway) gwv1.Gat
...
+	routeReport := r.route(&route)
+	routeStatus := gwv1.RouteStatus{}
+	for _, parentRef := range route.Spec.ParentRefs {
+		parentStatusReport := routeReport.parentRef(&parentRef)
+		addMissingParentRefConditions(parentStatusReport)
+
+		// get status of current parentRef status if it exists
+		var currentParentRefConditions []metav1.Condition
+		currentParentRefIdx := slices.IndexFunc(route.Status.Parents, func(s gwv1.RouteParentStatus) bool {
+			return reflect.DeepEqual(s.ParentRef, parentRef)
```
lgadban, line 73: thoughts? as the ParentRef structs are very small, figured this would be easier than implementing a custom comparison

yuval-k: i'm ok with this for now; in the past we saw `reflect.DeepEqual` in the hot path of cpu profiles. if it comes up again we can change to a custom comparison

lgadban: sounds good, my main reasoning is these are "proper" Go structs and not e.g. proto generated structs, so their footprint (even when reflected) seems minimal

lgadban: just for posterity: https://github.com/kubernetes-sigs/gateway-api/blob/95a6b8a1f7b27bc0e632e4f44fc3e277a80276bc/apis/v1/shared_types.go#L35-L155

6 fields, ptrs to strings, string, and an int

yuval-k (approved):

PR #8919, opened 2023-11-17
loadbalancer: Slowstart configuration fixes

nfuden:
# Description

Slow start was not correctly implemented.
It was not tested fully and the runtime override flag was not set as expected.

_

## Code changes
Adds configuration for clustername.slowstart.agression in a way that mimicks upstream tests
Adds fallback in case clustername is not available to have a stable 'hopefully' unique runtime key. It shouldnt be fully random as that would cause uneeded churn. 

## Testing steps

This should be tested but as of the draft pr is not


projects/gloo/pkg/plugins/loadbalancer/plugin.go
```diff
@@ -167,18 +169,34 @@ func configureLeastRequestLb(out *envoy_config_cluster_v3.Cluster, cfg *v1.LoadB
...
-func toSlowStartConfig(cfg *v1.LoadBalancerConfig_SlowStartConfig) *envoy_config_cluster_v3.Cluster_SlowStartConfig {
+func toSlowStartConfig(clusterInfo *envoy_config_cluster_v3.Cluster, cfg *v1.LoadBalancerConfig_SlowStartConfig) *envoy_config_cluster_v3.Cluster_SlowStartConfig {
 	if cfg == nil {
 		return nil
 	}
 	out := envoy_config_cluster_v3.Cluster_SlowStartConfig{
 		SlowStartWindow: cfg.GetSlowStartWindow(),
 	}
+
+	proposedRuntimeKey := clusterInfo.GetName()
```
jbohanon, line 180: Do we want to offer the user the option to set their own runtime key or is that just a bit footgun for copypasta config?

nfuden: I think we can be opinionated on this but I also dont want to do more work on this until we truly can pick this up. 

This was put up as a poc more than anything

PR #8922, opened 2023-11-20
Renew Hashicorp Vault Tokens

sheidkamp:
# Description

Implement token renewal as demonstrated here: https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go

## API changes
Added `settings.secretOptions.sources[].vault.aws.watcherIncrement` which gets used as the `Increment` argument when creating the `LifeTimewatcher` to monitor the token. If the value is 0 or not defined, the argument will not be passed.

## Code changes
Added renewal logic to `projects/gloo/pkg/bootstrap/clients/vault.go`

## Docs changes
Will be coming along

## Interesting decisions

-  Put the renewal logic in the `vault.go` bootstrap client in `gloo`  instead of in `solo-kit` because we are not currently handling any login/token/session logic in `solo-kit`
- Put the new `watcherIncrement` API/Helm field in the `VaultAwsAuth` section because it is part of the Vault/AWS connection

## Testing steps
Follow the steps in https://docs.solo.io/gloo-edge/main/guides/integrations/vault/aws-irsa-auth/ to see the 403 errors.

Repeat with the code in this branch and with the watcher increment set in settings (this shows only the new value to add, not the complete YAML block):
```
  secretOptions:
    sources:
    - vault:
        aws:
          watcherIncrement: 300
```

and the errors will not occur

## Notes for reviewers

Be sure to verify intended behavior by actually running it!

# Checklist:


projects/gloo/api/v1/settings.proto
```diff
@@ -181,6 +181,8 @@ message Settings {
         string secret_access_key = 6;
         // The Session Token as provided by the security credentials on the AWS IAM resource
         string session_token = 7;
+        // Value in seconds to pass the Vault LifetimeWatcher constructor as the Increment parameter. If undefined or 0, the argument will not be passed.
```
jbohanon, line 184: Should we put a more user-facing description here? i.e. what the Increment parameter is configuring?

jbohanon (commented):
I won't get a chance to run it until tomorrow morning but so far looks good! We should also remove the TODO I had put in `vault.go`

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -49,20 +39,20 @@ func NewVaultSecretClientFactory(clientInit VaultClientInitFunc, pathPrefix, roo
 	}
 }
 
-func VaultClientForSettings(vaultSettings *v1.Settings_VaultSecrets) (*api.Client, error) {
+func VaultClientForSettings(vaultSettings *v1.Settings_VaultSecrets) (*vault.Client, error) {
```
sam-heilbron, line 42: Since we rely on more logging functionality, are are also now spinning off go-routines, I think it might be valuable to pass the context down into this function so it can be used by its dependents.

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+			contextutils.LoggerFrom(context.Background()).Fatalf("unable to start managing token lifecycle: %v", tokenErr)
+		}
+	}
+}
+
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(client *vault.Client, token *vault.Secret, watcherIncrement int) error {
+	renew := token.Auth.Renewable // You may notice a different top-level field called Renewable. That one is used for dynamic secrets renewal, not token renewal.
```
sam-heilbron, line 236: A vault secret also has a `LeaseDuration` property. What is that intended for? Any reason we can't use that to determine that value to pass to the LifetimeWatcher.increment?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
 		return nil, errors.New("no auth info was returned after login")
 	}
 
+	go renewToken(client, awsAuth, int(aws.GetWatcherIncrement()))
```
sam-heilbron, line 212: This goroutine will span additional waches/goroutines. Our setup model works such that, when the Settings object is modified, we re-run our setup code (which includes this). Let's be sure that re-running that will not cause goroutines to leak/grow. 

We could explore https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/consul/consul_suite_test.go#L14 and see if we can perform a similar check that we don't have goroutine leaks

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+	}
+}
+
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(client *vault.Client, token *vault.Secret, watcherIncrement int) error {
+	renew := token.Auth.Renewable // You may notice a different top-level field called Renewable. That one is used for dynamic secrets renewal, not token renewal.
+	if !renew {
+		contextutils.LoggerFrom(context.Background()).Infof("Token is not configured to be renewable. Re-attempting login.")
```
sam-heilbron, line 238: As a user, how should I interpret this log message and what should I do about it? Does this belong more at the debug leve?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		// duration is under a built-in threshold and either renewing is not
+		// extending it or renewing is disabled. In any case, the caller
+		// needs to attempt to log in again.
+		case err := <-watcher.DoneCh():
+			if err != nil {
+				contextutils.LoggerFrom(context.Background()).Infof("Failed to renew token: %v. Re-attempting login.", err)
+				return nil
+			}
+			// This occurs once the token has reached max TTL.
+			contextutils.LoggerFrom(context.Background()).Infof("Token can no longer be renewed. Re-attempting login.")
```
sam-heilbron, line 270: What is the impact on the user when this occurs?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+
+	watcher, err := client.NewLifetimeWatcher(lifetimeWatcherInput)
+	if err != nil {
+		return fmt.Errorf("unable to initialize new lifetime watcher for renewing auth token: %w", err)
+	}
+
+	go watcher.Start()
+	defer watcher.Stop()
+
+	for {
```
sam-heilbron, line 258: Given the asynchronous nature of this feature, how will users monitor that this is behaving as expected? Should we include a metric to signal the last updated time? Or the number of successful updates vs failed updates?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
 	return client, nil
 }
+
+// Once you've set the token for your Vault client, you will need to periodically renew its lease.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func renewToken(client *vault.Client, awsAuth *awsauth.AWSAuth, watcherIncrement int) {
+	for {
+		vaultLoginResp, err := client.Auth().Login(context.Background(), awsAuth)
+		if err != nil {
+			contextutils.LoggerFrom(context.Background()).Fatalf("unable to authenticate to Vault: %v", err)
```
sam-heilbron, line 223: Why do we return a fatal error in this instance?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+	if !renew {
+		contextutils.LoggerFrom(context.Background()).Infof("Token is not configured to be renewable. Re-attempting login.")
+		return nil
+	}
+
+	lifetimeWatcherInput := &vault.LifetimeWatcherInput{
+		Secret: token,
+	}
+
+	if watcherIncrement > 0 {
```
sam-heilbron, line 246: If the increment value is 0, we are intending to not watch the secret, is that correct? In that case, should we just return early here so avoid creating a NewLIfetimeWatcher that does nothing?

sam-heilbron (commented):
A good start! A few questions around the implementation.
Also, https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/vault.go has a lot of deprecated code that is not used anywhere, should we just remove it?

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: I will look into what happens in this case, and if there's a use case for not having a watcher (lifetime of the pod < ttl?) or if there's a better default.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: This is what Hashicorp did, and we should probably not just accept their demo code and consider how we want this to work.

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +209,70 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		}
+	}
+}
+
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(client *vault.Client, token *vault.Secret, watcherIncrement int) error {
+	renew := token.Auth.Renewable // You may notice a different top-level field called Renewable. That one is used for dynamic secrets renewal, not token renewal.
+	if !renew {
```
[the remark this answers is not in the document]
sheidkamp: 👍

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: I think this would be for some deeper debugging, and it would be appropriate to knock this down a level.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: I will look into that.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: 👍

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: 👍

projects/gloo/api/v1/settings.proto
```diff
@@ -181,6 +181,8 @@ message Settings {
...
+        int32 watcher_increment = 8;
```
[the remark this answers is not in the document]
sheidkamp: yes

helm-overrides.yaml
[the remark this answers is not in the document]
sheidkamp: Yup, local file. Removed.

projects/gloo/api/v1/settings.proto
sheidkamp: Probably... I will update after taking a closer look at the docs and appropriate defaults.

projects/gloo/api/v1/settings.proto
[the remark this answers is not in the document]
sheidkamp: yes

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: 👍

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: `LeaseDuration` looks like it's involved in leader elections and is not related to token leases.
```
	// LeaseDuration is the duration that non-leader candidates will
	// wait to force acquire leadership. This is measured against time of
	// last observed ack.
```

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -195,27 +207,26 @@ Note that you must adjust both the `pathPrefix` and `rootKey` options when you u
...
-			address: http://vault-internal.vault:8200
-			aws:
-				iamServerIdHeader: ${IAM_SERVER_ID_HEADER_VALUE}
-				mountPath: aws
-				region: ${AWS_REGION}
-			# assumes kv store is mounted on 'dev'
-			pathPrefix: dev
-			# root key by default is 'gloo'.
-			rootKey: gloo
+  kubeResourceOverride:
```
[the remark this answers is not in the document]
sheidkamp: The yaml that was there represents how the data ends up looking in the settings object in kubernetes, but we needed to update this to what helm expects. This change is not due to changed functionality, it is due to the existing YAML not working as intended.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: I think I understand a little better how this is used. The increment is used here is the LifetimeWatcher code:
https://github.com/hashicorp/vault/blob/main/api/lifetime_watcher.go#L382

If it is 0, if it is 0 the effective value is the lease duration. It is used to calculate the "grace period", which is used to jitter the sleep period, so the value only has a small effect on behavior. I'm not sure we even need it, but also we don't need special code if its 0.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: Added

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: Added similar to the bootstrap client test suite.

install/helm/gloo/generate/values.go
```diff
@@ -253,13 +253,14 @@ type VaultTlsConfig struct {
...
+	SecretAccessKey   string  `json:"secretAccessKey,omitempty" desc:"Optional. The Secret Access Key as provided by the security credentials on the AWS IAM resource. In cases such as receiving temporary credentials through assumed roles with AWS Security Token Service (STS) or IAM Roles for Service Accounts (IRSA), this field can be omitted. https://developer.hashicorp.com/vault/docs/auth/aws#iam-authentication-inferences."`
+	SessionToken      string  `json:"sessionToken,omitempty" desc:"The Session Token as provided by the security credentials on the AWS IAM resource."`
+	WatcherIncrement  *uint32 `json:"watcherIncrement,omitempty" desc:"The number of seconds passed as the WatcherInterval argument to the vault LifetimeWatcher. Argument is not passed of value is 0, See https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go. Defaults to 0."`
```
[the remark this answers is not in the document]
sheidkamp: updated

projects/gloo/api/v1/settings.proto
sheidkamp: updated

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		}
+	}
+}
+
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// based on https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(ctx context.Context, client *vault.Client, token *vault.Secret, watcherIncrement int) error {
+	if !token.Auth.Renewable {
+		contextutils.LoggerFrom(ctx).Debugf("Token is not configured to be renewable. Re-attempting login.")
```
sam-heilbron, line 237: If a token is not renewable, will it ever be? Why would logging back in change the behavior? Won't we find ourselves in an endless loop of: login > token not renewable > login > token not renewable ..etc

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+				contextutils.LoggerFrom(ctx).Debugf("Failed to renew token: %v. Re-attempting login.", err)
+				return nil
+			}
+			// This occurs once the token has reached max TTL.
+			contextutils.LoggerFrom(ctx).Debugf("Token can no longer be renewed. Re-attempting login.")
+			return nil
+
+		// Successfully completed renewal
+		case renewal := <-watcher.RenewCh():
+			contextutils.LoggerFrom(ctx).Debugf("Successfully renewed: %#v.", renewal)
```
sam-heilbron, line 269: Is there a reason you use `%#v` and `%v` in the log lines in this block? (ie any reason we don't just use one format) https://pkg.go.dev/fmt

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,8 +194,7 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
 		return nil, err
 	}
 
-	// TODO(jbohanon) set up auth token refreshing with client.NewLifetimeWatcher()
-	authInfo, err := client.Auth().Login(context.Background(), awsAuth)
+	authInfo, err := client.Auth().Login(ctx, awsAuth)
```
sam-heilbron, line 197: Should we wrap this call in a simple retry with backoff mechanism? That way if there are transient failures we are protected.

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+	go renewToken(ctx, client, awsAuth, int(aws.GetLeaseIncrement()))
+
 	return client, nil
 }
+
+// Once you've set the token for your Vault client, you will need to periodically renew its lease.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func renewToken(ctx context.Context, client *vault.Client, awsAuth *awsauth.AWSAuth, watcherIncrement int) {
+	for {
+		vaultLoginResp, err := client.Auth().Login(ctx, awsAuth)
```
sam-heilbron, line 221: In the above call to `client.Auth().Login` we check if the return authInfo is nil and if so, return an error. Is that required here as well?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+			contextutils.LoggerFrom(ctx).Fatalf("unable to start managing token lifecycle: %v", tokenErr)
+		}
+	}
+}
+
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// based on https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(ctx context.Context, client *vault.Client, token *vault.Secret, watcherIncrement int) error {
+	if !token.Auth.Renewable {
```
sam-heilbron, line 236: Let's use the TokenIsRenewable method defined on the object: https://github.com/hashicorp/vault/blob/main/api/secret.go#L235

projects/gloo/pkg/bootstrap/vault.go
```diff
@@ -1,6 +1,8 @@
 package bootstrap
 
 import (
+	"context"
```
sam-heilbron, line 4: I think we can just delete this whole file since it's unused, what do you think?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		contextutils.LoggerFrom(ctx).Debugf("Token is not configured to be renewable. Re-attempting login.")
+		return nil
+	}
+
+	watcher, err := client.NewLifetimeWatcher(&vault.LifetimeWatcherInput{
```
sam-heilbron, line 241: nit: There are some other inputs in the LifetimeWatcherInput and we are inheriting the defaults. Personally I like explicitly defining these values for 2 reasons:
1. It's clear to other devs what the behavior is, and what toggles exist
2. We are protected from library changes that impact those default values

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		Secret:    token,
+		Increment: watcherIncrement,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to initialize new lifetime watcher for renewing auth token: %w", err)
```
sam-heilbron, line 246: Looking at the constructor, we will only receive an error here if:
- The input is nil
- The input.secret is nil

I don't think we should ever hit this case, though I'm also a bit nervous about this error leading to a fatal error

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -219,5 +208,65 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
 	return client, nil
 }
+
+// Once you've set the token for your Vault client, you will need to periodically renew its lease.
+// taken from https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func renewToken(ctx context.Context, client *vault.Client, awsAuth *awsauth.AWSAuth, watcherIncrement int) {
+	for {
+		vaultLoginResp, err := client.Auth().Login(ctx, awsAuth)
+		if err != nil {
+			contextutils.LoggerFrom(ctx).Fatalf("unable to authenticate to Vault: %v", err)
```
sam-heilbron, line 223: At this point in the code, we are guaranteed to have 1 successful login of the user. As a resut, I'm not sure I follow why we opt to fatal on errors, instead of continuining to retry. What are the conditions that a fatal would lead to a subsequent success that a retry wouldn't?

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: We ran into something similar in extauth with oidc I think. The approach I prefer is to keep the code "formal" and properly collect and react to the theoretical error, and also including a comment explaining why it should never happen.

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: and at the request of the FE, gonna add metrics.

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
-		// if using inferred credentials, add error information regarding setting credentials
-		if len(possibleErrStrings) > 0 {
-			err = errors.Wrapf(err, "using implicit credentials, consider setting aws secret access key and access key id")
+		if ctx.Err() != nil {
+			return nil, errors.Wrap(ctx.Err(), "unable to login to AWS auth method")
+		} else {
+			// This should never happen because we are retrying on all other errors
+			return nil, errors.Wrap(err, "unable to login to AWS auth method")
+
 		}
```
sam-heilbron, line 234: Can you help me understand what this guard is intended to handle? This allow us to include the context error if one exists, but wouldn't that only occur if the context is cancelled, which would in turn cancel all the operations? 

If the context is relevant, what if we baked this ctx.Err check into the loginWithRetry function? That way the signature of this function is clear: if an error was returned, the operation wasn't completed and the error contains all relevant context

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+// Starts token lifecycle management. Returns only fatal errors as errors,
+// otherwise returns nil so we can attempt login again.
+// based on https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go
+func manageTokenLifecycle(ctx context.Context, client *vault.Client, secret *vault.Secret, watcherIncrement int) (bool, error) {
+	if renewable, err := secret.TokenIsRenewable(); !renewable || err != nil {
+		if err != nil {
+			contextutils.LoggerFrom(ctx).Debugf("Error in checking if token is renewable: %v. Re-attempting login.", err)
+		} else {
+			contextutils.LoggerFrom(ctx).Debugf("Token is not configured to be renewable. Re-attempting login.")
+		}
```
sam-heilbron, line 330: nit: What if we consolidated these two into a single message which includes both the value of the error and value of renewable?

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+		return true, nil
```
sam-heilbron, line 331: Am I understanding this correctly: If a token is not renewable, we will return true from `manageTokenLifecycle` , this will cause us to login again and repeat the whole process. Are all tokens supposed to be renewable? If one is not, this seems like it will cause an infinite loop

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+
+// The possibleErrors parameter is included for error messaging on the initial login attempts
+func loginWithRetry(ctx context.Context, client *vault.Client, awsAuth *awsauth.AWSAuth, possibleErrors bool) (*vault.Secret, error) {
+	// var count = 0
+	var vaultLoginResp *vault.Secret
+	err := contextutils.NewExponentialBackoff(contextutils.ExponentialBackoff{}).Backoff(ctx, func(ctx context.Context) error {
+		var vaultErr error
+		vaultLoginResp, vaultErr = client.Auth().Login(ctx, awsAuth)
+		if vaultErr != nil {
+			if possibleErrors {
```
sam-heilbron, line 256: This makes an assumption and limitation that possibleErrors will always be due to the implicit credentials. What if instead of passing in a boolean, we passed in an error? Then on login error, we append the provided one and log that. This will alllow us to pass in nil when there are no additional errors, or a custom error about using implicit credentials

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+var (
+	errVaultAuthentication = errors.New("unable to authenticate to Vault")
+	errNoAuthInfo          = errors.New("no auth info was returned after login")
+)
+
+// The possibleErrors parameter is included for error messaging on the initial login attempts
+func loginWithRetry(ctx context.Context, client *vault.Client, awsAuth *awsauth.AWSAuth, possibleErrors bool) (*vault.Secret, error) {
+	// var count = 0
+	var vaultLoginResp *vault.Secret
+	err := contextutils.NewExponentialBackoff(contextutils.ExponentialBackoff{}).Backoff(ctx, func(ctx context.Context) error {
```
sam-heilbron, line 252: Elsewhere in our codebase, we've used https://github.com/avast/retry-go. I haven't explored the differences between the two implementations, but wanted to make you aware of it

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+	for {
+		vaultLoginResp, err := loginWithRetry(ctx, client, awsAuth, false)
+		// The only errors we should ever hit are context errors because we are retrying on all other errors
+		if err != nil {
+			if ctx.Err() != nil {
+				contextutils.LoggerFrom(ctx).Errorf("renew token context error: %v", ctx.Err())
+				return
+			} else {
+				// This should never happen because we are retrying on all other errors
+				contextutils.LoggerFrom(ctx).Fatalf("unable to authenticate to Vault: %v", err)
```
sam-heilbron, line 301: Why do we choose to Fatal here, where the other case returns an error? At this point, in the renewtoken function, aren't we guaranteed that a login was successful once and thus we should be successful with enough retries on subsequent times? I would have thought we could just Errorf here, and then the operation would be retried

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -204,20 +220,161 @@ func configureAwsIamAuth(aws *v1.Settings_VaultAwsAuth, client *api.Client) (*ap
...
+			}
+		}
+
+		retry, tokenErr := manageTokenLifecycle(ctx, client, vaultLoginResp, watcherIncrement)
+
+		// The only error this function can return is if the vaultLoginResp is nil, and we have checked against that
```
sam-heilbron, line 307: This comment makes a reference to use checking if thevaultLoginResp is nil, but I don't see that code in this function, or inside of manageTokenLifecycle, can you point me towards it?

sheidkamp (commented):
updated based on PR

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -341,7 +341,7 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 
 	getVaultInit := func(vaultSettings *v1.Settings_VaultSecrets) bootstrap_clients.VaultClientInitFunc {
 		return func() *vaultapi.Client {
-			c, err := bootstrap_clients.VaultClientForSettings(vaultSettings)
+			c, err := bootstrap_clients.VaultClientForSettings(ctx, vaultSettings)
```
nfuden, line 344: is this the right context? shouldnt this be the context of whatever  invokes it rather than the top level syncer context?

nfuden: It looks like we will setup a new instance and have this old uncancelled context each run. Which given we spin off sub go routines will lead to a go routine leak and contention

sheidkamp:
Superceeded

PR #8945, opened 2023-12-04
Revert to proxy ir

lgadban:


projects/gateway2/Makefile
```diff
@@ -43,10 +43,15 @@ tests/conformance/conformance_test.go:
...
 CONFORMANCE_ARGS:=-gateway-class=gloo-gateway -supported-features=Gateway,ReferenceGrant,HTTPRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRouteResponseHeaderModification,HTTPRoutePortRedirect,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRoutePathRedirect
 
+# TODO (Law[revert]): need to add CLI tests here, see CI pr.yaml workflow
 .PHONY: test
 test:
 	go test -ldflags=$(LDFLAGS) ./...
 
+.PHONY: test-full
+test-full:
+	go test -ldflags=$(LDFLAGS) -count=1 ./...
```
EItanya, line 53: What's the difference here?

EItanya (commented):
These changes LGTM, just awaiting reset of HEAD branch

projects/gateway2/Makefile
lgadban: the `-count=1` flag is a way of preventing go from caching test results; I ran into several instances where test results were cached and didn't accurately reflect the state of the codebase.

As there are valid reasons for caching tests, @yuval-k suggested that rather than modifying the existing `test` target, we can add a separate target specifically for the non-cached behavior.

EItanya (approved):

PR #8956, opened 2023-12-08
Vault Token Renewal: Separate Client Auth

sam-heilbron:
# Description
- Introduce the abstractions that will enable token renewal
- Demonstrate how we can mock external clients, or write unit tests to specific functionality

# Context
When reviewing this code, I would step through it in the following order:

## 1. ClientAuth
[ClientAuth](https://github.com/solo-io/gloo/blob/vault-token-renewal/auth/projects/gloo/pkg/bootstrap/clients/vault/auth.go#L23) represents the interface that a vault client must comply with. 

For a client to be authorized, it must have an access token set. That access token is provided by some source, and the "ClientAuth" responsibility is to extract this token bind it to the client.

At the moment we have 2 types of authorization strategies for clients:
1. [A token is extracted from the Settings object, and set directly on the client](https://github.com/solo-io/gloo/blob/32d85ecdc03ac8c9595a45ecd5e9e31120e7d6ff/projects/gloo/pkg/bootstrap/clients/vault.go#L135)
2. [A token is read from AWS, using the configuration provided in the Settings object](https://github.com/solo-io/gloo/blob/32d85ecdc03ac8c9595a45ecd5e9e31120e7d6ff/projects/gloo/pkg/bootstrap/clients/vault.go#L208)

The two implementations aren't easy to reason through, and we now have to figure out how to manage the renewal of these tokens, so we create the `ClientAuth` interface to represent this. This interface just extends the provided vault interface, by including our renewal function.

We can now have a single way that a [vault client is constructed and authorized](https://github.com/solo-io/gloo/blob/vault-token-renewal/auth/projects/gloo/pkg/bootstrap/clients/vault/client.go#L14), and the subtleties of those implementations are isolated to the implementation structs.

## 2. LocalClientAuth
This is the implementation for gathering tokens from a local location. It is the simplest implementation, as it just involves initializing the ClientAuth with a token, that it returns when `Login` is invoked.

## 3. RemoteClientAuth
This is the implementation for gathering tokens from a remote location. It allows injecting an `AuthMethod`, so in practice we inject the AWS implementation when users have configured it.

As part of this, we introduced metrics to capture the behavior, as well as retries to make the client login code resilient to network errors. Previously a single failed request would result in a failed login, but we provide some default retries.

## 4. Testing Strategy
A challenge that we have with this code is two-fold:
- We are integrating with an external system, and we want to be able to integrate with it correctly
- There is a single method that this package should export `Give me a vault client based on some settings`, but that doesn't allow us to inject the dependencies that we may need

I watched https://www.youtube.com/watch?v=EZ05e7EMOLM for some inspiration

_I would like to discuss this further, as there are some tradeoffs to any approach_

Option 1 (current):
I wanted to test the implementation of the remoteClientAuth, and ensure that I could mock the behavior of that remote client. I also wanted to avoid exporting the remoteClientAuth implementation, since it is not something external consumers of this package should be aware of. To write tests against this implementation though, I had to place the unit tests in the same package as the code. See https://github.com/golang/go/issues/25223#issuecomment-433611722 for some thoughts around this.

Option 2 (seriously considering):
I export just the methods that this module should export, and one of those allows injecting custom `ClientAuth` implementation. Using this, I can write tests against behaviors of the client, based on the injected ClientAuth implementation. This way I avoid just writing tests against the implementation details of the different ClientAuth implementations.

A benefit of this, is that I can test the behaviors of the client. A downside is that the low level details of how the renewal happens is not explicitly tested.

Option 3: 
Perhaps a combination of both approaches is the correct solution. Most tests should be aimed at the behavior that this module provides (an authenticated vault client), but occasionally we want to validate specific implementation details (like channel closing) and we should still support that.

## Testing steps
- A full suite of unit tests
- Relying on existing integration tests
- True e2e tests will not be introduced as part of this, but https://github.com/solo-io/solo-projects/issues/5356 would enable this

## Notes for reviewers
-

# Checklist:


sheidkamp:
Changes used in another branch

PR #8970, opened 2023-12-12
V1.15 CVE bump go version

sheidkamp:
# Description

Updating cloud builder dependencies to pull in fix for https://nvd.nist.gov/vuln/detail/CVE-2023-39326

# Validation

This vulnerability does not seem to be picked up by Trivy, so pull down the image:
`docker pull quay.io/solo-io/gloo:1.15.17-8970`
and inspect with Docker desktop


ci/check-go-version.sh
```diff
@@ -11,5 +11,5 @@ if [[ "$goVersion" == "$goModVersion"* ]]; then
 else
     echo "Your Go version ($goVersion) does not match the version from go.mod ($goModVersion)".
     echo "Please update your Go version to $goModVersion and re-run."
-    exit 1;
+    #exit 1;
```
nfuden, line 14: can we just remove this whole check? or make it scream if your version isnt go mod or 1.21+?

sheidkamp: I was mostly validating that this was what was causing buildbot to fail, and also that the results no longer contained the PR (and it doesn't).

I'll add a new branch of logic to handle this situation. Its a bit of an ugly fix, but this is only in the LTS branch and will eventually fade away.

ci/check-go-version.sh
inFocus7, line 1: Do we still need the `check-go-version.sh` file if we're removing it?

sheidkamp: removed

inFocus7 (approved):
🚀

PR #8976, opened 2023-12-13
Vault client refactor

sheidkamp:
# Description

This PR introduces a refactoring of the Hashicorp vault client code in preparation for adding renewal functionality. The changes are primarily to introduce a dependency injection of the Vault authorization method in the client creation to allow for testing.

## Code changes
The signature for `VaultClientForSettings` has changes from 
```
VaultClientForSettings(vaultSettings *v1.Settings_VaultSecrets) (*api.Client, error)
```
to 
```
VaultClientForSettings(ctx context.Context, vaultSettings *v1.Settings_VaultSecrets, vaultAuth vault.ClientAuth) (*api.Client, error)
```

The `pkg/bootstrap/clients/vault` package was created to support vault functionality, and it provides the ClientAuth interface:
```
type ClientAuth interface {
	vault.AuthMethod
}
```

This interface allows the authorization to by mocked with `mockgen` . This interface will be expanded later to support token renewal logic

## Testing steps

I manually verified behavior by loading this into an AWS EKS cluster (which has since been deleted) and inspecting the metrics around the logins, . The most recent changes have not been manually verified.

## Notes for reviewers

We added retry functionality around the login, but have set the retry times to 1, so we should have no functional changes to validate with e2e tests at this time.

# Checklist:


projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+// Since the vault.AuthMethod interface does not have corresponding mocks, we have to define our own.
+// todo - just mock the ClientAuth interface we define
+//go:generate mockgen -destination mocks/mock_auth.go -package mocks github.com/hashicorp/vault/api AuthMethod
+
+type ClientAuth interface {
+	vault.AuthMethod
+	// Start Renewal should be called after a successful login to start the renewal process
+	// it starts a go routine that will renew the token at the appropriate time, and so it does
+	// not return a value, it just goes off and does its thing
+	// StartRenewal(ctx context.Context, client *vault.Client, secret *vault.Secret)
```
nfuden, line 29: for reviewers who dont have context, there will be a fast follow to implement this

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+var _ ClientAuth = &staticTokenAuth{}
+var _ ClientAuth = &remoteTokenAuth{}
+
+var (
+	ErrEmptyToken = errors.New("unable to authenticate to vault with empty token. check Settings configuration")
```
nfuden, line 36: i wonder if this should actually just not have the check settings configuration piece and instead have that as wrapper on the settings code itself so that if the error comes from there it is wrapped with the check Settings configuration warning.

projects/gloo/pkg/bootstrap/clients/vault.go
```diff
@@ -49,175 +31,6 @@ func NewVaultSecretClientFactory(clientInit VaultClientInitFunc, pathPrefix, roo
...
-		}
-
-		return nil, err
-	}
-	if authInfo == nil {
-		return nil, errors.New("no auth info was returned after login")
-	}
-
-	return client, nil
+func VaultClientForSettings(ctx context.Context, vaultSettings *v1.Settings_VaultSecrets, vaultAuth vault.ClientAuth) (*api.Client, error) {
```
sam-heilbron, line 34: I like the idea that we expose a function at this package level, to generate an api.Client, since it abstract some of the details of the inner vault package. However, given that this function signature also exposes the vault.ClientAuth interface, can you help me understand what value this function provides beyond just requiring user to invoke `vault.NewAuthorizedClient` directly?

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+	vault "github.com/hashicorp/vault/api"
+	awsauth "github.com/hashicorp/vault/api/auth/aws"
+	"github.com/solo-io/gloo/pkg/utils"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+// In an ideal world, we would re-use the mocks provided by an external library.
+// Since the vault.AuthMethod interface does not have corresponding mocks, we have to define our own.
+// todo - just mock the ClientAuth interface we define
```
sam-heilbron, line 21: I think we can remove this todo. I initially wrote this when I hoped to mock a ClientAuth directly, but in the end we want to be able to mock the AuthMethod directly, so I don't think the todo is accurate

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+//go:generate mockgen -destination mocks/mock_auth.go -package mocks github.com/hashicorp/vault/api AuthMethod
+
+type ClientAuth interface {
+	vault.AuthMethod
+	// Start Renewal should be called after a successful login to start the renewal process
+	// it starts a go routine that will renew the token at the appropriate time, and so it does
+	// not return a value, it just goes off and does its thing
```
sam-heilbron, line 28: nit: I think that while true for certain implementations of the ClientAuth interface, it leaks certain implementation details which aren't always true. For example, the StaticTokenAuth will have a no-op, right?

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+	case *v1.Settings_VaultSecrets_AccessToken:
+		return newStaticTokenAuth(tlsCfg.AccessToken), nil
+
+	case *v1.Settings_VaultSecrets_Aws:
+		awsAuth, err := newAwsAuthMethod(tlsCfg.Aws)
+		if err != nil {
+			return nil, err
+		}
+
+		return NewRemoteTokenAuth(awsAuth), nil
```
sam-heilbron, line 52: Can you help me understand why this is exported and the other ClientAuth constructors are not?

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+func (r *remoteTokenAuth) Login(ctx context.Context, client *vault.Client) (*vault.Secret, error) {
+	var (
+		loginResponse *vault.Secret
+		loginErr      error
+	)
+
+	loginErr = retry.Do(func() error {
+		// If the context is cancelled, we should not retry, but we also can't return an error or we will retry
+		// so we return nil and rely on the caller to check the context
+		if ctx.Err() != nil {
```
sam-heilbron, line 122: I'd check out https://github.com/solo-io/ext-auth-service/pull/693 for an alternative approach to handling context cancellation in a retry block

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
```diff
@@ -0,0 +1,223 @@
...
+					Auth: &api.SecretAuth{
+						ClientToken: "a-client-token",
+					},
+				}, nil).AnyTimes()
+				clientAuth = NewRemoteTokenAuth(internalAuthMethod, retry.Attempts(5))
+			})
+
+			It("should return a context error", func() {
+				go func() {
+					time.Sleep(2 * time.Second)
```
sam-heilbron, line 156: Is there any concern about this introducing flakes if this test runs in a busy environment? I read the test as initiating a login attempt that will fail, sleep for a second, fail again, sleep for more than a second, and then on the third attempt, succeed to login. It seems like a fairly tight window we're trying to assert will occur.

One alternative, would be that we have the login method always return an error, and have the ClientAuth attempt login 10 times. Then we can use the same logic and assert both that the login failed with a particular error, and that the number of failures is less than the number of retries

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
```diff
@@ -0,0 +1,223 @@
...
+				assertions.ExpectStatLastValueMatches(MLastLoginSuccess, BeZero())
+				assertions.ExpectStatSumMatches(MLoginFailures, Equal(2))
+
+			})
+
+		})
+
+	})
+
+	Context("NewClientAuth", func() {
```
sam-heilbron, line 174: This function was changed to `ClientAuthFactory`

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
```diff
@@ -0,0 +1,223 @@
...
+		// These tests validate that the constructor maps the Gloo Settings into the appropriate ClientAuth interface
+		// it does not test the underlying implementations, as those are handled in the above tests
+
+		DescribeTable("should error on invalid inputs",
+			func(vaultSettings *v1.Settings_VaultSecrets, expectedError types.GomegaMatcher) {
+				clientAuth, err := ClientAuthFactory(vaultSettings)
+				Expect(err).To(expectedError)
+				Expect(clientAuth).To(BeNil())
+			},
+			Entry("partial accessKey / secretAccessKey", &v1.Settings_VaultSecrets{
```
sam-heilbron, line 184: Are there any other cases that would cause the ClientAuthFactory to return an error? I chose this first one to demonstrate an example, but I didn't do a thorough review

projects/gloo/pkg/bootstrap/clients/vault_test.go
```diff
@@ -0,0 +1,209 @@
...
+			// It("startRenewal should return nil", func() {
+			// 	err := clientAuth.StartRenewal(ctx, nil)
+			// 	Expect(err).NotTo(HaveOccurred())
+			// })
+
+		})
+
+	})
+
+	Context("NewRemoteTokenAuth", func() {
```
sam-heilbron, line 110: Can you help me understand what behaviors you see the tests in this file providing that our `vault` package does not have?

projects/gloo/pkg/bootstrap/clients/vault_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+})
+
+func resetViews() {
+	views := []*view.View{
+		vault.MLastLoginFailureView,
+		vault.MLastLoginSuccessView,
+		vault.MLoginFailuresView,
+		vault.MLoginSuccessesView,
+		vault.MLastLoginFailureView,
```
sam-heilbron, line 201: `MLastLoginFailureView,` is defined twice

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -341,7 +342,7 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 
 	getVaultInit := func(vaultSettings *v1.Settings_VaultSecrets) bootstrap_clients.VaultClientInitFunc {
 		return func() *vaultapi.Client {
-			c, err := bootstrap_clients.VaultClientForSettings(vaultSettings)
+			vaultAuth, err := vault.ClientAuthFactory(vaultSettings)
```
sam-heilbron, line 345: Can you help me understand what this `getVaultInit` function wrapper provides us? I wonder if, given the refactor in this PR, the vault init function is still necessary

test/gomega/transforms/metrics.go
```diff
@@ -0,0 +1,38 @@
+package transforms
+
+import (
+	errors "github.com/rotisserie/eris"
+	"go.opencensus.io/stats/view"
+)
+
+func WithLastValueTransform() func(rows []*view.Row) (int, error) {
```
sam-heilbron, line 8: nit: Exported functions in go should contain a comment : https://tip.golang.org/doc/comment

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -257,7 +258,8 @@ var _ = Describe("Bootstrap Clients", func() {
 
 		setVaultClientInitMap := func(idx int, vaultSettings *v1.Settings_VaultSecrets) {
 			vaultClientInitMap[idx] = func() *vaultapi.Client {
-				c, err := clients.VaultClientForSettings(vaultSettings)
+				vaultAuth, _ := vault_client.ClientAuthFactory(vaultSettings)
```
sam-heilbron, line 261: I don't think we want to drop the error here. If there is a reason to do it, I think it warrants a comment

test/services/gloo.go
```diff
@@ -328,8 +329,8 @@ func constructTestOpts(ctx context.Context, runOptions *RunOptions, settings *gl
 		// The test author has configured the secret source to be Vault, instead of an in memory cache
 		// As a result, we need to construct a client to communicate with that vault instance
 		vaultSecretSource := settings.GetVaultSecretSource()
-
-		vaultClient, err := bootstrap_clients.VaultClientForSettings(vaultSecretSource)
+		vaultAuth, _ := vault_client.ClientAuthFactory(vaultSecretSource)
```
sam-heilbron, line 332: Same as above, I think we should just assert than an error wasn't encountered

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,209 @@
...
+		utils.Measure(ctx, MLastLoginFailure, time.Now().Unix())
+		utils.MeasureOne(ctx, MLoginFailures)
+		return nil, eris.Wrapf(loginErr, "unable to authenticate to vault")
+	}
+
+	if loginResponse == nil {
+		contextutils.LoggerFrom(ctx).Error(ErrNoAuthInfo)
+		utils.Measure(ctx, MLastLoginFailure, time.Now().Unix())
+		utils.MeasureOne(ctx, MLoginFailures)
+		return nil, ErrNoAuthInfo
```
sam-heilbron, line 150: Is there a unit test for this case?

sam-heilbron (commented):
Looks good! All small thoughts that can easily be handled in a follow-up if others are happy with the PR. I'm OOO tomorrow, but no objections from me to have this merge

projects/gloo/pkg/bootstrap/clients/vault.go
sheidkamp: This is the dependency injection so we can unit test the public facing method.

projects/gloo/pkg/syncer/setup/setup_syncer.go
jbohanon: The decision to pass in setup funcs was made during the implementation of the [multiple secret sources feature](https://github.com/solo-io/gloo/pull/8263). It solves a nuanced issue where we would end up with a nil vault api client. To support multiple secret sources, possibly multiple Vault secret sources, we defer the api client creation to the ResourceClient creation and check the success/failure of the client creation then. Previously, we initialized the api client in the setup_syncer. I'm happy to be convinced that this is no longer necessary because it is a bit convoluted, but we'd have to validate that the refactor doesn't affect the multiple secret sources feature implementation (hopefully the tests would catch this).

test/helpers/metrics.go
```diff
@@ -15,7 +14,9 @@ import (
 // "the metric was never recorded" from "a value of 0 was recorded"
 func ReadMetricByLabel(metricName string, labelKey string, labelValue string) (int, error) {
 	rows, err := view.RetrieveData(metricName)
-	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to retrieve data for %s", metricName)
+	}
```
jbohanon, line 19: What purpose does this change serve in this PR?

projects/gloo/pkg/bootstrap/clients/vault/client.go
jbohanon, line 1: Can we change the wording in these function names and comments to be a bit more accurate? All of this logic involves _authentication_ (we are who we say we are i.e. login) not _authorization_ (we have access to that which we are requesting access)

projects/gloo/pkg/bootstrap/clients/vault_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+		clientAuth vault.ClientAuth
+	)
+
+	BeforeEach(func() {
+		ctx, cancel = context.WithCancel(context.Background())
+
+		// The tests below will be responsible for assigning this variable
+		// We re-set it here, just to be safe
+		//clientAuth = nil
```
jbohanon, line 36: In this case can we assert in a `JustBeforeEach` that `clientAuth != nil`?

projects/gloo/pkg/bootstrap/clients/vault_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+		When("token is empty", func() {
+
+			BeforeEach(func() {
+				vaultSettings := &v1.Settings_VaultSecrets{
+					AuthMethod: &v1.Settings_VaultSecrets_AccessToken{
+						AccessToken: "",
+					},
+				}
+				clientAuth, _ = vault.ClientAuthFactory(vaultSettings)
```
jbohanon, line 58: Like Sam said above we should probably assert no error here instead of dropping it. Just to be safe

jbohanon (commented):
Great work! I love how much more testable this is and how many tests you've added

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
```diff
@@ -0,0 +1,222 @@
...
+				ctrl := gomock.NewController(GinkgoT())
+				internalAuthMethod := mocks.NewMockAuthMethod(ctrl)
+				internalAuthMethod.EXPECT().Login(ctx, gomock.Any()).Return(nil, eris.New("mocked error message")).AnyTimes()
+
+				clientAuth = NewRemoteTokenAuth(internalAuthMethod, retry.Attempts(3))
+			})
+
+			It("should return the error", func() {
+				secret, err := clientAuth.Login(ctx, nil)
+				Expect(err).To(MatchError("unable to authenticate to vault: mocked error message"))
```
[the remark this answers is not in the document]
sheidkamp: Will do this in the next round of commits

test/gomega/transforms/metrics.go
sheidkamp: Added here and below

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
sheidkamp: Done

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,213 @@
...
+		possibleErrStrings = append(possibleErrStrings, "secret access key must be defined for AWS IAM auth")
+	}
+
+	// if we have only partial configuration set
+	if len(possibleErrStrings) == 1 {
+		return nil, errors.New("only partial credentials were provided for AWS IAM auth: " + possibleErrStrings[0])
+	}
+
+	// At this point, we either have full auth configuration set, or are in an ec2 environment, where vault will infer the credentials.
+	loginOptions := []awsauth.LoginOption{awsauth.WithIAMAuth()}
```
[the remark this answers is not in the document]
jbohanon: Might be worth investigating and writing an issue if needed. I think @inFocus7 wrote this feature so he might have more context on the assumptions we can make there

test/helpers/metrics.go
sam-heilbron: This function blended a combination of mechanisms for handling errors. On one hand, we asserted that an error did not occur, and a ginkgo.Fail would be invoked here. On the other hand, we return an error from this function. 
I made this change so that we have a consistent mechanism for handling errors in this function: we return it.

sam-heilbron: It's not required for this PR, though I was making changes to metrics assertions, and I came across this code and figured it was worth it to change

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: All good, thanks for the explanations! I agree it's not a change I'd want to include in this PR, I was trying to better understand it's purpose and now I know it

sam-heilbron: Happy to resolve this

projects/gloo/pkg/bootstrap/clients/vault/auth.go
nfuden: I still see the check settings? am I missing something?

projects/gloo/pkg/bootstrap/clients/vault/auth.go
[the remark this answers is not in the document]
nfuden: 1 is a guaranteed error, 2 we have to at least try to see if there are injected unknown reasons we have auth. 
We will fail higher up

projects/gloo/pkg/bootstrap/clients/vault/auth.go
sheidkamp: Note: pull context check from code

projects/gloo/pkg/bootstrap/clients/vault/auth.go
sheidkamp: TODO: closer look

test/services/gloo.go
sheidkamp: TODO

projects/gloo/pkg/bootstrap/clients/vault/auth_test.go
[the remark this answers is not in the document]
sheidkamp: Updated

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -0,0 +1,213 @@
...
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+// In an ideal world, we would re-use the mocks provided by an external library.
+// Since the vault.AuthMethod interface does not have corresponding mocks, we have to define our own.
+// todo - just mock the ClientAuth interface we define
+//go:generate mockgen -destination mocks/mock_auth.go -package mocks github.com/hashicorp/vault/api AuthMethod
+
+type ClientAuth interface {
```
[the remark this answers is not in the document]
jbohanon: In an offline discussion, we decided this warrants a larger conversation about the role/value of solo-kit and what to do with _that_ code.

projects/gloo/pkg/bootstrap/clients/vault.go
sam-heilbron: WOrks for me, thanks for the explanation

sam-heilbron (approved):
great work 🚀

jbohanon (dismissed):

projects/gloo/pkg/bootstrap/clients/vault_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	When("context is cancelled before login succeeds", func() {
+
+		retryAttempts := uint(5)
+		BeforeEach(func() {
+			ctrl = gomock.NewController(GinkgoT())
+			internalAuthMethod := mocks.NewMockAuthMethod(ctrl)
+			// The auth method will return an error twice, and then a success
+			// but we plan on cancelling the context before the success
+			internalAuthMethod.EXPECT().Login(ctx, gomock.Any()).Return(nil, eris.New("error")).AnyTimes()
```
sam-heilbron, line 175: This comment is no longer accurate, since we have changed the implementation

sam-heilbron (approved):

nfuden (approved):

projects/gloo/pkg/bootstrap/clients/vault_test.go
sheidkamp: Will update in follow up

Issue #8968, opened 2023-12-12
Gloo fails after upgrade from 1.14.27 to 1.15.x


ndtmdsma:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.15.x

### Kubernetes Version

v1.24

### Describe the bug

Upgrading from Gloo 1.14.x to 1.15.x causes Gloo to fail entirely. Everything starts, glooctl check says everything is OK, but Gloo doesn't route anything. I get warning as below for **all** upstreams:

```
[warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:128] gRPC config for type.googleapis.com/envoy.config.cluster.v3.Cluster rejected: Error adding/updating cluster(s) app1_gloo-system: Didn't find a registered config subscription factory implementation for name: 'envoy.config_subscription.rest', kube-svc:app2-0-80_ns2: Didn't find a registered config subscription factory implementation for name: 'envoy.config_subscription.rest', consul-svc:app3_gloo-system: Didn't find a registered config subscription factory implementation for name: 'envoy.config_subscription.rest'
```

It's a very long list.

### Expected Behavior

Gloo working after upgrade

### Steps to reproduce the bug

Upgrade Gloo from 1.14.27 to 1.15.x (I tried all patch versions). 

### Additional Environment Detail

Using consul upstreams discovery (but gloo fails for both consul and kube upstreams). 

Running in EKS.

Helm values: 
```  
    crds:
      create: true
    discovery:
      deployment:
        floatingUserId: false
        fsGroup: 10101
        image:
          extended: false
          repository: discovery
          tag: 1.15.17
        replicas: 1
        runAsUser: 10101
        stats:
          enabled: true
          serviceMonitorEnabled: true
          podMonitorEnabled: true
      enabled: true
      fdsMode: WHITELIST
      serviceAccount:
        disableAutomount: false
      logLevel: "info"
      udsOptions:
        enabled: true
    gateway:
      enabled: true
      certGenJob:
        enabled: true
        image:
          repository: certgen
        restartPolicy: OnFailure
        setTtlAfterFinished: true
        ttlSecondsAfterFinished: 60
        runAsUser: 10101
        cron:
          enabled: false
          schedule: "* * * * *"
        runOnUpdate: false
      rolloutJob:
        enabled: true
        image:
          repository: kubectl
        restartPolicy: OnFailure
        runAsUser: 10101
        ttlSecondsAfterFinished: 60
      cleanupJob:
[...]

PR #8731, opened 2023-09-28
Better support for unreleased envoy-gloo assets

ben-taussig-solo:
# Description
 - Update CI to better support unreleased assets from envoy-gloo CI builds
 - envoy-gloo CI builds and tags assets with a commit SHA. This leads to issues when we test unreleased assets in gloo CI, as gloo CI assumes that envoy-gloo assets are tagged in major.minor.patch-label format.
 - This PR introduces a try/except structure to the portion of CI that pulls envoy-gloo assets. First, we try to pull the asset using the major.minor.patch-label format. If that fails, we try to pull the asset using the commit SHA format.

# Context
 - essentially a port of this PR from the closed-source repo: https://github.com/solo-io/solo-projects/pull/5257

PR #8982, opened 2023-12-13
[1.15.x] Resolve rest EDS upgrade failure

ben-taussig-solo:
# Description
 - This PR revises `test/e2e/happypath_test.go`
 - Prior to this PR, this test contained an improper attempt to construct Ginkgo `Describe` blocks in a loop. 
   - This test attempted to concisely evaluate the same set of tests with and without REST EDS enabled, but the loop was not properly constructed and the tests were simply being run twice with REST EDS disabled.
 - This PR removes the loop structure and instead contains two separate `Describe` blocks, one for the REST EDS disabled case and one for the REST EDS enabled case.
 - This PR also extracts the common test logic into helper functions to reduce code duplication.

### Test Structure
 - I have tried to keep the test structure as close to the original as possible -- below I've highlighted the ginkgo spec tree before and after this PR.
 - These are essentially identical, except that the original test contained a Describe block which was not properly constructed and has been removed.
 - Note that the revised test structure contains two essentially identical `Describe` blocks, one for the REST EDS disabled case and one for the REST EDS enabled case.
#### Original Test structure
```go
Describe("Happy path", func() {
  BeforeEach(func() {
  AfterEach(func() {
    Describe(fmt.Sprintf("%s: (%s)", testCase.Title, testCase.TransportApiVersion.String()), func() {
      Describe("in memory", func() {
        BeforeEach(func() {
        It("should not crash", func() {
        It("should not crash multiple methods", func() {
        It("correctly configures envoy to emit virtual cluster statistics", func() {
        It("it correctly passes the suppress envoy headers config", func() {
        It("it correctly DID NOT pass the suppress envoy headers config", func() {
        It("passes a health check", func() {
        Context("ssl", func() {
          BeforeEach(func() {
          Context("simple ssl", func() {
            BeforeEach(func() {
            It("should work with ssl", func() {
          Context("sni", func() {
            BeforeEach(func() {
            It("should work with ssl", func() {
        Context("sad path", func() {
          It("should error the proxy with two listeners with the same bind address", func() {
```

#### Revised Test structure
```go
Describe("Happy path", func() {
  BeforeEach(func() {
  AfterEach(func() {
  Describe("in memory", func() {
    Describe("Rest Eds Enabled", func() {
      BeforeEach(func() {
      It("should not crash", func() {
      It("should not crash with multiple methods", func() {
      It("correctly configures envoy to emit virtual cluster statistics", func() {
      It("correctly passes suppressEnvoyHeaders config", func() {
      It("correctly did not pass suppressEnvoyHeaders config", func() {
      It("passes health check", func() {
      Context("ssl", func() {
        BeforeEach(func() {
        Context("simple ssl", func() {
          BeforeEach(func() {
          It("should work with ssl", func() {
        Context("sni", func() {
          BeforeEach(func() {
          It("should work with ssl", func() {
      Context("sad path", func() {
        It("should error the proxy with two listeners with the same bind address", func() {
  Describe("Rest Eds Disabled", func() {
    ... same as above ...
```

### Test logic
 - In order to reduce code duplication, I have extracted common logic into helper functions.
 - Prior to the PR, these tests essentially followed the pattern below:
   1. Declare scope-level variables at the top of a `Describe` block
   2. Initialize the scope-level variables in a `BeforeEach` block
   3. Run the test logic, occasionally referencing the scope-level variables
- The reliance on scope-level variables complicated the process of extracting common logic into helper functions. 
- To address this, most of what were previously scope-level variables now live in a struct called `testArgs`, which is passed to the helper functions.
- `testArgs` is initialized in a `BeforeEach` block, and is passed as an argument to the helper functions called from the `It` blocks.


test/kube2e/upgrade/upgrade_test.go
```diff
@@ -73,6 +73,11 @@ var _ = Describe("Kube2e: Upgrade Tests", func() {
 			It("uses helm to add a second gateway-proxy in a separate namespace without errors", func() {
 				addSecondGatewayProxySeparateNamespaceTest(ctx, crdDir, testHelper, chartUri, targetReleasedVersion, strictValidation)
 			})
+			It("doesn't error when settings.enableRestEds is set to true", func() {
```
sam-heilbron, line 76: I like the idea of a kubernetes test. I just want to call out that we have an existing integration test that is attempting to valdate the behavior of REST EDS: https://github.com/solo-io/gloo/blob/main/test/e2e/happypath_test.go#L54. Regardless of your decision around where an effective test should live, I think we should find an answer for why the existing tests did not catch this behavior

Makefile
```diff
@@ -48,7 +48,7 @@ VERSION ?= 1.0.1-dev
 
 SOURCES := $(shell find . -name "*.go" | grep -v test.go)
 
-ENVOY_GLOO_IMAGE ?= quay.io/solo-io/envoy-gloo:1.26.5-patch1
+ENVOY_GLOO_IMAGE ?= quay.io/solo-io/envoy-gloo:2d9b6aa2ab4fbc8c47ebb55e438c05a654546fec
```
sam-heilbron, line 51: We have a mechanism in Gloo Edge (https://github.com/solo-io/gloo/blob/main/.github/workflows/codegen.yaml#L26) to introduce code that will prevent CI from passing. Given that this is a temporary change, to demonstrate other functionality, can we add this DO_NOT_SUBMIT to somewhere in the code? THat will prevent this PR from merging accidentally

ci/cloudbuild/run-tests.yaml
```diff
@@ -30,9 +30,15 @@ steps:
   args:
   - '-c'
   - |
-    ENVOY_VERSION=v$$(make print-ENVOY_GLOO_IMAGE | cut -d: -f2)
-    gsutil cp gs://solo-public-artifacts.solo.io/envoy/$$ENVOY_VERSION/envoy.stripped /workspace/envoy
-    chmod +x /workspace/envoy
+    { # try to pull release assets first
```
sam-heilbron, line 33: Can you help me understand these changes? What problem does this solve / how does it manifest itself

changelog/v1.15.18/rest-eds-upgrade-failure.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: DEPENDENCY_BUMP
+    dependencyOwner: envoy-gloo
+    dependencyRepo: solo-io
+    dependencyTag: NOT_YET_RELEASED
+    description: >-
+      Update envoy-gloo to resolve issues configuring rest EDS
+  - type: NON_USER_FACING
+    description: >- 
+      Update CI to better support unreleased assets from envoy-gloo-ee CI builds
```
sam-heilbron, line 10: nit: envoy-gloo not, envoy-gloo-ee right?

changelog/v1.15.18/rest-eds-upgrade-failure.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: DEPENDENCY_BUMP
+    dependencyOwner: envoy-gloo
+    dependencyRepo: solo-io
+    dependencyTag: NOT_YET_RELEASED
+    description: >-
+      Update envoy-gloo to resolve issues configuring rest EDS
```
sam-heilbron, line 7: Given that we are pulling in functionality from an external project (envoy-gloo), I think we should have an integration test which demonstrates that this functionality in Edge works as expected. Since we have uncovererd a bug, I think this is a chance to expand our tests in Gloo Edge to prevent future regressions. Can we add a test that will fail without this dependency change, and that succeeds with it?

sam-heilbron (changes requested):
Generally looks good.

changelog/v1.15.18/rest-eds-upgrade-failure.yaml
ben-taussig-solo: yes, good catch

changelog/v1.15.18/rest-eds-upgrade-failure.yaml
ben-taussig-solo: I think the plan as discussed in stand-up was to merge data-plane bumps without testing updates and then do the testing updates as a fast-follow. I need to sync with Manuel about that. I agree that we need to have testing against this functionality (and also that we need to repair existing gaps in our tests as described [here](https://github.com/solo-io/gloo/issues/8968#issuecomment-1856085120))

sam-heilbron: I'm a bit concerned about `I think the plan as discussed in stand-up was to merge data-plane bumps without testing updates and then do the testing updates as a fast-follow.`

What is the value of introducing code without corresponding tests that demonstrate it to behave the way we would expect? If it comes down to a timing issue (we need this functionality merged into Gloo Edge by some time) then I think we can regroup on how to most efficiently verify (can be manual) this functionality. 

As an aside, I would expect that we can write some automated tests for this in <1 day, so I'd be surprised if our delivery of this feature being delayed a day was unnacceptable to product.

ci/cloudbuild/run-tests.yaml
ben-taussig-solo: this is a backport of this PR to v1.15.x. This allows us to test unreleased envoy-gloo assets in gloo OSS CI
https://github.com/solo-io/gloo/pull/8731

Makefile
ben-taussig-solo: resolved here: https://github.com/solo-io/gloo/pull/8982/commits/0024443b15a1095fc8b11d42a4b1bf8da971b113

changelog/v1.15.18/rest-eds-upgrade-failure.yaml
ben-taussig-solo: resolved here: https://github.com/solo-io/gloo/pull/8982/commits/b62f7043a0ac483b399a4b9605a02bcbbe0a5290

Makefile
```diff
@@ -48,7 +48,9 @@ VERSION ?= 1.0.1-dev
 
 SOURCES := $(shell find . -name "*.go" | grep -v test.go)
 
-ENVOY_GLOO_IMAGE ?= quay.io/solo-io/envoy-gloo:1.26.5-patch1
+# pre-release envoy-gloo image
+# DO_NOT_SUBMIT
```
sam-heilbron, line 52: I would inspect the do-not-submit script to confirm how this works. If this comment were having an impact, I would expect the static analysis job would fail during the final step.

ci/cloudbuild/run-tests.yaml
sam-heilbron: Super helpful! Personally it's hard to find this history/rationale. It could be that I didn't look deeply enough into the commit history, so I'm fine with that as the solution. Do you think a comment outlining why we perform this action would be useful?

Secondly, this adds complexity into our cloudbuild definition. If I'm a developer executing code locally, is there a situation that I would ever want to perform these actions? If so, I'd be in favor of defining this complexity in a re-usable place. This way, if/when we change cloud providers we don't have to re-write this logic again

Makefile
ben-taussig-solo: Looking at the [codegen logs](https://github.com/solo-io/gloo/actions/runs/7249379897/job/19747239624?pr=8982), we can see that the script is running but it doesn't seem to actually be scanning any files (or at least we're failing to increment the file count)

```
Run ./ci/do-not-submit.sh changelog/v1.15.18/rest-eds-upgrade-failure.yaml Makefile ci/cloudbuild/run-tests.yaml
0 go files checked, none contain DO_NOT_SUBMIT
````

I think that suggests that the checks we're doing against the filenames passed into the script are not working as intended

For now I'm just going to leave this PR as a draft

Comment (sam-heilbron) [reply Makefile:52]: `I think that suggests that the checks we're doing against the filenames passed into the script are not working as intended` I'm not sure I agree with this. To me it means that the script behaves a certain way, and we're not using it that way. I think our options are: expand the script to fit our use case, adjust our use case to fit the script.

https://github.com/solo-io/gloo/blob/main/ci/do-not-submit.sh#L8

It seems like the comment format isn't following what the script expects

File: ci/cloudbuild/run-tests.yaml (hunk 3 above)
Comment (ben-taussig-solo) [reply ci/cloudbuild/run-tests.yaml:33]: Updated commentary in [536529d](https://github.com/solo-io/gloo/pull/8982/commits/536529d0cf282afd76e2fee7eb20c338577b994e)

```sh
  - |
    { # try to pull release assets first
      # release assets are in v1.x.y (semver, prefixed with a literal v) format
      ENVOY_VERSION=v$$(make print-ENVOY_GLOO_IMAGE | cut -d: -f2)
      gsutil cp gs://solo-public-artifacts.solo.io/envoy/$$ENVOY_VERSION/envoy.stripped /workspace/envoy
      chmod +x /workspace/envoy
    } || { 
      # if that fails, pull from CI
      # CI assets are in SHA format -- the difference from the above is that the SHA is not prefixed with a literal v
      ENVOY_VERSION=$$(make print-ENVOY_GLOO_IMAGE | cut -d: -f2)
      gsutil cp gs://solo-public-artifacts.solo.io/envoy/$$ENVOY_VERSION/envoy /workspace/envoy
      chmod +x /workspace/envoy
```

I also put up a PR against main with the same change here: https://github.com/solo-io/gloo/pull/8997

Makefile
ben-taussig-solo: I put an issue up discussing potential improvements to this functionality: https://github.com/solo-io/gloo/issues/8998

As such, I am resolving the thread.

PR #8984, opened 2023-12-14
add route options to v2.0.x

lgadban:
Introduces support for `RouteOptions` CR to be used as an `ExtensionRef` for GWAPI filters.

changes included:
* add new reconciler for `RouteOptions` (including RBAC permissions)
* expand query engine to support `LocalObjectReference`, which is the reference type for extensionRef
* introduces a new framework for "Extension Plugins"
  * converts the ExtensionRef plugin to be a "standard http filter plugin"
  * new registry to contain extension plugins, e.g. to handle the logic of `RouteOptions`, regardless of how the plugin was referenced
  * this will be useful when we use alternative mechanisms to inline extensionRef for including plugins, such as policy/meta-resource attachment (e.g. to handle `VHostOptions`)


projects/gateway2/query/query.go
```diff
@@ -23,6 +23,7 @@ var (
 	ErrNoMatchingListenerHostname = fmt.Errorf("no matching listener hostname")
 	ErrNoMatchingParent           = fmt.Errorf("no matching parent")
 	ErrNotAllowedByListeners      = fmt.Errorf("not allowed by listeners")
+	ErrLocalObjRefMissingKind     = fmt.Errorf("localObjRef provided with empty kind")
```
EItanya, line 26: Should we advertise that this defaults to `gateway.solo.io/RouteOptions` rather than erroring. Could be a good initial usage

projects/gateway2/translator/httproute/filterplugins/extensionref/extension_ref_plugin.go
```diff
@@ -0,0 +1,63 @@
...
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type GetPluginFunc = func(gk schema.GroupKind) (filterplugins.ExtPlugin, error)
+
+type Plugin struct {
+	queries   query.GatewayQueries
+	getPlugin GetPluginFunc
```
lgadban, line 16: this was a temporary solution to not introduce the registry as a dependency to the plugin, but after seeing it sketched out, passing in the registry may be a better solution

projects/gateway2/query/query.go
lgadban: I'm undecided on this. It could be a nice default value but could paint users into a corner later.

Deferring to others to weigh-in

lgadban:
converting to draft to explore non-GroupKind key'd map for extension plugins

PR #8986, opened 2023-12-14
cli: Refactor getting dumps from envoy and log these dumps on test failures

davidjumani:
# Description

Refactor getting dumps from the proxy. Also export this new method which is quite useful in debugging failures, especially in CI.
This way the config dump, stats, cluster, listener, heap and other info can be dumped if tests fail in CI making debugging them easier

[Example dump](https://github.com/solo-io/gloo/suites/19067766467/artifacts/1116445953)

## API changes
N/A

## Code changes
- Exposed `GetEnvoyDump` in gloo/cli/pkg/cmd/gateway
- Use the newly exposed method to dump data when kube2e tests fail
- Refactored tests in `utils` package to `utils_test` for consistency

## CI changes
N/A

## Docs changes
N/A

# Context
At times failures are difficult to debug in CI as we do not know if there was an issue with the upstream cluster, etc. This way as part of the tests fail handler, the envoy dump can be fetched to make debugging easier

## Interesting decisions
 
N/A

## Testing steps

N/A

## Notes for reviewers

N/A

# Checklist:


projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -35,7 +35,7 @@ func dumpCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra.
 	return cmd
 }
 
-func getEnvoyCfgDump(opts *options.Options) (string, error) {
+func GetEnvoyDump(opts *options.Options, path string, timeout time.Duration) (string, error) {
```
nfuden, line 38: nit: exported functions should have comments

nfuden: Also I dont think this is a dump any more so i think the name needs changing

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -95,7 +95,10 @@ func getEnvoyCfgDump(opts *options.Options) (string, error) {
 			return "", errors.Errorf("timed out trying to connect to Envoy admin port")
 		}
 	}
+}
 
+func getEnvoyCfgDump(opts *options.Options) (string, error) {
+	return GetEnvoyDump(opts, "/config_dump", 5*time.Second)
```
jbohanon, line 101: Is there a purpose for this indirection? Can we make this tail call from `GetEnvoyCfgDump` instead?

projects/gloo/pkg/utils/gloo_resources_test.go
```diff
@@ -1,9 +1,10 @@
-package utils
+package utils_test
```
nfuden, line 1: why are we moving to blackbox testing?

projects/gloo/pkg/utils/log_filtering_test.go
```diff
@@ -1,11 +1,12 @@
-package utils
+package utils_test
```
nfuden, line 1: why are we making this change? it doesnt seem in scope for the pr nor does it seem to be advantageous to blackbox test here

projects/gloo/pkg/utils/ssl.go
```diff
@@ -478,7 +478,7 @@ func convertVersion(v ssl.SslParameters_ProtocolVersion) (envoyauth.TlsParameter
 	return envoyauth.TlsParameters_TLS_AUTO, TlsVersionNotFoundError(v)
 }
 
-func verifySanListToMatchSanList(sanList []string) []*envoymatcher.StringMatcher {
+func VerifySanListToMatchSanList(sanList []string) []*envoymatcher.StringMatcher {
```
nfuden, line 481: Why are we making this exported? every exported function makes it harder to change our signatures in the future

nfuden: If this is to be exported it should have a comment

projects/gloo/pkg/utils/utils_test.go
```diff
@@ -1,25 +1,26 @@
...
 
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/core/matchers"
+	"github.com/solo-io/gloo/projects/gloo/pkg/utils"
 )
 
-var _ = Describe("PathAsString", func() {
+var _ = Describe("utils.PathAsString", func() {
```
nfuden, line 10: This is already in the utils package why would its test have another level of utils in its desc?

projects/gloo/pkg/utils/ssl_test.go
```diff
@@ -26,11 +27,11 @@ var _ = Describe("Ssl", func() {
 		tlsSecret              *v1.TlsSecret
 		secret                 *v1.Secret
 		secrets                v1.SecretList
-		configTranslator       *sslConfigTranslator
-		resolveCommonSslConfig func(cs CertSource, secrets v1.SecretList) (*envoyauth.CommonTlsContext, error)
+		configTranslator       utils.SslConfigTranslator
```
jbohanon, line 30: This type changed from pointer to non. Some of the function receivers we call on this are pointer receivers, so this should probably still be a pointer

test/helpers/kube_dump.go
```diff
@@ -219,18 +225,62 @@ func kubeList(namespace string, target string) ([]string, error) {
...
 
+// EnvoyDumpOnFail creates a small dump of the envoy configs when a test fails.
+// This is useful for debugging test failures.
+// The dump is written to _output/envoy-dump.
+// The dump includes:
+// - config dump
+// - stats
+// - clusters
+// - listeners
+func EnvoyDumpOnFail(out io.Writer, namespace string) func() {
```
jbohanon, line 236: why do we need to accept an `io.Writer` here? If this is to satisfy an interface, can we rename `out` to `_`, and if not can we just remove it?

test/helpers/kube_dump.go
```diff
@@ -219,18 +225,62 @@ func kubeList(namespace string, target string) ([]string, error) {
...
+		Top: options.Top{
+			Ctx: contextWithCancel,
+		},
+		Proxy: options.Proxy{
+			Name: "gateway-proxy",
+		},
+	}
+	cfg, err := gateway.GetEnvoyDump(opts, path, 30*time.Second)
+	if err != nil {
+		f.WriteString("*** Unable to get envoy " + path + " dump ***\n")
```
ben-taussig-solo, line 264: Nit: can we write the original error here as well? I think that would make it easier to debug if things go wrong

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -120,62 +123,5 @@ func statsCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra
 }
 
 func getEnvoyStatsDump(opts *options.Options) (string, error) {
-	adminPort := strconv.Itoa(int(defaults.EnvoyAdminPort))
```
sam-heilbron, line 123: Port-forwarding is a common action that takes place in our CLI and our tests. I think having a shared utility to handle this would be valuable. Recently I added https://github.com/solo-io/dev-portal/tree/main/pkg/utils/kubeutils to the dev-portal, and I wonder if we can introduce a similar package in Gloo

test/helpers/kube_dump.go
```diff
@@ -219,18 +225,62 @@ func kubeList(namespace string, target string) ([]string, error) {
...
+			Namespace: namespace,
+		},
+		Top: options.Top{
+			Ctx: contextWithCancel,
+		},
+		Proxy: options.Proxy{
+			Name: "gateway-proxy",
+		},
+	}
+	cfg, err := gateway.GetEnvoyDump(opts, path, 30*time.Second)
```
ben-taussig-solo, line 262: How do you feel about creating a wrapper for `GetEnvoyDump` that takes in ctx, namespace, and proxy name arguments instead of an `options.Options` struct? I think it's potentially unclear to future users that these specific options need to be specified

test/helpers/kube_dump.go
jbohanon: I see now it is a `PreFailHandler`. In this case, can we plumb the `GinkgoWriter` through to the `setupOutDir` func and use it when printing via the `fmt.Fprint*` functions?

test/kube2e/istio/istio_suite_test.go
```diff
@@ -61,6 +61,7 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 
 	skhelpers.RegisterPreFailHandler(helpers.KubeDumpOnFail(GinkgoWriter, testHelper.InstallNamespace))
+	skhelpers.RegisterPreFailHandler(helpers.EnvoyDumpOnFail(GinkgoWriter, testHelper.InstallNamespace))
```
nfuden, line 64: The fact that we always need to remember to have both of these calls indicates to me that there should be a wrapper caller like helpers.StandardGlooDumpOnFail or something

projects/gloo/pkg/utils/log_filtering_test.go
ben-taussig-solo: +1

projects/gloo/pkg/utils/gloo_resources_test.go
davidjumani: Some tests in this module are blackbox while some are not. Doing this for consistency. Also, utilizing the newly exported `GetEnvoyDump` method leads to cyclic dependencies [here](https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=5ededf0d-c0f1-4505-af52-f6882033d5cb). Added this to also avoid it

projects/gloo/pkg/utils/ssl.go
davidjumani: This was required for moving the tests to their own package to avoid cyclic dependencies. It is used only in the tests as of now

test/helpers/kube_dump.go
```diff
@@ -219,18 +225,62 @@ func kubeList(namespace string, target string) ([]string, error) {
...
+// The dump is written to _output/envoy-dump.
+// The dump includes:
+// - config dump
+// - stats
+// - clusters
+// - listeners
+func EnvoyDumpOnFail(out io.Writer, namespace string) func() {
+	return func() {
+		setupOutDir(envoyOutDir)
+		recordEnvoyDump(fileAtPath(filepath.Join(envoyOutDir, "config.log")), "/config_dump", namespace)
```
ben-taussig-solo, line 239: can we use `getEnvoyCfgDump` and `getEnvoyStatsDump` here? How do you feel about creating `getEnvoyClustersDump` and `GetEnvoyListenersDump`?

projects/gloo/pkg/utils/ssl_test.go
davidjumani: This was a pointer to a struct but is now an interface

test/helpers/kube_dump.go
davidjumani: I did consider it, however with every dump that we require it would in turn require creating a wrapper method. Instead it would be simpler to expose the GetEnvoyDump method so the required path can be passed to fetch the relevant data

projects/gloo/cli/pkg/cmd/gateway/dump.go
davidjumani: Looks like something already exists https://github.com/solo-io/gloo/blob/main/pkg/cliutil/uri.go#L210
I'll look into using it

projects/gloo/pkg/utils/ssl.go
nfuden: Then we probably shouldnt have moved those tests if there is no reason to.

test/helpers/kube_dump.go
davidjumani: From what I see, it is passed in case it needs to be used. In enterprise, it is used to dump the kube events when tests fail. I'd prefer not to make too many changes, especially around CI right now since we're nearing a release

projects/gloo/cli/pkg/cmd/gateway/dump.go
davidjumani: Leftover method. Have cleaned it up

projects/gloo/cli/pkg/cmd/gateway/dump.go
davidjumani: Updated the name and comment

projects/gloo/pkg/utils/ssl.go
davidjumani: Without moving those tests to their own package, it caused a cyclic dependency issue. I've reverted the other changes so only the required tests have been moved to their _test module

test/helpers/kube_dump.go
davidjumani: Added it here as well as other instances when fetching the kube and docker state

projects/gloo/pkg/utils/gloo_resources_test.go
davidjumani: Updated to only move the required tests to avoid cyclic dependencies

projects/gloo/pkg/utils/log_filtering_test.go
davidjumani: Updated to only move the required tests to avoid cyclic dependencies

projects/gloo/pkg/utils/utils_test.go
davidjumani: Updated to only move the required tests to avoid cyclic dependencies

test/helpers/kube_dump.go
jbohanon: Sure. Admittedly I don't know the benefit of using the GinkgoWriter vs regular stdout. I imagine it does some smart stuff with interleaving test output, but not sure. 

Either way, if we're not using the writer, then can we be explicit about it? Using the [blank identifier](https://go.dev/ref/spec#Declarations_and_scope) immediately informs the reader not to worry about that parameter, and has the bonus effect of not creating a binding for that parameter at runtime.
```go
func EnvoyDumpOnFail(_ io.Writer, namespace string) func() {
```

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -35,10 +36,11 @@ func dumpCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra.
 	return cmd
 }
 
-func getEnvoyCfgDump(opts *options.Options) (string, error) {
+// GetEnvoyAdminData returns the response from the envoy admin interface based on the `path` specified within the defined timeout
+func GetEnvoyAdminData(ctx context.Context, proxyName, namespace, path string, timeout time.Duration) (string, error) {
```
jbohanon, line 40: nit: might be good to add a check here that the provided `path` begins with `/`.

test/helpers/kube_dump.go
davidjumani: Changes made

projects/gloo/cli/pkg/cmd/gateway/dump.go
davidjumani: Updated the docs and added a check

jbohanon (approved):
:rocket:

jbohanon (approved):

test/helpers/kube_dump.go
ben-taussig-solo: Looks great, thanks!

test/helpers/kube_dump.go
ben-taussig-solo: makes sense!

projects/gloo/pkg/utils/utils_test.go
nfuden: Soooo should we remove the .utils?

nfuden (approved):

PR #8989, opened 2023-12-15
workflows: update to have 1.16 in some of the workflows 

nfuden:
Make a placeholder for 1.17 to make sure that we dont mess up on 1.16 targeted things. 
Also update some of our workflows to be 1.16 aware

### DIFF
--- a/.github/workflows/docs-gen.yaml
+++ b/.github/workflows/docs-gen.yaml
@@ -7,6 +7,7 @@ on:
   push:
     branches:
     - 'main'
+    - 'v1.16.x'
     - 'v1.15.x'
     - 'v1.14.x'
     - 'v1.13.x'
--- a/.github/workflows/docs-gen.yaml
+++ b/.github/workflows/docs-gen.yaml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
     - 'main'
-    - 'v1.16.x'
     - 'v1.15.x'
     - 'v1.14.x'
     - 'v1.13.x'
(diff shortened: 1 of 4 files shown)


docs/active_versions.json
```diff
@@ -2,6 +2,7 @@
   "latest": "v1.15.x",
```
inFocus7, line 2: are we planning on updating this after the first non-rc release of `1.16`?

.github/workflows/docs-gen.yaml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
     - 'main'
+    - 'v1.16.x'
     - 'v1.15.x'
     - 'v1.14.x'
     - 'v1.13.x'
```
inFocus7, line 13: Will we want to remove `v1.12.x` from our docs as well?

inFocus7 (commented):
a few nit comments. 

also noting that we may want to update `DOCS-ISSUE.yml` so latest is 1.16.x, and `main` references `1.17.x beta`. although this could wait til after the official release of 1.16.0

.github/workflows/docs-gen.yaml
nfuden: I erred on the side of not removing 1.12 until we hit ga

docs/active_versions.json
nfuden: yes

.github/workflows/docs-gen.yaml
inFocus7: makes sense! 👍🏼 technically it's still in our support until 1.16's release, so we could hold off on removing it

inFocus7 (approved):

sam-heilbron (approved):

PR #9005, opened 2023-12-19
Introduce Basic Linter

sam-heilbron:
# Description

Introduce a linter, inspired by https://github.com/solo-io/gloo-mesh-enterprise/blob/main/.golangci.yaml

# Context

Introducing a linter will always be a large lift, that won't be completed in a single PR. I aim to introduce the structure of the lint logic, with minimal requirements. Hopefully this will make it easier for future devs to add incremental enhancements which overall improve the health of the codebase.

In this PR I only introduce the ginkgo linter, and then I don't even make tests required as part of CI. In essence, this does nothing. However I do it for the following reasons:
- Devs can now run `make analyze` locally and fix tests lint issues
- We will have the lint config in our CI pipeline that other devs can expand

## Auto-Fix Challenges
The `ginkgolinter` has auto-fix logic: 
```
ginkgolinter -fix ./...
```

However, when I run that locally, I get a panic:
```
panic: runtime error: index out of range [0] with length 0

goroutine 23146 [running]:
github.com/nunnatsa/ginkgolinter.checkMatchErrorAssertion(0xc008a26a40?, 0xbe405e9?)
        /Users/samheilbron/go/pkg/mod/github.com/nunnatsa/ginkgolinter@v0.15.2-0.20231218063552-f0f42939d765/ginkgo_linter.go:408 +0xaf
github.com/nunnatsa/ginkgolinter.doCheckMatchError(0xc001fae0d0, 0xc008a26a80, 0xc008a26a40, {0x13d30c0, 0xc008a269c0}, {0x13d41d0?, 0x15bfc00?}, {0xc00502c380, 0x62})
        /Users/samheilbron/go/pkg/mod/github.com/nunnatsa/ginkgolinter@v0.15.2-0.20231218063552-f0f42939d765/ginkgo_linter.go:375 +0x279
github.com/nunnatsa/ginkgolinter.checkMatchError(0x13d30c0?, 0x13d30c0?, {0x13d30c0?, 0xc008a269c0?}, {0x13d41d0?, 0x15bfc00?}, {0xc00502c380?, 0x0?})
        /Users/samheilbron/go/pkg/mod/github.com/nunnatsa/ginkgolinter@v0.15.2-0.20231218063552-f0f42939d765/ginkgo_linter.go:340 +0x65
[...]

PR #9007, opened 2023-12-20
Resolve Ginkgo Lint Errors

sam-heilbron:
# Description

Enable linting on tests as part of our CI check

# Context
We introduced a linter in: https://github.com/solo-io/gloo/pull/9005

There were known tests to not follow the lint requirements, so we ignored those in the first PR.

This PR aims to address those issues, fix the tests, and ensure that future commits all are required to fit the ginkgolinter standard.
 
## Helm Template changes
- [ ] Signed off by @sheidkamp 

Slack context: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1703096096869749

We discovered that some tests weren't forming proper assertions against our helm templates, and fixing those assertions made us aware of an inconsistency in our templates.

Specifically, the tests which asserted that by default, we set `runAsUser` to `10101`, was broken for certain deployments. We opted to solve this minimally. The solution is self-explanatory.

## Gloo Resources Test Changes
We uncovered a number of issues with these tests, beyond the fact that the assertions were doing nothing. We have updated the test, but are finding that it is still failing.

We can either:
- (A) Disable the test (current status) and merge these changes
- (B) Block this PR until the test can be fixed

_I'm leaning towards A, in the name of making incremental progress, but happy to be convinced otherwise._

# Testing steps
- [x] `make analyze` returns no issues
- [x] CI passes, including lint action

## Kube2 Local Testing
For the kube2e tests, I found it valuable to run the tests locally to verify that modifying the assertions did not break the tests. To iterate quickly, these are the steps I followed:

1. Create a kind cluster, and don't build images locally:
```
CLUSTER_NAME=lint VERSION=1.16.0-rc1 SKIP_DOCKER=true  ci/kind/setup-kind.sh
```
2. Run the tests, using a released version:
```
KUBE2E_TESTS=gateway TEAR_DOWN=true WAIT_ON_FAIL=1 RELEASED_VERSION=1.16.0-rc1 make run-kube-e2e-tests
```

Note: This works when I did not have to build any images

# Notes for reviewers
-

# Checklist:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -22,9 +22,17 @@ linters:
 linters-settings: {}
 
 run:
-  # In the past we have found performance to be better when using a subset of total CPUs.
+  # Number of CPUs to use when running golangci-lint.
+  # Default: the number of logical CPUs in the machine
   concurrency: 4
+
+  # Include test files or not.
+  # Default: true
   tests: true
+
+  # Timeout for analysis, e.g. 30s, 5m.
+  # Default: 1m
+  timeout: 5m
 
   skip-dirs:
     # don't lint ruleguard files
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -8,7 +8,7 @@
 linters:
   # Disable all linters.
   disable-all: true
-  # Enable specific linter
+  # Enable specific linters
   # https://golangci-lint.run/usage/linters/#enabled-by-default
   enable:
     # Given that we have not had linting enabled in this repository, it would be a large change to turn on all linters immediately.
@@ -28,12 +28,18 @@ issues:
 run:
   # Number of CPUs to use when running golangci-lint.
   # Default: the number of logical CPUs in the machine
+  # The GME team found performance to be better when using a subset of total CPUs, so we chose 4, like they did.
   concurrency: 4
 
   # Include test files or not.
+  # We have found that tests which break the linter are often symptomatic of a poorly written test.
+  # To avoid unintentionally introducing tests that don't do what we expect, we run the linter against them.
   tests: true
 
   # Timeout for analysis, e.g. 30s, 5m.
+  # We select 5 minutes somewhat arbitrarily.
+  # The idea is that we want a timeout that is long enough to allow the job to run.
+  # If we find that the job is timing out, we can explore ways to make this job run faster, or increase the timeout.
   timeout: 5m
 
   skip-dirs:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -42,3 +42,27 @@ run:
 
     # don't lint gomock intermediate files
     - 'gomock_reflect_\d*'
+
+# output configuration options
+output:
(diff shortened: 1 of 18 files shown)


.golangci.yaml
```diff
@@ -21,15 +20,49 @@ linters:
...
 
+issues:
+  # Maximum count of issues with the same text.
+  # Set to 0 to disable.
+  max-same-issues: 0
+
 run:
-  # In the past we have found performance to be better when using a subset of total CPUs.
+  # Number of CPUs to use when running golangci-lint.
+  # Default: the number of logical CPUs in the machine
```
nfuden, line 30: Isn't the subset thing valuable info? Should we add back in the old comment?

.golangci.yaml
```diff
@@ -21,15 +20,49 @@ linters:
...
   concurrency: 4
-  # test code shouldn't follow same rules, if we find tests are failing a lot we can revisit.
-  tests: false
+
+  # Include test files or not.
+  tests: true
+
+  # Timeout for analysis, e.g. 30s, 5m.
+  timeout: 5m
```
nfuden, line 37: Is there a reason we chose this number?

.golangci.yaml
sam-heilbron: I opted to remove it because the learning "we found better performance" is not actually based on anything we on the Edge team did. The value is the same as before, and is the default. I'm happy to re-add if you think it's valuable

.golangci.yaml
sam-heilbron: I included a comment on the github workflow for how 5 minutes was chosen. I'm happy to copy that comment here. Somewhat arbitrary value, chosen as a way to capture if/when the job takes "too long"

sam-heilbron: Pushed an update to the file to try to clarify, thanks for raising this! let me know what you think

.golangci.yaml
sam-heilbron: I put this back in, let me know what you think

changelog/v1.17.0-beta2/linter-ginkgo-p2.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/6686
+    resolvesIssue: false
+    description: >-
```
nfuden, line 5: Should we add a callous for runasuser?

nfuden: *call out

sam-heilbron: good call!

jbohanon (approved):

nfuden (approved):

jbohanon:
Bulldozer?

sheidkamp (approved):
Love to see code getting cleaned up

.golangci.yaml
jbohanon: Gonna resolve this since @nfuden approved

PR #9009, opened 2023-12-21
Vault renewal with refactor

sheidkamp:
# Description

Implement token renewal as demonstrated here: https://github.com/hashicorp/vault-examples/blob/main/examples/token-renewal/go/example.go

## API changes
Added `settings.secretOptions.sources[].vault.aws.leaseInterval` which gets used as the `Increment` argument when creating the `LifeTimewatcher` to monitor the token. If the value not defined, it will default to 0, which will be treated by Vault as the ttl of the secret.

## Code changes
Modified the `ClientAuth` interface by adding `StartRenewal`:
```
type ClientAuth interface {
	// vault.AuthMethod provides Login(ctx context.Context, client *Client) (*Secret, error)
	vault.AuthMethod
	// ManageTokenRenewal should be called after a successful login to start the renewal process
	// This method may have many different types of implementation, from just a noop to spinning up a separate go routine
	ManageTokenRenewal(ctx context.Context, client *vault.Client, secret *vault.Secret)
}

```
The renewal logic is in `projects/gloo/pkg/bootstrap/clients/vault/renewal.go`
New interfaces:
```
type TokenRenewer interface {
	StartRenewal(ctx context.Context, client *vault.Client, clientAuth ClientAuth, secret *vault.Secret) error
}
```

The `TokenRenewer` interface is used to renew tokens. We have implemented `VaultTokenRenewer` to handle the vault logic, and it can be replaced during testing. 
```
type VaultTokenRenewer struct {
	auth                     vault.AuthMethod
	leaseIncrement           int
	getWatcher               getWatcherFunc
	retryOnNonRenewableSleep int
}
```

The `TokenWatcher` interface is implemented by `vault.LifetimeWatcher`, and is broken out for testing purposes so we can send our own messages.
```
type TokenWatcher interface {
	DoneCh() <-chan error
	RenewCh() <-chan *vault.RenewOutput
        Start()
        Stop()
}
```
The TokenWatcher interface matches the shape of the LifetimeWatcher in our code and was implemented to let us switch out the implementation for testing.

We added the following metrics:

gloo_solo_io_vault_last_login_success - Time of last successful vault login
gloo_solo_io_vault_last_login_failure - Time of last failed vault login

gloo_solo_io_vault_last_renew_success - Time of last successful vault renewal
gloo_solo_io_vault_last_renew_failure - Time of last failed vault renewal - note, these are somewhat common and will happen when the token ttl has expired. This will be followed by a login attempt.

gloo_solo_io_vault_login_successes - Number of successful authentications of vault
gloo_solo_io_vault_login_failures - Number of failed authentications of vault

gloo_solo_io_vault_renew_successes - Number of successful vault renewals
gloo_solo_io_vault_renew_failures - Number of successful vault failures

## Docs changes
Updated Vault IRSA guide to include working instructions for basic setup.

# Context
When integrating with Vault users were getting 403 errors after the token lease had expired

## Testing steps
Follow the steps in https://docs.solo.io/gloo-edge/main/guides/integrations/vault/aws-irsa-auth/ using a post v1.15.3 (probably easiest to use the latest beta) release to see the 403 errors.

Repeat with the code in this branch and with the watcher increment set in settings (this shows only the new value to add, not the complete YAML block):
```
  secretOptions:
    sources:
    - vault:
        aws:
          leaseInterval: 300
```

and the errors will not occur

# Checklist:


projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -21,10 +20,10 @@ import (
 //go:generate mockgen -destination mocks/mock_auth.go -package mocks github.com/hashicorp/vault/api AuthMethod
 
 type ClientAuth interface {
+	// vault.AuthMethod provides Login(ctx context.Context, client *Client) (*Secret, error)
 	vault.AuthMethod
-	// Start Renewal should be called after a successful login to start the renewal process
-	// This method may have many different types of implementation, from just a noop to spinning up a separate go routine
-	// StartRenewal(ctx context.Context, client *vault.Client, secret *vault.Secret) error
+	// TokenRenewer provides the function: StartRenewal(ctx context.Context, client *vault.Client, secret *vault.Secret) error
+	TokenRenewer
```
sheidkamp, line 26: This is the piece I was thinking could be simplified by just inlining the function definition since we don't need to mock. I think we still want to separate out the token renewal functionality, but it doesn't need to be exposed here

projects/gloo/pkg/bootstrap/clients/vault/renewal.go
```diff
@@ -0,0 +1,184 @@
...
+		retryLogin := ctx.Err() == nil
+		return retryLogin, nil
+	}
+
+	watcher, stop, err := r.getWatcher(client, secret, watcherIncrement)
+
+	// The only errors the constructor can return are if the input parameter is nil or if the secret is nil, and we
+	// are always passing input and have validated the secret is not nil in the calling
+	if err != nil {
+		return false, ErrInitializeWatcher(err)
```
jbohanon, line 151: It looks like this will double-wrap `err` with the same wrapper. I think you could probably remove the wrapping from the `VaultGetWatcher`, wdyt?

test/e2e/vault_aws_test.go
```diff
@@ -99,16 +103,29 @@ var _ = Describe("Vault Secret Store (AWS Auth)", decorators.Vault, func() {
...
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(secret.GetOauth().GetClientSecret()).To(Equal("test"))
-			}, "5s", ".5s").Should(Succeed())
+				g.ExpectWithOffset(1, err).NotTo(HaveOccurred())
+				g.ExpectWithOffset(1, secret.GetOauth().GetClientSecret()).To(Equal("test"))
+			}
+
+			Eventually(getSecret, "5s", ".5s").Should(Succeed())
+			// Sleep and try again. There is a 10s ttl on the lease, so this should fail
+			time.Sleep(12 * time.Second)
```
jbohanon, line 124: could we reduce the TTL on the secret and this sleep any further?

projects/gloo/pkg/bootstrap/clients/vault/renewal_test.go
```diff
@@ -0,0 +1,287 @@
...
+				Auth: &vault.SecretAuth{
+					Renewable:   false,
+					ClientToken: "test-token-nonrenewable",
+				},
+				LeaseDuration: 100,
+			}
+		}
+	)
+
+	var getTestWatcher = func(client *vault.Client, secret *vault.Secret, watcherIncrement int) (TokenWatcher, func(), error) {
```
jbohanon, line 72: nit: unused parameters could have blank identifiers instead of names

jbohanon (commented):
Looks great, mostly comments about reasoning about the code

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -103,7 +112,7 @@ func (s *StaticTokenAuth) Login(ctx context.Context, _ *vault.Client) (*vault.Se
 }
 
 // NewRemoteTokenAuth is a constructor for RemoteTokenAuth
-func NewRemoteTokenAuth(authMethod vault.AuthMethod, retryOptions ...retry.Option) ClientAuth {
+func NewRemoteTokenAuth(authMethod vault.AuthMethod, t TokenRenewer, aws *v1.Settings_VaultAwsAuth, retryOptions ...retry.Option) ClientAuth {
```
sam-heilbron, line 115: The defaultRetryOptions still only use 1 attempt. I would have expected that a test failed because we should be validating that we can handle retries appropriately. Is this the default attempts that you would expect? I'm good with it if we decide 1 is the default, but it means that when we invoke `NewRemoteTokenAuth` in production, that we pass a production-ready number of retries

projects/gloo/pkg/bootstrap/clients/vault/renewal.go
```diff
@@ -0,0 +1,184 @@
...
+	"context"
+	"math/rand"
+	"time"
+
+	vault "github.com/hashicorp/vault/api"
+	"github.com/solo-io/gloo/pkg/utils"
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+type TokenWatcher interface {
```
sam-heilbron, line 13: nit: For exported types, let's include some comments about about what this is used for, and the expectations of any of the functions. This way all the logic can live on the interface, so future devs can introduce new implementations if they need to

test/e2e/vault_aws_test.go
```diff
@@ -99,16 +103,29 @@ var _ = Describe("Vault Secret Store (AWS Auth)", decorators.Vault, func() {
 		})
 
 		It("can read secret using resource client", func() {
-			Eventually(func(g Gomega) {
-				secret, err := testContext.TestClients().SecretClient.Read(
+			var (
```
sam-heilbron, line 106: Nice, I like the test! Up to you on how you want to structure the tests, though I imagine we want to test the following cases:
Positive:
- Can read with token
- Can read with token that has been renewed
- Can read with token that has exceeded its ttl, and a new token has been captured via a subsequent login
Negative:
- Can not read with bad token

projects/gloo/pkg/bootstrap/clients/vault/renewal.go
```diff
@@ -0,0 +1,184 @@
...
+	DoneCh() <-chan error
+	RenewCh() <-chan *vault.RenewOutput
+}
+
+type TokenRenewer interface {
```
sam-heilbron, line 18: nit: I don't quite follow how the TokenRenewer provides a noticeable difference from the ClientAuth. Since login and renewal are so tightly coupled (renewal relies on login) I think it makes sense to have a single interface that co-locates both. This doesn't need to be handled as part of this PR, but I just wanted to raise the comment/question

projects/gloo/pkg/bootstrap/clients/vault/client_test.go
```diff
@@ -124,41 +132,55 @@ var _ = Describe("ClientAuth", func() {
...
 
 		When("internal auth method returns an error, and then a success", func() {
+			var (
+				client *vaultapi.Client
+				err    error
+			)
 
 			BeforeEach(func() {
+				client = nil
+				err = nil
```
sam-heilbron, line 159: Is this necessary? I assumed that the variables are reset each iteration anyway

projects/gloo/pkg/bootstrap/clients/vault/renewal_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+func (t *testWatcher) DoneCh() <-chan error {
+	return t.DoneChannel
+}
+
+func (t *testWatcher) RenewCh() <-chan *vault.RenewOutput {
+	return t.RenewChannel
+}
+
+var _ = Describe("Vault Token Renewal", func() {
```
sam-heilbron, line 33: Do you think there is value in a test that cancels the context, and spins up a new context with differnet behavior? Then we can more easily catch/validate that goroutines are not leaked

sam-heilbron (commented):
Looking good! 

I tried to share thoughts on all areas of the code since I'll be out for the next two weeks. The changes look good and I'm happy with however the group decides to move forward (incorporating or ignoring my comments)

projects/gloo/pkg/bootstrap/clients/vault/renewal_test.go
sheidkamp: I don't think it adds much value here, as the behavior is all mocked by sending calls to the channels. We would essentially just be running the same test twice.

There are 2 ways we can see goroutines are no leaked:
* Some of the tests call the blocking version (`RenewToken`) of the function called by the goroutine to validate what error is returned. These rely on the context beign cancelled to return or they would hang indefinitely.
* We can remove the call to `cancel()` in one of the tests that uses the function that runs the renewal in the go routine with `ManageTokenRenewal`, for example `"Renewal should work"` and we will see that we get the error about unexpected go routines (note if you want to try this yourself: if you focus a test or group of tests, the go routine detection does not work, so you have to run the full test suite).

projects/gloo/pkg/bootstrap/clients/vault/renewal.go
```diff
@@ -0,0 +1,185 @@
...
+		retryOnNonRenewableSleep := watcherIncrement
+		defaultRetry := 60
+		if retryOnNonRenewableSleep == 0 {
+			retryOnNonRenewableSleep = defaultRetry
+		}
+
+		contextutils.LoggerFrom(ctx).Errorw("Token is not configured to be renewable.", "retry", retryOnNonRenewableSleep, "Error", err, "TokenIsRenewable", renewable)
+
+		// The units don't make sense but this is the way the docs recommend doing it
+		time.Sleep(time.Duration(retryOnNonRenewableSleep) * time.Second)
```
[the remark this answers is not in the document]
sheidkamp: I think this would be appropriate for a followup. If there is time waiting for further rounds of review, I will take a look at this, but I think the worst case is that the long running go routine sticks around for 60 seconds (the sleep time in configurable at the struct level, but hardcoded at the instantiation level).

sheidkamp (commented):
PR Updated

projects/gloo/pkg/bootstrap/clients/vault/renewal.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+
+	if err != nil {
+		return nil, nil, ErrInitializeWatcher(err)
+	}
+
+	go watcher.Start()
+	//defer watcher.Stop()
+
+	return watcher, watcher.Stop, nil
```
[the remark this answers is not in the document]
sheidkamp: I had done this to sweep the go routine out of the way for testing, but now that I have the tests setup, this is not necessary and have made this change

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -115,12 +115,21 @@ aws iam attach-role-policy --role-name $VAULT_AUTH_ROLE_NAME --policy-arn=${VAUL
...
 
+This is a basic approach that may work for you. Note that is uses dev mode and is intended for use with this guide only:
+
+```shell
+helm repo add hashicorp https://helm.releases.hashicorp.com
+helm repo update
+
+helm install vault hashicorp/vault --set "server.dev.enabled=true" --namespace vault --create-namespace
+```
+
```
jbohanon, line 132: there is a step here missing where we specify how to run the vault commands listed below. Perhaps a section or note indicating that you must exec into the vault pod to execute these commands

jbohanon: hmm actually on second thought the guide previously sets up some useful env vars that I don't think make it through to the exec shell. so maybe a link to where to get the vault client for local execution and instructions on how to wire it up to run against the vault installed in eks

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -182,40 +191,43 @@ vault write auth/aws/role/dev-role-iam \
     max_ttl=24h
 ```
 
+If this fails see [Access denied due to identity-based policies – implicit denial](#access-denied-due-to-identity-based-policies--implicit-denial)
```
jbohanon, line 194: this is a great link to the relevant section. however, that section more or less just states what the problem is, not how to fix it. A little clarity on what commands/console work needs to be done to resolve would be useful for people (like me) who largely got to this point in the guide through copypasta

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
sheidkamp: I have actually been using a combination of opening a shell on the vault pod for the first set of commands, to avoid having to send the policy file in a `kubectl exec` and then exiting and using `kubectl exec` to run the commands that rely on environment variables.

I have updated the document to reflect those steps.

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
sheidkamp: I used the console and have added the steps to create the policy.

jbohanon:
Note: to make the token lease ttl shorter for testing the feature, when following the IRSA guide and creating the role, change `max_ttl=24h` to something more workable like `max_ttl=5m`

projects/gloo/pkg/bootstrap/clients/vault/client_test.go
```diff
@@ -90,25 +98,25 @@ var _ = Describe("ClientAuth", func() {
 					},
 				}
 
-				clientAuth, _ = vault.ClientAuthFactory(vaultSettings)
+				clientAuth, _ = ClientAuthFactory(vaultSettings)
```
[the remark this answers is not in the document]
jbohanon: It looks like we are still not capturing the error here

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -251,7 +280,21 @@ Code: 400. Errors:
 
 To resolve this issue, add the `iam:GetRole` action to a policy attached to the assumed-role identity. In the previous example, you would add the `iam:GetRole` action to the identity `arn:aws:sts::account-id:assumed-role/foo-role/bar`.
 
+One way to do this is via the console:
+* Log into the AWS console and go to the IAM dashbord
```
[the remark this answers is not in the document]
sheidkamp: updated

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -57,7 +57,7 @@ cat <<EOF > trust-relationship.json
 }
 EOF
 
-export VAULT_AUTH_ROLE_NAME="dev-role-iam"
+export VAULT_AUTH_ROLE_NAME="dev-role-iam-${CLUSTER_NAME}"
 export VAULT_AUTH_ROLE_ARN=$([[ $(aws iam list-roles --query "Roles[?RoleName=='${VAULT_AUTH_ROLE_NAME}'].Arn" --output text) == "" ]] \
```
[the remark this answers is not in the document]
sheidkamp: Works as is for me, and the zsh docs (https://zsh.sourceforge.io/Doc/Release/Conditional-Expressions.html#Conditional-Expressions) say both should work inscide `[[  ... ]]` but updating here and for `VAULT_AUTH_POLICY_ARN` as well

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -175,47 +201,50 @@ vault write auth/aws/config/client \
 Finally, bind the Vault authentication and policy to your role in AWS. To use IAM roles, the following command sets the `auth_type` to `iam`.
 
 ```shell
-vault write auth/aws/role/dev-role-iam \
+kubectl -n vault exec vault-0 -- vault write auth/aws/role/dev-role-iam \
```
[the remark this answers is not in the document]
sheidkamp: Updated to use `${VAULT_AUTH_ROLE_NAME}`

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -115,13 +123,33 @@ aws iam attach-role-policy --role-name $VAULT_AUTH_ROLE_NAME --policy-arn=${VAUL
...
 ## Vault
 
-After you set up your AWS resources, you can configure Vault with AWS authentication.
+After you set up your AWS resources, you can configure Vault with AWS authentication. This guide has only been verfied with Vault installed in the same EKS cluster we created in the [AWS section](#aws).
 
 ### Step 1: Set up Vault
 
 Install Vault by choosing one of the installation methods in Vault's [Installing Vault](https://developer.hashicorp.com/vault/docs/install) documentation.
 
+This is a basic approach that may work for you. Note that is uses dev mode and is intended for use with this guide only:
```
nfuden, line 132:
```suggestion
This is a basic approach that may work for you. Note that is uses dev mode and is intended for use with this guide only. (https://developer.hashicorp.com/vault/docs/concepts/dev-server)
```

projects/gloo/pkg/bootstrap/clients/vault/auth.go
```diff
@@ -37,10 +37,13 @@ var (
 		return errors.Wrap(err, "unable to authenticate to vault")
 	}
 	ErrPartialCredentials = func(err error) error {
-		return eris.Wrap(err, "only partial credentials were provided for AWS IAM auth: ")
+		return errors.Wrap(err, "only partial credentials were provided for AWS IAM auth: ")
+	}
+	ErrAccessKeyId       = errors.New("access key id must be defined for AWS IAM auth")
```
nfuden, line 42: nit: the name of this + package name doesnt indicate why aws is involved

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -251,7 +288,21 @@ Code: 400. Errors:
 
 To resolve this issue, add the `iam:GetRole` action to a policy attached to the assumed-role identity. In the previous example, you would add the `iam:GetRole` action to the identity `arn:aws:sts::account-id:assumed-role/foo-role/bar`.
 
+One way to do this is via the console:
+* Log into the AWS console and go to the IAM dashboard
+* Search for and select the IAM user, in this example `dev-role-iam`
```
nfuden, line 293:
```suggestion
* Search for and select the properly enabled role in this example `dev-role-iam`
```

nfuden (approved):

bewebi (commented):
Having gone through the docs guide, everything works and makes sense to me
I would recommending having the docs team take a pass for style type stuff

projects/gloo/pkg/bootstrap/clients/vault/auth.go
sheidkamp: updated to include "for vault authorization"

nfuden (approved):

bewebi (approved):
Doc review of the guide for style will be a follow up

PR #9021, opened 2023-12-27
add RoutePlugins

lgadban:
* introduces RoutePlugins which will be run for each pass of route translation
* converts all existing HttpFilterPlugins to be RoutePlugins
  * simple util functions used by plugins to determine if they are applied via gwAPI route filter 

alternative approach to https://github.com/solo-io/gloo/pull/9008


projects/gateway2/translator/listener/gateway_listener_translator.go
```diff
@@ -282,7 +282,7 @@ func (mfc *httpFilterChain) translateHttpFilterChain(
 	ctx context.Context,
 	parentName string,
 	gatewayNamespace string,
-	plugins registry.HTTPFilterPluginRegistry,
+	routePlugins registry.RoutePluginRegistry,
```
ilackarms, line 285: any reason we're going with separate registries instead of just one?

lgadban: So the RoutePluginRegistry is currently the only registry as we only have route plugins.
When we add ListenerPlugins I would want to fold them into a single registry

ilackarms (approved):

PR #9025, opened 2023-12-29
generalize PluginRegistry

lgadban:


projects/gateway2/translator/plugins/headermodifier/header_modifier_plugin.go
```diff
@@ -13,13 +13,15 @@ import (
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+var _ plugins.RoutePlugin = &Plugin{}
```
lgadban, line 16: thoughts on this compile-time assertion?

I really like it for enforcing the plugin developers' intent, as with the generic registry approach, there's no compiler feedback if e.g. the API for the various plugins changes.

Given the early stages of this codebase, I could see the method signatures changing frequently, and having the compiler do the work for us of identifying breaks is really nice (I exercised it with the change of the RoutePlugin signature in this PR)

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -9,6 +9,9 @@ import (
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// Empty type for base plugins, currently no base methods.
+type Plugin = interface{}
```
ilackarms, line 13: i think i prefer type definition over alias here 
alias should really be used when we want two types to be interchangeable, type alias is not really its own type
here we actually want the plugins to be 'typed' interfaces

lgadban: makes sense, good catch!

projects/gateway2/translator/plugins/headermodifier/header_modifier_plugin.go
ilackarms: yeah i like this pattern. one thing i'd change is making plugin structs private by default. when they're exported, it should be for an explicit reason, like when another plugin wraps them
this is a strange behavior though and probably should be avoided by extracting shared code (though it's something we've done in the past)

ilackarms (approved):

PR #9026, opened 2024-01-02
Docs: support only one authorization header in the BasicAuth config

samuelvl:
Document the limitation of supporting only one authorization header in the BasicAuth config as stated in https://github.com/solo-io/gloo-mesh-enterprise/issues/13744#issuecomment-1872348959


docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto.sk.md
```diff
@@ -430,7 +430,10 @@ This is used with custom auth servers.
 ---
 ### BasicAuth
 
-
+ 
+This is the legacy/simple basic auth config. It supports the APR and SHA-1 hashing algorithms.
+When using this config only one `Authorization` header can be passed. This means that this config cannot be used in
+conjunction with other configs that rely on the `Authorization` header as well.
```
artberger, line 436: Maybe provide what they should do in case of the conflict? how about:

```suggestion
This legacy basic auth config supports the APR and SHA-1 hashing algorithms.

When using basic auth, requests can pass only one `Authorization` header. You cannot use basic auth config in conjunction with other auth configs that rely on the `Authorization` header as well. In case of such a conflict, use a different type of auth config or configure a different header, such as `Auth`.
```

artberger (approved):

k8so (approved):

PR #9030, opened 2024-01-02
AWS IRSA Vault guide for Docs review

sheidkamp:
# Description

This guide was updated with https://github.com/solo-io/gloo/pull/9009, and I don't believe it has been through a review by the docs team.

Please review the entire document instead of just the changeset.

# Updates
* Simplified the logic around the creation of the VAULT_AUTH_ROLE and VAULT_AUTH_POLICY. Previously we were checking if the role/policy existed already and returning the arn if it did.  This made the logic difficult to read and was also assuming the existing entity was properly configured for the purposes of the guide.
* Updated the troubleshooting section to use the command line instead of the console.

### DIFF
--- a/docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
+++ b/docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
@@ -28,12 +28,12 @@ export ACCOUNT_ID=$(aws sts get-caller-identity --query "Account" --output text)
 
 This environment variable can only be set after setup
 ```shell
-export OIDC_PROVIDER=$(aws eks describe-cluster --name $CLUSTER_NAME --region $AWS_REGION --query "cluster.identity.oidc.issuer" --output text | sed -e "s/^https:\/\///")
+export OIDC_PROVIDER=$(aws eks describe-cluster --name ${CLUSTER_NAME} --region ${AWS_REGION} --query "cluster.identity.oidc.issuer" --output text | sed -e "s/^https:\/\///")
 ```
 
 At this point double check that you have OIDC provider associated with the cluster. The output should be either match `created IAM Open ID Connect provider for cluster ` or `IAM Open ID Connect provider is already associated with cluster`
 ```shell
-eksctl utils associate-iam-oidc-provider --cluster $CLUSTER_NAME --approve
+eksctl utils associate-iam-oidc-provider --cluster ${CLUSTER_NAME} --approve
 ```
 
 ### Step 2: Set up a Role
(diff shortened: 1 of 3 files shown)


Rachael-Graham:
Taking a look!

Rachael-Graham:
@sheidkamp Okay, I committed my suggestions for docs. Please take a look and LMK if anything is amiss!

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -15,30 +15,29 @@ Start by creating the necessary permissions in AWS.
 
 ### Step 1: Create a cluster with OIDC
 
-To configure IRSA, create or use an EKS cluster with an associated OpenID Provider (OIDC). For setup instructions, follow [Creating an IAM OIDC provider for your cluster](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html) in the AWS docs. Only the first page of the linked guide is neccessary.
+To configure IRSA, create or use an EKS cluster with an associated OpenID Provider (OIDC). Be sure to use a cluster that you have not yet installed Gloo Edge into.
```
sheidkamp, line 18:
```suggestion
To configure IRSA, create or use an EKS cluster with an associated OpenID Provider (OIDC). Be sure to use a cluster name that has not previously been used to create the AWS vault permissions in this guide.
```

The issue isn't if the cluster has Gloo Edge installed (it's easy to uninstall), the issue is the names of the AWS entities created, which are based on the clustername. So the following situation would cause problems:

1) Create a cluster named "gloo-vault"
2) Run through this guide
3) Delete the cluster (the roles and policies do not automatically get cleaned up)
4) Create another cluster named "gloo-vault"


Though this cluster is brand new does not have gloo installed, the AWS namespace has been polluted.  We should make sure to be cleat that this is the issue

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
```diff
@@ -286,23 +290,60 @@ Code: 400. Errors:
...
-
-
-## Summary
-
-Now, Gloo Edge securely accesses Vault secrets using temporary credentials obtained through AWS IAM Roles for Service Accounts (IRSA).
-This enhances security, streamlines access control, and simplifies authorization within your Kubernetes environment.
\ No newline at end of file
+To create and associate the necessary policy:
+
+1. Set an environment variable with the assumed role. You can find the value in your error message.
```
sheidkamp, line 295:
```suggestion
1. Set an environment variable with the assumed role. You can find the value in your error message. In the example above, the `<role-name>` would be `foo-role`
```

I think its important somewhere to explicitly state the string that would be used for <role> in the example. There string in the error in the example is arn:aws:sts::account-id:assumed-role/foo-role/bar and the variable should only be set to foo-role (not the complete string, don't include the /bar). The real string will look something like: arn:aws:sts::802411188784:assumed-role/eksctl-username-clustername-nodegrou-NodeInstanceRole-toYayfo5xAZz/i-011db30b6bff745d8

docs/content/guides/integrations/vault/aws-irsa-auth/_index.md
sheidkamp, line 18: Should probably have a link for this up here too: https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html

sheidkamp (commented):
Thank you, It looks so much better! Just a few notes.

Rachael-Graham (approved):

bewebi (approved):

PR #9031, opened 2024-01-03
fix: slow start aggression

sopida-chotwanwirach:
# Description

The aggression setting in slow start config is missing a required runtime key as specified in envoy [config.core.v3.RuntimeDouble](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-runtimedouble) documentation. The runtime_key is governed by the following validation rule.
```
string runtime_key = 2 [(validate.rules).string = {min_len: 1}];
```

Even though the runtime key is required for RuntimeDouble type, this field is not in use as slow start aggression is not supported in [envoy runtime configuration](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_runtime#config-cluster-manager-cluster-runtime). 

## Code changes
- Add a placeholder runtime key to satisfy the requirement for RuntimeDouble

# Context

I ran into the bug when creating an upstream with slow start mode.
```
kind: Upstream
spec:
  loadBalancerConfig:
    roundRobin:
      slowStartConfig:
        aggression: 2
        minWeightPercent: 20
        slowStartWindow: 60s
```
```
(ClusterValidationError.RoundRobinLbConfig: embedded message failed validation | 
caused by RoundRobinLbConfigValidationError.SlowStartConfig: embedded message failed validation |
caused by SlowStartConfigValidationError.Aggression: embedded message failed validation |
caused by RuntimeDoubleValidationError.RuntimeKey: value length must be at least 1 runes)
```

## Testing steps

The bug is reproducible with the unit tests added in the PR.
```
 [FAILED] Unexpected error:
      <clusterv3.Cluster_SlowStartConfigValidationError>:
      invalid Cluster_SlowStartConfig.Aggression: embedded message failed validation | caused by: invalid RuntimeDouble.RuntimeKey: value length must be at least 1 runes
      {
          field: "Aggression",
          reason: "embedded message failed validation",
          cause: <corev3.RuntimeDoubleValidationError>{
              field: "RuntimeKey",
              reason: "value length must be at least 1 runes",
              cause: nil,
              key: false,
          },
          key: false,
      }
```
After the fix, the error goes away.

# Checklist:


nfuden:
Sorry for not seeing this earlier. 
Whats the though on slowstart being the same across clusters?

Currently I think it may make more sense to have slowstart be a per cluster naming like seen here https://github.com/solo-io/gloo/pull/8919/commits/ab61e954d57be194ef4125b5d752a186d770a7e4

What do you think?

sopida-chotwanwirach:
Thank you for reviewing the PR @nfuden!

It makes sense to make the aggression configurable per cluster. I can add that change in.
For the odd case where the cluster name is missing, should the runtime key fallback to something like a global key so that the flow is: `upstream.<cluster_name>.slowstart.aggression` -> `upstream.slowstart.aggression` -> default value?

nfuden:
Yeah I think that makes sense. 
I figure that if the cluster name somehow doesnt exist you probably still want to be able to toggle it and it can be painful in larger deployments to extract small bits of configuration to determine what key to use and a random key could cause configuration churn

jenshu (approved):
thanks for fixing!

nfuden (approved):

PR #9040, opened 2024-01-08
add RouteOption route plugin

lgadban:
Add support for watching and translating `RouteOptions`.

Controller now watches `RouteOptions` similar to other API types and will `Kick()` the reconciler when an event happens. I explored trying to make the controller updates generic, but there didn't seem like a straightforward way to do so that wasn't contrived. Given that the only other type which will follow this specific model is `VHostOptions`, while the rest of the policies will come from the Platform APIs, this method of adding to the controller seems to make sense to me.

New `RouteOptions` plugin added, along with util methods to handle the `ExtensionRef` filter along with the actual retrieval of the referenced object, which ultimately happens inside query engine.

Original PR: https://github.com/solo-io/gloo/pull/8984

### DIFF
--- a/projects/gateway2/controller/controller.go
+++ b/projects/gateway2/controller/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	sologatewayv1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1/kube/apis/gateway.solo.io/v1"
 	"github.com/solo-io/gloo/projects/gateway2/deployer"
 	"github.com/solo-io/gloo/projects/gateway2/query"
 	corev1 "k8s.io/api/core/v1"
@@ -52,6 +53,7 @@ func NewBaseGatewayController(ctx context.Context, cfg GatewayConfig) error {
 		controllerBuilder.watchHttpRoute,
 		controllerBuilder.watchReferenceGrant,
 		controllerBuilder.watchNamespaces,
+		controllerBuilder.watchRouteOptions,
 		controllerBuilder.addIndexes,
 	)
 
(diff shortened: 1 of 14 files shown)


projects/gateway2/controller/scheme/scheme.go
```diff
@@ -13,7 +14,7 @@ import (
 func NewScheme() *runtime.Scheme {
 	scheme := runtime.NewScheme()
 	for _, f := range []func(*runtime.Scheme) error{
-		apiv1.AddToScheme, apiv1beta1.AddToScheme, corev1.AddToScheme, appsv1.AddToScheme,
```
ilackarms, line 16: this is one of those arcane things it's good to track via codegen or something else

projects/gateway2/controller/controller.go
```diff
@@ -52,6 +53,7 @@ func NewBaseGatewayController(ctx context.Context, cfg GatewayConfig) error {
 		controllerBuilder.watchHttpRoute,
 		controllerBuilder.watchReferenceGrant,
 		controllerBuilder.watchNamespaces,
+		controllerBuilder.watchRouteOptions,
 		controllerBuilder.addIndexes,
```
ilackarms, line 57: this is something that is likely to become arcane if we don't come up with a consistent, top-level way of keeping this in sync

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -0,0 +1,84 @@
...
+
+func formatNotFoundMessage(routeCtx *plugins.RouteContext, filter *gwv1.HTTPRouteFilter) string {
+	return fmt.Sprintf(
+		"extensionRef '%s' of type %s.%s in namespace '%s' not found",
+		filter.ExtensionRef.Group,
+		filter.ExtensionRef.Kind,
+		filter.ExtensionRef.Name,
+		routeCtx.Route.GetNamespace(),
+	)
+}
```
ilackarms, line 84: could see this living in a util package

lgadban: I initially agreed, but after working through it, this is actually only relevant for extensionRef types and this is likely going to be the only extensionRef-type plugin we support, as the rest of the policies will be implemented via policy attachment (i.e. `targetRef`).

Not sure if makes sense to extract it to a util package if it won't be used anywhere else.
What do you think?

ilackarms (approved):

projects/gateway2/controller/controller.go
lgadban: adding for posterity, this was discussed offline -- we will not implement anything to keep this consistent as part of this work, but we will keep an eye on it as the codebase progresses

PR #9051, opened 2024-01-09
Add unparam linter

jbohanon:
# Description

Add `unparam` linter to guard against unintended behavior with return vals that are always the same and increase readability by using blank identifiers for unused aparmeters


install/test/helm_suite_test.go
```diff
@@ -373,6 +373,7 @@ func makeUnstructured(yam string) *unstructured.Unstructured {
 	return runtimeObj.(*unstructured.Unstructured)
 }
 
+//nolint:unparam // TODO(jbohanon) unravel usage and see if values can be replaces with `gloo-system`
```
jenshu, line 376: do you want to address this todo before merging?

jbohanon: No I think not. As I got further in, there were several spots where it was simpler/more correct to add `nolint` so I should just remove the TODO

sam-heilbron (approved):
lgtm! one question/nit

test/kube2e/helm/helm_test.go
```diff
@@ -661,7 +661,7 @@ func installGloo(testHelper *helper.SoloTestHelper, chartUri string, fromRelease
 // CRDs are applied to a cluster when performing a `helm install` operation
 // However, `helm upgrade` intentionally does not apply CRDs (https://helm.sh/docs/topics/charts/#limitations-on-crds)
 // Before performing the upgrade, we must manually apply any CRDs that were introduced since v1.9.0
-func upgradeCrds(testHelper *helper.SoloTestHelper, fromRelease string, crdDir string) {
+func upgradeCrds(_ *helper.SoloTestHelper, fromRelease string, crdDir string) {
```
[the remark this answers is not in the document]
jbohanon: Yeah there were a lot of unexported functions whose signatures could be improved. I could make another pass and try to remove as many of the blank identifiers as possible

projects/gateway/pkg/translator/aggregate_translator.go
```diff
@@ -321,7 +321,6 @@ func (a *AggregateTranslator) processMatchableHttpGateway(
 // For example in hybrid a similar function is called computeMatchedTcpListener
 func (a *AggregateTranslator) processMatchableTcpGateway(
 	params Params,
-	proxyName string,
```
nfuden, line 324: If we are removing this can we either add a TODO or explain why we dont think we will need this variable?

jbohanon: It's removed because it wasn't used in this function. I assume the function was copied from `processMatchableHttpGateway` which does use it to call the virtual service translator. I don't see why we need to document why we're removing an unused parameter from an unexported function that is not otherwise contractually required (e.g. to match a func type signature)

sam-heilbron (approved):

projects/gateway/pkg/translator/aggregate_translator.go
nfuden: Yeah im not agains the actual removal, more of a: if we can see what the gap is by not having this function as it was straightlifted from a similar function then it may be good to clean up from the old messy update.

That being said this doesnt have to be a blocking thing and we can look more into it later

Issue #9033, opened 2024-01-05
gloo install fails with gloo-resource-rollout job not found error, using helm + tilt


jenshu:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

v1.15.9

### Kubernetes Version

unknown

### Describe the bug

Customer is using helm and tilt to install GlooEE v1.15.9 (which uses Gloo OSS v1.15.17) and running into error:
```
Running cmd: ['helm', 'get', 'manifest', '--namespace', 'api-gateway', 'api-gateway']
Running cmd: ['kubectl', 'get', '-oyaml', '-f', '-']
Error from server (NotFound): jobs.batch "gloo-resource-rollout" not found
```
The error is thrown from helm-apply-helper [here](https://github.com/tilt-dev/tilt-extensions/blob/8381f694df7210551d523274f6336000ae934952/helm_resource/helm-apply-helper.py#L82)

In this version of GlooEE we have removed the helm hooks that used to be in the rollout job, and customer believes that is causing an ordering issue when trying to apply the resources. The rollout job and associated rbac resources did not get created.

As a workaround, they added back the helm hooks that were present in Gloo OSS v1.15.15 and the install succeeded.

### Expected Behavior

Install should succeed without workarounds.

### Steps to reproduce the bug

See above and check with customer.

nfuden:
https://docs.tilt.dev/helm.html#pros-and-cons-of-helm

So basically if we want to support this behavior in Tilt we need to not use hooks 

Before digging in to far, should this be an install doc with a work around? Or is there a well known tilt startegy for handling hooks in general?

davidjumani:
Have created a [test release](https://github.com/solo-io/solo-projects/actions/runs/7493661034) and awaiting user confirmation

davidjumani:
Fixed in OSS [v1.17.0-beta5](https://github.com/solo-io/gloo/releases/tag/v1.17.0-beta5), [v1.16.3](https://github.com/solo-io/gloo/releases/tag/v1.16.3) && [v1.15.22](https://github.com/solo-io/gloo/releases/tag/v1.15.22)
This will be in Enterprise 1.17.0, 1.16.1, && 1.15.12

PR #9094, opened 2024-01-24
move resource rollout job cleanup to pre-upgrade hook

davidjumani:
# Description

Tilt supports Helm out-of-the-box via the helm and helm_resource [extension](https://docs.tilt.dev/helm.html#installing-existing-charts). The way helm_resource works is as follows
- It deploys the specified helm chart via helm upgrade --install
- Once installed, it verifies that all the resources in helm get manifest exist

Since the resource-rollout-job is no longer a hook, it needs to be deleted if it exists prior to an upgrade / install since it will not be re-created if it already exists (the default behaviour of helm). This is done as part of the [resource-rollout-check-job](https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/5-resource-rollout-check-job.yaml#L104) as a post-install hook for simplicity.
As a consequence, when tilt executes helm get manifests as part of deploying a helm chart, the resource-rollout-job manifest is generated but not found in kubernetes and causes tilt deployments to fail.

This fix moves the deletion of the resource rollout job to a pre upgrade hook. Thus the rollout job exists when tilt checks all the resources after installation / upgrade

Steps to reproduce the issue
- Setup your local cluster `./ci/kind/setup-kind.sh`
- [Install tilt](https://docs.tilt.dev/install.html#macos)
- Create the following `Tiltfile`
  ```
  # -*- mode: Python -*-
  update_settings( k8s_upsert_timeout_secs = 600)
  load('ext://helm_resource', 'helm_resource')
  helm_resource(name='gloo', chart="install/helm/gloo", flags=["--values=./test/kube2e/helm/artifacts/helm.yaml", "--set=gatewayProxies.gatewayProxy.service.type=ClusterIP"])
  ```
- Run `tilt up`
  In the web UI you should see the following error :
  ```
       Error from server (NotFound): jobs.batch "gloo-resource-rollout" not found
       Traceback (most recent call last):
         File "/Users/djumani/Library/Application Support/tilt-dev/tilt_modules/github.com/tilt-dev/tilt-extensions/helm_resource/helm-apply-helper.py", line 82, in <module>
           completed.check_returncode()
         File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 460, in check_returncode
           raise CalledProcessError(self.returncode, self.args, self.stdout,
       subprocess.CalledProcessError: Command '['kubectl', 'get', '-oyaml', '-f', '-']' returned non-zero exit status 1.
  ```

## API changes
N/A

## Code changes
N/A

## CI changes
N/A

## Docs changes
N/A

# Context
https://github.com/solo-io/gloo/issues/9033

## Interesting decisions
N/A

## Testing steps
Following the steps mentioned in the description :
### Clean install
- Run `tilt down` to clear the prior installation 
- Run `tilt up`
  The deployment should succeed

### Upgrade
Modify the Tiltfile by changing some random helm value (via the flag parameter in the tiltfile)
eg: (removing the `--set=gatewayProxies.gatewayProxy.service.type=ClusterIP` helm flag
```
helm_resource(name='gloo', chart="install/helm/gloo", flags=["--values=./test/kube2e/helm/artifacts/helm.yaml"])
```
Tilt should automatically detect the changes and upgrade gloo

## Notes for reviewers
N/A

N/A

# Checklist:


nfuden:
Can we include a test to show that this works to upgrade when mtls and validation is configured?

davidjumani:
> Can we include a test to show that this works to upgrade when mtls and validation is configured? Or at least a manual test?

Since that depends on the helm values passed, they can be added to the value files and tested via the same steps

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
```diff
@@ -0,0 +1,154 @@
...
+kind: Job
+metadata:
+  labels:
+    app: gloo
+    gloo: resource-rollout
+  name: gloo-resource-rollout-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "5" # run this job after the role/rolebinding is created
```
ben-taussig-solo, line 16: What is the significance of using 5 as the hook weight? When is this job run in relation to other pre-upgrade jobs?

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
```diff
@@ -0,0 +1,154 @@
...
+    {{- include "gloo.jobHelmDeletePolicySucceededAndBeforeCreation" .Values.gateway.rolloutJob | nindent 4 }}
```
ben-taussig-solo, line 17: I see where this is defined (https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/_helpers.tpl#L88-L97) but I'm having trouble understanding what this does in context. It looks like depending on the values of some helm parameters, we will set one of the following annotations:
 - `"helm.sh/hook-delete-policy": before-hook-creation`
 - `"helm.sh/hook-delete-policy": hook-succeeded,before-hook-creation`

In either case, what does this mean for this job?

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
```diff
@@ -0,0 +1,154 @@
...
+spec:
+  {{- include "gloo.jobSpecStandardFields" .Values.gateway.rolloutJob | nindent 2 -}}
+  template:
+    metadata:
+      labels:
+        gloo: resource-rollout
```
ben-taussig-solo, line 23: Should this be changed to `resource-rollout-cleanup`? I'm not sure if it matters but I want to be careful that we don't interact negatively with the original resource rollout job

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
```diff
@@ -0,0 +1,154 @@
...
+  labels:
+    app: gloo
+    gloo: rbac
+  name: gloo-resource-rollout-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight":  "0"
+---
+kind: Role
```
ben-taussig-solo, line 81: Some questions on the RBAC stuff at the end of this file:
 - I think the rules might be more permissive than necessary -- it seems like we only need to be able to get and delete a specific job
 - Do we need all of the different RBAC resources (Role, RoleBinding, ClusterRole, ClusterRoleBinding)? Can we get away with just specifying one or two of these?

ben-taussig-solo (commented):
Is there documentation on jobs created by Gloo Edge, or about how Gloo handles pre/post install? If so, can we confirm that that documentation is still correct?

ben-taussig-solo (commented):
What is the burden for testing this work? Do we have the infrastructure to easily create tests which repeatedly install and upgrade Gloo Edge? Testing would give me increased confidence that this will not interact negatively with the original resource rollout check job

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: This is to ensure that it is run after the RBAC is created in https://github.com/solo-io/gloo/pull/9094/files/8a562adda5662b2bc27ff66fe57e2d7f71fac1d3#diff-eab8ec55d49121923be89c2d035e626273c3445f3841ea23d7dde578c3ea00a2R68-R154 as they have a weight of 0

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: It just defines when to remove the job if it already exists. Ie: whether to remove it it up before the hook is created and / or after it succeeds. It should have no effect on this job as the default hook-delete-policy is before-hook-creation and the k8s jobs have a ttl

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: Good point. Let me make the changes

davidjumani: From looking at the resource rollout job RBAC, it does have roles as well as cluster roles. I haven't dug into why the need for both but for safety I'd prefer to keep both here to mimic it

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: I'd prefer to keep it this way so all the resource-rollout job related resources have the same label - Ref the [resource-rollout-check](https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/5-resource-rollout-check-job.yaml#L23) job as well

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
jenshu: i don't think the clusterrole/clusterrolebinding below are needed, they seem to just be providing cluster-wide access to get/list/delete jobs, which is not needed since you already have the namespace-specific Role here

davidjumani:
> Is there documentation on jobs created by Gloo Edge, or about how Gloo handles pre/post install? If so, can we confirm that that documentation is still correct?

I haven't found any documentation but can create one as part of this fix

> How much work is it to test this change? Do we have the infrastructure to easily create tests which repeatedly install and upgrade Gloo Edge? Testing would give me increased confidence that this will not interact negatively with the original resource rollout check job

Do you mean e2e tests ? The upgrade tests should verify that it does work. Local testing also can provide validation as I do not think a dedicated e2e test for tilt is necessary as tilt is used mostly for local development

Upgrade Tests :
```
Ran 8 of 8 Specs in 660.848 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

ArgoCD Tests :
```
Ran 1 of 1 Specs in 381.899 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

```

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
davidjumani: Makes sense, Removed

ben-taussig-solo: Looks good to me, resolving this thread

install/helm/gloo/templates/5-resource-rollout-cleanup-job.yaml
ben-taussig-solo: Makes sense -- resolving this thread

changelog/v1.17.0-beta5/move-resource-rollout-cleanup.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/gloo/issues/9033
+  resolvesIssue: false
+  description: When deploying edge with tile + helm, the resource-rollout-job is deleted before the install completes which causes the deployment to fail. The cleanup for the resource-rollout job is now moved to a pre-upgrade hook to ensure that the job exists so deployments now succeed.
```
bewebi, line 5:
```suggestion
  description: When deploying edge with tilt + helm, the resource-rollout-job is deleted before the install completes which causes the deployment to fail. The cleanup for the resource-rollout job is now moved to a pre-upgrade hook to ensure that the job exists so deployments now succeed.
```

bewebi (approved):
I tried it out locally in addition to reviewing the changes and LGTM

Thanks for introducing me/us to Tilt as well!

nfuden (approved):
Tested with a simple tilt install and upgrades locally via helm in kind.  Seems to work as expected

Issue #8931, opened 2023-11-26
Gloo validating webhook blocks deletion of unrelated secrets


shahar-h:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.15.16

### Kubernetes Version

v1.27.3

### Describe the bug

When an upstream with a bad state exists in cluster no secret can be deleted - even secrets that are not referenced by Gloo Edge resources.

### Expected Behavior

We need to be able to delete secrets that are not referenced by Gloo Edge resources.

### Steps to reproduce the bug

1. Create a kind cluster:
```bash
kind create cluster --image kindest/node:v1.27.3
```
2. Install Gloo Edge chart with below values:
```bash
cat <<EOF | helm upgrade --install gloo gloo/gloo --version 1.15.16 -n gloo-system --create-namespace --wait --values -
gatewayProxies:
  gatewayProxy:
    service:
      type: NodePort
      httpPort: 31500
      httpsPort: 32500
      httpNodePort: 31500
      httpsNodePort: 32500
gateway:
  validation:
    allowWarnings: true
    alwaysAcceptResources: false
    failurePolicy: Fail
EOF
```
3. Create an upstream with non-exisitng service:
```bash
cat <<EOF | kubectl create -f -
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: my-us
  namespace: gloo-system
spec:
  kube:
    serviceName: my-svc
    serviceNamespace: gloo-system
    servicePort: 18081
    serviceSpec:
      grpc: {}
EOF
```
4. Create a virtualService that refers the previously created upstream:
```bash
cat <<EOF | kubectl create -f -
apiVersion: gateway.solo.io/v1
kind: VirtualService
metadata:
  name: my-vs
spec:
  virtualHost:
    domains:
    - valid.local
    options:
    routes:
    - matchers:
        - prefix: /
      routeAction:
        single:
          upstream:
            name: my-us
            namespace: gloo-system
EOF
```
Upstream status should have a warning:
```yaml
status:
  statuses:
    gloo-system:
      reason: "warning: \n  1 error occurred:\n\t* Upstream name:\"my-us\" namespace:\"gloo-system\"
        references the service \"my-svc\" which does not exist in namespace \"gloo-system\"\n\n"
      reportedBy: gloo
      state: Warning
```
5. Edit the default settings resource(gloo-system/default) and change `spec.gateway.validation.allowWarnings` to `false`.
6. Create some secret: 
```bash
[...]

PR #9098, opened 2024-01-25
Sheidkamp/secret delete retranslate - for solo-projects ingestion

sheidkamp:
# Description

## API changes
None

## Code changes
- Add `Bar()` function
- ...

## CI changes
None

## Docs changes
- ...

# Context

## Interesting decisions
 
While the client request was specifically around the deletion of secrets, modification of resources with `allow_warnings=false` when there are already errors is in general an area where logic may be updated further. If there is more than one warning, it is impossible to make updates to address these issues, which is unfriendly UX

Because of this, the validation code has been written to handle any type of resource and updates/inserts in addition to deletes. The function `shouldRetryValidationOnErrors` was created to handle the logic determining which resources and operations and is currently only enabling this validation approach for the deletion of secrets.

* Gating new functionality behind an environment variable
This feature will be enabled by default for 1.17, and we have added the ability to disable it with an environment variable flag. We took this approach because the new functionality adds additional rounds of validation and CPU load when modifying resources when `allow_warnings` is `false`. An environment variable was used instead of an API field so if/when this gating is removed, we will not be left with a deprecated field in the API. This additional work comes in two ways:
  * When `allow_warnings` is false and a potentially updatable-despite-warnings resource is being modified, we will collect all errors instead of stopping after the first.
  * When `allow_warnings` is false, a potentially updatable-despite-warnings resource is being modified, and there are warnings in the validation, we will run another round of validation to 

## Testing steps

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8931


sheidkamp:
Moved to a different branch

PR #9115, opened 2024-01-29
[DNM] mod: upgrade go-control-plane (minus ext_proc)

sam-heilbron:
# Description

This demonstrates the impact of upgrading the go-control-plane used by the Gloo control plane. 

We are moving off our fork in this PR _only_ to demonstrate that there are no other breaking changes which we need to account for. This will not merge.

# Context

Slack: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1706540297508999
Enterprise changes: https://github.com/solo-io/solo-projects/pull/5729

## Testing steps
- CI passes

## Notes for reviewers
-

# Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #9116, opened 2024-01-29
mod: upgrade go-control-plane-fork-v2

sam-heilbron:
# Description

Upgrade the go.mod dependency to rely on a later version of the go-control-plane-fork-v2

# Context

## Work in Progress
This PR can only merge after ALL of the following pieces of work are completed:
- [ ] https://github.com/solo-io/gloo/pull/9115 is confirmed to work, and PR closed
- [ ] https://github.com/solo-io/solo-projects/pull/5729 is confirmed to work, and PR closed
- [ ] https://github.com/solo-io/go-control-plane-fork-v2/pull/3 is merged and published (tagged)
- [ ] https://github.com/solo-io/solo-projects/pull/5725 merges and updates a number of dependencies
- [ ] https://github.com/solo-io/solo-projects/pull/5728 imports this PR, and passes CI, demonstrating that this works

## Interesting decisions
 -

## Testing steps
- CI passes on a number of PRs, linked above

## Notes for reviewers
- Slack context: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1706540297508999

# Checklist:


docs/content/static/content/osa_provided.md
```diff
@@ -7,7 +7,7 @@ Name|Version|License
 [census-instrumentation/opencensus-proto](https://github.com/census-instrumentation/opencensus-proto)|v0.2.0|Apache License 2.0
 [xds/go](https://github.com/cncf/xds)|v0.0.0-20230607035331-e9ce68804cb4|Apache License 2.0
 [cratonica/2goarray](https://github.com/cratonica/2goarray)|v0.0.0-20190331194516-514510793eaa|MIT License
-[envoyproxy/go-control-plane](https://github.com/envoyproxy/go-control-plane)|v0.0.0-20231026174848-be784d54c527|Apache License 2.0
+[envoyproxy/go-control-plane](https://github.com/envoyproxy/go-control-plane)|v1.29.0-patch1|Apache License 2.0
```
ben-taussig-solo, line 10: Nit: I don't think we need to fix this here, but this is incorrect. It seems like our code generation is not handling replaces properly

ben-taussig-solo (approved):

ashishb-solo (commented):
Seems fine to me, but I'll let @ben-taussig-solo and @jbohanon approve since they have outstanding comments on this pull request

docs/content/static/content/osa_provided.md
sam-heilbron: Good call! Want to write up an issue to track that? That way at least we're aware of it

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/census-instrumentation/opencensus-proto v0.4.1
 	github.com/cncf/xds/go v0.0.0-20230607035331-e9ce68804cb4
 	github.com/cratonica/2goarray v0.0.0-20190331194516-514510793eaa
-	github.com/envoyproxy/go-control-plane v0.11.1
+	github.com/envoyproxy/go-control-plane v0.12.0
```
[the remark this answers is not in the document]
jbohanon: https://go.dev/ref/mod#minimal-version-selection outlines how, with a replace directive or without, by specifying a later version of the go-control-plane in solo-projects, all assets built will rely on that version (in this case, the portal image , whose code will be imported from the gme repo).

My concern is that because the code in the dependency was written against a different, older dependency, we're kinda introducing undefined behavior by having it run code from the newer dependency. As long as it compiles, anything else will be missed. This highlights the need for good, thorough tests of features which require this dependency to make sure that everything works as expected with the version in our `go.mod`. Also,  if/when we unpin off of our fork, it's not going to matter because we will likely need a newer version than our GME dependency anyway.

Another reason is that it makes it hard to know in the future why we were on that version and if it is safe to update it, if so what is it safe to update to... etc. Whereas just plopping in a comment with `// matches GME's version` and trusting that `go mod tidy` will bump it whenever we consume a newer version of GME takes away some mental overhead. However, this dep bump should be handled by `go mod tidy` if our GME dep's `go-control-plane` version changes to one newer than we have specified.

jbohanon (approved):

Issue #9118, opened 2024-01-30
Gateway secret validation flake on k8s 1.24


bewebi:
### Which tests failed?

```
Kube2e: gateway Validation Configuration when allowWarnings=false secret validation [It] should act as expected with secret validation
```

### Initial Investigation

There is a race condition where the test makes assertions before we can be sure that the condition is met

I have [a branch](https://github.com/solo-io/gloo/compare/gateway-validation-flake-fix#diff-909288de4ab619d47d75a89fc9d6f4acfbd645ec16063a9cad8dcde18d591762R2145) where the test passes consistently due to adding an additional check that resources are accepted before starting and wrapping the final assertion in an `Eventually`

### Additional Information

The test is only run on k8s 1.24 on nightly runs and has failed repeatedly lately
The most recent failure is [here](https://github.com/solo-io/gloo/actions/runs/7706526122/job/21002249112)

PR #9119, opened 2024-01-30
Gateway validation flake fix

bewebi:
# Description

- Add a wait for resources to be accepted and wrap an assertion in an `Eventually` to resolve a test flake

# Context

This test has been failing often on k8s 1.24 on nightlies
I was also able to reproduce this locally

It's not clear why it fails in k8s 1.24 but not 1.28 or what caused the test to start failing per se

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9118

### DIFF
--- a/test/kube2e/gateway/gateway_test.go
+++ b/test/kube2e/gateway/gateway_test.go
@@ -2112,6 +2112,11 @@ spec:
 				// It isn't until later - either a few minutes and/or after forcing an update by updating the VS - that the error status appears.
 				// The reason is still unknown, so we retry on flakes in the meantime.
 				It("should act as expected with secret validation", FlakeAttempts(3), func() {
+					By("waiting for the modified VS to be accepted")
+					helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
+						return resourceClientset.VirtualServiceClient().Read(testHelper.InstallNamespace, testRunnerVs.GetMetadata().GetName(), clients.ReadOpts{Ctx: ctx})
+					})
+
 					By("failing to delete a secret that is in use")
 					err := resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
 					Expect(err).To(HaveOccurred())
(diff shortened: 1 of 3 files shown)


test/kube2e/gateway/gateway_test.go
```diff
@@ -2138,8 +2143,9 @@ spec:
 					})
 
 					// Although these tests delete the secret handled by our SnapshotWriter, because we set `IgnoreNotFound` when deleting snapshot resources, this won't cause an issue.
-					err = resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
-					Expect(err).NotTo(HaveOccurred())
+					Eventually(func() error {
+						return resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+					}).WithPolling(50 * time.Millisecond).WithTimeout(30 * time.Second).ShouldNot(HaveOccurred())
```
ashishb-solo, line 2148: nit: 50ms seems to be a really fast polling internal

test/kube2e/gateway/gateway_test.go
```diff
@@ -2112,6 +2112,11 @@ spec:
 				// It isn't until later - either a few minutes and/or after forcing an update by updating the VS - that the error status appears.
 				// The reason is still unknown, so we retry on flakes in the meantime.
 				It("should act as expected with secret validation", FlakeAttempts(3), func() {
```
ashishb-solo, line 2114: do you think we should try to remove `FlakeAttempts()`?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2112,6 +2112,11 @@ spec:
 				// It isn't until later - either a few minutes and/or after forcing an update by updating the VS - that the error status appears.
 				// The reason is still unknown, so we retry on flakes in the meantime.
```
ashishb-solo, line 2113: could you update this comment?

ashishb-solo (changes requested):
looks great! just a few minor things that might need to be addressed

test/kube2e/gateway/gateway_test.go
bewebi: The test still flakes, likely for the reason described in the comment
Running locally, it still requires multiple attempts just under 14% of the time, never requiring more than two in my runs

test/kube2e/gateway/gateway_test.go
bewebi: I'm not sure that this PR is related to the comment, and I _think_ it is still accurate

test/kube2e/gateway/gateway_test.go
bewebi: What would you find more reasonable?

ashishb-solo: i just thought that .05 seconds/check with a 30 second timeout seemed a little excessive (600 total checks). maybe .5 seconds instead? but regardless, i think it's not a big deal unless there are other changes to push

ashishb-solo (approved):

ashishb-solo (approved):

ben-taussig-solo (approved):

PR #9122, opened 2024-01-31
Secret Deletion validation against Previous State

sheidkamp:
# Description
Allow the deletion of secrets despite unrelated errors/warnings.

When validating a modification to a snapshot:
  * the snapshot is cloned
  * the resource is modified or deleted in the clone
  * the modified snapshot is translated and the errors and warnings are extracted from the reports
  * if there are errors, or if there warnings and warnings are not allowed, the modification is rejected.

If the warnings or errors were present in the original snapshot, no updates are possible, unless it is an update that completely removes the blocking issues.

In the case of secret deletion, an alternative form of validation will be used when the initial snapshot validation fails:
* validation will run against a clone of the original snapshot
* the output of that validation will be compared to the validation of the modified snapshot
* if the output is the same (or if there are only changes to warnings and warnings are allowed), accept the deletion.

## Code changes

* https://github.com/solo-io/gloo/blob/sheidkamp/secret-delete-narrow/projects/gateway/pkg/validation/validator.go is the center of the changes. 
  * The `validateProxiesAndExtensions` function was extracted from `validateSnapshot`. This function attempts the translation of the snapshot and performs gloo/extention validation, and returns proxyReports, translated proxies, errors and warnings.
    * This function uses the new `opts.collectAllErrorsAndWarnings` to indicate that it should not stop validating on the first error, and it should collect and return everything.
  * The `compareValidationWithoutModification` function was created to re-run validation and compare the output to the previous run
* https://github.com/solo-io/gloo/blob/sheidkamp/secret-delete-narrow/projects/gloo/pkg/utils/validation/proxy_validation.go
  * `GetListenerError` was extracted from `GetProxyError` for reuse in comparing proxy reports.

Because of this extra overhead, we allow this functionality to be turned off by setting the environment variable `ENABLE_VALIDATION_AGAINST_SNAPSHOT` to false. This functionality will be enabled by default in main, and disabled by default in LTS branches. The long-term plan is to deprecate the environment variable: https://github.com/solo-io/gloo/issues/9146, so we will not add a setting or explicit helm value to manage this field. `gloo.deployment.customenv` can be used to set the environment variable with helm.

## Docs changes
- None

# Context
Users ran into this issue when attempting to delete secrets that had no relation to any resource because they had/errors warning that preventing the validating webhook from ever accepting the delete.

### General Approach
There were three candidates for the overall approach to determine if a secret is in use:

1. Traversing the snapshot to look for the secret. This has the advantage of being able to directly validate whether or not a secret is in use. In terms of implementation there were 2 possible approaches:
    1. Code the snapshot traversal logic by hand - this was undesirable, as it would require maintenance every time a secret was added/moved/changed in our protobuf definitions. Overall a "brittle" approach.
    2. Use code generation to generate the code needed to traverse the snapshot - this would be ideal, but would be time consuming to setup.
2. Compare the output of validation with the output of the previous run, and if its the same, conclude that we did not degrade the state of the system by deleting the secret. Also two possible approaches here:
    1. Save the results of the previous validation to compare against - this has a few issues. Where to store them, as the validator will be destroyed and reinstantiated if settings change, and how to handle the first run when there are no previous validations to compare against. There is also the more subtle issue of validation short-circuiting on the first error for a proxy, which is good for performance, and bad for validation, as we need the complete accounting of issues for the comparison
    2. If the secret validation fails, rerun validation against the original snapshot and compare the output - this approach has the downside of requiring extra processing time when deleting secrets. This extra processing time comes from continuing validation after the first error is found, and the revalidation after the fact.
3. Hook into the secrets client to track all secrets that are being used and check against that list.
    1. This approach relies on the secrets client being used appropriately/consistently, and we are not confident there will not be edge cases.

We ended up taking the approach of rerunning validation against the original snapshot and comparing the results after confirming with the client that the extra overhead was acceptable.

### Comparing validation artifacts
The snapshot object consists of arrays of resources, so we can be confident that the successive validation attempts will loop over the resources in the same order.

There are 3 outputs of the validation process that we can compare:
1. Proxies - These are the proxies generated by translation We will compare the proxies by comparing lengths of the lists generated by the two validation runs, and then if they are the same length, compare their hashes. If these match, we have the same proxies
2. ProxyReports - These are created during gloo validation:
  ```
// The Proxy Report should contain one report for each sub-resource of the Proxy
// E.g., each listener will have a corresponding report. Within each listener report is
// a route report corresponding to each route on the listener.
  ```
  These resources are stored as slices, so the corresponding reports will be in a consistent order. These are easiest to compare when there `allowWarnings` is false. In that case, we can use `reflect.DeepEqual` to check if they are identical.
  If `allowWarnings` is true, we loop over the ProxyReports and their component ListenerReports. We test the equality of the ListenerReport types, and the equality of the errors they contain. If these are equal, the comparison passes

3. Errors - errors come from two places, ProxyReports and ResourceReports. The ProxyReports errors are retrieved by a function that loops over slices of the ListenerReports and their sub-components, and so will return the same order each time. ResourceReports are stored as a map, with InputResources as the keys to the corresponding report. InputResources are only identified by their name and namespace, which are not guaranteed to be unique. As part of this feature development, the reporter code in `solo-kit` was updated to always return the same errors in the same order for a given set of reports. These errors are all stored as a `multierror` which compiles a number of errors into one error with wrapping under the hood. Because we have ensured that errors are generated in the same manner each time, we can simply compare the resulting error strings from the validation runs to look for differences. When `allowWarnings=false`, warnings are treated as errors, and collected/compared along with them.

There are two special errors that may be encountered that are not a result of the snapshot itself. If `Sync` has not yet been run, a `SyncNotYetRunError` error is appended, and if the wrong number of gloo reports are created (`// This was likely caused by a development error.`), a `GlooValidationResponseLengthError`  is appended. These errors both short circuit the snapshot validation process and cause it to fail, as we will not have valid validation results to use for comparison.

## Dependency changes
* updated Eris to `v.0.5.4` to get `As` functionality
* updated `solo-kit` to `0.34.2` to pull in changes to guarantee consistent ordering of errors/warnings in report validation

## Testing steps
New unit tests here added [here](https://github.com/solo-io/gloo/blob/2d6491794399b3b34ed0978f9c0f7899b99660fb/projects/gateway/pkg/validation/validator_test.go#L217) to capture the different cases identified in the [use case matrix](https://docs.google.com/spreadsheets/d/1vAFLCyYAXJ05p2NRj-DZu4lJkK7nz_d09puzJ0GAsEg/edit#gid=0)

A basic smoke test can be performed by running through the reproduction steps identified in the issue.

Outstanding testing:
* ~~kube2e tests~~
* manual tests for errors from proxy validation and extension validation, ~~and automated versions if feasible~~ Unit tests in place, still need manual steps
* ~~improved unit testing around `solo-kit` reporter libraries.~~ UPDATED

## Notes for reviewers
Solo-kit changes: https://github.com/solo-io/solo-kit/pull/550

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8931


projects/gateway/pkg/validation/validator.go
```diff
@@ -50,6 +51,22 @@ func (r *Reports) GetProxies() []*gloov1.Proxy {
 type ProxyReports []*validation.ProxyReport
 type UpstreamReports []*validation.ResourceReport
 
+type GlooValidationResponseLengthError struct {
```
davidjumani, line 54: nit: A comment as to why this needs to be a struct (assuming it is required for errors.As) instead of a method that returns an error

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +520,185 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
 	return reports, nil
 }
 
+// shouldRetryValidationOnWarnings contains the logic to determine if validation should be retried against the original snapshot
+// and the results of that valdidation compared to the original validation output in order to determine whether to accept the modification.
+// Currently we only support this for the deletion of secrets.
+func (v *validator) shouldRetryValidation(ctx context.Context, opts *validationOptions) bool {
```
davidjumani, line 526: Not sure if the method name in the comment or the method name itself needs to be changed

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +520,185 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+}
+
+func (v *validator) passedValidation(errs error, warnings error) bool {
+	return errs == nil && (warnings == nil || v.allowWarnings)
+}
+
+// compareValidationWithoutModification is used to compare the output of validation against validation of the orginal snapshot
+// this is used in special cases. specifically the deletion of a secret.  In these cases, the usual validation logic is overriden,
+// and instead of relying on the presence of errors and warnings to determine whether to accept the modification, the output of
+// validation of the request (proxies, proexReports, errors, and warnings) is compared yo the output of the validation of the original snapshot.
```
davidjumani, line 545:
```suggestion
// validation of the request (proxies, proexReports, errors, and warnings) is compared to the output of the validation of the original snapshot.
```

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +520,185 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+// This is somewhat redundant as we have already extracted and compared the errors, but provides
+// and extra layer of validation.
+func compareReports(proxyReports1, proxyReports2 ProxyReports, allowWarnings bool) bool {
+	// If warnings are not allowed, we can compare the proxy reports directly
+	if !allowWarnings {
+		return reflect.DeepEqual(proxyReports1, proxyReports2)
+	}
+
+	// Warnings are allowed, so the proxy reports must be compared manually
+	if len(proxyReports1) != len(proxyReports2) {
```
davidjumani, line 625: Could there be a case where one report has additional warnings but the same errors ?

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +520,185 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+}
+
+// compareProxies compares two lists of proxies and returns true if they are the same
+// First the length of the lists is compared, then the hash of each proxy is compared
+// If all of these are the same, the proxies are considered the same
+func compareProxies(proxy1, proxy2 []*gloov1.Proxy) bool {
+	sameProxies := len(proxy1) == len(proxy2)
+	if sameProxies {
+		for i := range proxy1 {
+			if proxy1[i].MustHash() != proxy2[i].MustHash() {
```
davidjumani, line 672: Is it guaranteed that the proxies are ordered in the same way ?

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +520,185 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+			"Validation against original snapshot failed, accepting modification",
+			zap.Bool("sameProxies", sameProxies),
+			zap.Bool("sameReports", sameReports),
+			zap.Bool("sameErrors", sameErrors),
+			zap.Bool("sameWarnings", sameWarnings),
+			zap.String("resource", opts.Resource.GetMetadata().String()),
+		)
+	} else {
+		contextutils.LoggerFrom(ctx).Debugw(
+			"Validation against original snapshot succeded, accepting modification",
```
davidjumani, line 591: Is the resource accepted or determined on line 489 ?

projects/gateway/pkg/validation/validator.go
```diff
@@ -255,89 +459,49 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
 
-	extensionReports := v.extensionValidator.Validate(ctx, snapshotClone)
-	if len(extensionReports) > 0 {
-		if err = v.getErrorsFromResourceReports(extensionReports); err != nil {
-			err = errors.Wrapf(err, failedExtensionResourceReports)
-			errs = multierr.Append(errs, err)
+	if !v.passedValidation(errs, warnings) && !overrideErrors {
+
+		// If we have warnings and they are not allowed, they are errors.
+		if warnings != nil {
```
davidjumani, line 492: Missing a !allowWarnings ?

projects/gateway/pkg/validation/validator.go
```diff
@@ -473,33 +816,46 @@ func (v *validator) validateResource(opts *validationOptions) (*Reports, error)
 	}
 }
 
-// getErrorsFromGlooValidation returns an error comprising of the gloo reports. The errors will include warnings if
-// allowWarnings is not set.
-func (v *validator) getErrorsFromGlooValidation(reports []*gloovalidation.GlooValidationReport) error {
-	var errs error
+// getErrorsFromGlooValidation returns errors and warnings from the Gloo validation reports. It uses the warningHandling field to determine
+// how to handle warnings. This function is consistent with our general warning handling approach:
+// * fill in &
```
davidjumani, line 821: ?

projects/gateway/pkg/validation/validator.go
```diff
@@ -473,33 +816,46 @@ func (v *validator) validateResource(opts *validationOptions) (*Reports, error)
...
-				for _, warning := range warnings {
-					errs = multierr.Append(errs, errors.New(warning))
+
+			if proxyWarnings := validationutils.GetProxyWarning(proxyReport); len(proxyWarnings) > 0 {
+				// `opts` is not passed down but warningHandling can be used to determine how to handle warnings
+				if warningHandling == reporter.SeparateWarnings {
+					for _, warning := range proxyWarnings {
+						warnings = multierror.Append(warnings, errors.New(warning))
+					}
+				} else if warningHandling != reporter.IgnoreWarnings {
```
davidjumani, line 849:
```suggestion
				} else if warningHandling == reporter.Strict {
```

davidjumani:
We would also need to enable / disable the feature based on a helm value - My preference is to add this as a field in the setting object so the pod will not need to be restarted for changes to take effect, or keep it as an env var and populate the deployment template accordingly

projects/gateway/pkg/validation/validator.go
sheidkamp: updated

sheidkamp (commented):
One round of updates

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -842,10 +842,13 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 	// create a validator to validate extensions
 	extensionValidator := syncerValidation.NewValidator(syncerValidatorExtensions, opts.Settings)
 
+	// allow by default
+	disableValidationAgainstSnapshot := os.Getenv("DISABLE_VALIDATION_AGAINST_SNAPSHOT") != "true"
```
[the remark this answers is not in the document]
sheidkamp: yes

projects/gateway/pkg/validation/validator.go
sheidkamp: Test added

projects/gateway/pkg/validation/validator.go
```diff
@@ -217,6 +239,180 @@ func (v *validator) validateSnapshotThreadSafe(opts *validationOptions) (
...
+	if opts.collectAllErrorsAndWarnings {
+		warningHandling = reporter.SeparateWarnings
+	} else if v.allowWarnings {
+		warningHandling = reporter.IgnoreWarnings
+	}
+
+	gatewaysByProxy := utils.GatewaysByProxyName(snapshot.Gateways)
+	// translate all the proxies
+	for proxyName, gatewayList := range gatewaysByProxy {
+		proxy, reports := v.translator.Translate(ctx, proxyName, snapshot, gatewayList)
```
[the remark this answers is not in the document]
sheidkamp: I implemented something along these lines, passing the `errs` and `warnings` slices as pointers and updated them in place.

The function signatures are a little messy, lmk if you think this makes it more readable.

projects/gateway/pkg/validation/validator.go
```diff
@@ -217,6 +239,180 @@ func (v *validator) validateSnapshotThreadSafe(opts *validationOptions) (
...
+			errs = multierror.Append(errs, err)
+			continue
+		}
+
+		// Validate the reports returned by the glooValidator
+		proxyReport := glooReports[0].ProxyReport
+		proxyReports = append(proxyReports, proxyReport)
+
+		// Get the errors from the proxyReport
+		if err := validationutils.GetProxyError(proxyReport); err != nil {
```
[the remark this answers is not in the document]
sheidkamp: As part of the refactoring suggested above, I have pulled this logic into the `appendProxyErrorsAndWarnings` function. Not exactly the same approach, but this loop is now cleaner.

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +563,184 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+
+	// Loop over the proxy reports and compare their errors. Ignore warnings
+	for i := range proxyReports1 {
+		pr1 := proxyReports1[i]
+		pr2 := proxyReports2[i]
+
+		// Check that the listener reports are the same types
+		l1 := pr1.GetListenerReports()
+		l2 := pr2.GetListenerReports()
+		if len(l1) != len(l2) {
```
sheidkamp, line 679: Following up on the above comment, this is where we create the listener reports: https://github.com/solo-io/gloo/blob/85a8f3f509f47d93e877b932e9785998215210c5/projects/gloo/pkg/utils/validation/proxy_validation.go#L72

the `listenerReport` array is `make`d with a size of `len(listeners)`, and listenerReports are placed into them, so from the same snapshot (plus or minus a secret), there will the same number of listener reports for a proxy regardless of the presence or absence of errors and warnings.

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +563,184 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
 	return reports, nil
 }
 
+// shouldRetryValidation contains the logic to determine if validation should be retried against the original snapshot
+// and the results of that valdidation compared to the original validation output in order to determine whether to accept the modification.
+// Currently we only support this for the deletion of secrets.
+func (v *validator) shouldRetryValidation(ctx context.Context, opts *validationOptions) bool {
+	if v.disableValidationAgainstSnapshot {
```
nfuden, line 570: The name of this variable versus what it does seems odd to me.

It doesn't really disable validation against snapshot. It rather stops us from checking against the previous state of the world.

On that note I wonder about shouldRetryValidation's naming as well.

sheidkamp: Happy to rename variables, `disableValdiationAgainstPreviousSnapshot` is starting to get a little long - any suggestions? I'll think on names for this and `shoudRetryValidation`.

projects/gateway/pkg/validation/validator.go
```diff
@@ -255,89 +438,42 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+	proxies, proxyReports, errs := v.validateProxiesAndExtensions(ctx, snapshotClone, opts)
 
-		err = v.getErrorsFromGlooValidation(glooReports)
-		if err != nil {
-			err = errors.Wrapf(err, failedResourceReports)
-			errs = multierr.Append(errs, err)
-			continue
-		}
+	passedSnapshotValidation := false
+	// We want to compare the validation output if the retryValidation flag and we are currently not passing validation
```
ben-taussig-solo, line 453: nit/typo
```suggestion
	// We want to compare the validation output if validateAgainstCurrentSnapshot is true and we are currently not passing validation
```

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +492,174 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+	}
+	return false
+}
+
+// compareValidationWithoutModification is used to compare the output of validation against validation of the orginal snapshot
+// This is used in special cases, specifically the deletion of a secret.  In these cases, the usual validation logic is overriden,
+// and instead of relying on the presence of errors to determine whether to accept the modification, the output of
+// validation of the modified snapshot (proxies, proxyReports, errors) is compared to the output of the validation of the original snapshot.
+// If outputs are the same, it is assumed that the modification did not degrade the system and  is accepted
+func (v *validator) compareValidationWithoutModification(ctx context.Context, opts *validationOptions, proxies []*gloov1.Proxy, proxyReports ProxyReports, errs error) bool {
```
ben-taussig-solo, line 515: How would you feel about creating a `validationOutput` struct so that we can pass the validation results around in a single object? I think it would make this function signature a bit cleaner/easier to understand

```go
type validationOutput struct {
    Proxies []*gloov1.Proxy
    ProxyReports []*gloov1.ProxyReport
    Upstreams []*gloov1.Upstream
}
```

Then we could do

```diff
- func (v *validator) compareValidationWithoutModification(ctx context.Context, opts *validationOptions, proxies []*gloov1.Proxy, proxyReports ProxyReports, errs error) bool {
+ func (v *validator) compareValidationWithoutModification(ctx context.Context, opts *validationOptions, modifiedSnapshotOutput validationOutput) bool {
```


Please push back if this is overkill or adding unnecessary complexity -- given that we only call `compareValidationWithoutModification` in one place, it might not be worth it if it requires us to change a bunch of other code.

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +492,174 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+
+	snapshotCloneUnmodified, err := v.copySnapshotNonThreadSafe(ctx, opts.DryRun)
+	if err != nil {
+		// If storage is broken default to to disallowing the update.
+		// Don't override the initial errors without being confident that the update is valid
+		return false
+	}
+
+	// Get the validation output without the modification.
+	proxiesNoMod, proxyReportsNoMod, errorsNoMod := v.validateProxiesAndExtensions(ctx, snapshotCloneUnmodified, opts)
```
ben-taussig-solo, line 536: nit: How do you feel about computing these values beforehand and passing them into `compareValidationWithoutModification` (even if we just do it inline before we call `compareValidationWithoutModification`)? It seems a little misleading to me that `compareValidationWithoutModification` is doing validation and comparison instead of just comparison

projects/gateway/pkg/validation/validator.go
sheidkamp: I think makes sense, especially combined with your other comment about restructuring the comparison function. Implemented.

projects/gateway/pkg/validation/validator.go
sheidkamp: I have split this so `validatateSnapshot` calls `validateAgainstCurrentSnapshot` which does some error checking then gets the output of `validateProxiesAndExtensions` and feeds that into `compareValidationOutputs` which just does the comparisons.

sheidkamp:
Labelled as "work in progress" until `solo-kit` changes are merged and pulled.

projects/gateway/pkg/validation/validator.go
sheidkamp: Renamed this method and the variable that stores the result to `useValidationAgainstCurrentSnapshot`

projects/gateway/pkg/utils/gateway_filtering.go
```diff
@@ -17,6 +19,41 @@ func GatewaysByProxyName(gateways v1.GatewayList) map[string]v1.GatewayList {
 	return result
 }
 
+// GatewaysAndProxyName is a struct that holds a list of gateways and a proxy name
+type GatewaysAndProxyName struct {
```
davidjumani, line 23: Is there a specific reason for adding this struct that I'm missing instead of just a map of names to gateways ? Since it doesn't have any methods defined on it I think that a map should suffice

projects/gateway/pkg/validation/validator.go
```diff
@@ -217,6 +239,169 @@ func (v *validator) validateSnapshotThreadSafe(opts *validationOptions) (
...
+	return validationOutput{
+		proxies:      proxies,
+		proxyReports: proxyReports,
+		err:          errs,
+	}
+}
+
+// appendProxyErrors appends the errors and from the proxyReport to the (multi)error, passed in by pointer
+// It returns a boolean to indicate whether the caller should continue processing the next proxy
+func (v *validator) appendProxyErrors(errs *error, proxyReport *validation.ProxyReport, proxy *gloov1.Proxy, opts *validationOptions) bool {
```
davidjumani, line 383: Since this just returns the opposite of `opts.collectAllErrors`, I don't think we need to have this to have a return value unless I'm missing something

projects/gateway/pkg/validation/validator.go
davidjumani: nit: `shouldValidaeAgainstCurrentSanpshot` seems more readable

projects/gateway/pkg/validation/validator.go
```diff
@@ -114,21 +129,28 @@ type validationOptions struct {
 	Delete      bool
 	Resource    resources.Resource
 	Gvk         schema.GroupVersionKind
+	// This flag is used when re-validating a snapshot when deleting secrets and is used in setting
+	//  the `resource` parameter passed to the glooValidator, which will remove the resource if it is present
+	validateUnmodified bool
```
davidjumani, line 134: Maybe this is already handled in the code, but if this is set to true, is it guaranteed that `Resource` is nil ?

davidjumani (commented):
LGTM, couple of nits and questions

sheidkamp:
/kick doesn't seem to be an error message

projects/gateway/pkg/validation/validator.go
```diff
@@ -356,6 +502,182 @@ func (v *validator) validateSnapshot(opts *validationOptions) (*Reports, error)
...
+	if opts.Delete && opts.Gvk.Kind == "Secret" {
+		return true
+	}
+	return false
+}
+
+// validateAgainstCurrentSnapshot is used to compare the output of validation against validation of the orginal snapshot.
+// It takes the output of the first valdidation with the modification and then gets the output of the validation of the original snapshot
+// and compares the two using compareValidationOutputs, unless there are breaking errors in the first validation output
+func (v *validator) validateAgainstCurrentSnapshot(ctx context.Context, opts *validationOptions, validationOutput validationOutput) bool {
```
ben-taussig-solo, line 523: nit: I don't love the variable name and the type being the same (`validationOutput validationOutput`). What do you think about `validationOutputWithMod`? `modifiedValidationOutput`?

projects/gateway/pkg/validation/validator.go
ben-taussig-solo: Looks great, thanks!

test/kube2e/gateway/gateway_test.go
```diff
@@ -2212,6 +2215,174 @@ spec:
...
+        single:
+          upstream:
+            name: my-us
+            namespace:  ` + testHelper.InstallNamespace
+
+				updateFailurePolicy(admissionregv1.Fail)
+				// Allow warnings during setup so that we can install the resources
+				kube2e.UpdateAllowWarningsSetting(ctx, true, testHelper.InstallNamespace)
+
+				// This should work regardless of whether the warnings are allowed or not
```
ben-taussig-solo, line 2279: Is this because nothing is routing to the bad upstream yet?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2212,6 +2215,174 @@ spec:
...
+			const (
+				secretName       = "tls-secret"
+				unusedSecretName = "tls-secret-unused"
+			)
+
+			var (
+				invalidUpstreamYaml string
+				vsYaml              string
+
+				updateFailurePolicy = func(failurePolicy admissionregv1.FailurePolicyType) {
```
ben-taussig-solo, line 2229: Is there a particular reason we define `updateFailurePolicy` here and `UpdateAllowWarningsSetting` in the kube2e utils?

projects/gateway/pkg/validation/validator.go
davidjumani: In that case I think it would be better renamed to `appendProxyErrorsAndProceed` or something along those lines but again this is a nit

projects/gateway/pkg/validation/validator.go
ben-taussig-solo: resolved here: https://github.com/solo-io/gloo/pull/9122/commits/645f2b1c8687b627241f99f718eaa3896ded73d8

test/kube2e/gateway/gateway_test.go
ben-taussig-solo: moved to utils here: https://github.com/solo-io/gloo/pull/9122/commits/645f2b1c8687b627241f99f718eaa3896ded73d8#diff-909288de4ab619d47d75a89fc9d6f4acfbd645ec16063a9cad8dcde18d591762

ben-taussig-solo (dismissed):
Looks good to me!

davidjumani (commented):
LGTM, just one nit

davidjumani (dismissed):

projects/gateway/pkg/validation/validator_test.go
```diff
@@ -245,28 +364,75 @@ var _ = Describe("Validator", func() {
...
+				} else {
+					Expect(err).To(HaveOccurred())
+				}
+			},
+				// Regular validation
+				reportValidationEntries,
+			)
+
+			DescribeTable("Breaking errors don't trigger revalidation", func(validator validationFunc, expectedErrString string) {
+				//FIt("Fails if it gets a hasn't Sync'ed", func() {
```
davidjumani, line 408: Cleanup

ben-taussig-solo (approved):

davidjumani (approved):

nfuden (approved):

PR #9041, opened 2024-01-09
Implement toggle to control istio x-forwarded-for header

SantoDE:
# Description

Fixes #8855 

With our reworked istio-integration that we introduced with 1.15, its not always necessary to rewrite the xff header. We have some users that run into trouble because of the header being present. Now there is a toggle to control, whether or not users want to have the header.

## API changes
- Added a new field called appendIstioXff to Settings

## Code changes
- Honour the new field and toggle it on the generated envoy config

# Context
Users ran into this, that now in some situations the header was added where it previously might not have been. To get more control over that header, there is now the toggle.

## Interesting decisions
 
We chose to do things this way because ...

## Testing steps

I manually verified behavior by toggling the behaviour through helm installations with a values file like this
```
global:
  istioIntegration:
    enableIstioSidecarOnGateway: true
    disableAutoinjection: true
    appendIstioXff: true
  istioSDS:
    enabled: true
  glooMtls:
    istioProxy:
      image:
        registry: docker.io/istio
        repository: proxyv2
        tag: 1.19.5
    sds:
      image:
        repository: sds
        tag: 1.15.10
```

## Notes for reviewers

Be sure to verify intended behavior by ...

Please proofread comments on ...

This is a complex PR and may require a huddle to discuss ...

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8855

PR #9125, opened 2024-01-31
Istio X Forwarded Host toggle

bewebi:
# Description

Expose a `global.istioIntegration.AppendXForwardedHost` Helm value which allows users opt out of setting the `x-forwarded-host` header on requests using the Istio plugin

## API changes
- Added `IstioOptions` message to the `Settings` proto with a `AppendXForwardedHost` boolean field

## Code changes
- Istio Integration plugin gets a new `appendXForwardedHost` field which gets its value from the `Gloo.IstioOptions.AppendXForwardedHost` field on `Init()`
  - The output `routeAction` only has `Route.AppendXForwardedHost` set to `true` if the plugin's corresponding field is `true`
  - Add unit tests for this behavior

## Helm changes
- Expose `global.IstioIntegration.AppendXForwardedHost` Helm value
  - This value is propagated directly as `gloo.IstioOptions.AppendXForwardedHeader` in the Settings template
  - This value is set to `true` in `values-template.yaml`, making `true` the default
  - Add helm test for this behavior

# Context

This replaces #9041

## Interesting decisions
 
The location of the value to config was chosen to make the most sense in the immediate context
For Helm it made the most sense to add the Istio X Forwarded Host valeu to the existing `IstioIntegration` section
For the Settings proto we follow the pattern of creating messages for plugin-specific config under the `Gloo` message
Users are expected to configure via Helm, so the fact that the location of the value doesn't match the Settings CR 1:1 is not expected to lead to significant confustion
Both are included in reference docs

## Testing steps

I have manually verified the intended behavior using the following steps:
- Use `./ci/kind/setup-kind.sh` to setup a KinD cluster
- Run `make build-test-chart` to create a local Helm chart
- Follow the [docs guide for setting up Edge with Istio](https://docs.solo.io/gloo-edge/latest/guides/integrations/service_mesh/istio/), but using the local chart
  - The installation command becomes
  ```
  helm install gloo ./install/helm/gloo --namespace gloo-system --create-namespace -f value-overrides.yaml
  ```
- Observe that the Settings CR has `gloo.istioOptions.appendXForwardedHost` set to `true`
- At step 3 of the "Set up and verify the mTLS connection" section of the guide, observe that the response contains the `X-Forwarded-Host` header:
```
{
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.default",
    "User-Agent": "curl/8.1.2",
    "X-B3-Sampled": "0",
    "X-B3-Spanid": "c2f0a67a569f15dd",
    "X-B3-Traceid": "1694608ef69f94e4c2f0a67a569f15dd",
    "X-Forwarded-Host": "www.example.com"
  }
}
```
- Modify `value-overrides.yaml` to add `global.istioIntegration.XForwardedHost=false`
- Upgrade the helm installation:
```
helm upgrade gloo ./install/helm/gloo --namespace gloo-system --create-namespace -f value-overrides.yaml
```
- Observe that the Settings CR has `gloo.istioOptions.appendXForwardedHost` set to `false`
- Execute the same `curl` from the guide and observe that the response does not contain the `X-Forwarded-Host` header:
```
{
  "headers": {
    "Accept": "*/*",
    "Host": "httpbin.default",
    "User-Agent": "curl/8.1.2",
    "X-B3-Sampled": "0",
    "X-B3-Spanid": "0f74a457a0753ca9",
    "X-B3-Traceid": "0bf1353677a9eaeb0f74a457a0753ca9"
  }
}
```

Further testing and validation is expected to be customer success' responsibility

# Checklist:


install/helm/gloo/templates/18-settings.yaml
```diff
@@ -72,6 +72,10 @@ spec:
       fallbackToFirstFunction: {{ .Values.settings.aws.fallbackToFirstFunction }}
   {{- end }}
 {{- end }}
+  {{- if .Values.global.istioIntegration }}
+    istioOptions:
+      appendXForwardedHost: {{ .Values.global.istioIntegration.appendXForwardedHost }}
+  {{- end }}
```
jbohanon, line 78: What happens here if a user puts
```yaml
global:
  istioIntegration: {}
```

projects/gloo/api/v1/settings.proto
```diff
@@ -725,6 +725,14 @@ message GlooOptions {
     // Set escapeCharacters for all TransformationTemplates on all vhosts and routes.
     // This setting can be overridden in individual TransformationTemplates
     google.protobuf.BoolValue transformation_escape_characters = 17;
+
+    message IstioOptions {
+        // Set to false externto disable adding X-Forwarded-Host header in Istio integration
```
jbohanon, line 730:
```suggestion
        // Set to false to disable adding X-Forwarded-Host header in Istio integration
```

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -60,7 +66,11 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 	routeAction.Route.HostRewriteSpecifier = &envoy_config_route_v3.RouteAction_HostRewriteLiteral{
 		HostRewriteLiteral: hostInMesh,
 	}
-	routeAction.Route.AppendXForwardedHost = true
+
+	if p.appendXForwardedHost {
```
jbohanon, line 70: Do we want to explicitly set it to false if our setting isn't enabled? I'm not sure if we have guarantees that the out route struct is at zero state when we receive it. If we do know that, then I'm fine with this as-is

jbohanon (commented):
Great work! 

Do we want to have e2e/kube2e tests to validate the behavior as the end-user sees it?

projects/gloo/pkg/plugins/istio_integration/plugin.go
bewebi: I'm not sure its guaranteed per se, but the only time we have a `Route_Route` type as an `Action` and therefore reach this line, the action will be in a zero state (see [here](https://github.com/solo-io/gloo/blob/6de8d8fabbf7b8959a515452efd30bde50b473d3/projects/gloo/pkg/translator/route_config.go#L283))

I'm not sure if we could reason about the case where we somehow did receive an Action that already had `AppendXForwardedHost` set to `true` when the Settings would suggest it should be otherwise...that implies a situation where some feature has been introduced without accounting for all possible permutations of config, thereby introducing a bug

In practice I don't think it matters either way...if it seems more thorough to include the `false` case I'm happy to

install/helm/gloo/templates/18-settings.yaml
bewebi: I believe the value from `values-template.yaml` fills in, which is `true`
When I `helm template` with this I get a Settings CR with `gloo.istioOptions.appendXForwardedHost` set to `true`

bewebi:
I'm unclear if automated e2e tests are warranted
I was hoping to get context/feedback from the customer success side before investing time in that

projects/gloo/pkg/plugins/istio_integration/plugin.go
jbohanon: Hmm. Interesting situation. I am fine with it I think. It almost seems to imply we should check that it's not `true` and raise an error, but that seems way overly defensive. I am good with it as is.

jbohanon (dismissed):

projects/gloo/pkg/plugins/istio_integration/plugin.go
```diff
@@ -29,7 +30,12 @@ func (p *plugin) Name() string {
 	return ExtensionName
 }
 
-func (p *plugin) Init(_ plugins.InitParams) {
+func (p *plugin) Init(params plugins.InitParams) {
+	if xfh := params.Settings.GetGloo().GetIstioOptions().GetAppendXForwardedHost(); xfh != nil {
+		p.appendXForwardedHost = xfh.GetValue()
+	} else {
+		p.appendXForwardedHost = true
+	}
```
davidjumani, line 38: If the default is false, would it be better to just set this instead of the if statement ?
I believe the getters handle nils
```
p.appendXForwardedHost = params.Settings.GetGloo().GetIstioOptions().GetAppendXForwardedHost().GetValue()
```

bewebi: The default is `true`
If the field is missing from Settings (which we don't expect will ever be the case fwiw) we set to `true`

davidjumani: Got it, nice defensive check

davidjumani (approved):

davidjumani (approved):

jbohanon (approved):

PR #9130, opened 2024-02-02
Correct the extauth grpc passthroughauth retryproto comment

davidjumani:
# Description

Correct the extauth grpc passthroughauth retryproto comment.
It incorrectly mentions that a default retry will be configured when that is the intended behaviour

## API changes
- N/A

## Code changes
- N/A

## CI changes
- N/A

## Docs changes
- N/A

# Context
- N/A

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1389,9 +1389,8 @@ message PassThroughGrpc {
     // TLS config for the Grpc passthrough, if not configured the connection will use insecure.
     PassThroughGrpcTLSConfig tlsConfig =3;
 
-    // Indicates the retry policy for re-establishing the gRPC stream
-    // This field is optional. If max interval is not provided, it will be set to ten times the provided base interval.
-    // If not set, gRPC streams default base interval:500ms, maximum interval:30s will be applied.
+    // Indicates the retry policy for re-establishing the gRPC stream.
+    // This field is optional and failed calls will not not retry unless configured.
```
sheidkamp, line 1393: The extauth code works this way if retry policy is nil, but are we always setting a retry policy here: https://github.com/solo-io/solo-projects/blob/c434d34dffc641b87e47458781b013f8e900ae6a/projects/extauth/pkg/config/translator.go#L745 ?

sheidkamp (commented):
one question

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
nfuden: Are we also setting it always in the kube controller variant?

sheidkamp:
```suggestion
    // This field is optional and failed calls will not retry unless configured.
```

sheidkamp: @nfuden - we are setting it here: https://github.com/solo-io/ext-auth-service/blob/32bee3462fc556c40a2c9681a3ada7b70d08ffb3/pkg/controller/translation/pass_through.go#L89 

but in that case it should be nil, and the Retry opts are not added:
 https://github.com/solo-io/ext-auth-service/blob/32bee3462fc556c40a2c9681a3ada7b70d08ffb3/pkg/config/passthrough/grpc/grpc_client_manager.go#L73

sheidkamp (approved):
lgtm

jbohanon (approved):

PR #9117, opened 2024-01-30
Merge v2.0.x to main

jenshu:
# Description

Initially, everything for Gloo Gateway v2 was developed separately against the v2.0.x branch. This PR merges all of it back into `main`. Roughly, this merge involved:
- pulling all the changes from last few months from `main` into the `v2.0.x` branch, resolving conflicts, and adding back deleted code (mostly in glooctl)
- fix function signatures and typos to get `make generated-code` working
- fixing all the lint errors in the v2 code (since we recently added a linter to `main`)
- re-enabling CI checks and ensure all checks pass for both GE and GGv2
    - as part of this, need to support 2 translation modes in some places (see https://github.com/solo-io/gloo/pull/9105 which is incorporated into the current PR)

## Notes for reviewers

The goal of this PR is to merge all work from `v2.0.x` into `main` and ensure CI is passing for both Gloo Edge and Gloo Gateway v2 (with the minimal changes necessary), so that further GGv2 development can be done against `main`.  Not currently looking for thorough review of the PRs that are being pulled directly from v2.0.x (as they had already gotten reviewed when they were initially merged). 

As long as regression tests / build-bot pass, I'm confident we didn't break anything in GE.  GG is likely broken in some places (e.g. glooctl) but will be handled in follow-up PRs (along with cleanup and deduping of GE/GG code).

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #9131, opened 2024-02-02
Revert mock update

jenshu:
The [merge from v2 branch](https://github.com/solo-io/gloo/pull/9117) brought in a dependency update from `github.com/golang/mock` to `go.uber.org/mock`. While we should probably update that dependency eventually (as the golang one is no longer maintained), this update currently causes test compilation issues when pulling into solo-projects, as all our other repos still use the golang mock. It will take some coordination to get the updates into all the other repos. For now just revert the update in gloo to avoid the conflicts.

Note: not including in changelog because there hasn't been a gloo release with the uber mock update so there should be no user facing change


davidjumani (approved):
Did a quick look at the branch. No references to `go.uber.org/mock` found

sheidkamp (approved):
Checked for `go.uber.org/mock` and successfully ran codegen.

bewebi (approved):

PR #9139, opened 2024-02-07
Use hostname for hashing

asayah:
# Description

exposing use_hostname_for_hashing, the hostname will be used for hashing when using maglev for example, useful when using multiple host in the upstreams that resolve to the same IP.

## API changes
- Added use_hostname_for_hashing field to LoadBalancerConfig resource
- ...

## Code changes
- exposing use_hostname_for_hashing
- ...

# Checklist:


jbohanon:
Awesome! Any appetite to expose it as a helm value? I don't even know if we have a helm section for lb config

asayah:
> Awesome! Any appetite to expose it as a helm value? I don't even know if we have a helm section for lb config

LBs are configured per upstream basis, if it was global it would be useful as a helm value

asayah:
> Is there an e2e test we could add to validate this behavior?

I added a test

jbohanon (approved):
I still think an e2e test validating user-facing behavior would be good but since customer is confirming this build solves their use case, I am good with this merging

nfuden (approved):

PR #9148, opened 2024-02-09
Combine Gloo Gateway and Gloo Edge binaries/helm charts

jenshu:
# Description

Up until now, Gloo Gateway (GG) has lived in its own helm chart and was deployed separately from Gloo Edge (GE).
This PR moves GG into the same helm chart/deployment as GE, so that GE and GG get installed together (via the existing GE helm chart). 

## Helm changes
- GG helm templates and values have been moved into the GE chart. By default, installing the GE chart does not enable GG. To enable it, set the env value `GG_EXPERIMENTAL_K8S_GW_CONTROLLER` to `true` in the `gloo` deployment. (This is intended to be a temporary toggle; it will be made into an actual helm/settings value in a follow-up)
- Note that the Kubernetes Gateway API CRDs are not included in the GE helm chart, but rather they are expected to be applied as an extra step by the user if using GG. 
- The dynamically generated GG proxy deployment now uses the gloo envoy wrapper

## Code changes
- GE and GG now share and xds server and xds cache. The proxies produced by each translator are prefixed with the translator name to ensure cache entries don’t conflict.

## CI changes
- GG conformance tests now install the full GE helm chart
- The setup-kind script has been updated to apply the Gateway API CRDs so that they exist for regression tests

## CLI changes
- Disabled GG glooctl commands for now, as they weren’t fully functional

## What is not done in this PR

In an effort to get this change merged asap, some work will be done in follow-up PRs:
- re-enabling GG controller in GE regression tests (the enabling of GG currently causes failures in GE that need further investigation)
- moving new toggles (e.g. to turn on/off GG) to settings / helm values
- ensuring health/metrics ports are working for GG
- cleanup and/or consolidation w/GE of unused/stale GG make targets / helm values
- additional unit / e2e tests with consolidated GE/GG
- un-hardcoding the gloo-system namespace in the GG deployer
- steps for enabling debug logs on dynamic proxy deployments
- steps for dev workflow via make targets
- support non-cluster-scoped rbac for GG 
- clean up all the excess logging

# Checklist:


sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=32df8235-b4c3-4898-bc0a-d456544e0488

SEeing 404s during prepare-workspace:
```
Step #1 - "run-tests": Step #0 - "prepare-workspace": go: github.com/hashicorp/go-kms-wrapping/entropy/v2@v2.0.0: verifying module: github.com/hashicorp/go-kms-wrapping/entropy/v2@v2.0.0: reading https://sum.golang.org/tile/8/1/148: 404 Not Found
Step #1 - "run-tests": Step #0 - "prepare-workspace": 	server response: not found
Step #1 - "run-tests": Step #0 - "prepare-workspace": go: github.com/ianlancetaylor/demangle@v0.0.0-20200824232613-28f6c0f3b639: verifying module: github.com/ianlancetaylor/demangle@v0.0.0-20200824232613-28f6c0f3b639: reading https://sum.golang.org/tile/8/1/023: 404 Not Found
Step #1 - "run-tests": Step #0 - "prepare-workspace": 	server response: not found
Step #1 - "run-tests": Step #0 - "prepare-workspace": go: github.com/yashtewari/glob-intersection@v0.1.0: verifying go.mod: github.com/yashtewari/glob-intersection@v0.1.0/go.mod: reading https://sum.golang.org/tile/8/1/148: 404 Not Found
Step #1 - "run-tests": Step #0 - "prepare-workspace": 	server response: not found
Step #1 - "run-tests": Step #0 - "prepare-workspace": go: gopkg.in/natefinch/lumberjack.v2@v2.2.1: verifying module: gopkg.in/natefinch/lumberjack.v2@v2.2.1: reading https://sum.golang.org/tile/8/0/x060/428: 404 Not Found
Step #1
```

projects/gateway2/controller/translator.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"context"
+
+	gloov1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins/registry"
+	"github.com/solo-io/gloo/projects/gloo/pkg/translator"
+)
+
+func newGlooTranslator(ctx context.Context) translator.Translator {
```
sam-heilbron, line 12: I think we could just inject the translator that is created in the setup_syncer, rather than reconstructing it. This will also ensure we getting the SEttings values for free

lgadban (commented):
can we add these to follow up section?
* steps for enabling debug logs on dynamic proxy deployments
* steps for dev workflow via make targets

ci/kind/setup-kind.sh
```diff
@@ -59,7 +59,10 @@ fi
 # 4. Build the gloo command line tool, ensuring we have one in the `_output` folder
 USE_SILENCE_REDIRECTS=true make -s build-cli-local
 
-# 5. Install additional resources used for particular KUBE2E tests
+# 5. Apply the Kubernetes Gateway API CRDs
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml
```
lgadban, line 63: just adding a note for now; this is a different set of CRDs than what we were installing prevously (standard vs. experimental)
e.g. https://github.com/solo-io/gloo/blob/v2.0.x/projects/gateway2/Makefile#L30

install/helm/gloo/values-template.yaml
```diff
@@ -2,6 +2,124 @@ namespace:
...
+      # requests:
+      #   cpu: 100m
+      #   memory: 128Mi
+
+
+    nodeSelector: {}
+
+    tolerations: []
+
+    affinity: {}
```
lgadban, line 72: do we care about any of this now? we are effectively removing the deployment.yaml in this PR, feels like we should remove the values for it as well

projects/gateway2/cmd/controller.go
```diff
@@ -25,6 +25,6 @@ func init() {
 	controllerCmd.Flags().StringVar(&cfg.Release, "release-name", "", "The release name for gateway resources")
 	controllerCmd.Flags().BoolVar(&cfg.Dev, "dev-mode", false, "Turn on dev mode (more verbose logging, etc.)")
 	controllerCmd.Flags().BoolVar(&cfg.AutoProvision, "auto-provision", false, "Auto provision gateway resources")
-	controllerCmd.Flags().StringVar(&cfg.XdsServer, "xds-server", "", "Xds host to set on auto provisioned gateway resources")
-	controllerCmd.Flags().Uint16Var(&cfg.XdsPort, "xds-port", 8080, "Xds port to set on auto provisioned gateway resources")
+	// controllerCmd.Flags().StringVar(&cfg.XdsServer, "xds-server", "", "Xds host to set on auto provisioned gateway resources")
```
lgadban, line 28: do we need a TODO or anything here to track?

projects/gateway2/deployer/deployer.go
```diff
@@ -186,6 +183,9 @@ func (d *Deployer) GetObjsToDeploy(ctx context.Context, gw *api.Gateway) ([]clie
 	// Set owner ref
 	trueVal := true
 	for _, obj := range objs {
+		fmt.Printf("xxxxx objToDeploy: kind=%v, namespace=%s, name=%s\n", obj.GetObjectKind(),
```
lgadban, line 186: nit: should this use a logger rather than `fmt.Printf()`?
I understand we will clean this up so fine if we punt, but this is definitely a valuable debug log to keep around

projects/gateway2/deployer/deployer_test.go
```diff
@@ -58,7 +58,7 @@ var _ = Describe("Deployer", func() {
 	)
 	BeforeEach(func() {
 		var err error
-		d, err = deployer.NewDeployer(scheme.NewScheme(), false, "foo", "xds", 8080)
+		d, err = deployer.NewDeployer(scheme.NewScheme(), false, "foo", 8080)
```
lgadban, line 61: should this still be hardcoded to `foo` ?

projects/gateway2/deployer/deployer_test.go
```diff
@@ -204,7 +204,7 @@ var _ = Describe("Deployer", func() {
 
 	It("should propagate version.Version to get deployment", func() {
 		version.Version = "testversion"
-		d, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", "xds", 8080)
+		d, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", 8080)
```
lgadban, line 207: same as above

projects/gateway2/deployer/deployer_test.go
```diff
@@ -311,21 +311,21 @@ var _ = Describe("Deployer", func() {
...
+				"namespace": gw.Namespace,
+			},
+		}))
 
 	})
 
 	It("support segmenting by release", func() {
 
-		d1, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", "xds", 8080)
+		d1, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", 8080)
```
lgadban, line 325: same

projects/gateway2/deployer/deployer_test.go
```diff
@@ -311,21 +311,21 @@ var _ = Describe("Deployer", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 
-		d2, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", "xds", 8080)
+		d2, err := deployer.NewDeployer(scheme.NewScheme(), false, "foo", 8080)
```
lgadban, line 328: same

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -95,7 +93,7 @@ spec:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "gloo-gateway.gateway.fullname" . }}
+  name: {{ include "gloo-gateway.gateway.serviceAccountName" . }}
```
lgadban, line 96: why?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -182,18 +175,18 @@ data:
                   "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
 
       clusters:
-        - name: {{ $xdsClusterName }}
+        - name: xds_cluster
```
lgadban, line 178: question in general around this -- why remove the existing logic?
just to simplify?

projects/gloo/pkg/xds/cache.go
```diff
@@ -2,70 +2,41 @@ package xds
...
+	return OwnerNamespaceNameID(owner, namespace, name)
 }
 
 // SnapshotCacheKeys returns a list with the SnapshotCacheKey for each Proxy
-func SnapshotCacheKeys(proxies v1.ProxyList) []string {
+func SnapshotCacheKeys(owner string, proxies v1.ProxyList) []string {
 	var keys []string
 	// Get keys from proxies
 	for _, proxy := range proxies {
 		// This is where we correlate Node ID with proxy namespace~name
```
lgadban, line 23: nit: this comment should updated to include owner convention?

install/test/helm_test.go
```diff
@@ -174,7 +174,12 @@ var _ = Describe("Helm Test", func() {
 			})
 
 			It("should have all resources marked with a namespace", func() {
-				prepareMakefile(namespace, helmValues{})
+				prepareMakefile(namespace, helmValues{
+					valuesArgs: []string{
```
sam-heilbron, line 178: I'm a bit concerned that we need to do this. Since this work is in flight (ie unfinished elemenets will be merged afterwards), why don't we default the gateway2 API to off? Then we can leave these tests unmodified

projects/gateway2/cmd/controller.go
sam-heilbron: I'd almost rather remove. Do we even intend to have users (developers) invoke just the gateway2 components from the command line? Given that we inject config from the gloo setup function now, I don't think anyone should even use this, so I'd be happy to delete it entirely

projects/gateway2/controller/gw_controller.go
```diff
@@ -74,6 +74,8 @@ func (r *gatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 	}
 
+	log.Info("xxxxx deploying objects", "Objects", objs)
```
sam-heilbron, line 77: nit: This is ok for now, though I imagine we'd want to do at the debug level. I think we'll have to regroup on how aggressive we should be logging. I like the vision for how we can debug inconsistencies, but given the eventually consistent nature of edge, I fear this will flood the control plane logs

projects/gateway2/deployer/deployer_test.go
sam-heilbron: tbh it feels a bit unnecessary that we expose the "controllerName" in the deployer constructor API. This will always be a gloo-gateway deployer

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
sam-heilbron: Correct. This mirrors the way that we setup the xds_cluster and rest_xds_cluster. I found it more confusing with a dynamic value, when in reality all we need is a unique name, and to have the cluster name and load assignment match

projects/gateway2/tests/conformance/test-values.yaml
```diff
@@ -0,0 +1,5 @@
+gloo:
```
sam-heilbron, line 1: nit: I think we should be applying more of our "standard" helm values as well when we install the gateway. I had created `projects/gateway2/examples/helm-install.yaml` with some values pulled from other kube2e tests

projects/gateway2/xds/xds_syncer.go
```diff
@@ -229,13 +229,13 @@ func (s *XdsSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapshot) rep
 			allKeys[key] = false
 		}
 		// Get all valid node ID keys for Proxies
-		for _, key := range xds.SnapshotCacheKeys(snap.Proxies) {
+		for _, key := range xds.SnapshotCacheKeys(xds.GlooGatewayPrefix, snap.Proxies) {
```
sam-heilbron, line 232: This is the most interesting part of this PR that I'd like to learn better. Could you map out in excalidraw (or somewhere) what memory is shared and why this is necessary? In theory the keys should be unique so I'm curious why we need to add an additional prefix

projects/gloo/pkg/xds/cache.go
```diff
@@ -2,70 +2,41 @@ package xds
...
-func SnapshotCacheKey(proxy *v1.Proxy) string {
+func SnapshotCacheKey(owner string, proxy *v1.Proxy) string {
 	namespace, name := proxy.GetMetadata().Ref().Strings()
-	return fmt.Sprintf("%v~%v", namespace, name)
+	return OwnerNamespaceNameID(owner, namespace, name)
 }
 
 // SnapshotCacheKeys returns a list with the SnapshotCacheKey for each Proxy
-func SnapshotCacheKeys(proxies v1.ProxyList) []string {
+func SnapshotCacheKeys(owner string, proxies v1.ProxyList) []string {
```
sam-heilbron, line 19: Shouln't the "owner" be discoverable on the proxy resource itself as an annotation? That way this function can just parse the proxies and determine the owner from wherever we define it

projects/gloo/pkg/xds/node_hash.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Prefix indicating which translator created this Proxy.
+const ClassicEdgePrefix = "edge"
+const GlooGatewayPrefix = "gateway"
+
+// OwnerNamespaceNameID returns the string identifier for an Envoy node in a provided namespace.
+// Envoy proxies are assigned their configuration by Gloo based on their Node ID.
+// Therefore, proxies must identify themselves using the same naming
+// convention that we use to persist the Proxy resource in the snapshot cache.
+// The naming convention that we follow is "OWNER~NAMESPACE~NAME"
```
sam-heilbron, line 26: When we had previous conversations around this, I thought we agreed that while naming collision could occur, we didn't think it would ever actually happen. Can you help me understand how our previous conclusion was incorrect?

projects/gloo/pkg/xds/cache.go
jenshu: good point, we can get it from there assuming all proxies in the list were created by the same translator (which in our case, they are)

projects/gloo/pkg/xds/node_hash.go
jenshu: 1. i don't think we can assume proxy names won't collide, since they can be user-provided (it's based on the k8s gateway name for GG, and the edge gateway name for GE) and while maybe not realistic in practice, it's certainly possible for it to happen
2. but the main driver for adding owner/translator to the name was the fact that each envoy syncer (GE and GG) has a garbage collection step where it checks the list of current Proxies that it has, and deletes all others from the xds cache. This causes issues when GE and GG run together, because they delete each other's snapshots. Now with the owner/translator name, they know to only delete their own snapshots

projects/gateway2/controller/translator.go
jenshu: took note for future improvement

ci/kind/setup-kind.sh
jenshu: think i got this from the [quickstart docs](https://docs.solo.io/gloo-gateway/v2/quickstart/) - helm tab

projects/gloo/pkg/xds/node_hash.go
sam-heilbron: awesome discovery!

install/helm/gloo/values-template.yaml
jenshu: removed 387384c0fb23700a817e820d80d1d1b0eb64261f

projects/gateway2/cmd/controller.go
jenshu: removed c40de0e2c69227cc27290d3b84f879ec52f8fb96

projects/gateway2/tests/conformance/test-values.yaml
jenshu: i'm not sure if it will make a difference for the conformance tests since they don't really rely on any gloo edge stuff like settings

projects/gateway2/controller/gw_controller.go
jenshu: yeah i need to clean up all these logging statements

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
jenshu: this is the value that's used a few lines above in the deployment `serviceAccountName`, and also it seems to be the right value because we expose a `Values.gateway.serviceAccount.name` (which that helper function uses) and it seems like it should be used here

.github/workflows/pr.yaml
```diff
@@ -34,7 +34,13 @@ jobs:
       with:
         install_only: true
         version: v0.20.0
-    - name: Setup Env in KinD
-      run: make -C ./projects/gateway2/ build docker kind
+    - name: Set up cluster and helm chart
+      run: ./ci/kind/setup-kind.sh
+    - name: Additional cluster setup
+      run: ./projects/gateway2/kind.sh
+    - name: Install Gloo Gateway
```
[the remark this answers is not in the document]
jenshu: do you mean adding it to the regression test matrix?

install/helm/gloo/values-template.yaml
```diff
@@ -2,6 +2,124 @@ namespace:
   create: false
 k8s:
   clusterName: cluster.local
+
+####### start gateway2
+gateway2:
```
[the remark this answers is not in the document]
jenshu: cleaned up in 387384c0fb23700a817e820d80d1d1b0eb64261f

.github/workflows/pr.yaml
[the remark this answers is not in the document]
sam-heilbron: Kind of. I more mean that all of our other kube2e tests work by:
- Using the kind script to seed a cluster
- Configure some onfailure options to capture the state of a cluster if a test fails
- Using a SetupSuite node to install gloo edge in the cluster
- Run tests directly

The conformance tests differ, in that we perform most of those actions in this script. Again, not a requirement for this PR, just something to consider as we move forward

install/test/helm_test.go
jenshu: reverted all except the namespaced rbac test 9b04a54771a2db5cc75e6457cd88b1276b89c7f3

projects/gloo/pkg/xds/cache.go
jenshu: get owner from proxy 16c7d7678294f63788c9d9b26840248acfc5f631

projects/gateway2/deployer/deployer_test.go
jenshu: hmm not sure if we should remove that yet, we do expose controllerName and className as helm values so if we want to hardcode, we should remove all the places where it's currently configurable

projects/gloo/pkg/xds/cache.go
jenshu: updated in 16c7d7678294f63788c9d9b26840248acfc5f631

.github/workflows/pr.yaml
```diff
@@ -34,7 +34,13 @@ jobs:
       with:
         install_only: true
         version: v0.20.0
-    - name: Setup Env in KinD
-      run: make -C ./projects/gateway2/ build docker kind
+    - name: Set up cluster and helm chart
+      run: ./ci/kind/setup-kind.sh
+    - name: Additional cluster setup
+      run: ./projects/gateway2/kind.sh
```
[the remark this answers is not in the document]
jenshu: shift to the existing setup-kind script, you mean? if so, yes probably

ci/kind/setup-kind.sh
lgadban: sounds right, just pointing out that previously during the dev flow we were installing the experimental APIs.
since conformance is passing, it's fine, but will need to revisited when we start implementing the experimental channel

install/helm/gloo/generate/values.go
```diff
@@ -10,7 +10,9 @@ type HelmConfig struct {
 }
 
 type Config struct {
-	Namespace      *Namespace              `json:"namespace,omitempty"`
+	Namespace *Namespace `json:"namespace,omitempty"`
+	// This is an Alpha API and is subject to change in subsequent 1.17 beta releases
+	Gateway2       interface{}             `json:"gateway2,omitempty" desc:"Gloo Gateway settings, subject to change."`
```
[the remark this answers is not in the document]
lgadban: previously it was exposed to users via helm values -- at least in the standalone chart.
We need to do the same now, where we support helm values in the top-level gloo chart that can tune the deployer (and therefore the dynamic proxies)
IMO, we should track this somewhere (maybe in the follow ups) as it is a regression of what was available in the standalone ggv2 install

projects/gateway2/deployer/deployer_test.go
lgadban: my point is more that I see the change from "xds" to "foo" but I don't actually see it being checked/asserted anywhere. Is this intentional?

projects/gateway2/controller/controller.go
```diff
@@ -84,8 +86,8 @@ func (c *controllerBuilder) watchGw(ctx context.Context) error {
 	// setup a deployer
 	log := log.FromContext(ctx)
 
-	log.Info("creating deployer", "ctrlname", c.cfg.ControllerName, "server", c.cfg.XdsServer, "port", c.cfg.XdsPort)
-	d, err := deployer.NewDeployer(c.cfg.Mgr.GetScheme(), c.cfg.Dev, c.cfg.ControllerName, c.cfg.XdsServer, c.cfg.XdsPort)
+	log.Info("creating deployer", "ctrlname", c.cfg.ControllerName, "server", c.cfg.ControlPlane.GetBindAddress())
+	d, err := deployer.NewDeployer(c.cfg.Mgr.GetScheme(), c.cfg.Dev, c.cfg.ControllerName, c.cfg.ControlPlane.GetBindPort())
```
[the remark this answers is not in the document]
lgadban: also why was the port removed from the log?

projects/gateway2/deployer/deployer_test.go
lgadban: Ignore comment, I misread, so we're just removing the xds string as we no longer take the host (but still take in the port)
I think this needs cleanup in general but fine for now.

My question now boils down to "foo" in these tests is pointless right? We don't actually test based on the controller name?
A super-nit would be lets just change it to empty string, but also fine to ignore

jenshu: noted, can clean up later

install/helm/gloo/generate/values.go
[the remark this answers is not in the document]
jenshu: noted, will track it somewhere

projects/gateway2/controller/controller.go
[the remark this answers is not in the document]
jenshu: the port was already in the bind address so it was redundant, but we can add it back too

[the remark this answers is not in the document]
jenshu: added back log in f4e1f5922b96ac519cbe064602948db38675da85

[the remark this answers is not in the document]
sam-heilbron: Yeah so in theory (things will break if this is not the case) the ControlPlane instance needs to mirror the Gloo Service port that exposes that API. 

Our goal here is to inject the value of the xds socket address, so that when the deployer creates a proxy dynamically, it correctly configures the xds address of gloo. This way on startup, that proxy can get configuration from the control plane.

Before we were approaching this with some port magic (ie hope that the values stay in sync, which is fairly safe, because we don't really allow users to configure the control plane port).

However, since there is go code to actually expose a listener on a port (https://github.com/solo-io/gloo/blob/combine-binaries/projects/gloo/pkg/syncer/setup/setup_syncer.go#L161) I figured we could re-use that code to identify the true identity of the "control plane"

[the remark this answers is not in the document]
sam-heilbron: If there is a better way to pass around the specifics of the ControlPlane, I'm all ears

projects/gloo/pkg/xds/cache.go
jenshu: actually had to revert this since it caused some test failures. maybe some proxies didn't have the expected labels. but i did update to use the existing translator name consts instead of defining new ones

.github/workflows/pr.yaml
```diff
@@ -34,7 +34,13 @@ jobs:
...
+    - name: Install Gloo Gateway
+      run: helm upgrade -i -n gloo-system gloo ./_test/gloo-1.0.0-ci.tgz --create-namespace -f ./projects/gateway2/tests/conformance/test-values.yaml
```
npolshakova, line 42: In a follow up we should maybe make this a top-level Make target?

install/helm/gloo/templates/44-gatewayclass.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.gateway2.controlPlane.enabled -}}
+kind: GatewayClass
+apiVersion: gateway.networking.k8s.io/v1
+metadata:
+  labels:
+    {{- include "gloo-gateway.gateway.const_labels" . | nindent 4 }}
```
npolshakova, line 6: [nit] Do we want to keep the camel casing here consistent and change `const_labels` to `constLabels`?

projects/gateway2/deployer/builder.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func WithDevMode(devMode bool) Option {
+	return func(p *properties) {
+		p.dev = devMode
+	}
+}
+
+// scheme *runtime.Scheme, dev bool, controllerName, host string, port uint16
```
npolshakova, line 71: [nit] remove comment?

projects/gateway2/controller/controller.go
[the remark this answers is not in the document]
lgadban: my question is basically does the ControlPlane bind port map to the xds port value exposed on the service?
see:

- https://github.com/solo-io/gloo/blob/combine-binaries/install/helm/gloo/templates/9-gateway-proxy-configmap.yaml#L173

- https://github.com/solo-io/gloo/blob/combine-binaries/install/helm/gloo/templates/2-gloo-service.yaml#L17

projects/gloo/pkg/xds/cache.go
lgadban: this comment fix was reverted -- not super important; resolve this comment if you don't want to add more churn by adding the update back in

projects/gateway2/deployer/builder.go
jenshu: removed 56857e2

.github/workflows/pr.yaml
jenshu: yes probably should

projects/gloo/pkg/xds/cache.go
jenshu: oops haha fixed 56857e2

install/helm/gloo/generate/values.go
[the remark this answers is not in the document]
sam-heilbron: We spoke in a huddle and agreed to handle this in a follow-up. @jenshu will track this work

install/test/helm_test.go
sam-heilbron: This value (`gateway2.controlPlane.enabled`) renders the two manifests related to RBAC. When this value is enabled, the unit tests fail.

We agreed that we need to regroup on supporting namespace RBAC after this, but it is not a blocker for this PR to merge.

sam-heilbron (approved):
LGTM!

lgadban (approved):
LGTM!

Issue #8554, opened 2023-08-08
k8s regression test (istio) flake


sheidkamp:
### Which tests failed?

Failed test was `[It] with matching port and target port`  in /home/runner/work/gloo/gloo/test/kube2e/istio/istio_integration_test.go:113

### Initial Investigation

State never moves from "Pending" to "Accepted":

```
  [FAILED] Timed out after 30.001s.
  Expected
      <string>: Status
  to match fields: {
  .State:
  	Expected
  	    <core.Status_State>: 0
  	to equal
  	    <core.Status_State>: 1
  }
  ```

### Additional Information

failure https://github.com/solo-io/gloo/actions/runs/5798585599/attempts/1

**Comment by inFocus7** [comment]
Seen in [this run](https://github.com/solo-io/gloo/actions/runs/6512069557/job/17689008413?pr=8795)

**Comment by sheidkamp** [comment]
Hit this one again: https://github.com/solo-io/gloo/actions/runs/7410687254/job/20163550838?pr=9030

Passed on the 4th try

**Comment by npolshakova** [comment]
It looks like this is caused by an issue with the eds watcher only triggering once. In the istio integration test, it fails the status check when this error happens on the control plane due to the upstream not being picked up when the eds plugin is run:

```
{"level":"error","ts":"2024-03-11T23:07:00.229Z","logger":"gloo.v1.event_loop.setup.v1.event_loop.syncer.kubernetes_eds","caller":"kubernetes/eds.go:215","msg":"upstream gloo-system.gloo-system-testserver-1234: port 1234 not found for service testserver","version":"1.0.0-ci","stacktrace":"github.com/solo-io/gloo/projects/gloo/pkg/plugins/kubernetes.(*edsWatcher).List\n\t/Users/ninapolshakova/solo/gloo/projects/gloo/pkg/plugins/kubernetes/eds.go:215\ngithub.com/solo-io/gloo/projects/gloo/pkg/plugins/kubernetes.(*edsWatcher).watch.func1\n\t/Users/ninapolshakova/solo/gloo/projects/gloo/pkg/plugins/kubernetes/eds.go:236\ngithub.com/solo-io/gloo/projects/gloo/pkg/plugins/kubernetes.(*edsWatcher).watch.func2\n\t/Users/ninapolshakova/solo/gloo/projects/gloo/pkg/plugins/kubernetes/eds.go:263"}
[...]

PR #9165, opened 2024-02-19
Add istio auto mtls support

npolshakova:
# Description

## API changes
- Added `autoMtlsEnabled` to Gloo settings under IstioOptions.
- Added `disableIstioAutoMtls` to Gloo UpstreamSslConfig to explicitly disable auto mtls on a particular Upstream. This supports istio PeerAuthorization policies with mtls mode set to `mode: DISABLE`. 

## Code changes
- Cluster translator will add two transport socket matches for tlsMode and plaintext if istioAutoMtls is enabled
- Kubernetes eds plugin will add the istio tlsMode metadata in computeGlooEndpoints if istioAutoMtls is enabled to the endpoint. This has access to the Pods, which will have the istio injection label
- The same sds/istio-proxy helm values will control the sds sidecar and istio-proxy on the Gloo Edge side. The Gloo Gateway will have a temporary helm value to enable Istio integration until the helm chart merge is complete. 
- When the Upstream sslConfig is set, that config will overwrite istio auto mtls transport socket match configuration. 
- In the automtls case, if istioSds.Enabled is not set, we will error because this results in an invalid case (transport socket match is translated, but no sds cluster is present):

image

## CI changes
- [x] TODO: Add e2e test for Gloo Edge with automtls enabled
- [x] TODO: Add e2e test for Gloo Gateway with automtls enabled 

## Tests 

The new istio integration tests add a new test that edits the upstream. Since other tests are creating/deleting/editing upstream, we need to make sure the upstream is present before making a change. This is fixed in https://github.com/solo-io/gloo/pull/9165/commits/ada8df3053b13b49cadc31bf5d1c9d4743729df3, but eventually, we should move away from relying on Discovery to create the upstream and manually create/edit the upstreams.

```
Ran 11 of 11 Specs in 66.573 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

All tests passed...
Will keep running them until they fail.
This was attempt #12
Looking good!

```

## Docs changes
- TODO: Need to add Istio automtls docs. This should probably go in after helm chart clean to have more accurate helm chart values: https://github.com/solo-io/solo-projects/issues/5881

## Try it out:

To run the gloo-gateway (no istio) setup, run this in the `gloo/projects/gateway2` directory:

```
make build docker kind install-gloo-gateway
```

Create a Gateway resource and you should only see one container:
```
❯ kubectl apply -f - <<EOF
kind: Gateway
metadata:
  name: http
spec:
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: http
    port: 8080
    protocol: HTTP
EOF
```

To run the istio setup:

```
make build docker kind install-istio-gloo-gateway
```

This will install istio and bookinfo in the mesh. It will also apply the Gateway, HTTPRoute and AuthorizationPolicy to test mtls with the istio integration. 

Send traffic by port-forwarding the gateway:
```
❯ k port-forward deployment/gloo-proxy-http 8080:8080
```

And then sending a request through the gateway with the example HTTPRoute from the setup script:
```
❯ curl -I localhost:8080 -H "host: www.example.com" -v
```

# Demo:

https://drive.google.com/file/d/1Jonyy1AbHjS5GcUkgMZ-ZujvzgBpsFqL/view?usp=sharing 

# Checklist:

- [ ] I have made corresponding changes to the documentation (Will be done as part of https://github.com/solo-io/solo-projects/issues/5881)

Passing new ci workflow: https://github.com/solo-io/gloo/actions/runs/8114585555/job/22180563260?pr=9165 

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5695


projects/gateway2/README.md
```diff
@@ -0,0 +1,177 @@
+## Gloo Gateway
+
+# Test Locally 
+
+To create the local test environment in kind, run:
+
+```shell
+ CLUSTER_NAME=solo-test-cluster ci/kind/setup-kind.sh; CLUSTER_NAME=solo-test-cluster make kind-build-and-load
```
lgadban, line 8: nit: is there a compelling reason to use a custom cluster name rather than just relying on the default?

npolshakova: Nope- this is just copied from my notes for setting up the e2e test environment. The default should work!

projects/gateway2/helm/gloo-gateway/values.yaml
```diff
@@ -51,3 +51,22 @@ gateway:
     # The name of the service account to use.
     # If not set and create is true, a name is generated using the fullname template
     name: ""
+
+global:
+  # TODO(npolshak): Add support for other Istio integration values
+  istioSDS:
+    enabled: false
+  glooMtls:
```
npolshakova, line 59: TODO: remove glooMtls from values and template

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1281,18 +1281,35 @@ func constructOpts(ctx context.Context, params constructOptsParams) (bootstrap.O
...
 }
 
+func constructIstioBootstrapOpts() bootstrap.IstioValues {
+	istioValues := bootstrap.IstioValues{
+		// TODO: This value should be inherited at installation time, to determine if the k8s controller is enabled
+		// In the interim, we use an env variable to control the value
+		IstioSDS: bootstrap.IstioSDS{
+			Enabled: isEnvTruthy(constants.IstioMtlsEnabled),
+		},
+		GlooMtls: bootstrap.GlooMtls{
```
npolshakova, line 1305: TODO: remove Gloo Mtls

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -30,48 +31,199 @@ spec:
       securityContext:
         {{- toYaml .Values.gateway.podSecurityContext | nindent 8 }}
       containers:
-      - name: {{ .Chart.Name }}
-        securityContext:
+        - name: {{ .Chart.Name }}
```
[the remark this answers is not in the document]
npolshakova: Nope, I think this was just carried over from copying the templating.

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -7,88 +8,249 @@ metadata:
...
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.name
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: metadata.namespace
+          {{- if $gateway.istioSDS.enabled }}
```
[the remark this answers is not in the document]
npolshakova: Nope- just fixed this as part of fixing the indenting. It was left over from the old templating that checked glooMtls.enabled vs. istioSDS.enabled

projects/gloo/constants/istio.go
```diff
@@ -0,0 +1,22 @@
+package constants
+
+const (
+	// Env variable to enable the Istio mtls integration to add the istio-proxy sidecar
```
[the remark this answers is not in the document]
npolshakova: Yep, that's a good point!

projects/gloo/constants/istio.go
```diff
@@ -0,0 +1,22 @@
+package constants
```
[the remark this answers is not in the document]
npolshakova: Yeah I can add them here! Which ones are you thinking? Most of the constants here are currently values that get used in the sslConfig or automtls translation. I think `IstioMtlsEnabled` is the env value, but I can add `ENABLE_ISTIO_INTEGRATION` here as well.

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -1281,18 +1286,30 @@ func constructOpts(ctx context.Context, params constructOptsParams) (bootstrap.O
...
+		IstioValues:                constructIstioBootstrapOpts(),
+		GlooOptions:                settings.GetGloo(),
 	}
 }
 
+func constructIstioBootstrapOpts() bootstrap.IstioValues {
+	istioValues := bootstrap.IstioValues{
+		// TODO: This value should be inherited at installation time, to determine if the k8s controller is enabled
+		// In the interim, we use an env variable to control the value
+		IstioSDSEnabled: isEnvTruthy(constants.IstioMtlsEnabled),
```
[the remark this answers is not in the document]
npolshakova: Yep, I think that's a good idea- the helm value that this is based off isn't experimental but I think the env approach should be an interim solution!

projects/gloo/pkg/translator/clusters.go
```diff
@@ -116,24 +139,27 @@ func (t *translatorInstance) initializeCluster(
...
 
-	if sslConfig := upstream.GetSslConfig(); sslConfig != nil {
-		applyDefaultsToUpstreamSslConfig(sslConfig, t.settings.GetUpstreamOptions())
-		cfg, err := utils.NewSslConfigTranslator().ResolveUpstreamSslConfig(*secrets, sslConfig)
-		if err != nil {
-			reports.AddError(upstream, err)
-		} else {
-			typedConfig, err := utils.MessageToAny(cfg)
+	if sslConfig != nil {
+		if !sslConfigDisabled(sslConfig) {
```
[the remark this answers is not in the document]
npolshakova: Yep! At one point I wanted to add a log here that the automtls was disabled, but I think that's a pretty explicit user choice so I don't think we need it?

projects/gloo/pkg/translator/clusters.go
```diff
@@ -100,6 +104,25 @@ func (t *translatorInstance) initializeCluster(
...
+		} else {
+			// Istio automtls will only be applied when the upstream has no sslConfig and istio integration is enabled on the helm chart
+			socketmatches = []*envoy_config_cluster_v3.Cluster_TransportSocketMatch{
+				// add istio mtls match
+				createIstioMatch(),
+				// plaintext match. Note: this needs to come after the tlsMode-istio match
+				createDefaultIstioMatch(),
+			}
+		}
+	}
```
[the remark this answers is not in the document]
npolshakova: So I think this is more similar to the `sslConfigTranslator utils.SslConfigTranslator` that gets passed in. I'm not sure if a plugin makes sense since this depends on the sslConfig that gets fed into that translator. Maybe if there was a combined sslConfig plugin, that would make sense to go there? I could create a `utils.IstioAutoMtlsTranslator` that gets passed in here? 

But I think a plugin here would be messier because it doesn't seem like there are any ClusterPlugins currently and it feels like there is a lot of shared logic between calling the SslConfigTranslator vs. translating for automtls?

I might also be missing something on the edge side, so this might be easier to chat about while looking at the code on a call!

projects/gloo/pkg/translator/clusters.go
```diff
@@ -157,6 +183,105 @@ func (t *translatorInstance) initializeCluster(
...
+
+func createDefaultIstioMatch() *envoy_config_cluster_v3.Cluster_TransportSocketMatch {
+	// TODO: upstream has https://github.com/istio/istio/blob/fa321ebd2a1186325788b0f461aa9f36a1a8d90e/pilot/pkg/xds/filters/filters.go#L78
+	typedConfig, _ := utils.MessageToAny(&socketsRaw.RawBuffer{})
+	rawBufferTransportSocket := &envoy_config_core_v3.TransportSocket{Name: wellknown.TransportSocketRawBuffer,
+		ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: typedConfig},
+	}
+
+	return &envoy_config_cluster_v3.Cluster_TransportSocketMatch{
+		Name:            "tlsMode-disabled",
```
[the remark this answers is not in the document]
npolshakova: Good catch!

projects/gloo/pkg/translator/clusters.go
```diff
@@ -157,6 +183,105 @@ func (t *translatorInstance) initializeCluster(
...
+	}
+
+	typedConfig, _ := utils.MessageToAny(sslSds)
+	transportSocket := &envoy_config_core_v3.TransportSocket{
+		Name:       wellknown.TransportSocketTls,
+		ConfigType: &envoy_config_core_v3.TransportSocket_TypedConfig{TypedConfig: typedConfig},
+	}
+
+	return &envoy_config_cluster_v3.Cluster_TransportSocketMatch{
+		Name:            "tlsMode-" + constants.IstioMutualTLSModeLabel,
```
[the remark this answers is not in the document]
npolshakova: Yep, this was copied from the istio code, but I'm fine using the const for `tlsMode` here as well!

projects/gloo/pkg/translator/endpoints.go
```diff
@@ -168,6 +173,20 @@ func getLbMetadata(upstream *v1.Upstream, labels map[string]string, zeroValue st
 		return nil
 	}
 
+	if autoMtlsEnabled {
```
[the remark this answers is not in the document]
npolshakova: Do you think it's worth having a separate plugin for eds? I was on the fence here because it seemed a little overkill for just getting the metadata label?

projects/gloo/pkg/translator/clusters.go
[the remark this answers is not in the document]
npolshakova: Added the plugin approach and I think it's a lot cleaner now: https://github.com/solo-io/gloo/pull/9165/commits/0e97ecd68a6a289561e4bad2d26fb43adfb827c0

projects/gloo/pkg/translator/endpoints.go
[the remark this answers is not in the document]
npolshakova: Added a new `getEndpointMetadata` that adds both the istio metadata and keeps `getLbMetadata` separate

projects/gloo/pkg/translator/clusters.go
```diff
@@ -100,6 +100,9 @@ func (t *translatorInstance) initializeCluster(
 	}
 
 	circuitBreakers := t.settings.GetGloo().GetCircuitBreakers()
+
+	sslConfig := upstream.GetSslConfig()
```
lgadban, line 104: do we still need this change now that the auto mtls stuff is in a plugin?

projects/gloo/pkg/translator/clusters.go
```diff
@@ -157,6 +160,16 @@ func (t *translatorInstance) initializeCluster(
 	return out
 }
 
+func sslConfigDisabled(sslConfig *ssl.UpstreamSslConfig) bool {
```
lgadban, line 163: can we add a godoc and rename this function to be a bit more descriptive?
I'm struggling to see the reasoning for this -- are we saying that if the sslConfig is NOT nil but autoMtls is disabled AND all the values on the sslConfig are the zero-values, return true?
we might want to talk this one through

projects/gloo/pkg/plugins/istio_automtls/plugin.go
```diff
@@ -0,0 +1,162 @@
...
+	p.ctx = params.Ctx
+}
+
+func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *envoy_config_cluster_v3.Cluster) error {
+	var socketmatches []*envoy_config_cluster_v3.Cluster_TransportSocketMatch
+
+	sslConfig := in.GetSslConfig()
+	// if the upstream has mtls enabled, and the sslConfig is not nil, then we need to add the istio mtls match
+	// If the upstream has sslConfig, the upstream config will overwrite the auto mtls config
+	if p.settings.GetGloo().GetIstioOptions().GetEnableAutoMtls() && sslConfig == nil {
```
lgadban, line 57: do we not need to check the disableAutoMtls flag here?

lgadban (commented):
still need to review endpoints.go

projects/gloo/pkg/translator/clusters.go
npolshakova: We can move it down closer to the check, but I think this should be on it's own line to make the if statement cleaner.

projects/gloo/pkg/plugins/istio_automtls/plugin.go
npolshakova: If the `disableAutoMtls` is set it'll also result in a non-nil `sslConfig`

projects/gloo/pkg/translator/clusters.go
npolshakova: Yeah let's talk through this one- currently if the sslConfig is not nil, the SslConfigTranslator will return a envoyauth.UpstreamTlsContext [here](https://github.com/solo-io/gloo/blob/112ed69f2156b2456da4af6ef4f3c1fc89809adc/projects/gloo/pkg/utils/ssl.go#L78). So if just `disableAutoMtls` is set, it breaks regular upstream. So this check if `GetDisableIstioAutoMtls` is set AND no other sslConfig is set, skip translation.

install/helm/gloo/values-template.yaml
```diff
@@ -18,6 +18,9 @@ settings:
   create: true
   linkerd: false
   enableRestEds: false
+  gloo:
+    istioOptions:
+      enableAutoMtls: false
```
yuval-k, line 23: where is this used? i don't think this is automatically merged into the settings in the helm chart? I looked at `install/helm/gloo/templates/18-settings.yaml`

yuval-k: (asking as i would like to be able to enabled this in the helm chart)

npolshakova: Ah good catch- currently setting here in the helm chart doesn't do anything. It needs to be set in the settings CR. I'll add it to `18-settings.yaml` and the helm chart

projects/gloo/pkg/plugins/istio_automtls/plugin.go
```diff
@@ -0,0 +1,162 @@
...
+func (p *plugin) Init(params plugins.InitParams) {
+	p.settings = params.Settings
+	p.ctx = params.Ctx
+}
+
+func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *envoy_config_cluster_v3.Cluster) error {
+	var socketmatches []*envoy_config_cluster_v3.Cluster_TransportSocketMatch
+
+	sslConfig := in.GetSslConfig()
+	// if the upstream has mtls enabled, and the sslConfig is not nil, then we need to add the istio mtls match
```
lgadban, line 55: this doesn't sound right?

lgadban:
```suggestion
	// if istio auto mtls is enabled, and the sslConfig is nil, then we need to add the istio mtls match
```

projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -468,12 +487,21 @@ func generateFilteredEndpointList(
 
 		var ep *v1.Endpoint
 		// While istio integration requires the Service VIP, headless services require the pod IP, as there is no Cluster IP
-		if istioIntegrationEnabled && !addr.IsHeadless {
+		if istioIntegrationEnabled && !addr.IsHeadless && !autoMtls {
```
lgadban, line 490: this is confusing to me; should we be modifying this conditional to include the autoMtls logic as well?
seems like we're crossing concerns a bit

npolshakova: Addressed in https://github.com/solo-io/gloo/pull/9165/commits/6284d3f51079e2ed9a712a3410b104c3670a4ace after chating

projects/gloo/pkg/plugins/istio_automtls/plugin.go
npolshakova: Addressed in https://github.com/solo-io/gloo/pull/9165/commits/6284d3f51079e2ed9a712a3410b104c3670a4ace after chatting

projects/gloo/pkg/translator/clusters.go
npolshakova: resolved moved disableAutoMtls to upstream instead of sslConfig

projects/gloo/pkg/translator/endpoints.go
```diff
@@ -45,11 +50,12 @@ func (t *translatorInstance) computeClusterEndpoints(
 func loadAssignmentForUpstream(
 	upstream *v1.Upstream,
 	clusterEndpoints []*v1.Endpoint,
+	enableAutoMtls bool,
 ) *envoy_config_endpoint_v3.ClusterLoadAssignment {
 	clusterName := UpstreamToClusterName(upstream.GetMetadata().Ref())
 	var endpoints []*envoy_config_endpoint_v3.LbEndpoint
 	for _, addr := range clusterEndpoints {
-		metadata := getLbMetadata(upstream, addr.GetMetadata().GetLabels(), "")
+		metadata := getEndpointMetadata(upstream, addr.GetMetadata().GetLabels(), "", enableAutoMtls)
```
lgadban, line 58: nit: feels like we shouldn't change the entrypoint (see below comment)

projects/gloo/pkg/translator/endpoints.go
```diff
@@ -172,6 +177,27 @@ func getLbMetadata(upstream *v1.Upstream, labels map[string]string, zeroValue st
 	return meta
 }
 
+func getEndpointMetadata(upstream *v1.Upstream, labels map[string]string, zeroValue string, enableAutoMtls bool) *envoy_config_core_v3.Metadata {
```
lgadban, line 180: IMO this should just be `getEndpointMetadata(labels map[string]string)` and conditionally called from `getLbMetadata(...)` above

lgadban (approved):
added a nit, won't hold up approval over it, seems like the factoring can be improved in  endpoints.go but don't feel too strongly.
LMK if you want to talk through it

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -32,34 +16,59 @@ import (
...
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	. "github.com/solo-io/solo-kit/test/matchers"
+
 	"github.com/solo-io/gloo/pkg/utils/api_conversion"
 	"github.com/solo-io/gloo/pkg/utils/settingsutil"
 	"github.com/solo-io/gloo/projects/gloo/constants"
 	gloo_envoy_core "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/api/v2/core"
+	v3 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/config/core/v3"
+	envoy_v3 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/filters/http/buffer/v3"
+	csrf_v31 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/filters/http/csrf/v3"
```
sam-heilbron, line 50: nit: I think some of these aliases could be a bit more expressive

test/kube2e/istio/artifacts/gloo-gateway-helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+gateway:
+  persistProxySpec: true
+  logLevel: info
+  validation:
+    allowWarnings: true
+    alwaysAcceptResources: false
+gateway2:
+  gatewayClass:
+    className: gloo-gateway
+    controllerName: solo.io/gloo-gateway
```
sam-heilbron, line 50: It's probably fine to leave for now, but right now these Helm API values do nothing

sam-heilbron (dismissed):
LGTM! A few nits/questions, but feel free to resolve them and either not handle it, or handle in a follow-up

projects/gloo/cli/pkg/cmd/install/v2/install.go
```diff
@@ -47,12 +52,15 @@ func install(opts *options.Options, installOpts *Options) error {
 		return err
 	}
 
+	// TODO(npolshak): Need a better way for the cli to access helm values here
```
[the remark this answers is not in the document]
npolshakova: And just not provide the `IstioValues`? Yeah that works!

.github/workflows/composite-actions/istio-tests/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+      uses: ./.github/workflows/composite-actions/prep-go-runner
+    - name: Install kind
+      uses: helm/kind-action@v1.5.0
+      with:
+        install_only: true
+        version: ${{ matrix.kube-version.kind }}
+    - uses: azure/setup-kubectl@v3
+      id: kubectl
+      with:
+        version: ${{ matrix.kube-version.kubectl }}
```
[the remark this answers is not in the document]
npolshakova: Yeah I think we might want to add the istio versions here as part of the follow up issue to add test coverage for istio versions, so we still might want it as a separate action.

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -235,8 +235,12 @@ spec:
           - name: LOG_LEVEL
             value: {{ .Values.gloo.logLevel }}
         {{- end}}
+        {{- if .Values.global.istioSDS.enabled }}
+          - name: GG_EXPERIMENTAL_ISTIO_MTLS_SDS_ENABLED
+            value: "true"
+        {{- end}}
         {{- if .Values.global.istioIntegration.enableIstioSidecarOnGateway }}
-          - name: ENABLE_ISTIO_INTEGRATION
```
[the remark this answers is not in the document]
npolshakova: Yep, now it's: 
```
lookupResult, found := os.LookupEnv(constants.IstioInjectionEnabled)
```

projects/gateway2/deployer/deployer.go
```diff
@@ -55,6 +55,8 @@ type Inputs struct {
 	ControllerName string
 	Dev            bool
 	Port           int
+	Settings       *v1.Settings
```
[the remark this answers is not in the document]
npolshakova: Yep- settings isn't currently used in the deployer so I can remove!

projects/gateway2/deployer/deployer_test.go
```diff
@@ -68,6 +70,62 @@ var _ = Describe("Deployer", func() {
...
+			Dev:            false,
+			ControllerName: "foo",
+			Port:           8080,
+			Settings: &v1.Settings{
+				Gloo: &v1.GlooOptions{
+					IstioOptions: &v1.GlooOptions_IstioOptions{
+						EnableAutoMtls: &wrappers.BoolValue{Value: true},
+					},
+				},
+			},
```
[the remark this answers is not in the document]
npolshakova: I don't think it's needed anymore! Will remove

projects/gloo/api/v1/upstream.proto
```diff
@@ -184,7 +184,9 @@ message Upstream {
     // In most cases preconnect hurts more than it helps.
     PreconnectPolicy preconnect_policy= 32;
 
-
+    // If set to true, the proxy will not allow automatic mTLS detection for Istio upstreams.
```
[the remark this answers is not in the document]
npolshakova: How does this sound? 

```
 If set to true, the proxy will not allow automatic mTLS detection for Gloo Upstreams selecting Istio injected workloads.
```

projects/gloo/pkg/translator/translator_test.go
npolshakova: I don't think these are net new- just reorganized the imports. But I can rename them to be more consistent with the rest of the code base: 

```
	bufferv3 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/filters/http/buffer/v3"
	gloocsrf "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/extensions/filters/http/csrf/v3"
```

It looks like `v3 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/envoy/config/core/v3"` is used as `v3` everywhere, so I'm leaning towards leaving it.

test/kube2e/istio/artifacts/gloo-gateway-helm.yaml
npolshakova: Yep, copied over from the conformance tests, but I think these are just the defaults?

projects/gloo/pkg/translator/endpoints.go
npolshakova: Some of the earlier feedback I got was having `getLbMetadata` call Istio specific logic made it hard to follow what `getLbMetadata` was responsible for. How about addressing https://github.com/solo-io/gloo/pull/9165#discussion_r1518396454 instead? So it would still be getEndpointMetadata but change the function that gets passed

projects/gloo/pkg/translator/route_config.go
```diff
@@ -554,13 +560,13 @@ func (m *multiRouteConfigurationTranslator) ComputeRouteConfiguration(params plu
...
 // TODO(marco): when we update the routing API we should move this to a RouteActionPlugin
-func getSubsetMatch(destination *v1.Destination) *envoy_config_core_v3.Metadata {
+func getSubsetMatch(destination *v1.Destination, enableAutoMtls bool) *envoy_config_core_v3.Metadata {
 	var routeMetadata *envoy_config_core_v3.Metadata
 
 	// TODO(yuval-k): should we add validation that the route subset indeed exists in the upstream?
 	// First convert the subset information on the base destination, if present
 	if destination.GetSubset() != nil {
-		routeMetadata = getLbMetadata(nil, destination.GetSubset().GetValues(), "")
+		routeMetadata = getEndpointMetadata(nil, destination.GetSubset().GetValues(), "", enableAutoMtls)
```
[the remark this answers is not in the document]
npolshakova: I think if istio auto mtls is enabled, we want to get both `getLbMetadata` and the istio metadata. The istio metadata is set for `envoy.transport_socket_match` and the load balance metadata is set for `envoy.lb`.

sam-heilbron (dismissed):

projects/gloo/pkg/translator/route_config.go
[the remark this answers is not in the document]
lgadban: how confident are we that the istio metadata is relevant here? Seems like this is mainly for subset loadbalancing, which is a separate concern from what istio automtls is actually providing

projects/gloo/pkg/translator/route_config.go
```diff
@@ -82,6 +82,7 @@ type httpRouteConfigurationTranslator struct {
 	report                   *validationapi.HttpListenerReport
 	routeConfigName          string
 	requireTlsOnVirtualHosts bool
+	enableAutoMtls           bool
```
lgadban, line 85: no longer needed right? same with all changes to this file now it looks like

lgadban (changes requested):

projects/gloo/pkg/translator/translator.go
```diff
@@ -56,6 +55,11 @@ type translatorInstance struct {
 	listenerTranslatorFactory *ListenerSubsystemTranslatorFactory
 }
 
+type IstioTranslatorConfig struct {
```
lgadban, line 58: what is this used for?

npolshakova: I think this can be removed- the auto mtls plugin now is initialized with these values

npolshakova:
https://github.com/solo-io/gloo/actions/runs/8238795640/job/22541536443?pr=9165 flaked due to https://github.com/solo-io/gloo/issues/8554. Since we're adding a new istio regression run, https://github.com/solo-io/gloo/pull/9165/commits/136a43d3963de84cdfc96a11da6a56017a6c08d5 mitigates the status flake:

```
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 6 Skipped
PASS | FOCUSED

All tests passed...
Will keep running them until they fail.
This was attempt #31
No, seriously... you can probably stop now.
```

projects/gateway2/discovery/translator.go
```diff
@@ -25,19 +25,21 @@ type Translator interface {
 	ReconcileEndpoints(ctx context.Context, req ctrl.Request) (ctrl.Result, error)
 }
 
-func NewTranslator(cli client.Client, inputChannels *xds.XdsInputChannels) Translator {
+func NewTranslator(cli client.Client, inputChannels *xds.XdsInputChannels, enableIstioAutoMtls bool) Translator {
```
lgadban, line 28: what does `enableIstioAutoMtls` do now? seems unused...

lgadban (changes requested):

.github/workflows/istio-tests.yaml
```diff
@@ -0,0 +1,67 @@
...
+    needs: prepare_env
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    if: ${{ !github.event.pull_request.draft }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # upgrade tests are run on LTS but not on main branch, for main they are run nightly
+        # this is the github action version of ternary op
+        kube-e2e-test-type: [ 'istio-auto', 'gloo-gateway' ]
```
lgadban, line 52: does this mean we only test legacy with istio-auto now?

projects/gateway2/controller/controller.go
```diff
@@ -34,6 +34,8 @@ type GatewayConfig struct {
 	Kick           func(ctx context.Context)
 
 	ControlPlane bootstrap.ControlPlane
+	Settings     *v1.Settings
```
sam-heilbron, line 37: I don't see this used anywhere in this component, can we remove it?

projects/gateway2/controller/start.go
```diff
@@ -111,15 +106,17 @@ func Start(ctx context.Context, cfg StartConfig) error {
...
 		Kick:           inputChannels.Kick,
+		Settings:       cfg.Opts.Settings,
 	}
 	if err = NewBaseGatewayController(ctx, gwCfg); err != nil {
 		setupLog.Error(err, "unable to create controller")
 		return err
 	}
 
-	if err = discovery.NewDiscoveryController(ctx, mgr, inputChannels); err != nil {
+	if err = discovery.NewDiscoveryController(ctx, mgr, inputChannels, cfg.Opts.Settings.GetGloo().GetIstioOptions().GetEnableAutoMtls().GetValue()); err != nil {
```
sam-heilbron, line 119: This new `enableAutoMtls` value is not used inside of this discovery controller, can we remove it?

projects/gloo/pkg/bootstrap/opts.go
```diff
@@ -65,8 +62,17 @@ type Opts struct {
 	GlooGateway GlooGateway
 }
 
+type IstioValues struct {
+	// Based on istioSDS helm values. Config used for installing Gloo Edge with Istio SDS cert rotation features for Istio mTLS
+	IstioSDSEnabled bool
```
sam-heilbron, line 67: nit (non-blocking): The name here stutters (https://sben.dev/stutter/) a bit. Given that these are IstioValues, could we just remove the `istio` prefix from the field names?

projects/gloo/pkg/bootstrap/opts.go
```diff
@@ -65,8 +62,17 @@ type Opts struct {
...
+	// Based on enableIstioSidecarOnGateway helm value. If true, the gateway will have the istio sidecar injected.
+	IstioSidecarOnGatewayEnabled bool
+}
+
 type GlooGateway struct {
 	EnableK8sGatewayController bool
+	IstioValues                IstioValues
+	GlooOptions                *v1.GlooOptions
```
sam-heilbron, line 75: I can't see these values being used, am I missing something?

projects/gloo/pkg/plugins/istio_automtls/istio_automtls_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package istio_automtls_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestGzip(t *testing.T) {
```
sam-heilbron, line 10:
```suggestion
func TestIstioAutoMtls(t *testing.T) {
```

projects/gloo/pkg/plugins/istio_automtls/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+func (p *plugin) Name() string {
+	return ExtensionName
+}
+
+func (p *plugin) Init(params plugins.InitParams) {
+	p.settings = params.Settings
+	p.ctx = params.Ctx
```
sam-heilbron, line 55: Each of the plugin functions, in this instance ProcessUpstream, should pass params that have a context on them. Is there a reason we want to persist the context on the plugin instead of just using what is passed in? The go recommendation (https://go.dev/blog/context-and-structs) is to avoid storing context on a struct, and prefer to use it passed as an argument

.github/workflows/istio-tests.yaml
npolshakova: No, the regression tests run the normal Istio test with no auto mtls: https://github.com/solo-io/gloo/actions/runs/8251250967/job/22569675558?pr=9165. 

This new suite will eventually have the matrix of different Istio versions.

projects/gateway2/controller/controller.go
npolshakova: Yeah I think this got added back in the merge 😢

projects/gloo/pkg/bootstrap/opts.go
npolshakova: Yep, it was supposed to reflect the helm value back when this was a copy of the istio helm values from values.go. Can rename!

projects/gloo/pkg/bootstrap/opts.go
npolshakova: I think `GlooOptions` isn't getting used anymore either- the plugin now has access to the settings and is getting the autoMtlsEnabled value from that.

projects/gloo/pkg/plugins/istio_automtls/plugin.go
npolshakova: The azure plugin does this as well, so that's where this came from. But I think all the uses of the ctx use the param context and that makes more sense that using the one that's persisted on the plugin. Especially any logging happens on `ProcessUpstream` and there's no logging on the plugin initialization.

.github/workflows/istio-tests.yaml
lgadban: perfect, ty!

projects/gloo/pkg/plugins/kubernetes/eds_test.go
```diff
@@ -56,19 +64,109 @@ var _ = Describe("Eds", func() {
...
 
-			os.Unsetenv("ENABLE_ISTIO_INTEGRATION")
-			Expect(isIstioIntegrationEnabled()).To(BeFalse())
+			os.Setenv(constants.IstioInjectionEnabled, "false")
+			Expect(isIstioInjectionEnabled()).To(BeFalse())
+		})
 
-			os.Setenv("ENABLE_ISTIO_INTEGRATION", "false")
-			Expect(isIstioIntegrationEnabled()).To(BeFalse())
+		It("autoMtls should translate EDS metadata", func() {
```
lgadban, line 82: should this test really be in this file? there are no functional changes to `eds.go` and the behavior tested here is in `translator/endpoints.go`.
If this does make sense then all good!

test/kube2e/istio/istio_suite_test.go
```diff
@@ -117,6 +181,44 @@ func installGloo() {
 	// Ensure gloo reaches valid state and doesn't continually resync
 	// we can consider doing the same for leaking go-routines after resyncs
 	kube2e.EventuallyReachesConsistentState(testHelper.InstallNamespace)
+
+	// Ensure discovery reaches a valid state
+	// Note: discovery is only used in the "classic", non-k8s-gateway api setup
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
```
lgadban, line 188: why?

lgadban (approved):

projects/gloo/pkg/plugins/kubernetes/eds_test.go
npolshakova: We don't need a net new test but should make sure the tlsMode label is picked up. This was added when we had the conditional parsing for the tlsMode data.

npolshakova: Actually, it doesn't look like any tests in this file check that `Metadata.Labels`. So yes, we need a test to check the labels are getting set. I can rename the test, but we do want to keep the checks here since it adds new coverage.

test/kube2e/istio/istio_suite_test.go
npolshakova: We should be able to reuse the test context, this was in a different file before.

projects/gloo/pkg/plugins/kubernetes/eds_test.go
```diff
@@ -56,19 +64,109 @@ var _ = Describe("Eds", func() {
...
+				map[*core.ResourceRef]*kubeplugin.UpstreamSpec{
+					up.Metadata.Ref(): up.GetKube(),
+				})
+
+			Expect(endpoints).To(HaveLen(1), "expected to have 1 endpoint")
+			Expect(warnsToLog).To(BeEmpty(), "expected no warnings")
+			Expect(errorsToLog).To(BeEmpty(), "expected no errors")
+
+			// Check endpoint has automtls metadata
+			Expect(endpoints[0].Metadata.Labels).To(HaveKeyWithValue(constants.IstioTlsModeLabel, "istio"))
```
lgadban, line 169: my point is more that the assertion here relies on the metadata changes present in the endpoint translator (i.e. `translator/endpoints.go`) so it seems like this functionality should be tested there, e.g. https://github.com/solo-io/gloo/blob/npolshak/istio-auto-mtls/projects/gloo/pkg/translator/translator_test.go#L1819

I definitely agree that we need to test this metadata propagation and i'm fine leaving as-is, more so calling out that it seems a little confusing to have this logic tested as part of the EDS plugin

lgadban (approved):
excellent work!

projects/gloo/pkg/plugins/kubernetes/eds_test.go
npolshakova: It's also tested in the translator tests, but I think have a direct call to `FilterEndpoints` is better than just relying the parent translator tests. If there's any issues with `FilterEndpoints` missing metadata it will be easier to debug here than the translator. It doesn't rely on any imports from the translator, it's only scoped to things in the eds.go file. 

https://github.com/solo-io/gloo/blob/npolshak/istio-auto-mtls/projects/gloo/pkg/translator/translator_test.go#L1819  runs through the whole translate().

sam-heilbron (approved):

npolshakova:
retrigger bulldozer

PR #9174, opened 2024-02-23
[WIP] setup: ensure that setup funcs do not block

sam-heilbron:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/pkg/utils/kubeutils/pods.go
+++ b/pkg/utils/kubeutils/pods.go
@@ -1,0 +1,93 @@
+package kubeutils
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/solo-io/skv2/pkg/multicluster/kubeconfig"
+)
+
+// Inspired by: https://github.com/solo-io/gloo-mesh-enterprise/blob/main/pkg/utils/kubeutils/pods.go
+
+// GetPodsForDeployment gets all pods backing a deployment
+func GetPodsForDeployment(
+	ctx context.Context,
+	kubeConfig string,
+	kubeContext string,
+	deploymentName string,
+	deploymentNamespace string,
+) ([]string, error) {
+	config, err := kubeconfig.GetRestConfigWithContext(kubeConfig, kubeContext, "")
+	if err != nil {
+		return nil, err
+	}
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	deployment, err := kubeClient.AppsV1().Deployments(deploymentNamespace).Get(ctx, deploymentName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	matchLabels := deployment.Spec.Selector.MatchLabels
+	listOptions := (&client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(matchLabels),
+		FieldSelector: fields.Set{"status.phase": "Running"}.AsSelector(),
+	}).AsListOptions()
+
+	podList, err := kubeClient.CoreV1().Pods(deploymentNamespace).List(ctx, *listOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	pods := make([]string, len(podList.Items))
+	for i := range podList.Items {
+		pods[i] = podList.Items[i].Name
+	}
+
+	return pods, nil
+}
+
+// GetPodsForService gets all pods backing a deployment
+func GetPodsForService(
+	ctx context.Context,
+	kubeConfig string,
+	kubeContext string,
+	serviceName string,
+	serviceNamespace string,
+) ([]string, error) {
+	config, err := kubeconfig.GetRestConfigWithContext(kubeConfig, kubeContext, "")
+	if err != nil {
+		return nil, err
+	}
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	service, err := kubeClient.CoreV1().Services(serviceNamespace).Get(ctx, serviceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	matchLabels := service.Spec.Selector
+	listOptions := (&client.ListOptions{LabelSelector: labels.SelectorFromSet(matchLabels)}).AsListOptions()
+
+	podList, err := kubeClient.CoreV1().Pods(serviceNamespace).List(ctx, *listOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	pods := make([]string, len(podList.Items))
+	for i := range podList.Items {
+		pods[i] = podList.Items[i].Name
+	}
+
+	return pods, nil
+}
--- a/pkg/utils/kubeutils/port_forward.go
+++ b/pkg/utils/kubeutils/port_forward.go
@@ -1,0 +1,188 @@
+package kubeutils
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/rotisserie/eris"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+
+	"github.com/solo-io/go-utils/cliutils"
+	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/skv2/pkg/multicluster/kubeconfig"
+)
+
+// Inspired by: https://github.com/solo-io/gloo-mesh-enterprise/blob/main/pkg/utils/kubeutils/port_forward.go
+
+type properties struct {
+	kubeConfig        string
+	kubeContext       string
+	resourceName      string
+	resourceNamespace string
+	localPort         string
+	remotePort        string
+	stdout            io.Writer
+	stderr            io.Writer
+}
+
+type PortForwardOption func(*properties)
+
+func WithKindCluster(kindClusterName string) PortForwardOption {
+	return func(config *properties) {
+		config.kubeContext = fmt.Sprintf("kind-%s", kindClusterName)
+	}
+}
+
+func WithResource(name, namespace string) PortForwardOption {
+	return func(config *properties) {
+		config.resourceName = name
+		config.resourceNamespace = namespace
+	}
+}
+
+func WithPorts(localPort, remotePort int) PortForwardOption {
+	return func(config *properties) {
+		if localPort == 0 {
+			config.localPort = ""
+		} else {
+			config.localPort = strconv.Itoa(localPort)
+		}
+		config.remotePort = strconv.Itoa(remotePort)
+	}
+}
+
+func WithWriters(out, err io.Writer) PortForwardOption {
+	return func(config *properties) {
+		config.stdout = out
+		config.stderr = err
+	}
+}
+
+func buildPortForwardProperties(options ...PortForwardOption) *properties {
+	//default
+	cfg := &properties{
+		kubeConfig:        "",
+		kubeContext:       "",
+		resourceName:      "",
+		resourceNamespace: "",
+		localPort:         "",
+		remotePort:        "",
+		stdout:            os.Stdout,
+		stderr:            os.Stderr,
+	}
+
+	//apply opts
+	for _, opt := range options {
+		opt(cfg)
+	}
+
+	return cfg
+}
+
+// PortForwardFromDeployment opens a port-forward against the specified deployment. Returns the local port.
+// If localPort is unspecified, a free port will be chosen at random.
+// Canceling the context will stop the port-forward.
+func PortForwardFromDeployment(ctx context.Context, options ...PortForwardOption) (string, error) {
+	config := buildPortForwardProperties(options...)
+
+	pods, err := GetPodsForDeployment(ctx, config.kubeConfig, config.kubeContext, config.resourceName, config.resourceNamespace)
+	if err != nil {
+		return "", err
+	}
+
+	if len(pods) == 0 {
+		return "", eris.Errorf("No pods found for deployment %s: %s", config.resourceNamespace, config.resourceName)
+	}
+
+	config.resourceName = pods[0]
+	return portForwardFromPod(ctx, config)
+}
+
+// PortForwardFromSvc opens a port-forward against the specified service. Returns the local port.
+// If localPort is unspecified, a free port will be chosen at random.
+// Canceling the context will stop the port-forward.
+func PortForwardFromSvc(ctx context.Context, options ...PortForwardOption) (string, error) {
+	config := buildPortForwardProperties(options...)
+
+	pods, err := GetPodsForService(ctx, config.kubeConfig, config.kubeContext, config.resourceName, config.resourceNamespace)
+	if err != nil {
+		return "", err
+	}
+
+	if len(pods) == 0 {
+		return "", eris.Errorf("No pods found for service %s: %s", config.resourceNamespace, config.resourceName)
+	}
+
+	config.resourceName = pods[0]
+	return portForwardFromPod(ctx, config)
+}
+
+func portForwardFromPod(ctx context.Context, options *properties) (string, error) {
+	logger := contextutils.LoggerFrom(ctx)
+	// select random open local port if unspecified
+	if options.localPort == "" {
+		freePort, err := cliutils.GetFreePort()
+		if err != nil {
+			return "", err
+		}
+		options.localPort = strconv.Itoa(freePort)
+	}
+	logger.Debugf("forwarding port %s of pod %s in namespace %s to local port %s",
+		options.remotePort, options.resourceName, options.resourceNamespace, options.localPort)
+
+	config, err := kubeconfig.GetRestConfigWithContext(options.kubeConfig, options.kubeContext, "")
+	if err != nil {
+		return "", err
+	}
+
+	// the following code is based on this reference, https://github.com/kubernetes/client-go/issues/51
+	roundTripper, upgrader, err := spdy.RoundTripperFor(config)
+	if err != nil {
+		return "", err
+	}
+
+	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", options.resourceNamespace, options.resourceName)
+	hostIP := strings.TrimLeft(config.Host, "https:/")
+	serverURL := url.URL{Scheme: "https", Path: path, Host: hostIP}
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, &serverURL)
+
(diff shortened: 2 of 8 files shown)


sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=82a9c9df-c788-440d-93c2-95e58ea89080

I have not seen this failure before, but it seems to be around status propagation on the Proxy

sam-heilbron:
The build-bot failure is a real one. There is some issue with status propagation in e2e tests. This had passed previously, so it seems like I broke something in a recent commit, I will investigate

projects/gateway2/deployer/deployer.go
```diff
@@ -135,9 +133,7 @@ func (d *Deployer) renderChartToObjects(ctx context.Context, gw *api.Gateway) ([
 			},
 		},
 	}
-	if d.dev {
-		vals["develop"] = true
-	}
```
lgadban, line 140: how do we set this now?
it wasn't pretty, but this (at least theoretically) was how we can enable debug logs in envoy at startup in the deployed proxies
ref: https://github.com/solo-io/gloo/blob/sam/combine-binaries/projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml#L40-L43

projects/gateway2/controller/main.go
```diff
@@ -76,49 +69,35 @@ func Start(cfg ControllerConfig) {
 		mgr.GetScheme(),
 	)
 	if err := mgr.Add(xdsSyncer); err != nil {
-		setupLog.Error(err, "unable to add xdsSyncer runnable")
-		os.Exit(1)
-	}
-
-	// sam-heilbron: I don't think this is necessary, as we should have a shared cache
-	if cfg.Dev {
-		go xdsSyncer.ServeXdsSnapshots()
```
lgadban, line 85: unless this changed elsewhere, this is a real endpoint -- used for debugging the xds being served by ggv2 syncer
see: https://github.com/solo-io/gloo/blob/sam/combine-binaries/projects/gateway2/xds/xds_syncer.go#L314

does this comment mean the served xds should be available in another debug endpoint thanks to the shared cache?

projects/gateway2/controller/main.go
```diff
@@ -13,35 +14,29 @@ import (
...
 
 	ControlPlane bootstrap.ControlPlane
 }
 
-func Start(cfg ControllerConfig) {
-	setupLog.Info("xxxxx starting gw2 controller xxxxxx")
-	var opts []zap.Opts
-	if cfg.Dev {
-		setupLog.Info("starting log in dev mode")
-		opts = append(opts, zap.UseDevMode(true))
```
lgadban, line 42: IIRC this was needed to get debug logs working with controller-runtime

lgadban (commented):
Awesome PR description, thanks!

This is a first pass of things that I am second guessing the removal of.
Will do another pass of the actual setup func logic next.

projects/gateway2/controller/main.go
sam-heilbron: I had thought that since we have https://github.com/solo-io/gloo/blob/sam/combine-binaries/projects/gloo/pkg/syncer/translator_syncer.go#L90, we would get this endpoint and we wouldn't need both. Was I incorrect in that assumption?

projects/gateway2/deployer/deployer.go
sam-heilbron: That's fair. I was probably overly aggressive removing this. I figured we'll need to allow configuring more values on these dynamic proxies and I probably underestimated the importance of keeping it here for now. Do we want to keep this configurable for now (I'm happy to add it) or would just enabling develop logs by default be enough?

pkg/utils/kubeutils/port_forward.go
```diff
@@ -0,0 +1,188 @@
...
+		opt(cfg)
+	}
+
+	return cfg
+}
+
+// PortForwardFromDeployment opens a port-forward against the specified deployment. Returns the local port.
+// If localPort is unspecified, a free port will be chosen at random.
+// Canceling the context will stop the port-forward.
+func PortForwardFromDeployment(ctx context.Context, options ...PortForwardOption) (string, error) {
```
jenshu, line 92: are the functions in this file used anywhere?
are they intended for e2e tests?

projects/gateway2/controller/main.go
```diff
@@ -13,35 +14,29 @@ import (
...
-	}
-	ctrl.SetLogger(zap.New(opts...))
+// Start runs the controllers responsible for processing the K8s Gateway API objects
+// It is intended to be run in a goroutine as the function will block until the supplied
+// context is cancelled
+func Start(cfg ControllerConfig) error {
+	ctx := cfg.Ctx
+	logger := contextutils.LoggerFrom(ctx)
```
jenshu, line 38: i know we need to clean up the logging in general, but it would be good to have a named logger here for easier debugging (if it isn't already named)

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -456,7 +456,10 @@ func RunGloo(opts bootstrap.Opts) error {
 	return RunGlooWithExtensions(opts, glooExtensions)
 }
 
-// Called directly by GlooEE
```
jenshu, line 459: nit: i liked having the callout that this function is used externally, so we know the implications of removing it or changing the function signature

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -470,9 +473,12 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 	}
 
 	watchOpts := opts.WatchOpts.WithDefaults()
+	watchOpts.Ctx = contextutils.WithLogger(watchOpts.Ctx, "setup")
 	opts.WatchOpts.Ctx = contextutils.WithLogger(opts.WatchOpts.Ctx, "gloo")
```
jenshu, line 477: these lines are a little confusing to me, why are we setting different loggers on `opts.WatchOpts` and `watchOpts` (which is derived from `opts.WatchOpts`)?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -645,7 +649,7 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 	}
 	if warmTimeout.GetSeconds() != 0 || warmTimeout.GetNanos() != 0 {
 		warmTimeoutDuration := prototime.DurationFromProto(warmTimeout)
-		ctx := opts.WatchOpts.Ctx
+		ctx := watchOpts.Ctx
```
jenshu, line 652: thanks for this change, i agree we should use `watchOpts` everywhere rather than the original `opts.WatchOpts`.
i see at least one more usage of opts.WatchOpts on line 776, should we update that too?
and also in `K8sGatewayControllerStartFunc`

jenshu (commented):
looks great overall! just some minor comments

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: I agree it would be nice to consolidate all the usages so they are all pointing to the same context. When creating the StartFunc type, I decided to use `Opts` and `Extensions` since those are the core inputs that all StartFunctions will require

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: Yeah I agree this is a bit confusing. This line is actually just changing the order of when code is executed. I'm not exactly sure why we would have different contexts, with different loggers. At first I tried to consolidate a lot of these changes, but that broke tests, so I backed them out

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: Good call! I tried to simplify by calling out that it was the entrypoint (both for other open source components, and enterprise ones), but I like the direct callout about GlooEE depending on it, I'lll re-add that

projects/gateway2/controller/main.go
sam-heilbron: I'm happy to re-add this, it was never actually controlled by any value when calling Start (see the setup_syncer invocation of this function), so it always defaulted to false

projects/gateway2/controller/main.go
sam-heilbron: I re-added the logger functionality that I had removed. On top of that, every "top-level" StartFunc, is provided a namedLogger for the injected context

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: On second thought, I like your idea. I moved the context to an explicit argument

pkg/utils/kubeutils/port_forward.go
sam-heilbron: They are not used anywhere yet. I noted this in the PR body about how I had found them useful.

At the beginning of this PR, I had intentions of incorporating them, but then I backed away from those changes. I'm happy to just remove these files from this PR to simplify/clarify

projects/gateway2/controller/main.go
sam-heilbron: I re-added this, and added a comment about how the value is not actually user-controlled, but I agree having this functionality available to developers (and eventually users) will be good

pkg/utils/kubeutils/pods.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/solo-io/skv2/pkg/multicluster/kubeconfig"
+)
+
+// Inspired by: https://github.com/solo-io/gloo-mesh-enterprise/blob/main/pkg/utils/kubeutils/pods.go
+
+// GetPodsForDeployment gets all pods backing a deployment
+func GetPodsForDeployment(
```
npolshakova, line 18: Are these just nice-to-haves to help with debugging? They're not currently being used by the regression tests, right?

npolshakova: Oh nevermind- answered here: https://github.com/solo-io/gloo/pull/9174/files#r1501723494

projects/gateway2/controller/start.go
```diff
@@ -1,47 +1,53 @@
...
-	AutoProvision         bool
-	Dev                   bool
-
+type StartConfig struct {
+	Dev          bool
 	ControlPlane bootstrap.ControlPlane
 }
 
-func Start(cfg ControllerConfig) {
-	setupLog.Info("xxxxx starting gw2 controller xxxxxx")
```
npolshakova, line 38: Is there any way to identify whether the ggv2 controller is running without this log line? I'm not a fan of the `gw2` name, but maybe it would still be helpful to create some sort of identifier for this controller?

npolshakova: Or is this handled by the `k8s-gateway-controller` name get initialized with the logger in the start function `namedCtx := contextutils.WithLogger(ctx, name)`?

sam-heilbron: At the moment, I've identified these asynchronous start functions using a named context, and we start with a debug line saying that a specific task is starting: https://github.com/solo-io/gloo/blob/sam/combine-binaries/projects/gloo/pkg/syncer/setup/start_func.go#L31. Future log lines will also be attached to that name, so I think/hope it will be clear what logs are associated with that controller.

It's a good question though, so let me run this at the debug log level and paste what log lines appear so we can discuss if it's clear enough, how does that sound?

projects/gateway2/controller/main.go
lgadban: oh interesting, I assume it should work!

projects/gateway2/controller/start.go
```diff
@@ -1,47 +1,53 @@
...
+	gatewayClass = apiv1.ObjectName(wellknown.GatewayClassName)
+
 	setupLog = ctrl.Log.WithName("setup")
 )
 
-type ControllerConfig struct {
-	// The name of the GatewayClass to watch for
-	GatewayClassName      string
-	GatewayControllerName string
-	Release               string
```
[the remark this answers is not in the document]
sam-heilbron: It was not

projects/gateway2/controller/start.go
```diff
@@ -1,47 +1,53 @@
...
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	apiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+const (
+	// AutoProvision controls whether the controller will be responsible for provisioning dynamic
+	// infrastructure for the Gateway API.
+	AutoProvision = true
```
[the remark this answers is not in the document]
sam-heilbron: Similar to "release", I found that we were passing a lot of variable between structs, and it was obfuscating what was actually happening and what was actually configurable. In this case, there is no user defined option to disable this, plus it's an intentional design decision to provision proxies ourselves, so I figured it was simpler to not pass around the option in a bunch of structs, and to just define a constant.

projects/gateway2/controller/controller.go
```diff
@@ -87,7 +89,15 @@ func (c *controllerBuilder) watchGw(ctx context.Context) error {
...
 
 	log.Info("creating deployer", "ctrlname", c.cfg.ControllerName, "server", c.cfg.ControlPlane.GetBindAddress(), "port", c.cfg.ControlPlane.GetBindPort())
-	d, err := deployer.NewDeployer(c.cfg.Mgr.GetScheme(), c.cfg.Dev, c.cfg.ControllerName, c.cfg.ControlPlane.GetBindPort())
+
+	deployerOptions := []deployer.Option{
+		deployer.WithScheme(c.cfg.Mgr.GetScheme()),
+		deployer.WithChartFs(helm.GlooGatewayHelmChart),
+		deployer.WithControllerName(c.cfg.ControllerName),
+		deployer.WithXdsServer(c.cfg.ControlPlane.GetBindPort()),
+		deployer.WithDevMode(c.cfg.Dev),
```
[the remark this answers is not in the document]
sam-heilbron: Yeah perhaps I was premature in this. What I found while developing this code, was that we had a constructor that accepted options, and in future PRs we kept changing the constructor to add more options. See https://github.com/solo-io/gloo/pull/9165 as an example of another instance where the constructor signature would change.

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -456,7 +456,13 @@ func RunGloo(opts bootstrap.Opts) error {
 	return RunGlooWithExtensions(opts, glooExtensions)
 }
 
-// Called directly by GlooEE
+// RunGlooWithExtensions is the core entrypoint to the Gloo components.
+// THIS FUNCTION MUST NOT BLOCK:
```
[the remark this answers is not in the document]
sam-heilbron: Good call! Should we also add a log line after each of the StartFunc are executed to show that they have all begun and likely add a test that proves that it cannot block?

projects/gateway2/controller/start.go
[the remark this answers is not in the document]
lgadban: this will definitely be supported though and will need to be configurable eventually.
I'm fine with this change but FWIW it will eventually need to be a variable associated with the controller

projects/gateway2/deployer/deployer.go
lgadban: Not defaulting for now

projects/gateway2/controller/main.go
sam-heilbron: Sam followup: make sure that the legacy xds snapshot endpoint works, and will serve up configuration for legacy and new proxies. Not required for this PR however.

pkg/utils/kubeutils/port_forward.go
sam-heilbron: Per Slack: we will incorporate this in a follow-up, but we don't like the idea of pulling in code that is not used anywhere. 
Sam todo before resolving: remove this code

projects/gateway2/controller/start.go
sam-heilbron: Sam todo: in the StartFunc, replace 'main' with the actual name of the logger. And change the log level to Info there

projects/gateway2/controller/start.go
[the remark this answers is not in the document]
sam-heilbron: Per slack: To reduce churn, we're not going to re-add this. We'll leave it as a constant for now, but we will need to build this into the product in the future

projects/gateway2/controller/controller.go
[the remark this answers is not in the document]
sam-heilbron: Sam todo: follow up with Nina RE struct or builder pattern.

projects/gloo/pkg/syncer/setup/setup_syncer.go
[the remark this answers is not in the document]
sam-heilbron: sam todo:
- add DEBUG log line at the end of StartAsync funcs 
- add INFO log line to https://github.com/solo-io/gloo/blob/sam/combine-binaries/projects/gloo/pkg/syncer/setup/setup_syncer.go#L1020 (gloo is starting)

projects/gateway2/controller/controller.go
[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/pull/9174/commits/21bebdcb9eba342046ac4bc845a14c3e39cdcff7

projects/gateway2/controller/start.go
sam-heilbron: https://github.com/solo-io/gloo/pull/9174/commits/ee8a3ece7d146cfc7b1626d93ffc981323494b6b
https://github.com/solo-io/gloo/pull/9174/commits/92f8ddb68c08a8c2b9c79181c41b4bc911c39a61

pkg/utils/kubeutils/port_forward.go
sam-heilbron: https://github.com/solo-io/gloo/pull/9174/commits/7b0701d4e3cde515ba9e850bb5737238dd0f30f2

jenshu (dismissed):

projects/gateway2/deployer/deployer.go
```diff
@@ -36,28 +38,39 @@ type gatewayPort struct {
 	TargetPort uint16 `json:"targetPort"`
 }
 
+// A Deployer is responsible for deploying proxies
+// See builder.go for the Deployer constructor
```
lgadban, line 42: update to `deployer.Inputs` ?

sam-heilbron: https://github.com/solo-io/gloo/pull/9174/commits/4cc9a63fd2d6cd91e3aa9048b4cd0ad0eb897f46

lgadban (approved):
LGTM

jenshu (approved):

PR #9177, opened 2024-02-23
Update github to trigger on ready for review

npolshakova:
In draft: 
image

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,5 +1,7 @@
 name: GGII Tests
-on: pull_request
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
   test_gloo_gateway2:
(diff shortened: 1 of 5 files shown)


.github/workflows/pr.yaml
```diff
@@ -1,5 +1,7 @@
 name: GGII Tests
-on: pull_request
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
jenshu, line 4: this only affects the GG workflow; we should also add this to the [GE regression tests](https://github.com/solo-io/gloo/blob/d43d6b964eeda88cb4258f6444683bc4ea838711/.github/workflows/regression-tests.yaml#L2)

should probably also add a check in this file below to only run these tests if it's not a draft, like [this](https://github.com/solo-io/gloo/blob/d43d6b964eeda88cb4258f6444683bc4ea838711/.github/workflows/regression-tests.yaml#L43)

sam-heilbron (commented):
LGTM!

Given these changes, I think we can update the README: https://github.com/solo-io/gloo/tree/main/.github/workflows to reference:
- The new GGII tests, and that they do not run on a draft PR
- Remove the special callout that when conveting a draft PR -> ready for review requires manual intervention

sam-heilbron (approved):
Thanks for this! 🚀

PR #9190, opened 2024-02-26
Updates to various sections in the support ticket docs

codeverifier:
# Docs changes

- Updated sections in the "Submit a request" docs


nmnellis (commented):
In the debugging doc i would like to see a set of kubectl commands that a customer can run to gather all the information we think is needed. if someone has an issue i want to point them to this doc and say "run these commands in sections a,b,e" and include them in the ticket.

PR #9201, opened 2024-02-27
k8sgateway/plugins: expose an extension point to inject enterprise-only plugins

ilackarms:
# Description
Expose extension points, for injecting custom functionality when processing k8s gateway resources.

# Context
## Developer Portal
We need custom functionality to execute during the translation of k8s gateway resources into a Proxy resource. That will be used to power a developer portal. 

The implementation to rely on these extension points can be found here: https://github.com/solo-io/solo-projects/pull/5850. However, we will be breaking that PR into smaller pieces, and first we will use https://github.com/solo-io/solo-projects/pull/5883 to prove out this injection functionality.

### Extension Point: Gateway Queries
[Gateway Queries](https://github.com/solo-io/gloo/blob/main/projects/gateway2/query/query.go#L111) provide a framework for developers to query cached data from etcD. The Portal plugin which must run during (and after) translation, requires an extended implementation of this interface, to query specific Portal data. 

### Extension Point: Plugin Registry
We need to inject some custom plugins to be executed during Portal (Enterprise) translation. As a result, we need a mechanism to construct a registry with a set of Plugins. This is already doable, but at the moment, to construct those plugins, we pass a `GatewayQueries` object. However, what we really need is a type that:
- Implements the GatewayQueries interface
- Implements a custom PortalQueries interface (this is necessary to construct the Portal plugin)

### Proposed Solution: Extension Manager
There are two dimensions in which the extensions can extend (queries, plugins) and those two implementations are dependent on one another. As a result, I propose we encapsulate those into a single `ExtensionManager`.


projects/gateway2/translator/plugins/registry/plugin_registry.go
```diff
@@ -1,43 +1,71 @@
...
 
-type PluginRegistry struct {
-	routePlugins []plugins.RoutePlugin
+type PluginRegistry interface {
+	GetRoutePlugins() []plugins.RoutePlugin
+	GetNamespacePlugins() []plugins.NamespacePlugin
 }
 
-func (h *PluginRegistry) GetRoutePlugins() []plugins.RoutePlugin {
+type PluginRegistryFactory func(ctx context.Context, mgr controllerruntime.Manager, queries query.GatewayQueries) PluginRegistry
```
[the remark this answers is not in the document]
lgadban: Also, why are `ctx` and `mgr` being passed?
Doesn't seem like this is really intended to be "request" scoped re: `ctx`
And the impl drops the value of `mgr`

projects/gateway2/xds/xds_syncer.go
```diff
@@ -377,9 +340,22 @@ func (s *XdsSyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl ap
...
 			gw.Status = *status
-			if err := s.cli.Status().Patch(ctx, &gw, client.Merge); err != nil {
+			if err := s.mgr.GetClient().Status().Patch(ctx, &gw, client.Merge); err != nil {
 				logger.Error(err)
 			}
 		}
 	}
 }
+
+func (s *XdsSyncer) syncGwNamespaces(ctx context.Context, namespaces []string, pluginRegistry registry.PluginRegistry) {
```
lgadban, line 350: I'm concerned with namespace plugins being in the same plugin registry as the e.g. route plugins. I see the original intent as being translation plugins, whereas this is something different.
Are we comfortable with having a single registry that includes plugins that do "anything"?

projects/gloo/pkg/syncer/setup/extensions.go
```diff
@@ -0,0 +1,48 @@
...
+	k8sgatewayregistry "github.com/solo-io/gloo/projects/gateway2/translator/plugins/registry"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins"
+	"github.com/solo-io/gloo/projects/gloo/pkg/syncer"
+	xdsserver "github.com/solo-io/solo-kit/pkg/api/v1/control-plane/server"
+)
+
+// Extensions contains the set of extension points for Gloo
+type Extensions struct {
+	// K8sGatewayPluginRegistryFactory is responsible for creating a K8sGateway PluginRegistry
+	// This is the set of plugins which are executed when converting K8s Gateway resources into a Proxy resource
```
lgadban, line 14: This comment illustrates my point above. These plugins, now, are NOT only responsible for converting K8s GW resources into `Proxies`. It's more than just translation

sam-heilbron: Good point! So 2 aspects of this: we do need a way to inject custom RoutePlugins (this is what portal will need to execute), and we do need a way to inject custom "post translation processing" (this is the second half of what portal needs). 

Up until now, the registry just managed all the translator units (plugins), though I'm happy to diferentiate those concepts into a more "translation-specific registry" and "post-processing translation"

projects/gateway2/translator/plugins/registry/plugin_registry.go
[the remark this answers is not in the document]
sam-heilbron: Sorry I should have linked https://github.com/solo-io/solo-projects/pull/5850. This is an unfortunate detail about what enterprise plugins need to be constructred

projects/gateway2/translator/plugins/registry/plugin_registry.go
```diff
@@ -1,43 +1,71 @@
...
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/mirror"
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/redirect"
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/routeoptions"
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/urlrewrite"
+	controllerruntime "sigs.k8s.io/controller-runtime"
 )
 
-type PluginRegistry struct {
-	routePlugins []plugins.RoutePlugin
+type PluginRegistry interface {
```
sam-heilbron, line 16: there are not different implementations of the same interface, so this should be a struct.

projects/gateway2/translator/plugins/registry/plugin_registry.go
```diff
@@ -1,43 +1,71 @@
...
+	GetRoutePlugins() []plugins.RoutePlugin
+	GetNamespacePlugins() []plugins.NamespacePlugin
 }
 
-func (h *PluginRegistry) GetRoutePlugins() []plugins.RoutePlugin {
+type PluginRegistryFactory func(ctx context.Context, mgr controllerruntime.Manager, queries query.GatewayQueries) PluginRegistry
+
+type pluginRegistry struct {
+	routePlugins     []plugins.RoutePlugin
+	namespacePlugins []plugins.NamespacePlugin
```
sam-heilbron, line 25: convert this to a "post processing plugin", and leave in the existing registry

projects/gateway2/controller/start.go
```diff
@@ -33,8 +33,9 @@ var (
 )
 
 type StartConfig struct {
-	Dev          bool
-	ControlPlane bootstrap.ControlPlane
+	Dev                   bool
+	ControlPlane          bootstrap.ControlPlane
+	PluginRegistryFactory registry.PluginRegistryFactory
```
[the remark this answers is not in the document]
sam-heilbron: No, we inject a factory which will be invoked at runtime (inside the controller) to construct the plugins at the same time they previously were built

sam-heilbron:
The build-bot failure (https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=217cdf6b-11fd-4ba0-92bf-9396f7ccc59b) is unrelated to this work:
```
Step #0 - "publish-artifacts": Step #2 - "docker-login": Error response from daemon: Get "https://quay.io/v2/": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

However, I will spend some time this morning looking into it

projects/gateway2/translator/plugins/registry/plugin_registry.go
```diff
@@ -10,34 +10,56 @@ import (
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/urlrewrite"
 )
 
+// PluginRegistry is used to provide Plugins to relevant translators using
+// during the conversion of K8s Gateway resources into a Gloo Proxy resource.
+// The registry is a concrete type because we only have one implementation for a
+// registry to provide these plugins. This can still support Enterprise-only functionality
+// because the registry initialized with the list of plugins that it will manage.
```
lgadban, line 17: typo: "relevant translators using during the conversion"

also wording isn't quite correct, now that we have postTranslationPlugins, the plugins are not used purely during conversion to a Proxy object

last, I think commenting on the reasoning for the concrete-only type seems unnecessary but that is just a nit

projects/gateway2/extensions/manager.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	e.queryEngine = query.NewData(e.manager.GetClient(), e.manager.GetScheme())
+	return e.queryEngine
+}
+
+func (e *extensionManager) CreatePluginRegistry(ctx context.Context) registry.PluginRegistry {
+	if !e.pluginRegistry.IsNil() {
+		return e.pluginRegistry
+	}
```
lgadban, line 46: why is this necessary? where can this be called multiple times?

projects/gateway2/extensions/manager.go
```diff
@@ -0,0 +1,52 @@
...
+	manager controllerruntime.Manager
+
+	queryEngine    query.GatewayQueries
+	pluginRegistry registry.PluginRegistry
+}
+
+func (e *extensionManager) CreateGatewayQueries(ctx context.Context) query.GatewayQueries {
+	if e.queryEngine != nil {
+		return e.queryEngine
+	}
```
lgadban, line 37: why?

projects/gateway2/xds/xds_syncer.go
```diff
@@ -377,9 +345,22 @@ func (s *XdsSyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl ap
...
 			gw.Status = *status
-			if err := s.cli.Status().Patch(ctx, &gw, client.Merge); err != nil {
+			if err := s.mgr.GetClient().Status().Patch(ctx, &gw, client.Merge); err != nil {
 				logger.Error(err)
 			}
 		}
 	}
 }
+
+func (s *XdsSyncer) syncPostTranslationPlugins(ctx context.Context, namespaces []string, pluginRegistry registry.PluginRegistry) {
```
lgadban, line 355: nit: this isn't a sync, just applying the plugins

projects/gateway2/extensions/manager.go
sam-heilbron: I'm happy to remove this. I know that in the current usage, we rely on invoking the "create" method just once at the beginning. I figured that this manager should have a clear responsibility/expectation, regardless of how it is used today. That way, if other devs use it in other ways, the original assumptions are well understood. If an extension manager should produce new extension each time it is requested, we can make that happen.

projects/gateway2/translator/plugins/registry/plugin_registry.go
sam-heilbron: Sounds good I can update

projects/gateway2/xds/xds_syncer.go
```diff
@@ -152,28 +147,36 @@ func (s *XdsSyncer) Start(
...
 				proxies = append(proxies, proxy)
+				uniqueGatewayNamespaces = stringutils.AppendIfMissing(uniqueGatewayNamespaces, proxy.GetMetadata().GetNamespace())
 				//TODO: handle reports and process statuses
 			}
 		}
 		proxyApiSnapshot.Proxies = proxies
 		s.syncEnvoy(ctx, proxyApiSnapshot)
 		s.syncStatus(ctx, rm, gwl)
 		s.syncRouteStatus(ctx, rm)
+		s.applyPostTranslationPlugins(ctx, uniqueGatewayNamespaces, pluginRegistry)
```
inFocus7, line 179: Is it possible that the number of **repeated** gateway namespaces could ever be "big"? I guess it would be dependent on number of proxies, but idk how many a customer/user would have on average. 

i'm just wondering if this implementation is the best option, or if we would want to instead use a set like `var uniqueGatewayNamespaces map[string]interface{}`. That way the `AppendIfMissing` could be O(1) instead of O(N).. although we would likely have to convert it to a slice [O(N)] (or loop through the map/set - unsure the time complexity on that) for the `applyPostTranslationPlugins` portion.

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -464,15 +460,8 @@ func RunGloo(opts bootstrap.Opts) error {
...
-	if extensions.ApiEmitterChannel == nil {
-		return errors.Errorf("Extensions.ApiEmitterChannel must be defined, found nil")
-	}
-	if extensions.PluginRegistryFactory == nil {
-		return errors.Errorf("Extensions.PluginRegistryFactory must be defined, found nil")
-	}
-	if extensions.SyncerExtensions == nil {
-		return errors.Errorf("Extensions.SyncerExtensions must be defined, found nil")
+	if err := extensions.Validate(); err != nil {
+		return err
```
inFocus7, line 464: The `extensions` we're import and using here here is from the `gateway2 > extensions` package ("github.com/solo-io/gloo/projects/gateway2/extensions"). The original code references extensions which are set up in the syncer setup package ("/projects/gloo/syncer/setup/setup_syncer.go"). 

Should we be using the syncer setup extension's `Validate()` here instead? Or am i misunderstanding the extension points/logic? 👀

projects/gateway2/extensions/manager.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (e *extensionManager) CreateGatewayQueries(ctx context.Context) query.GatewayQueries {
+	return query.NewData(e.manager.GetClient(), e.manager.GetScheme())
+}
+
+func (e *extensionManager) CreatePluginRegistry(ctx context.Context) registry.PluginRegistry {
+	gatewayQueries := e.CreateGatewayQueries(ctx)
+	plugins := registry.BuildPlugins(gatewayQueries)
+	return registry.NewPluginRegistry(plugins)
+}
```
inFocus7, line 39: if `CreatePluginRegistry` already sets up gateway queries, why do we need to expose/use `CreateGatewayQueries` outside of it*? 

\* by 'it', i mean the `CreatePluginRegistry` function

inFocus7 (commented):
have some questions. also do we have existing any tests for this just to make sure no functionality was broken? (i haven't deep dived into our ggv2 code in gloo to know what to look out for, as far as things potentially breaking)

projects/gateway2/xds/xds_syncer.go
```diff
@@ -377,9 +345,22 @@ func (s *XdsSyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl ap
...
 	}
 }
+
+func (s *XdsSyncer) applyPostTranslationPlugins(ctx context.Context, namespaces []string, pluginRegistry registry.PluginRegistry) {
+	logger := contextutils.LoggerFrom(ctx)
+	for _, ns := range namespaces {
+		for _, postTranslationPlugin := range pluginRegistry.GetPostTranslationPlugins() {
+			err := postTranslationPlugin.ApplyPostTranslationPlugin(ctx, &gwplugins.NamespaceContext{Namespace: ns})
+			if err != nil {
+				logger.Errorf("Error applying post-translation plugin: %v", err)
```
nfuden, line 361: Should this only be logged or should it be aggregated and eventually returned?

projects/gloo/pkg/syncer/setup/extensions.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	errors "github.com/rotisserie/eris"
+	"github.com/solo-io/gloo/projects/gateway2/extensions"
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins"
+	"github.com/solo-io/gloo/projects/gloo/pkg/syncer"
+	xdsserver "github.com/solo-io/solo-kit/pkg/api/v1/control-plane/server"
+)
+
+// Extensions contains the set of extension points for Gloo
```
nfuden, line 11: What is an extension point? I get no additional understanding from this comment

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -31,3 +33,16 @@ type RoutePlugin interface {
 		outputRoute *v1.Route,
 	) error
 }
+
+type NamespaceContext struct {
+	// this is the namespace where the gateway lives
+	Namespace string
+}
+
+type PostTranslationPlugin interface {
```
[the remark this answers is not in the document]
sam-heilbron: These are all in the gateway2 package and only apply to the new k8s gateway translation

projects/gateway2/extensions/manager.go
sam-heilbron: The portal implementation requires that a PluginRegistry is created using both GatewayQueries and PortalQueries

projects/gloo/pkg/syncer/setup/setup_syncer.go
sam-heilbron: So there is extensions object that is the set of extensions that the ControlPlane supports, and it is defined in a new extensions.go file. That object supports a 'validate' method on it, which historically was just defined inline in the control plane setup_syncer file.

I created a new property in the control plane extensions, and it's specific to the k8s gateway extensions.

projects/gateway2/xds/xds_syncer.go
sam-heilbron: I considered it, though I think we're working in a small enough space of namespaces that I wouldn't expect this to be an issue.

projects/gateway2/xds/xds_syncer.go
sam-heilbron: At the moment we only have a single plugin (the soon to be portal post-processing plugin) so I don't see an issue with logging on failure. There isn't a plan to have many of these post-processing plugins, and if that day comes I think we'd want to re-evaluate the whole structure. I'd rather keep the code simple and log immediately, given that there is only one instance of it

projects/gateway2/extensions/manager.go
```diff
@@ -0,0 +1,39 @@
...
+// ExtensionManager is responsible for providing implementations for translation utilities
+// which have Enterprise variants.
+type ExtensionManager interface {
+	CreateGatewayQueries(ctx context.Context) query.GatewayQueries
+	CreatePluginRegistry(ctx context.Context) registry.PluginRegistry
+}
+
+type ExtensionManagerFactory func(manager controllerruntime.Manager) ExtensionManager
+
+// NewExtensionManager returns the Open Source implementation of ExtensionManager
```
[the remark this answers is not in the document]
sam-heilbron: Good call, I updated

projects/gloo/pkg/syncer/setup/setup_syncer.go
inFocus7: Ah i think i get it now. I was confused since `extensions` is the name of the imported ggv2 package(?) which supports `Valudate`. **_But_** we also have function parameters named `extensions` which are based on the set of extensions (`Extensions`) that were moved out.

I was getting confused due to the same name between the `extensions` parameter & the `extensions` import - if that makes sense. I pulled the code locally and understood the difference better

projects/gloo/pkg/syncer/setup/extensions.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if e.PluginRegistryFactory == nil {
+		return errors.Errorf("Extensions.PluginRegistryFactory must be defined, found nil")
+	}
+	if e.ApiEmitterChannel == nil {
+		return errors.Errorf("Extensions.ApiEmitterChannel must be defined, found nil")
+	}
+	if e.SyncerExtensions == nil {
+		return errors.Errorf("Extensions.SyncerExtensions must be defined, found nil")
+	}
```
inFocus7, line 49: super nit: could we create a wrapper (ex. **`extension.name` must be defined, found nil**) or reusable variable for these errors since they're all very similar. This would also help for the extension_test's expected error messages so that they're not hard-coded.

sam-heilbron: good call, pushed an update, let me know what you think

inFocus7 (approved):
🚀

projects/gateway2/translator/listener/gateway_listener_translator.go
```diff
@@ -187,6 +188,8 @@ type mergedListener struct {
 	httpFilterChain   *httpFilterChain
 	httpsFilterChains []httpsFilterChain
 	listenerReporter  reports.ListenerReporter
+	listener          gwv1.Listener
```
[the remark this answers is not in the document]
lgadban: am I missing the concern? `gwv1` is the k8s gateway API package, this is a GWAPI Listener

projects/gateway2/xds/xds_syncer.go
```diff
@@ -152,28 +147,36 @@ func (s *XdsSyncer) Start(
...
 		rm := reports.NewReportMap()
 		r := reports.NewReporter(&rm)
+
+		var uniqueGatewayNamespaces []string
 		for _, gw := range gwl.Items {
-			proxy := t.TranslateProxy(ctx, &gw, queries, r)
+			proxy := gatewayTranslator.TranslateProxy(ctx, &gw, gatewayQueries, r)
 			if proxy != nil {
 				proxies = append(proxies, proxy)
+				uniqueGatewayNamespaces = stringutils.AppendIfMissing(uniqueGatewayNamespaces, proxy.GetMetadata().GetNamespace())
```
lgadban, line 171: seems like this is namespace plugin specific logic; also seems like postTranslationPlugins are still namespace plugins in all but name

projects/gateway2/xds/xds_syncer.go
```diff
@@ -152,28 +146,41 @@ func (s *XdsSyncer) Start(
...
 				//TODO: handle reports and process statuses
 			}
 		}
 		proxyApiSnapshot.Proxies = proxies
 		s.syncEnvoy(ctx, proxyApiSnapshot)
 		s.syncStatus(ctx, rm, gwl)
 		s.syncRouteStatus(ctx, rm)
+		s.applyPostTranslationPlugins(ctx, pluginRegistry, &gwplugins.PostTranslationContext{
+			TranslatedProxies: translatedProxies,
+		})
```
lgadban, line 183: seems like now that we have the proper abstraction, post-translation should happen right after we have our list of proxies and before we do the various `sync...` calls.

I can see a possibility at some point where also introduce postSyncPlugins?

lgadban (commented):
looks great, a couple final comments

projects/gateway2/xds/xds_syncer.go
ilackarms: what's the difference between sync and translation?

lgadban: I'm imaging that is a postTranslationPlugin wanted to, it could modify the `Proxy` before it is translated to xds/synced to envoy. The way the ordering is done currently, that would not be possible, as the sync to envoy happens before the post translation plugins run

ilackarms: im not a fan of anything that would modify the proxy, only the gateway namespaces are required, and they should be read-only

ilackarms: nothing in the plugin inputs i.e. `*Context` should ever be modified

projects/gateway2/xds/xds_syncer.go
```diff
@@ -377,9 +349,22 @@ func (s *XdsSyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl ap
...
 			gw.Status = *status
-			if err := s.cli.Status().Patch(ctx, &gw, client.Merge); err != nil {
+			if err := s.mgr.GetClient().Status().Patch(ctx, &gw, client.Merge); err != nil {
 				logger.Error(err)
 			}
 		}
 	}
 }
+
+func (s *XdsSyncer) applyPostTranslationPlugins(ctx context.Context, pluginRegistry registry.PluginRegistry, translationContext *gwplugins.PostTranslationContext) {
```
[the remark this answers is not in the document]
sam-heilbron: Good call! I'll incorporate this once we agree on the comment below around what the behavior of the post translation processing should be (when it should run, what it should be called..etc)

projects/gateway2/xds/xds_syncer.go
sam-heilbron: Per a conversation with Scott, we opted to do the following:
1. remove the proxy from the postTranslationContext. It's not needed at the moment, and if in the future we want to explore including it, we can have that discussion
2. run the postTranslation plugins before any of the sync operations

projects/gateway2/xds/xds_syncer.go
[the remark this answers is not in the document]
sam-heilbron: I pushed up a change to remove this from the xdsSyncer

projects/gateway2/extensions/manager.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+
+	"github.com/solo-io/gloo/projects/gateway2/query"
+	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/registry"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+)
+
+// ExtensionManager is responsible for providing implementations for translation utilities
+// which have Enterprise variants.
```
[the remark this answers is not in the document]
nfuden: Fair enough. I was thinking of how users can override with their own code thats not solo owned/specific but I think this is sufficient

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -24,10 +26,28 @@ type RouteContext struct {
...
 }
+
+type PostTranslationContext struct {
+	// TranslatedGateways is the list of Gateways that were generated in a single translation run
+	TranslatedGateways []TranslatedGateway
+}
+
+type TranslatedGateway struct {
+	// Gateway is the input object that produced the Proxy
+	Gateway gwv1.Gateway
```
lgadban, line 44: is this not pointer to prevent modification in the plugin?

sam-heilbron: Partially, though it was more based on the fact that during translation we have the value and so I passed it to the plugin as is. I figured this has the benefit of:
- Modification: It means that the postProcessing plugin can't modify it, though I don't really think that is a concern
- Concurrency: The post processing may required some async process

I don't have strong feelings on it though, and could be convinced to pass the pointer

lgadban (approved):

inFocus7 (approved):
🚀

lgadban (approved):

sam-heilbron:
bulldozer please

PR #9208, opened 2024-02-29
xds/hash: support non-proxy workload key generation

sam-heilbron:
# Description

Fix the implementation of node_hash to support the assumption that Enterprise services make when communicating with the control plane.

# Context

## Previous Solution
In classic Gloo Edge (1.16 and below), we only supported a single type of Proxy workload (data plane). This meant that when workloads started, they would identify themselves via the `node.metadata.role` property in the Node metadata. Our xDS cache used a node hash to inspect the value of the `node.metadata.role` and use that string value as the exact cache key.

Node Metadata:
```
node:
  metadata:
    role: gloo-system~gateway-proxy
```

In-Memory Cache:
```
gloo-system~gateway-proxy: <xDS Snapshot>
```

We relied on the same functionality for non-proxy workloads (like extauth). Node Metadata:
```
node:
  metadata:
    role: extauth
```
In-Memory Cache:
```
extauth: <xDS Snapshot>
```

## Flawed Existing Solution
As part of work to support the k8s gateway API (https://github.com/solo-io/gloo/pull/9148/) we needed to distinguish between xDS Snapshots that were generated by the classic "Edge" controller, and the new "k8s Gateway" controller. This was because each controller was responsible for garbage collection, and we needed one controller to not clean up cache entries that were populated by the other controller.

To support this, we added a prefix for proxy-workloads in their cache entry:
```
node:
  metadata:
    role: gloo-system~gateway-proxy
```
In-Memory Cache:
```
gloo-gateway-translator~gloo-system~gateway-proxy: <xDS Snapshot>
```

However, this meant that the prefix was also applied to non-proxy workloads
```
node:
  metadata:
    role: extauth
```
In-Memory Cache:
```
gloo-gateway-translator~extauth: <xDS Snapshot>
```
**The problem with this is that the extauth syncer makes an assumption about the cache key, and assumes the key and the node.metadata.role are the same. Therefore, it stores cache entries at the `extauth` key. This means we have a disconnect: `extauth` is the key with the values, but an xds request for the `extauth` role, leads to a lookup at the `gloo-gateway-translator~extauth` key**

## New Solution
We have 3 types of workloads that can ask for configuration from the control plane:
- Classic Edge proxies
- New Gateway proxies dynamically provisioned from the k8s API
- Enterprise services

Unfortunately, when it was originally designed, the mechanism that "classic edge proxies" and "enterprise services" identified themselves was by the same field `node.metadata.role` but with different formats: proxy workloads use `NAMESPACE~NAME` , but enterprise services just use `NAME`.

Ideally, each workload would have a unique way of representing its metadata, so that the node_hash could easily come up with the relevant cache key. To support the legacy proxies, we need to make the node-hash aware of the implementation detail that non-proxy workloads should not be keyed by the gateway translator prefix. It's not ideal, but a necessary solution.

In the future, we should have enterprise services identify themselves using custom metadata, so that we can look for a particular field.

## Interesting decisions
 -

## Testing steps
- [x] Unit tests to validate the changes
- [x] Existing OSS integration tests prove that proxy workloads do work
- [x] [Enterprise PR](https://github.com/solo-io/solo-projects/pull/5854/) to prove that non-proxy workloads can get configuration from control plane

# Checklist:

### DIFF
--- a/projects/gloo/pkg/xds/node_hash_test.go
+++ b/projects/gloo/pkg/xds/node_hash_test.go
@@ -28,11 +28,16 @@ var _ = Describe("NodeHash", func() {
 				"non-role-field": structpb.NewStringValue("non-role-value"),
 			},
 		}, Equal(xds.FallbackNodeCacheKey)),
-		Entry("metadata with role", &structpb.Struct{
+		Entry("metadata with proxy workload", &structpb.Struct{
 			Fields: map[string]*structpb.Value{
-				"role": structpb.NewStringValue("role-value"),
+				"role": structpb.NewStringValue("proxy-namespace~proxy-name"),
 			},
-		}, Equal("gloo-gateway-translator~role-value")),
+		}, Equal("gloo-gateway-translator~proxy-namespace~proxy-name")),
+		Entry("metadata with non-proxy workload", &structpb.Struct{
+			Fields: map[string]*structpb.Value{
+				"role": structpb.NewStringValue("no-tilde-in-role"),
+			},
+		}, Equal("no-tilde-in-role")),
 	)
 
 	DescribeTable("GlooGatewayNodeHash",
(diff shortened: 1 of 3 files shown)


sam-heilbron:
Gloo regression tests failed with: https://github.com/solo-io/gloo/actions/runs/8104067463/job/22149930970?pr=9208. I retried the test to see if it's a flake

projects/gloo/pkg/xds/node_hash_test.go
```diff
@@ -28,11 +28,16 @@ var _ = Describe("NodeHash", func() {
 				"non-role-field": structpb.NewStringValue("non-role-value"),
 			},
 		}, Equal(xds.FallbackNodeCacheKey)),
-		Entry("metadata with role", &structpb.Struct{
+		Entry("metadata with proxy workload", &structpb.Struct{
```
npolshakova, line 31: This looks good- at some point we should add a test here for the `misconfigured-node` case

npolshakova (approved):

jenshu (approved):
looks good, and regression tests have passed in the enterprise PR using this branch

Issue #9211, opened 2024-03-02
glooctl version improvements


codeverifier:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

1.16.x

### Kubernetes Version

1.25.x

### Describe the bug

This ticket covers a number of issues and enhancements as laid out below to improve `glooctl version`

- [ ] incorrect print format for json & yaml when using `glooctl version -o xxx`. none of variants produce valid output.
- [ ] container image versions are not displayed correctly. container image versions are shown as `latest` when valid images are used.
- [ ] lack of kubernetes version in output causes some grief when it comes to diagnosing issues.

### Expected Behavior

all 3 issues fixed above

### Steps to reproduce the bug

run `glooctl version -o xxx`

PR #9212, opened 2024-03-03
Fixing #9211

codeverifier:
- Added kubernetes version
- Fixed prettifying output
- Fixed image version display
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9211


codeverifier:
output now looks like,

```
{
  "glooVersion": {
    "client": {
      "version": "1.0.1-dev"
    },
    "server": [
      {
        "type": "Gateway",
        "enterprise": true,
        "kubernetes": {
          "containers": [
            {
              "Tag": "1.14.15",
              "Name": "observability-ee",
              "Registry": "quay.io/solo-io"
            },
            {
              "Tag": "1.14.15",
              "Name": "gloo-ee",
              "Registry": "quay.io/solo-io",
              "OssTag": "1.14.29"
            },
            {
              "Tag": "1.14.15",
              "Name": "gloo-ee-envoy-wrapper",
              "Registry": "quay.io/solo-io"
            }
          ],
          "namespace": "gloo-system"
        }
      }
    ]
  },
  "kubernetesVersion": {
    "major": "1",
    "minor": "24",
    "gitVersion": "v1.24.13+k3s1",
    "gitCommit": "3f79b289229dc0452c5cde26cccfa069805c5610",
    "gitTreeState": "clean",
    "buildDate": "2023-04-20T01:39:02Z",
    "goVersion": "go1.19.8",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

projects/gloo/cli/pkg/cmd/version/cmd.go
```diff
@@ -80,53 +99,30 @@ func getClientVersion() *version.ClientVersion {
...
 
 func printVersion(sv ServerVersion, w io.Writer, opts *options.Options) error {
-	vrs, _ := GetClientServerVersions(opts.Top.Ctx, sv)
+	vrs, k8sV, _ := GetClientServerVersions(opts.Top.Ctx, sv)
+	wrappedVersions, _ := getWrappedVersions(vrs, k8sV)
 	// ignoring error so we still print client version even if we can't get server versions (e.g., not deployed, no rbac)
 	switch opts.Top.Output {
 	case printers.JSON:
-		clientVersion, err := GetJson(vrs.GetClient())
+		formattedVer, err := getFormattedJson(wrappedVersions) // GetJson(vrs)
```
jenshu, line 107:
```suggestion
		formattedVer, err := getFormattedJson(wrappedVersions)
```

projects/gloo/cli/pkg/cmd/version/cmd.go
```diff
@@ -80,53 +99,30 @@ func getClientVersion() *version.ClientVersion {
...
-			if err != nil {
-				return err
-			}
-			fmt.Fprintf(w, "%s\n", string(serverVersionStr))
+			fmt.Fprintf(w, "%s\n\n", undefinedServer)
 		}
+		fmt.Fprintf(w, "%s", string(formattedVer))
 	case printers.YAML:
-		clientVersion, err := GetYaml(vrs.GetClient())
+		formattedVer, err := getFormattedYaml(wrappedVersions) // GetJson(vrs)
```
jenshu, line 116:
```suggestion
		formattedVer, err := getFormattedYaml(wrappedVersions)
```

changelog/v1.17.0-beta11/9211-improve-glooctl-version.yaml.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
```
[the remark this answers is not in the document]
codeverifier: good catch

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -110,17 +116,25 @@ func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, error) {
 			},
 		},
 	}
-	return []*version.ServerVersion{serverVersion}, nil
+	return []*version.ServerVersion{serverVersion}, k8ServerVersion, nil
```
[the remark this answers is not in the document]
codeverifier: historical reasoning is beyond me. but i agree that this can be simplified.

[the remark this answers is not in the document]
jenshu: according to [here](https://github.com/solo-io/gloo/blob/112ed69f2156b2456da4af6ef4f3c1fc89809adc/projects/gloo/api/grpc/version/version.proto#L55-L56) it was indeed intended to return multiple gloo instances on a cluster. it currently won't do that because if you don't pass in a namespace, it defaults to `gloo-system` and it doesn't seem like there's a way to default to "all namespaces". we _could_ fix that by allowing some way to specify "all namespaces" and return multiple gloo versions, but i think that's out of the scope of this PR

and looking at the version.proto, another possible place we could insert the k8s version is adding a new field to [Version](https://github.com/solo-io/gloo/blob/112ed69f2156b2456da4af6ef4f3c1fc89809adc/projects/gloo/api/grpc/version/version.proto#L53) - so we could have a gloo `ClientVersion`, list of gloo `ServerVersion`s, and a `KubernetesVersion`

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -41,27 +42,32 @@ func NewKube(namespace, kubeContext string) *kube {
 	}
 }
 
-func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, error) {
+func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, *kube1vVersion.Info, error) {
```
sam-heilbron, line 45: Taking a look at the API that exists (https://github.com/solo-io/gloo/blob/main/projects/gloo/api/grpc/version/version.proto#L30) I think we may want to approach this slightly differently.

At the moment we have a ServerVersion concept, and there is one concrete type that we implement, a KubernetesServerVerision.

At the moment that type (https://github.com/solo-io/gloo/blob/main/projects/gloo/api/grpc/version/version.proto#L46) has a list of contains and namespace where GLoo is installed.

I think we could expand this API to also have cluster metadata. The downside is that this would lead us to duplicate the values if we did have multiple installs of gloo in the same cluster (each ServerVersion would have a reference to the same cluster metadata), but I think the benefit is that it allows us to use the existing API

sam-heilbron (commented):
Thanks for tackling this! I left some thoughts about ways to expand the API using the proto as the source of truth. Let me know if that doesn't make sense and I'm happy to huddle, or outline what I had in mind

projects/gloo/cli/pkg/cmd/version/clients.go
jenshu: see https://github.com/solo-io/gloo/pull/9212#discussion_r1516178018, i think it might be more appropriate to put the new field (KubernetesVersion / KubernetesServerVersion) in Version, so there is just one instance returned, rather than one per gloo instance

sam-heilbron: nice! yeah totally agreed with @jenshu . i didn't look far enough down in the proto 😮‍💨 , but where she proposes is exactly where I'd put it

codeverifier: I was worried you guys might say that 😅, i just didnt want to touch the proto.
but agreed. made the change.

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -113,18 +114,75 @@ func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, error) {
...
 	splitRepoName := strings.Split(name, "/")
 	registry := strings.Join(splitRepoName[:len(splitRepoName)-1], "/")
 	img.Repository = &splitRepoName[len(splitRepoName)-1]
 	img.Registry = &registry
 	return img
 }
+
+func handleContainerImageStrFormat(str string) []string {
+	arr := strings.Split(str, ":")
+	// check for special case of image string following <registry:port/name>
```
jenshu, line 180: it's a bit hard to follow what this function is doing. can you document here what an example input/output would be?

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -113,18 +114,75 @@ func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, error) {
...
+	if len(arr) >= 2 && strings.Index(str, "/") > strings.Index(str, ":") {
+		copyArr := make([]string, len(arr)-2)
+		copyArr[0] = arr[0] + ":" + arr[1]
```
jenshu, line 183: seems to be a bug here: if `arr` is of length 2, then `copyArr` is initialized with length=0, and this line results in a panic trying to reference `copyArr[0]`

jenshu (changes requested):

projects/gloo/cli/pkg/cmd/version/clients.go
codeverifier: good catch, il add a test case for it.

jenshu (commented):
thanks for the comments and tests!

jenshu (dismissed):

projects/gloo/cli/pkg/cmd/version/clients.go
```diff
@@ -113,18 +114,92 @@ func (k *kube) Get(ctx context.Context) ([]*version.ServerVersion, error) {
 	return []*version.ServerVersion{serverVersion}, nil
 }
 
+func (k *kube) GetClusterVersion() (*version.KubernetesClusterVersion, error) {
+	cfg, err := kubeutils.GetConfig("", "")
+	if k.kubeContext != "" {
+		cfg, err = kubeutils.GetConfigWithContext("", "", k.kubeContext)
+	}
```
sam-heilbron, line 121: These lines feel redundant. `GetConfig` just calls `GetConfigWithContext` and passes an empty string, so I think we can just simplify this to:
```suggestion
	cfg, err = kubeutils.GetConfigWithContext("", "", k.kubeContext)
```

projects/gloo/cli/pkg/cmd/version/version_test.go
```diff
@@ -36,23 +36,47 @@ var _ = Describe("version command", func() {
 		It("will error if an error occurs while getting the version", func() {
 			fakeErr := eris.New("test")
 			client.EXPECT().Get(ctx).Return(nil, fakeErr).Times(1)
+			//client.EXPECT().GetClusterVersion().Return(nil, nil).Times(0)
```
sam-heilbron, line 39: nit: Can we delete this, or should we re-add the assertion?

sam-heilbron (dismissed):
nice work!

projects/gloo/cli/pkg/cmd/version/clients.go
codeverifier: i assume this was probably left for backwards compatibility or something and never refactored. i can remove it.

jenshu (approved):

sam-heilbron (approved):

sam-heilbron:
bulldozer

PR #9215, opened 2024-03-04
Add importas linter to enforce common package import alias

npolshakova:
# Description

Adds a basic importas linter setup to enforce common package import alias scenarios. We can expand this over time to cover gloo-specific imports. 

You can test this locally by running `make analyze` to run the linters.

## CI changes
- The linter step now runs a check to make sure imports are named consistently for corev1, appsv1 and metav1.


Makefile
```diff
@@ -150,7 +150,7 @@ check-spelling:
 # The analyze target runs a suite of static analysis tools against the codebase.
 # The options are defined in .golangci.yaml, and can be overridden by setting the ANALYZE_OPTIONS variable.
 .PHONY: analyze
-ANALYZE_OPTIONS ?= --fast --verbose
+ANALYZE_OPTIONS ?= --fast --timeout=5m --tests=true --verbose --concurrency=4
```
jenshu, line 153: it seems without the concurrency flag, it will use the number of CPUs by default. is there a reason to limit it to 4?

.golangci.yaml
```diff
@@ -22,9 +22,18 @@ linters:
     - nakedret
     - predeclared
     - usestdlibvars
+    - importas
 
 # Settings dedicated to specific linters.
 linters-settings:
+  importas:
+    alias:
```
jenshu, line 30: could also add `batchv1` since you've already fixed it in some places

sam-heilbron (dismissed):

sam-heilbron (dismissed):

.golangci.yaml
```diff
@@ -22,9 +22,20 @@ linters:
...
+      - pkg: k8s.io/api/apps/v1
+        alias: appsv1
+      - pkg: k8s.io/api/core/v1
+        alias: corev1
+      - pkg: k8s.io/apimachinery/pkg/apis/meta/v1
+        alias: metav1
+      - pkg: k8s.io/api/batch/v1
+        alias: batchv1
```
sheidkamp, line 38: Nit: add `batchv1` to the list of changes in the "CI changes" section of the PR description

Makefile
```diff
@@ -150,7 +150,7 @@ check-spelling:
 # The analyze target runs a suite of static analysis tools against the codebase.
 # The options are defined in .golangci.yaml, and can be overridden by setting the ANALYZE_OPTIONS variable.
 .PHONY: analyze
-ANALYZE_OPTIONS ?= --fast --verbose
+ANALYZE_OPTIONS ?= --fast --timeout=5m --tests=true --verbose
```
sheidkamp, line 153: nit: these look like the same values from the `.golangci.yaml` file. Did adding them here change any behavior?

sheidkamp (approved):
A nit and a question, but LGTM!

Makefile
npolshakova: Ah good catch- I don't think we need these then!

sam-heilbron (approved):

sheidkamp (approved):

sam-heilbron:
bulldozer

Issue #8849, opened 2023-10-30
Distroless


nfuden:
### Gloo Edge Product

Open Source

### Gloo Edge Version

1.15.x

### Is your feature request related to a problem? Please describe.

Alpine is no longer supported for our purposes so we moved to ubuntu to mimic upstream and decided to hold off at the time on adding a distroless build. Turns out some of the libraries in the image used by upstream are include some packages we dont want to have around.

### Describe the solution you'd like

Get a distroless option for those who dont want full ubuntu shenanagins

This SHOULD be an opt in version for both non-fips and fips. This does NOT need the debug images to be supported.
This MUST be toggleable as part of the top level helm configuration rather than a per deployment flag.
Our distroless SHOULD use most of the same work flow as our ubuntu versions
This MAY include our apiserver components but it MAY leave that to a future issue. 
Distroless MUST be supported for gloo, discovery, envoy-gloo-ee-wrapper, extauth, and rate limit servers. 
Distroless images MUST NOT include any problematic licenses such as those found but not used in ubuntu images.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206768562311555/1206790677120279) by [Unito](https://www.unito.io)

sam-heilbron:
We no longer consider this a release blocker for 1.16. This is intended to be released in a subsequent patch release. Confirmed with @SantoDE

sam-heilbron:
https://github.com/solo-io/gloo/issues/6084 is a duplicate image. I am closing that one, in favor of this more recently created issue, but I wanted to keep the context

davidjumani:
List of problematic libraries that should not be included are :
- berkleydb

davidjumani:
The following images now have a distroless variant :
- caching-ee
- discovery-ee
- discovery-ee-fips
- ext-auth-plugins
- extauth-ee
- extauth-ee-fips
- gloo-ee
- gloo-ee-envoy-wrapper
- gloo-ee-envoy-wrapper-fips
- gloo-ee-fips
- observability-ee
- rate-limit-ee
- rate-limit-ee-fips
- sds-ee
- sds-ee-fips

Having discussed on [slack](https://solo-io-corp.slack.com/archives/C028QH1MKFE/p1711124854593859), have decided to :
- Add support for fed images

davidjumani:
Adding distroless variants to all images created by gloo edge in https://github.com/solo-io/gloo/pull/9278

davidjumani:
[...]

PR #9221, opened 2024-03-05
feat: Support distroless images

davidjumani:
# Description
Adds the ability to use distroless images. To use distroless images set `global.image.variant=distroless` for standard distroless images and `global.image.variant=fips-distroless` for fips distroless images.

This PR only contains the helm changes required for enterprise

The distroless variants contain the [gcr.io/distroless/base-debian11:latest](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md) image along with the following files

For Envoy / gloo :
- openssl
- libssl
- lib64z1

For the probes / lifecycle hooks :
- sh
- wget
- sleep
- nc

## API changes
- N/A

## Code changes
- Adds the `global.image.variant` helm value. This controls the variant of the images to deploy

## CI changes
- Use the distroless variant for PR regression tests and the standard variant for nightlies. This will provide good spread and soaking for distroless as well as ensure that the standard variant is not neglected

## Docs changes
- N/A

# Context
https://github.com/solo-io/gloo/issues/8849

## Interesting decisions
 - N/A

## Testing steps
In enterprise

## Notes for reviewers
 - N/A

 - N/A

# Checklist:


install/helm/gloo/templates/_helpers.tpl
```diff
@@ -29,16 +29,35 @@ ClusterRole
 Expand the name of a container image, adding -fips to the name of the repo if configured.
```
bewebi, line 29: This description could use an update

bewebi (commented):
Looking good, a lot more readable than previously

install/helm/gloo/templates/_helpers.tpl
davidjumani: Updated!

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -26,19 +26,38 @@ ClusterRole
...
 {{- /*
 In consideration of https://github.com/solo-io/gloo/issues/7326, we want the ability for -fips images to use their own digests,
 rather than falling back (incorrectly) onto the digests of non-fips images
 */ -}}
-{{ .registry }}/{{ .repository }}-fips:{{ .tag }}@{{ .fipsDigest }}
-{{- else -}}
-{{ .registry }}/{{ .repository }}{{ ternary "-fips" "" ( and (has .repository (list "gloo-ee" "extauth-ee" "gloo-ee-envoy-wrapper" "rate-limit-ee" "discovery-ee" "sds-ee" )) (default false .fips)) }}:{{ .tag }}{{ ternary "-extended" "" (default false .extended) }}{{- if .digest -}}@{{ .digest }}{{- end -}}
-{{- end -}}
-{{- end -}}
+{{- $image = printf "%s-fips:%s@%s" $image .tag .fipsDigest -}}
```
ben-taussig-solo, line 38: This looks good to me. I assume we must already have helm tests validating that these fips tags are still generated correctly -- do you know if that's true?

davidjumani: Yes, there are tests in enterprise that I've modified to test for distroless as well

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -26,19 +26,38 @@ ClusterRole
...
+{{- else -}} {{- /* if and .fips .fipsDigest */ -}}
+{{- /* Handle only the images that have a fips and distroless variant */ -}}
+{{- if (has .repository (list "gloo-ee" "extauth-ee" "gloo-ee-envoy-wrapper" "rate-limit-ee" "discovery-ee" "sds-ee")) -}}
```
ben-taussig-solo, line 41: Can we define this list outside of this if statement? Is there any centralized location where we define which images support fips and which images support distroless?

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -26,19 +26,38 @@ ClusterRole
...
+{{- if .fips -}}
+{{- $image = printf "%s-fips" $image -}}
+{{- end -}}{{- /* if .fips */ -}}
+{{- end -}}{{- /* if (has .repository...)) */ -}}
+{{- $image = printf "%s:%s" $image .tag -}}
+{{- if (has .repository (list "gloo-ee" "extauth-ee" "gloo-ee-envoy-wrapper" "rate-limit-ee" "discovery-ee" "sds-ee" "observability-ee" "caching-ee")) -}}
```
ben-taussig-solo, line 47:
```suggestion
{{- /* Handle only the images that have a distroless variant */ -}
{{- if (has .repository (list "gloo-ee" "extauth-ee" "gloo-ee-envoy-wrapper" "rate-limit-ee" "discovery-ee" "sds-ee" "observability-ee" "caching-ee")) -}}
```

Also, similar to above, can we define this list outside of this if statement?

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -26,19 +26,38 @@ ClusterRole
...
+{{- if .distroless -}} {{- /* Add distroless suffix to the tag since it contains the same binaries in a different container */ -}}
+{{- $image = printf "%s-distroless" $image -}}
+{{- end -}}{{- /* if .distroless */ -}}
+{{- end -}}{{- /* if (has .repository...)) */ -}}
+{{- if .extended -}}
```
ben-taussig-solo, line 52: nit (formatting): Personally this section would be easier to read if there was an empty line or a comment before each new `{{- if ...-}}` block

Something like:
```diff
{{- if .distroless -}} {{- /* Add distroless suffix to the tag since it contains the same binaries in a different container */ -}}
{{- $image = printf "%s-distroless" $image -}}
{{- end -}}{{- /* if .distroless */ -}}
{{- end -}}{{- /* if (has .repository...)) */ -}}

{{- if .extended -}}
{{- $image = printf "%s-extended" $image -}}
{{- end -}}{{- /* if .extended */ -}}
```

install/helm/gloo/templates/_helpers.tpl
davidjumani: Updated!

install/helm/gloo/templates/_helpers.tpl
davidjumani: Personally, I wouldn't say I like a blank line, especially in helm .tmpl files since the output is used elsewhere in generating the charts and mishandled blank lines can break deployments

ben-taussig-solo: makes sense, thanks for the context -- resolving this thread

install/helm/gloo/templates/_helpers.tpl
```diff
@@ -26,19 +26,39 @@ ClusterRole
...
 {{- /*
 In consideration of https://github.com/solo-io/gloo/issues/7326, we want the ability for -fips images to use their own digests,
 rather than falling back (incorrectly) onto the digests of non-fips images
 */ -}}
-{{ .registry }}/{{ .repository }}-fips:{{ .tag }}@{{ .fipsDigest }}
-{{- else -}}
-{{ .registry }}/{{ .repository }}{{ ternary "-fips" "" ( and (has .repository (list "gloo-ee" "extauth-ee" "gloo-ee-envoy-wrapper" "rate-limit-ee" "discovery-ee" "sds-ee" )) (default false .fips)) }}:{{ .tag }}{{ ternary "-extended" "" (default false .extended) }}{{- if .digest -}}@{{ .digest }}{{- end -}}
-{{- end -}}
-{{- end -}}
+{{- $image = printf "%s-fips:%s@%s" $image .tag .fipsDigest -}}
```
ben-taussig-solo, line 38: if we hit this condition (i.e., `{- if and .fips .fipsDigest -}`), won't we append another `-fips` on line 43?

so on line 38 we'll have `{image}-fips:{tag}@{digest}`
and on line 43 we'll update that to be `{image}-fips:{tag}@{digest}-fips`

davidjumani:
/kick
```
Step #0 - "publish-artifacts": Step #2 - "docker-login": Error response from daemon: Get "https://quay.io/v2/": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

install/helm/gloo/templates/_helpers.tpl
davidjumani: That should not happen because of the else on line 39

bewebi (dismissed):

docs/content/reference/values.txt
```diff
@@ -9,12 +9,12 @@
...
 |settings.integrations.knative.proxy.image.registry|string||The image hostname prefix and registry, such as quay.io/solo-io.|
 |settings.integrations.knative.proxy.image.pullPolicy|string||The image pull policy for the container. For default values, see the Kubernetes docs: https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting|
 |settings.integrations.knative.proxy.image.pullSecret|string||The image pull secret to use for the container, in the same namespace as the container pod.|
-|settings.integrations.knative.proxy.image.extended|bool||If true, deploys an extended version of the container with additional debug tools.|
-|settings.integrations.knative.proxy.image.fips|bool||If true, deploys a version of the data-plane containers that is built with FIPS-compliant crypto libraries. (Enterprise-only feature.)|
+|settings.integrations.knative.proxy.image.variant|string||Specifies the version of the data-plane containers to deploy. Can take the values 'standard', 'fips', 'distroless', 'fips-distroless'. Defaults to standard. (Enterprise-only feature.)|
```
ben-taussig-solo, line 15: Not strictly relevant to this PR but I'm pretty sure we haven't supported the knative integration for some time -- I'm not sure if we actually build this image anymore

This isn't a blocker for this PR in my eyes but I'm going to spin off another issue about this

ben-taussig-solo: Created that here: https://github.com/solo-io/gloo/issues/9245

Closing this thread

install/test/helm_test.go
```diff
@@ -3007,16 +3007,21 @@ spec:
 						testManifest.Expect("Deployment", namespace, "gateway-proxy").NotTo(BeNil())
 					})
 
-					It("supports deploying the fips envoy image", func() {
+					DescribeTable("supports deploying the specified variant of the envoy image", func(variant string) {
 						prepareMakefile(namespace, helmValues{
 							valuesArgs: []string{
-								"global.image.fips=true",
```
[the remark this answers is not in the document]
davidjumani: I've re-added a single test to support the deprecated `global.image.fips` value since the flow has not changed that much with these changes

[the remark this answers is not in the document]
ben-taussig-solo: new test looks great! approving the PR

ben-taussig-solo (dismissed):
LGTM!

bewebi (dismissed):

bewebi (dismissed):

bewebi (approved):

.github/workflows/nightly-tests.yaml
```diff
@@ -47,6 +47,8 @@ jobs:
         kube-e2e-test-type: ['gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'glooctl', 'upgrade']
         kube-version: [ { node: 'v1.24.15@sha256:7db4f8bea3e14b82d12e044e25e34bd53754b7f2b0e9d56df21774e6f66a70ab', kubectl: 'v1.24.17', kind: 'v0.20.0', helm: 'v3.12.2' },
                         { node: 'v1.28.0@sha256:b7a4cad12c197af3ba43202d3efe03246b3f0793f162afb40a33c923952d5b31', kubectl: 'v1.28.4', kind: 'v0.20.0', helm: 'v3.13.2' }]
+        image-variant:
+          - standard
```
ben-taussig-solo, line 51: I left the same comment on the enterprise PR, but why not just run nightlies on both distroless and standard?

davidjumani: Since they are essentially the same binary just in a different container, I thought it would be good to run distroless on PRs since they have more action. Running them on both would just be a waste of resources

test/kube2e/upgrade/upgrade_suite_test.go
```diff
@@ -41,6 +42,8 @@ var (
 	firstReleaseOfMinor                bool
 
 	skipIfFirstMinorFunc func()
+
+	variant string
```
ben-taussig-solo, line 46: Nit: I think it might be cleaner to define `variant` in `test/kube2e/upgrade/upgrade_test.go`, in the same location where we define `strictValidation`

https://github.com/solo-io/gloo/blob/fe6feb6138231485122cc77f1b68888d0a1fe0a1/test/kube2e/upgrade/upgrade_test.go#L41-L42

Right now, I think it's slightly confusing that the value of `variant` is set in `upgrade_suite_test.go` but only consumed in `upgrade_test.go`

davidjumani: The way I see it is that it needs to be fetched only once so I've set it in the `upgrade_suite_test.go` to set up the test while `strictValidation` does change values between tests. Happy to change if you think it is better in `upgrade_test.go`

install/helm/gloo/values-template.yaml
```diff
@@ -224,7 +224,6 @@ global:
   image:
     registry: quay.io/solo-io
     pullPolicy: IfNotPresent
-    extended: false
```
ben-taussig-solo, line 227: removed because false is the default value?

davidjumani: Since it is no longer supported

Makefile
```diff
@@ -461,6 +480,18 @@ gloo-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfi
...
+$(GLOO_OUTPUT_DIR)/Dockerfile.gloo.distroless: $(GLOO_DIR)/cmd/Dockerfile.distroless
+	cp $< $@
+
+# Explicitly specify the base image is amd64 as we only build the amd64 flavour of gloo envoy
+.PHONY: gloo-distroless-docker
+gloo-distroless-docker: $(GLOO_OUTPUT_DIR)/gloo-linux-$(GOARCH) $(GLOO_OUTPUT_DIR)/Dockerfile.gloo.distroless
+	docker buildx build --load $(PLATFORM) $(GLOO_OUTPUT_DIR) -f $(GLOO_OUTPUT_DIR)/Dockerfile.gloo.distroless \
+		--build-arg GOARCH=$(GOARCH) \
+		--build-arg ENVOY_IMAGE=$(ENVOY_GLOO_IMAGE) \
+		--build-arg BASE_IMAGE=$(DISTROLESS_BASE_IMAGE)-amd64 \
```
ben-taussig-solo, line 492: correct me if I'm looking in the wrong place but it seems like the Gloo dockerfile doesn't support a `BASE_IMAGE` arg: https://github.com/solo-io/gloo/blob/15e6e83cdfd8bf8c9bd2ce609ffec2d4a79ce5fb/projects/gloo/cmd/Dockerfile#L1-L16

test/kube2e/upgrade/upgrade_suite_test.go
davidjumani: Moved them to _test.go to make it more readable

ben-taussig-solo: Ah, I see your point -- the new change looks good, resolving this thread

.github/workflows/nightly-tests.yaml
ben-taussig-solo: Fair enough, I can understand not wanting to double the amount we spend on nightly tests with one PR.

Makefile
```diff
@@ -721,8 +784,12 @@ docker: check-go-version
...
-docker: certgen-docker
 docker: sds-docker
+docker: gloo-distroless-docker
+docker: discovery-distroless-docker
+docker: gloo-envoy-wrapper-distroless-docker
+docker: sds-distroless-docker
+docker: certgen-docker
 docker: ingress-docker
 docker: access-logger-docker
 docker: kubectl-docker
```
ben-taussig-solo, line 795: why don't we have distroless variants of the ingress, access-logger, and kubectl images?

davidjumani: Based on the requirements mentioned [here](https://github.com/solo-io/gloo/issues/8849), we're targeting the primary images we produce. This will be the first phase and subsequently move other images to distroless as and when it seems fit

projects/envoyinit/cmd/Dockerfile.envoyinit.distroless
```diff
@@ -0,0 +1,42 @@
+ARG ENVOY_IMAGE
+ARG BASE_IMAGE
+
+FROM busybox:1.35.0-uclibc as busybox
```
ben-taussig-solo, line 4: any particular reason we're using this specific busybox tag? what do we get from uclibc?

ben-taussig-solo: if there's one image that needs uclibc and we're just using the same busybox version on the rest for consistency, that is fine with me

Makefile
ben-taussig-solo: makes sense! resolving the thread.

ben-taussig-solo (approved):
Looks good to me, thanks for answering all the questions!

projects/envoyinit/cmd/Dockerfile.envoyinit.distroless
ben-taussig-solo: note: this is a nit and I've approved the PR in the background, just looking for clarification

davidjumani: Good question. I forgot to add the doc in the dockerfile but it can stay on the PR.
The uclibc variant of distroless has a statically compiled shell that works as a standalone binary (at least in the cases we need). I tried other variants and unfortunately, it wouldn't work as a standalone when copied over to the distroless container. The first pass was to use the distroless debug image but since it had a lot of unnecessary binaries I pivoted to copy over only the necessary binaries to keep the container as minimal as possible

ben-taussig-solo: makes sense, thanks!

bewebi (approved):

PR #6408, opened 2022-05-03
[WIP] Add grpc server for proxy endpoint

elcasteel:
# Description
Show interface for GRPC call to get proxies. 
TODO: starting server, configuration, integration with glooctl. 

# Context

Users ran into this bug doing ... \ Users needed this feature to ...

# Checklist:

- [ ] I included a concise, user-facing changelog (for details, see https://github.com/solo-io/go-utils/tree/master/changelogutils) which references the issue that is resolved.
- [ ] If I updated APIs (our protos) or helm values, I ran `make -B install-go-tools generated-code` to ensure there will be no code diff
- [ ] I followed guidelines laid out in the Gloo Edge [contribution guide](https://docs.solo.io/gloo-edge/latest/contributing/)
- [ ] I opened a draft PR or added the work in progress label if my PR is not ready for review
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #6661, opened 2022-07-01
Delete helm upgrade e2e tests once those gloo versions are no longer supported


jenshu:
### Gloo Edge Version

1.12.x (beta)

### Describe the bug

There are certain tests in the [kube2e helm test suite](https://github.com/solo-io/gloo/blob/d4cd6a26680ecf2d4b67952ea13163c48aee9a20/test/kube2e/helm/helm_test.go) that test upgrading from specific gloo edge versions to the most recent version (see any tests referencing 1.11.x)

This is just a reminder to delete those tests once the versions are no longer supported.

### Steps to reproduce the bug

N/A

### Expected Behavior

N/A

sam-heilbron:
This has been resolved in Gloo 1.17 in both open source and enterprise

PR #9226, opened 2024-03-07
cli: get proxy API can query proxies from k8s-gateway

sam-heilbron:
# Description

Enhance the CLI utilities that return proxies to users to also support proxies created by the k8s gateway translation

# Context

## Background
Before these changes, Gloo Gateway exposed a debug endpoint (https://github.com/solo-io/gloo/pull/6408) that would allow users/operators to query the control plane, and see the state of proxies that were persisted in memory.

### Drawbacks
**Smart CLI**: The CLI code is aware of where proxies are stored, and attempts to [either query etcD or the control plane](https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/common/get.go#L133). This means there is complexity that is built into the CLI (client) that could otherwise be built into the server (API).

**Namespace Doesn't Work**: The existing code looks for proxies in the [`opts.Metadata.Namespace` namespace](https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/common/get.go#L134). This is supplied via `-n` in the command line, and is used to represent the namespace where Gloo is installed. However, it is also being used to represent the namespace to look for the Proxies. This is ok on a default install, but if you configure Gloo to persist proxies in a namespace that is different from where the Control Plane is installed, you can no longer use `glooctl get proxy -n [....]`.

**K8s Gateway Translation Proxies**: The existing code only works for Proxy resources generated by the Edge Gateway translator. There is a new translation that runs, and it takes K8s Gateway resources and produces a Proxy. During this translation, operators have no way to inspect the state of the generated Proxy.

### Port Forwarder
The Port Forwarder is a solution to make it easier to port-forward to a Kubernetes pod. It is inspired by both the Istio and Gloo Mesh Enterprise projects, and in code I include `Inspired by` and a link to the relevant code.

**CliPortForwarder**: This is an implementation that mirrors what we used to do, rely on the Kubernetes CLI
**ApiPortForwarder**: This is an implementation that mirrors what GME and Istio do, rely on the Kubernetes API directly

### ProxyEndpointServer Improvements
The previous server only supported a single ProxyClient. This worked because the Control Plane only ever stored Proxies in a single place (either in memory, or in etcD). With the new K8s Gateway translation we have:
- Proxies generated by the k8s gateway translation, that are maintained in memory
- Proxies generated by the edge gateway translation, that are either maintained in memory or etcD, depending on the Setting `persistProxySpec`

As a result, our server needs to support "looking" for proxies in multiple locations. To enable this, I expand the API to support registry multiple ProxyReaders. 

In addition, I create amend the GetProxies API to include a "source" field. This allows consumers of the API (in this case, the CLI) to request proxies either generated by the k8s gateway translation, the edge translation, or both.

### CLI Improvements
The CLI code had more knowledge of where proxies were persisted than it should have. I migrated that "smarts" to the server, and simplified the CLI code to just query the endpoint with the request properties.

In addition, I exposed new flags for the user to provide, so more easily filter the proxies returned.

### Helm Tests
When I made changes to the CLI, I noticed that our Helm tests failed. They failed because the ProxyDebugBindAddr was empty in the tests. The old CLI code allowed for this to occur, because if there was an error identifying the port to query (which an empty address would cause that error), it would fallback and query etcD directly.

It confused me that the bind address would be empty, because there is explicitly a default value applied in our Helm chart. Then I realized that our helm tests attempt to upgrade from Gloo 1.9. So what I believed was happening was:
1. Install gloo
2. Default settings CR is created, without the proxyDebugBindAddr, because it didn't exist yet
3. We fail to parse an empty BindAddr, and error during `glooctl check`

I went with the quick solution, to just start our upgrades at a more recent version, though I left a comment outlining that the tests should be more dynamic.

In addition, I removed tests that relied on v1.11, which is now outside of the support window.

### Unit Tests
I introduced a full set of unit tests to demonstrate the behavior of the ProxyEndpointServer

### Kube2e Tests
All kube tests install gloo, and then perform a `glooctl check`. Under the hood, that command invokes the CLI code that I changed (the port-forwarding, and querying the proxy endpoint). Since the kube2e tests are all passing, I'm confident in the various components that I changed

### Manual CLI Tests
Produce a local CLI:
```
make build-cli-local
```

Various commands using the CLI. Try running these before proxies are created, and then add Edge and K8s Gateway resources to produce proxies and invoke these commands.
```
 _output/glooctl-darwin-arm64 check
```
```
 _output/glooctl-darwin-arm64 get proxy
```
```
 _output/glooctl-darwin-arm64 get proxy --proxy-ns=default
```
```
 _output/glooctl-darwin-arm64 get proxy --proxy-ns=default --kube
```
```
 _output/glooctl-darwin-arm64 get proxy --proxy-ns=default --edge
```
```
 _output/glooctl-darwin-arm64 get proxy --all
```

## Notes for reviewers
-

# Checklist:


projects/gateway2/xds/xds_syncer.go
```diff
@@ -383,3 +390,21 @@ func (s *XdsSyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl ap
 		}
 	}
 }
+
+// syncProxyCache persists the proxies that were generated during translations and stores them in an in-memory cache
+// This cache is utilized by the debug.ProxyEndpointServer
+func (s *XdsSyncer) syncProxyCache(ctx context.Context, proxyList gloo_solo_io.ProxyList) {
+	ctx = contextutils.WithLogger(ctx, "proxyCache")
```
npolshakova, line 397: Can we specify this is the ggv2 proxyCache here with the logger?

test/kube2e/helm/helm_test.go
```diff
@@ -339,83 +333,6 @@ var _ = Describe("Kube2e: helm", func() {
 			Expect(*validationWebhook.Webhooks[0].TimeoutSeconds).To(Equal(int32(5)))
 		})
 
-		// Below are tests with different combinations of upgrades with failurePolicy=Ignore/Fail.
-		Context("failurePolicy upgrades", func() {
```
npolshakova, line 343: Why is this test getting removed? Won't it still work for the --edge gateway proxies?

npolshakova:
Looks good- one minor question: You mentioned `glooctl get proxy -n [....]` doesn't work for proxies created in a namespace that is different from where the Control Plane is installed. So if I create a k8s Gateway resource in the `test` namespace, and the control plane is installed in `gloo-system` I wouldn't be able to filter via namespace or list all proxies across namespaces?

projects/gloo/cli/pkg/common/get.go
```diff
@@ -124,138 +111,6 @@ func GetSettings(opts *options.Options) (*gloov1.Settings, error) {
 	return client.Read(opts.Metadata.GetNamespace(), defaults.SettingsName, clients.ReadOpts{Ctx: opts.Top.Ctx})
 }
 
-func GetProxies(name string, opts *options.Options) (gloov1.ProxyList, error) {
-	settings, err := GetSettings(opts)
-	if err != nil {
-		return nil, err
-	}
-	proxyEndpointPort := computeProxyEndpointPort(opts.Top.Ctx, settings)
```
npolshakova, line 132: Did this move or did you end up removing the logic on the client side?

sam-heilbron:
> Looks good- one minor question: You mentioned `glooctl get proxy -n [....]` doesn't work for proxies created in a namespace that is different from where the Control Plane is installed. So if I create a k8s Gateway resource in the `test` namespace, and the control plane is installed in `gloo-system` I wouldn't be able to filter via namespace or list all proxies across namespaces?

Yeah great point. I was trying to get at the _historically_ this did not work, because the code used the value of `-n` to determine where the settings object was (ie where gloo was installed) and where the proxies are persisted.

In my PR, I introduced a flag in the get proxy api, called `proxy-ns`. This way you could run `glooctl get proxy -n gloo-system -proxy-ns test` to inspect the state of proxies that are managed by a gloo install running in gloo-system, but proxies persisted in `test`. 

An alternative would be to use this api (https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/flagutils/metadata.go#L30), but tbh it's so buried in the API, I opted to just add a new one directly in the proxy api

test/kube2e/helm/helm_test.go
sam-heilbron: Inside this text of removed testing, you'll notice a link to https://github.com/solo-io/gloo/issues/6661. This test was specifically testing that we can upgrade from a version of gloo that is now outside of our support window, and that issue was saying "once it's outside of our support wiindow, we can remove the test". 

@jenshu is this safe to do?

projects/gateway2/xds/xds_syncer.go
sam-heilbron: Sure thing! The other `sync` operations in this file also don't prefix with k8s-gateway or anything, shoudl I update those as well?

projects/gloo/cli/pkg/common/get.go
sam-heilbron: Since there were so many functions specific to proxies, I just split this code out into a `proxy.go` file in the same package. The `computeProxyEndpointPort` function is still there, though I added some comments to help clarify behavior I had run into. Specifically, I hit errors on upgrade tests when installing an older version of gloo, that pre-dated this API

sam-heilbron:
> Looks good- one minor question: You mentioned `glooctl get proxy -n [....]` doesn't work for proxies created in a namespace that is different from where the Control Plane is installed. So if I create a k8s Gateway resource in the `test` namespace, and the control plane is installed in `gloo-system` I wouldn't be able to filter via namespace or list all proxies across namespaces?

Great call. The only way to do that would have been to cheat and supply an empty namespace: `glooctl get proxy --proxy-ns ""`. I don't think that's a good UX, so I added a `--all`  flag: 
`glooctl get proxy` will return you proxies in the default (gloo-system) namespace
`glooctl get proxy --proxy-ns test` will return you proxies in the test namespace
`glooctl get proxy --all` will return you proxies in all namespaces

projects/gateway2/xds/xds_syncer.go
sam-heilbron: I ended up leaving this logger with the exact name, but modified the context.Logger in the function that calls this to include some k8s gateway reference, let me know what you think

projects/gloo/pkg/debug/proxy_endpoint.go
```diff
@@ -11,43 +15,119 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, reader := range proxyReaders {
+		proxy, readErr := reader.Read(namespace, name, clients.ReadOpts{Ctx: ctx})
+		if readErr == nil {
+			return proxy, nil
+		} else {
+			// continue, this just means that one of the readers didn't have a proxy with that name
```
npolshakova, line 80: Should we check if it's a "not found" error here and report the real errors otherwise?

npolshakova (commented):
Looks good! Just a couple minor questions!

projects/gloo/pkg/debug/proxy_endpoint.go
```diff
@@ -11,43 +15,119 @@ import (
...
 		})
-		if err != nil {
-			return nil, err
-		}
-		return &debug.ProxyEndpointResponse{Proxies: proxies}, nil
-	} else {
-		proxy, err := p.proxyClient.Read(req.GetNamespace(), req.GetName(), clients.ReadOpts{Ctx: ctx})
-		if err != nil {
-			return nil, err
+		if listErr == nil {
```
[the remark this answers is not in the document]
sam-heilbron: Yeah on this one, I wasn't sure what to do. If we make 3 list calls, and one succeeds, then I figure we should return the list of proxies from the one that succeded, and we kind of have to swallow the other failures.

projects/gloo/pkg/debug/proxy_endpoint.go
sam-heilbron: good call! updated

projects/gloo/cli/pkg/common/proxy.go
```diff
@@ -0,0 +1,194 @@
...
+			// Some proxies can become very large and exceed the default 100Mb limit
+			// For this reason we want remove the limit but will settle for a limit of MaxInt32
+			// as we don't anticipate proxies to exceed this
+			grpc.MaxCallRecvMsgSize(math.MaxInt32),
+		)
+		proxyEndpointResponse = r
+		return err
+	},
+		retry.LastErrorOnly(true),
+		retry.Delay(100*time.Millisecond),
```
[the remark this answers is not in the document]
sam-heilbron: i like it!

projects/gloo/cli/pkg/cmd/options/options.go
```diff
@@ -100,9 +100,17 @@ type Upgrade struct {
 }
 
 type Get struct {
+	Proxy    GetProxy
 	Selector InputMapStringString
 }
 
+type GetProxy struct {
```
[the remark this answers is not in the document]
sam-heilbron: I updated the comment around this object to try to answer this question, let me know if there's more I can add to it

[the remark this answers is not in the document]
npolshakova: Thanks!

projects/gloo/pkg/debug/proxy_endpoint.go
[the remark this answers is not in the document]
npolshakova: If we swallow the errors on other cli commands, I think this behavior is fine then. I think it's also reasonable to display both the results and errors, but if we're not doing that in other places I think this is more consistent.

npolshakova (dismissed):

projects/gloo/cli/pkg/cmd/options/options.go
```diff
@@ -100,9 +100,32 @@ type Upgrade struct {
...
+	// An empty string is interpreted as all namespaces
+	Namespace string
+
+	// K8sGatewaySource is true if the GetProxy request should only consider Proxy resources
+	// that were generated by the K8s Gateway Translation
+	K8sGatewaySource bool
+
+	// EdgeGatewaySource is true if the GetProxy request should only consider Proxy resources
+	// that were generated by the Edge Gateway Translation
+	EdgeGatewaySource bool
```
jenshu, line 123:
```suggestion
	// K8sGatewaySource is true if the GetProxy request should consider Proxy resources
	// that were generated by the K8s Gateway Translation
	K8sGatewaySource bool

	// EdgeGatewaySource is true if the GetProxy request should consider Proxy resources
	// that were generated by the Edge Gateway Translation
	EdgeGatewaySource bool
```
according to the implementation, both flags can be true at the same time, so probably more accurate to remove the word `only` in the 2 descriptions above

projects/gloo/pkg/syncer/setup/start_func.go
```diff
@@ -44,12 +49,29 @@ func ExecuteAsynchronousStartFuncs(
...
+		inMemoryProxyClient, err := v1.NewProxyClient(ctx, &factory.MemoryResourceClientFactory{
+			Cache: memory.NewInMemoryResourceCache(),
+		})
+		if err != nil {
+			return err
+		}
+
+		if opts.ProxyDebugServer.Server != nil {
+			// If we have a debug server running, let's register the proxy client used by
+			// the k8s gateway translation. This will enable operators to query the debug endpoint
```
jenshu, line 62: in what circumstances would the debug server not be running, and in that case what's the behavior of `glooctl get proxy`?

pkg/utils/kubeutils/names.go
```diff
@@ -0,0 +1,5 @@
+package kubeutils
+
+const (
+	GlooDeploymentName = "gloo"
```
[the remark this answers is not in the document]
sam-heilbron: It's only used in CLI code for now, but we make lots of references in our tests to the deployment names, so I figured it could be useful to store the constants outside the CLI. What do you think?

projects/gloo/api/grpc/debug/proxy_endpoint.proto
```diff
@@ -16,14 +16,20 @@ service ProxyEndpointService {
...
   string name = 2;
 
   // Optional. Selector to use to filter returned proxies. This will be ignored if a name is provided.
   map<string, string> selector = 3;
+
+  // Optional. The translator source that produced the requested Proxy. Available values are:
+  //  1. edge-gw: If the Proxy was created using the Edge Gateway resources
+  //  2. k8s-gw: If the Proxy was created using the K8s Gateway resources
+  // If this is omitted, all sources will be considered.
+  string source = 4;
```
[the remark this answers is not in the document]
sam-heilbron: Yeah I intentionally avoided the Enum route. I figured this was an additioanl type of filter, and it would be better easier to add new sources (if they ever existed) without future updates to the API if we left this a string. In the end, our APi is really just consumed by the CLI (we should still think of the API as though users are invoking it), so I figured that this complexity is also hidden to end users

projects/gloo/pkg/syncer/setup/start_func.go
sam-heilbron: It should never be happening, but I wanted to avoid any nil pointer exception so I guarded the logic behind the conditional

docs/content/reference/cli/glooctl_get_proxy.md
```diff
@@ -17,7 +17,11 @@ glooctl get proxy [flags]
 ### Options
 
 ```
-  -h, --help   help for proxy
+      --all               get all proxies
+      --edge              include proxies produced from edge gateway resources
+  -h, --help              help for proxy
+      --kube              include proxies produced from k8s gateway resources
+      --proxy-ns string   namespace where proxies are persisted (default "gloo-system")
```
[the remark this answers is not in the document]
sam-heilbron: `--namespace` is a flag that represents where gloo is installed. So if you run `glooctl get proxy --namespace test`, we will assume that Gloo is installed in the `test` namespace, and we will look for a Sttings resources in that same namespace. 
At the moment, if --proxy-ns is unset, it will just default to `gloo-system` which is what each of our other CLI flags do. If we were to try to be more opinionated about "dynamically" defaulting the value (ie if ns is provided, but proxy-ns isn't, make proxy-ns the same), I think we would need to build that logic into the CLI itself, and not just rely on default values. 
I'm ok if we want to take that approach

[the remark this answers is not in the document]
jenshu: i'm good with this for now

pkg/utils/kubeutils/names.go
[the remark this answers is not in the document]
jenshu: sounds good to me

npolshakova (approved):

changelog/v1.17.0-beta12/support-get-proxy-api.yaml
```diff
@@ -0,0 +1,13 @@
+changelog:
+  - type: NON_USER_FACING
```
nfuden, line 2: I havent tested it but it seems like this would be a breaking change. 
Namely a new glooctl version querying an older gloo version that has its proxies not available via the grpc endpoint will not work. Its not a huge deal as thats a relatively old version. That being said I think it needs to be called out because historically we have tried to make sure that older users can still use a new glooctl.

sam-heilbron: Good point! Thought for `historically we have tried to make sure that older users can still use a new glooctl.` I feel like our stance has been that CLI and server should be in sync, no?

jenshu (approved):

nfuden (approved):

PR #9227, opened 2024-03-07
k8s-gateway: support configurable deployer and plugin translation

sam-heilbron:
# Description

Support more injectable functionality, to enable Gloo Gateway Enteprise

# Context

## Deployer API
There are ongoing discussions around how to expose an API for the Deployer. In the meantime, we need a simple way to define what binaries to run in the proxy workload. 

I create a minimal "experimental" API, using an environment variable, so that we can run the Enterprise control plane, and have it configure proxies that rely on the Enterprise binary as well

## Plugin Translation
The factory that produces the plugins to be used in translation needs to come from the Extensions object that Gloo is setup with. This is the injection point where we supply enterprise plugins.

I discovered this when I ran Gloo Gateway Enterprise, and realized that none of the Enterprise plugins were being evaluated.

## Interesting decisions
-

## Testing steps
I have been using https://github.com/solo-io/solo-projects/issues/5874#issuecomment-1977969418 to validate these (and other changes). It links to https://github.com/solo-io/edge-shared-resources/pull/3, which is where I've documented some scripts/resources.

Note: This script is out of date. I used it when all I needed to do was inject the image.tag, but I've included the image repo as part of this as well

I also included unit tests in the deployer to demonstrate how the default and override behaviors will work.

## Notes for reviewers
-

# Checklist:


jenshu (commented):
looks good, do you have a solo-projects PR that verifies the image is passed through correctly and that enterprise plugins are being called?

projects/gateway2/deployer/deployer.go
```diff
@@ -309,3 +312,20 @@ func ConvertYAMLToObjects(scheme *runtime.Scheme, yamlData []byte) ([]client.Obj
 
 	return objs, nil
 }
+
+func getDeployerImageValues() map[string]any {
+	image := os.Getenv("GG_EXPERIMENTAL_DEPLOYER_IMAGE")
+	if image == "" {
+		// Fallback to the default values
+		return map[string]any{
+			"tag":        "",
```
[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/blob/main/projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml#L44 Yeah the image.tag will fallback to the chart version if it's not supplied. I could probably go in and clean this up a bit...at first I tried to just pass nil values if the env variable wasn't set, hoping that the defaults would be used, but that broke things.

projects/gateway2/controller/start.go
```diff
@@ -66,7 +73,10 @@ func Start(ctx context.Context, cfg StartConfig) error {
 	// TODO: replace this with something that checks that we have xds snapshot ready (or that we don't need one).
 	mgr.AddReadyzCheck("ready-ping", healthz.Ping)
 
-	glooTranslator := newGlooTranslator(ctx)
+	glooTranslator := translator.NewDefaultTranslator(
+		cfg.Settings,
```
npolshakova, line 77: Are the settings dynamically picked up by the ggv2 controller? Or only picked up on startup?

projects/gateway2/deployer/deployer.go
```diff
@@ -309,3 +312,20 @@ func ConvertYAMLToObjects(scheme *runtime.Scheme, yamlData []byte) ([]client.Obj
 
 	return objs, nil
 }
+
+func getDeployerImageValues() map[string]any {
+	image := os.Getenv("GG_EXPERIMENTAL_DEPLOYER_IMAGE")
```
npolshakova, line 317: Do you need to supply the full image value?  
```
  - env:
    - name: GG_EXPERIMENTAL_DEPLOYER_IMAGE
      value: "quay.io/solo-io/my-gloo:1.0.0"
```

So this wouldn't support overwriting the whole image like we do with the helm chart: 

```
  image:
    repository: proxyv2
    registry: docker.io/istio
    tag: 1.18.2
    pullPolicy: IfNotPresent
    extended: false
```

I think we should add a comment that this is a temporary workaround until we fix the helm value piping.

projects/gateway2/controller/start.go
sam-heilbron: The settings object is the core of translation, so anytime the settings resource changes, the entire "setup" look is triggered. This means that we call controller.Start everytime these change. In this code, we are just piping the values that the controller should be using, instead of passing empty objects.

projects/gateway2/deployer/deployer.go
sam-heilbron: The proxy-deployment chart is structured slightly differently: https://github.com/solo-io/gloo/blob/main/projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml#L44, in that it just accepts a repository and tag. I'm happy to add a comment outlining that this is just temporary

npolshakova: Sounds good- if that's all it needs I think this is fine. Let's add a comment that the env is temporary so we remember to switch it later.

projects/gateway2/controller/start.go
```diff
@@ -35,6 +39,9 @@ var (
 type StartConfig struct {
 	Dev          bool
 	ControlPlane bootstrap.ControlPlane
+
+	Settings              *v1.Settings
+	PluginRegistryFactory plugins.PluginRegistryFactory
```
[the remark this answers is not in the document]
sam-heilbron: Yeah great point. I'll go with the `gloo` prefix for now as you propose

projects/gateway2/deployer/deployer.go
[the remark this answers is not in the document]
sam-heilbron: I've pushed a unit test to prove out this functionality (both the default behavior, and the override behavior)

lgadban (approved):

npolshakova (approved):

sam-heilbron:
bulldozer pleeease

Issue #8580, opened 2023-08-14
Generated CRDs should get validation rules from protos


elcasteel:
### Gloo Edge Product

Open Source

### Gloo Edge Version

1.15.*

### Is your feature request related to a problem? Please describe.

We have annotations in some of our protos to indicated fields that are required/have constraints. However, these constraints are not added to the CRD schemas so we can't rely on the k8s apiserver to enforce them for us. Thus we have a lot of checks in Gloo code for things that are already defined on the proto. This approach is fairly error-prone as we sometimes miss adding in requirements. 
See: https://github.com/solo-io/gloo/issues/8573 for a recent example.

PR #9233, opened 2024-03-13
Add GatewayParameters CRD for dynamic proxy configuration

jenshu:
# Description

Add support for customization of the dynamically provisioned proxy resources in Gloo Gateway. The customization is done via a new `GatewayParameters` CRD which enables configuring various aspects of the resources (deployment, service, etc) that are created by the GG deployer. A GatewayParameters is attached to a Gateway via an annotation `gateway.gloo.solo.io/gateway-parameters-name` on the Gateway, which points to the name of a GatewayParameters in the same namespace as the Gateway.

Example configuration of a Gateway and GatewayParameters is shown below (not valid values, just showing what all the configurable fields are):

```
kind: Gateway
metadata:
  name: http
  namespace: gloo-system
  annotations:
    gateway.gloo.solo.io/gateway-parameters-name: "my-gw-params"
spec:
  listeners:
  - protocol: HTTP
    port: 8080
    name: http
    allowedRoutes:
      namespaces:
        from: All
---
kind: GatewayParameters
metadata:
  name: my-gw-params
  namespace: gloo-system
spec:
  kube:
    deployment:
      replicas: 3
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
      securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        fsGroup: 2000
      imagePullSecrets:
      - name: my-secret
      nodeSelector:
        node-sel-key: node-sel-value
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: disktype
                operator: In
                values:
                - ssd
      tolerations:
      - key: "key1"
        operator: "Equal"
        value: "value1"
        effect: "NoSchedule"
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
      image:
        registry: quay.io/solo-io-test-123
        repository: gloo-envoy-test-repo
        tag: 1.2.3
        pullPolicy: Always
      securityContext:
        allowPrivilegeEscalation: true
        privileged: true
        runAsGroup: 4444
      resources:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
          memory: "128Mi"
          cpu: "500m"
    service:
      type: ClusterIP
      clusterIP: None
      extraLabels:
        svc-label-key: svc-label-val
      extraAnnotations:
        svc-anno-key: svc-anno-val
```

## Code changes
- Added the protos and codegen to generate the GatewayParameters CRD and Go types.
    - Note: GE classic already has code to generate go structs etc from protos using solo-kit. For GG, we are generally using skv2 and controller-runtime, not the solo-kit/snapshot framework, so decided to keep the (skv2-based) GG codegen separate from (solo-kit-based) GE codegen.
    - imported some k8s core and apimachinery protos, to avoid needing to redefine the structs in our protos
- The deployer now checks for a GatewayParameters annotation on the Gateway to get any custom config values, and merges those with the (previously used) default values.
- Updated the way we inject the envoy image values (repo and tag) from GlooEE. It is now done via a new `GetEnvoyImage` func (which returns different values in OSS vs EE) in K8sGatewayExtensions, which is passed to the GG deployer. Removed the temporary K8sGatewayExtensions workaround that was being used before. 

## Notes:
- Modifying a GatewayParameters object that is already being referenced by a Gateway will cause the deployer to re-run and update the proxy deployment.
- Adding/removing/modifying the params annotation on a Gateway will not in itself cause the deployer to re-run, since annotation changes don't trigger Gateway reconciliation. Once the Gateway parametersRef field is available and we move over to using that instead of the annotation, this issue should be resolved.

## Testing:
- There are deployer unit tests covering many scenarios
- e2e enterprise tests are in progress and tracked separately

# Checklist:


projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,56 @@
...
+  Autoscaling autoscaling = 4;
+}
+
+message ProxyDeployment {
+  // The number of desired pods. Defaults to 1.
+  google.protobuf.UInt32Value replicas = 1;
+
+  Pod pod_template = 2;
+
+  Container proxy_container = 3;
```
npolshakova, line 53: So I think this would work, but I'm having a hard time visualizing it without writing it out:

One of the issues with the current `Container` is it's missing some fields the k8s Container [field](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3879) (which is just the [EphemeralContainer](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3784) api) has which are currently configurable in the gateway deployment through helm values. 

For example, you can set the log level via the [arg](https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L395) for the istio-proxy:

```
        args:
        - proxy
        - sidecar
        - --domain
        - $(POD_NAMESPACE).svc.cluster.local
        - --configPath
        - /etc/istio/proxy
        - --serviceCluster
        - istio-proxy-prometheus
        - --drainDuration
        - 45s
        - --parentShutdownDuration
        - 1m0s
        - --proxyLogLevel={{ $global.glooMtls.istioProxy.logLevel }}
        - --proxyComponentLogLevel=misc:error
```

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+  ResourceRequirements resources = 4;
+}
+
+message EnvoyLogging {
+  // Envoy log level. Options include "trace", "debug", "info", "warn", "error",
+  // "critical" and "off". Defaults to "info".
+  string log_level = 1;
+}
+
+message GatewayConfigStatus {}
```
lgadban, line 75: IMO this should be a K8s style status -- in which case I would also argue it should be omitted for now and tracked separately

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+
+  k8s.io.api.core.v1.SecurityContext security_context = 3;
+
+  ResourceRequirements resources = 4;
+}
+
+message EnvoyLogging {
+  // Envoy log level. Options include "trace", "debug", "info", "warn", "error",
+  // "critical" and "off". Defaults to "info".
+  string log_level = 1;
```
lgadban, line 72: probably should be enum?

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/service.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/serviceaccount.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+
+import "google/protobuf/wrappers.proto";
+
+// A GatewayConfig contains configuration that is used to dynamically provision
+// the data plane (Envoy proxy instance), based on a Kubernetes Gateway or
+// GatewayClass.
+message GatewayConfigSpec {
```
lgadban, line 22: wondering if this should just be ProxyConfig at the top-level? and no `GatewayConfig`

the data plane point seems valid though, but eventually will probably grow beyond just the Proxy (e.g. it may contain ExtAuth and RateLimit config etc)

maybe `GlooDataplaneConfig` ?

projects/gateway2/api/v1alpha1/pod.proto
```diff
@@ -0,0 +1,23 @@
...
+
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
+
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+
+message Pod {
+  map<string, string> annotations = 1;
+
```
lgadban, line 13: should we have labels as well?

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+}
+
+message ProxyDeployment {
+  // The number of desired pods. Defaults to 1.
+  google.protobuf.UInt32Value replicas = 1;
+
+  Pod pod_template = 2;
+
+  EnvoyContainer envoy_container = 3;
+}
```
lgadban, line 56: `ProxyDeployment` should probably contain annotations and labels as well?

lgadban (commented):
first pass over just the GatewayConfig API -- looks pretty close to me already, great job!

projects/gateway2/api/v1alpha1/serviceaccount.proto
```diff
@@ -0,0 +1,16 @@
...
+option go_package = "github.com/solo-io/gloo/projects/gateway2/pkg/api/gateway.gloo.solo.io/v1alpha1";
+
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
+
+import "google/protobuf/wrappers.proto";
+
+message ServiceAccount {
+  // Whether to create a ServiceAccount. Defaults to true.
```
sam-heilbron, line 12: I'm still a bit confused why a user would disable this. Is it just so they can create their own service account and not rely on a default gloo definition? If so, I'd be curious what they'd want different in a service account that we don't offer, and if we just shouldn't let users configure it

projects/gateway2/deployer/deployer.go
```diff
@@ -58,7 +58,7 @@ type Inputs struct {
 }
 
 // NewDeployer creates a new gateway deployer
-func NewDeployer(scheme *runtime.Scheme, inputs *Inputs) (*Deployer, error) {
+func NewDeployer(scheme *runtime.Scheme, cli client.Client, inputs *Inputs) (*Deployer, error) {
```
sam-heilbron, line 61: Given that scheme comes from client, should we just pass a client to the constructor?

projects/gateway2/deployer/deployer.go
```diff
@@ -85,7 +86,17 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
 		},
 	}
 
-	objs, err := d.renderChartToObjects(ctx, fakeGw)
+	// these are the minimal values that render the Deployment, Service, ServiceAccount, and ConfigMap
+	vals := map[string]any{
```
sam-heilbron, line 90: Given that we define a proto for this API, could we use the generated go struct to build this value?

projects/gateway2/deployer/deployer.go
```diff
@@ -131,41 +143,41 @@ func (d *Deployer) renderChartToObjects(ctx context.Context, gw *api.Gateway) ([
...
+				},
+				"xds": map[string]any{
+					// The xds host/port MUST map to the Service definition for the Control Plane
+					// This is the socket address that the Proxy will connect to on startup, to receive xds updates
+					//
+					// NOTE: The current implementation in flawed in multiple ways:
+					//	1 - This assumes that the Control Plane is installed in `gloo-system`
+					//	2 - The port is the bindAddress of the Go server, but there is not a strong guarantee that that port
+					//		will always be what is exposed by the Kubernetes Service.
+					"host": fmt.Sprintf("gloo.%s.svc.%s", defaults.GlooSystem, "cluster.local"),
```
sam-heilbron, line 168: I was thinking more about this, and shouldn't be we able to use POD_NAMESPACE to determine where the control plane is running? (instead of relying on the hardcoded gloo-system)

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+  }
+  // Initial envoy configuration.
+  EnvoyBootstrap bootstrap = 2;
+}
+
+// Configuration for the set of Kubernetes resources that will be provisioned
+// for a given Gateway.
+message KubernetesProxyConfig {
+  // The workload type of the proxy
+  oneof workload_type {
```
sam-heilbron, line 40: What are the other workload types that you see this growing into?

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/bootstrap.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/container.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/pod.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/service.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/serviceaccount.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+
+import "google/protobuf/wrappers.proto";
+
+// A GatewayConfig contains configuration that is used to dynamically provision
```
sam-heilbron, line 19: How does a GatewayConfig get "attached" to a Kubernetes Gateway?

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Configuration for the proxy instance that implements a given Gateway.
+message ProxyConfig {
+  // The type of environment/platform in which the proxy will be provisioned.
+  oneof environment_type {
+    KubernetesProxyConfig kube = 1;
+  }
+  // Initial envoy configuration.
+  EnvoyBootstrap bootstrap = 2;
```
sam-heilbron, line 33: Does this need to be user facing? Or is there some API that is internal-only (known only to the control plane), and users don't need to be aware of it

projects/gateway2/api/v1alpha1/service.proto
```diff
@@ -0,0 +1,12 @@
...
+package gateway.gloo.solo.io;
+option go_package = "github.com/solo-io/gloo/projects/gateway2/pkg/api/gateway.gloo.solo.io/v1alpha1";
+
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
+
+message Service {
+  // Defaults to `ClusterIP`.
+  string type = 1;
```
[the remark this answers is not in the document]
jenshu: yeah this should probably be an enum.
the k8s proto has it as a string, but the go type from which the proto was generated uses an enum

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: i can try removing for now - wasn't sure if skv2 or something else would blow up if there wasn't a status message

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: yes can move to enum

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: yeah i was leaving open the option to add more config in the future.
hmmm, [Gloo][Gateway]DataPlaneConfig?
gateway.gloo.solo.io is the pkg name so maybe DataPlaneConfig is sufficient?

projects/gateway2/deployer/deployer.go
jenshu: the code changes are still WIP but im using json marshal/unmarshal

projects/gateway2/deployer/deployer.go
jenshu: yep! that is where i was planning to get it from

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: right now DaemonSet is the main one (GE supports that too)

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: a Gateway contains a gatewayClassName (e.g. `gloo-gateway`) pointing to an instance of a GatewayClass, and a GatewayClass has a parametersRef pointing to an instance of a GatewayConfig, e.g.
```
kind: GatewayClass
metadata:
  name: gloo-gateway
spec:
  controllerName: solo.io/gloo-gateway
  parametersRef:
    group: gateway.gloo.solo.io
    kind: GatewayConfig
    name: my-gateway-config
    namespace: gloo-system
```

this is for gatewayclass-specific customization; adding gateway-specific config will be done in another (TBD) way later

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: i would say it should be user facing, in case user wants to override the xds server address/port (we allow that in GE). do you think there's a reason to not make it configurable? if we are not sure for now, can also remove it for now and expose it later

projects/gateway2/api/v1alpha1/serviceaccount.proto
jenshu: yeah they could provide their own serviceaccount, or just use `default`. if we don't think this adds value, can definitely remove it. lawrence had a similar comment above

projects/gateway2/deployer/deployer.go
jenshu: yeah good call, i think we can do that

projects/gateway2/api/v1alpha1/container.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+package gateway.gloo.solo.io;
+option go_package = "github.com/solo-io/gloo/projects/gateway2/pkg/api/gateway.gloo.solo.io/v1alpha1";
+
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
```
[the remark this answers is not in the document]
jenshu: it seems to be standard in all other GE protos. i'm not sure if we need them though

projects/gateway2/api/v1alpha1/container.proto
```diff
@@ -0,0 +1,21 @@
...
+
+message Image {
+  string registry = 1;
```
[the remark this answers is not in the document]
jenshu: yep! that's one of the TODOs called out above that i still need to do

install/helm/gloo/templates/44-rbac.yaml
```diff
@@ -66,6 +66,11 @@ metadata:
     {{- include "gloo-gateway.gateway.constLabels" . | nindent 4 }}
     {{- include "gloo-gateway.controlPlane.labels" . | nindent 4 }}
 rules:
+- apiGroups:
```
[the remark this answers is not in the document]
jenshu: this whole file is gated by `gateway2.controlPlane.enabled`

projects/gateway2/api/v1alpha1/pod.proto
jenshu: maybe, i need to think about it. currently we hardcode the same labels (via helm helper functions) in a bunch of places, so if we made it configurable we would need to ensure it gets propagated everywhere to match

projects/gateway2/api/v1alpha1/gateway_config.proto
lgadban: Workload is overloaded in Istio terminology, I would prefer to avoid it.

The reason I suggested DataPlaneConfig is because eventually we may need to spin up an ExtAuth/RateLimit deployment alongside the Proxy (e.g. datalplanePerProxy usagecase)
I can see DataPalaneConfig containing:
- ProxyConfig
- ExtAuthServerConfig
- RateLimitServerConfig

GatewayConfig can easily be confused for control plane config — which actually may make sense if we want this API to eventually be able to control control plane config AND proxy config

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+// Configuration for the set of Kubernetes resources that will be provisioned
+// for a given Gateway.
+message KubernetesProxyConfig {
+  // The workload type of the proxy
+  oneof workload_type {
+    ProxyDeployment deployment = 1;
+  }
+
+  Service service = 2;
+  ServiceAccount service_account = 3;
```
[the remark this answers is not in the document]
jenshu: in GE you can specify annotations and whether to automount the SA token

projects/gateway2/api/v1alpha1/gateway_config.proto
sam-heilbron: Yeah I there are a few parts to this:
1. What users can do today in Gloo Gateway
2. What users should do today in Gloo Gateway
3. What this means for this CR

1. Today, you're right, users have access to control not only the pod port of the xds server, but also the service port
2. I think Lawrence has raised this in the past, but realistically we should be able to hardcode the service port, and if users want to customize the pod port, they can do that
3. Regardless of what users can/can't configure, I think they'd want to control that in one place. What I'm proposing is that instead of the user needing to specify on this CR, the address where Gloo is running, the deployer should be able to determine what port it is accessible on (via the Kubernetes Service), and then configure proxy workloads to talk back to it.

projects/gateway2/api/v1alpha1/gateway_config.proto
sam-heilbron: I see. Since we don't yet support it truly (https://github.com/solo-io/solo-projects/issues/5629) I think we're still a bit away from needing to configure it here. However, it's a good idea and perhaps worth the though experiment to see what the configuration would look like. It might help us determine how much of those two APIs would need to be shared or separate

projects/gateway2/api/v1alpha1/gateway_config.proto
```diff
@@ -0,0 +1,75 @@
...
+  EnvoyContainer envoy_container = 3;
+}
+
+message EnvoyContainer {
+  Image image = 1;
+
+  // Logging configuration.
+  EnvoyLogging logging = 2;
+
+  k8s.io.api.core.v1.SecurityContext security_context = 3;
```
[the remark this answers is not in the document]
sam-heilbron: Sorry yeah it was less focused on the logging. More I was trying to get at the fact that we have a SecurityContext field (part of the k8s api) on a proto message, that is called ProxyDeployment. It's not that critical, more wanted to point out that this "ProxyDeployment" as it's written now, will not be able to be reused in non-k8s environments, but I think that's ok. So you can ignore me/resolve this convo

projects/gateway2/api/v1alpha1/container.proto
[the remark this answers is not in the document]
sam-heilbron: True

projects/gateway2/api/v1alpha1/pod.proto
jenshu: the deployer essentially "owns" the deployment it created, so it would overwrite any changes you manually make

projects/gateway2/api/v1alpha1/gateway_config.proto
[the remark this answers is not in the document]
jenshu: yep, the "Deployment" and "Container" messages are k8s-specific and only used in k8s environments

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: resolution: use DataPlaneConfig

projects/gateway2/api/v1alpha1/pod.proto
jenshu: resolution: will we expose extra_labels / extra_annotations on only the Pod template and Service

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: resolution: will we expose extra_labels / extra_annotations on only the Pod template and Service

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: yep it's the reason i kept `Pod` as a reusable message; both deployment and daemonset have a podspec, and where they differ is the toplevel fields e.g. replicas for deployment

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: as discussed, removed configurability of xds host/port ac987b1048b0f4562d8c91b4b8273d4b66bba544

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: ac987b1048b0f4562d8c91b4b8273d4b66bba544

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: b3cbe5552d8d7e50a5fb2f38e7aac1d1457c4c68

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: as discussed, kept as string and also added component_log_level ac987b1048b0f4562d8c91b4b8273d4b66bba544

projects/gateway2/api/v1alpha1/gateway_config.proto
jenshu: as discussed, added list of Conditions (currently unimplemented) ac987b1048b0f4562d8c91b4b8273d4b66bba544

projects/gateway2/api/v1alpha1/gateway_config.proto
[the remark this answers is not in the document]
jenshu: moved k8s-specific messages to a separate pkg and added more proto comments ac987b1048b0f4562d8c91b4b8273d4b66bba544

projects/gateway2/api/v1alpha1/serviceaccount.proto
jenshu: removed serviceaccount 43146a131de9dbbd71ff59520cfed3dad34b1216

projects/gateway2/api/v1alpha1/container.proto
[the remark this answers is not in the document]
jenshu: i commented most/all the fields ac987b1048b0f4562d8c91b4b8273d4b66bba544
still need to do a bit of cleanup + document the default values

projects/gateway2/deployer/deployer.go
jenshu: 0b165314e7c66b8f3304f3d0c8c439174ecae349

Makefile
```diff
@@ -288,7 +288,7 @@ generate-all: generated-code
 # Generates all required code, cleaning and formatting as well; this target is executed in CI
 .PHONY: generated-code
 generated-code: check-go-version clean-solo-kit-gen ## Run all codegen and formatting as required by CI
-generated-code: go-generate-all generate-cli-docs getter-check mod-tidy
+generated-code: go-generate-all mod-tidy generate-cli-docs getter-check mod-tidy
```
sam-heilbron, line 291: We perform a `mod-tidy` at the end of this recipe. Is there a reason that is no longer sufficient?

generate.go
```diff
@@ -11,10 +11,15 @@ import (
...
 func main() {
-	log.Printf("starting generate")
+	log.Printf("starting generate for gloo")
 
 	protoImports := sk_anyvendor.CreateDefaultMatchOptions(
-		[]string{"projects/**/*.proto", sk_anyvendor.SoloKitMatchPattern},
+		[]string{
+			"projects/gloo/**/*.proto",
+			"projects/gateway/**/*.proto",
+			"projects/ingress/**/*.proto",
```
sam-heilbron, line 20: I like this change as it is more explicit, than just saying all protos in all folders. out of curiosity, did this fix a functional or performance issue?

install/helm/gloo/crds/gateway.gloo.solo.io_crds.yaml
```diff
@@ -0,0 +1,1477 @@
...
+                  - kube
+                properties:
+                  kube:
+                    description: The proxy will be deployed on Kubernetes.
+                    properties:
+                      autoscaling:
+                        description: Autoscaling configuration.
+                        properties:
+                          horizontalPodAutoscaler:
+                            description: |-
```
sam-heilbron, line 67: In the Gloo CRDs, https://github.com/solo-io/gloo/blob/main/generate.go#L53, we remove the descriptions because they aren't functional, and lead to larger files. Are there any toggles that exist in skv2 to perform something similar? (I don't think we need to build it if it doesn't exist, just curious what config does exist)

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
+    // Use a Kubernetes deployment as the proxy workload type.
+    ProxyDeployment deployment = 1;
+  }
+
+  // Configuration for the Kubernetes Service that exposes the Envoy proxy over
+  // the network.
+  kube.gateway.gloo.solo.io.Service service = 2;
+
+  // Autoscaling configuration.
+  kube.gateway.gloo.solo.io.Autoscaling autoscaling = 3;
```
sam-heilbron, line 49: Autoscaling seems like it is tightly coupled with the deployment configuration. In fact, it doesn't work with a DaemonSet:

```
Horizontal pod autoscaling does not apply to objects that can't be scaled (for example: a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset).)
```

Should this live more closely to the ProxyDeployment?

projects/gateway2/api/v1alpha1/kube/service.proto
```diff
@@ -0,0 +1,45 @@
...
+  // and
+  // https://kubernetes.io/docs/concepts/services-networking/service/#headless-services
+  // on the implications of setting `clusterIP`.
+  string clusterIP = 2;
+
+  // Additional labels to add to the Service object metadata.
+  map<string, string> extra_labels = 3;
+
+  // Additional annotations to add to the Service object metadata.
+  map<string, string> extra_annotations = 4;
```
sam-heilbron, line 26: We expose extra labels and annotations on the Service and the Pod template. Is there a reason you can see those two drifting?

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
+
+  // The compute resources required by this container. See
+  // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+  // for details.
+  kube.gateway.gloo.solo.io.ResourceRequirements resources = 5;
+}
+
+message DataPlaneConfigStatus {
+  // The current conditions of the DataPlaneConfig. This is not currently
+  // implemented.
```
sam-heilbron, line 97: Given that this is not implemented, do you think we should just leave this empty for now, and lave the proposed API commented out, so it's clear what the conditions will be?

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
+  // passthrough value for the `--component-log-level` argument to Envoy, e.g.
+  // `upstream:debug,connection:trace`. See
+  // https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/run-envoy#debugging-envoy
+  // for more information.
+  string component_log_level = 3;
+
+  // The security context for this container. See
+  // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#securitycontext-v1-core
+  // for details.
+  k8s.io.api.core.v1.SecurityContext security_context = 4;
```
sam-heilbron, line 87: What is the purpose of exposing both the SecurityContext on the EnvoyContainer, and the SecurityContext on the Pod?

Do we support the ability to inject other containers in this pod?

sam-heilbron (commented):
Looking good!

I think we want to keep making progress on the configurability of the deployer, while still iterating on the API. 

My recommendation would be that we introduce the minimal API for the DataPlaneConfig, one that just introduce things like envoy log level and image. Then we can start mapping out how changes to that CR is picked up and affects the system as a whole, we can write integration tests, and we can start using this in Enterprise to provision enterprise workloads.

We can then keep iterating on the rest of the API.

@nrjpoddar Could you review this as well as there are a lot of user facing changes being introduced

projects/gateway2/deployer/deployer.go
jenshu: defined go structs to make it less error prone a4922f800700cc2511e89d7f273e96f055fb5645

generate.go
jenshu: i'll add a comment here, but essentially i had to exclude the gateway2 folder from being built, because some of the imported protos from k8s were causing compilation failures in solo-kit (e.g. it complained about using the name `*Entry` on a non-map struct). also, since we are not using the gateway2 protos in classic GE, there was no need to build them here

Makefile
jenshu: it was failing codegen in the middle without this. let me run again and paste the error

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: i'm ok with that
@lgadban do you have opinions on this?

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: they are setting pod-level vs container-level security contexts

and yes we will likely need to add config for the istio containers

projects/gateway2/api/v1alpha1/data_plane_config.proto
nrjpoddar, line 49: Do we need to provide Autoscaling from the get go? I was hoping that we can start small and see what all settings are required and add incrementally.

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
+// Configuration for the Proxy deployment in Kubernetes.
+message ProxyDeployment {
+  // The number of desired pods. Defaults to 1.
+  google.protobuf.UInt32Value replicas = 1;
+
+  // Configuration for the pods created by the deployment.
+  kube.gateway.gloo.solo.io.Pod pod_template = 2;
+
+  // Configuration for the container running Envoy.
+  EnvoyContainer envoy_container = 3;
```
nrjpoddar, line 61: Even if a daemonset is used as `workload_type` I am assuming that these fields would need to reused?

nrjpoddar (commented):
I'm a bit worried about teasing out fields one by one from deployment/pod/container/service spec and creating our own proto structs. Not sure how scalable this will be. cc @yuval-k @EItanya can you take a look too?

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: this was previously exposed in the GG chart (when it was a standalone chart) so i wanted to maintain the configurability

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: yes the pod_template and envoy_container would be reused. i could move those 2 fields up one level to the KubernetesProxyConfig, and then ProxyDeployment will only have `replicas` - does that work?

projects/gateway2/api/v1alpha1/kube/service.proto
jenshu: in the GE classic chart, we allow 2 separate values to control annotations on service vs pod.
i don't know specific customer use cases but i don't think we should be too prescriptive - they may want different labels/annotations on each type of resource, but if they want the same ones, they can set the same ones in both places

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: i saw that there are [multiple types of autoscaling](https://github.com/kubernetes/autoscaler/tree/master) so defined a top-level autoscaling (for now only HPA is supported). HPA doesn't apply to daemonsets but i'm not sure about the other types of autoscaling. however if we want to simplify we can remove the general 'autoscaling' message in favor of a more specific HPA message that lives alongside deployment. do you have a preference?

install/helm/gloo/crds/gateway.gloo.solo.io_crds.yaml
jenshu: yes we can remove the descriptions in skv2. if they aren't there, where would be the best place for the user to see descriptions of the fields?

Makefile
jenshu: it seems to be working now so removed it 78dcdadb49b944ac2472cfb22c82b0014340834b

generate.go
jenshu: added comment 76a58f6076b19d6aecc744099a188523d74455fa

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: i think that's better; i moved them e3b2491

projects/gateway2/api/v1alpha1/data_plane_config.proto
lgadban: removing the status for now makes sense

projects/gateway2/deployer/deployer.go
```diff
@@ -96,86 +108,149 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
...
+		return nil, nil
+	}
+
+	if string(paramsRef.Group) != v1alpha1.DataPlaneConfigGVK.Group ||
+		string(paramsRef.Kind) != v1alpha1.DataPlaneConfigGVK.Kind {
+		return nil, UnsupportedParametersRefKind(string(gwClassName), paramsRef)
+	}
+
+	dpc := &v1alpha1.DataPlaneConfig{}
+	err = d.cli.Get(ctx, client.ObjectKey{Namespace: string(*paramsRef.Namespace), Name: paramsRef.Name}, dpc)
```
yuval-k, line 171: We need to consider where we attach the DataplaneConfig. Things like replicas=1 (and probably some other helm values) should be attached to a specific gateway, rather than the gateway class.

lgadban: the plan, at least initially, was to only support GatewayClass-based config then follow up a per-Gateway config.
I just summarized and added links here: https://soloio.slab.com/posts/g-gv-2-dynamic-proxy-configration-afijdvkg#hk0gh-decisions-per-this-thread

lgadban: If we really want, we can go outside of the spec, and do something like `targetRef` attachment of DataPlaneConfig.
See also: https://soloio.slab.com/posts/g-gv-2-dynamic-proxy-configration-afijdvkg#thread-pszok8tb

danehans: +1 on using GC paramRefs for gateway config. If needed by customers in the future, we could support the [infrastructure](https://github.com/kubernetes-sigs/gateway-api/blob/v1.0.0/apis/v1/gateway_types.go#L224-L230) field to provide Gateway-level config overrides.

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
+  kube.gateway.gloo.solo.io.Service service = 4;
+
+  // Autoscaling configuration.
+  kube.gateway.gloo.solo.io.Autoscaling autoscaling = 5;
+}
+
+// Configuration for the Proxy deployment in Kubernetes.
+message ProxyDeployment {
+  // The number of desired pods. Defaults to 1.
+  google.protobuf.UInt32Value replicas = 1;
```
danehans, line 61: Does gloo use [protovalidate](https://github.com/bufbuild/protovalidate) or any other tools for API validation? If not, it's a good idea to create an issue for this topic and link here. It's important that we provide API validation to ensure clients adhere to predefined schemas and rules.

projects/gateway2/api/v1alpha1/kube/pod.proto
```diff
@@ -0,0 +1,44 @@
...
+
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+
+// Configuration for a Kubernetes Pod template.
+message Pod {
+  // Additional labels to add to the Pod object metadata.
+  map<string, string> extra_labels = 1;
+
+  // Additional annotations to add to the Pod object metadata.
+  map<string, string> extra_annotations = 2;
```
danehans, line 17: Note that this could be the source of conflicts if we decide to use `gateway.spec.infrastructure` for gateway-level config overrides in the future.

projects/gateway2/deployer/deployer.go
```diff
@@ -96,86 +108,149 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
...
+	// Get the DataPlaneConfig from the GatewayClass
+	paramsRef := gwc.Spec.ParametersRef
+	if paramsRef == nil {
+		// there is no custom data plane config (just use default values)
+		logger.V(1).Info("no parametersRef found for GatewayClass", "GatewayClass", gwClassName)
+		return nil, nil
+	}
+
+	if string(paramsRef.Group) != v1alpha1.DataPlaneConfigGVK.Group ||
+		string(paramsRef.Kind) != v1alpha1.DataPlaneConfigGVK.Kind {
```
danehans, line 166: The `namespace` field must be specified since DataPlaneConfig is namespace-scoped. Furthermore, if we expect the namespace value to be set to a specific namespace, that should be checked too.

danehans (commented):
In general, the APIs introduced in this PR have a large surface. I prefer creating APIs with a small surface and iterating over time based on user feedback. From my understanding, we are porting over existing API types but is it necessary to immediately support all the functionality for the Gateway API equivalent? This could be a good opportunity to not support functionality that exists in gloo APIs. Is this to provide feature parity between the two sets of APIs?

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,106 @@
...
+message DataPlaneConfigSpec {
+  ProxyConfig proxy_config = 1;
+}
+
+// Configuration for the proxy instance that implements a given Gateway.
+message ProxyConfig {
+  // The type of environment/platform in which the proxy will be provisioned.
+  oneof environment_type {
+    // The proxy will be deployed on Kubernetes.
+    KubernetesProxyConfig kube = 1;
```
yuval-k, line 31: The ProxyConfig states how gloo should deploy the data plane; i don't see us doing that in envs that are not k8s.. so might be worth flatting it (i.e. not having kube)

yuval-k (commented):
we discussed using parameterRef on the gateway when that's added to the spec; and an annotation in the mean time.

howardjohn:
> we discussed using parameterRef on the gateway when that's added to the spec; and an annotation in the mean time.

https://github.com/kubernetes-sigs/gateway-api/pull/2924 (sent this PR before I saw this :slightly_smiling_face: )

projects/gateway2/api/v1alpha1/data_plane_config.proto
```diff
@@ -0,0 +1,99 @@
...
```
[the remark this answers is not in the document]
jenshu: extracted the envoy log levels into a separate EnvoyBootstrap message

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: do you think this issue adequately covers this? https://github.com/solo-io/gloo/issues/8580

projects/gateway2/deployer/deployer.go
jenshu: updated to use Gateway annotation with the name of the GatewayParameters object (must be in same namespace)

projects/gateway2/api/v1alpha1/data_plane_config.proto
jenshu: per discussions, decided to keep `kube`

jenshu:
letting ci run

projects/gateway2/deployer/values.go
```diff
@@ -41,13 +66,34 @@ type helmImage struct {
...
+	MaxReplicas                       *uint32 `json:"maxReplicas,omitempty"`
+	TargetCPUUtilizationPercentage    *uint32 `json:"targetCPUUtilizationPercentage,omitempty"`
+	TargetMemoryUtilizationPercentage *uint32 `json:"targetMemoryUtilizationPercentage,omitempty"`
+}
+
 type helmIstioSds struct {
 	Enabled *bool `json:"enabled,omitempty"`
 }
+
+type helmServiceAccount struct {
```
jbohanon, line 95: It doesn't look like this is implemented yet. Is that intended to be part of this PR or a TODO? Or did I miss it somewhere?

jenshu: so the structs in this file are intended to represent the helm chart that the deployer internally uses; some of the values (e.g. service account, xds host/port) are not configurable by the user. the only user-facing fields are in GatewayParameters

jbohanon: Yeah I don't see in the deployer where we set service account config except in a map in `GetGvksToWatch`, but I didn't think that was actual config

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/kube/service.proto";
+
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/apimachinery/pkg/apis/meta/v1/generated.proto";
+
+import "google/protobuf/wrappers.proto";
+
+// A GatewayParameters contains configuration that is used to dynamically
+// provision Gloo Gateway's data plane (Envoy proxy instance), based on a
+// Kubernetes Gateway or GatewayClass.
```
ilackarms, line 21: it would be just based on a Gateway, right? Or is this object designed to be used as parameters for both Gateway and GatewayClass?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+message GatewayParametersSpec {
+  ProxyConfig proxy_config = 1;
```
ilackarms, line 23: do we need the extra level of nesting here? could we just put `kube` at top level?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jenshu: ah yeah i can update this to only mention Gateway

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jenshu: per discussions, want to leave it flexible enough to add other data plane config at this level later (e.g. extauthConfig, ratelimitConfig)

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jenshu: caa26a5

projects/gateway2/extensions/extensions.go
```diff
@@ -38,3 +42,18 @@ func (e *k8sGatewayExtensions) CreatePluginRegistry(_ context.Context) registry.
 	plugins := registry.BuildPlugins(queries, e.mgr.GetClient())
 	return registry.NewPluginRegistry(plugins)
 }
+
+// GetEnvoyImage returns the image repo and tag to use for the envoy container image
+// in the proxy deployment.
+func (e *k8sGatewayExtensions) GetEnvoyImage() Image {
```
sam-heilbron, line 48: Given that we are expanding the implemtation of this interface, we can create corresponding tests to match>

sam-heilbron (commented):
Looking good!

I understand we're in the process of building out true end to end tests and that will not be handled in this PR. I think what could still be valuable is an explicit list of behaviors that we expect this component to have, and ideally the steps that users can take to verify that it does.
Could this also include a demonstration for how a user would configure a proxy and our istio integration?

install/helm/gloo/crds/gateway.gloo.solo.io_crds.yaml
```diff
@@ -0,0 +1,613 @@
+# Code generated by skv2. DO NOT EDIT.
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    crd.solo.io/specHash: 2c6d037668ce167d
+  labels:
+    app: gateway2
+    app.kubernetes.io/name: gateway2
```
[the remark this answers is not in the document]
jenshu: ah yeah that must have come from the AppName in the skv2 command. 

looking at some of our other CRDs (excluding GE classic since they don't have these labels): we use `app: gloo-fed`, `app: skv2`, `app: gloo-mesh-apis`, they are all somewhat module-specific. what do you think of `app: gloo-gateway`?

projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto
```diff
@@ -0,0 +1,6191 @@
+/*
```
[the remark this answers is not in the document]
jenshu: yep! all the `generated.proto` are imported because the gateway parameters proto uses them. this is so that we can reuse already-defined messages instead of redefining them in our protos. in some messages there are many levels of nesting (e.g. security context) and it would be tedious to maintain our own copy of those messages

projects/gateway2/api/external/kubernetes/apimachinery/pkg/runtime/generated.proto
```diff
@@ -0,0 +1,136 @@
+/*
```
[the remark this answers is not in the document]
jenshu: yeah the comments at the top of the files are easy to miss. i can add a separate README

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/kube/autoscaling.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/kube/container.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/kube/pod.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1/kube/service.proto";
+
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/api/core/v1/generated.proto";
+import "github.com/solo-io/gloo/projects/gateway2/api/external/kubernetes/apimachinery/pkg/apis/meta/v1/generated.proto";
+
+import "google/protobuf/wrappers.proto";
```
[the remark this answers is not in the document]
jenshu: i can add that to the README

projects/gloo/constants/gloo_gateway.go
```diff
@@ -2,11 +2,4 @@ package constants
 
 const (
 	GlooGatewayEnableK8sGwControllerEnv = "GG_EXPERIMENTAL_K8S_GW_CONTROLLER"
-
-	// GlooGatewayDeployerImage is an experimental API that allows users to inject the image
-	// that the deployer should provision. The value should be in the form:
-	//	[image_repository:image_tag] --> quay.io/solo-io/gloo-envoy-wrapper:1.17.0-beta1
-	// This API is intended to be short-lived, as the long-term vision is to support a CR to define
-	// configuration for our deployer
-	GlooGatewayDeployerImage = "GG_EXPERIMENTAL_DEPLOYER_IMAGE"
```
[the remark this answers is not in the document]
jenshu: i have an enterprise [PR](https://github.com/solo-io/solo-projects/pull/5973) open that shows the injection works, and i think it's safe to remove

install/helm/gloo/crds/gateway.gloo.solo.io_crds.yaml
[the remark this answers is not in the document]
jenshu: renamed to gloo-gateway a359bebebb9c0d7e60ef2b4453fb1267fb5d0239

projects/gateway2/api/external/kubernetes/apimachinery/pkg/runtime/generated.proto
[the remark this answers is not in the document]
jenshu: added readmes 5742d72229c617f940f223b2b292f7c17ae72bdb

projects/gateway2/api/v1alpha1/gateway_parameters.proto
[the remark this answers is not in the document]
jenshu: added readmes 5742d72229c617f940f223b2b292f7c17ae72bdb

projects/gloo/cli/pkg/cmd/install/util.go
```diff
@@ -50,6 +50,8 @@ func init() {
 		"virtualhostoptions.gateway.solo.io",
 		"routeoptions.gateway.solo.io",
 		"graphqlapis.graphql.gloo.solo.io",
+		// gloo gateway CRDs:
```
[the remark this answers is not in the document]
jenshu: updated comment 9b8c468df53d070c73c807ec09ea7e77d9f9eca5

projects/gateway2/controller/controller.go
```diff
@@ -101,6 +104,7 @@ func (c *controllerBuilder) watchGw(ctx context.Context) error {
 		Dev:            c.cfg.Dev,
 		IstioValues:    c.cfg.IstioValues,
 		ControlPlane:   c.cfg.ControlPlane,
+		Extensions:     c.cfg.Extensions,
```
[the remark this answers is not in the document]
jenshu: yes good call

projects/gateway2/controller/controller.go
```diff
@@ -35,6 +36,8 @@ type GatewayConfig struct {
 
 	ControlPlane bootstrap.ControlPlane
 	IstioValues  bootstrap.IstioValues
+
+	Extensions extensions.K8sGatewayExtensions
```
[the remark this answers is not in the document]
jbohanon: I think given how foundational the deployer is to the implementation, it is reasonable to assume that more functionality will need to be injected from EE here.

projects/gateway2/deployer/deployer.go
```diff
@@ -141,14 +180,59 @@ func (d *Deployer) getValues(ctx context.Context, gw *api.Gateway) *helmConfig {
 				Host: &d.inputs.ControlPlane.Kube.XdsHost,
 				Port: &d.inputs.ControlPlane.Kube.XdsPort,
 			},
-			Image: getDeployerImageValues(ctx),
+			Image: getDefaultEnvoyImageValues(d.inputs.Extensions.GetEnvoyImage()),
```
jbohanon, line 183: it _shouldn't_ ever happen that the deployer's inputs or Extensions are nil, but this is NPE if they are.

We don't currently check anywhere that we received a non-nil inputs and that it contains a non-nil Extensions.

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+  // The type of environment/platform in which the proxy will be provisioned.
+  oneof environment_type {
+    // The proxy will be deployed on Kubernetes.
+    KubernetesProxyConfig kube = 1;
+  }
+}
+
+// Configuration for the set of Kubernetes resources that will be provisioned
+// for a given Gateway.
+message KubernetesProxyConfig {
```
jbohanon, line 37: Does it make sense to expose service account config here?

projects/gateway2/extensions/extensions.go
```diff
@@ -13,6 +14,9 @@ import (
 type K8sGatewayExtensions interface {
 	// CreatePluginRegistry returns the PluginRegistry
 	CreatePluginRegistry(ctx context.Context) registry.PluginRegistry
+
+	// GetEnvoyImage returns the envoy image and tag used by the proxy deployment.
+	GetEnvoyImage() Image
```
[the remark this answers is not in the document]
jenshu: sounds good, i think we can do that refactor later if/when we determine that the deployer needs more info

projects/gateway2/deployer/deployer.go
```diff
@@ -129,7 +143,32 @@ func (d *Deployer) renderChartToObjects(ctx context.Context, gw *api.Gateway, va
...
+		logger.V(1).Info("no GatewayParameters found for Gateway",
+			"gatewayName", gw.GetName(),
+			"gatewayNamespace", gw.GetNamespace())
+		return nil, nil
+	}
+
+	// the GatewayParameters must live in the same namespace as the Gateway
+	gwpNamespace := gw.GetNamespace()
+	gwp := &v1alpha1.GatewayParameters{}
+	err := d.cli.Get(ctx, client.ObjectKey{Namespace: gwpNamespace, Name: gwpName}, gwp)
```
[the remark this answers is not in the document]
jenshu: the pre-indexing is mostly only useful if you're doing some sort of query (e.g. find all resources with field key=val), but in this case it's a simple lookup by name/namespace, and would likely not benefit from indexing. controller-runtime clients [read from a cache by default](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2#hdr-Clients_and_Caches) so i don't think we need to worry about performance for this particular lookup

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jenshu: we decided against it, at least for the first pass

projects/gateway2/deployer/deployer.go
jenshu: hmm yeah, should we add back that dpanic you had before?

projects/gateway2/extensions/extensions.go
jenshu: we're testing the plumbing through of the image value in the deployer tests in this PR, as well as the e2e tests in https://github.com/solo-io/solo-projects/pull/5973 - do you think that's sufficient?

i couldn't think of any good standalone tests to add for just the extensions, since this is just returning a static value

projects/gateway2/deployer/deployer.go
jbohanon: Added errors and tests for these cases

projects/gateway2/controller/controller.go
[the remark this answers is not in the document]
sam-heilbron: Fair enough! That's why I had proposed https://github.com/solo-io/gloo/pull/9233#discussion_r1552462052, because I figured if/when we need more enterprise-only concepts, it could easily be passed in. Happy to leave this as is

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+  // The envoy container image. See
+  // https://kubernetes.io/docs/concepts/containers/images
+  // for details.
+  //
+  // Default values, which may be overridden individually:
+  //   registry: quay.io/solo-io
+  //   repository: gloo-envoy-wrapper (OSS) / gloo-ee-envoy-wrapper (EE)
+  //   tag: <gloo version> (OSS) / <gloo-ee version> (EE)
+  //   pullPolicy: IfNotPresent
+  kube.gateway.gloo.solo.io.Image image = 2;
```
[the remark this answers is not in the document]
jenshu: TODO: capture the reason for exposing this somewhere (e.g. design doc)

projects/gateway2/extensions/extensions.go
```diff
@@ -38,3 +42,18 @@ func (e *k8sGatewayExtensions) CreatePluginRegistry(_ context.Context) registry.
...
+// in the proxy deployment.
+func (e *k8sGatewayExtensions) GetEnvoyImage() Image {
+	return Image{
+		Repository: "gloo-envoy-wrapper",
+		Tag:        version.Version,
+	}
+}
+
+// Image contains an image repository (e.g. "gloo-envoy-wrapper") and tag (e.g. "1.17.0")
+type Image struct {
```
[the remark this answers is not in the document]
jenshu: TODO: file issue to add fips support (or not) and add tests

howardjohn:
https://github.com/kubernetes-sigs/gateway-api/pull/2924 has merged

sam-heilbron (approved):

generate.go
lgadban: super-nit: I was still a bit confused here even though the comment is really helpful. The README added to gateway2/api cleared it up for me; if you have more changes coming, adding a pointer from this comment to the README might  be much appreciated for a new reader

projects/gateway2/api/README.md
```diff
@@ -0,0 +1,19 @@
+## APIs for Gloo Gateway - Kubernetes Gateway API Integration
```
lgadban, line 1: extremely helpful README, thank you!

projects/gateway2/api/external/kubernetes/README.md
```diff
@@ -0,0 +1,42 @@
+## Kubernetes Protos in Gloo Gateway
```
lgadban, line 1: same as above, this is really useful for a new reviewer, and I suspect a new maintainer. Thank you!

projects/gateway2/api/external/kubernetes/apimachinery/pkg/runtime/generated.proto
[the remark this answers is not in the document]
lgadban: good call-out @sam-heilbron  and great work on the READMEs @jenshu !

projects/gateway2/deployer/deployer.go
```diff
@@ -141,14 +189,59 @@ func (d *Deployer) getValues(ctx context.Context, gw *api.Gateway) *helmConfig {
...
+	gwp, err := d.getGatewayParametersForGateway(ctx, gw)
+	if err != nil {
+		return nil, err
+	}
+	// if there is no GatewayParameters, return the values as is
+	if gwp == nil {
+		return vals, nil
+	}
+
+	kubeProxyConfig := gwp.Spec.GetProxyConfig().GetKube()
```
lgadban, line 209: so for each config field in the GP obj, will we need to manually pull that out as a var here to set as a val?

i.e. this list of config vars has to be manually updated/kept in sync with the GwParam API?

I think it's fine for now but just trying to understand the maintenance cost

projects/gateway2/api/v1alpha1/gateway_parameters.proto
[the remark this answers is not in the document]
lgadban: FWIW we definitely need to expose this. Like mentioned above, the registry is commonly changed.

lgadban (commented):
This looks really good!
Nothing too controversial in my eyes.
Just a couple of questions/comments; will do another quick pass tomorrow morning

projects/gateway2/deployer/deployer.go
jenshu: that's right; added a comment here a02bff6

generate.go
jenshu: added a pointer to readme ccd1280

howardjohn (commented):
Overall LGTM

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -0,0 +1,112 @@
...
+
+  // The compute resources required by this container. See
+  // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+  // for details.
+  kube.gateway.gloo.solo.io.ResourceRequirements resources = 4;
+}
+
+// Configuration for the Envoy proxy instance that is provisioned from a
+// Kubernetes Gateway.
+message EnvoyBootstrap {
```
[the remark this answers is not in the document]
jenshu: if we collapse overall and component log level into one field, we'll need to parse that out.
how about if we keep them separate, but change component_log_level to a `map<string,string>`?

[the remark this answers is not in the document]
jbohanon: I agree that having an overall log_level field is valuable. I find it's usually only specific debugging scenarios that call for component log level

projects/gateway2/api/v1alpha1/gateway_parameters.proto
ilackarms: i wanted @howardjohn feedback here

lgadban: so the idea is that we should NOT allow users to provision the rest of our enterprise data plane (extauth, ratelimit) via `Gateway` ?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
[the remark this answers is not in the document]
jenshu: d37aaee

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jenshu: 2ee32cc

yuval-k (approved):

sam-heilbron (approved):

PR #9237, opened 2024-03-14
Updates for deflaking dataplane per proxy tests

bewebi:
# Description

Changes made as part of the fix for solo-io/solo-projects#5903 which concerns flaking tests in the dataplane-per-proxy suite

## Helm changes
- Change the Helm hook for resource cleanup resources to `pre-delete` from `post-delete`

## Test utils
- Add `AllowInsecure` option to kube2e `CurlOpts`

# Context

Corresponding solo-projects PR is solo-io/solo-projects#5908

### pre-delete webhook

As part of the deflaking of dataplane-per-proxy tests we are setting `gloo.gateway.validationfailurePolicy="Fail"`
Until now, when the validation failure policy was "Fail", Helm uninstall would block forever due to the following sequence:
- `gloo` deployment, which serves the validation webhook, is deleted
- Secrets and/or Upstreams remain
- When Helm attempts to delete these resources, k8s attempts to reach the validation webhook and fails
- Due to the failure policy of "Fail", validation of those resources' deletion fails and the operation is rejected
- Helm uninstall fails, as does manual deletion of the resources or namespace

The resolution for this is to modify/delete the ValidatingWebhookConfiguration resource
This is bad UX and non-obvious to devs and customers in this situation

Therefore we move the resource cleanup job, which removes the ValidatingWebhookConfiguration, and other resources, to a pre-delete hook
This way, validation is disabled completely before deletion of resources is attempted, allowing all resources to be deleted efficiently

### AllowInsecure

At one point in development we were attempting to request the validation webhook directly via `curl` using the `-k` flag, to enable `AllowInsecure`
While this is not currently used in the EE PR, it doesn't hurt to add to the helper

## Interesting decisions
 
By moving the deletion of certain resources on Helm uninstall prior to the deletion of the `gloo` deployment, there is a possibility that these deletions will trigger translation by the Gloo controller
This would lead to invalid and/or unpredictable outputs
Our assumption is that this is irrelevant, as the user is attempting to totally remove Gloo, completely breaking everything, so a temporary breaking change on the way to complete deletion does not constitute unexpected behavior

## Testing steps

To reproduce the issue:
- Helm install Gloo with the following validation values:
```yaml
gateway:
  validation:
    enabled: true
    allowWarnings: false
    alwaysAcceptResources: false
    failurePolicy: "Fail"
```
- Helm uninstall and observe that the uninstall fails
- Manually remove the `ValidatingWebhookConfiguration` resource and re-attempt uninstall
- Observe success

To reproduce fix:
- Check out this branch and build the Helm chart locally (using `make build-test-chart`)
- Helm install the local chart using the same values
- Helm uninstall and observe success

The EE argocd suite has been run against this branch locally and the Tilt integration has also been manually tested with this branch

## Notes for reviewers

See slack conversation [here](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1710795296360169?thread_ts=1710793274.597129&cid=G01EERAK3KJ) for additional context

# Checklist:


jenshu (approved):

bewebi:
bulldozer please 🙏

PR #9241, opened 2024-03-14
initialize k8s gw extensions factory just once, allow returning error

ilackarms:
this is so extension plugins can initialize their own watches and register their own GVKs to the mgr.Scheme a single time during control plane startup


projects/gateway2/controller/start.go
```diff
@@ -85,6 +85,13 @@ func Start(ctx context.Context, cfg StartConfig) error {
 		cfg.GlooPluginRegistryFactory(ctx))
 	var sanz sanitizer.XdsSanitizers
 	inputChannels := xds.NewXdsInputChannels()
+
+	k8sGwExtensions, err := cfg.ExtensionsFactory(mgr)
+	if err != nil {
+		setupLog.Error(err, "unable to create k8s gw extensions")
+		return err
+	}
+
```
inFocus7, line 94: actually, what should we do if - for some reason - the `k8sGwExtensions` is `nil`?

inFocus7 (commented):
single question. not a blocker since we're setting this up and not something a user could break - afaik - but maybe worth knowing what we should do.

projects/gateway2/extensions/extensions.go
```diff
@@ -17,7 +17,7 @@ type K8sGatewayExtensions interface {
 }
 
 // K8sGatewayExtensionsFactory returns an extensions.K8sGatewayExtensions
-type K8sGatewayExtensionsFactory func(mgr controllerruntime.Manager) K8sGatewayExtensions
+type K8sGatewayExtensionsFactory func(mgr controllerruntime.Manager) (K8sGatewayExtensions, error)
```
sam-heilbron, line 20: Given this signature change, I think we should modify our open source implementation `NewK8sGatewayExtensions` to fit this signature as well, and return a nil error

inFocus7: :sounds-good: i pushed up changes 👍🏼

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -439,8 +441,10 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 
 func RunGloo(opts bootstrap.Opts) error {
 	glooExtensions := Extensions{
-		K8sGatewayExtensionsFactory: extensions.NewK8sGatewayExtensions,
-		PluginRegistryFactory:       registry.GetPluginRegistryFactory(opts),
+		K8sGatewayExtensionsFactory: func(mgr ctrl.Manager) (extensions.K8sGatewayExtensions, error) {
```
sam-heilbron, line 444: I don't think change is necessary anymore, right? We can just pass the function because the signature is the same

projects/gloo/pkg/syncer/setup/extensions_test.go
```diff
@@ -21,19 +23,25 @@ var _ = Describe("Extensions", func() {
 			K8sGatewayExtensionsFactory: nil,
 		}, MatchError(ErrNilExtension("K8sGatewayExtensionsFactory"))),
 		Entry("missing PluginRegistryFactory", Extensions{
-			K8sGatewayExtensionsFactory: extensions.NewK8sGatewayExtensions,
-			PluginRegistryFactory:       nil,
+			K8sGatewayExtensionsFactory: func(mgr ctrl.Manager) (extensions.K8sGatewayExtensions, error) {
+				return extensions.NewK8sGatewayExtensions(mgr)
+			},
```
sam-heilbron, line 28: I don't think these changes are necessary either, given that the function signature already matches

sam-heilbron (approved):

inFocus7:
bulldozer wasn't kicking in a few minutes so i manually squashed & merged to continue release process

PR #9242, opened 2024-03-18
syncer: reconcile proxies in memory cache

sam-heilbron:
# Description

Update the xDS syncer of the k8s gateway integration to correctly update Proxy entries

# Context

## Logging
We include a name in the k8s gateway logger, so that you can easily identify where logs lines originated from:
```
{"level":"debug","ts":"2024-03-18T03:35:07.411Z","logger":"k8s-gw-syncer.translator","caller":"translator/translator.go:140","msg":"verifying upstream groups: gw"}
```

Before this change, the place where that name was added, was inside of a function that is called within a loop. The result was that the log line would grow like:
```
{"level":"debug","ts":"2024-03-18T03:35:07.411Z","logger":"k8s-gw-syncer.k8s-gw-syncer.k8s-gw-syncer.k8s-gw-syncer.k8s-gw-syncer.k8s-gw-syncer.translator","caller":"translator/translator.go:140","msg":"verifying upstream groups: gw"}
```

## Proxy Reconciliation
We maintain Proxies that have been generated by the k8s gateway integration, in an in-memory store. This allows the "admin" endpoint to serve up proxies to users `glooctl get proxy` so they can inspect those Proxies at runtime.

If a Proxy with the same name does not exist in the cache, we need to write it. If a Proxy with the same name does exist in the cache, we need to overwrite it. In order to perform both cases, we need to use a ProxyReconciler.

Otherwise, you would see log lines produced by https://github.com/solo-io/gloo/blob/main/projects/gateway2/xds/xds_syncer.go#L385 (just look for `proxyCache`).

## Interesting decisions
 -

## Notes for reviewers
-

# Checklist:

### DIFF
--- a/projects/gloo/cli/pkg/common/proxy.go
+++ b/projects/gloo/cli/pkg/common/proxy.go
@@ -9,6 +9,8 @@ import (
 	"strconv"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/solo-io/gloo/pkg/utils/kubeutils/portforward"
 
 	"github.com/solo-io/solo-kit/pkg/errors"
@@ -80,7 +82,7 @@ func getProxiesFromControlPlane(opts *options.Options, name string, proxyEndpoin
 		// By supplying no filters, we are requesting all the proxies
 		proxyRequest = &debug.ProxyEndpointRequest{
 			Name:      "",
-			Namespace: "",
+			Namespace: metav1.NamespaceAll,
 			Source:    "",
 			Selector:  nil,
 		}
--- a/projects/gloo/pkg/api/v1/proxy_extensions.go
+++ b/projects/gloo/pkg/api/v1/proxy_extensions.go
@@ -30,8 +30,3 @@ type ProxyReader interface {
 	Read(namespace, name string, opts clients.ReadOpts) (*Proxy, error)
 	List(namespace string, opts clients.ListOpts) (ProxyList, error)
 }
-
-// ProxyWriter exposes the subset of methods from a v1.ProxyClient that are write-only
-type ProxyWriter interface {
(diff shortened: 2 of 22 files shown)


inFocus7 (approved):

projects/gloo/pkg/api/v1/proxy_extensions.go
```diff
@@ -23,3 +24,14 @@ func (p *Proxy) UnmarshalStatus(status v1.Status, unmarshaler resources.StatusUn
...
 }
+
+// ProxyReader exposes the subset of methods from a v1.ProxyClient that are read-only
+type ProxyReader interface {
+	Read(namespace, name string, opts clients.ReadOpts) (*Proxy, error)
+	List(namespace string, opts clients.ListOpts) (ProxyList, error)
+}
+
+// ProxyWriter exposes the subset of methods from a v1.ProxyClient that are write-only
+type ProxyWriter interface {
```
jenshu, line 35: is this used?

projects/gloo/cli/pkg/common/proxy.go
```diff
@@ -80,7 +82,7 @@ func getProxiesFromControlPlane(opts *options.Options, name string, proxyEndpoin
 		// By supplying no filters, we are requesting all the proxies
 		proxyRequest = &debug.ProxyEndpointRequest{
 			Name:      "",
-			Namespace: "",
+			Namespace: metav1.NamespaceAll,
```
jenshu, line 85: was `""` not using all namespaces?

sam-heilbron: It was, the constant is just defined as an empty string. I just found using the constant to be a more expressive way of showing that all namespaces should be searched

jenshu (approved):

inFocus7 (approved):
🚀

PR #9250, opened 2024-03-20
logging and verifyGlooValidationWorks

sheidkamp:
# Description
Attempts to fix/diagnose the `gateway Validation Configuration when allowWarnings=false secret validation [It] should act as expected with secret validation`

- Added `verifyGlooValidationWorks` to the beginning of flaky secret delete test to validate that validation is ready before starting the test
- Fixed the artifact dumping for the kube2e tests

## Interesting decisions
Removed the `FlakeAttempts` decorator when adding the logging in order to expose the flake and the new logging in order to diagnose the root cause/permanent fix instead of hiding it with retries.

## Context
Flake example: https://github.com/solo-io/gloo/actions/runs/8321982247/job/22769048982#step:3:329
Earlier flake fix attempt: https://github.com/solo-io/gloo/pull/8795

## Testing steps
by adding `err = nil` at line 2126 of `test/kube2e/gateway/gateway_test.go` and running the gateway test, we can trigger the logging that would occur on a failure.

Validate that the output of `ls -l _output/kube2e-artifacts` looks something like:
```
total 208
-rw-------  1 sethheidkamp  staff    335 Mar 20 15:30 docker-state.log
drwxr-xr-x  6 sethheidkamp  staff    192 Mar 20 15:30 envoy-dump
drwxr-xr-x  9 sethheidkamp  staff    288 Mar 20 15:30 gloo-system
-rw-------  1 sethheidkamp  staff  96535 Mar 20 15:30 kube-state.log
-rw-------  1 sethheidkamp  staff     52 Mar 20 15:30 process-state.log
```

Any method of triggering a failure may be used.

# Checklist:


test/kube2e/gateway/gateway_test.go
```diff
@@ -2113,14 +2113,31 @@ spec:
...
+					verifyGlooValidationWorks()
+
 					By("waiting for the modified VS to be accepted")
 					helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
 						return resourceClientset.VirtualServiceClient().Read(testHelper.InstallNamespace, testServerVs.GetMetadata().GetName(), clients.ReadOpts{Ctx: ctx})
 					})
 
 					By("failing to delete a secret that is in use")
 					err := resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+					if err == nil { // Extra logging to help debug flakes
```
sam-heilbron, line 2126: https://github.com/solo-io/gloo/blob/main/test/helpers/kube_dump.go#L26

We register a PreFailHandler (https://onsi.github.io/ginkgo/) to dump state of the cluster before cleaning it up.

This solution you have works, though perhaps another way to approach this would be to expand what is output in that handler. We could either include it always, or just optional include it, and for this test, we turn on that option

test/kube2e/util.go
```diff
@@ -148,6 +148,13 @@ func UpdateSettings(ctx context.Context, updateSettings func(settings *v1.Settin
 	UpdateSettingsWithPropagationDelay(updateSettings, waitForSettingsToPropagate, ctx, installNamespace)
 }
 
+func GetSettings(ctx context.Context, installNamespace string) *v1.Settings {
+	settingsClient := clienthelpers.MustSettingsClient(ctx)
+	settings, err := settingsClient.Read(installNamespace, "default", clients.ReadOpts{})
+	Expect(err).NotTo(HaveOccurred())
```
sam-heilbron, line 154: nit: This is in a utility function, so we'd likely need an offest

sam-heilbron (dismissed):

test/kube2e/util.go
```diff
@@ -148,6 +148,13 @@ func UpdateSettings(ctx context.Context, updateSettings func(settings *v1.Settin
 	UpdateSettingsWithPropagationDelay(updateSettings, waitForSettingsToPropagate, ctx, installNamespace)
 }
 
+func GetSettings(ctx context.Context, installNamespace string) *v1.Settings {
+	settingsClient := clienthelpers.MustSettingsClient(ctx)
+	settings, err := settingsClient.Read(installNamespace, "default", clients.ReadOpts{})
```
jenshu, line 153: you can also use the const defined in `projects/gloo/pkg/defaults/settings.go`

sheidkamp: updated

test/kube2e/util.go
sheidkamp: updated

test/kube2e/gateway/gateway_test.go
sheidkamp: Looks like the artifact generation was broken. I have made an update that fixes it and removed this logging.

sam-heilbron (approved):
nice work!

jenshu (approved):

sheidkamp:
dozer?

PR #9260, opened 2024-03-21
Expose disable dynamic stats

ashishb-solo:
# Description

Expose an option to disable dynamic stats for benchmarking/profiling purposes.

## API changes
- Added `dynamic_stats` field to the router message

## Docs changes
- Inline protobuf documentation for dynamic stats field

# Context

This request came from a user who was interested in understanding the differences in performance between http1 and http2.

## Testing steps

Not yet implemented

## Notes for reviewers

# Checklist:


projects/gloo/api/v1/options/router/router.proto
```diff
@@ -9,4 +9,7 @@ import "google/protobuf/wrappers.proto";
 message Router{ 
     // Do not add any additional x-envoy- headers to requests or responses. This only affects the router filter generated x-envoy- headers, other Envoy filters and the HTTP connection manager may continue to set x-envoy- headers.
     google.protobuf.BoolValue suppress_envoy_headers = 1;
-}
\ No newline at end of file
+
+    // Disable dynamic stats. Defaults to false. This should only be set to true for profiling or benchmarking scenarios - it is strongly recommended to NOT set this in production
```
ben-taussig-solo, line 13: - I think this should say `enable` dynamic stats, but I am not sure
 - Can we update this to say something like "Set to true to enable dynamic stats" -- I think at the moment it is slightly unclear what will happen if the value is set to true/false

projects/gloo/api/v1/options/router/router.proto
```diff
@@ -9,4 +9,7 @@ import "google/protobuf/wrappers.proto";
...
+    google.protobuf.BoolValue dynamic_stats = 2;
```
ben-taussig-solo, line 14: given that this defaults to false and that we don't seem to be making use of null values, can we just use a bool here instead of a `google.protobuf.BoolValue`?

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -176,13 +177,16 @@ var _ = Describe("Listener Subsystem", func() {
 			},
 		),
 		Entry(
-			"Add suppress envoy headers to the router",
+			"Set dynamic_stats to false and suppress_envoy_headers to true",
```
ben-taussig-solo, line 180: can we decouple the preexisting test around supress_envoy_headers from the new testing around dynamic stats? correct me if I'm wrong but they don't seem to affect each other

projects/gloo/pkg/translator/network_filters.go
```diff
@@ -240,6 +240,8 @@ func (h *hcmNetworkFilterTranslator) computeHttpFilters(params plugins.Params) [
 		routerV3.SuppressEnvoyHeaders = true
 	}
 
+	routerV3.DynamicStats = h.listener.GetOptions().GetRouter().GetDynamicStats()
```
sam-heilbron, line 243: nit: I see that we are following the existing convention of defining the router options in this network_filters file. In general, for any filter that has any amount of complexity, we've tended to define a plugin that holds all the logic. Do you think since we're now adding more fields, it's worth it to introduce a router plugin to hold this logic?

projects/gloo/api/v1/options/router/router.proto
sam-heilbron: I agree with Ben. I find the name of the field and the comment "disable dynamic stats" to be a bit confusing

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -202,7 +206,48 @@ var _ = Describe("Listener Subsystem", func() {
 				}},
 			},
 			func(listener *envoy_config_listener_v3.Listener, routeConfigs []*envoy_config_route_v3.RouteConfiguration) {
-				By("Should be able to add and translate the router to an envoy config")
+				By("configuring the envoy router from gloo settings")
```
sam-heilbron, line 209: nit: I know we're following prior art, so feel free to disregard.

It feels like the listener_subsystem_test isn't the appropriate place for these tests. Per the comment:
```
// These tests validate that the ListenerSubsystemTranslatorFactory produces Translators
// which in turn create Envoy Listeners and RouteConfigurations with expected values
// The tests are non-exhaustive, as we expect each translator to more rigorously test the
// edge cases. Instead, these tests focus on the high level Envoy resources that are created.
```

It seems like these tests either belong in a network_filters_test which proves that router filter is added, or better yet, a plugin that performs the translation and corresponding unit tests to prove it out

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -202,7 +206,48 @@ var _ = Describe("Listener Subsystem", func() {
...
+				hcm := &envoy_http_connection_manager_v3.HttpConnectionManager{}
+				err = translator.ParseTypedConfig(hcmFilter, hcm)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(hcm.HttpFilters).To(HaveLen(2))
+
+				routeFilter := hcm.GetHttpFilters()[1]
+				typedRouterFilter := routerv3.Router{}
+				routeFilter.GetTypedConfig().UnmarshalTo(&typedRouterFilter)
+				Expect(typedRouterFilter.GetDynamicStats().GetValue()).To(BeFalse())
+				Expect(typedRouterFilter.SuppressEnvoyHeaders).To(BeTrue())
```
sam-heilbron, line 224: Let's the use the Getter method for SuppressEnvoyHeaders

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -202,7 +206,48 @@ var _ = Describe("Listener Subsystem", func() {
...
+			},
+		),
+		Entry(
```
sam-heilbron, line 227: nit: It feels like we could simplify these tests into a simple Table driven test that has cases for:
- unset values
- set true
- set false

projects/gloo/api/v1/options/router/router.proto
ashishb-solo: Thanks, I was struggling on how to make it clear. I took another stab at it in 78e943e6fddc54a601c20dd0de3c4227667d092d - could you let me know if you think it's an improvement? Hopefully it's more explicit about what the possible states do (true, false, nil)

projects/gloo/pkg/translator/listener_subsystem_test.go
ashishb-solo: Sounds like a good idea to me. Were you thinking of doing this in a separate table, or is it possible to have nested table tests in ginkgo/gomega?

projects/gloo/pkg/translator/network_filters.go
ashishb-solo: I like the idea and thought the same thing while writing this, but think it might be a better idea to do this some other time due to (a) time constraints and (b) the fact that the complexity of the current solution is still fairly manageable (in my opinion)

I think I'll just add a TODO in this translation area for the time being, since there are more complicated options in the router filter that we may want to implement at some point. Do you think that's fair for now?

projects/gloo/pkg/translator/listener_subsystem_test.go
ashishb-solo: Created a new table test here: https://github.com/solo-io/gloo/pull/9260/files#diff-1f2deb4318fe8a371631759130cc64f355caf466ded0bd46a996d5f71df48e85R65

projects/gloo/pkg/translator/listener_subsystem_test.go
ashishb-solo: This moved around a bit in the test refactor, but you can find the updated version here: https://github.com/solo-io/gloo/pull/9260/files#diff-1f2deb4318fe8a371631759130cc64f355caf466ded0bd46a996d5f71df48e85R151

projects/gloo/pkg/translator/listener_subsystem_test.go
ashishb-solo: It took quite a few commits to get this in, but I've moved these tests out to here: https://github.com/solo-io/gloo/pull/9260/files#diff-1f2deb4318fe8a371631759130cc64f355caf466ded0bd46a996d5f71df48e85R1

projects/gloo/pkg/translator/network_filters.go
ashishb-solo: i put in a TODO here, let me know how you feel: d9283ab417f7803b4d0396765cc803d6889087c8

projects/gloo/pkg/translator/listener_subsystem_test.go
ashishb-solo: broken out tests can be found here: https://github.com/solo-io/gloo/pull/9260/files#diff-1f2deb4318fe8a371631759130cc64f355caf466ded0bd46a996d5f71df48e85R143

projects/gloo/api/v1/options/router/router.proto
ben-taussig-solo: The update looks much clearer, thanks!

projects/gloo/pkg/translator/listener_subsystem_test.go
ben-taussig-solo: those new table entries are very easy to read, thanks for this!

projects/gloo/api/v1/options/router/router.proto
ashishb-solo: I agree that it's a bit silly to have two values that do the same thing. I'm inclined to just leave it as is and stay as close as possible to the Envoy API, but happy to change it if we see a good reason to do so

projects/gloo/pkg/translator/network_filters_test.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		translatorFactory = translator.NewListenerSubsystemTranslatorFactory(pluginRegistry, sslutils.NewSslConfigTranslator())
+	})
+
+	AfterEach(func() {
+		cancel()
+	})
+
+	DescribeTable("GetAggregateListenerTranslators (success)",
```
sam-heilbron, line 65: nit: Given that these tests are in a new file (network_filters) and are specifically testing the behavior of the httpNetworkFilterTranslator, should we change this name, and perhaps change the inputs to call `ComputeNetworkFilters` ?

projects/gloo/pkg/translator/network_filters_test.go
```diff
@@ -0,0 +1,174 @@
...
+			hcm := &envoy_http_connection_manager_v3.HttpConnectionManager{}
+			err = translator.ParseTypedConfig(hcmFilter, hcm)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(hcm.HttpFilters).To(HaveLen(1))
+
+			routeFilter := hcm.GetHttpFilters()[0]
+			typedRouterFilter := routerv3.Router{}
+			routeFilter.GetTypedConfig().UnmarshalTo(&typedRouterFilter)
+			// Perform assertions on generated Envoy router filter
+			assertionHandler(&typedRouterFilter)
```
sam-heilbron, line 140: nit: The solution you have works well. An alternative, is to just pass a GomegaMatcher, and then call:
```
Expect(typedRouterFilter).To(expectedMatcher)
```

An example of that is here: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/syncer/setup/extensions_test.go#L17

Not a requirement, just sharing an alternative way

sam-heilbron (approved):

projects/gloo/pkg/translator/network_filters_test.go
ashishb-solo: Nice catch on the table name. I've fixed that.

Regarding calling `ComputeNetworkFilters`, I would need an instance of `hcmNetworkFilterTranslator` and I don't see any code that directly creates or instantiates that. So I'm inclined to leave it as is given the amount of work involved in doing that (we still have to push a release out based on this, so I'd ideally like to merge this today/early tomorrow).

However, I did simplify the test down a bit by switching from using an AggregateListener to a HttpListener and this had a modest reduction on the size of the table function.

Both of these changes can be found in https://github.com/solo-io/gloo/pull/9260/commits/0ff31e7f157b9422e82b1a52e155025bb5b2ba5e

projects/gloo/pkg/translator/network_filters_test.go
ashishb-solo: Thanks for the heads up. I like the way it is because it allows us to reuse the same architecture to test different fields of the `typedRouterFilter` struct (suppress headers and dynamic stats). I wouldn't be surprised if Gomega provides a way to specify which field the matcher ought to check on, but I would prefer to leave it as is now

projects/gloo/pkg/translator/network_filters_test.go
sam-heilbron: Yeah I think for constructing the hcmNetworkFilterTranstlator, you could use `NewHttpListenerNetworkFilterTranslator`. Not necessary, I think what you have makes sense

projects/gloo/pkg/translator/network_filters_test.go
sam-heilbron: Yeah that's fair. I think the way you could solve this, using a GomegaMatcher, would be something like:
```
Expect(typedRouterFilter).To(matchers.MatchProto(&routerV1.Router{
	SuppressEnvoyHeaders: &wrappers.BoolValue{
		Value: true,
	},
}))
```

Where MatchProto comes from: https://github.com/solo-io/solo-kit/blob/main/test/matchers/proto.go#L40. And allows us to compare top level fields of a proto message.

ben-taussig-solo (approved):

PR #9276, opened 2024-03-27
GG deployer helm cleanup and use correct xds values

jenshu:
Extracting out some of the internal deployer changes from https://github.com/solo-io/gloo/pull/9233 

The deployer now gets the xds host/port from the gloo service, rather than using a potentially incorrect bind address. This is done via the ControlPlane struct that the setup syncer initializes.

The rest of the changes involve mostly code cleanup/refactoring:
- Define go structs for the (non-user-facing) deployer helm values
- Remove some unused deployer helm values
- Change HPA to v2 (v2beta1 is no longer served as of k8s 1.25)

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5949


jenshu:
/kick transformation test https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=83ea239f-a0ba-4eb2-a21b-b26381b9f283

projects/gateway2/controller/controller_suite_test.go
```diff
@@ -112,6 +114,26 @@ var _ = BeforeSuite(func() {
 		},
 	})
 	Expect(err).NotTo(HaveOccurred())
+
+	// the deployer needs the gloo service to exist, in the namespace defined by POD_NAMESPACE
```
sam-heilbron, line 118: I found this comment really helpful!

projects/gateway2/deployer/deployer.go
```diff
@@ -96,86 +91,58 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
...
 
-	// convert to json for helm (otherwise go template fails, as the field names are uppercase)
-	var portsAny []any
-	err := jsonConvert(gwPorts, &portsAny)
+	return objs, nil
+}
+
+func (d *Deployer) getValues(ctx context.Context, gw *api.Gateway) (*helmConfig, error) {
+	xdsHost := GetDefaultXdsHost()
+	xdsPort, err := GetDefaultXdsPort(ctx, d.cli)
```
sam-heilbron, line 122: These xds functions are invoked each time we call `getValues` which is invoked each time we reconcile a Gateway. Should we instead determine these values once, at startup, and then just refer to the xdshost/port as we need it?

projects/gateway2/deployer/deployer.go
```diff
@@ -193,25 +160,36 @@ func (d *Deployer) Render(ctx context.Context, name, ns string, vals map[string]
...
+		return nil, fmt.Errorf("failed to get values to render objects: %w", err)
+	}
+	logger.V(1).Info("got deployer helm values", "values", vals)
+
+	// convert to json for helm (otherwise go template fails, as the field names are uppercase)
+	var convertedVals map[string]any
+	err = jsonConvert(vals, &convertedVals)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert helm values: %w", err)
+	}
```
sam-heilbron, line 184: Should we explore using https://github.com/mitchellh/mapstructure/blob/main/mapstructure.go#L306 or some other library to perform this on our behalf? I figure there are trade-offs between simplicity and performance. Given that these values are quite small, I don't anticipate true performance issues

projects/gateway2/deployer/defaults.go
```diff
@@ -0,0 +1,45 @@
...
+// These values are used by the deployer to render the dynamically-generated proxy resources (deployment, etc)
+
+var (
+	NoXdsPortFoundError = func(portName string, svcNamespace string, svcName string) error {
+		return eris.Errorf("no port with the name %s found in service %s.%s", portName, svcNamespace, svcName)
+	}
+)
+
+// GetDefaultXdsPort gets the xDS port from the gloo Service.
+func GetDefaultXdsPort(ctx context.Context, cli client.Client) (int32, error) {
```
sam-heilbron, line 24: Another thought I had on this. We already have a ControlPlane object with metadata about itself: https://github.com/solo-io/gloo/blob/e9aa9633d927b3d40d41be0170e8275489ee4d67/projects/gloo/pkg/bootstrap/opts.go#L83. What if we performed this operation when we construct it, and populate an optional KubernetesMetadata field, or something of the like. That way, anyone who wants to "use" the control plane, can get information off that object, instead of needing to calculate it itself. It means that our components can be "dumb" because they don't need to know anything about the service name, or namespace, or xds host info

sam-heilbron: The proposal is just to take the existing logic you built into the deployer, and move it to the ControlPlane setup, so the deployer can just use it. It also gets at an earlier comment regarding "lets compute this once"

projects/gateway2/deployer/deployer.go
```diff
@@ -85,7 +71,16 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
 		},
 	}
 
-	objs, err := d.renderChartToObjects(ctx, fakeGw)
+	// these are the minimal values that render the Deployment, Service, ServiceAccount, and ConfigMap
```
sam-heilbron, line 74: Can you help me understand (and perhaps update the comment to explain) why we need to render these additional resources to capture the GVKs to watch? 
What is the expected result of GetGvksToWatch?

Should we update https://github.com/solo-io/gloo/blob/main/projects/gateway2/deployer/deployer_test.go#L123 to be more explicit?

projects/gateway2/deployer/deployer.go
```diff
@@ -193,25 +160,36 @@ func (d *Deployer) Render(ctx context.Context, name, ns string, vals map[string]
...
 
 func (d *Deployer) GetObjsToDeploy(ctx context.Context, gw *api.Gateway) ([]client.Object, error) {
-	objs, err := d.renderChartToObjects(ctx, gw)
+	logger := log.FromContext(ctx)
+
+	vals, err := d.getValues(ctx, gw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get values to render objects: %w", err)
+	}
+	logger.V(1).Info("got deployer helm values", "values", vals)
```
sam-heilbron, line 177: I'm less familiar with the verbosity levels of this logger. How verbose is this log?

projects/gateway2/deployer/deployer_test.go
```diff
@@ -123,35 +131,6 @@ var _ = Describe("Deployer", func() {
 		Expect(gvks).NotTo(BeEmpty())
 	})
 
-	It("rbac should have our gvks", func() {
```
sam-heilbron, line 126: Can you help me understand why this is safe to remove?

projects/gateway2/deployer/deployer_test.go
```diff
@@ -92,6 +99,7 @@ var _ = Describe("Deployer", func() {
 				APIVersion: "gateway.solo.io/v1beta1",
 			},
 			Spec: api.GatewaySpec{
+				GatewayClassName: "gloo-gateway",
```
sam-heilbron, line 102: nit: Let's use https://github.com/solo-io/gloo/blob/e9aa9633d927b3d40d41be0170e8275489ee4d67/projects/gateway2/wellknown/controller.go#L5

sam-heilbron (commented):
Looking good!

projects/gateway2/deployer/deployer.go
jenshu: that is debug https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md

projects/gateway2/deployer/deployer_test.go
jenshu: this is a remnant of when GG had a standalone chart and had separate toggles for enabling control plane and data plane. it's not longer relevant now that the control plane doesn't live in this chart. basically it continued passing because the list of objects was empty

as part of the future cleanup of the GG control plane helm api we can add back a test like this in the helm unit tests (if one doesn't already exist)

projects/gateway2/deployer/defaults.go
jenshu: refactored this in abc11d5 and 243e50a, so that the xds host/port are just retrieved once each time the setup func is run in GE. still need to add some tests

projects/gateway2/deployer/deployer.go
jenshu: see https://github.com/solo-io/gloo/pull/9276#discussion_r1542184982

projects/gateway2/deployer/deployer.go
jenshu: yeah i had looked into that briefly, but without having time to investigate the pros/cons or performance benefits, i think that can be looked at later. and i agree the values would generally be small. and the jsonConvert func is pretty simple and would be easy to swap out later

projects/gateway2/controller/controller_suite_test.go
jenshu: it's gone now!

projects/gateway2/deployer/deployer_test.go
jenshu: i'm going to make the deployer_test changes in [the other PR](https://github.com/solo-io/gloo/pull/9277) since there's a bunch of refactoring going on in this file

projects/gateway2/deployer/deployer.go
jenshu: added comment 8c6085d

i'm going to make the deployer_test changes in [the other PR](https://github.com/solo-io/gloo/pull/9277) since there's a bunch of refactoring going on in this file

projects/gateway2/deployer/deployer.go
```diff
@@ -224,23 +206,17 @@ func (d *Deployer) GetObjsToDeploy(ctx context.Context, gw *api.Gateway) ([]clie
...
-		if err := cli.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner(d.inputs.ControllerName)); err != nil {
+		logger.V(1).Info("deploying object", "kind", obj.GetObjectKind(), "namespace", obj.GetNamespace(), "name", obj.GetName())
+		if err := d.cli.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner(d.inputs.ControllerName)); err != nil {
 			return fmt.Errorf("failed to apply object %s %s: %w", obj.GetObjectKind().GroupVersionKind().String(), obj.GetName(), err)
 		}
 	}
 	return nil
 }
 
-func (d *Deployer) Deploy(ctx context.Context, gw *api.Gateway, cli client.Client) error {
```
jenshu, line 236: not used anywhere

projects/gateway2/deployer/deployer.go
```diff
@@ -318,26 +294,3 @@ func ConvertYAMLToObjects(scheme *runtime.Scheme, yamlData []byte) ([]client.Obj
 
 	return objs, nil
 }
-
-func getDeployerImageValues() map[string]any {
```
jenshu, line 322: moved to values_helpers.go

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -262,6 +246,75 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
 	}
 	watchNamespaces := utils.ProcessWatchNamespaces(settings.GetWatchNamespaces(), writeNamespace)
 
+	consulClient, err := bootstrap_clients.ConsulClientForSettings(ctx, settings)
```
jenshu, line 249: these are moved up from below, without changes.
needed the clients to be initialized before bootstrap opts, and bootstrap opts initialized before ControlPlane creation

jenshu, line 249: consul/vault/secretsources are moved up from below, without changes.
needed the clients to be initialized before bootstrap opts, and bootstrap opts initialized before ControlPlane creation

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -312,9 +376,12 @@ func (s *setupSyncer) Setup(ctx context.Context, kubeCache kube.SharedCache, mem
...
 			callbacks = s.extensions.XdsCallbacks
 		}
-		s.controlPlane = NewControlPlane(ctx, s.makeGrpcServer(ctx), xdsTcpAddress, callbacks, true)
+		s.controlPlane = NewControlPlane(ctx, s.makeGrpcServer(ctx), xdsTcpAddress,
+			bootstrap.KubernetesControlPlaneConfig{XdsHost: xdsHost, XdsPort: xdsPort}, callbacks, true)
 		s.previousXdsServer.cancel = cancel
-		s.previousXdsServer.addr = xdsAddr
+		s.previousXdsServer.addr = xdsBindAddr
+		s.previousXdsHost = xdsHost
+		s.previousXdsPort = xdsPort
```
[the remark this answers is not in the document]
jenshu: yep probably a little cleaner, fixed in 1d72080

projects/gateway2/deployer/values_helpers.go
```diff
@@ -0,0 +1,57 @@
...
+		targetPort := ports.TranslatePort(listenerPort)
+		portName := string(l.Name)
+		protocol := "TCP"
+
+		var port helmPort
+		port.Port = &listenerPort
+		port.TargetPort = &targetPort
+		port.Name = &portName
+		port.Protocol = &protocol
+		gwPorts = append(gwPorts, port)
```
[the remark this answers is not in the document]
jenshu: fixed in 1d72080

projects/gateway2/deployer/deployer.go
```diff
@@ -85,7 +72,25 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
...
+	// update the values here to enable them.
+	vals := map[string]any{
+		"gateway": map[string]any{
+			"serviceAccount": map[string]any{
+				"create": true,
+			},
+		},
+	}
+
+	objs, err := d.renderChartToObjects(ctx, fakeGw, vals)
```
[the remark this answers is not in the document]
jenshu: renamed in 1d72080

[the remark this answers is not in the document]
jenshu: oh let me add a comment too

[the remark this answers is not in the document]
jenshu: added comment cb0684f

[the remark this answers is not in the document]
jbohanon: That is much clearer, thank you so much!

jbohanon (dismissed):
:rocket:

projects/gateway2/deployer/deployer.go
```diff
@@ -193,25 +166,36 @@ func (d *Deployer) Render(ctx context.Context, name, ns string, vals map[string]
...
 	return objs, nil
 }
 
 func (d *Deployer) GetObjsToDeploy(ctx context.Context, gw *api.Gateway) ([]client.Object, error) {
-	objs, err := d.renderChartToObjects(ctx, gw)
+	logger := log.FromContext(ctx)
+
+	vals, err := d.getValues(ctx, gw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get values to render objects: %w", err)
```
npolshakova, line 181: [nit] Can we specify which gateway failed to deploy in the logs? 

```suggestion
		return nil, fmt.Errorf("failed to get values to render gateway %s.%s: %w", gw.Name, gw.Namespace, err)
```

projects/gateway2/deployer/deployer.go
```diff
@@ -193,25 +166,36 @@ func (d *Deployer) Render(ctx context.Context, name, ns string, vals map[string]
...
+	}
+	logger.V(1).Info("got deployer helm values", "values", vals)
```
npolshakova, line 183: [nit] Can we log which gateway here as well? 

```suggestion
	logger.V(1).Info(fmt.Sprintf("got deployer helm values for gateway %s.%s", gw.Name, gw.Namespace), "values", vals)
```

jenshu: added in d6c419d

projects/gateway2/deployer/values_helpers.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if image == "" {
+		// If the env is not defined, return the default
+		return defaultImageValues
+	}
+
+	imageParts := strings.Split(image, ":")
+	if len(imageParts) != 2 {
+		// If the user provided an invalid override, fallback to the default
+		return defaultImageValues
```
npolshakova, line 51: [nit] Do we want to log that the image is invalid and we're reverting to the default here?

projects/gateway2/deployer/deployer.go
jenshu: realized getValues doesn't need to return error anymore 90f3523

projects/gateway2/deployer/values.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// istio values
+	IstioSDS *helmIstioSds `json:"istioSDS,omitempty"`
+
+	// xds values
+	Xds *helmXds `json:"xds,omitempty"`
+}
+
+type helmPort struct {
+	Port       *uint16 `json:"port,omitempty"`
```
npolshakova, line 29: Is port required and the others are optional? It looks like we only match on the port [here](https://github.com/solo-io/gloo/blob/d6c419d34fb6fec3b4453e62901aedc0cad01494/projects/gateway2/deployer/values_helpers.go#L20).

So I couldn't provide only the port name here?

npolshakova: Oh I think I see- `XdsPort` is required and it matches that?

npolshakova (approved):
LGTM!

projects/gateway2/deployer/values.go
jenshu: those are 2 different ports. helmPort comes from the ports on the gateway, which are used to populate the ports exposed on the proxy service and container. i'll add more comments to hopefully clarify

jenshu: 1bde966

projects/gateway2/deployer/values_helpers.go
jenshu: 1bde966

jbohanon (approved):

jenshu:
i'm monitoring the setup syncer flakes (https://github.com/solo-io/gloo/actions/runs/8512608565/job/23314570766?pr=9276) but it passed this time so letting it merge

PR #9281, opened 2024-03-28
add RouteOption policy attachment

lgadban:
- Add RouteOption resource policy attachment to HTTPRoute
- When RouteOptions are referenced as a filter and attached, the filter (i.e. more specific) takes precedence
- Introduces a new QueryEngine specific for RouteOptions
  - Also uses an index for finding attachment via targetRef
  - Currently only considers attachment within the same namespace 

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/5714


projects/gateway/api/v1/external_options.proto
```diff
@@ -247,4 +248,9 @@ message RouteOption {
 
   // Route options. See Route for delegation behavior.
   gloo.solo.io.RouteOptions options = 3;
+  
+  // When using Kubernetes Gateway API mode, targetRef can be used to attach this RouteOption
+  // to a gateway.networking.k8s.io/HTTPRoute object. The `options` specified will then be used
+  // for all `Rules` in the `HTTPRoute`.
+  core.skv2.solo.io.PolicyTargetReference target_ref = 5;
```
ilackarms, line 255: i believe we should go with `target_refs` here unless we want to make the change in a follow up

projects/gateway2/controller/controller.go
```diff
@@ -82,6 +84,13 @@ func (c *controllerBuilder) addIndexes(ctx context.Context) error {
 	})
 }
 
+// TODO: move to RtOpt plugin when breaking the logic to RouteOption-specific controller
```
ilackarms, line 87: 👍

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -10,22 +10,28 @@ import (
 )
 
 // Plugin is an empty type for base plugins, currently no base methods.
-type Plugin interface{}
+type Plugin interface {
+	GetName() string
```
ilackarms, line 14: a comment here would be nice (in particular explaining why it's necessary for every plugin to impement GetName, and that the return value should be unique amongst plugins)

though i wonder if the return value should be unique against plugins would it be better to store these in a map somewhere

ilackarms (changes requested):

projects/gateway2/translator/plugins/plugins.go
ilackarms: no strong feeling on this though but the comment would be nice

projects/gateway/api/v1/external_options.proto
lgadban: yeah this should be considered out of scope for this PR

projects/gateway2/translator/plugins/plugins.go
lgadban: was not intended to be that formal, was simply a way of providing a sane logging flow.
going to remove so it can be addressed holistically later

projects/gateway/api/v1/external_options.proto
jenshu: just a note that if we make this change in a follow-up, it will be a breaking change

jenshu, line 255: should this field go in the VirtualHostOptions in options.proto instead of here?

lgadban: I don't think the intent would be to release without adding any necessary functionality, more so that there was a limited scope for this work and the follow on work that depends on this

lgadban: This is specifically for RouteOptions (notably VHostOptions is a different set of config that is associated with gwapi.Listener/envoy VirtualHost that is coming in the next set of work related to policy)

>or maybe target refs only applies in the standalone case?

Correct, `targetRef` is purely for the case of attaching a policy object, in this case a `RouteOption` resource, to another resource (e.g. HTTPRoute)
So this is indeed only relevant when using a `RouteOption` standalone object.

jenshu: sorry yeah i meant RouteOptions.
ok makes sense!

sam-heilbron:
/kick 

Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=935fe1f3-ada2-4021-917a-39f2fe3b3464
Error:
```
Step #1 - "run-tests": Step #8 - "run-e2e-tests": [38;5;9m[1mSummarizing 1 Failure:[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   [38;5;9m[FAIL][0m [0mStaged Transformation [38;5;243mno auth [0mcan enable enhanced logging [38;5;9m[1m[It] staged logging enabled, individual logging enabled[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   [38;5;243m/workspace/gloo/test/e2e/staged_transformation_test.go:644[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests": 
```

GitHub issue: https://github.com/solo-io/gloo/issues/9292

sam-heilbron (dismissed):

sam-heilbron (dismissed):

projects/gateway2/translator/plugins/routeoptions/query/indexers.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func IterateIndices(f func(client.Object, string, client.IndexerFunc) error) error {
+	return f(&solokubev1.RouteOption{}, RouteOptionTargetField, routeOptionTargetRefIndexer)
+}
+
+func routeOptionTargetRefIndexer(obj client.Object) []string {
+	rtOpt, ok := obj.(*solokubev1.RouteOption)
+	if !ok {
+		panic(fmt.Sprintf("wrong type %T provided to indexer. expected gateway.solo.io.RouteOption", obj))
```
jenshu, line 24: should this log an error instead of panicking?

projects/gateway2/translator/plugins/utils/sort_objects.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NOTE: this sort utility is adapted from GME policy sorting
+// ref: pkg/translator/utils/sort_creation_timestamp.go
+
+// SortByCreationTime accepts a slice of client.Object instances and sorts it by creation timestamp in ascending order.
+// It panics if the argument isn't a slice, or if it is a slice of a type that does not implement client.Object.
+func SortByCreationTime(objs interface{}) {
+	// Validate the argument
+	if err := validate(objs); err != nil {
+		panic(err)
```
jenshu, line 20: should this log an error instead of panicking?

projects/gateway2/translator/plugins/routeoptions/query/indexers.go
lgadban: this is using the pattern in existing query indexers, such as here: https://github.com/solo-io/gloo/blob/main/projects/gateway2/query/indexers.go#L28

notably this function can not return an error, and if we hit this case, the Listing won't work and we are broken.
This would only happen if the indexing is setup incorrectly, which is essentially a developer error.

My opinion is to leave it -- wdyt?

projects/gateway2/translator/plugins/utils/sort_objects.go
lgadban: this also was adapted [from elsewhere](https://github.com/solo-io/gloo-mesh-enterprise/blob/main/pkg/translator/utils/sort_creation_timestamp.go#L21) -- and also is essentially a developer error, i.e. passing an incorrect object to `Sort...(...)`

we could switch this one to a dpanic and pass in a logger. wdyt?

projects/gateway2/translator/plugins/routeoptions/query/indexers.go
jenshu: that sounds reasonable

jenshu:
LGTM, pending @ilackarms review

projects/gateway2/extensions/extensions.go
```diff
@@ -32,9 +31,10 @@ type k8sGatewayExtensions struct {
 
 // CreatePluginRegistry returns the PluginRegistry
 func (e *k8sGatewayExtensions) CreatePluginRegistry(_ context.Context) registry.PluginRegistry {
-	plugins := registry.BuildPlugins(query.NewData(
+	queries := query.NewData(
 		e.mgr.GetClient(),
 		e.mgr.GetScheme(),
-	))
+	)
+	plugins := registry.BuildPlugins(queries, e.mgr.GetClient())
```
ilackarms, line 38: i expect we will want to just pass `client`+`scheme` eventually since the plugins may be initializing their own queries (such as portal) but we can wait til we need it to make that change

projects/gateway2/translator/plugins/routeoptions/query/indexers.go
ilackarms: in general i think we usually prefer dpanics over panics but i agree we should keep consistent, changing it all at once if we decide to do so

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -40,37 +46,106 @@ func (p *plugin) ApplyRoutePlugin(
 	routeCtx *plugins.RouteContext,
 	outputRoute *v1.Route,
 ) error {
-	filter := utils.FindExtensionRefFilter(routeCtx, gk)
-	if filter == nil {
+	var routeOptions *v1.RouteOptions
+	// check for RouteOptions applied to full Route
+	routeOptions = p.handleAttachment(ctx, routeCtx)
```
ilackarms, line 51: nit: `routeOptions := ...`

projects/gateway2/translator/plugins/routeoptions/query/indexers.go
lgadban: the challenge with a dpanic here is we don't have control of the signature to this function, so we would need to use some weirdness to get access to the logger to do a dpanic.

agreed on keeping it consistent for now as well

ilackarms (approved):

jenshu (approved):

PR #9283, opened 2024-03-28
Declarative test setup

npolshakova:
The declarative environment builder tool can setup Kubernetes test cluster(s) with IstioOperators, Helm Charts, and any Applications required for testing. This is based on the declarative setup tool used in GME.

### How do I get started? 

A make target is provided to run the setup. A config file needs to be provided via `CONFIG` variable.

To test the simple gloo gateway setup:

```bash
CONFIG=test/setup/example_configs/gloo-gateway-setup.yaml make setup-declarative-env
```

This can also be run directly with: 
```bash
go run ./test/setup setup -c test/setup/example_configs/gloo-gateway-setup.yaml
```
Make sure the images and helm chart are present before running the go command directly.

### What happens if I already have a (partial) setup? 

Running the declarative setup will not automatically fail if you have different resources already setup. This is slightly different from the regression tests which require you to set `SKIP_TEARDOWN` or `SKIP_INSTALL`.

For example:
1. If you have a kind cluster already running, the declarative setup will use the existing kind cluster
2. If you already have a namespace, the declarative setup will use the existing namespace, add any labels defined in the config file and continue. 
3. If you have gloo already installed, the declarative setup will run a `helm upgrade -i` and upgrade the existing installation. 

### How do I write my own config? 

Configuration is typically provided via yaml file.  Usage is usually coupled with `make` targets
since a lot of default environment variables are defined there.

The configuration file is broken into a list of clusters, each with a distinct set of capabilities:

```yaml
clusters:
    - name: management-cluster
      management: true
      kindConfig:
        ...
      istioOperators:
        ...
      charts:
        ...
      namespaces:
        ...
      apps:
        ...
      images:
        ...
```

* kindConfig: The configuration for the KinD cluster
* istioOperators: Installation mechanism for Istio uses istio operators. This is a list of istio operator configs to install into the cluster. Uses the istioctl set by the `ISTIO_VERSION` environment variable, or the installed istioctl binary if not set for installation.
* charts: helm charts to install into the cluster.
    * name: name of the chart.
    * namespace: namespace to install the chart into
    * path: if true, the chart is installed from the local filesystem. Otherwise, it is installed from the configured helm repo.
    * values: values to pass to the chart.
* namespaces: namespaces to create in the cluster. Labels can be specified on the namespace. 
* apps: applications to install into the cluster. Follows the k8s resources apis and includes deployments, serviceaccounts, services, configmaps, etc. 
* images: images to load into the cluster. If an image override is defined on the helm chart, the setup command will also load that into the kind cluster so images only need to be defined in one place.


Makefile
```diff
@@ -921,6 +921,18 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
 	helm package --destination $(TEST_ASSET_DIR) $(HELM_DIR)
 	helm repo index $(TEST_ASSET_DIR)
 
+# Sets up environment for running e2e tests
+.PHONY: setup-declarative-env
```
sam-heilbron, line 925: I had imagined that the benefit for a declarative environment, is that the config we supply/declare is what defines what should be run. What if instead of having env variables that gate the steps, we just say the processor of the config says "if docker is to true, build the images", "if chart needs to be built, build the chart"...etc? That way we isolate complexity to this declarative setup?

test/setup/cmd/setup.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to load configuration")
+	}
+
+	if opts.printConfig {
+		return config.Print()
+	}
+
+	var istioctlBinary string
+	if os.Getenv(defaults.IstioctlVersionEnv) != "" {
```
sam-heilbron, line 74: How do we separate what is defined in the config and what is defined in the command line? I would have imagined that we could instead have the config support a definition of the istio version and we could parse it from there. Is there a reason we wouldn't want to do that?

test/setup/cmd/setup.go
```diff
@@ -0,0 +1,349 @@
...
+	for _, cluster := range set {
+		cluster := cluster
+
+		out = append(out, func() error {
+			var (
+				err        error
+				kubeConfig string
+			)
+
+			if cluster.KindConfig != nil {
```
sam-heilbron, line 158: More and more I kind of think we should split environment setup into 2 separate pieces of config:
1. My cluster(s) definition
2. My infrastructure setup (helm charts, additional services)

I think this will make it easier to support non-kind environments, and also allow devs to easily re-use cluster that may exist.

test/setup/helpers/helpers.go
```diff
@@ -0,0 +1,72 @@
...
+func GlooDirectory() string {
+	data, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
+	if err != nil {
+		panic(err)
+	}
+	return strings.TrimSpace(string(data))
+}
+
+// Merge maps, merge priority is left to right (i.e. right maps will overwrite left maps)
+func MergeValueMaps(maps ...map[string]interface{}) map[string]interface{} {
```
sam-heilbron, line 18: I think these utilities could be defined outside of our test package, and in a top level utils package, what do you think?

test/setup/helpers/timer.go
```diff
@@ -0,0 +1,30 @@
...
+	"strings"
+	"sync"
+	"time"
+)
+
+// synchronous timer aggregator
+var components []string
+var lock sync.Mutex
+
+func ListTimers() {
```
sam-heilbron, line 14: In our codegen library, we have https://github.com/solo-io/solo-kit/blob/main/pkg/code-generator/metrics/metrics.go. Perhaps we could re-use it?

test/setup/defaults/defaults.go
```diff
@@ -0,0 +1,20 @@
+package defaults
+
+const (
+	DefaultIstioTag           = "1.18.2"
+	DefaultIstioImageRegistry = "docker.io/istio"
+	IstioctlVersionEnv        = "ISTIOCTL_VERSION"
+
+	DefaultGlooImageRegistry = "quay.io/solo-io"
+
+	DefaultGlooImageName         = "gloo"
```
sam-heilbron, line 10: I like this! I think this deserves to be defined in Gloo code (ie not test code) and then used in tests. I had started to create a file here: https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/names.go

test/setup/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Load a configuration from a string.
+func Load(ctx context.Context, configPath string, reader io.Reader) (*types.Config, error) {
+	var (
+		err    error
+		config types.Config
+		logger = contextutils.LoggerFrom(ctx)
+	)
+
+	if os.Getenv("CLUSTER_NAME") == "" {
```
sam-heilbron, line 26: For env variables that are only used by tests, I try to define them in a single place https://github.com/solo-io/gloo/blob/main/test/testutils/env.go so that devs can see what is available. Can you re-use/expand these?

test/setup/helm/helm.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+type Opts struct {
+	Cluster    string
+	Client     k8s.Interface
+	Controller client.Client
+	Logger     *zap.SugaredLogger
+}
+
+func (o Opts) InstallChart(ctx context.Context, chart *types.Chart, cluster *kubernetes.Cluster) error {
```
sam-heilbron, line 33: A lot of this mirrors our CLI installer (which uses helm under the hood): https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/cmd/install/installer.go. I think we should try to consolidate these as much as possible, and ideally the logic for the installer should live in the CLI, and we can just invoke a top-level go function from here.

test/setup/istio/istioctl.go
```diff
@@ -0,0 +1,186 @@
...
+	"istio.io/istio/operator/pkg/apis/istio/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/solo-io/go-utils/contextutils"
+)
+
+func Install(
```
sam-heilbron, line 27: Should this be defined in the Istio commands of our CLI: https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/cmd/istio/root.go. Again, we can just call the function we need, instead of relying on the CLI artifact, but it might allow us to easily re-use code

npolshakova: I'd say no, mostly because the `istioctl` cli tool should be responsible for installing Istio. Even on the gloo-mesh side, `meshctl` doesn't install Istio. I think it would give `glooctl` too much responsibility if we owned the Istio installation and it would be be more work to maintain. 

Having this as a non user facing testing tool gives us flexibility, but doesn't require we maintain this in a real released artifact (so we don't have to test across istio versions, etc.)

test/setup/helm/helm.go
npolshakova: I don't think so, this is installing _any_ chart defined in the config file. It should be more flexible that the cli installer. If I wanted to install GME or Otel along side Edge, this helm installer would work for that. 

The user facing `glooctl` CLI tool depends on all of these imports: 
```
	"github.com/solo-io/gloo/install/helm/gloo/generate"
	"github.com/solo-io/gloo/pkg/cliutil"
	"github.com/solo-io/gloo/pkg/cliutil/helm"
	"github.com/solo-io/gloo/pkg/version"
	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/options"
	"github.com/solo-io/gloo/projects/gloo/cli/pkg/constants"
	"github.com/solo-io/gloo/projects/gloo/cli/pkg/helpers"
```

This file is agnostic to any gloo dependencies or constants and is only used for test env setup. It's not meant to be a released cli tool.

test/setup/config/config.go
npolshakova: Sounds good!

test/setup/helpers/timer.go
npolshakova: Or we can move these to the metrics util? Do you have an example that uses the solo-kit metrics? All of the examples seem to initialize the metrics aggregator in init() and then call it only once per operation like [here](https://github.com/solo-io/solo-kit/blob/5693efe646dd2249ff9e6a6436d1251e14541710/pkg/code-generator/cmd/main.go#L227)? Versus we want to time it per cluster:

```
timerFn := helpers.TimerFunc((fmt.Sprintf("[%s] %s helm installation", cluster.GetName(), name)))
defer timerFn()
```

test/setup/helpers/helpers.go
npolshakova: Yep agreed! The merge functionality is pretty useful!

Makefile
npolshakova: Yeah, I the separate docker/helm/declarative-setup is how GME handles it. I think it's possible to have the declarative setup also build the images/package the helm chart, but it's a little tricky because I think we'd want it to be identical to what we run in CI so we'd want to somehow reuse the make targets. I don't think it makes sense to hard code all the images/helm chart paths in the declarative setup code to do this since then that could drift.

test/setup/cmd/setup.go
npolshakova: The declarative setup does already re-use clusters that exists. If the KindConfig is not set, it'll just reuse the cluster with the same name from your kubeconfig. Eventually I think we'd want to add support loading images built from source to remote registries (AWS, GCP, etc ...), but if you have your kubeconfig pointing to a real cluster, this setup should still work (if you've manually uploaded the images)!

npolshakova: Having separate config makes it a little harder to read config for multi cluster setups (so that's why we have it in one file in GME: https://github.com/solo-io/gloo-mesh-enterprise/blob/main/test/gloo-core/gloo-core-multi-cluster.yaml#L8). Maybe on the edge side it makes more sense to separate the files, but I think that would still make it tricky to test multi cluster fed setups.

test/setup/defaults/defaults.go
npolshakova: Just moving this- is there any reason we put the default names in `kubeutils`?

npolshakova:
Closing in favor of new framework

PR #9307, opened 2024-04-02
hardcode envoy's runtime flag for sanitize TE to false

jbohanon:
# Description

Envoy's change to remove hop-by-hop TE headers broke upstreams whose gRPC client libraries expect a `TE: trailers` header. This is fixed in Envoy 1.29.3, which is not yet released. In the interim we will disable the sanitization.

# Context

Implementation: https://github.com/envoyproxy/envoy/pull/30535
Upstream Fix: https://github.com/envoyproxy/envoy/pull/32255
Upstream Fix backport: https://github.com/envoyproxy/envoy/pull/33018 

See slack conversation [here](https://solo-io-corp.slack.com/archives/C03MFATU265/p1712070803092839)

## Interesting decisions
 
We chose to do things this way in case 1.29.3 is not released at the time 1.17 goes GA or in case the bump is not made

## Testing steps

Helm tests upgraded, could stand to test that before/after a grpc upstream did or did not receive the headers

# Checklist:

### DIFF
--- a/changelog/v1.17.0-beta16/envoy-sanitize-te-false.yaml
+++ b/changelog/v1.17.0-beta16/envoy-sanitize-te-false.yaml
@@ -1,0 +1,9 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/gloo/issues/9232
+  resolvesIssue: false
+  description: >-
+    Hardcode Envoy's reloadable feature for TE header sanitization to false. This feature as it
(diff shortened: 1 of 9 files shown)


jbohanon:
Really hard to test this as it appears the restriction for the `:te` header exists in the c++ library but not in the go library for grpc

nfuden (approved):

ben-taussig-solo (approved):

PR #9312, opened 2024-04-02
gateway config crd tests

jbohanon:
# Description

Change deployer tests to be structured as a table. We generally have one gatewayclass, one gateway, one deployer, and zero or one gatewayparameters as input, and expect a set of resources we can make assertions on as output. This test setup allows us to pass an expected error or a validation function into the Table

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:


projects/gateway2/deployer/deployer_test.go
```diff
@@ -20,20 +24,54 @@ import (
...
+	panic("not implemented") // TODO: Implement
+}
+
+// GetEnvoyImage returns the envoy image and tag used by the proxy deployment.
+func (f *fakeK8sGatewayExtensions) GetEnvoyImage() extensions.Image {
+	return extensions.Image{
+		Repository: "gloo-envoy-wrapper",
+		Tag:        version.Version,
 	}
+}
```
jenshu, line 40: would these tests work with a "real" instance of the gatewayextensions? e.g. something like
```
mgr, err := ctrl.NewManager(&rest.Config{}, ctrl.Options{})
k8sGatewayExt, err = extensions.NewK8sGatewayExtensions(mgr)
```
i guess it would initialize the plugin registry, which we don't really need for these tests, but the GetEnvoyImage part would at least be functional and we wouldn't need to redefine the struct here

projects/gateway2/deployer/deployer_test.go
```diff
@@ -20,20 +24,54 @@ import (
...
 
-	// Create an empty ClusterRole object
-	var ret T
+type clientObjects []client.Object
 
-	// Unmarshal the JSON into the ClusterRole object
-	if err := json.Unmarshal(jsonBytes, &ret); err != nil {
-		panic(err)
+func (objs *clientObjects) deployment() *appsv1.Deployment {
```
jenshu, line 44: i think it would be worthwhile to check for a specific deployment name/namespace, just to be more precise (same with serviceaccount and the other ones below)

jenshu: and maybe they could be named findDeployment(), findServiceAccount(), etc?

projects/gateway2/deployer/deployer_test.go
```diff
@@ -42,7 +80,6 @@ var _ = Describe("Deployer", func() {
 
 		gwc *api.GatewayClass
```
jenshu, line 81: i'm not sure we need this gwc anymore; i think it was a remnant from when we were planning to use gatewayclass.parametersRef; can probably remove

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+					},
+					Extensions: &fakeK8sGatewayExtensions{},
+				},
+				gwc:     defaultGatewayClass(),
+				glooSvc: defaultGlooSvc(),
+				gw:      defaultGateway(),
+				gwp:     defaultGatewayParams(),
+			}, &expectedOutput{
+				validationFunc: func(objs clientObjects) error {
+					Expect(objs.deployment().Spec.Template.Spec.Containers).To(HaveLen(3))
```
jenshu, line 369: i know this test was already like this, but could we validate the names of the 3 containers?

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+			err            error
+			validationFunc func(objs clientObjects) error
 		}
-	}
-	return ""
-}
+
+		var (
+			defaultGwpName = "default-gateway-params"
+			defaultGlooSvc = func() *corev1.Service {
```
jenshu, line 194: i think we should be able to remove gloo svc from the tests, since the xds info is now initialized in setup syncer and passed to the deployer as input (rather than the deployer retrieving the info from the svc)

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+						Ports: []corev1.ServicePort{
+							{
+								Name: "grpc-xds",
+								Port: 1234,
+							},
+						},
+					},
+				}
+			}
+			defaultGatewayClass = func() *api.GatewayClass {
```
jenshu, line 210: same with gatewayclass, i think it can be removed since the deployer no longer tries to look up the gatewayclass

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+				dInputs: defaultDeployerInputs(),
+				gwc:     defaultGatewayClass(),
+				glooSvc: defaultGlooSvc(),
+				gw:      defaultGateway(),
+				gwp:     &gw2_v1alpha1.GatewayParameters{},
+			}, &expectedOutput{
+				validationFunc: func(objs clientObjects) error {
+					Expect(objs).NotTo(BeEmpty())
+					// Check we have Deployment, ConfigMap, ServiceAccount, Service
+					// TODO(jbohanon): validate everything??
```
jenshu, line 334: i think it would be good to validate that at least some key fields are equal to the defaults from values.yaml

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+				gwc:     defaultGatewayClass(),
+				glooSvc: defaultGlooSvc(),
+				gw:      defaultGatewayWithGatewayParams(defaultGwpName),
+				gwp:     defaultGatewayParams(),
+			}, &expectedOutput{
+				err: nil,
+				validationFunc: func(objs clientObjects) error {
+					Expect(objs).NotTo(BeEmpty())
+					// Check we have Deployment, ConfigMap, ServiceAccount, Service
+					Expect(objs).To(HaveLen(4))
```
jenshu, line 350: should check the object names and some key fields too

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+					Expect(node).To(HaveKeyWithValue("metadata", map[string]any{
+						"gateway": map[string]any{
+							"name":      gw.Name,
+							"namespace": gw.Namespace,
+						},
+					}))
+					return nil
+				},
+			}),
+			// Entry("no gateway class", &input{
```
jenshu, line 513: these can be removed

projects/gateway2/deployer/deployer_test.go
```diff
@@ -426,20 +174,407 @@ var _ = Describe("Deployer", func() {
...
+										Deployment: &gw2_v1alpha1.ProxyDeployment{
+											Replicas: &wrappers.UInt32Value{Value: 3},
+										},
+									},
+									EnvoyContainer: &gw2_v1alpha1.EnvoyContainer{
+										Bootstrap: &gw2_v1alpha1.EnvoyBootstrap{
+											LogLevel:          "debug",
+											ComponentLogLevel: "router:info,listener:warn",
+										},
+									},
```
jenshu, line 275: can you add a test that validates that the output resources contain these values? and perhaps some other values like envoy image, and service type, and annotations?

jenshu (commented):
looks great so far! have some suggestions for cleanup and more validation

i think this is a good set of tests to start with; we can always add to them later

projects/gateway2/deployer/deployer_test.go
jbohanon: I'll take a look! I didn't know how to get a controller-runtime manager and didn't dive too deep, but it probably is cleaner to just use the real struct. Thanks for the snippet

projects/gateway2/deployer/deployer_test.go
jbohanon: Ah so it will need to be on the default deployer inputs and any we override though, right?

jenshu: yep, we'll need to pass in the ControlPlane as part of the input (looks like we already do in some places):
```
			ControlPlane: bootstrap.ControlPlane{
				Kube: bootstrap.KubernetesControlPlaneConfig{XdsHost: "something.cluster.local", XdsPort: 1234},
			},
```

jenshu:
LGTM!

PR #9316, opened 2024-04-03
testutils: create curl and kubectl utilities

sam-heilbron:
# Description
Update testing and CLI utilities for execute curl requests

# Context
## Background
We are in the process of building out an improved e2e testing framework. For that framework to be effective, we need some standard utilities that can be used across tests. This largely is:
- Curl commands
- Kubectl commands
- Port-Forwarding
- Performing assertions on Curl responses
- Executing Curl requests from a kubernetes pod

Once we have these common utilities, it will be easier to write effective tests. 

Some of the concepts in this PR have been extracted/inspired by: https://github.com/solo-io/gloo/pull/9284

## CurlRequestBuilder
This logic previously was co-located with another concept, a test container. In the future, we will be using this curl logic in other places, so I opted to split this into a standalone component.

It does not have tests at the moment, because I used it where we previously has the testContainer logic.

## Kubectl
This was inspired by https://github.com/solo-io/gloo-mesh-enterprise/tree/main/test/utils/kubectl. I opted to make this general utility however (instead of just isolating it to tests), since there are instances where we exexute kubectl commands within our Gloo CLI.

I updated our tests to rely on this command, that is why I did not write explicit tests for this component.

## Curl Transformation
This just ports logic that we have previously built into Enterprise: https://github.com/solo-io/solo-projects/blob/main/test/gomega/transforms/curl.go

There were uses for it in open source. 

I added the explicit tests we had before, and used the transform within the testContainer.

## TestContainer Assertion
Previously we had a fairly weak assertion: that a curl response string would contain a substring.

The purpose of our [http response matcher](https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_http_response.go) is to be able to be more explicit with responses.

Some tests could only work with a substring match, but others were able to be more explicit and we could assert that the response was a 200 as well. To make this work, I had to add some simple logic to the testContainer to generate a gomega.Matcher based on what was provided by the caller.

## Interesting decisions
 -

## Testing steps
Each of the above sections attempt to call out what tests were introduced for these utilities, or why there weren't tests added.

## Notes for reviewers
-

## Next Steps
1. I want to take our new Kube CLI tool, and use it to build a `glooctl export report` or something of the like, and have our tests use that functionality on failure.
2. I want to create a single example test, similar to Nina's (https://github.com/solo-io/gloo/pull/9284) and get feedback on a single test

# Checklist:


test/testutils/curl_request.go
```diff
@@ -0,0 +1,215 @@
+package testutils
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/onsi/ginkgo/v2"
```
npolshakova, line 8: Do we need the ginkgo dependency here? If we don't include it, it's easier to reuse these utils for non-ginkgo tests.

test/testutils/curl_request.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (c *CurlRequestBuilder) BuildArgs() []string {
+	ginkgo.GinkgoHelper()
+
+	if err := c.errorIfInvalid(); err != nil {
+		// We error loudly here
+		// These types of errors are intended to prevent developers from creating resources
+		// which are semantically correct, but lead to test flakes/confusion
+		ginkgo.Fail(err.Error())
```
npolshakova, line 167: It looks like we're only using ginkgo here to fail- can we just return the error instead?

test/kube2e/istio/istio_integration_test.go
```diff
@@ -879,7 +884,10 @@ var _ = Describe("Gloo + Istio integration tests", func() {
 						Verbose:           false,
 						WithoutStats:      true,
 						ReturnHeaders:     false,
-					}, "upstream connect error or disconnect/reset before headers. reset reason: connection termination", 1, time.Minute*1)
+					}, &testmatchers.HttpResponse{
+						StatusCode: 503,
```
npolshakova, line 888: [nit] http.StatusServiceUnavailable

test/testutils/curl_request.go
sam-heilbron: Yeah it's a great point. We don't necessarily need it, and I'm happy to try to find ways to remove it. Given that ginkgo/gomega is used to widely in our codebase (and in other testing utilities), I wonder what value you see as removing this from this specific utility.

test/testutils/curl_request.go
sam-heilbron: We could, and I'm happy to do that. My hope was basically that we know there are cases that should never pass, and so instead of deferring the assertion to the caller of this function, we could bake the logic directly in. 

What do you think about the tradeoff of baking the logic into this component versus delegating it to the user?

test/testutils/curl_request.go
npolshakova: Yeah, we had something similar on the GME side and had to switch to remove all the non-ginkgo logic: https://github.com/solo-io/gloo-mesh-enterprise/pull/13425 

The biggest values of removing it was being able to reuse the kubectl logic for non ginkgo/gomega tests and isolate the test util functionality vs. the assertion functionality.

test/testutils/curl_request.go
npolshakova: Yeah I think it depends- it looks like the `errorIfInvalid()` is called only when the `c.service` is unset? But then the`DefaultCurlRequestBuilder` doesn't have the service set? It looks like setting the service field is already user controlled, not something that's set by the builder, so it's already the user's responsibility. 

If `c.service` was set by the builder based on some env/dependency, I think this pattern of erroring loudly makes sense. It does have the tradeoff that we're trying our test util code to ginkgo.

npolshakova (approved):

pkg/cliutil/install/kubernetes.go
```diff
@@ -45,24 +34,13 @@ func SetVerbose(b bool) {
...
-		kubectl.Stderr = cliutil.GetLogger()
-	}
-	return kubectl.Run()
+	_, err := KubectlOut(stdin, args...)
+	return err
 }
 
+// Deprecated: Prefer kubectl.Cli
 func KubectlOut(stdin io.Reader, args ...string) ([]byte, error) {
 	kubectl := exec.Command("kubectl", args...)
```
jbohanon, line 46: Does it make sense to pass this through now to make this call into `kubectl.Cli`?

pkg/utils/cmdutils/cmd.go
jbohanon, line 1: Is this abstraction useful at this point? I don't see anywhere that we are accepting a `Cmd` and thus exercising the flexibility allowed by this interface. I see a ton of places where we return a `Cmd`, but these could all probably return the implementation without any other downsides?

pkg/utils/cmdutils/run_error.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+)
+
+// RunError represents an error running a Cmd
+type RunError struct {
+	Command    []string // [Name Args...]
+	Output     []byte   // Captured Stdout / Stderr of the command
+	Inner      error    // Underlying error if any
+	StackTrace error
+}
```
jbohanon, line 15: Does it make sense to have unexported types with a constructor or builder for this type since we really don't want the fields listed modified by any consumers of returned `RunError`s?

pkg/utils/cmdutils/cmd.go
```diff
@@ -0,0 +1,31 @@
...
+	SetEnv(...string) Cmd
+
+	// SetStdin sets the io.Reader used for stdin
+	SetStdin(reader io.Reader) Cmd
+
+	// SetStdout sets the io.Writer used for stdout
+	SetStdout(io.Writer) Cmd
+
+	// SetStderr sets the io.Reader used for stderr
+	SetStderr(io.Writer) Cmd
```
jbohanon, line 25: nit:
These are more akin to `WithX` functions imo since they return their owner for chaining

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -0,0 +1,129 @@
...
+
+// NewCli returns an implementation of the kubectl.Cli
+func NewCli(receiver io.Writer) *Cli {
+	return &Cli{
+		receiver:    receiver,
+		kubeContext: "",
+	}
+}
+
+func (c *Cli) SetKubeContext(kubeContext string) *Cli {
```
jbohanon, line 36: same nit:
`WithKubeContext`

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -0,0 +1,129 @@
...
+	cmd := cmdutils.Command(ctx, "kubectl", args...)
+
+	//disable DEBUG=1 from getting through to kube
+	env := os.Environ()
+	for i, pair := range env {
+		if strings.HasPrefix(pair, "DEBUG") {
+			env = append(env[:i], env[i+1:]...)
+			break
+		}
+	}
```
jbohanon, line 55: could this be extracted to `cmdutils`?

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -0,0 +1,129 @@
...
+	applyArgs := append([]string{"-f", "-"}, extraArgs...)
+	return c.ApplyCmd(ctx, content, applyArgs...).Run().Cause()
+}
+
+func (c *Cli) ApplyFile(ctx context.Context, file string, extraArgs ...string) error {
+	applyFileArgs := append([]string{"-f"}, extraArgs...)
+	return c.ApplyCmd(ctx, []byte(file), applyFileArgs...).Run().Cause()
+}
+
+func (c *Cli) DeleteCmd(ctx context.Context, content []byte, extraArgs ...string) cmdutils.Cmd {
```
jbohanon, line 87: If this weren't implementing an interface, would it be cleaner to not expose `DeleteCmd` and add confusion? Does it add any value over the `Delete` and `DeleteFile` functinos?

test/kube2e/gateway/gateway_test.go
```diff
@@ -698,20 +703,17 @@ var _ = Describe("Kube2e: gateway", func() {
...
-				dialContext := context.Background()
-				portFwd := exec.Command("kubectl", "port-forward", "-n", testHelper.InstallNamespace,
-					"deployment/gloo", "9966")
-				portFwd.Stdout = os.Stderr
-				portFwd.Stderr = os.Stderr
-				err := portFwd.Start()
-				Expect(err).ToNot(HaveOccurred())
+				portForwarder, err := kubeCli.StartPortForward(ctx,
+					portforward.WithDeployment(kubeutils.GlooDeploymentName, testHelper.InstallNamespace),
+					portforward.WithRemotePort(9966),
```
jbohanon, line 708: I get that it was in the tests before, but do we know where this number came from and if it can be constant-ified?

test/kube2e/helper/curl.go
```diff
@@ -103,105 +110,134 @@ func (t *testContainer) CurlEventuallyShouldOutput(opts CurlOpts, substr string,
...
+	switch a := expectedOutput.(type) {
+	case types.GomegaMatcher:
+		// To handle all edge cases, developers must define the expected matcher explicitly
+		return a
+	case *matchers.HttpResponse:
+		// If a response is provided, we assume we can convert the string into an http.Response
+		return WithTransform(transforms.WithCurlHttpResponse, matchers.HaveHttpResponse(a))
+
+	case string:
+		// If a string is provided, we assume it was a successful response
```
jbohanon, line 168: I'm not sure why this is a valid assumption. If this is just to support the existing/legacy pattern of passing the expected response string in, we should call that out and state that devs are expected to pass a matcher in.

test/kube2e/helper/test_container.go
```diff
@@ -32,9 +32,9 @@ type TestContainer interface {
 	TerminatePodAndDeleteService() error
 	CanCurl() bool
 	// Checks the response of the request eventually meets expectation
-	CurlEventuallyShouldRespond(opts CurlOpts, substr string, ginkgoOffset int, timeout ...time.Duration)
+	CurlEventuallyShouldRespond(opts CurlOpts, response interface{}, ginkgoOffset int, timeout ...time.Duration)
 	// Checks all of the output of the curl command eventually meets expectation
-	CurlEventuallyShouldOutput(opts CurlOpts, substr string, ginkgoOffset int, timeout ...time.Duration)
+	CurlEventuallyShouldOutput(opts CurlOpts, output interface{}, ginkgoOffset int, timeout ...time.Duration)
```
jbohanon, line 37: Wouldn't hurt to callout here that the interface is expected to be a matcher

test/kube2e/istio/istio_integration_test.go
```diff
@@ -284,7 +286,7 @@ var _ = Describe("Gloo + Istio integration tests", func() {
 				Verbose:           false,
 				WithoutStats:      true,
 				ReturnHeaders:     true,
-			}, fmt.Sprintf("HTTP/1.1 %d", expected), 1, time.Minute*1)
+			}, ContainSubstring(fmt.Sprintf("HTTP/1.1 %d", expected)), 1, time.Minute*1)
```
jbohanon, line 289: Could this and L523 be a `*testmatchers.HttpResponse` as demonstrated elsewhere in this file?

pkg/cliutil/install/kubernetes.go
sam-heilbron: I had originally tried to do that as part of this PR, and then the number of impacted areas expanded dramatically, and some of the logging functionality was different. My plan instead was to introduce this new CLI, and then in follow-up work, figure out how to deprecate and delete this old code, so that we can have a single implementation that works for all cases

pkg/utils/cmdutils/cmd.go
sam-heilbron: Yeah it's a fair questions. I like having it for a few reasons (listed below), but I could be convinced otherwise:
1. We use the exec.Cmd in a lot of places in the code, in slightly different ways (with different logging mostly). This would provide a way to standardize it. Also, it means that if we build customizations into what the Run function returns, or how env variables are processed, all usage of it can inherit those benefits
2. I like the fluent API, that allows you to chain SetStdin, SetStderr to calls (cmd.SetStdin().SetStdout().Run())
3. My hope was that when we introduce functionality to execute a curl request via the kube CLI, developers would want to construct a Cmd, and then feed that into the kube CLI, and we would capture the arguments and pass them to the kube CLI

pkg/utils/kubeutils/kubectl/cli.go
sam-heilbron: Good point, I'm happy to remove. I had added it because I viewed this API as un-opinionated, and thus any calls that I user could want to make , I wanted to support. But given that it isn't used yet, and might just cause confusion, I'm happy to remove

pkg/utils/kubeutils/kubectl/cli.go
sam-heilbron: Good call!

jbohanon (approved):

npolshakova (approved):

sam-heilbron:
bulldozer

PR #9322, opened 2024-04-04
Refactor Makefile docker variant targets and fix for m1 machines

davidjumani:
# Description

- Refactor the Makefile grouping targets based on variants to simplify large make targets
- Adds the amd64 platform for all distroless base images since we only publish amd64 images and use amd64 envoy. This also unblocks distroless dev on m1. 
When building distroless images on m1, it fails when using the distroless-base image even if it exists as the arch is amd64. It tries to fetch the arm64 distroless-base image from quay that doesn't exist and shows this error
`
ERROR: failed to solve: [quay.io/solo-io/distroless-base:1.0.1-dev](http://quay.io/solo-io/distroless-base:1.0.1-dev): failed to resolve source metadata for [quay.io/solo-io/distroless-base:1.0.1-dev](http://quay.io/solo-io/distroless-base:1.0.1-dev): unexpected status from HEAD request to https://quay.io/v2/solo-io/distroless-base/manifests/1.0.1-dev: 401 UNAUTHORIZED
`
This fixes it by ensuring we only use amd64 images for distroless builds

# Checklist:


Makefile
```diff
@@ -867,82 +867,102 @@ docker-push-%-distroless:
...
+.PHONY: docker-distroless-retag
+docker-distroless-retag: docker-retag-gloo
+docker-distroless-retag: docker-retag-discovery
+docker-distroless-retag: docker-retag-gloo-envoy-wrapper
+docker-distroless-retag: docker-retag-sds
+docker-distroless-retag: docker-retag-certgen
+docker-distroless-retag: docker-retag-ingress
+docker-distroless-retag: docker-retag-access-logger
+docker-distroless-retag: docker-retag-kubectl
+
```
sheidkamp, line 955: Should these be `-distroless`?

sheidkamp (changes requested):
One spot where it looks like there was a c/p error. Otherwise LGTM.

Makefile
```diff
@@ -867,82 +867,102 @@ docker-push-%-distroless:
...
+.PHONY: docker-distroless
+docker-standard: check-go-version
+docker-standard: gloo-distroless-docker
+docker-standard: discovery-distroless-docker
+docker-standard: gloo-envoy-wrapper-distroless-docker
+docker-standard: sds-distroless-docker
+docker-standard: certgen-distroless-docker
+docker-standard: ingress-distroless-docker
+docker-standard: access-logger-distroless-docker
+docker-standard: kubectl-distroless-docker
```
sheidkamp, line 890: these `docker-standards` should be `docker-distroless`

sheidkamp (changes requested):
one more c/p error

sheidkamp (approved):

Makefile
davidjumani: Good catch! Thanks

Makefile
davidjumani: Good catch! Thanks

ci/kind/setup-kind.sh
```diff
@@ -18,6 +18,8 @@ KUBE2E_TESTS="${KUBE2E_TESTS:-gateway}"  # If 'KUBE2E_TESTS' not set or null, us
 # The version of istio to install for glooctl tests
 # https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases
 ISTIO_VERSION="${ISTIO_VERSION:-1.18.2}"
+# Set the default image variant to standard
+IMAGE_VARIANT="${IMAGE_VARIANT:-standard}"
```
bewebi, line 22: Doesn't this need to get passed as an env var to step #2 below?

sheidkamp (approved):

bewebi (dismissed):
LGTM

nfuden (approved):

sheidkamp (approved):

PR #9327, opened 2024-04-05
cli: Introduce glooctl export report

sam-heilbron:
# Description

Introduce `glooctl export report`

# Context
## Background
As I work on e2e tests for Gloo Gateway, I have observed that many of our tools/utilities have different implementations spread across the codebase. The following actions are things that we have different implementations of:
- Executing a curl request
- Asserting a curl response returned expected data
- Opening a port-forward
- Querying the envoy admin api
- Capturing the state of a cluster

I think we need to have standard ways of executing actions and then re-use that everywhere it is needed. I have merged previous PRs that attempt to improve this, and this PR is an extension of:
- https://github.com/solo-io/gloo/pull/9316
- https://github.com/solo-io/gloo/pull/9314
- https://github.com/solo-io/gloo/pull/9226

## CLI
I'm happy to discuss other names for this tool. My hope was that the CLI code is as minimal as possible, and just passes user-defined options to some re-usable code.

### envoyutils.AdminCtl
This is a reusable utility for executing curl requests agains the Envoy Admin API. It is intended to store all details about accessing the Envoy API, and allow developer to use the utility methods that it exposes.

This is used by the export utility (defined below) to capture details from a running envoy instance.

### Export utility
This is a utility which will be invoked by our CLI (`glooctl export report`) or our tests, when they fail. I could see this utility living in other places in the codebase too, but for now I just placed it in a shared `cliutils` location.

It is built with the intention of just collecting lots of small modules, which should be re-used in other areas of our codebase.

**The current implementation just tries to define a structure, and is not feature-complete. I'd rather get consensus on the existing code, merge it, and then expand the usage of it in follow-up PRs.**

## Testing steps
1. Install Gloo Gateway in `gloo-system` namespace
2. Build CLI Locally:
```
make build-cli-local
```

3. Use the local CLI to export a report:
```
_output/glooctl-darwin-arm64 export report --output-dir ./_output
```

4. Untar the output file, and inspect the content to find envoy data captured in files

## Notes for reviewers
-

# Checklist:


pkg/cliutil/export/cmd.go
```diff
@@ -0,0 +1,33 @@
+package export
+
+import (
+	"github.com/solo-io/gloo/pkg/utils/cmdutils"
+	"os"
+	"path/filepath"
+)
+
+func RunCommandOutputToFile(cmd cmdutils.Cmd, path string) func() error {
```
nfuden, line 9: nit: export comment?

pkg/utils/cmdutils/local.go
```diff
@@ -31,7 +31,7 @@ func (c *LocalCmder) Command(ctx context.Context, name string, arg ...string) Cm
 	}
 }
 
-// LocalCmd wraps os/exec.Cmd, implementing the kind/pkg/exec.Cmd interface
+// LocalCmd wraps os/exec.Cmd, implementing the cmdutils.Cmd interface
```
nfuden, line 34: do we no longer implement the kind exec interface? https://pkg.go.dev/sigs.k8s.io/kind/pkg/exec

pkg/utils/envoyutils/adminctl/cli.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/solo-io/gloo/pkg/utils/cmdutils"
+	"github.com/solo-io/gloo/test/testutils"
+	"github.com/solo-io/go-utils/contextutils"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
```
nfuden, line 14: Can we minimally include a link to the source we are building against ? https://github.com/envoyproxy/envoy/blob/63bc9b564b1a76a22a0d029bcac35abeffff2a61/source/server/admin/admin.cc#L127

Perhaps a todo to autogenerate from the makeHandler calls?

pkg/utils/errutils/concurrent.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"sync"
+)
+
+// AggregateConcurrent runs fns concurrently, returning a NewAggregate if there are > 1 errors
+func AggregateConcurrent(funcs []func() error) error {
+	// run all fns concurrently
+	ch := make(chan error, len(funcs))
```
nfuden, line 11: consider a chan of struct{fxnIdx int, err error} to make sure that error output is ordered consistently

projects/gloo/cli/pkg/cmd/export/report/command.go
```diff
@@ -0,0 +1,35 @@
...
+
+	return cmd
+}
+
+func doExportReport(exportOptions *options.Options) error {
+	// The CLI should publish to a consistent file name (report.tar.gz)
+	// We include the time from when the report was captured, to easily distinguish between
+	// separate exports
+	outputFile := filepath.Join(
+		exportOptions.OutputDir,
```
nfuden, line 32: If this is to be in the same location but have potentially different types of subchecks run based on exportOptions I think that we MUST include the options used to run this in the report file.

nfuden (commented):
This generally makes sense to me and I really like the seperation of export and how it will be composed of sub functions that may run in parallel. 
Left some more implementation based thoughts rather than large picture thoughts.

Issue #9337, opened 2024-04-08
Glooctl Test Suite (Kube2e)


sam-heilbron:
### Which tests failed?

Logs: https://github.com/solo-io/gloo/actions/runs/8606475721/job/23584985419?pr=9335
Error:
```
  gateway-proxy-575476ff7b-qv2zz      2/3     ErrImagePull   0          62s
  gateway-proxy-6db76c97cc-lxcm9      1/1     Running        0          2m5s
```

### Initial Investigation

We are failing to pull an image, and thus we cannot route traffic through the proxy.

In the logs I see:
```
Gloo version found, using "1.0.0" for SDS sidecar
```

This means that when we inject the sds sidecar, we are trying to pull an image with `1.0.0` as the tag. That is not built as part of the test (we tag imgages with `1.0.0-ci`).

I believe this is due to changes introduced in: https://github.com/solo-io/gloo/pull/9212

### Additional Information

- [Internal Slack context](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1712610591897009)

PR #9338, opened 2024-04-08
glooctl: resolve test flake

sam-heilbron:
# Description

Fix a bug in the way that we inject an SDS sidecar in our CI tests

# Context
To determine the SDS image tag we:
1. Get the gloo deployment
2. Extract the gloo-oss-tag annotation from the pod template
3. Convert that string into a version object
4. Convert version object into a string

There is subtle behavior in our version string that wipes the label from the stringified value, if the labelVersion is 0.

This meant that because we tagged images with `1.0.0-ci`, the stringified version was `1.0.0`. As a result, when injecting the SDS sidecar, we would try to pull an image with `1.0.0` as the image tag and would get an `ErrImagePull`.

## Interesting decisions
 -

## Testing steps
- I have added unit tests to demonstrate this behavior
- I have run the `glooctl` test suite locally to prove this works

## Notes for reviewers
-

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9337


projects/gloo/cli/pkg/prerun/version_warning_test.go
```diff
@@ -186,27 +186,6 @@ var _ = Describe("version command", func() {
 		err = prerun.WarnOnMismatch(ctx, binaryName, versionGetter, logger)
 	})
 
-	It("should warn when the versions differ in gateway pod", func() {
```
sam-heilbron, line 189: The gateway pod does not exist anymore, so this code was not doing anything

projects/gloo/cli/pkg/prerun/version_warning.go
```diff
@@ -148,31 +144,31 @@ type ContainerVersion struct {
 	Version       *versionutils.Version
 }
 
-// return an array of open source gloo versions found in the cluster
-// this is determined by looking at either the version of gateway (if the pod is present) or the annotation in the gloo pod.
+// GetOpenSourceVersions returns an array of open source gloo versions found in the cluster
+// this is determined by looking at either the annotation on the gloo pod.
```
jenshu, line 148:
```suggestion
// this is determined by looking at the annotation on the gloo pod.
```

projects/gloo/cli/pkg/cmd/istio/util.go
```diff
@@ -133,7 +135,20 @@ func GetGlooVersion(ctx context.Context, namespace string) (string, error) {
 	if len(openSourceVersions) > 1 {
 		fmt.Printf("Found multiple gloo versions, picking %s", openSourceVersions[0].String())
 	}
-	return openSourceVersions[0].String(), nil
+
```
jenshu, line 138: just curious why is this func under istio rather than somewhere more generic?

jenshu (commented):
looks good, can you just make sure that all instances of `1.0.0-ci` are replaced with the new string (if appropriate)? i saw a few more instances in the repo

projects/gloo/cli/pkg/cmd/istio/util.go
sam-heilbron: I agree, it feels like a misplaced utility. I decided not to change it as part of this PR, to keep the scope limited to just the fix, but I agree it would be good to move to a more standard location.

jenshu (dismissed):

jbohanon (dismissed):
Curious if/how our PR version tags are handled since they are rejected by semver validation in versionutils

jenshu (approved):

sam-heilbron:
During this PR, I encountered a flake in the same suite (https://github.com/solo-io/gloo/actions/runs/8608274784/attempts/1?pr=9338) but it was a different error. As a result, I am not addressing that flake, as it is more important that we fix the one that is known to occur frequently.

Once we solve the flake/failure that this PR attempts to resolve, we can look into addressing the most recent one

sam-heilbron:
I hit a new error: 
```
  [FAILED] Timed out after 60.000s.
  The function passed to Eventually failed at /home/runner/work/gloo/gloo/test/kube2e/helper/curl.go:159 with:
  Expected
      <*http.Response>: {
          Status:     <string>: ""
          StatusCode: <int>: 0
          Body:       <string>: "upstream connect error or disconnect/reset before headers. reset reason: remote connection failure, transport failure reason: TLS_error:|268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER:TLS_error_end:TLS_error_end< HTTP/1.1 503 Service Unavailable\r\n"
      }
  to have HTTP status
      <int>: 200 

  expected: {
    "StatusCode": 200,
    "Body": {
      "Substr": "[{\"id\":1,\"name\":\"Dog\",\"status\":\"available\"},{\"id\":2,\"name\":\"Cat\",\"status\":\"pending\"}]",
      "Args": null
    },
    "Headers": null,
    "Custom": null
  }
  In [It] at: /home/runner/work/gloo/gloo/test/kube2e/glooctl/istio_test.go:141 @ 04/09/24 01:07:50.352

```
This is certainly an issue, though I don't believe the changes in this PR caused this. More likely, I think the bug that this PR fixes were masking this. I am re-running CI to see how frequent this issue is.

jbohanon (approved):

PR #9341, opened 2024-04-09
Route option status

lgadban:
Introduces `status` to `RouteOption` resources used for policy attachment

- Introduces `StatusPlugins` that will run post-xds translation and receive a full set of translation reports from Gloo
- Update `RouteOptionsPlugin` to be a `StatusPlugin` and report status on the `RouteOption` objects directly based on the translation reports from xds translation
  - Currently utilizes legacy Edge status format
  - Requires a `gatewayv1.RouteOptionClient` to get RouteOptions and `reporter.StatusReporter` to persist the status
  - We will adapt this to use native K8s status and K8s GWAPI `PolicyStatus` next
- Add `SourceMetadata` to `validation.RouteReport` to allow for tracking of RouteOptions used on a given gloov1.Route (and therefore determining which RouteOption was used for processing errors on Routes)


projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -156,3 +258,45 @@ func formatNotFoundMessage(routeCtx *plugins.RouteContext, filter *gwv1.HTTPRout
...
+					}
+				}
+			}
+		}
+	}
+	return routeErrors
+}
+
+// if the Route error has a RouteOption source associated with it, extract the source and return it
+func extractRouteOptionSourceKeys(routeErr *validation.RouteReport_Error) *types.NamespacedName {
```
jbohanon, line 286: Nit:
It looks like we are returning a pointer here to indicate error states within the function. I think it would be more idiomatic to return by value and with an error (i.e. `(types.NamespacedName, error)`). In this way, we can check `err != nil` to potentially handle different cases more gracefully, as well as avoiding potentially footgunning ourselves with a NPE later

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -156,3 +258,45 @@ func formatNotFoundMessage(routeCtx *plugins.RouteContext, filter *gwv1.HTTPRout
 		routeCtx.Route.GetNamespace(),
 	)
 }
+
+// given a ProxyReport, extract and aggregate all Route errors that have RouteOption source metadata
+// and key them by the source RouteOption NamespacedName
+func extractRouteErrors(proxyReport *validation.ProxyReport) map[types.NamespacedName][]*validation.RouteReport_Error {
+	routeErrors := make(map[types.NamespacedName][]*validation.RouteReport_Error)
+	for _, lr := range proxyReport.GetListenerReports() {
```
jbohanon, line 266: This is so many nested for loops. I understand that it would be out of scope to look at revisiting the structure of the proxy report object, but I think this if this is the only way to access these objects then it might be warranted soon.

jbohanon (commented):
Not a full review; just trying to familiarize myself with the space

projects/gateway2/controller/start.go
```diff
@@ -48,7 +50,10 @@ type StartConfig struct {
 
 	// ProxyClient is the client that writes Proxy resources into an in-memory cache
 	// This cache is utilized by the debug.ProxyEndpointServer
-	ProxyClient v1.ProxyClient
+	ProxyClient       v1.ProxyClient
+	RouteOptionClient gatewayv1.RouteOptionClient
```
sam-heilbron, line 54: nit: Since this is the entry-point to this controller, I've found it helpful to have each of the properties well-commented, so that future devs know what these clients are used for, could you add a brief outline for the client and status reporter?

projects/gateway2/extensions/extensions.go
```diff
@@ -16,17 +18,29 @@ type K8sGatewayExtensions interface {
 }
 
 // K8sGatewayExtensionsFactory returns an extensions.K8sGatewayExtensions
-type K8sGatewayExtensionsFactory func(mgr controllerruntime.Manager) (K8sGatewayExtensions, error)
+type K8sGatewayExtensionsFactory func(
```
sam-heilbron, line 25: There are 3 stages of configuration:
1. Construct a factory 
2. Each run, use that factory to construct a set of Extensions
3. Use the extensions to build things, like plugins


Do we need the RouteOptionClient and StatusReporter to be set at the second stage? What if instead, we constructed a factory that had these already defined?

projects/gateway2/reports/reporter.go
```diff
@@ -27,6 +27,7 @@ type RouteReport struct {
 	observedGeneration int64
 }
 
+// TODO: rename to e.g. RouteParentRefReport
```
sam-heilbron, line 30: nit: An alternative would be to create a RouteParetnRefReport struct, and type alias it to this, and deprecate the old format. Then you could start using what you wanted

projects/gateway2/translator/listener/gateway_listener_translator.go
```diff
@@ -518,5 +519,6 @@ func makeVhostName(
 	domain string,
 ) string {
 	// TODO is this a valid vh name?
+	// Looks like it... see: projects/gloo/pkg/utils/sanitize.go
```
sam-heilbron, line 522: Given this question (first comment) and clarifying comment (I found helpful), should we just call sanitize name to ensure it is valid?

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -49,9 +50,25 @@ type TranslatedGateway struct {
...
 		postTranslationContext *PostTranslationContext,
 	) error
 }
+
+type StatusContext struct {
+	ProxiesWithReports []translatorutils.ProxyWithReports
+}
+
+// Plugin that recieves proxy reports post-xds translation to handle any status reporting necessary
+type StatusPlugin interface {
```
sam-heilbron, line 67: Can you help me understand how the responsibility of a StatusPlugin differs from a PostTranslationPlugin?

projects/gateway2/translator/routeutils/utils.go
```diff
@@ -0,0 +1,28 @@
+package routeutils
+
+import (
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func AppendSourceToRoute(route *v1.Route, source client.Object) {
```
sam-heilbron, line 9: nit: Does this need to live in the gateway2 project structure? Or could this live in the gloo project code?

projects/gateway2/xds/xds_syncer.go
```diff
@@ -274,18 +290,21 @@ func (s *XdsSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapshot) rep
 			Messages: map[*core.ResourceRef][]string{},
 		}
 
-		xdsSnapshot, reports, _ := s.translator.Translate(params, proxy)
-
-		// Messages are aggregated during translation, and need to be added to reports
```
sam-heilbron, line 279: It looks like this was removed. If this is no longer necessary/implemented, could we add a comment a few lines up where the messages are initialized explaining why?

projects/gloo/pkg/syncer/setup/start_func.go
```diff
@@ -56,6 +59,13 @@ func K8sGatewayControllerStartFunc() StartFunc {
 			return err
 		}
 
+		routeOptionClient, err := gateway.NewRouteOptionClient(ctx, opts.RouteOptions)
+		if err != nil {
+			return err
+		}
+		statusClient := statusutils.GetStatusClientForNamespace(opts.StatusReporterNamespace)
```
sam-heilbron, line 66: I left a comment earlier about where we inject these structs. I figure we have a couple options here as well:
1. In the setup_syncer code that calls `K8sGatewayControllerStartFunc` we construct these (https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/syncer/setup/setup_syncer.go#L612), so we could just pass them as arguments to the StartFunc, instead of re-creating them
2. We could create the Extensions object with these two populated

projects/gloo/pkg/translator/route_config.go
```diff
@@ -390,6 +390,15 @@ func (h *httpRouteConfigurationTranslator) runRoutePlugins(
 			if isWarningErr(err) {
 				continue
 			}
+			if staticMetadata := in.GetMetadataStatic(); staticMetadata != nil {
```
sam-heilbron, line 393: Should we add a comment to clarify the why metadataStatic would be nil?

projects/gateway2/xds/xds_syncer.go
```diff
@@ -191,7 +192,8 @@ func (s *XdsSyncer) Start(ctx context.Context) error {
 			TranslatedGateways: translatedGateways,
 		})
 
-		s.syncEnvoy(ctx, proxyApiSnapshot)
+		proxiesWithReports := s.syncEnvoy(ctx, proxyApiSnapshot)
+		s.applyStatusPlugins(ctx, pluginRegistry, proxiesWithReports)
```
sam-heilbron, line 196: We now invoke:
- ApplyStatusPlugins
- SyncStatus
- SyncRouteStatus

Is this the desired breakdown of responsibilities?

projects/gateway2/translator/plugins/plugins.go
lgadban: PostTranslationPlugins currently run post gateway translation but before xds translation and do not receive the reports from xds translation.

We can change this if desired but impacts the current lifecycle hooks.

projects/gateway2/reports/reporter.go
lgadban: this is simply a reminder that the current name is not great and should be changed but don't want to add that diff to this PR as it is completely unrelated. (It was more relevant when trying to use kube style statuses)
No deprecation would be needed, it's not publicly used etc

projects/gateway2/xds/xds_syncer.go
lgadban: currently, yes.
ApplyStatusPlugins will be responsible for policy status
SyncStatus can be renamed to SyncGatewayStatus and is responsible for kube `Gateway` resources
SyncRouteStatus is responsible `HTTPRoute` status

The latter 2 are core to the GWAPI translation and likely won't be plugins for the near future.
We can wrap these calls in a function or something to show the intent more but not sure how helpful it is right now. Ideally we can take a more holistic look later

projects/gateway2/xds/xds_syncer.go
lgadban: to be honest I'm not sure what the intent for these was. AFAICT It originally was copy/paste from envoy_translator_syncer and I don't see any plugins actually use Messages during a codesearch. 

What are they used for?
I can add this line back in but it is irrelevant to the ggv2 implementation at this point.

projects/gateway2/translator/routeutils/utils.go
lgadban: since it interacts with a `client.Object` it's not particularly relevant to the edge code; see https://github.com/solo-io/gloo/blob/route-option-status/projects/gateway/pkg/translator/config_source.go#L97

this is in a util function because initially i was also tracking HTTPRoute objects in addition to RouteOptions (i.e. it was called from more than 1 place and made sense to pull out into a separate util file)

projects/gloo/pkg/translator/route_config.go
lgadban: fixed

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
lgadban: Yes, I completely agree

ilackarms (approved):

sam-heilbron (approved):

PR #9346, opened 2024-04-10
kube2e: resolve recent test flakes

sam-heilbron:
# Description
Fix recent kube2e test failures.

# Context
### Timeout
Nightly tests failed:
- [Internal Slack link](https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1712728408782959)
- [Github action logs](https://github.com/solo-io/gloo/actions/runs/8626220304/job/23644076229)

In https://github.com/solo-io/gloo/pull/9335 we added retry logic to the curl request mechanism, and identified that this problematic test required retries. However, we didn't also bump the total timeout for the request, so the retries aren't being respected.

### Other Failures
While exploring this code, and observing other failures I also made the following changes:
- I noticed that we were not retrying on `connrefused` so I added curl support for that.
- We have tests that assert we need to define a ConnectionTimeout, but [we added a default for that](https://github.com/solo-io/gloo/blob/main/pkg/utils/requestutils/curl/request.go#L24) so the comment is no longer accurate, as a result, I removed the comment in this file
- We have tests that set a 5 second connection timeout, and 15 second eventually, which means that we are more likely to flake, since the Eventually only provides time for 3 attempts. I reduced the timeout to 1 second in this file, since requests that take longer than that should likely be evaluated

### TCP Gateway Failures
The TCP tests failed with:
```
  [FAILED] in [It] - /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:1348 @ 04/10/24 16:51:15.375
  pod "http-echo" deleted
  << Timeline

  [FAILED] Timed out after 30.001s.
  The function passed to Eventually failed at /home/runner/work/gloo/gloo/test/kube2e/helper/curl.go:127 with:
  Expected
      <string>: waiting for reply
  to contain substring
      <string>: "hostname":"gateway-proxy"
  In [It] at: /home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:1348 @ 04/10/24 16:51:15.375

  Full Stack Trace
    github.com/solo-io/gloo/test/kube2e/gateway_test.glob..func1.11.3.2()
    	/home/runner/work/gloo/gloo/test/kube2e/gateway/gateway_test.go:1348 +0x29b
```

I ran the tests, using WAIT_ON_FAIL to halt the cluster on failure. Then I recreated the request that was being made:
```
 kubectl exec -n gloo-system -i testserver --  curl -v --connect-timeout 1 --max-time 1 --no-retry-connrefused http://gateway-proxy:8000/
```
it returned:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 10.96.98.106...
* TCP_NODELAY set
* Connected to gateway-proxy (10.96.98.106) port 8000 (#0)
> GET / HTTP/1.1
> Host: gateway-proxy:8000
> User-Agent: curl/7.58.0
> Accept: */*
>
< HTTP/1.1 200 OK
< X-Powered-By: Express
< Content-Type: application/json; charset=utf-8
< Content-Length: 252
< ETag: W/"fc-GU7B+DOFeSnQI2GFaTJXEyC8Log"
< Date: Wed, 10 Apr 2024 17:57:21 GMT
< Connection: keep-alive
<
{ [252 bytes data]
100   252  100   252    0     0    992      0 --:--:-- --:--:-- --:--:--  1032
* Connection #0 to host gateway-proxy left intact
{"path":"/","headers":{"host":"gateway-proxy:8000","user-agent":"curl/7.58.0","accept":"*/*"},"method":"GET","body":{},"fresh":false,"hostname":"gateway-proxy","ip":"::ffff:10.244.0.77","ips":[],"protocol":"http","query":{},"subdomains":[],"xhr":false}%
```
To me this demonstrated that the request was valid, but our test code to perform the assertion was not. Since we are returning data, I just used the utility that asserts a given response (which ALL our other test code uses).

## Glooctl Test
Nightly tests failed:
- [Internal Slack link](https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1712728408782959)
- [Github action logs](https://github.com/solo-io/gloo/actions/runs/8626220304/job/23644078865)

_This test hasn't been AS flakey, so I've decided to address it in a follow-up PR after monitoring the continued pain it provides_

## Interesting decisions
-

## Testing steps
- I confirmed that CI passes

Because i'm not changing live code, and only modified test code, I was able to streamline testing this locally by creating a kind cluster, and then using a previously released version:
```
KUBE2E_TESTS=gateway VERSION=1.17.0-beta17  TEAR_DOWN=true WAIT_ON_FAIL=1 make run-kube-e2e-tests
```

## Notes for reviewers
-

# Checklist:


pkg/utils/requestutils/curl/request.go
```diff
@@ -90,9 +90,9 @@ func (c *requestConfig) generateArgs() []string {
 	if c.headersOnly {
 		args = append(args, "-I")
 	}
-
-	// We prefer to be explicit, and always set the method
-	args = append(args, "--request", c.method)
+	if c.method != "" {
```
sam-heilbron, line 93: Given that this is a common utility, I opted to make it as minimal as possible, and default to it being unset, so that users must set it if they want it.

This also mirrors https://github.com/solo-io/gloo/blob/5d9e6c20fe69131e40585b0fe58b81439f2a1062/test/kube2e/helper/curl.go#L162C2-L164C3

jenshu (approved):

pkg/utils/requestutils/curl/request.go
sam-heilbron: Yeah the linked code was the "previous" implementation. I found/decided that setting the more verbose option was easier to reason through.

I also figured that if the user sets the value (even if it's to get) we should just be transparent and reflect that in the curl request, so there is no confusion. Wdyt?

sheidkamp: I like this way better. "Get" seems like a little bit of a magic constant in the linked code and unnecessary, and I think `--request` also scans faster than `-X`

pkg/utils/requestutils/curl/request.go
```diff
@@ -112,6 +112,11 @@ func (c *requestConfig) generateArgs() []string {
 	if c.retryMaxTime != 0 {
 		args = append(args, "--retry-max-time", fmt.Sprintf("%d", c.retryMaxTime))
 	}
+	if c.retryConnectionRefused {
+		args = append(args, "--retry-connrefused")
+	} else {
+		args = append(args, "--no-retry-connrefused")
+	}
```
nfuden, line 119: nit: This does nothing as only the last instance of an arg details its behavior:
see
curl hi == curl --verbose --no-verbose
and
curl hi --verbose == curl hi --verbose --no-verbose --verbose

This can be further seen in https://github.com/curl/curl/blob/4f15443d2c202cdd74dd2ca4112cea6f3a9635da/src/tool_getparam.c#L1250-L1255 to recheck how this is used in the source
```suggestion
```

sam-heilbron:
https://github.com/solo-io/gloo/actions/runs/8635765714/job/23674275293?pr=9346 helm tests passed, but no signal was created to notify the action is succeeded. So I am going to cancel it and retry.

nfuden (approved):
lgtm 1 nit suggestion. Approving per conversations blocking merge

nfuden (approved):

sheidkamp (approved):

pkg/utils/requestutils/curl/option.go
```diff
@@ -112,10 +115,20 @@ func WithRetries(retry, retryDelay, retryMaxTime int) Option {
...
+	return func(config *requestConfig) {
+		config.retryConnectionRefused = retryConnectionRefused
+	}
+}
+
 // WithoutRetries returns the Option to disable retries for the curl request
 func WithoutRetries() Option {
 	return func(config *requestConfig) {
 		WithRetries(0, -1, 0)(config)
+		WithRetryConnectionRefused(false)
```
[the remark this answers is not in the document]
sam-heilbron: I opted to keep retryOnConnectionRefused and the other retry properties to be separate. So I wanted to be sure that disabling retries meant that you turned it all off. I'm open to saying we should just have a single WithRetries that exposes all the retry options you as a dev may want

sam-heilbron:
bulldozer

PR #9347, opened 2024-04-10
Use kube-builder for GatewayParameters CRD

yuval-k:
suggestion: use kube-builder for CRD generation.
pros: Kube builder generation is simpler, faster, and has more features than skv2 CRD generation.
cons: need to figure out docs integration


projects/gateway2/api/doc.go
```diff
@@ -0,0 +1 @@
+package api
```
jenshu, line 1: is this file needed?

projects/gateway2/api/v1alpha1/kube.go
```diff
@@ -0,0 +1,265 @@
+package v1alpha1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Kubernetes autoscaling configuration.
+type Autoscaling struct {
```
jenshu, line 8: i guess we can remove autoscaling for now if we don't support it yet

projects/gateway2/docs/conf/config.json
```diff
@@ -0,0 +1,28 @@
+{
```
jenshu, line 1: can you add a readme on what the files in this directory are used for, and how/when to edit them?

projects/gateway2/deployer/merge.go
```diff
@@ -52,57 +50,65 @@ func deepMergeSlices[T any](dst, src []T) []T {
...
 		return src
 	}
 
-	dstKube := dst.Spec.GetKube()
-	srcKube := src.Spec.GetKube()
+	dstKube := dst.Spec.Kube
+	srcKube := src.Spec.Kube
 
-	dstKube.EnvoyContainer = deepMergeEnvoyContainer(dstKube.GetEnvoyContainer(), srcKube.GetEnvoyContainer())
+	dstKube.EnvoyContainer = deepMergeEnvoyContainer(dstKube.EnvoyContainer, srcKube.EnvoyContainer)
```
jenshu, line 74: can we use getters for all the fields, to avoid potential nil panics?

projects/gateway2/deployer/merge.go
```diff
@@ -52,57 +50,65 @@ func deepMergeSlices[T any](dst, src []T) []T {
...
-			break
-		}
-		dstWorkload = deepMergeDeploymentWorkloadType(dstWorkload, srcWorkload)
-	default:
-		// TODO(jbohanon) log or something? Shouldn't happen unless a new type is added
-		break
-	}
+	dst.Spec.Kube = deepMergeDeploymentWorkloadType(dstKube, srcKube)
+
+	// TODO: Figure out ONEOF behavior
```
jenshu, line 95: is this TODO still needed?

projects/gateway2/generate.go
```diff
@@ -1,69 +1,8 @@
...
 
-	"github.com/solo-io/skv2/codegen"
-	"github.com/solo-io/skv2/codegen/model"
-	"github.com/solo-io/skv2/codegen/skv2_anyvendor"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-)
-
-//go:generate go run ./generate.go
 func main() {
-	log.Println("starting generate for k8s gateway controller")
```
lgadban, line 14: we should replicate this behavior somehow -- we need logging in the generate process that shows progress etc. for troubleshooting purposes

maybe we `go run` a helper file that then runs the `controller-gen` process etc

changelog/v1.17.0-beta36/kubebuilder.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
```
jenshu, line 2: changelog needs to move to 1.18
and should probably be user facing due to gwp api changes

projects/gateway2/api/doc.go
EItanya: Not sure, @yuval-k

install/helm/gloo/crds/gateway.gloo.solo.io_gatewayparameters.yaml
```diff
@@ -0,0 +1,3851 @@
...
+    singular: gatewayparameters
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            description: |-
+              APIVersion defines the versioned schema of this representation of an object.
```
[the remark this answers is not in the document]
EItanya: I honestly have no idea, is it important?

projects/gateway2/docs/conf/config.json
EItanya: This stuff is autogenerated, I don't think we need it

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
```diff
@@ -0,0 +1,510 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=gateway-gloo,shortName=gwp
+// +kubebuilder:subresource:status
+type GatewayParameters struct {
```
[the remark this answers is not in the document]
EItanya: Manually created

projects/gateway2/deployer/merge.go
EItanya: I added getters manually where I could, but in general we're not generating the code anymore so there are no more getters to use unless we add them.

projects/gateway2/generate.go
EItanya: To be honest that seems like a rabbit hole

projects/gateway2/deployer/merge.go
EItanya: Yes, we need to figure out how we want to handle `oneof` given that we don't have proto anymore. One way could be via `cel` validation. Gateway API handles it with enums that match to fields. @yuval-k @ilackarms @lgadban  do y'all have input on that?

install/helm/gloo/crds/gateway.gloo.solo.io_gatewayparameters.yaml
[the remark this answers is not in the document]
sam-heilbron: What we had seen in the past was the CRDs grew (the Proxy CRD in particular) to a size that was too large for etcD. https://github.com/solo-io/gloo/blob/main/test/kube2e/helm/helm_test.go#L424 was created to identify this in the future. So while I agree that the descriptions expand the size of the CRD, I _think_ that test shows we won't hit size limits with this. If we do, then we may have to solve the description challenge, but for now I don't think it's necessary

projects/gateway2/api/external/kubernetes/README.md
```diff
@@ -1,42 +0,0 @@
-## Kubernetes Protos in Gloo Gateway
```
sam-heilbron, line 1: It kind of feels like this README is actually fairly useful, but we need to update it with the context from this PR: ie why we use Go types and what implication that has for future work

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
```diff
@@ -0,0 +1,510 @@
+package v1alpha1
```
sam-heilbron, line 1: How will we "publish" these types to solo-apis? Does that mechanism need to be updated as well, or do we get that for free?

projects/gateway2/api/v1alpha1/kube.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// A container image. See https://kubernetes.io/docs/concepts/containers/images
+// for details.
+type Image struct {
+	// The image registry.
+	// +kubebuilder:validation:Optional
```
sam-heilbron, line 11: Do we want to add this into 1.17? We had held off from adding this to our existing APIs to avoid rushing in additional validation late in the cycle that could impact users. But this is a new API so I can see an argument for being more aggressive with adding it. wdyt?

projects/gateway2/api/v1alpha1/kube.go
```diff
@@ -0,0 +1,196 @@
...
+	Digest string `json:"digest,omitempty"`
+	// The image pull policy for the container. See
+	// https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy
+	// for details.
+	// +kubebuilder:validation:Optional
+	PullPolicy corev1.PullPolicy `json:"pull_policy,omitempty"`
+}
+
+func (in *Image) GetRegistry() string {
```
sam-heilbron, line 29: I wonder if a comment here, or a mention in the README about how we don't have `getters` for free could provie some context to users. I imagine the github.com/saiskee/gettercheck doesn't catch this?

sam-heilbron (commented):
looking good!

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
jenshu: we currently don't publish this to solo-apis, so that will need to be a separate set of work

projects/gateway2/deployer/merge.go
lgadban: Did some poking around, IMO we should follow the K8s convention
i.e. use a type field that is effectively a string enum

Looking at GwApi, I see examples of simple CEL validation to ensure that fields for a specific type are not used incorrectly.

That being said, for now, we only support a single type here, which is workload of type `Deployment` correct?
So the validation etc. can be punted for now

However I think I am missing why this TODO is only in `merge.go` -- assuming it was just the first place you saw the usage of the `oneof` proto?

projects/gateway2/api/v1alpha1/kube.go
EItanya: These are very very light validations. I do think we should get as much validation in as soon as possible, because backfilling validation is very very hard.

install/helm/gloo/crds/gateway.gloo.solo.io_gatewayparameters.yaml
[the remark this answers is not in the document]
jenshu: removed descriptions ece2d6a

projects/gateway2/api/doc.go
jenshu: removed

projects/gateway2/deployer/merge.go
jenshu: probably the most relevant place for a oneof/enum currently is at the top level where we accept either `kube` or `selfManaged`

projects/gateway2/api/v1alpha1/kube.go
jenshu: there's an issue for validation https://github.com/solo-io/solo-projects/issues/6014

in the interest of merging this soon, i think the existing minimal validations are good enough

projects/gateway2/deployer/merge.go
jenshu: added `XValidation` tag for the top-level oneof 79d4321b6589c96a5df8559fd3b1c01f91722935

projects/gateway2/deployer/merge.go
jenshu: updated to use getters everywhere

projects/gateway2/generate.go
jenshu: @lgadban can we resolve / handle this later?

projects/gateway2/api/v1alpha1/kube.go
jenshu: updated readme 6163db5

projects/gateway2/api/external/kubernetes/README.md
jenshu: incorporated some of this info (updated version) into the api readme 6163db5

projects/gateway2/api/README.md
```diff
@@ -1,19 +1,32 @@
...
-- CRDs generated in [install/helm/gloo/crds](/install/helm/gloo/crds)
+These are the steps required to add a new CRD to be used in the Kubernetes Gateway integration:
 
-### Adding a new CRD
+1. If creating a new API version (e.g. `v1`, `v2alpha1`), create a new directory for the version and create a `doc.go` file with the `// +kubebuilder:object:generate=true` annotation, so that Go types in that directory will be converted into CRDs when codegen is run.
+    - The `groupName` marker specifies the API group name for the generated CRD. 
+2. Create a `_types.go` file in the API version directory. Following [gateway_parameters_types.go](/projects/gateway2/api/v1alpha1/gateway_parameters_types.go) as an example:
+    - Define a struct for the resource (containing the metadata fields, `Spec`, and `Status`)
+        - Tip: For spec fields, try to use pointer values when appropriate, as it makes inheritance easier (allows us to differentiate between zero values and nil).
+        - Define getters for each field, as these are not generated automatically.
```
jbohanon, line 14: This is a lot of extra work that we lose getting for free. Can we have an intermediate generation step that makes the getters? 

Otherwise I fear we will do what is shown in the test file in this PR where we just don't make the getters and deref everywhere.

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
sam-heilbron: Is this something we need to solve (https://github.com/solo-io/gloo/issues/9569) to consider this feature complete? In the past, the value of publishing these APIs was that users could import just our API and build their own controllers to produce those resources. Since this is a public API already, I don't think following the same pattern that we've done is the past is strictly necessary. Curious how you'd like to move forward

projects/gateway2/deployer/merge.go
EItanya: Awesome!!! Speaking of `XValidation` I think it's important that we put list limits in the API ASAP. It doesn't need to happen in this PR specifically because I don't want to block, but I want to make sure we track it. Backfilling length requirements is really really hard as we've seen

EItanya (dismissed):
These changes LGTM, should wait to merge until we have a passing SP PR

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
jenshu: we should definitely think about the solo-apis part, but it seems independent of this kubebuilder change

i've added a [comment](https://github.com/solo-io/gloo/issues/9569#issuecomment-2214686390) to that issue to keep in mind the differences between how we'd need to sync gwparams vs our other apis

projects/gateway2/api/README.md
jenshu: as discussed offline - would be nice to autogenerate these getters (if we could find / come up with such a tool), but the k8s types that we're importing don't have getters on them anyway so we just need to be careful with our nil checks everywhere

projects/gateway2/deployer/merge.go
jenshu: filed this for length validation https://github.com/solo-io/solo-projects/issues/6507

jenshu:
solo-projects pr that consumes this https://github.com/solo-io/solo-projects/pull/6505
passing ci except a couple failures that seem unrelated

EItanya (approved):

jbohanon (approved):

jenshu:
a

PR #9350, opened 2024-04-10
Bernie/kube dump logs

bewebi:
# Description

Fix return from `kubeExecute()` helper so that it returns the actual response from the command instead of only returning error output

## CI changes
- `kubeExecute()` writes stdout to a local buffer and returns the contents of that buffer if there is no error
  - This is called as part of `KubeDumpOnFail()` in tests and is not in a production code path

# Context

We intend to dump resources and logs on kube2e test failure, however we haven't actually been getting logs, resources etc. to date
This is because we weren't capturing the output to stdout on k8s commands, only error output, and were throwing away the output we actually care about
With this change we return the output from the k8s commands and ultimately write it to disk (test artifacts in CI) as intended

## Testing steps

This change has been used successfully in solo-io/solo-projects#6018
For example see the dumps on [this action](https://github.com/solo-io/solo-projects/actions/runs/8696156924) populated as intended
Compare with the dump on [this action](https://github.com/solo-io/solo-projects/actions/runs/8634082143) which didn't use this branch and which does not have pod logs, CRs, etc.

# Checklist:

### DIFF
--- a/test/helpers/kube_dump.go
+++ b/test/helpers/kube_dump.go
@@ -150,7 +150,7 @@ func recordKubeDump(namespaces ...string) {
 
 // recordPods records logs from each pod to _output/kube2e-artifacts/$namespace/pods/$pod.log
 func recordPods(podDir, namespace string) error {
-	pods, err := kubeList(namespace, "pod")
+	pods, _, err := kubeList(namespace, "pod")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


test/helpers/kube_dump.go
```diff
@@ -226,8 +228,13 @@ func kubeGet(namespace string, kubeType string, name string) (string, error) {
...
 
-	runError := cli.Command(context.Background(), args...).Run()
-	return runError.OutputString(), runError.Cause()
+	var outLocation threadsafe.Buffer
+	runError := cli.Command(context.Background(), args...).WithStdout(&outLocation).Run()
+	if runError != nil {
+		return runError.OutputString(), runError.Cause()
+	}
+
+	return outLocation.String(), nil
```
ashishb-solo, line 237: so my understanding here is that with this change, we return stdout on success and stderr on failure - but we ignore stdout on error and stderr on success, correct?

in my opinion, a more elegant solution would be to combine them into a single stream - or perhaps return both stderr and stdout always. is this a possibility?

ashishb-solo: that said, i'm happy to accept this as is, since it's definitely an improvement over the current behavior!

bewebi: So on 234 we'd have something like
```
return fmt.Sprintf("%s\n\nERROR OUTPUT:\n%s", outLocation.String(), runError.OutputString())
```
?

that seems reasonable enough to me

Another option would be to return both separately and have the callers write to separate files

ashishb-solo: i like both solutions. i would suggest going with whichever you feel is easiest to implement

bewebi: I like the separate files better, but it's not easier per se...I'll take a pass and abandon if it gets out of hand

ashishb-solo: personally, my favorite solution would be to combine them into a single output (just like we have in CI) so we preserve locality between stdout/stderr messages, but i had the same concern that it may not be easy to do

ashishb-solo (approved):

test/helpers/kube_dump.go
bewebi: FWIW In most cases we don't write at all if an error is returned regardless

PR #9351, opened 2024-04-10
utils: Improve kubeutils, cmdutils, errutils

sam-heilbron:
# Description

Improve utils

# Context
These are various utilities that we know we need, and they come with corresponding tests.

## KubeUtils
Slack context: https://solo-io-corp.slack.com/archives/C03MFATU265/p1712773070410799?thread_ts=1712758049.132859&cid=C03MFATU265 (@jenshu)

I will also need this when working on my e2e framework, because it relies on applying resources defined in a file.

## CmdUtils
Slack context: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1712768107313859 (@bewebi)

This will be useful when we improve the way that we export reports from a Kube cluster, either when debugging tests, or live clusters. For example, https://github.com/solo-io/gloo/pull/9327 demonstrates how this would be used in the future.

## Interesting decisions
 -

## Testing steps
- I added tests for the new functionality

## Notes for reviewers
-

# Checklist:


pkg/utils/cmdutils/file.go
```diff
@@ -0,0 +1,34 @@
+package cmdutils
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// RunCommandOutputToFile executes a Cmd, and pipes the stdout of the Cmd to the file
+// If the file does not exist on the host, it will be created
+func RunCommandOutputToFile(cmd Cmd, path string) func() error {
```
bewebi, line 10: Why not make this a member function of `Cmd`? And why does it have to return a `func() error` instead of actually executing?

pkg/utils/cmdutils/file.go
```diff
@@ -0,0 +1,34 @@
...
+			return err
+		}
+		defer func() {
+			_ = f.Close()
+		}()
+		// We intentionally do not output stderr to the output file
+		// Otherwise, the output file will contain errors that may be misleading to users
+		// For example, if a curl request failed first, and then succeeded, we do not want to
+		// populate the failures in the output file
+		return cmd.WithStdout(f).Run().Cause()
```
bewebi, line 23: We may want to consider using a `MultiWriter` so that whatever stdout had been specified still gets the output

pkg/utils/cmdutils/file_test.go
```diff
@@ -0,0 +1,52 @@
...
+		})
+
+	})
+
+})
+
+func writeActiveProcessesToFileAndAssertOutput(ctx context.Context, fileName string) {
+	GinkgoHelper()
+
+	cmdFn := RunCommandOutputToFile(
```
bewebi, line 44: nit: I like factoring out the shared logic, however it may be clearer if the function being tested is invoked directly from the spec

pkg/utils/errutils/concurrent.go
```diff
@@ -0,0 +1,37 @@
+package errutils
+
+import (
+	"sync"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
```
bewebi, line 6: nit/observation: naming an import `utilerrors` in the `errutils` package could be confusing

more humorous than something that has to be changed though

pkg/utils/errutils/concurrent.go
```diff
@@ -0,0 +1,37 @@
...
+	for err := range ch {
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) > 1 {
+		return utilerrors.NewAggregate(errs)
+	} else if len(errs) == 1 {
+		return errs[0]
+	}
```
bewebi, line 35: I don't see much of a downside to wrapping even singleton errors in an `Aggregate`
Not much of an upside either, but it would let us set the return type to `Aggregate` and therefore give the caller easy access to it's `Errors()` function

pkg/utils/errutils/concurrent_test.go
```diff
@@ -0,0 +1,61 @@
...
+			var sum = uint32(0)
+
+			goodFns := []func() error{
+				goodFn(&sum),
+				goodFn(&sum),
+				goodFn(&sum),
+			}
+			badFns := []func() error{
+				errorFn(eris.Errorf("Erroring function. ID=%d", 1)),
+				errorFn(eris.Errorf("Erroring function. ID=%d", 2)),
```
bewebi, line 40: nit: the spec says `even if 1 fails` and all we're testing below is that any error occurred, so the extra failing func is extraneous

bewebi (commented):
Good improvements!
Some nits mainly

pkg/utils/cmdutils/file.go
sam-heilbron: RE why not make this a member function of Cmd: It's an interesting idea. Ultimately, I tried to keep the Cmd interface focused, and as tightly scoped to the implementation of the exec.Cmd library

RE why return a function instead of executing: The idea here is that we can then aggregate these Cmd objects and execute them in parallel. https://github.com/solo-io/gloo/blob/sam/gg-kube-test-export-report/pkg/cliutil/export/envoy.go#L33 is an example of what I think we should work towards, as a way of capturing output

pkg/utils/cmdutils/file.go
sam-heilbron: I like the idea, though I don't exactly see how it would be implemented. Do you see that as a requirement for this PR? If not, do you want to open a follow-up PR to introduce that?

pkg/utils/cmdutils/file.go
bewebi: Seems reasonable on both fronts

It might be worthwhile to rename to `RunCommandOutputToFileFunc()` so that `RunCommandOutputToFile()` can be reserved for a wrapper function that calls `RunCommandOutputToFileFunc(cmd, path)()`

pkg/utils/errutils/concurrent.go
sam-heilbron: I had done that mostly because everywhere else we imports "errors" and it varies what package that actually is. Since this is a tightly scoped utility, im happy to remove the alias because it's clear what we're using (and it's a type that doesn't exist in other libraries)

pkg/utils/cmdutils/file.go
bewebi: Something like is done [here](https://github.com/solo-io/gloo/blob/c8aa6c3ce3762f6215714ab6ce75df41b1f20e80/pkg/utils/cmdutils/local.go#L81)

I'm realizing the `Cmd` interface doesn't allow for this however, so it would require taking or casting to `LocalCmd` which, while possible, starts to get messy

Not too significant a hurdle, but not necessary for this PR

pkg/utils/errutils/concurrent.go
bewebi: I was more amused than anything haha

pkg/utils/cmdutils/file_test.go
bewebi: I only meant to move the invocation out of the factored out func, could have left `AssertOutput` factored out, but not a big deal either way

bewebi (approved):

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=68450fee-3fd7-4e73-b0b8-273b898baf7b build reporting failure from start

pkg/utils/cmdutils/file.go
```diff
@@ -5,7 +5,7 @@ import (
 	"path/filepath"
 )
 
-// RunCommandOutputToFileFunc executes a Cmd, and pipes the stdout of the Cmd to the file
+// RunCommandOutputToFileFunc returns a Cmd that pipes the stdout of the Cmd to the file specified.
```
bewebi, line 8:
```suggestion
// RunCommandOutputToFileFunc returns a func that runs the given Cmd pipes its stdout to the file specified.
```

pkg/utils/cmdutils/file.go
```diff
@@ -0,0 +1,34 @@
...
+		// Otherwise, the output file will contain errors that may be misleading to users
+		// For example, if a curl request failed first, and then succeeded, we do not want to
+		// populate the failures in the output file
+		return cmd.WithStdout(f).Run().Cause()
+	}
+}
+
+// FileOnHost is a helper to create a file at path even if the parent directory doesn't exist
+// in which case it will be created with ModePerm
+func fileOnHost(path string) (*os.File, error) {
```
jenshu, line 29: nit: can the func name be a bit more descriptive so it's easy to tell what it does, e.g. `getOrCreateFileOnHost`?

pkg/utils/errutils/concurrent.go
```diff
@@ -0,0 +1,35 @@
+package errutils
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/errors"
+)
+
+// AggregateConcurrent runs fns concurrently, returning a NewAggregate if there are > 1 errors
```
jenshu, line 9:
```suggestion
// AggregateConcurrent runs fns concurrently, returning a NewAggregate if there are > 0 errors
```

pkg/utils/cmdutils/local.go
```diff
@@ -19,7 +19,9 @@ var (
 
 // Command is a convenience wrapper over DefaultCmder.Command
 func Command(ctx context.Context, command string, args ...string) Cmd {
-	return DefaultCmder.Command(ctx, command, args...)
+	return DefaultCmder.Command(ctx, command, args...).
```
jenshu, line 22: is DefaultCmder ever intended to be used directly, or can we make it private and have users always call this convenience wrapper? that might reduce confusion on which one to use

bewebi (dismissed):

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -60,38 +61,58 @@ func (c *Cli) RunCommand(ctx context.Context, args ...string) error {
...
+		Run().
+		Cause()
 }
 
 func (c *Cli) Copy(ctx context.Context, from, to string) error {
-	return c.copyCmd(ctx, from, to).Run().Cause()
+	return c.RunCommand(ctx, "cp", from, to)
 }
 
 func (c *Cli) StartPortForward(ctx context.Context, options ...portforward.Option) (portforward.PortForwarder, error) {
```
jenshu, line 118: what is the StartPortForward in this file used for? it seems to not be kubectl-specific

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -60,38 +61,58 @@ func (c *Cli) RunCommand(ctx context.Context, args ...string) error {
...
+	args := append([]string{"apply", "-f", "-"}, extraArgs...)
+	return c.Command(ctx, args...).
+		WithStdin(bytes.NewBuffer(content)).
+		Run().
+		Cause()
 }
 
-func (c *Cli) deleteCmd(ctx context.Context, content []byte, extraArgs ...string) cmdutils.Cmd {
-	args := append([]string{"delete"}, extraArgs...)
+func (c *Cli) ApplyFile(ctx context.Context, fileName string, extraArgs ...string) error {
```
jenshu, line 72: nit: can you add descriptions to each of the functions in this file?

test/kube2e/gateway/gateway_test.go
```diff
@@ -1909,7 +1910,10 @@ var _ = Describe("Kube2e: gateway", func() {
 
 			var stderr threadsafe.Buffer
 
-			_ = kubeCli.ApplyCmd(ctx, []byte(yaml), "-f", "-").WithStderr(&stderr).Run()
+			_ = kubeCli.Command(ctx, "apply", "-f", "-").
```
jenshu, line 1913: could we use `kubeCli.Apply` here?

sam-heilbron: We cannot, the apply command runs the command, without overriding the Stderr location. So we have a few options:
1. Update the kubeCli to default it's output to this buffer for this test, and then undo it
2. Follow the structure I took

I opted to follow the second pattern because I think it keeps the concerns of the test pretty close together. It's easy to see where we're piping the stderr to, what command is being run, and what the corresponding behavior is

pkg/utils/kubeutils/kubectl/cli.go
sam-heilbron: I think it's pretty kubectl specific. We are using the kubernetes CLI under the hood to open a port forward to a defined location. https://github.com/solo-io/gloo/pull/9335 demonstrates how we used this to fix a test flake, because it provides us default retry logic to open the port-forward that was forgotten in the previous implementation

pkg/utils/cmdutils/local.go
sam-heilbron: Sounds good, I can change

pkg/utils/kubeutils/kubectl/cli.go
jenshu: haha totally forgot it's a kubectl command

jenshu (approved):

PR #9352, opened 2024-04-11
VirtualHostOptions plugin

jbohanon:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/projects/gateway/pkg/translator/converter.go
+++ b/projects/gateway/pkg/translator/converter.go
@@ -157,6 +157,7 @@ func (r *reporterHelper) addWarning(resource resources.InputResource, err error)
 // Since virtual services and route tables are often owned by different teams, it breaks multitenancy if
 // this function cannot return successfully; thus ALL ERRORS are added to the resource reports
 func (rv *routeVisitor) ConvertVirtualService(virtualService *gatewayv1.VirtualService, gateway *gatewayv1.Gateway, proxyName string, snapshot *gloov1snap.ApiSnapshot, reports reporter.ResourceReports) []*gloov1.Route {
+	panic("converting virtualservice")
 	wrapper := &visitableVirtualService{VirtualService: virtualService}
 	return rv.visit(
 		wrapper,
--- a/projects/gateway2/translator/listener/gateway_listener_translator.go
+++ b/projects/gateway2/translator/listener/gateway_listener_translator.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"sort"
 
+	"github.com/solo-io/gloo/projects/gateway2/translator/plugins"
 	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/registry"
 	"github.com/solo-io/gloo/projects/gateway2/translator/sslutils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,7 +34,7 @@ func TranslateListeners(
 ) []*v1.Listener {
 	validatedListeners := validateListeners(gateway, reporter.Gateway(gateway))
 
-	mergedListeners := mergeGWListeners(queries, gateway.Namespace, validatedListeners, routesForGw, reporter.Gateway(gateway))
+	mergedListeners := mergeGWListeners(queries, gateway.Namespace, validatedListeners, *gateway, routesForGw, reporter.Gateway(gateway))
 	translatedListeners := mergedListeners.translateListeners(ctx, pluginRegistry, queries, reporter)
 	return translatedListeners
 }
@@ -42,10 +43,12 @@ func mergeGWListeners(
 	queries query.GatewayQueries,
 	gatewayNamespace string,
 	listeners []gwv1.Listener,
+	parentGw gwv1.Gateway,
 	routesForGw query.RoutesForGwResult,
 	reporter reports.GatewayReporter,
 ) *mergedListeners {
 	ml := &mergedListeners{
+		parentGw:         parentGw,
 		gatewayNamespace: gatewayNamespace,
 		queries:          queries,
 	}
@@ -68,6 +71,7 @@ func mergeGWListeners(
 
 type mergedListeners struct {
 	gatewayNamespace string
+	parentGw         gwv1.Gateway
 	listeners        []*mergedListener
 	queries          query.GatewayQueries
 }
@@ -176,6 +180,16 @@ func (ml *mergedListeners) translateListeners(
 	var listeners []*v1.Listener
 	for _, mergedListener := range ml.listeners {
 		listener := mergedListener.translateListener(ctx, pluginRegistry, queries, reporter)
+
+		// run listener plugins
+		for _, listenerPlugin := range pluginRegistry.GetListenerPlugins() {
+			listenerPlugin.ApplyListenerPlugin(ctx, &plugins.ListenerContext{
+				Gateway:    &mergedListener.parentGateway,
+				GwListener: &mergedListener.listener,
+				Reporter:   reporter,
+			}, listener)
+		}
+
 		listeners = append(listeners, listener)
 	}
 	return listeners
@@ -189,6 +203,7 @@ type mergedListener struct {
 	httpsFilterChains []httpsFilterChain
 	listenerReporter  reports.ListenerReporter
 	listener          gwv1.Listener
+	parentGateway     gwv1.Gateway
 
 	// TODO(policy via http listener options)
 }
--- a/projects/gateway2/translator/plugins/plugins.go
+++ b/projects/gateway2/translator/plugins/plugins.go
@@ -38,6 +38,25 @@ type RoutePlugin interface {
 	) error
 }
 
+type ListenerContext struct {
+	// top-level Gateway
+	Gateway *gwv1.Gateway
+	// gw Listener being processed
+	GwListener *gwv1.Listener
+	// Reporter for the correct ParentRef associated with this Listener
+	Reporter reports.Reporter
+}
+type ListenerPlugin interface {
+	Plugin
+
+	// ApplyListenerPlugin is called for each Listener in a Gateway
+	ApplyListenerPlugin(
+		ctx context.Context,
+		routeCtx *ListenerContext,
+		outputListener *v1.Listener,
+	) error
+}
+
 type PostTranslationContext struct {
 	// TranslatedGateways is the list of Gateways that were generated in a single translation run
 	TranslatedGateways []TranslatedGateway
--- a/projects/gateway2/translator/plugins/virtualhostoptions/query/indexers.go
+++ b/projects/gateway2/translator/plugins/virtualhostoptions/query/indexers.go
@@ -1,0 +1,49 @@
+package query
+
+import (
+	"fmt"
+
+	solokubev1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1/kube/apis/gateway.solo.io/v1"
+	"github.com/solo-io/gloo/projects/gateway2/wellknown"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	VirtualHostOptionTargetField = "virtualHostOption.targetRef"
+)
+
+func IterateIndices(f func(client.Object, string, client.IndexerFunc) error) error {
+	return f(&solokubev1.VirtualHostOption{}, VirtualHostOptionTargetField, virtualHostOptionTargetRefIndexer)
+}
+
+func virtualHostOptionTargetRefIndexer(obj client.Object) []string {
+	rtOpt, ok := obj.(*solokubev1.VirtualHostOption)
+	if !ok {
+		panic(fmt.Sprintf("wrong type %T provided to indexer. expected gateway.solo.io.VirtualHostOption", obj))
+	}
+
+	var res []string
+	targetRef := rtOpt.Spec.GetTargetRef()
+	if targetRef == nil {
+		return res
+	}
+	if targetRef.GetGroup() != gwv1.GroupName {
+		return res
+	}
+	if targetRef.GetKind() != wellknown.HTTPRouteKind {
+		return res
+	}
+
+	ns := targetRef.GetNamespace().GetValue()
+	if ns == "" {
+		ns = rtOpt.GetNamespace()
+	}
+	targetNN := types.NamespacedName{
+		Namespace: ns,
+		Name:      targetRef.GetName(),
+	}
+	res = append(res, targetNN.String())
+	return res
+}
--- a/projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
+++ b/projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
@@ -1,0 +1,39 @@
+package query
+
+import (
+	"context"
+
+	solokubev1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1/kube/apis/gateway.solo.io/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type VirtualHostOptionQueries interface {
+	// Populates the provided VirtualHostOptionList with the VirtualHostOption resources attached to the provided Gateway.
+	// Note that currently, only VirtualHostOptions in the same namespace as the Gateway can be attached.
+	GetVirtualHostOptionsForGateway(ctx context.Context, gw *gwv1.Gateway, list *solokubev1.VirtualHostOptionList) error
+}
+
+type virtualHostOptionQueries struct {
+	c client.Client
+}
+
+func NewQuery(c client.Client) VirtualHostOptionQueries {
+	return &virtualHostOptionQueries{c}
+}
+
+func (r *virtualHostOptionQueries) GetVirtualHostOptionsForGateway(ctx context.Context, gw *gwv1.Gateway, list *solokubev1.VirtualHostOptionList) error {
+	nn := types.NamespacedName{
+		Namespace: gw.Namespace,
+		Name:      gw.Name,
+	}
+	return r.c.List(
+		ctx,
+		list,
+		client.MatchingFieldsSelector{Selector: fields.OneTermEqualSelector(VirtualHostOptionTargetField, nn.String())},
+		client.InNamespace(gw.GetNamespace()),
+	)
+}
--- a/projects/gateway2/translator/plugins/virtualhostoptions/query/query_test.go
+++ b/projects/gateway2/translator/plugins/virtualhostoptions/query/query_test.go
@@ -1,0 +1,215 @@
+package query_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	sologatewayv1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1"
+	solokubev1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1/kube/apis/gateway.solo.io/v1"
+	gwscheme "github.com/solo-io/gloo/projects/gateway2/controller/scheme"
+	"github.com/solo-io/gloo/projects/gateway2/translator/plugins/virtualhostoptions/query"
+	"github.com/solo-io/gloo/projects/gateway2/wellknown"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	corev1 "github.com/solo-io/skv2/pkg/api/core.skv2.solo.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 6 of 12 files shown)


projects/gateway/pkg/translator/converter.go
```diff
@@ -157,6 +157,7 @@ func (r *reporterHelper) addWarning(resource resources.InputResource, err error)
 // Since virtual services and route tables are often owned by different teams, it breaks multitenancy if
 // this function cannot return successfully; thus ALL ERRORS are added to the resource reports
 func (rv *routeVisitor) ConvertVirtualService(virtualService *gatewayv1.VirtualService, gateway *gatewayv1.Gateway, proxyName string, snapshot *gloov1snap.ApiSnapshot, reports reporter.ResourceReports) []*gloov1.Route {
+	panic("converting virtualservice")
```
jbohanon, line 160: oops remove this
```suggestion
```

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
```diff
@@ -0,0 +1,39 @@
...
+func (r *virtualHostOptionQueries) GetVirtualHostOptionsForGateway(ctx context.Context, gw *gwv1.Gateway, list *solokubev1.VirtualHostOptionList) error {
+	nn := types.NamespacedName{
+		Namespace: gw.Namespace,
+		Name:      gw.Name,
+	}
+	return r.c.List(
+		ctx,
+		list,
+		client.MatchingFieldsSelector{Selector: fields.OneTermEqualSelector(VirtualHostOptionTargetField, nn.String())},
+		client.InNamespace(gw.GetNamespace()),
```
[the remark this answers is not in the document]
jbohanon: I am not sure why this restriction is imposed. It exists on the RouteOptions queries, and it was brought over as it was there. If this is not needed on VirtualHostOptions queries, I can remove it and look into reference grants.

@lgadban do you have more context on the namespace restriction?

projects/gateway2/translator/plugins/virtualhostoptions/query/indexers.go
```diff
@@ -0,0 +1,49 @@
...
+	targetRef := rtOpt.Spec.GetTargetRef()
+	if targetRef == nil {
+		return res
+	}
+	if targetRef.GetGroup() != gwv1.GroupName {
+		return res
+	}
+	if targetRef.GetKind() != wellknown.HTTPRouteKind {
+		return res
+	}
```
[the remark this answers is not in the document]
jbohanon: Good catch; will fix

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
```diff
@@ -0,0 +1,39 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+type VirtualHostOptionQueries interface {
+	// Populates the provided VirtualHostOptionList with the VirtualHostOption resources attached to the provided Gateway.
+	// Note that currently, only VirtualHostOptions in the same namespace as the Gateway can be attached.
+	GetVirtualHostOptionsForGateway(ctx context.Context, gw *gwv1.Gateway, list *solokubev1.VirtualHostOptionList) error
```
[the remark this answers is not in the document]
jbohanon: It looks like it was done this way because the controller-runtime client's `List` accepts the list pointer as an output parameter. I agree it's not needed so I will restructure this to return the list

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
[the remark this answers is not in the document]
lgadban: Enforcing same namespace was the simplest possible solution (without requiring reference grants) and as we are not merging Route/VHost options, there can only be one policy object attached, so it seems reasonable.

We can add reference grants for cross-namespace if we need to, but i think it is lower priority than e.g. adding status reporting for VHostOption resources

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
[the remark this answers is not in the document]
lgadban: Yes this was done to match the `List` semantics. If we change to instantiate the List inside the query, can we also update the RouteOption query to do the same? Would prefer them to be consistent

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var optToUse *solokubev1.VirtualHostOption
+	// If there is not a section name or the specified section name matches our listener, apply the vhost options
+	if targetedOpt, ok := optsWithSectionName[string(listenerCtx.GwListener.Name)]; ok {
+		optToUse = targetedOpt
+	} else if len(optsWithoutSectionName) == 1 {
+		optToUse = optsWithoutSectionName[0]
+	}
+	var vhs []*v1.VirtualHost
+	switch outListener.GetListenerType().(type) {
```
lgadban, line 73: Possible nit: I'm fairly certain we are only rendering out aggregateListeners; we may want to codify that and error if we encounter any of other type. The defensiveness here is useful but might just be more confusing to build in support for types that may never be used

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestRouteOptionsPlugin(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "RouteOptions Plugin Suite")
```
lgadban, line 12: leftover RouteOptions

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
```diff
@@ -0,0 +1,128 @@
...
+	ctx context.Context,
+	listenerCtx *plugins.ListenerContext,
+	outListener *v1.Listener,
+) error {
+
+	// attachedOptions represents all VirtualHostOptions targeting the Gateway on which this listener resides
+	attachedOptions := getAttachedVirtualHostOptions(ctx, listenerCtx.Gateway, p.vhOptQueries)
+	if len(attachedOptions) == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
jbohanon: Since we are selecting for a single options here, by adding this selection logic into the query it no longer makes sense for the query to return a List. @lgadban does it still make sense to have the queries match format between RouteOptions and VirtualHostOptions given this deviation, or have I misinterpreted the comment here?

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
jbohanon: Good call, added an error for anything but an aggregate listener

TODO: test that error

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_suite_test.go
jbohanon: Thanks; removed

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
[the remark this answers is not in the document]
EItanya: My question here is slightly different I think. I'm not arguing that we should return a single `opt` here, but rather all matching opts considering the entirety of the matching logic. The plugin can decide how to update statuses etc, but I don't think half of the selection logic should be done by the query engine, and the other half by the plugin.

projects/gateway2/translator/plugins/virtualhostoptions/query/query_test.go
```diff
@@ -0,0 +1,198 @@
...
+			TargetRef: &corev1.PolicyTargetReferenceWithSectionName{
+				Group:     gwv1.GroupVersion.Group,
+				Kind:      wellknown.GatewayKind,
+				Name:      "test",
+				Namespace: wrapperspb.String("default"),
+			},
+			Options: &v1.VirtualHostOptions{},
+		},
+	}
+}
```
lgadban, line 140: should we add a test for sectionName being present?

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin_test.go
```diff
@@ -0,0 +1,211 @@
...
+			Options: &v1.VirtualHostOptions{
+				Retries: &retries.RetryPolicy{
+					RetryOn:    "5xx",
+					NumRetries: 5,
+				},
+			},
+		},
+	}
+}
+
```
lgadban, line 167: should we add a test for sectionName being present?

lgadban (commented):
mostly LGTM, a few comments/questions

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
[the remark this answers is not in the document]
lgadban: I think I had the same concern but I believe it has been addressed in a more recent commit.

Current state is the field used for selection during Listing is targetRef but only the NamespacedName of the `Gateway`. Then the sectionName processing is done within the query, post-cache List.

I suppose there is theoretically another layer of indexing which could be used List attached policy by sectionName, but that would require Listing per Listener.

lgadban:
Worth calling out a couple of things:
* the current implementation is a Listener plugin, but it's really in the context of a gloov1.Listener, not a gwapi.Listener

Separately, there is a (somewhat) similar naivety in this implementation (similar to the current state of RouteOptions) where:
* iterate for every gloov1.Listener
* run a new query for VHostOpts attached to a parent gwapi.Gateway
* results in a new DeepCopy of the gwapi.Gateway, regardless of how many gloov1.Listeners are from the same parent gwapi.Gateway

EItanya:
> results in a new DeepCopy of the gwapi.Gateway, regardless of how many gloov1.Listeners are from the same parent gwapi.Gateway

I don't believe this is a problem with this PR specifically, but rather that DeepCopy needs to be disabled for the project which is a work item in itself

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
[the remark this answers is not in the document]
EItanya: Yea I think the indexing is fine for now, especially given that needing to index by sectionName could end up taking up more memory than it gives value. I added another comment about returning all selected options from the query though because I don't think it's the responsibility of the query to decide how many options is the right number.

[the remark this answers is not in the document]
jbohanon: Agreed. I made modifications to the query to keep the selection logic but defer the discernment to the plugin. Is this a more appropriate separation of concerns?

projects/gateway2/translator/listener/gateway_listener_translator.go
```diff
@@ -178,6 +182,16 @@ func (ml *mergedListeners) translateListeners(
 	var listeners []*v1.Listener
 	for _, mergedListener := range ml.listeners {
 		listener := mergedListener.translateListener(ctx, pluginRegistry, queries, reporter)
+
+		// run listener plugins
+		for _, listenerPlugin := range pluginRegistry.GetListenerPlugins() {
+			listenerPlugin.ApplyListenerPlugin(ctx, &plugins.ListenerContext{
```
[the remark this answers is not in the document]
jbohanon: Added!

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -39,6 +39,25 @@ type RoutePlugin interface {
 	) error
 }
 
+type ListenerContext struct {
+	// top-level Gateway
+	Gateway *gwv1.Gateway
+	// gw Listener being processed
+	GwListener *gwv1.Listener
+	// Reporter for the correct ParentRef associated with this Listener
+	Reporter reports.Reporter
```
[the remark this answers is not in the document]
jbohanon: Good call; I removed the item from the context struct.

projects/gateway2/translator/plugins/virtualhostoptions/query/query_test.go
jbohanon: I added some tests, lmk what you think of the coverage.

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin_test.go
jbohanon: I added a test here, but I felt like the query tests covered a lot of the corner case, so equal coverage here would be largely redundant. wdyt?

jbohanon:
> * run a new query for VHostOpts attached to a parent gwapi.Gateway

Is it worth caching the List result from a query on the query object keyed on the gateway NN? I'm not sure how the `controller-runtime` query engine works or if that would be an over-extension of the query implementation

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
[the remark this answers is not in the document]
jbohanon: If this merges as-is I will follow up with another PR to make the RouteOptions consistent

projects/gateway2/translator/listener/gateway_listener_translator.go
```diff
@@ -185,11 +186,14 @@ func (ml *mergedListeners) translateListeners(
...
 		// run listener plugins
 		for _, listenerPlugin := range pluginRegistry.GetListenerPlugins() {
-			listenerPlugin.ApplyListenerPlugin(ctx, &plugins.ListenerContext{
+			err := listenerPlugin.ApplyListenerPlugin(ctx, &plugins.ListenerContext{
 				Gateway:    &mergedListener.parentGateway,
 				GwListener: &mergedListener.listener,
 				Reporter:   reporter,
 			}, listener)
+			if err != nil {
+				contextutils.LoggerFrom(ctx).Errorf("error in ListenerPlugin: %v", err)
```
EItanya, line 195: If these start erroring continuously this could get very noisy, is there any kind of reporting we want to do with this?

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
```diff
@@ -16,21 +15,26 @@ type VirtualHostOptionQueries interface {
...
 }
 
 type virtualHostOptionQueries struct {
 	c client.Client
 }
 
+type VirtualHostOptionsQueryResult struct {
+	OptsWithSectionName    []*solokubev1.VirtualHostOption
+	OptsWithoutSectionName []*solokubev1.VirtualHostOption
+}
```
EItanya, line 28: Ok I'm confused, why are we distinguishing between these 2?

jbohanon: An option with a section name should always override an option without a section name, but that seemed to fall under discernment re: prior conversation

projects/gateway2/translator/listener/gateway_listener_translator.go
jbohanon: It's a good point. @lgadban is this something that upcoming status reporting would handle? I'm guessing not since the current implementation of the RouteOptions still prints this error after status reporting was added? Or was that an oversight?

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
EItanya: Ahhhh I see the problem, very interesting. I may have sent you on a bit of a goose chase. What if we return them in order of specificity? That way that logic to determine can be `libraried` but it's also clear which one the caller should favor? @lgadban what do you think of that?

jbohanon: I think that could still make it difficult to determine invalid config states such as multiple options targeting with section name, or multiple options targeting the gateway but without section name. 
IMO the four valid config states are:
- No options at all
- one option attached to gateway without section name, none with section name
- one option attached to gateway with section name, none without section name
- two options attached to gateway, one without section name, one with section name

projects/gateway2/translator/listener/gateway_listener_translator.go
lgadban: The intention was to log in addition to reporting as (IMO at least) they are used for different things. Probably worth a larger discussion for the holistic approach, that being said... as we don't have VHostOption status persistence, I think it makes even more sense to log in this specific case.

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
lgadban: I am okay with the current logic as this query is specifically receiving a Listener **and a Gateway**. There should only ever be a single `VirtualHost`

If we do not want to do the sectionName/Listener logic in the query, I would say let's change the query to be `getAttachedVHostToGateway` and return all VHostOptions attached to a Gateway. This also gives us the option to report conflicted status (eventually) on the VHostOptions that are not being attached (as we should only process the earliest possible VHost)

EItanya: Why is section name treated differently, isn't it just a more specific way of targeting?

lgadban: Yeah, nevermind, I didn't see the change where we are returning the two lists.

If we keep the query as-is, we should only return a single VirtualHostOption, the most-specific. Is this what you were initially suggesting @EItanya ?

This does mean that eventually, if there are more than one at the level of specificity, we will need to e.g. pass a reporter to report on the VHostOpts that were targeted but weren't used because an earlier object existed etc

jbohanon: I guess to @EItanya's point if they are sorted by specificity (not sure what is more specific if two target the same gateway with same section name), then it makes sense to just return the sorted list, use the head of it, and report on all of the others

lgadban: that's fair but I struggle to see value of doing it that way. It would only be to not pass in a report mechanism (but we do pass that in other gateway queries, notably)

jbohanon: I guess the question is whose responsibility is it to determine what constitutes conflicting vhost options attachments, the plugin or the query

jbohanon: The decision made at the EOD meeting yesterday was to run with the suggestion of returning a sorted list in order to unblock this initial implementation. We can change this implementation detail later as needed

projects/gateway2/translator/listener/gateway_listener_translator.go
jbohanon: We will keep this logging in AT LEAST until status reporting is implemented, so it will stay in for this PR.

jbohanon:
> > * run a new query for VHostOpts attached to a parent gwapi.Gateway
> 
> Is it worth caching the List result from a query on the query object keyed on the gateway NN? I'm not sure how the `controller-runtime` query engine works or if that would be an over-extension of the query implementation

This is outside the scope of this PR

projects/gateway2/translator/plugins/virtualhostoptions/query/query.go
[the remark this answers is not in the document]
jbohanon: Support for reference grants can be assessed and added later as needed

projects/gateway2/translator/plugins/plugins.go
```diff
@@ -39,6 +39,23 @@ type RoutePlugin interface {
...
+	// gw Listener being processed
+	GwListener *gwv1.Listener
+}
+type ListenerPlugin interface {
+	Plugin
+
+	// ApplyListenerPlugin is called for each Listener in a Gateway
+	ApplyListenerPlugin(
+		ctx context.Context,
+		routeCtx *ListenerContext,
```
ilackarms, line 54: `listenerCtx`

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
```diff
@@ -0,0 +1,73 @@
...
+var (
+	ErrUnexpectedListenerType = eris.New("unexpected listener type")
+	errUnexpectedListenerType = func(l *v1.Listener) error {
+		return eris.Wrapf(ErrUnexpectedListenerType, "expected AggregateListener, got %T", l.GetListenerType())
+	}
+)
+
+func NewPlugin(gwQueries gwquery.GatewayQueries, client client.Client) *plugin {
+	return &plugin{
+		gwQueries,
```
ilackarms, line 31: nit: prefer to always use field names when constructing structs

EItanya (approved):

lgadban (approved):

jbohanon:
Paging Dr. Dozer

Issue #8527, opened 2023-07-28
E2e http_tunneling_test flake


jbohanon:
### Gloo Edge Version

1.15.x (beta)

### Describe the bug
```log
Step #8 - "run-e2e-tests":   [FAILED] in [AfterEach] - /workspace/gloo/test/e2e/http_tunneling_test.go:476 @ 07/28/23 17:34:04.421
Step #8 - "run-e2e-tests":   << Timeline
Step #8 - "run-e2e-tests": 
Step #8 - "run-e2e-tests":   [FAILED] Unexpected error:
Step #8 - "run-e2e-tests":       <*errors.errorString | 0x72bb630>: {s: "EOF"}
Step #8 - "run-e2e-tests":       EOF
Step #8 - "run-e2e-tests":   occurred
Step #8 - "run-e2e-tests":   In [AfterEach] at: /workspace/gloo/test/e2e/http_tunneling_test.go:476 @ 07/28/23 17:34:04.421
Step #8 - "run-e2e-tests": 
Step #8 - "run-e2e-tests":   Full Stack Trace
Step #8 - "run-e2e-tests":     github.com/solo-io/gloo/test/e2e_test.connectProxy.func2()
Step #8 - "run-e2e-tests":     	/workspace/gloo/test/e2e/http_tunneling_test.go:476 +0x1e3
Step #8 - "run-e2e-tests":     created by github.com/solo-io/gloo/test/e2e_test.connectProxy
Step #8 - "run-e2e-tests":     	/workspace/gloo/test/e2e/http_tunneling_test.go:470 +0xe06
```

jbohanon:
https://github.com/solo-io/gloo/pull/8487

sam-heilbron:
Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=1e9d12a3-d955-4b3d-9cae-924b85dbdeed
Error:
```
 < HTTP/1.1 200 OK
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   < content-type: text/plain; charset=UTF-8
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   < cache-control: no-cache, max-age=0
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   < x-content-type-options: nosniff
Step #1 - "run-tests": Step #8 - "run-e2e-tests":   < date: Sun, 14 Apr 2024 23:35:14 GMT
[...]

PR #9358, opened 2024-04-12
test/kube: Introduce framework

sam-heilbron:
# Description

Introduce a framework to execute end-to-end tests for Gloo Gateway

# Context

## Background
See https://github.com/solo-io/gloo/blob/sam/gg-kube-test-framework/test/kubernetes/README.md#historical-challenges

## Test Utilities
See https://github.com/solo-io/gloo/blob/sam/gg-kube-test-framework/test/kubernetes/README.md#test-utilities

## Framework
See https://github.com/solo-io/gloo/blob/sam/gg-kube-test-framework/test/kubernetes/e2e/README.md

## Additional Changes
I merged in changes from https://github.com/solo-io/gloo/pull/9366 that has its own context

## Testing steps
- New GitHub action is passing. I intentionally have not made this required, as it is used in this PR to prove the functionality. In a follow-up PR, I intend to create actual test suites that run.

## Notes for reviewers
- The `glooctl` suite is currently failing for reasons unrelated to this change. I have opened a thread in the team slack channel: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1713193713670019. Current state: This is due to distroless changes that causes the SDS image to not be pulled from quay

# Checklist:


sam-heilbron:
/kick https://github.com/solo-io/gloo/issues/8527#issuecomment-2054228707

test/kubernetes/e2e/features/deployer/gateway-parameters.yaml
```diff
@@ -0,0 +1,39 @@
...
+    envoyContainer:
+      bootstrap:
+        logLevel: debug
+        componentLogLevels:
+          upstream: debug
+          connection: trace
+      image:
+        registry: quay.io/solo-io
+        repository: gloo-envoy-wrapper
+        tag: "1.17.0-beta18"
```
nfuden, line 39: do we want this to be version locked? if so we may want some utility like https://github.com/xorcare/golden to handle toil of updates

sam-heilbron: I plan on updating this in follow up PRs. I was using this to prove to myself that it could be set, but I agree, it would be better to just assert that it could be overriden.

nfuden (dismissed):

test/kubernetes/testutils/operations/operation.go
```diff
@@ -0,0 +1,64 @@
...
+	Action() actions.ClusterAction
+
+	// Assertion returns the assertions.ClusterAssertion that will run after the Action is executed
+	// This is a function that asserts behavior of the cluster
+	Assertion() assertions.ClusterAssertion
+}
+
+// ReversibleOperation combines two Operation, that are the inverse of one another
+// We recommend that developers write tests using ReversibleOperation.
+// This is because when these are executed, they leave the cluster in the state they found it.
```
jbohanon, line 31: This is dependent on the `Do` and `Undo` being implemented and is not guaranteed by the mere structure of the framework. I think this comment implies that `ReversibleOperation`s are somehow infallibly or magically idempotent

test/kubernetes/testutils/operations/operation.go
```diff
@@ -0,0 +1,64 @@
...
+	Do   Operation
+	Undo Operation
+}
+
+var _ Operation = new(BasicOperation)
+
+// BasicOperation is an implementation of the Operation interface, with the minimal properties required
+type BasicOperation struct {
+	OpName       string
+	OpAction     func(ctx context.Context) error
```
jbohanon, line 43: Should this be defined as a `actions.ClusterAction` instead of the raw func signature?

test/kubernetes/testutils/operations/README.md
```diff
@@ -0,0 +1,17 @@
+# Operations
```
jbohanon, line 1:
```suggestion
# Operations

_This document is a work in progress_
```

test/kubernetes/testutils/actions/provider/provider.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/solo-io/gloo/test/kubernetes/testutils/cluster"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
+)
+
+// ActionsProvider is the entity that creates actions.
+// These actions are executed against a running installation of Gloo Gateway, within a Kubernetes Cluster.
+// This provider is just a wrapper around sub-providers, so it exposes methods to access those providers
+type ActionsProvider struct {
+	kubeCtlProvider *kubectl.Provider
+	glooCtlProvider glooctl.Provider
```
jbohanon, line 18: nit: these are not camel case anywhere else that I have ever seen
```suggestion
	kubectlProvider *kubectl.Provider
	glooctlProvider glooctl.Provider
```

test/kubernetes/testutils/operations/operator.go
```diff
@@ -0,0 +1,146 @@
...
+	}()
+
+	for _, op := range operations {
+		undoOperations = append(undoOperations, op.Undo)
+
+		doErr := o.executeOperation(ctx, op.Do)
+		if doErr != nil {
+			return doErr
+		}
+	}
```
jbohanon, line 118: Does it make sense to append to `undoOperations` after execution the operation returns nil? If there is an error in the execution of the operation, it seems unsafe to assume that the `Undo` can be applied cleanly

test/kubernetes/testutils/operations/operator.go
```diff
@@ -0,0 +1,146 @@
...
+	o.writeProgress(operation, "executing operation")
+	if err := action(ctx); err != nil {
+		return err
+	}
+
+	o.writeProgress(operation, "asserting operation")
+	assertion := operation.Assertion()
+	if assertion == nil {
+		// We want to make it impossible for developers to accidentally define operations that do not assert any behavior
+		// If a developer wants to provide a no-op implementation, they will, but this check ensures that is it intentional
```
jbohanon, line 135: nit:
```suggestion
		// If a developer wants to provide a no-op implementation, they will, but this check ensures that it is intentional
```

test/kubernetes/testutils/operations/operation.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/solo-io/gloo/test/kubernetes/testutils/assertions"
+)
+
+// Operation defines the properties of an operation that can be applied to a Kubernetes cluster
+// An Operation is intended to be simple, and encapsulate two concepts:
+//
+//	Action - A mutation that is applied to a cluster
+//	Assertion - An assertion that the mutation behaved appropriately
+type Operation interface {
```
jbohanon, line 16: Can we have `Operation` implement `Stringer` so that reporting e.g. progress, the Operation definition can have some impact on what information is displayed rather than just `operation.Name()`? 

For this implementation, it pretty clearly makes sense to just have the Operation stringified as the `operation.Name()`, but it may be useful.

But then again, YAGNI

test/kubernetes/testutils/assertions/assertion.go
```diff
@@ -0,0 +1,8 @@
+package assertions
+
+import "context"
+
+// ClusterAssertion is a function which asserts a given behavior at a point in time
+// If it succeeds, it will not return anything
+// If it fails, it will panic
```
jbohanon, line 7: If a user is defining a custom assertion that doesn't utilize testing frameworks that exist, this should be more strongly worded to indicate that the only valid failure mode is a panic (as opposed to, say a `log.Fatal`)

```suggestion
// If it fails, it must panic
```

test/kubernetes/testutils/actions/glooctl/export.go
```diff
@@ -0,0 +1,20 @@
...
+func (p *providerImpl) ExportReport() actions.ClusterAction {
+	p.requiresGlooGatewayContext()
+
+	return func(ctx context.Context) error {
+		p.testingFramework.Logf("invoking `glooctl export report` for Gloo Gateway installation in %s", p.glooGatewayContext.InstallNamespace)
+
+		// TODO: implement `glooctl export report`
+		// This would be useful for developers debugging tests and administrators inspecting running installations
+
+		return nil
```
jbohanon, line 18: This function should `panic("not implemented")` or at least return an error indicating such.

Alternately, un-export the function until implemented

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,169 @@
...
+
+	i.TestingFramework.Logf("TEST: %s", test.Name)
+	test.Test(ctx, i)
+}
+
+// preFailHandler is the function that is invoked if a test in the given TestInstallation fails
+func (i *TestInstallation) preFailHandler() {
+	exportReportOp := &operations.BasicOperation{
+		OpName:   "glooctl-export-report",
+		OpAction: i.Actions.GlooCtl().ExportReport(),
```
jbohanon, line 145: Having this here is confusing since this is not implemented

test/kubernetes/testutils/actions/README.md
```diff
@@ -0,0 +1,14 @@
+# Actions
+
+A [ClusterAction](./action.go) is a function that will be executed against the cluster, to mutate it's state.
```
jbohanon, line 3: nit:
```suggestion
A [ClusterAction](./action.go) is a function that will be executed against the cluster to mutate its state.
```

test/kubernetes/testutils/actions/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Actions are intended to mirror actions that users of the product can take. We group these actions in a package associated with the tools that users would rely on (`glooctl`, `kubectl`..etc).
+
+If you intend to introduce a new action, please follow this approach:
+- We want to avoid writing custom code that is _just_ used by our tests. The core action should live in a utility package ([utils](/pkg/utils), [cli-utils](/pkg/cliutil), ...etc)
+- The `ClusterAction` that you add should live in the package associated with that tool
```
jbohanon, line 9:
```suggestion
- The `ClusterAction` that you add should live in the actions package associated with that tool
```

test/kubernetes/testutils/actions/action.go
```diff
@@ -0,0 +1,8 @@
+package actions
+
+import "context"
+
+// ClusterAction is a function that will be executed against the cluster
+// If it succeeds, it will not return anything
+// If it fails, it will return an error
```
jbohanon, line 7: Same, though it's a bit more obvious here than with an `Assertion`, I think it would be useful to use comment language to advise against a `log.Fatal` or a `panic` so as to provide consistent behavior.

Or perhaps the executor of the action could recover a panic and handle it as returning an error.

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+	err := i.Operator.ExecuteOperations(context.Background(), exportReportOp)
+	if err != nil {
+		i.TestingFramework.Errorf("Failed to executed preFailHandler operation for TestInstallation (%s): %+v", i.Name, err)
+	}
+}
+
+// TestFn is a function that executes a test, for a given TestInstallation
+type TestFn func(ctx context.Context, suite *TestInstallation)
```
jbohanon, line 158: nit clarity: 
```suggestion
type TestExecutor func(ctx context.Context, suite *TestInstallation)
```

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,169 @@
...
+		OpName:   "glooctl-export-report",
+		OpAction: i.Actions.GlooCtl().ExportReport(),
+		OpAssertion: func(ctx context.Context) {
+			// This action is performed on test failure, and is not modifying the cluster
+			// As a result, there is no assertion that we perform
+		},
+	}
+	err := i.Operator.ExecuteOperations(context.Background(), exportReportOp)
+	if err != nil {
+		i.TestingFramework.Errorf("Failed to executed preFailHandler operation for TestInstallation (%s): %+v", i.Name, err)
```
jbohanon, line 153: nit: 
```suggestion
		i.TestingFramework.Errorf("Failed to execute preFailHandler operation for TestInstallation (%s): %+v", i.Name, err)
```

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,169 @@
...
+	installOperation := &operations.BasicOperation{
+		OpName:      "uninstall-gloo-gateway",
+		OpAction:    uninstallAction,
+		OpAssertion: i.Assertions.UninstallationWasSuccessful(),
+	}
+	return i.Operator.ExecuteOperations(ctx, installOperation)
+}
+
+// RunTests will execute a batch of Test against the installation
+func (i *TestInstallation) RunTests(_ context.Context, _ ...Test) {
```
jbohanon, line 125: Is this to implement an interface..? If not and this exists purely for documentation/intended use, then I think this should be removed in favor of other documentation. If it is to implement an interface, I would put that in the godoc comment with some big warnings and maybe put a static compilation check for interface implementation in this file near the type def (I didn't see one here).

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func (s *TestSuite) RegisterTestInstallation(name string, glooGatewayContext *gloogateway.Context) *TestInstallation {
+	if s.activeInstallations == nil {
+		s.activeInstallations = make(map[string]*TestInstallation, 2)
+	}
+
+	installation := &TestInstallation{
+		// Create a reference to the TestSuite, and all of it's metadata
```
jbohanon, line 59: nit:
```suggestion
		// Create a reference to the TestSuite, and all of its metadata
```

jbohanon (commented):
I got most of the way through the framework. Nothing huge here, a lot of nits and wording. Will finish in the morning

test/kubernetes/testutils/operations/operator.go
sam-heilbron: Yeah in the follow-up PR/commit I ended up changing this and clarified why. We could make it configurable in the future, but I'd rather reduce the knobs until we actually need them

test/kubernetes/README.md
```diff
@@ -0,0 +1,53 @@
...
+### Operator
+An [Operator](testutils/operations/operator.go) is a centralized tool that executes Operations against a cluster. It is intentionally unaware of _what_ those operations do, and is just used to coordinate the actions that are taken, and provide a centralized ledger for the events that occurred.
+
+## End-To-End Testing
+### Historical Challenges
+_We document the historical challenges we have experienced with writing and managing end-to-end tests as a way of avoiding making the same mistakes:_
+
+#### Not Representative of User Behaviors
+- Tests were too aware of implementation details, and would have custom behaviors built into them (ie. deleting Proxy CRs after deleting the sourcing VirtualService and Gateway).
+- Actions were taken on cluster that users couldn't do. For example, we would patch the Settings resource, but in a realistic scenario this would be done by modifying a manifest and performing a Helm upgrade.
```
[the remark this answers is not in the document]
sam-heilbron: It's a good point! I was trying less to say what users coudl/couldn't do, and more what was standard practice. I updated the phrasing a bit

test/kubernetes/e2e/README.md
```diff
@@ -0,0 +1,44 @@
+# End-to-End Testing Framework
+
+## TestSuite
+A [TestSuite](./framework.go) is the structure that manages tests running against a single Kubernetes Cluster.
```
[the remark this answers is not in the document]
sam-heilbron: Good point! i had originally contemplated calling this a TestCluster and then kind of forgot about it. Given the naming of TestInstallation, I actually like using TestCluster again, since it represents all the tests for a single cluster

test/kubernetes/e2e/framework.go
sam-heilbron: Update to remove. This was based on a conversation withNathan and we agreed some documentation was better than none. I opted to remove this function, and just added a comment above the impelmented `RunTest` function

test/kubernetes/e2e/framework.go
sam-heilbron: I intentionally did this. I wanted to show that capturing state about a cluster really should be an action that is available in the CLI. Since that doesn't yet exist, I wanted to create the piping for it. I intend as fast follow PRs to implement that function

test/kubernetes/testutils/actions/action.go
sam-heilbron: ha great minds think alike: https://github.com/solo-io/gloo/blob/sam/gg-kube-test-framework/test/kubernetes/testutils/operations/operator.go#L76

I intercept failures intentionally, so that the test authors can write assertions against them if need be.

I'm happy to update the documentation around recommendations for defining assertions

test/kubernetes/testutils/operations/operation.go
sam-heilbron: i like it!

test/kubernetes/testutils/operations/operation.go
sam-heilbron: Good call, i updated the wording to try to clarify that this is just a guideline for devs, but nothing magical

test/kubernetes/e2e/framework.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (c *TestCluster) UnregisterTestInstallation(installation *TestInstallation) {
+	delete(c.activeInstallations, installation.Name)
+}
+
+// TestInstallation is the structure around a set of tests that validate behavior for an installation
+// of Gloo Gateway.
+type TestInstallation struct {
+	*TestCluster
```
jbohanon, line 91: Why does a `TestInstallation` need to be composed from a `*TestCluster`?

jbohanon: This is especially confusing to me given the latter seems to be structured to have parentship/ownership of the former

jbohanon: as far as I can tell it is just to access the `TestingFramework` field of the parent `TestCluster`. If this is the case, could just that be passed down? At the minimum, I think we should store the reference to the `*TestCluster` as a named field to indicate that we are keeping it for reference in lieu of composing the type, which implies some sort of inheritance model.

pkg/utils/kubeutils/pods.go
```diff
@@ -36,17 +77,20 @@ func GetPodsForDeployment(
...
+	podList, err := kubeClient.CoreV1().Pods(deploy.GetNamespace()).List(ctx, *listOptions)
 	if err != nil {
 		return nil, err
 	}
 
-	pods := make([]string, len(podList.Items))
+	podNames := make([]string, 0, len(podList.Items))
 	for i := range podList.Items {
-		pods[i] = podList.Items[i].Name
+		pod := podList.Items[i]
```
jbohanon, line 87: nit: this could be inlined into the for loop setup
```go
for i, pod := range podList.Items {
```

changelog/v1.17.0-beta22/gg-kube-test-framework.yaml
```diff
@@ -0,0 +1,61 @@
...
+    issueLink: https://github.com/solo-io/solo-projects/issues/6029
+    resolvesIssue: false
+    description: >-
+      Iterate on the test framework, and introduce some basic tests for asserting behavior about provisioned proxies.
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/5940
+    resolvesIssue: false
+    description: >-
+      Provide the gloo ServiceAccount with the necessary permissions to watch the VirtualHostOptions object.
+      This would only affect users who had configured `kubeGateway.enabled=true`, an alpha feature in 1.17.
```
jbohanon, line 49: This fix has been extracted into https://github.com/solo-io/gloo/pull/9368

Also I don't think it warrants a `FIX` changelog entry since the feature hasn't been released

test/kubernetes/e2e/framework.go
sam-heilbron: Is t there an inheritance model? An installation runs in a cluster and so it should have accesss to all details of that cluster

test/kubernetes/e2e/k8sgateway/deployer_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+		It("provisions resources appropriately", func() {
+			testInstallation.RunTest(ctx, deployer.ProvisionDeploymentAndService)
+		})
+
+		It("configures proxies from the GatewayParameters CR", func() {
+			testInstallation.RunTest(ctx, deployer.ConfigureProxiesFromGatewayParameters)
+		})
+
+	})
```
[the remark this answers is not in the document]
sam-heilbron: True, but the goal was to intentionally decouple what tests are run from what installation. That way a single test can be run by N different installations. I agree we can come up with improved ways to safeguard devs misusing this pattern, but I don't see that as a requirement to the framework existing

test/kubernetes/testutils/actions/kubectl/provider.go
```diff
@@ -0,0 +1,42 @@
...
+// Provider provides a mechanism to generation operations that are performed via kubectl
+type Provider struct {
+	kubeCli *kubectl.Cli
+}
+
+func NewProvider() *Provider {
+	return &Provider{
+		kubeCli: nil,
+	}
+}
```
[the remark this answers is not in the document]
sam-heilbron: If/when this changes, I find it easier to have a standardized constructor, since it will be identifiable which code relies on it

sam-heilbron:
/kick cloudbuild reporting error but is running https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=78a0724b-fef9-4b87-ba3d-3997a1863ce5

test/kubernetes/e2e/framework.go
jbohanon: I agree that an installation should have access to the cluster that it's running inside of, I disagree that it should be _composed of_ the cluster that it is running inside of. 

It's perhaps a pretty pedantic semantic difference, but naming the field helps clarify that _this installation is running within the context of this owning cluster_ as opposed to _this installation (is)/(is synonymous with)/(is an extension of) this owning cluster_

jbohanon: Per offline discussion, this will be changed to a field on the `TestInstallation` instead of a composition inclusion

nfuden (approved):

test/kubernetes/e2e/framework.go
sam-heilbron: captured here: https://github.com/solo-io/solo-projects/issues/6047#issuecomment-2059650113

test/kubernetes/testutils/actions/glooctl/export.go
sam-heilbron: yeah that's why I originally didn't implement it as a panic. If the preFailHandler is called, that means a test failed. And if a test failed, we _should_ have the offending assertion. I don't know the exact behavior with this panicing as well, but I figure it's something that will be easily identifiable if it's an issue or not

PR #9366, opened 2024-04-15
kube/test: framework follow-ups

sam-heilbron:
# Work In Progress
Builds off of: https://github.com/solo-io/gloo/pull/9358

This PR is actually in a near-finalized state, but it is targeted for another branch. I do not want this PR to merge into that branch, but it does depend on changes from it. My plan is as follows:
1. Get the PR for the target branch merged
2. Update this PR to point to main
3. Remove the work in progress label

# Description
- Create a unified k8s gateway test suite
- Introduce a test that asserts we can curl the admin server for a provisioned proxy
- Introduce some utilities for other tests
- Fix some bugs that were unearthed by the end-to-end test

# Context
## Port Forwarding
I noticed that when we opened a port-forward to the envoy admin API, requests would fail. Upon debugging, I noticed that our utility for selecting available pods would return any pods that were running. I made two changes:
1. Allow our utility to be more configurable, so you as a developer can pass a predicate to filter out pods
2. Update our ClusterAssertion to require that pods are ready, not just running

## RBAC
When running a version of Gloo that was built off the main branch, I hit:
```
gloo-74c586d64d-rc6ql W0416 00:29:48.206527       1 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list *v1.VirtualHostOption: virtualhostoptions.gateway.solo.io is forbidden: User "system:serviceaccount:k8s-gw-deployer-test:gloo" cannot list resource "virtualhostoptions" in API group "gateway.solo.io" at the cluster scope
gloo-74c586d64d-rc6ql E0416 00:29:48.206577       1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: Failed to watch *v1.VirtualHostOption: failed to list *v1.VirtualHostOption: virtualhostoptions.gateway.solo.io is forbidden: User "system:serviceaccount:k8s-gw-deployer-test:gloo" cannot list resource "virtualhostoptions" in API group "gateway.solo.io" at the cluster scope
```
The fix I provided is not feature complete. I would expect there to be more rigorous testing to prevent this, and an analysis of the RBAC around this feature. I do not plan on doing that in this PR.

## Readiness Probe
Slack context: https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1713225744015549

When trying to curl the provisioned proxies, I noticed that the pod wasn't ready. Inspecting the code showed that our readinessProbe pointed to a port that didn't exist.

I updated the probe to mirror that of our "class" proxy deployment. 

## Helm Utility
I was building this before encountering issues with getting the tests to complete. This follows the pattern established in: https://github.com/solo-io/gloo/pull/9316. The idea is that we can have a centralized client with all the details/complexity isolated to it.

I intend to use this utility to build ClusterActions in tests, so that I can migrate our helm tests to use the new framework.

## Interesting decisions
-

## Testing steps
- I have kube e2e tests passing
- I have not introduced unit tests for the small bug fixes I introduced. If they break existing unit tests, I will fix those, and then work with the engineers assigned the original tasks to expand our tests.

## Notes for reviewers
-

# Checklist:


pkg/utils/requestutils/curl/option.go
```diff
@@ -67,10 +68,30 @@ func WithPort(port int) Option {
 // WithHost returns the Option to set the host for the curl request
 func WithHost(host string) Option {
 	return func(config *requestConfig) {
+		if host == "localhost" {
+			host = "127.0.0.1"
```
jbohanon, line 72: Should check if use ipv6 is set and set to `::1` if so. This introduces an issue where the option definition order changes behavior that I don't love, but I don't think this is acceptable as-is. 

Maybe the setter for use ipv6 needs to assert that the host is not an ipv4 address and vice versa?

jbohanon (changes requested):

PR #9373, opened 2024-04-16
Add status syncer to fix RouteOptions statuses 

npolshakova:
# Description

Statuses are back on RouteOptions with targetRefs!

```
kind: RouteOption
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"gateway.solo.io/v1","kind":"RouteOption","metadata":{"annotations":{},"name":"foo","namespace":"default"},"spec":{"options":{"extauth":{"configRef":{"name":"basic-auth","namespace":"gloo-system"}},"retries":{"numRetries":3,"perTryTimeout":"5s","retryOn":"connect-failure"},"timeout":"10s"},"targetRef":{"group":"gateway.networking.k8s.io","kind":"HTTPRoute","name":"example-route"}}}
  creationTimestamp: "2024-04-16T15:34:00Z"
  generation: 2
  name: foo
  namespace: default
  resourceVersion: "7924"
  uid: 1892c23e-9779-4130-85b4-792beba4bf6b
spec:
  options:
    extauth:
      configRef:
        name: basic-auth
        namespace: gloo-system
    retries:
      numRetries: 3
      perTryTimeout: 5s
      retryOn: connect-failure
    timeout: 10s
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: example-route
status:
  statuses:
    gloo-system:
      reportedBy: gloo-kube-gateway
      state: Accepted

```

# Tests 

The basic RouteOptions tests are based on new [test framework](https://github.com/solo-io/gloo/pull/9358) and assert statuses when the `targetRef` field is used.

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6084


lgadban:
just confirmed locally off this branch that errors are being propagated as well:
```
status:
  statuses:
    gloo-system:
      reason: "2 errors occurred:\n\t* *faultinjection.plugin: invalid abort status
        code '0', must be in range of [200,600). Route Name: http~example_com-route-0-matcher-0\n\t*
        *faultinjection.plugin: invalid abort status code '0', must be in range of
        [200,600). Route Name: http~example_com-route-1-matcher-0\n\n"
      reportedBy: gloo-kube-gateway
      state: Rejected
      subresourceStatuses:
        gloo-kube-gateway-api~default~default-gw:
          reason: "1 error occurred:\n\t* Route Error: ProcessingError. Reason: *faultinjection.plugin:
            invalid abort status code '0', must be in range of [200,600). Route Name:
            http~example_com-route-0-matcher-0; Route Error: ProcessingError. Reason:
            *faultinjection.plugin: invalid abort status code '0', must be in range
            of [200,600). Route Name: http~example_com-route-1-matcher-0\n\n"
          reportedBy: gloo-kube-gateway
          state: Rejected
```

pkg/utils/kubeutils/portforward/api_forwarder.go
```diff
@@ -129,6 +128,10 @@ func (f *apiPortForwarder) Address() string {
 	return net.JoinHostPort(f.properties.localAddress, strconv.Itoa(f.properties.localPort))
 }
 
+func (f *apiPortForwarder) LocalPort() int {
```
sam-heilbron, line 131: I like this addition, and am happy to keep it. Originally, this was just returned within the `Address` method. If we want to expose more methods to give granular access to properties, should we also expose a method for localAddress?

test/kubernetes/e2e/framework.go
```diff
@@ -62,6 +62,9 @@ func (c *TestCluster) RegisterTestInstallation(name string, glooGatewayContext *
 		// Name is a unique identifier for this TestInstallation
 		Name: name,
 
+		// Namespace is the namespace where the installation of Gloo Gateway is running
```
sam-heilbron, line 65: My hope with the gloogateway.Context (https://github.com/solo-io/gloo/blob/main/test/kubernetes/testutils/gloogateway/context.go#L3) was that it contains all the properties of the installation, and the TestInstallation can contain whatever references it needs. Instead of re-defining the properties on this new struct, what if we just made the entire gloogateway.Context available on this object? That way, as we add more properties to it, the TestInstallation gets them for free

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
```diff
@@ -1,11 +1,8 @@
 global:
   image:
     pullPolicy: IfNotPresent
-  glooRbac:
-    namespaced: true
-    nameSuffix: e2e-test-rbac-suffix
 settings:
-  singleNamespace: true
+  singleNamespace: false
```
sam-heilbron, line 5: Since this is an intentional change, and a requirement of the feature, I think it could be valuable to have a comment here indicating why we set this value, and perhaps linking any relevant github issues? Ideally it will help us avoid questions around why we set this

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
```diff
@@ -1,11 +1,8 @@
 global:
   image:
     pullPolicy: IfNotPresent
-  glooRbac:
```
sam-heilbron, line 4: Could we set this, but explicitly set it to false? That way we can add a comment about why this is required to be clusterscoped, and link the relevant issues/context

test/kubernetes/testutils/cluster/context.go
```diff
@@ -27,4 +33,21 @@ type Context struct {
 
 	// A set of clients for interacting with the Kubernetes Cluster
 	Clientset *kubernetes.Clientset
+
+	// A set of clients for interacting with the Edge resources
+	// TODO(npolshak): Add new clients as needed
+	RouteOptionClient gatewayv1.RouteOptionClient
```
sam-heilbron, line 39: I was hoping that a cluster.Context has no knowledge of Edge/Gateway concepts, it is just a kubernetes cluster with utilities. I figured that if we need a set of clients for edge purposes, those should be created per testInstallation, since clients would operate against resources of an installation

sam-heilbron (commented):
I just focused on the testing portion for this first pass. I'm happy to review the status syncing work as well, but trying to separate the two in my head

pkg/utils/kubeutils/portforward/api_forwarder.go
npolshakova: I'm leaning towards moving away from the port forward testing- it looks isn't as stable in ci: https://github.com/solo-io/gloo/actions/runs/8739466981/job/23980869526  

I'm moving towards creating a `curl` deployment as part of the setup and sending requests from there: https://github.com/solo-io/gloo/pull/9373/commits/dae5bbf755eda66fc87b3a3dff3d046cdbfb44f0

test/kubernetes/e2e/framework.go
npolshakova: But then we're tying the `gloogateway.Context` to the installation, so for classic edge apis you need to still set this value in order to get the namespace? Maybe we should rename the context?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
npolshakova: Do you think we need a docs change here? I created a related issue, but it's not specifically calling out that we need to document this as a requirement for Gloo Gateway: https://github.com/solo-io/solo-projects/issues/6058

test/kubernetes/testutils/cluster/context.go
npolshakova: I think I can't move it into the test itself because we invoke the test with: 
```
testInstallation.RunTest(ctx, route_options.ConfigureRouteOptionsWithTargetRef)
```

So it needs to be part of the testInstallation, it can't just be initialized on the test itself. Does that sound reasonable?

test/kubernetes/e2e/framework.go
sam-heilbron: I'm a bit confused, an installation is a running instance of Gloo. I figured it has 2 types of things: metadata about the install (gloogateway.Context), utilities to help tests interact with the install (TestInstalllation).

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
sam-heilbron: yeah great call! I like having our tests as the last line of defense, since sometimes we forget to publicly document behaviors, but I really like putting this in our user-facing docs too

test/kubernetes/testutils/cluster/context.go
sam-heilbron: `So it needs to be part of the testInstallation, it can't just be initialized on the test itself. Does that sound reasonable?`

Oh I agree with this. What I was proposing was that the TestInstallation have a set of clients. So tests can use them

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -115,35 +119,33 @@ var ConfigureRouteOptionsWithFilterExtenstion = e2e.Test{
...
-					Body:       "fault filter abort",
-					StatusCode: http.StatusTeapot,
-				}))
+			g.Expect(resp).To(ContainSubstring("fault filter abort"))
+			g.Expect(resp).To(ContainSubstring("HTTP/1.1 418"))
 		}, "10s", "1s", "curl should eventually return fault injection response").Should(Succeed())
 	}
 }
+
+func curl(ctx context.Context, ns, fromDeployment, fromContainer, fdqnAddr, host string) (string, error) {
```
sam-heilbron, line 136: I had imagined that we could invert the relationship here. Instead of having curl information, and using the kubecli, I thought that a kubeCli command that we expose is "ExecCurl" and we pass it the curlOptions that we want to exec. Then the details how how we do it are isolated to the kubeCli

npolshakova: Yeah I think that makes sense- that way all the args (ns, fromDeployment, fromContainer, fdqnAddr, host, etc.) can be combined in `curlOptions` and have sane defaults/error if fields aren't set. The tricky part here is you need to know to setup a curl deployment for this to work.

test/kubernetes/e2e/framework.go
npolshakova: If I'm running a gloo classic edge api with the new framework, do I need to define a `gloogateway.Context`? If I associate the `gloogateway.Context` with the TestInstalllation I think we should rename the resource. 

Since I'm using the installation namespace in the test assertion I need it to either be on the TestInstalllation or we need to also pass the context here and make it an interface (to support non Gloo Gateway tests): 

```
testInstallation.RunTest(ctx, route_options.ConfigureRouteOptionsWithTargetRef)
```

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
npolshakova: Created one: https://github.com/solo-io/solo-projects/issues/6064

test/kubernetes/e2e/framework.go
jbohanon: I think @sam-heilbron's intention (ccorrect me if I'm wrong) is to remove the distinction between "edge" and "gateway" as products, and lean in toward "gloo gateway" is the product and within that there are classic gateway resources and k8s gateway resources. From this perspective, ALL of our tests here are "gloo gateway" tests

sam-heilbron: Oh yeah sorry for the confusion. Yeah @jbohanon is correct. My hope was just that you install Gloo Gateway, and perhaps you have k8s gateway integration enabled, perhaps not. Perhaps you have istio enabled, perhaps not. But ulimately, there are some properties about that installation that we track in a `gloogateway.Context`. And as you show on your PR, there are situations where tests may want to consume that information

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -58,7 +60,15 @@ var ConfigureRouteOptionsWithTargetRef = e2e.Test{
 					// Check fault injection is applied
 					CheckFaultInjectionFromCluster()(ctx)
 
-					// TODO(npolshak) Check status on solo-apis client object once route option status support is added
+					// Check status on solo-apis client object
```
sam-heilbron, line 63: Can we use https://github.com/solo-io/gloo/blob/main/test/helpers/input_resources.go#L27 instead? Since it accepts "getter" method

npolshakova: The downside is I think this doesn't let you check other fields on the status like: 
```
g.Expect(routeOption.GetNamespacedStatuses().GetStatuses()[installation.Metadata.InstallNamespace].GetReportedBy()).To(Equal("gloo-kube-gateway"))
						g.Expect(routeOption.GetNamespacedStatuses().GetStatuses()[installation.Metadata.InstallNamespace].GetState()).To(Equal(core.Status_Accepted))
					
```

But that could be a separate check?

npolshakova: Added the change here: https://github.com/solo-io/gloo/pull/9373/commits/58f6eb5a047a28a9c4dae4833eae3d473fa8665f

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: Moved testing changes to: https://github.com/solo-io/gloo/pull/9388

projects/gateway2/status/status_syncer.go
```diff
@@ -0,0 +1,131 @@
...
+		proxiesToQueue v1.ProxyList,
+		pluginRegistry *registry.PluginRegistry,
+	)
+	HandleProxyReports(ctx context.Context, proxiesWithReports []translatorutils.ProxyWithReports)
+}
+
+// a threadsafe factory for initializing a status syncer
+// allows for the status syncer to be shared across multiple start funcs
+type statusSyncerFactory struct {
+	registryPerProxy   map[*v1.Proxy]*registry.PluginRegistry
```
npolshakova, line 35: "This instance of proxy" != this proxy. The proxy object will be different even with the same name/namespace (route attachments, changes to proxy).

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -133,6 +143,7 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
 			TranslatedGateways: translatedGateways,
 		})
 
+		s.queueStatusForProxies(proxies, &pluginRegistry)
```
npolshakova, line 146: This doesn't need to synchronously, we can push to a channel here.

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -131,6 +132,9 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
 		for _, gw := range gwl.Items {
 			proxy := gatewayTranslator.TranslateProxy(ctx, &gw, s.writeNamespace, r)
 			if proxy != nil {
+				// Add proxy id to the proxy metadata to track proxies for status reporting
+				proxy.GetMetadata().GetLabels()[utils.ProxyId] = uuid.New().String()
```
npolshakova: Store generation/count per proxy name.namespace (drop the older counts)

projects/gateway2/status/status_syncer.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/solo-io/gloo/projects/gateway2/translator/translatorutils"
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	"github.com/solo-io/gloo/projects/gloo/pkg/syncer"
+	"github.com/solo-io/gloo/projects/gloo/pkg/utils"
+)
+
+// HandleProxyReports should conform to the OnProxiesTranslatedFn and QueueStatusForProxiesFn signatures
+var _ syncer.OnProxiesTranslatedFn = (&statusSyncerFactory{}).HandleProxyReports
+
+// QueueStatusForProxiesFn queues a status sync for a given set of Proxy resources along with the plugins that produced them
```
lgadban, line 23: wrong comment?

projects/gateway2/status/status_syncer.go
```diff
@@ -0,0 +1,166 @@
...
+	HandleProxyReports(ctx context.Context, proxiesWithReports []translatorutils.ProxyWithReports)
+}
+
+// a threadsafe factory for initializing a status syncer
+// allows for the status syncer to be shared across multiple start funcs
+type statusSyncerFactory struct {
+	// maps a proxy sync action to the plugin registry that produced it
+	// sync iteration -> plugin registry
+	registryPerSync map[int]*registry.PluginRegistry
+	// maps a proxy to the sync iteration that produced it
```
lgadban, line 42: can we elaborate here explaining how the namespacedName -> syncCount mapping is sufficient given that a proxy's namespacedName is the same across many syncs even if the underlying Proxy itself has changed

lgadban (commented):
1st pass, simple questions on commentary

projects/gateway2/status/status_syncer.go
npolshakova: Copied it over from the proxy_syncer comment, will update both!

projects/gateway2/status/status_syncer.go
```diff
@@ -0,0 +1,167 @@
...
+		// remove the proxy from the queue
+		delete(f.resyncsPerProxy, proxyKey)
+	}
+
+	for syncCount, proxies := range proxiesToReport {
+		if plugins, ok := f.registryPerSync[syncCount]; ok {
+			newStatusSyncer(plugins).applyStatusPlugins(ctx, proxies)
+
+			// reinitialize the registry if there are no more proxies for the sync iteration
+			if len(f.resyncsPerProxy) == 0 {
```
lgadban, line 106: why is this happening inside the `range proxiesToReport`?

it seems like this is effectively saying "if we don't have any proxies to sync status for, throw away whatever plugins are maintained because they are no longer needed"

if `f.resyncPerProxy` is empty but there are multiple proxies in `proxiesToReport`, wouldn't this break after the first iteration of the loop?

projects/gateway2/status/status_syncer_test.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+		ctx := context.Background()
+		syncer.HandleProxyReports(ctx, proxiesWithReports)
+
+		// Ensure proxy one has been removed from the queue after handling reports, but proxy two is still present
+		proxiesMap = syncer.(*statusSyncerFactory).resyncsPerProxy
+		Expect(proxiesMap).ToNot(ContainElement(proxyOneNameNs))
+		Expect(proxiesMap).ToNot(ContainElement(proxyTwoNameNs))
+		registryMap = syncer.(*statusSyncerFactory).registryPerSync
+		Expect(registryMap).ToNot(ContainElement(123))
```
lgadban, line 166: how would proxyTwo sync it's status if the plugins for `123` are now gone?

projects/gateway2/status/status_syncer_test.go
```diff
@@ -0,0 +1,273 @@
...
+				Reports: translatorutils.TranslationReports{
+					ProxyReport:     &validation.ProxyReport{},
+					ResourceReports: reporter.ResourceReports{},
+				},
+			},
+		}
+		ctx := context.Background()
+		syncer.HandleProxyReports(ctx, proxiesWithReports)
+
+		// Ensure proxy one has been removed from the queue after handling reports, but proxy two is still present
```
lgadban, line 161: actually, my comment below might not be accurate; is this comment correct? seems like proxy two is NOT present

projects/gateway2/status/status_syncer_test.go
npolshakova: Both proxies are in the list here: 

```
// Handle the proxy reports only for proxy one (this is invoked as a callback in the envoy translator syncer)
		proxiesWithReports := []translatorutils.ProxyWithReports{
			{
				Proxy: proxyOne,
				Reports: translatorutils.TranslationReports{
					ProxyReport:     &validation.ProxyReport{},
					ResourceReports: reporter.ResourceReports{},
				},
			},
			{
				Proxy: proxyTwo,
				Reports: translatorutils.TranslationReports{
					ProxyReport:     &validation.ProxyReport{},
					ResourceReports: reporter.ResourceReports{},
				},
			},
		}
		ctx := context.Background()
		syncer.HandleProxyReports(ctx, proxiesWithReports)
```

The comment is off- it's from an older iteration of the test. I can update it to call HandleProxyReports twice though and first check 1) proxy one is correctly handled, the plugins are still present. 2) proxy two is correctly handled, everything is cleaned up

projects/gateway2/status/status_syncer_test.go
npolshakova: Yep, the comment is wrong- I'll update the test: https://github.com/solo-io/gloo/pull/9373#discussion_r1576637014

npolshakova: Ah, actually I already have that test as part of `should queue proxy, handle report and clean up after report is processed`. I just need to fix the comment here.

projects/gateway2/status/status_syncer.go
```diff
@@ -0,0 +1,167 @@
...
+		var proxySyncCount int
+		// Get the sync iteration that produced the proxy from the proxy metadata
+		if proxyWithReport.Proxy.GetMetadata().GetAnnotations() != nil {
+			if syncId, ok := proxyWithReport.Proxy.GetMetadata().GetAnnotations()[utils.ProxySyncId]; ok {
+				proxySyncCount, _ = strconv.Atoi(syncId)
+			}
+		}
+		proxyKey := getProxyNameNamespace(proxyWithReport.Proxy)
+
+		if f.resyncsPerProxy[proxyKey] > proxySyncCount {
```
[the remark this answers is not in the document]
npolshakova: Yep exactly, that's what the last unit test shows. If the syncCount is higher, we continue because we can ignore the older sync reports since they are not the latest report.

projects/gateway2/status/status_syncer_test.go
npolshakova: Actually that test is already present^. I'm just going to fix the comment here.

projects/gateway2/status/status_syncer.go
npolshakova: I think this should be moved to the end of the HandleProxyReports(). I agree, it doesn't need to be in this for loop.

projects/gateway2/status/status_syncer.go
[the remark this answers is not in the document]
npolshakova: Added a comment: https://github.com/solo-io/gloo/pull/9373/commits/e647278a6f0520a62061e9d43155e36e136858e6

projects/gateway2/status/status_syncer.go
npolshakova: Moved out of the loop: https://github.com/solo-io/gloo/pull/9373/commits/e647278a6f0520a62061e9d43155e36e136858e6

projects/gateway2/status/status_syncer_test.go
npolshakova: Updated comment: https://github.com/solo-io/gloo/pull/9373/commits/e647278a6f0520a62061e9d43155e36e136858e6

lgadban (changes requested):
can we update the changelog to resolve this issue: https://github.com/solo-io/solo-projects/issues/6084

projects/gateway2/controller/start.go
```diff
@@ -117,6 +122,9 @@ func Start(ctx context.Context, cfg StartConfig) error {
 		mgr,
 		k8sGwExtensions,
 		cfg.ProxyClient,
+		cfg.QueueStatusForProxies,
+		cfg.RouteOptionClient,
+		cfg.StatusReporter,
```
lgadban, line 127: why are RouteOptionClient and StatusReporter needed on the ProxySyncer?
can we remove from `NewProxySyncer(...)` as well as on the struct itself?

lgadban (commented):
mostly LGTM, just a few more items to possibly cleanup

ilackarms (dismissed):

lgadban (approved):

sam-heilbron (approved):

PR #9376, opened 2024-04-16
[1.16.x] Gateway validation flake fix (#9119)

sheidkamp:
# Description

Backport of https://github.com/solo-io/gloo/pull/9119

### DIFF
--- a/test/kube2e/gateway/gateway_test.go
+++ b/test/kube2e/gateway/gateway_test.go
@@ -2148,8 +2148,9 @@ spec:
 					})
 
 					// Although these tests delete the secret handled by our SnapshotWriter, because we set `IgnoreNotFound` when deleting snapshot resources, this won't cause an issue.
-					err = resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
-					Expect(err).NotTo(HaveOccurred())
+					Eventually(func() error {
+						return resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+					}).WithPolling(500 * time.Millisecond).WithTimeout(30 * time.Second).ShouldNot(HaveOccurred())
 				})
 
 				It("can delete a secret that is not in use", func() {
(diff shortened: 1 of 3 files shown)


test/kube2e/gateway/gateway_test.go
```diff
@@ -2148,8 +2148,9 @@ spec:
 					})
 
 					// Although these tests delete the secret handled by our SnapshotWriter, because we set `IgnoreNotFound` when deleting snapshot resources, this won't cause an issue.
-					err = resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
-					Expect(err).NotTo(HaveOccurred())
+					Eventually(func() error {
```
sam-heilbron, line 2151: It seems like the comment above this eventually is no longer accurate, am I misunderstanding what it is saying?

test/kube2e/gateway/gateway_test.go
```diff
@@ -2148,8 +2148,9 @@ spec:
...
+						return resourceClientset.KubeClients().CoreV1().Secrets(testHelper.InstallNamespace).Delete(ctx, secretName, metav1.DeleteOptions{})
+					}).WithPolling(500 * time.Millisecond).WithTimeout(30 * time.Second).ShouldNot(HaveOccurred())
```
sam-heilbron, line 2153: I know this is a backport, so happy to leave it as is. Is there a reason why we feel 30 seconds is necessary?  If a secret can't be deleted after 5 seconds, it feels like something is wrong

sam-heilbron (approved):
Approved since this is a direct backport. Had some questions that you can choose to address (or not)

jenshu (approved):

test/kube2e/gateway/gateway_test.go
sheidkamp: I agree... I can put together as quick PR to update this in main

test/kube2e/gateway/gateway_test.go
sheidkamp: I think its still true - earlier in this file:
https://github.com/solo-io/gloo/blob/ac53e453546c081be1dfddd606fcdadb40eb0785/test/kube2e/gateway/gateway_test.go#L132C1-L138C4

```
        JustAfterEach(func() {
                err := snapshotWriter.DeleteSnapshot(glooResources, clients.DeleteOpts{
                        Ctx:            ctx,
                        IgnoreNotExist: true,
                })
        })
```

sheidkamp: Although it is `IgnoreNotExist` not `IgnoreNotFound`

PR #9382, opened 2024-04-18
test/kube: Update framework, add resource clients

sam-heilbron:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/pkg/utils/kubeutils/kubectl/cli.go
+++ b/pkg/utils/kubeutils/kubectl/cli.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/solo-io/gloo/pkg/utils/requestutils/curl"
+	"github.com/solo-io/go-utils/threadsafe"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 
 	"github.com/solo-io/gloo/pkg/utils/cmdutils"
@@ -153,4 +156,28 @@ func (c *Cli) StartPortForward(ctx context.Context, options ...portforward.Optio
 		retry.Attempts(5),
 	)
 	return portForwarder, err
+}
+
+// CurlFromDeployment executes a curl from a deployment
+// The assumption is that the deployment has a container named `curl` that will be used to invoke the request
+func (c *Cli) CurlFromDeployment(ctx context.Context, deploymentMeta v1.ObjectMeta, options ...curl.Option) (string, error) {
+	var outLocation threadsafe.Buffer
+
+	args := []string{
+		"exec",
+		"-n", deploymentMeta.GetNamespace(),
+		fmt.Sprintf("deployment/%s", deploymentMeta.GetName()),
+		"-c",
+		"curl",
+		"--",
+		"curl",
+	}
+	args = append(args, curl.BuildArgs(options...)...)
+
+	err := c.Command(ctx, args...).WithStdin(&outLocation).Run().Cause()
+	if err != nil {
+		return "", err
+	}
+
+	return outLocation.String(), nil
 }
--- a/test/kube2e/glooctl/istio_test.go
+++ b/test/kube2e/glooctl/istio_test.go
@@ -37,6 +37,10 @@ var (
 
 var _ = Describe("Istio", Ordered, func() {
 
+	// These tests are known to be inconsistent, and cause toil in our CI pipeline
+	// https://github.com/solo-io/solo-projects/issues/6048 tracks the work to re-enable these tests
+	Skip("These tests are inconsistent. Temporarily disabling")
+
 	// Tests for: `glooctl istio [..]`
 	// These tests assume that Gloo and Istio are pre-installed in the cluster
 
--- a/test/kubernetes/e2e/framework.go
+++ b/test/kubernetes/e2e/framework.go
@@ -2,9 +2,9 @@ package e2e
 
 import (
 	"context"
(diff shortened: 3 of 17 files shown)


pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -154,3 +157,27 @@ func (c *Cli) StartPortForward(ctx context.Context, options ...portforward.Optio
...
+// The assumption is that the deployment has a container named `curl` that will be used to invoke the request
+func (c *Cli) CurlFromDeployment(ctx context.Context, deploymentMeta v1.ObjectMeta, options ...curl.Option) (string, error) {
+	var outLocation threadsafe.Buffer
+
+	args := []string{
+		"exec",
+		"-n", deploymentMeta.GetNamespace(),
+		fmt.Sprintf("deployment/%s", deploymentMeta.GetName()),
+		"-c",
+		"curl",
```
npolshakova, line 171: Let's not hard code the container here?

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -154,3 +157,27 @@ func (c *Cli) StartPortForward(ctx context.Context, options ...portforward.Optio
 	)
 	return portForwarder, err
 }
+
+// CurlFromDeployment executes a curl from a deployment
+// The assumption is that the deployment has a container named `curl` that will be used to invoke the request
+func (c *Cli) CurlFromDeployment(ctx context.Context, deploymentMeta v1.ObjectMeta, options ...curl.Option) (string, error) {
```
npolshakova, line 163: If we're using the curl.Options{}, I think we need to make it more clear the host would be the FDQN and you need to provide an arg for the host header (which is different). Maybe we can make those required args to this helper function? I don't see why you would ever call it without setting `host` to overwrite the default `"127.0.0.1"` since it's an in cluster curl.

sam-heilbron: Good point! Im happy to pull this out of this PR too. My hope was to show how it could be done by keeping the compleixty of the CLI command in this utility, instead of having other methods create the kubectl command themselves

test/kubernetes/e2e/framework.go
```diff
@@ -59,39 +50,49 @@ func (c *TestCluster) RegisterTestInstallation(name string, glooGatewayContext *
...
-	delete(c.activeInstallations, installation.Name)
+	delete(c.activeInstallations, installation.String())
 }
 
 // TestInstallation is the structure around a set of tests that validate behavior for an installation
 // of Gloo Gateway.
 type TestInstallation struct {
+	fmt.Stringer
+
 	*TestCluster
```
jbohanon, line 89: This seems like the time to change this from composition to a field if that's still the plan

test/kubernetes/e2e/framework.go
```diff
@@ -126,11 +137,9 @@ func (i *TestInstallation) UninstallGlooGateway(ctx context.Context, uninstallAc
 // We intentionally do not expose a RunTests method, because then we would
 // lose the ability to randomize tests through the testing framework
 func (i *TestInstallation) RunTest(ctx context.Context, test Test) {
-	if test.Name == "" {
-		i.TestingFramework.Fatal("All tests must include a name")
-	}
+	gomega.Expect(test.Name).NotTo(gomega.BeEmpty(), "All tests must include a name")
 
-	i.TestingFramework.Logf("TEST: %s", test.Name)
+	ginkgo.GinkgoWriter.Print("TEST: %s", test.Name)
```
jbohanon, line 142: Should we use the operator's progress writer here since we still have that field?

test/kube2e/glooctl/istio_test.go
```diff
@@ -37,6 +37,10 @@ var (
 
 var _ = Describe("Istio", Ordered, func() {
 
+	// These tests are known to be inconsistent, and cause toil in our CI pipeline
+	// https://github.com/solo-io/solo-projects/issues/6048 tracks the work to re-enable these tests
+	Skip("These tests are inconsistent. Temporarily disabling")
```
jbohanon, line 42: This should be on a separate, tightly scoped PR in my opinion. Will make reversion easier and git history cleaner, plus no need to hang it up on this PR merging

test/kubernetes/e2e/framework.go
sam-heilbron: yess great call (i had forgotten i agreeed to that)

test/kubernetes/e2e/framework.go
sam-heilbron: Love it

test/kube2e/glooctl/istio_test.go
sam-heilbron: I completely agree, I created https://github.com/solo-io/gloo/pull/9383

pkg/utils/kubeutils/kubectl/cli.go
sam-heilbron: I ended up pivoting and using another utility, let me know what you think

npolshakova (approved):
Tested as part of https://github.com/solo-io/gloo/pull/9388

jbohanon (approved):

PR #9384, opened 2024-04-18
Bump golangci-lint for parity with CI and increase static analysis timeout

davidjumani:
# Description

Bump golangci-lint to v1.57.2 for parity with CI and increase the static analysis timeout to 10 min

```
[15:32:18] ~/lab/gloo ➦ bump-linter $ make print-LINTER_VERSION
v1.57.2
[15:32:25] ~/lab/gloo ➦ bump-linter $ make install-go-tools
...
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.57.2
...
```

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #9386, opened 2024-04-18
Bump static analysis timeout

davidjumani:
# Description

Bump static analysis timeout to 10 min

# Checklist:

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -4,8 +4,8 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
-  process_skip_directives:
-    name: Process Skip Directives
+  prepare_env:
+    name: Prepare Environment
     runs-on: ubuntu-22.04
     timeout-minutes: 5
     outputs:
@@ -37,8 +37,8 @@ jobs:
     name: Build and Test Gloo Gateway
     runs-on: ubuntu-22.04
     timeout-minutes: 10
-    needs: process_skip_directives
-    if: ${{ !github.event.pull_request.draft && needs.process_skip_directives.outputs.should-auto-succeed-tests != 'true' }}
+    needs: prepare_env
+    if: ${{ !github.event.pull_request.draft && needs.prepare_env.outputs.should-auto-succeed-tests != 'true' }}
     steps:
     - uses: actions/checkout@v4
     - name: Setup Go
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yaml
```diff
@@ -4,11 +4,41 @@ on:
...
+        if [[ ! -z $skip_kube_tests && $skip_kube_tests = true ]] ; then
+          echo "auto_succeed=true" >> $GITHUB_OUTPUT
+        fi
+
   test_gloo_gateway:
     name: Build and Test Gloo Gateway
     runs-on: ubuntu-22.04
     timeout-minutes: 10
-    if: ${{ !github.event.pull_request.draft }}
+    needs: prepare_env
```
sam-heilbron, line 40: There actually aren't kube tests, but rather the unit tests. I have found it weird that we run these separate from our build-bot, but I don't think we can skip these

.github/workflows/pr.yaml
```diff
@@ -4,11 +4,41 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
+  prepare_env:
```
sam-heilbron, line 7: I wonder if instead of adding more complexity to this job, we could pull the conformance tests into the PR worflow, and re-use the existing prepare-environment

.github/workflows/pr.yaml
davidjumani: If they are tests, wouldn't it make sense to not run them, especially since the skip tests directive is intended to skip tests (as it is mostly used with doc changes)

sam-heilbron: kube tests is intended to skip our long running tests that require a kubernetes cluster. I like the idea of saying "perhaps we want a way to skip unit tests as well". I'm just afraid of co-mingling the two. What do you think?

.github/workflows/pr.yaml
davidjumani: The interesting part is that it does use include a composite action, however invoking it requires this boilerplate code
Other places are [gen-docs](https://github.com/solo-io/gloo/blob/e63b448213d7adb096cb580947652ec6a310e123/.github/workflows/docs-gen.yaml#L23), [istio-tests](https://github.com/solo-io/gloo/blob/e63b448213d7adb096cb580947652ec6a310e123/.github/workflows/istio-tests.yaml#L11) and [kubernetes-tests](https://github.com/solo-io/gloo/blob/e63b448213d7adb096cb580947652ec6a310e123/.github/workflows/pr-kubernetes-tests.yaml#L11)

.github/workflows/pr.yaml
davidjumani: Makes sense. I'll remove this bit for now

davidjumani:
Closing in favour of https://github.com/solo-io/gloo/pull/9384

PR #9388, opened 2024-04-18
Add route option e2e test

npolshakova:
# Description

Adds route option e2e test. Statuses will be added as part of https://github.com/solo-io/gloo/pull/9373 

Should go in after: https://github.com/solo-io/gloo/pull/9382


test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+				OpAction: installation.Actions.Kubectl().NewApplyManifestAction(targetRefManifest),
+				OpAssertion: func(ctx context.Context) {
+					// First check resources are created for Gateay
+					installation.Assertions.ObjectsExist(proxyService, proxyDeployment)
+
+					// Check fault injection is applied
+					CheckFaultInjectionFromCluster()(ctx)
+
+					// TODO(npolshak) Check status on solo-apis client object once route option status support is added
+				},
```
sam-heilbron, line 63: I don't think this is doing exactly what you may want. It seems like you want to define a list of assertions. Right now you are creating one assertion, and that calls a function to create the objects assert function, but doesn't execute it. I think what you want is
```suggestion
				OpAssertions: []assertions.ClusterAssertion {
					// First check resources are created for Gateay
					installation.Assertions.ObjectsExist(proxyService, proxyDeployment),

					// Check fault injection is applied
					CheckFaultInjectionFromCluster(),

					// TODO(npolshak) Check status on solo-apis client object once route option status support is added
				},
```

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		}
+
+		err := installation.Operator.ExecuteReversibleOperations(ctx, extensionFilterRoutingOp)
+		Expect(err).NotTo(HaveOccurred())
+	},
+}
+
+func CheckFaultInjectionFromCluster() assertions.ClusterAssertion {
```
sam-heilbron, line 113: nit: Since this is intentionally isolated to this test, could we make it private?

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+		Eventually(func(g Gomega) {
+			// curl gloo-proxy-gw.default.svc.cluster.local:8080 -H "Host: example.com" -v
+			var buf threadsafe.Buffer
+			kubeCli := kubectl.NewCli().WithReceiver(&buf)
+			resp := kubeCli.CurlFromEphemeralPod(ctx, curlPod.ObjectMeta,
+				curl.WithHost(fdqnAddr),
+				curl.WithHeader("Host", "example.com"),
+				curl.VerboseOutput())
+			g.Expect(resp).To(ContainSubstring("fault filter abort"))
+			g.Expect(resp).To(ContainSubstring("HTTP/1.1 418"))
```
sam-heilbron, line 126: Let's use the curl transform: https://github.com/solo-io/gloo/blob/main/test/gomega/transforms/curl.go

This will allow us to use the httpResponse matcher: https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_http_response.go#L79

Then we can easily assert status code and other properties of the response

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+		err := installation.Operator.ExecuteReversibleOperations(ctx, extensionFilterRoutingOp)
+		Expect(err).NotTo(HaveOccurred())
+	},
+}
+
+func CheckFaultInjectionFromCluster() assertions.ClusterAssertion {
+	return func(ctx context.Context) {
+		fdqnAddr := fmt.Sprintf("%s.%s.svc.cluster.local", proxyDeployment.GetName(), proxyDeployment.GetNamespace())
+
+		Eventually(func(g Gomega) {
```
sam-heilbron, line 117: I like that this assertion is private to this test, because it contains business logic that is only relevant here. I wonder though if we could create a generic assertion: 

```go
func CurlEventuallyRespondsAssertion(ctx context.Context, []curl.Option, expectedResponse types.GomegaMatcher) assertions.ClusterAssertion {
// ...generic implementation
}

```

test/kubernetes/e2e/k8sgateway/deployer_test.go
```diff
@@ -35,8 +36,11 @@ var _ = Describe("Deployer Test", Ordered, func() {
 			},
 		)
 
-		err := testInstallation.InstallGlooGateway(ctx, testInstallation.Actions.Glooctl().NewTestHelperInstallAction())
+		err = testInstallation.InstallGlooGateway(ctx, testInstallation.Actions.Glooctl().NewTestHelperInstallAction())
 		Expect(err).NotTo(HaveOccurred())
+
+		// Initialize required resource clients for test after CRDs are installed
```
sam-heilbron, line 42: My hope in the linked PR was to handle the resourceClient initialization inside of the InstallGlooGateway call, so that devs get that for free. What do you think?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
```diff
@@ -1,11 +1,15 @@
...
+  # Note: glooRbac.namespaced settings are not supported with Gloo Gateway https://github.com/solo-io/solo-projects/issues/6064
+  # Gateway API fundamentally expects HTTPRoutes and Gateways in any namespace and cross-namespace references to be supported
+  # Currently we are explicitly disabled namespaced roles for Gloo Gateway tests, but this can be left unset.
   glooRbac:
-    namespaced: true
-    nameSuffix: e2e-test-rbac-suffix
+    namespaced: false
 settings:
-  singleNamespace: true
+  # Gloo Gateway requires access to namespaces outside of the install namespace to watch and create Gateway resources
```
sam-heilbron, line 10: I find these really helpful! (could just be a me thing)

sam-heilbron (commented):
awesome work! Hopefully this is a sign of how quickly we'll be able to iterate on/add e2e tests

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: Yep- I thought we'd want a gloo edge classic API version of the same check but if this test is isolated to Gateway apis it makes sense to keep private.

test/kubernetes/e2e/k8sgateway/deployer_test.go
npolshakova: Yep, didn't realize it was already getting initialized. We can get rid of this line then

test/kubernetes/e2e/k8sgateway/deployer_test.go
```diff
@@ -58,4 +62,16 @@ var _ = Describe("Deployer Test", Ordered, func() {
 
 	})
 
+	Context("RouteOptions", func() {
```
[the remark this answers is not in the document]
jbohanon: @sam-heilbron what do you think of this? What was your vision for separation of concerns with test installations?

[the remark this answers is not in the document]
npolshakova: This was something we were discussing- I think we should rename the `deployer_test` to encompass a specific setup. That way the different setups ("basic Gloo Gateway", "Istio integration Gloo Gateway", "Gloo Gateway disabled") are all separate.

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: Can we force `verbose=true` to be added then? Right now I need to pass in the `-v` flag to get the `HTTP/1.1 418` in the response to match.

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+				},
+			},
+		}
+
+		err := installation.Operator.ExecuteReversibleOperations(ctx, trafficRefRoutingOp)
+		Expect(err).NotTo(HaveOccurred())
+	},
+}
+
+var ConfigureRouteOptionsWithFilterExtenstion = e2e.Test{
```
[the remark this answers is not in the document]
jbohanon: It's a nit so do what you will with it, but the object name wasn't changed in the commit you pushed

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: Something like we have for the testContainer but as a shared assertion? https://github.com/solo-io/gloo/blob/423eed988ac8bf7c9f10cf6c5f32fc8c8732093e/test/kube2e/helper/curl.go#L135C25-L135C52

test/kubernetes/e2e/k8sgateway/deployer_test.go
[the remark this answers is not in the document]
sam-heilbron: Yeah it's a great callout, and I had a vision for what it could look like, but that likely will morph over time. My main hope/focus is that the tests are structured in a way that makes it easy to find what is being tested and easily add new tests. I agree, that a concern of mine is that test files grow longer and then become less manageable. 

My current leaning is that a single file can contain references to all the tests/behaviors that we want to validate per set of helm values. And that features/[dir] would have the test files per feature. 

tldr - I think its great that we keep an eye on it, and my vote is we continue to place tests in a place that makes sense at the time. I think we can add these to this file for now, and in a week/two as we add more tests we can adjust as necessary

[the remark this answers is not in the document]
jbohanon: That's a good point about tying it to an install manifest. I think then I would argue that naming the test file in line with the manifest would help clarify this coupling/boundary definition.

e.g. I think this could be renamed `basic_gloo_gateway_test.go` and it could reference `manifests/basic_gloo_gateway_test_helm.yaml`. I think specifying `k8s-gateway` is redundant since we're in the `k8sgateway` directory/suite. It would need to be called out if k8s gateway were _disabled_ for a test file/manifest.

WDYT?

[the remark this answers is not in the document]
sam-heilbron: Yeah I think that's a great point, and I'm on board with it. If we're happy with all other changes for this PR, I'd vote for merging this as it's a step in the right direction and we can fast follow to change the name of the test, and perhaps better codify in the readme the intention of test structure

test/kubernetes/e2e/features/route_options/route_options.go
sam-heilbron: yeah exactly. Doesn't need to be part of this PR. I think we'll find patterns of applying assertions like this and then we can extract those patterns into generic cases. I really want to find those patterns as early as possible since as we scale this out next week to be used in more tests, I think it's important that we have a good foundation.

tldr - I don't see this generic approach as a blocker to this PR, but I think it's worth adding in the next PR that goes into this testing space

jbohanon (approved):

sam-heilbron (dismissed):
Awesome work! All the open comments are thoughts that I think we should do, but we plan to iterate on these tests so I dont consider them a blocker to this merging.

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func checkFaultInjectionFromCluster() assertions.ClusterAssertion {
+	return func(ctx context.Context) {
+		fdqnAddr := fmt.Sprintf("%s.%s.svc.cluster.local", proxyDeployment.GetName(), proxyDeployment.GetNamespace())
+
+		Eventually(func(g Gomega) {
+			// curl gloo-proxy-gw.default.svc.cluster.local:8080 -H "Host: example.com" -v
+			var buf threadsafe.Buffer
+			kubeCli := kubectl.NewCli().WithReceiver(&buf)
```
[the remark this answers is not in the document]
npolshakova: Added the generic assertion here: https://github.com/solo-io/gloo/pull/9388/commits/a854a86ba278d59cf46bf7867e979fe13e020704

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: Added the generic assertion here: https://github.com/solo-io/gloo/pull/9388/commits/a854a86ba278d59cf46bf7867e979fe13e020704

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/gloo/test/kube2e/helper"
+	"github.com/solo-io/go-utils/log"
+)
+
+func CurlEventuallyRespondsAssertion(ctx context.Context, curlFunc func() string, logResponses bool, expectedResponse interface{}, ginkgoOffset int, timeout ...time.Duration) ClusterAssertion {
+	return func(ctx context.Context) {
+		ginkgo.GinkgoHelper()
```
sam-heilbron, line 15: The usage of GinkgoHelper (a new concept to me) allows us to not need the offset

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/gloo/test/kube2e/helper"
+	"github.com/solo-io/go-utils/log"
+)
+
+func CurlEventuallyRespondsAssertion(ctx context.Context, curlFunc func() string, logResponses bool, expectedResponse interface{}, ginkgoOffset int, timeout ...time.Duration) ClusterAssertion {
```
sam-heilbron, line 13: How do you feel about passing curl.Options instead of a curl func? My hope was that it will make it clearer which utilities to use

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -0,0 +1,39 @@
...
+	return func(ctx context.Context) {
+		ginkgo.GinkgoHelper()
+		currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
```
sam-heilbron, line 16: For these timing intervals, I like them, though I'm (probably overly) concerned about adding complexity early. What if we defined constants with sensible defaults, and don't expose it to the user. And then if devs need a custom timeout, we can expose that later on?

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -0,0 +1,39 @@
...
+
+		EventuallyWithOffset(ginkgoOffset+1, func(g Gomega) {
+			res := curlFunc()
+			select {
+			default:
+				break
+			case <-tick:
+				if logResponses {
+					log.GreyPrintf("want %v\nhave: %s", expectedResponse, res)
+				}
```
sam-heilbron, line 28: I go back/forth on whether this is actually useful for debugging. Part of me thinks that instead of defining options that users can inject, we define a standard. I could see this going in one of three directions:
1. Always log
2. Never log
3. Write to the GinkgoWriter, if the verbosity is high

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -0,0 +1,39 @@
...
+			}
+
+			expectedResponseMatcher := helper.GetExpectedResponseMatcher(expectedResponse)
```
sam-heilbron, line 31: nit: I'd like us to decouple from the helper as much as possible. That was built for backward compatibility and supports all types of matchers. I'd almost rather that we required devs to injecet the expectedHttpResponse so that we don't let them write (weaker) substring matches. What do you think?

test/kubernetes/testutils/assertions/curl.go
sam-heilbron, line 13: Do we need to accept a context to this function? We return a closure that accepts a contex there

test/kubernetes/testutils/assertions/curl.go
npolshakova: I think this is the [same defaults](https://github.com/solo-io/gloo/blob/423eed988ac8bf7c9f10cf6c5f32fc8c8732093e/test/kube2e/helper/curl.go#L64) that are currently used for the test container curls. Since it's an optional arg (`timeout ...time.Duration`) I think it shouldn't be exposed to the user unless they dig into it. My test just uses the default for example: 

```
assertions.CurlEventuallyRespondsAssertion(ctx, curlFromPod(ctx), true, expectedFaultInjectionResp, 0),
```

test/kubernetes/testutils/assertions/curl.go
npolshakova: Yeah- let's always log for now then to simplify the `CurlEventuallyRespondsAssertion` interface, and we can remove it if it's too noisy later?

test/kubernetes/testutils/assertions/curl.go
npolshakova: Sounds good! Let's force it to be `*matchers.HttpResponse` here then!

npolshakova: Do we expect to assert non-http responses at some point?

test/kubernetes/testutils/assertions/curl.go
npolshakova: I thought it might be helpful for logging, but I think we can get rid of it!

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: Oh good point. Yeah we may and when that comes we'lll likely need a new assertion. I may be over indexing, but in the past weve built generic tools that leave a lot of flexibility for devs, but also lead to lots of different implementations. My hope with this direction is we provide a more structured approach and when we need flexibility, we build it out intentionally. What do you think?

test/kubernetes/testutils/assertions/curl.go
npolshakova: Discussed offline- keeping it generic for now. If we use the curlOpts here, this function should be less generic (ie. EphemeralCurlEventuallyRespondsAssertion) because we need to build the kubeCli and run the curl command as part of the assertion directly. In the future the assertions provider can build the kubeCli for the given install, but we will need to rename the assertion for the type of curl (ephemeral pods vs. port-forward, etc.)

test/kubernetes/testutils/assertions/curl.go
npolshakova: Sounds good! I think `*matchers.HttpResponse` makes sense for now, and we expand it later!

sam-heilbron (approved):

jbohanon (approved):

Issue #6543, opened 2022-06-08
gloo validation hook rejects upstream deletion because linked service is already gone


jdef:
### Gloo Edge Version

1.10.x

### Kubernetes Version

1.21.x

### Describe the bug

I delete a namespace that had a k8s service and a gloo upstream that referred to it. the service was deleted. the gloo validation hook rejected the deletion of the upstream, complaining that the service on which it depends was missing.

### Steps to reproduce the bug

1. create ns
2. create service
3. create upstream that refers to said service
4. delete the ns - watch the delete hang
5. list upstreams in the ns - observe they are still there
6. delete the upstream manually - observe the validation hook rejecting the DELETE request (because the linked service is gone)

### Expected Behavior

since i'm deleting an upstream, if the linked service is already gone i expect validation of the DELETE to succeed

jdef:
xref #2797

hengkuang777:
+1

sebastian-popa:
+1

PR #9392, opened 2024-04-22
Fix: validation rejects kube US deletion

bewebi:
# Description

Do not consider "fake" kube-svc Upstreams in the snapshot to validate when validating delete operations for Upstreams

## Code changes
- In `ValidateGloo()`, when deleting an Upstream, also remove an Upstream with the same namespace and name prefixed by `kube-svc:` from the resource list
- In `applyRequestToSnapshot()`, called by `Validate()`, add an Upstream ref with the same namespace and name prefixed by `kube-svc:` to the list of Upstream refs to be treated as deleted in the snapshot to validate
- Add test cases for both validation functions 
  - Add a kube client and cache to the `opts` used to instantiate the test plugin registry to ensure the kubernetes plugin is processed
  - Start with a snapshot that includes both a "real" and "fake" kube upstream without the referenced service
  - Delete only the "real" upstream"
  - Assert that there are no validation errors
- Export the `UpstreamNamePrefix` from the `upstreams/kubernetes` package and use as a single source of truth
- Fix various typos found during development, including switching `context.TODO()` to `context.Background()` 

# Context

## Buggy behavior

This bug is the root cause of a flake observed in solo-io/solo-projects#6018
The series of events is as follows:
- A test runs that applies kube services
- Discovery creates Upstreams for these services which are used in tests
- On cleanup for the test we delete the Services
- Discovery attempts to delete the Upstreams due to the Services no longer being present
- _Sometimes_ validation fails because the "fake" Upstream is invalid, due to the Service it points to no longer existing
- In most cases this eventually resolves, within 10-30s, however in some cases this is persistent
  - It's not clear what eventually allows validation to pass or why it's seemingly not deterministic
- When the Upstreams are not deleted there is test pollution that causes subsequent tests to fail  

_It is not clear why this PR in particular is causing this behavior to manifest as a test flake_

This behavior is also encompassed in #6543

## "fake" Upstreams

"fake" Upstreams are created within the Gloo controller to allow for direct routing to kube Services
They are not written to etcd

For more context on the need for "fake" Upstreams for kube Services, see [this Slack conversation](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1713474769770659)

## Interesting decisions

It does not seem that modifying the snapshot clone in [`validateSnapshot()`](https://github.com/solo-io/gloo/blob/89be5de330030d3142a7035476d4f877a4f7f219/projects/gateway/pkg/validation/validator.go#L405) has an impact with regards to this fix
In other words it didn't matter whether we also removed the `kube-svc:` Upstream [here](https://github.com/solo-io/gloo/blob/89be5de330030d3142a7035476d4f877a4f7f219/projects/gateway/pkg/validation/validator.go#L442)
Though we pass the `snapshotClone` to `validateProxiesAndExtensions()`, it's only used to compute proxies
When the validator is called [here](https://github.com/solo-io/gloo/blob/89be5de330030d3142a7035476d4f877a4f7f219/projects/gateway/pkg/validation/validator.go#L324), those proxies are passed along with the `resourceToModify` which is then used to re-compute the modified snapshot to validate based on the `validator`'s `lastestSnapshot`
Further investigation seems out of scope here, but it may be worth looking into what effect, if any, using a modified clone of the snapshot has as compared to the latest snapshot

## Testing steps

In addition to unit tests I manually verified using the following steps

- Install Gloo on a Kind cluster with strict validation by setting the following helm values
  - `gateway.validation.alwaysAcceptResources=false`
  - `gateway.validation.failurePolicy=fail`
  - `gateway.validation.allowWarnings=false`
- Apply a kubernetes Service
  - For my purposes I used the setup for the failing test from solo-proejcts, see Deployment and Service below
- Observe that an Upstream is created for the Service by Discovery
- Delete the Service 
- When reproducing the bug observe that:
  - The Upstream is not immediately deleted 
  - Logs for Discovery and Gloo have errors around validation
    - In Discovery this looks like:
      ```
      {"level":"error","ts":"2024-04-16T21:48:46.917Z","logger":"uds.v1.event_loop.uds","caller":"discovery/discovery.go:150","msg":"failed reconciling upstreams","version":"undefined","discovered_by":"kubernetesplugin","upstreams":20,"error":"deleting stale resource gloo-system-test-app-2-5678: deleting resource gloo-system-test-app-2-5678: admission webhook \"gloo.gloo-system.svc\" denied the request: resource incompatible with current Gloo snapshot: [failed validating the deletion of resource namespace: gloo-system name: gloo-system-test-app-2-5678: validating *v1.Upstream name:\"gloo-system-test-app-2-5678\" namespace:\"gloo-system\": 1 error occurred:\n\t* failed gloo validation resource reports: 2 errors occurred:\n\t* invalid resource gloo-system.kube-svc:gloo-system-test-app-2-5678\n\t* WARN: \n  [1 error occurred:\n\t* Upstream name:\"kube-svc:gloo-system-test-app-2-5678\" namespace:\"gloo-system\" references the service \"test-app-2\" which does not exist in namespace \"gloo-system\"\n\n]\n\n\n\n]","errorVerbose":"admission webhook \"gloo.gloo-system.svc\" denied the request: resource incompatible with current Gloo snapshot: [failed validating the deletion of resource namespace: gloo-system name: gloo-system-test-app-2-5678: validating *v1.Upstream name:\"gloo-system-test-app-2-5678\" namespace:\"gloo-system\": 1 error occurred:\n\t* failed gloo validation resource reports: 2 errors occurred:\n\t* invalid resource gloo-system.kube-svc:gloo-system-test-app-2-5678\n\t* WARN: \n  [1 error occurred:\n\t* Upstream name:\"kube-svc:gloo-system-test-app-2-5678\" namespace:\"gloo-system\" references the service \"test-app-2\" which does not exist in namespace \"gloo-system\"\n\n]\n\n\n\n]\ndeleting resource gloo-system-test-app-2-5678\ngithub.com/solo-io/solo-kit/pkg/errors.Wrapf\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/errors/errors.go:15\ngithub.com/solo-io/solo-kit/pkg/api/v1/clients/kube.(*ResourceClient).Delete\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/api/v1/clients/kube/resource_client.go:283\ngithub.com/solo-io/solo-kit/pkg/api/v1/reconcile.deleteStaleResource\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/api/v1/reconcile/reconciler.go:140\ngithub.com/solo-io/solo-kit/pkg/api/v1/reconcile.(*reconciler).Reconcile\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/api/v1/reconcile/reconciler.go:53\ngithub.com/solo-io/gloo/projects/gloo/pkg/api/v1.(*upstreamReconciler).Reconcile\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/api/v1/upstream_reconciler.sk.go:46\ngithub.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).Resync\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:146\ngithub.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).StartUds.func1\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:117\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.22.2/x64/src/runtime/asm_amd64.s:1695\ndeleting stale resource gloo-system-test-app-2-5678\ngithub.com/solo-io/solo-kit/pkg/errors.Wrapf\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/errors/errors.go:15\ngithub.com/solo-io/solo-kit/pkg/api/v1/reconcile.(*reconciler).Reconcile\n\t/home/runner/go/pkg/mod/github.com/solo-io/solo-kit@v0.34.2/pkg/api/v1/reconcile/reconciler.go:54\ngithub.com/solo-io/gloo/projects/gloo/pkg/api/v1.(*upstreamReconciler).Reconcile\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/api/v1/upstream_reconciler.sk.go:46\ngithub.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).Resync\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:146\ngithub.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).StartUds.func1\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:117\nruntime.goexit\n\t/opt/hostedtoolcache/go/1.22.2/x64/src/runtime/asm_amd64.s:1695","stacktrace":"github.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).Resync\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:150\ngithub.com/solo-io/gloo/projects/gloo/pkg/discovery.(*UpstreamDiscovery).StartUds.func1\n\t/home/runner/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240415211143-4482d7dae220/projects/gloo/pkg/discovery/discovery.go:117"}
      ```
    - In Gloo this looks like:
      ```
      {"level":"error","ts":"2024-04-19T20:05:37.012Z","logger":"gloo-ee.v1.event_loop.setup.gateway-validation-webhook","caller":"k8sadmission/validating_admission_webhook.go:356","msg":"Validation failed: 1 error occurred:\n\t* failed validating the deletion of resource namespace: gloo-system name: gloo-system-test-app-1-5678: validating *v1.Upstream name:\"gloo-system-test-app-1-5678\"  namespace:\"gloo-system\": 1 error occurred:\n\t* failed gloo validation resource reports: 4 errors occurred:\n\t* invalid resource gloo-system.gloo-system-test-app-2-5678\n\t* WARN: \n  [1 error occurred:\n\t* Upstream name:\"gloo-system-test-app-2-5678\"  namespace:\"gloo-system\" references the service \"test-app-2\" which does not exist in namespace \"gloo-system\"\n\n]\n\t* WARN: \n  [1 error occurred:\n\t* Upstream name:\"kube-svc:gloo-system-test-app-1-5678\"  namespace:\"gloo-system\" references the service \"test-app-1\" which does not exist in namespace \"gloo-system\"\n\n]\n\t* WARN: \n  [1 error occurred:\n\t* Upstream name:\"kube-svc:gloo-system-test-app-2-5678\"  namespace:\"gloo-system\" references the service \"test-app-2\" which does not exist in namespace \"gloo-system\"\n\n]\n\n\n\n\n\n","version":"PR-6018","Kind":"gloo.solo.io/v1, Kind=Upstream","Namespace":"gloo-system","Name":"gloo-system-test-app-1-5678","UID":"c21f7ce2-fe04-4ae2-ab22-8758939508b7","PatchOperation":"DELETE","UserInfo":{"username":"system:serviceaccount:gloo-system:discovery","uid":"956911bf-f758-44a1-b3d8-1d44cacf09af","groups":["system:serviceaccounts","system:serviceaccounts:gloo-system","system:authenticated"],"extra":{"authentication.kubernetes.io/pod-name":["discovery-59b55ffff-mpvrl"],"authentication.kubernetes.io/pod-uid":["def378d1-066d-4e3a-b033-65fc6a97f729"]}},"stacktrace":"github.com/solo-io/gloo/projects/gateway/pkg/services/k8sadmission.(*gatewayValidationWebhook).makeAdmissionResponse\n\t/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240418215505-b60f217310e1/projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go:356\ngithub.com/solo-io/gloo/projects/gateway/pkg/services/k8sadmission.(*gatewayValidationWebhook).ServeHTTP\n\t/go/pkg/mod/github.com/solo-io/gloo@v1.17.0-beta9.0.20240418215505-b60f217310e1/projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go:236\nnet/http.(*ServeMux).ServeHTTP\n\t/usr/local/go/src/net/http/server.go:2683\nnet/http.serverHandler.ServeHTTP\n\t/usr/local/go/src/net/http/server.go:3137\nnet/http.(*conn).serve\n\t/usr/local/go/src/net/http/server.go:2039"}
      ```
  - When testing the fix observe that the Upstream is immediately deleted and there are no errors in Discovery or Gloo logs

## Notes for reviewers

solo-io/solo-projects#6018 is pointing to this branch
Observe that the previously observed flake no longer occurs

# Checklist:


ashishb-solo (commented):
Overall, this looks fine to me. I just have one question: there are a lot of thoughts on how performance might be affected by this change? It seems to me like we might deleting (or attempting to delete) a lot more resources. Are there any circumstances under which this might cause a significant performance regression? ie. what if there are a lot of upstreams, or a lot of kubernetes service upstreams in the Gloo configuration?

bewebi:
To address @ashishb-solo's questions:

> do you have any thoughts on how performance might be affected by this change? 

There is no meaningful performance hit
In `ValidateGloo()` only performance hit is one extra call to `RemoveFromResourceList()`, which is O(n) where n=the number of Upstreams. This is already called once, and 2*O(n) == O(n), not to mention this is iterating through a list of refs in memory, so many thousands of Upstreams for this to be an expensive call in the first place
In `Validate()` we add an entry to `deletedUpstreamRefs()` for each US we're trying to delete, which will always be 1, and this list is converted to a `sets.String` when passed to `DeleteResources()` so there is one additional `Instert()` call, and no further performance implication

> It seems to me like we might deleting (or attempting to delete) a lot more resources. 

The changes here only affect validation when a delete operation is attempted a single Upstream CR. All we're doing is removing the corresponding `kube-svc` US from the _snapshot used for validation_ so that any errors with that do not impact the validation decision. The `kube-svc` US is not actually removed from the working snapshot in the course of validation. Changing the validation behavior will not change how often delete operations occur, other than that there may be fewer retries if the initial attempts are successful.

> Are there any circumstances under which this might cause a significant performance regression? ie. what if there are a lot of upstreams, or a lot of kubernetes service upstreams in the Gloo configuration?

As described above, this will not have an order of magnitude performance regression in any circumstance. If there are so many Upstreams in the snapshot that iterating through them a second time is a significant operation, there likely were already performance issues.

ashishb-solo (approved):
Thanks for your thorough reply regarding performance! I think `Validate` is safe. I do still have some reservations about the `ValidateGloo` performance, but I inspected `RemoveFromResourceList` and even though it's O(# of upstreams) it looks like it's a pretty fast function, so I think we should be safe.

Sorry for taking so long with this review. I've been pretty swamped with the other issue I was working on, and it took me a fair amount of time to read up on what's going on here since I don't have a lot of familiarity with this portion of the codebase.

nfuden (approved):
Ran through a few iterations on this. I dont love the fact that I don't know if I fully agree with functionality but as far as I can tell its correct.

Given this I have tried to break the implementation by constructing various validation examples and the proposed code worked correctly in each case.

Therefore looks good to me.

PR #9395, opened 2024-04-22
tests/kube/e2e: Expand utilities, run tests multiple times on PRs

sam-heilbron:
# Description

Expand the utilities for the new kubernetes e2e tests

# Context

## Ginkgo Flags
- I set the value of `--repeat` to 1 so that on PRs we run tests twice. The hope here is we are more aggressive about flakes, and thus can act on them more quickly
- I set the value of `--flake-attempts` to 0 to ensure that we do not retry flakes. My hope is that in other suites that are flakier we could increase that number, and so I'd rather be explicit in this suite.

## Cmd Data Race
I don't actually know if this will be resolved, which is why I set it to `resolvesIssue=false`. My hope is that by increasing the number of runs on a PR, we will see more frequently how problematic this is.

## Assertion Utilities 
Instead of requiring callers to construct their own Kube CLI, we can rely on the one in the assertions provider

## Kube DNS
We have a number of places in code that construct the FQDN for a service. I opted to define the mechanism in a shared utility, and call it from the places that use it.

## Interesting decisions
 -

## Testing steps
- Rely on existing CI

## Notes for reviewers
-

# Checklist:

### DIFF
--- a/pkg/utils/kubeutils/names.go
+++ b/pkg/utils/kubeutils/names.go
@@ -4,6 +4,7 @@ const (
 	GlooDeploymentName = "gloo"
 	GlooServiceName    = "gloo"
 
-	// The name of the port in the gloo control plane Kubernetes Service that serves xDS config.
+	// GlooXdsPortName is the name of the port in the Gloo Gateway control plane Kubernetes Service that serves xDS config.
+	// See: install/helm/gloo/templates/2-gloo-service.yaml
 	GlooXdsPortName = "grpc-xds"
 )
--- a/test/kubernetes/e2e/features/route_options/route_options.go
+++ b/test/kubernetes/e2e/features/route_options/route_options.go
@@ -56,10 +56,10 @@ var ConfigureRouteOptionsWithTargetRef = e2e.Test{
 				OpAction: installation.Actions.Kubectl().NewApplyManifestAction(targetRefManifest),
 				OpAssertions: []assertions.ClusterAssertion{
 					// First check resources are created for Gateway
-					installation.Assertions.ObjectsExist(proxyService, proxyDeployment, curlPod),
+					installation.Assertions.ObjectsExist(proxyService, proxyDeployment),
 
 					installation.Assertions.EphemeralCurlEventuallyResponds(
-						curlPod.ObjectMeta,
+						curlPod,
 						[]curl.Option{
 							curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
 							curl.WithHostHeader("example.com"),
@@ -94,11 +94,11 @@ var ConfigureRouteOptionsWithFilterExtenstion = e2e.Test{
 				OpAction: installation.Actions.Kubectl().NewApplyManifestAction(filterExtensioManifest),
 				OpAssertions: []assertions.ClusterAssertion{
 					// First check resources are created for Gateway
-					installation.Assertions.ObjectsExist(proxyService, proxyDeployment, curlPod),
+					installation.Assertions.ObjectsExist(proxyService, proxyDeployment),
 
 					// Check fault injection is applied
 					installation.Assertions.EphemeralCurlEventuallyResponds(
-						curlPod.ObjectMeta,
+						curlPod,
 						[]curl.Option{
 							curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
 							curl.WithHostHeader("example.com"),
--- a/test/kubernetes/e2e/features/route_options/route_options.go
+++ b/test/kubernetes/e2e/features/route_options/route_options.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"path/filepath"
 
+	v1 "k8s.io/apiserver/pkg/apis/example/v1"
+
 	"github.com/solo-io/gloo/pkg/utils/kubeutils"
 
 	. "github.com/onsi/gomega"
@@ -24,13 +26,21 @@ var (
 	targetRefManifest      = filepath.Join(util.MustGetThisDir(), "inputs/fault-injection-targetref.yaml")
 	filterExtensioManifest = filepath.Join(util.MustGetThisDir(), "inputs/fault-injection-filter-extension.yaml")
 
-	// When we apply the deployer-provision.yaml file, we expect resources to be created with this metadata
+	// When we apply the fault injection manifest files, we expect resources to be created with this metadata
 	glooProxyObjectMeta = metav1.ObjectMeta{
 		Name:      "gloo-proxy-gw",
 		Namespace: "default",
 	}
 	proxyDeployment = &appsv1.Deployment{ObjectMeta: glooProxyObjectMeta}
 	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+
+	// curlPod is the Pod that will be used to execute curl requests, and is defined in the fault injection manifest files
+	curlPod = &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "curl",
+			Namespace: "curl",
+		},
+	}
 
 	expectedFaultInjectionResp = &testmatchers.HttpResponse{
 		StatusCode: http.StatusTeapot,
@@ -50,10 +60,13 @@ var ConfigureRouteOptionsWithTargetRef = e2e.Test{
 					// First check resources are created for Gateay
 					installation.Assertions.ObjectsExist(proxyService, proxyDeployment),
 
-					installation.Assertions.CurlEventuallyResponds([]curl.Option{
-						curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
-						curl.WithHostHeader("example.com"),
-					}, expectedFaultInjectionResp),
+					installation.Assertions.EphemeralCurlEventuallyResponds(
+						curlPod.ObjectMeta,
+						[]curl.Option{
+							curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
+							curl.WithHostHeader("example.com"),
+						},
+						expectedFaultInjectionResp),
 
 					// TODO(npolshak) Check status on solo-apis client object once route option status support is added
 				},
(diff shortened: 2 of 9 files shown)


pkg/utils/kubeutils/names.go
```diff
@@ -4,6 +4,7 @@ const (
 	GlooDeploymentName = "gloo"
 	GlooServiceName    = "gloo"
 
-	// The name of the port in the gloo control plane Kubernetes Service that serves xDS config.
+	// GlooXdsPortName is the name of the port in the Gloo Gateway control plane Kubernetes Service that serves xDS config.
+	// See: install/helm/gloo/templates/2-gloo-service.yaml
```
npolshakova, line 8: Is the comment correct here? Is this only used in Gloo Gateway?

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -4,34 +4,80 @@ import (
...
+var (
+	// curlPodObjectMeta contains the ObjectMeta for the Pod that will be used to execute curl requests
+	curlPodObjectMeta = metav1.ObjectMeta{
+		Name:      "curl",
+		Namespace: "curl",
+	}
+)
+
+// CurlEventuallyResponds returns a ClusterAssertion to assert that a set of curl.Option will return the expected matchers.HttpResponse
+func (p *Provider) CurlEventuallyResponds(curlOptions []curl.Option, expectedResponse *matchers.HttpResponse, timeout ...time.Duration) ClusterAssertion {
```
npolshakova, line 26: Do we want to specify this `CurlEventuallyResponds` calls `p.clusterContext.Cli.CurlFromEphemeralPod`? I think right now it's not clear that the user needs to specify these options correctly otherwise the curl won't work from the pod: 

```
curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
curl.WithHostHeader("example.com"),
```

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -4,34 +4,80 @@ import (
...
 	"github.com/solo-io/gloo/test/gomega/matchers"
 	"github.com/solo-io/gloo/test/gomega/transforms"
 	"github.com/solo-io/gloo/test/kube2e/helper"
-	"github.com/solo-io/go-utils/log"
 )
 
-func CurlEventuallyRespondsAssertion(curlFunc func() string, expectedResponse *matchers.HttpResponse, timeout ...time.Duration) ClusterAssertion {
+var (
+	// curlPodObjectMeta contains the ObjectMeta for the Pod that will be used to execute curl requests
+	curlPodObjectMeta = metav1.ObjectMeta{
```
npolshakova, line 19: I'm not the biggest fan of hardcoding the curl deployment here. Maybe we can pass the `metav1.ObjectMeta` as an arg to `CurlEventuallyResponds`? The curl deployment is part of the setup manifest. None of the testing assertions should depend on objects from a specific manifest being present or not.

test/kubernetes/testutils/assertions/curl.go
npolshakova: For example, we can update the comment to clarify to will curl from an ephemeral container, add validation checks to the curlOptions and maybe rename it to `EphemeralCurlEventuallyResponds`. I'd also add the container info as something that can be configured (we might want curl pods in different namespaces, etc.)

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -4,34 +4,80 @@ import (
...
 	return func(ctx context.Context) {
 		ginkgo.GinkgoHelper()
 		currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+
 		// for some useful-ish output
 		tick := time.Tick(currentTimeout / 8)
 
 		Eventually(func(g Gomega) {
-			res := curlFunc()
+			res := p.clusterContext.Cli.CurlFromEphemeralPod(ctx, curlPodObjectMeta, curlOptions...)
```
npolshakova, line 35: We should check that the `curlPodObjectMeta` is present in the cluster and in a "ready" state before doing the curl. Otherwise, this will fail multiple times on the eventually.

pkg/utils/kubeutils/names.go
sam-heilbron: I was trying to adopt our new language, where the product is "Gloo Gateway" instead of "Gloo Edge"

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: That's fair! I like the callout and am happy to adjust. My thinking was, we start with the case that it always is (up to this point), and when we need to support objects not named curl, we can add functionality for it. However, I'm good with introducing that now

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: Yeah that's fair, and I guess raises the question of responsibility for assertions. On one hand, I'd like assertions to be discrete behaviors, that you as the caller/developer can string together. On the other, some have logic connections to others. 
Where do you fall on the divide of: small discrete assertions that devs have to string together, vs, aggregate assertions

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -4,34 +4,74 @@ import (
...
+			WithPolling(pollingInterval).
+			WithContext(ctx).
+			Should(Succeed())
+	}
+}
+
+// CurlFnEventuallyResponds returns a ClusterAssertion that behaves similarly to EphemeralCurlEventuallyResponds
+// The difference is that it accepts a generic function to execute the curl, instead of requiring the caller to pass explicit curl.Option
+// We recommend that developers rely on the typed EphemeralCurlEventuallyResponds but we want to provide the flexibility of other solutions as well
+func (p *Provider) CurlFnEventuallyResponds(curlFn func() string, expectedResponse *matchers.HttpResponse, timeout ...time.Duration) ClusterAssertion {
```
jenshu, line 51: it this currently used anywhere? i would prefer to not expose functions until we have a use for them, to reduce any potential confusion about which functions to use

test/kubernetes/e2e/features/route_options/route_options.go
```diff
@@ -24,35 +24,22 @@ var (
...
+	// When we apply the fault injection manifest files, we expect resources to be created with this metadata
 	glooProxyObjectMeta = metav1.ObjectMeta{
 		Name:      "gloo-proxy-gw",
 		Namespace: "default",
 	}
 	proxyDeployment = &appsv1.Deployment{ObjectMeta: glooProxyObjectMeta}
 	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
 
+	// curlPod is the Pod that will be used to execute curl requests, and is defined in the fault injection manifest files
 	curlPod = &corev1.Pod{
```
jenshu, line 36: is this a general pod that we can reuse for multiple tests?
if so, can we split it into some shared manifest, then we can simplify EphemeralCurlEventuallyResponds to not have the curlPod arg (since the pod will be the same for all tests)?

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -4,34 +4,74 @@ import (
...
+// EphemeralCurlEventuallyResponds returns a ClusterAssertion to assert that a set of curl.Option will return the expected matchers.HttpResponse
+// This implementation relies on executing from an ephemeral container.
+// It is the caller's responsibility to ensure the curlPodMeta points to a pod that is alive and ready to accept traffic
+func (p *Provider) EphemeralCurlEventuallyResponds(curlPodMeta metav1.ObjectMeta, curlOptions []curl.Option, expectedResponse *matchers.HttpResponse, timeout ...time.Duration) ClusterAssertion {
 	return func(ctx context.Context) {
 		ginkgo.GinkgoHelper()
 		currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+
 		// for some useful-ish output
 		tick := time.Tick(currentTimeout / 8)
```
jenshu, line 26: nit: not from this PR, but should we make this a more regular interval (perhaps the pollingInterval) that's not dependent on the currentTimeout value?

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: True, it's not. I will defer to @npolshakova who had introduced this originally, and raised points about the benefit of a function over curl.OPtions.

test/kubernetes/e2e/features/route_options/route_options.go
sam-heilbron: Slack context: https://solo-io-corp.slack.com/archives/C03MFATU265/p1713882000712249?thread_ts=1713815916.529489&cid=C03MFATU265

I tend to agree with you, that we would make this consistent. However, @npolshakova raised some good points about configurability. I figure we don't exactly know what this will look like, so I opted to make it configurable for now

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: I like it! If there are other code changes, I'm happy to make this now. If not, would you be open to one of us pushing this up in a follow-up PR?

test/kubernetes/e2e/features/route_options/route_options.go
npolshakova: If we have a generic pod that's reused, we're going back to the `testserver` pattern which limits the ability to make the curls from different namespaces, with different service accounts, etc. 

Allowing the pod metadata to be passed in EphemeralCurlEventuallyResponds also allows us to eventually paralyze the setups (each test can own its own curl pod in `test-id-curl` namespace, and use that for testing to ensure isolation).

test/kubernetes/testutils/assertions/curl.go
npolshakova: It's not currently used, but was originally introduced instead of `EphemeralCurlEventuallyResponds`. I'm fine deleting it until we have a use for it, but the idea here is that not all curl requests should be done from an ephemeral container. You can image port-forwarding the gateway/envoy admin page/etc., make a response and want to assert it with the matchers.HttpResponse.

npolshakova: I'm not sure if I agree with the comment- `We recommend that developers rely on the typed EphemeralCurlEventuallyResponds` - it really depends on what you're testing?

test/kubernetes/testutils/assertions/curl.go
npolshakova: I think EphemeralCurlEventuallyResponds is already doing two things: 1) It's responsible for building the ephemeral container to curl from, 2) asserting the curl response is eventually correct. 

If we want to keep these discrete assertions, the `EphemeralCurlEventuallyResponds` shouldn't be responsible for attaching the debug curl container. If we want to simplify the dev experience have it attach the debug container, it should also validate that 1) the pod it's attaching the debug container to exists, and 2) it's been successfully attached.

jenshu (dismissed):

test/kubernetes/testutils/assertions/curl.go
sam-heilbron: updated in https://github.com/solo-io/gloo/pull/9395/commits/6f9f1fbd65430e07f88617b35c10777b6c420999

jenshu (dismissed):

npolshakova (approved):

PR #9396, opened 2024-04-23
API for configuring GCP upstreams

EItanya:
# Description

This PR adds an API for configuring upstreams which can communicate securely with GCP Apis.

## API changes
- Added new `gcp` upstream type. This will enable authenticating requests with tokens from the google metadata service before sending them upstream.

## Docs changes
- Added generated docs for new `gcp` upstream typ .

# Context

Users have been requesting the ability to use gloo as the authentication point for their cloud run services.


ashishb-solo (dismissed):

nfuden (commented):
I think the only controversial thing here is whether we need to add this to the enterprise warn https://github.com/solo-io/gloo/blob/v1.16.x/projects/gloo/pkg/plugins/enterprise_warning/plugin.go#L2

EItanya:
Happy to add it there, I think that's fair

nfuden (approved):

ashishb-solo (approved):

PR #9397, opened 2024-04-23
Fix failure dump bugs

bewebi:
# Description

- Use the passed `path` to `recordEnvoyAdminData()` as an argument to `GetEnvoyAdminData()` instead of hard-coding one particular path
- Add any error returned when retrieving logs from pods to a multi-error and return after all pods have been dumped rather than always short-circuit the `for` loop for logging pods on the first iteration

# Context

We observed that the output in each file of the `envoy-dump` directory in failed test artifacts was the same despite claiming to come from different endpoints

We also observed that since #9350, only one pod's logs were being dumped

# Testing steps

To validate these fixes we can create a new branch off of this one, modify a test to ensure it fails, open a PR so that CI runs, and observe that the generated artifact contains all expected output

This has not (yet) been done as the fixes are seemingly straightforward enough to be validated through static code analysis

# Checklist:

### DIFF
--- a/test/helpers/kube_dump.go
+++ b/test/helpers/kube_dump.go
@@ -304,7 +304,7 @@ func recordEnvoyAdminData(f *os.File, path, proxyName, namespace string) {
 	defer f.Close()
 
 	fmt.Printf("Getting envoy for %s.%s and storing config dump: %s\n", proxyName, namespace, path)
-	cfg, err := gateway.GetEnvoyAdminData(context.TODO(), proxyName, namespace, "/config_dump", 30*time.Second)
+	cfg, err := gateway.GetEnvoyAdminData(context.Background(), proxyName, namespace, path, 30*time.Second)
 	if err != nil {
 		f.WriteString("*** Unable to get envoy " + path + " dump ***. Reason: " + err.Error() + " \n")
 		return
--- a/test/helpers/kube_dump.go
+++ b/test/helpers/kube_dump.go
@@ -166,6 +166,8 @@ func recordPods(podDir, namespace string) error {
 
 		logs, errOutput, err := kubeLogs(namespace, pod)
 		// store any error running the log command to return later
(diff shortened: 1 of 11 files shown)


sam-heilbron (dismissed):

test/helpers/kube_dump.go
```diff
@@ -155,27 +157,33 @@ func recordPods(podDir, namespace string) error {
...
 	for _, pod := range pods {
 		if err := os.MkdirAll(podDir, os.ModePerm); err != nil {
 			return err
 		}
 
-		f := fileAtPath(filepath.Join(podDir, pod+".log"))
-		errF := fileAtPath(filepath.Join(podDir, pod+"-error.log"))
 		logs, errOutput, err := kubeLogs(namespace, pod)
-
+		// store any error running the log command to return later
```
sam-heilbron, line 168: nit: I find that this comment repeats what the code below it does, but misses the chance to clarify _why_ we do this. 

For example, a question I have is: what type of error would be returned here and why wouldn't we write it to the errOutput if it occurs? If an err is returned, what is the desired impact on the execution of this function (ie why don't we return right away). These are behaviors that I think are useful to developers while reading the code, but can't easily be inferred from the code itself. 

One way to capture this would be:
```
We make a best effort to capture logs from Kubernetes Pods. We store logs from stdout and stderr in separate files, to make debugging easier. This code is invoked on test failure, and if we fail to capture logs, we want to aggregate that and return it to the caller, but we do not want to halt capturing logs from other pods
```

sam-heilbron (dismissed):

ashishb-solo (approved):

ashishb-solo (approved):

bewebi:
Documented gateway regression flake [here](https://github.com/solo-io/gloo/issues/7795#issuecomment-2075544667)

sam-heilbron (approved):

sam-heilbron:
comment to trigger bulldozer

PR #9398, opened 2024-04-24
Add port routing e2e tests

npolshakova:
Based on Istio regression varied ports e2e test: https://github.com/solo-io/gloo/blob/main/test/kube2e/istio/istio_integration_test.go#L268 

These tests are not Istio-specific and will be run with non-Istio Gloo Gateway manifest and the Istio setup.

The Istio regression tests will be deleted once all tests are ported to the new framework as part of https://github.com/solo-io/solo-projects/issues/6048


sam-heilbron (commented):
I like this, and the comments in the manifests make it easy to follow what is being tested here. A few minor thoughts/questions:
- Should we link in the PR body what tests originally inspired these and why they are not istio specific
- Can we clarify in the PR body what our intention is with the previous tests that existed (will we delete them eventually, if so, at what point)
- Should we have tests that demonstrate this using edge gw resources as well or does that not provide us any extra assurances

sam-heilbron (dismissed):

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
```diff
@@ -57,4 +58,8 @@ func TestK8sGateway(t *testing.T) {
 	t.Run("RouteOptions", func(t *testing.T) {
 		suite.Run(t, route_options.NewTestingSuite(ctx, testInstallation))
 	})
+
+	t.Run("Port Routing", func(t *testing.T) {
```
sam-heilbron, line 62: https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/example/error_logging_test.go#L51
```suggestion
	t.Run("PortRouting", func(t *testing.T) {
```

test/kubernetes/e2e/features/port_routing/types.go
```diff
@@ -0,0 +1,58 @@
...
+	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+
+	testService = &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-svc",
+			Namespace: "default",
+		},
+	}
+
+	// curlPod is the Pod that will be used to execute curl requests, and is defined in the fault injection manifest files
```
sam-heilbron, line 42:
```suggestion
	// curlPod is the Pod that will be used to execute curl requests, and is defined in the setup.yaml manifest file
```

test/kubernetes/e2e/features/port_routing/types.go
```diff
@@ -0,0 +1,58 @@
...
+var (
+	setupManifest = filepath.Join(util.MustGetThisDir(), "inputs/setup.yaml")
+
+	invalidPortAndValidTargetportManifest   = filepath.Join(util.MustGetThisDir(), "inputs/invalid-port-and-valid-targetport.yaml")
+	invalidPortAndInvalidTargetportManifest = filepath.Join(util.MustGetThisDir(), "inputs/invalid-port-and-invalid-targetport.yaml")
+	matchPodPortWithoutTargetportManifest   = filepath.Join(util.MustGetThisDir(), "inputs/match-pod-port-without-targetport.yaml")
+	matchPortandTargetportManifest          = filepath.Join(util.MustGetThisDir(), "inputs/match-port-and-targetport.yaml")
+	invalidPortWithoutTargetportManifest    = filepath.Join(util.MustGetThisDir(), "inputs/invalid-port-without-targetport.yaml")
+
+	// When we apply the deployer-provision.yaml file, we expect resources to be created with this metadata
```
sam-heilbron, line 27: This is out of date. This configuration is now distributed between `setup.yaml` and the test manifest file. I'm not sure how you want to clarify this

test/kubernetes/e2e/features/port_routing/suite.go
```diff
@@ -0,0 +1,140 @@
...
+
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+	s.testInstallation.Assertions.AssertEventualCurlResponse(
+		s.ctx,
+		curlPodExecOpt,
+		[]curl.Option{
+			curl.WithHost(kubeutils.ServiceFQDN(proxyService.ObjectMeta)),
+			curl.WithHostHeader("example.com"),
+		},
+		expectedServiceUnavailableResponse)
```
sam-heilbron, line 119: In this case, would we expect any other feedback provided to the user? Not sure if there is a status on resource

test/kubernetes/testutils/assertions/objects.go
```diff
@@ -15,7 +16,7 @@ func (p *Provider) EventuallyObjectsExist(ctx context.Context, objects ...client
 	for _, o := range objects {
 		p.Gomega.Eventually(ctx, func(innerG Gomega) {
 			err := p.clusterContext.Client.Get(ctx, client.ObjectKeyFromObject(o), o)
-			innerG.Expect(err).NotTo(HaveOccurred(), "object should be available in cluster")
+			innerG.Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("object %v %s.%s. should be available in cluster", o.GetObjectKind().GroupVersionKind().String(), o.GetName(), o.GetNamespace()))
```
sam-heilbron, line 19: I like this!

Given that apimachinery using namespace/name: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L38, should we attempt to at least have the namespace represented first?

Given that we use `ObjectKeyFromObject`, we could even use that and call `String()` on that objectKey?

test/kubernetes/testutils/assertions/objects.go
```diff
@@ -28,16 +29,16 @@ func (p *Provider) EventuallyObjectsNotExist(ctx context.Context, objects ...cli
...
+			WithTimeout(time.Second * 20).
 			WithPolling(time.Millisecond * 200).
 			Should(Succeed())
 	}
 }
 
 func (p *Provider) ExpectNamespaceNotExist(ctx context.Context, ns string) {
 	_, err := p.clusterContext.Clientset.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
-	p.Gomega.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "namespace should not be found in cluster")
+	p.Gomega.Expect(apierrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("namespace %s should not be found in cluster", ns))
```
sam-heilbron, line 43: Gomega actually calls fmt for us: https://github.com/onsi/gomega/blob/master/internal/assertion.go#L86, I think we can just pass in the format string as the first argument, and then any other properties 
```suggestion
	p.Gomega.Expect(apierrors.IsNotFound(err)).To(BeTrue(), "namespace %s should not be found in cluster", ns)
```

sam-heilbron (commented):
Looking good! All small comments/thoughts

test/kubernetes/e2e/features/port_routing/suite.go
npolshakova: Not really- we could check the status on the HTTPRoute (and eventually VirtualService when we add the classic edge API version), but the original port tests don't check this: https://github.com/solo-io/gloo/blob/main/test/kube2e/istio/istio_integration_test.go#L268 

Statuses are also notoriously flakey for VirtualServices and Upstreams in the regression tests, so I'd rather only check resource statuses when we really need to (checking that route options don't regress, etc.). The conformance tests and existing regression tests should validate statuses already.

test/kubernetes/e2e/features/port_routing/suite.go
```diff
@@ -0,0 +1,140 @@
...
+
+func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
+	return &portRoutingTestingSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+	}
+}
+
+func (s *portRoutingTestingSuite) SetupSuite() {
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, setupManifest)
```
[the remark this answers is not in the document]
npolshakova: When do we want to assert objects are created? Our current route options and deployers tests don't assert every object in the manifest is created, it is usually focused on the gateway deployment/svc which Gloo Edge's deployer is responsible for creating. I'm not sure what the benefit of checking if nginx/curl are created, but I agree we should check the Gateway resource is created.

test/kubernetes/e2e/features/port_routing/suite.go
```diff
@@ -0,0 +1,140 @@
...
+	s.Assert().NoError(err, "can apply setup manifest")
+}
+
+func (s *portRoutingTestingSuite) TearDownSuite() {
+	err := s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, setupManifest)
+	s.NoError(err, "can delete setup manifest")
```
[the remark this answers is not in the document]
npolshakova: I copied a lot of these assertions from the route options test- which one do we want to use? I don't have a preference but agree we should be consistent.

test/kubernetes/e2e/features/port_routing/suite.go
```diff
@@ -0,0 +1,140 @@
...
+	s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, proxyService, proxyDeployment)
```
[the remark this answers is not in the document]
npolshakova: Nope- the Gateway deployer creates gateways regardless if there's an HTTPRoute. So the flow is:

1. Create GW
2. Proxy provisioned
--- (per test config) ---
3. Attach HttpRoute with different port/targetport definition per test
4. Remove HttpRoute, proxy still exists without any routes
--- 
5. Delete GW
6. Proxy de-provisioned

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
npolshakova: Related to this- do you have any suggestions for a better name for the "feature" that "PortRouting"?

test/kubernetes/testutils/assertions/objects.go
npolshakova: We can- GME has objects logged as "name.namespace", but if Edge has "namespace.name" we can switch the order.

npolshakova: I'll update though, it looks like `ObjectKeyFromObject` is used in other places and will make the line a little easier to read

test/kubernetes/e2e/features/port_routing/suite.go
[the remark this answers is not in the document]
npolshakova: Added the gw service and deployment assertion in: https://github.com/solo-io/gloo/pull/9398/commits/648bd2ab464a74a6e503066d5fdc27cd460c7b19

test/kubernetes/e2e/features/port_routing/suite.go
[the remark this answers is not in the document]
npolshakova: Ended up changing everything to `s.NoError` in: https://github.com/solo-io/gloo/pull/9398/commits/648bd2ab464a74a6e503066d5fdc27cd460c7b19

test/kubernetes/e2e/features/port_routing/suite.go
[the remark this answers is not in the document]
npolshakova: Added a comment: https://github.com/solo-io/gloo/pull/9398/commits/648bd2ab464a74a6e503066d5fdc27cd460c7b19

sam-heilbron (approved):

test/kubernetes/testutils/assertions/objects.go
```diff
@@ -15,7 +16,7 @@ func (p *Provider) EventuallyObjectsExist(ctx context.Context, objects ...client
 	for _, o := range objects {
 		p.Gomega.Eventually(ctx, func(innerG Gomega) {
 			err := p.clusterContext.Client.Get(ctx, client.ObjectKeyFromObject(o), o)
-			innerG.Expect(err).NotTo(HaveOccurred(), "object should be available in cluster")
+			innerG.Expect(err).NotTo(HaveOccurred(), "object %v %s should be available in cluster", o.GetObjectKind().GroupVersionKind().String(), client.ObjectKeyFromObject(o).String())
```
nfuden, line 19:
```suggestion
			innerG.Expect(err).NotTo(HaveOccurred(), "object %s %s should be available in cluster", o.GetObjectKind().GroupVersionKind().String(), client.ObjectKeyFromObject(o).String())
```

test/kubernetes/testutils/assertions/objects.go
nfuden: Was this the case or is there a reason we resolved out this suggestion?

test/kubernetes/e2e/features/port_routing/inputs/invalid-port-and-invalid-targetport.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
```
nfuden, line 1: https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/cmd/go/internal/test/test.go;l=73

Can we move inputs into a testdata folder?

npolshakova: Good idea! I'll change it here for the port routing tests, but I'll move other inputs files in a separate pr to avoid merge conflicts with other tests in flight though

npolshakova: https://github.com/solo-io/gloo/pull/9398/commits/a5f4845101b8fdac4518b6f0a535e140c3606f0c

sam-heilbron (approved):

npolshakova:
/skip-changelog Looks like the changelog bot is stuck:

image

nfuden (approved):

PR #9391, opened 2024-04-22
Status Reporting for VirtualHostOptions in Kube GW

jbohanon:
# Description

Add status reporting to VirtualHostOptions when attached to kube gateway resources using targetRef

## API changes
- Added `metadata` field to `HttpListenerReport_Error` and `VirtualHostReport_Error` resources
- ...

## Code changes
- Make the VirtualHostOptions plugin a StatusPlugin
- ...

## Interesting decisions
 
Non-attached VHO are reported as warnings because...

## Testing steps

Kube e2e

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6044

PR #9399, opened 2024-04-24
Headless svc e2e test

npolshakova:
Based on Istio headless svc tests: https://github.com/solo-io/gloo/blob/main/test/kube2e/istio/istio_integration_test.go#L302

These tests are not Istio-specific and will be run with non-Istio Gloo Gateway manifest and the Istio setup. 

The Istio regression tests will be deleted once all tests are ported to the new framework as part of https://github.com/solo-io/solo-projects/issues/6048 

Follow up pr to generate manifests instead of hard coding values (installation namespace, host name, etc.): https://github.com/solo-io/gloo/pull/9416


test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,90 @@
...
+	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+
+	curlPod = &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "curl",
+			Namespace: "curl",
+		},
+	}
+
+	curlFromPod = func(ctx context.Context) func() string {
```
sam-heilbron, line 41: nit: Once https://github.com/solo-io/gloo/pull/9395 merges, I think we can simplify this to use the shared assertion, or at least to use the utility to build the FQDN

sam-heilbron (approved):

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
```diff
@@ -32,4 +32,7 @@ gloo:
     livenessProbeEnabled: true
 gatewayProxies:
   gatewayProxy:
-    healthyPanicThreshold: 0
\ No newline at end of file
+    healthyPanicThreshold: 0
+# Disable discovery, Gloo gateway doesn't need this
```
sam-heilbron, line 36: nit: And I might even add that it is not recommended in production. Also, in this case do you mean the "kube gateway integration" doesn't need this?

test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,143 @@
...
+					Do: &operations.BasicOperation{
+						OpName:   fmt.Sprintf("apply-manifest-%s", filepath.Base(k8sApiRoutingManifest)),
+						OpAction: installation.Actions.Kubectl().NewApplyManifestAction(k8sApiRoutingManifest),
+						OpAssertions: []assertions.ClusterAssertion{
+							// First check resources are created for Gateway
+							installation.Assertions.ObjectsExist(k8sApiproxyService, k8sApiProxyDeployment),
+
+							// Check headless svc can be reached
+							installation.Assertions.EphemeralCurlEventuallyResponds(curlPod, []curl.Option{
+								curl.WithHost(fmt.Sprintf("%s.%s.svc.cluster.local", k8sApiProxyDeployment.GetName(), k8sApiProxyDeployment.GetNamespace())),
```
sam-heilbron, line 92: Can we use https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/dns.go#L16 to generate the host instead? And just pass the deployment meta?

test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,143 @@
...
+				}
+			} else {
+				routingResourceOp = operations.ReversibleOperation{
+					Do: &operations.BasicOperation{
+						OpName:   fmt.Sprintf("apply-manifest-%s", filepath.Base(classicApiRoutingManifest)),
+						OpAction: installation.Actions.Kubectl().NewApplyManifestAction(classicApiRoutingManifest),
+						OpAssertions: []assertions.ClusterAssertion{
+							// Check headless svc can be reached
+							installation.Assertions.EphemeralCurlEventuallyResponds(curlPod, []curl.Option{
+								curl.WithHost(fmt.Sprintf("%s.%s.svc.cluster.local", defaults.GatewayProxyName, installation.Metadata.InstallNamespace)),
```
sam-heilbron, line 115: same comment as above about building the host

test/kubernetes/e2e/features/headless_svc/inputs/classic_api.yaml
```diff
@@ -0,0 +1,31 @@
...
+        routeAction:
+          single:
+            upstream:
+              name: default-headless-example-svc-8080
+              namespace: classic-edge-test # TODO: how to make this configurable?
+---
+apiVersion: gloo.solo.io/v1
+kind: Upstream
+metadata:
+  name: default-headless-example-svc-8080
```
sam-heilbron, line 23: nit: This Upstream name follows the pattern/structure of upstreams that the Control Plane discovers/creates. It might be easier to debug if this were named in a way that did not follow that pattern. What about something like `headless-nginx`

test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,143 @@
...
+								curl.WithHostHeader("headless.example.com"),
```
sam-heilbron, line 116: nit: This exposes a good challenge for our tests. This host header has meaning and must match the domain in the VS. Do you think a comment here is the best way to explain that to devs?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
npolshakova: Yep, I think I'll just make the comments consistent because the real reason we're disabling it is it's not recommended for production

test/kubernetes/e2e/features/headless_svc/headless_svc.go
jbohanon: If we're touching it anyway, can we expose a `ClusterInternalServiceFQDN` alongside a more generalized `ServiceFQDN` to make the centralized util more versatile outside of testing applications?

test/kubernetes/e2e/classicedge/classic_edge_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/solo-io/skv2/codegen/util"
+)
+
+var _ = Describe("Classic Edge Test", Ordered, func() {
+
+	var (
+		ctx context.Context
+
+		// testInstallation contains all the metadata/utilities necessary to execute a series of tests
+		// against an installation of Gloo Edge
```
jbohanon, line 21: Nit: 
```suggestion
		// against an installation of Gloo Gateway
```

test/kubernetes/e2e/classicedge/classic_edge_test.go
jbohanon, line 1: can we standardize this file's with the deployment yaml on `classic_edge_test` or `classic_gateway_test` or `edge_gateway_test`? not super picky on which though I think I prefer the lattermost

test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,143 @@
...
+					OpAssertion: func(ctx context.Context) {
+						// Check resources are deleted for headless svc
+						installation.Assertions.ObjectsNotExist(headlessService)
+					},
+				},
+			}
+
+			var routingResourceOp operations.ReversibleOperation
+			if useK8sApi {
+				routingResourceOp = operations.ReversibleOperation{
```
jbohanon, line 82: nit: could we pull these `ReversibleOperation`s out into their own vars to make the if/else here a little easier to read?

test/kubernetes/e2e/k8sgateway/deployer_test.go
jbohanon, line 1: can we go ahead and rename this file now since we're introducing the analogous edge gateway test suite?

test/kubernetes/e2e/features/headless_svc/headless_svc.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	expectedHealthyResponse = &testmatchers.HttpResponse{
+		StatusCode: http.StatusOK,
+		Body:       ContainSubstring("Welcome to nginx!"),
+	}
+)
+
+var ConfigureRoutingHeadlessSvc = func(useK8sApi bool) e2e.Test {
+	return e2e.Test{
```
jbohanon, line 57: nit: I found this kinda hard to read, and I am even pretty intimately familiar with the framework. I think readability would be improved by breaking out constituent chunks as much as possible to keep the `e2e.Test` definition as small as it can be. wdyt?

npolshakova: Do you have an example? I think an issue with the test framework is you can't group things by ginkgo `By`/`DescribeTables`/`Context`/etc.  So the only chunks that can be broken up into are separate `installation.Operator.ExecuteReversibleOperations`- which this does with the setup and separate routing config. I'm open to changing this because this is the first example with both k8s API resources and classic edge resources, but I couldn't really see an easier way to do this with the current framework.

test/kubernetes/e2e/classicedge/classic_edge_test.go
npolshakova: You mean rename the `deployer_test`? I'm on board, but I think Sam was going to do that in his PR?

test/kubernetes/e2e/classicedge/classic_edge_test.go
npolshakova: I want to clarify that this installs with kubeGateway disabled:
```
kubeGateway:
  # This is the field that enables the K8s Gateway Integration in Gloo Gateway
  enabled: false
```

I can update the comment though to `against an installation of Gloo Gateway with kubeGateway disabled`

test/kubernetes/e2e/classicedge/manifests/classic-gateway-test-helm.yaml
```diff
@@ -0,0 +1,31 @@
...
+  persistProxySpec: true
+  logLevel: info
+  validation:
+    allowWarnings: true
+    alwaysAcceptResources: false
+kubeGateway:
+  # This is the field that enables the K8s Gateway Integration in Gloo Gateway
+  enabled: false
+gloo:
+  logLevel: info
```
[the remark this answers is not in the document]
npolshakova: The current `k8s-gateway-test-helm.yaml` have it as log. I think it's because we wanted the manifests to be reflective of customer environments. 

I think if you're debugging a test you would want to also have `SKIP_CLEANUP` as an option and then look at the debug logs in the running installation. Having it default to `debug` doesn't add much to the tests themselves running in ci.

test/kubernetes/e2e/classicedge/manifests/classic-gateway-test-helm.yaml
```diff
@@ -0,0 +1,31 @@
...
+settings:
+  create: true
+  invalidConfigPolicy:
+    replaceInvalidRoutes: true
+    invalidRouteResponseCode: 404
+    invalidRouteResponseBody: Gloo Gateway has invalid configuration.
+gateway:
+  enabled: true
+  persistProxySpec: true
+  logLevel: info
```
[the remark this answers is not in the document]
npolshakova: I think this is the same issue as: https://github.com/solo-io/gloo/pull/9399#discussion_r1579680316

test/kubernetes/e2e/features/headless_svc/headless_svc.go
npolshakova: Let's add `ClusterInternalServiceFQDN` when we need it, I think `ServiceFQDN` should cover all the current test cases that use `curl.WithHost(fmt.Sprintf("%s.%s.svc.cluster.local", k8sApiProxyDeployment.GetName(), k8sApiProxyDeployment.GetNamespace()))` for now?

test/kubernetes/e2e/classicedge/classic_edge_test.go
jbohanon: Yeah I think that wording would be more clear

test/kubernetes/e2e/classicedge/classic_edge_test.go
jbohanon: In this comment, I was referring to the `classic_edge_test` and having the same name as the manifest it utilizes

test/kubernetes/e2e/k8sgateway/deployer_test.go
jbohanon: This was the comment where I was suggesting renaming `deployer_test.go` but as you say Sam intends to do it in his PR, resolving

test/kubernetes/e2e/classicedge/manifests/classic-gateway-test-helm.yaml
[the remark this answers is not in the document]
jbohanon: Yeah I had added [this log dump](https://github.com/solo-io/gloo/pull/9391/files#diff-60b09b6e73b4b5d3354e9e2eef1098b9632fac93753434815cdae91f56752f47) into my PR because I spent an uncomfortable amount of time debugging the routing action when it was due to a crashlooping gloo pod, so I figured defaulting this to debug would help. I don't think the logging would have a material effect on the test being comparable to user environment, but I'm not super worried about it either since
a. in the case of a panic/crashloop, _any_ verbosity should show it
b. `SKIP_CLEANUP` exists as you say, which I always forget about/struggle to get working properly

test/kubernetes/e2e/features/headless_svc/headless_svc.go
npolshakova: Is this easier to read? I think it's just personal preference, but I like have `err := installation.Operator.ExecuteReversibleOperations(ctx, commonSetup, routingResourceOp)` with a common routingResourceOp:

```
k8sRoutingResourceOp := operations.ReversibleOperation{
				Do: &operations.BasicOperation{
					OpName:   fmt.Sprintf("apply-manifest-%s", filepath.Base(k8sApiRoutingManifest)),
					OpAction: installation.Actions.Kubectl().NewApplyManifestAction(k8sApiRoutingManifest),
					OpAssertions: []assertions.ClusterAssertion{
						// First check resources are created for Gateway
						installation.Assertions.ObjectsExist(k8sApiproxyService, k8sApiProxyDeployment),

						// Check headless svc can be reached
						installation.Assertions.EphemeralCurlEventuallyResponds(curlPod, []curl.Option{
							curl.WithHost(kubeutils.ServiceFQDN(k8sApiProxyDeployment.ObjectMeta)),
							// The host headermust match the domain in the HTTPRoute
							curl.WithHostHeader("headless.example.com"),
							curl.WithPort(80),
						}, expectedHealthyResponse),
					},
				},
				Undo: &operations.BasicOperation{
					OpName:   fmt.Sprintf("delete-manifest-%s", filepath.Base(k8sApiRoutingManifest)),
					OpAction: installation.Actions.Kubectl().NewDeleteManifestAction(k8sApiRoutingManifest),
					OpAssertion: func(ctx context.Context) {
						// Check resources are deleted for Gateway
						installation.Assertions.ObjectsNotExist(k8sApiproxyService, k8sApiProxyDeployment)
					},
				},
			}

			classicRoutingResourceOp := operations.ReversibleOperation{
				Do: &operations.BasicOperation{
					OpName:   fmt.Sprintf("apply-manifest-%s", filepath.Base(classicApiRoutingManifest)),
					OpAction: installation.Actions.Kubectl().NewApplyManifestAction(classicApiRoutingManifest),
					OpAssertions: []assertions.ClusterAssertion{
						// Check headless svc can be reached
						installation.Assertions.EphemeralCurlEventuallyResponds(curlPod, []curl.Option{
							curl.WithHost(kubeutils.ServiceFQDN(metav1.ObjectMeta{Name: defaults.GatewayProxyName, Namespace: installation.Metadata.InstallNamespace})),
							// The host headermust match the domain in the VS
							curl.WithHostHeader("headless.example.com"),
							curl.WithPort(80),
						}, expectedHealthyResponse),
					},
				},
				Undo: &operations.BasicOperation{
					OpName:   fmt.Sprintf("delete-manifest-%s", filepath.Base(classicApiRoutingManifest)),
					OpAction: installation.Actions.Kubectl().NewDeleteManifestAction(classicApiRoutingManifest),
					OpAssertion: func(ctx context.Context) {
						// Check classic edge resources are deleted
						installation.Assertions.ObjectsNotExist(
							&v1.Upstream{
								ObjectMeta: metav1.ObjectMeta{
									Namespace: "headless-nginx-upstream",
									Name:      "classic-edge-test",
								},
							},
						)
					},
				},
			}

			if useK8sApi {
				err := installation.Operator.ExecuteReversibleOperations(ctx, commonSetup, k8sRoutingResourceOp)
			} else {
				err := installation.Operator.ExecuteReversibleOperations(ctx, commonSetup, classicRoutingResourceOp)
			}
```

npolshakova: I'm fine with that approach if it's more readable to others, but this goes back to rethinking if there's a better way to do kube API vs. classic API in the same test.

test/kubernetes/e2e/k8sgateway/deployer_test.go
npolshakova: I'm fine doing it as part of this PR, but I think @sam-heilbron might be renaming it in a different PR?

test/kubernetes/e2e/classicedge/manifests/classic-gateway-test-helm.yaml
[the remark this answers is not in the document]
npolshakova: Nice! Yeah we should add the log dumps to ci failures then- do you want to do that update in your PR then? I think changing it here doesn't really "prove" it does anything without the skip cleanup or post-failure log dump.

test/kubernetes/e2e/features/headless_svc/headless_svc.go
jbohanon: Yeah that's along the lines of what I was thinking, and I agree with your comment
```go
                        var routingResourceOp operations.ReversibleOperation
			if useK8sApi {
                                routingResourceOp = k8sRoutingResourceOp
			} else {
                                routingResourceOp = classicRoutingResourceOp
			}
                        err := installation.Operator.ExecuteReversibleOperations(ctx, commonSetup, routingResourceOp)
```

npolshakova:
/skip-changelog Looks like the changelog bot is stuck:

image

test/kubernetes/e2e/features/headless_svc/inputs/classic_api.yaml
jbohanon, line 1: is this file still needed?

test/kubernetes/e2e/features/headless_svc/resources.go
```diff
@@ -0,0 +1,156 @@
...
+	soloapis_gloov1 "github.com/solo-io/solo-apis/pkg/api/gloo.solo.io/v1"
+	"github.com/solo-io/solo-apis/pkg/api/gloo.solo.io/v1/core/matchers"
+	soloapis_kubernetes "github.com/solo-io/solo-apis/pkg/api/gloo.solo.io/v1/options/kubernetes"
+	gloocore "github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+)
+
+var (
+	headlessSvcSetupManifest  = filepath.Join(util.MustGetThisDir(), "inputs/setup.yaml")
+	k8sApiRoutingManifest     = filepath.Join(util.MustGetThisDir(), "inputs/k8s_api.gen.yaml")
+	classicApiRoutingManifest = filepath.Join(util.MustGetThisDir(), "inputs/classic_api.gen.yaml")
```
jbohanon, line 22:
```suggestion
	headlessSvcSetupManifest  = filepath.Join(util.MustGetThisDir(), "inputs", "setup.yaml")
	k8sApiRoutingManifest     = filepath.Join(util.MustGetThisDir(), "inputs", "k8s_api.gen.yaml")
	classicApiRoutingManifest = filepath.Join(util.MustGetThisDir(), "inputs", "classic_api.gen.yaml")
```

test/kubernetes/e2e/features/headless_svc/types.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var (
+	// When we apply the manifest file, we expect resources to be created with this metadata
+	k8sApiProxyObjectMeta = metav1.ObjectMeta{
+		Name:      "gloo-proxy-gw",
+		Namespace: "default",
+	}
+	k8sApiProxyDeployment = &appsv1.Deployment{ObjectMeta: k8sApiProxyObjectMeta}
+	k8sApiproxyService    = &corev1.Service{ObjectMeta: k8sApiProxyObjectMeta}
```
jbohanon, line 22: nit:
```suggestion
	k8sApiProxyService    = &corev1.Service{ObjectMeta: k8sApiProxyObjectMeta}
```

test/kubernetes/e2e/features/headless_svc/suite.go
```diff
@@ -0,0 +1,102 @@
...
+
+		s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, k8sApiProxyDeployment, k8sApiproxyService)
+
+		s.testInstallation.Assertions.AssertEventualCurlResponse(
+			s.ctx,
+			curlPodExecOpt,
+			[]curl.Option{
+				curl.WithHost(kubeutils.ServiceFQDN(k8sApiproxyService.ObjectMeta)),
+				// The host header must match the hostnames in the HTTPRoute
+				curl.WithHostHeader("headless.example.com"),
```
jbohanon, line 83: since we are defining these resources in code now, can this be a const for enforced consstency?

test/kubernetes/e2e/utils/generate_resources.go
```diff
@@ -0,0 +1,79 @@
...
+func objectToYaml(obj client.Object) ([]byte, error) {
+	jsonBytes, err := json.Marshal(obj)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal resource to JSON")
+	}
+
+	yamlBytes, err := yaml.JSONToYAML(jsonBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to convert resource JSON to YAML")
+	}
```
jbohanon, line 53: can we use protoutils instead of mashalling through json to yam?

test/kubernetes/e2e/features/headless_svc/inputs/classic_api.yaml
npolshakova: Nope- it's replaced with the generated one. It looks like when the generated manifest branch merged the old files didn't get cleaned up for some reason.

test/kubernetes/e2e/utils/generate_resources.go
npolshakova: I don't think so- it's a client.Object not a proto.Message. But maybe I'm missing something?

Makefile
```diff
@@ -268,8 +268,10 @@ run-kube-e2e-tests: test
 #----------------------------------------------------------------------------------
 # Go Tests
 #----------------------------------------------------------------------------------
-GO_TEST_ENV ?=
-GO_TEST_ARGS ?= -cpu=4 -race
+GO_TEST_ENV ?= GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore
+# Testings flags: https://pkg.go.dev/cmd/go#hdr-Testing_flags
+# The default timeout for a suite is 10 minutes, but this can be overridden by setting the -timeout flag
```
sam-heilbron, line 273: nit: Perhaps we specify instead of what the default is, _why_ we chose 20 (aka arbitrary threshold so that we are notified when tests take too long)

test/kubernetes/e2e/classicedge/classic_api_test.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		testInstallation.UninstallGlooGateway(ctx, func(ctx context.Context) error {
+			return testHelper.UninstallGlooAll()
+		})
+		testCluster.UnregisterTestInstallation(testInstallation)
+	})
+
+	// Install Gloo Gateway with only classic APIs enabled
+	// If the env var SKIP_GLOO_INSTALL=true, installation will be skipped
```
sam-heilbron, line 48: I think this env variable was actually removed in recent work. In general, I like the idea of the coment, but I wonder if we're better off with a debugging guide that includes these details. That way the code (env variable) and comments don't drift?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
```diff
@@ -32,4 +32,7 @@ gloo:
     livenessProbeEnabled: true
 gatewayProxies:
   gatewayProxy:
-    healthyPanicThreshold: 0
\ No newline at end of file
+    healthyPanicThreshold: 0
+# Disable discovery, not recommended for production
```
shashankram, line 36: nit: a comment on why this is being disabled for the test would help

shashankram (approved):

test/kubernetes/e2e/features/headless_svc/resources.go
```diff
@@ -0,0 +1,156 @@
...
+						ServiceNamespace: headlessService.GetNamespace(),
+						ServicePort:      8080,
+					},
+				},
+			},
+		}
+
+		headlessVs := &v1.VirtualService{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "VirtualService",
```
sam-heilbron, line 53: nit: We'll likely be needing to use more of this in our tests, we probably could have Group/Version/Kind constants for edge resources

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
```diff
@@ -59,6 +60,10 @@ func TestK8sGateway(t *testing.T) {
 		suite.Run(t, route_options.NewTestingSuite(ctx, testInstallation))
 	})
 
+	t.Run("HeadlessSvc", func(t *testing.T) {
```
shashankram, line 63: Good way to make sure your test doesn't result in a timeout :D

test/kubernetes/e2e/features/headless_svc/suite.go
```diff
@@ -0,0 +1,102 @@
...
+type headlessSvcSuite struct {
+	suite.Suite
+
+	ctx context.Context
+
+	// testInstallation contains all the metadata/utilities necessary to execute a series of tests
+	// against an installation of Gloo Gateway
+	testInstallation *e2e.TestInstallation
+
+	useK8sGatewayApi bool
```
sam-heilbron, line 26: Should this live here? An alternative would be that this is defined on the testInstallation.Metadata

test/kubernetes/e2e/features/headless_svc/suite.go
```diff
@@ -0,0 +1,102 @@
...
+	return &headlessSvcSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+		useK8sGatewayApi: useK8sGatewayApi,
+	}
+}
+
+// SetupSuite generates manifest files for the test suite
+func (s *headlessSvcSuite) SetupSuite() {
+	if s.useK8sGatewayApi {
```
sam-heilbron, line 39: I'm not quite sure I understand all the value of putting both edge/k8s gw tests in the same suite. As it is, all of the logic is separated, so there is no shared code. 
I do see the value in the tests be co-located so that they grow/change at the same time.

I wonder if we would benefit from 2 suites within this feature, and the Suite functions have the same name?

Not a blocker, I just fiugre we'll be following this pattern more so want it to be clear to devs

jbohanon (approved):

test/kubernetes/e2e/features/headless_svc/resources.go
```diff
@@ -0,0 +1,156 @@
...
+import (
+	"path/filepath"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/solo-io/skv2/codegen/util"
+	v1 "github.com/solo-io/solo-apis/pkg/api/gateway.solo.io/v1"
```
sam-heilbron, line 12: Why do we use the APIs defined in solo-apis instead of in Gloo? I assume this is because those types are the ones compatible with client.Object?

I think this comes at a cost in that new fields in our API will only be availabe after we: merge gloo, release gloo, merge changes into solo-apis, upgrade solo-apis in Gloo.

Given that the client.Object is then just used to write the object to a manifest file, what if we just had a way to write gloo resources to a file?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-test-helm.yaml
npolshakova: Can add in a follow up, but basically it's not recommended for production and new classic API tests define static upstreams. I might actually disable discovery in the other helm manifests since the k8s gateway API doesn't need the discovered upstreams either.

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
npolshakova: Haha since it's taking 18min and 45sec I think we'll probably need to bump the `-timeout=20m` again soon...

Makefile
npolshakova: Will do in a follow up- I think 20min might be too slow anyway since it's already taking over 18min.

test/kubernetes/e2e/features/headless_svc/resources.go
npolshakova: Let's add that when we need it. So far only this test has needed generating resources so I'd like to keep it minimal until we have more use cases.

test/kubernetes/e2e/features/headless_svc/suite.go
npolshakova: I thought that's the reason we're defining separate suites per feature- so we can define test specific fields to pass in. I don't think all tests need a `useK8sGatewayApi` flag since currently most of them won't use it.

test/kubernetes/e2e/features/headless_svc/resources.go
npolshakova: We need the resources to be []client.Object in order to generate the manifests because manifests contain both gateway API, portal and edge resources. The non-api edge resources don't have the same format as client.Objects and would need custom renders for fields (the custom edge metadata field, the Kind, the Version, the "Spec", etc.).

I agree the solo-apis dependency isn't ideal, but the classic edge apis should be pretty stable and not have many net new changes.  

I think we can look into how feasible it is to have two separate manifest generations with the classic edge apis vs. k8s gateway apis and portal in a follow up. But currently, this (porting over old Istio integration regression tests with the classic api to the new framework) is the only use case for having to define the classic edge apis in code and generating. Most new tests will be using the Gateway API and this function should be sufficient.

test/kubernetes/e2e/features/headless_svc/suite.go
npolshakova: So I think this is a good thing to look at it the follow up Istio permissive/strict auth policy. It feels a little repetitive having two separate setups and two separate suites for headless svc routing, especially when the setup config is the same. Let's move the conversation to the [Istio peerauth policy e2e pr](https://github.com/solo-io/gloo/pull/9418) because that also has a similar issue and will also need the headless service setup added as a test case.

test/kubernetes/e2e/classicedge/classic_api_test.go
npolshakova: I think this is just a comment! The env still exists, but yeah we can remove it in the follow up pr that adds the classic Istio edge setup.

PR #9400, opened 2024-04-24
test/kube/e2e: introduce testify, create example suite

sam-heilbron:
# Description

Introduce an example test suite, which relies on testify, and demonstrates how we can re-use an existing testing framework to remove a lot of the code that our custom framework introduced.

# Context

Adds testify:  https://github.com/stretchr/testify/tree/master

Introduces an example suite, not run in CI, to demonstrate how to group and execute our tests

**Follow-Up Work**:
- Merge this PR, once there is agreement
- Convert existing tests to this structure, and get them passing in CI
- Write enterprise test suites to demonstrate executing these tests

## Interesting decisions
 -

## Testing steps
This example suite does not run in CI. However, I changed some code that is run by our other suites, and those run in CI. So we expect those to pass.

If you want to run the example suite, try out the following:

```
kind create cluster
```

```
RELEASED_VERSION=1.17.0-beta20 CLUSTER_NAME=kind go test ./test/kubernetes/e2e/example/ -v
```

## Notes for reviewers

Be sure to verify intended behavior by ...

Please proofread comments on ...

This is a complex PR and may require a huddle to discuss ...

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #9401, opened 2024-04-25
test/kube/e2e: convert existing e2e tests

sam-heilbron:
# Description
Convert existing e2e tests to rely on testify

# Context
Follow up to: https://github.com/solo-io/gloo/pull/9400

## Framework Changes
- Remove the 'Operator' concept. This was intended to provide a centralize way for users to invoke actions/assertions against a cluster. However, it added extra abstraction that made debugging more challenging.
- Remove the e2e.Test concept. This was replaced by a testify.Suite. By using an existing library, we do not have to re-invent the wheel. The goal of an e2e.Test was to allow our tests to be defined outside of our test Suites. Ginkgo did not support this, but testify does.
- Do not use callbacks/closures for Assertions and Actions. Instead of returning functions that the operator can execute, execute the functions directly. The callback model made it challenging to debug.
- Remove ginkgo references. Mixing ginkgo/testify causes challenges when running and debugging locally

## Enterprise Equivalent Changes
These changes are imported into the Enterprise tests, to demonstrate how they allow us to re-use test cases, with different installations:
https://github.com/solo-io/solo-projects/pull/6106/

## Interesting decisions
 -

## Testing steps
- Existing tests, which run in CI, all pass

If you want to run tests locally:
```
 CLUSTER_NAME=kind TEST_PKG=test/kubernetes/e2e/example RELEASED_VERSION=1.17.0-beta20 GO_TEST_USER_FLAGS=-v make go-test
```

## Notes for reviewers
-

# Checklist:


test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/solo-io/gloo/pkg/utils/envoyutils/admincli"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/runtime"
+	"time"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/stretchr/testify/suite"
+)
+
+// FeatureSuite is the entire Suite of tests for the "deployer" feature
+type FeatureSuite struct {
```
shashankram, line 15: s/FeatureSuite/featureSuite/g

Can be unexported as it implements the suite interface

test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -0,0 +1,98 @@
+package deployer
```
shashankram, line 1: the package name is confusing. What is this feature?

test/kubernetes/e2e/features/deployer/manifests.go
```diff
@@ -0,0 +1,30 @@
+package deployer
```
shashankram, line 1: nit: files with just types/vars can be named types.go.

test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (s *FeatureSuite) TestProvisionDeploymentAndService() {
+	s.T().Cleanup(func() {
+		err := s.testInstallation.Actions.Kubectl().Client().DeleteFile(s.ctx, deployerProvisionManifestFile)
+		s.NoError(err, "can delete manifest")
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, proxyService, proxyDeployment)
+	})
+
+	err := s.testInstallation.Actions.Kubectl().Client().ApplyFile(s.ctx, deployerProvisionManifestFile)
+	s.NoError(err, "can apply manifest")
```
shashankram, line 52: s.Require().NoError() so that it doesn't attempt to assert on L53. I don't know how Cleanup() hook works with Require(), worth testing.

test/kubernetes/e2e/features/deployer/manifests.go
shashankram, line 1: Where's this test invoked? Can we add a top level test in test/kubernetes/e2e/k8sgateway that invokes this suite with the existing install params in deployer_test.go?

I am thinking we should have something like basic_test.go in test/kubernetes/e2e/k8sgateway which runs a set of feature tests against a standardized install of Gloo

sam-heilbron: Oh yeah I just hadn't done that yet, but it's in progress

test/kubernetes/e2e/features/deployer/suite.go
sam-heilbron: good call!

test/kubernetes/e2e/features/deployer/suite.go
sam-heilbron: It matches the name of the feature it is testing: https://github.com/solo-io/gloo/tree/main/projects/gateway2/deployer. What would you recommend as an alternative? Or perhaps i add a comment linking the feature it's testing?

test/kubernetes/e2e/features/deployer/manifests.go
sam-heilbron: I like it!

test/kubernetes/e2e/features/deployer/suite.go
shashankram: I am not aware of the feature. Makes sense to keep it in sync though.

test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -0,0 +1,104 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/gloo/pkg/utils/envoyutils/admincli"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/runtime"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/stretchr/testify/suite"
+)
+
+// testingSuite is the entire Suite of tests for the "deployer" feature
+type testingSuite struct {
```
shashankram, line 17: could you include the feature name in the type, such as deployerSuite, then we can selectively run this suite or its sub-tests as `go test -run TestBasicInstallation/deployerSuite/TestProvisionDeploymentAndService`. That way, we have a less ambiguous test IDs.

I also don't mind dropping the "Suite" from the name and making it CamelCase for consistency.

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/stretchr/testify/suite"
+)
+
+// testingSuite is the entire Suite of tests for the "example" feature
+type testingSuite struct {
```
shashankram, line 20: s/testingSuite/routeOptionsSuite

test/kubernetes/e2e/test.go
```diff
@@ -5,21 +5,18 @@ import (
...
 	"github.com/solo-io/gloo/test/kubernetes/testutils/cluster"
 	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
 	"github.com/solo-io/gloo/test/kubernetes/testutils/runtime"
 
 	"github.com/solo-io/gloo/test/kubernetes/testutils/assertions"
-	"github.com/solo-io/gloo/test/kubernetes/testutils/operations"
 )
 
-func NewTestCluster() *TestCluster {
+func MustTestCluster(t *testing.T) *TestCluster {
```
shashankram, line 17: just panic instead of passing testing.T for test suite prereqs? This seems unrelated to a test instance and would happen regardless of a test run. The `Must` in the name indicates that it should abort on error.

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -15,68 +15,38 @@ import (
...
+	// for some useful-ish output
+	tick := time.Tick(pollingInterval)
+
+	p.Gomega.Eventually(func(g Gomega) {
+		res := p.clusterContext.Cli.CurlFromEphemeralPod(ctx, client.ObjectKeyFromObject(curlPod), curlOptions...)
+		select {
+		default:
+			break
+		case <-tick:
+			ginkgo.GinkgoWriter.Printf("want %v\nhave: %s", expectedResponse, res)
```
shashankram, line 41: Using a ticker here is a bit unnecessary. Can just log it on every response.

test/kubernetes/e2e/features/deployer/suite.go
sam-heilbron: I'm good with that. I had avoided it initially, because it felt like it stuttered to both have `deployer` package and in the name

test/kubernetes/testutils/actions/glooctl.go
```diff
@@ -0,0 +1,113 @@
...
+func (p *providerImpl) WithGlooGatewayContext(ggCtx *gloogateway.Context) Glooctl {
+	p.glooGatewayContext = ggCtx
+	return p
+}
+
+// requiresGlooGatewayContext is invoked by methods on the Provider that can only be invoked
+// if the provider has been configured to point to a Gloo Gateway installation
+// There are certain actions that can be invoked that do not require that Gloo Gateway be installed for them to be invoked
+func (p *providerImpl) requiresGlooGatewayContext() {
+	p.require.NotNil(p.glooGatewayContext, "Provider attempted to create an action that requires a Gloo Gateway installation, but none was configured")
```
shashankram, line 61: A bit confused about the use of assertions in helpers. I would expect to only have assertions used in functions that are in `assertions/`. This helps decouple test helpers from the underlying assertion lib being used.
Helpers should either return error so that callers can handle it or panic

test/kubernetes/e2e/features/deployer/suite.go
shashankram: Yes I agree, but since we are importing these tests into a different suite, the naming becomes important. Overall I feel it would make sense to ID the test as `OuterSuite/FeatureSuite/TestName`, as it's very easy to identify specific tests to invoke or identify on failures.
This wouldn't be necessary if these suites were invoked as regular tests, which they aren't.

test/kubernetes/e2e/features/route_options/suite.go
sam-heilbron: Per https://github.com/solo-io/gloo/pull/9401#discussion_r1580169268 it seems like we would go with RouteOptionsSuite, are you good with that?

test/kubernetes/testutils/actions/glooctl.go
sam-heilbron: Yeah that's fair, this felt a bit strange too, so happy to remove it. I will have this panic as it means the developer was misusing the function, and there is no recoverable state. I had opted to use testing patterns (assertions) to handle that panic instead of panicing directly

shashankram:
One thing that is a must is to get rid of Ginkgo dependencies in the new framework as it messes with Go tests when running it via the debugger. To achieve this, this is my suggestion:
1. Consolidate helpers and assertions into separate directories.
2. Avoid assertions in helpers and return an error or panic in MustXyz() variants
3. Allow assertions to accept either require/assert, i.e. leave it to the caller

The goal should be to avoid importing the ginkgo pkg and instead hook up gomega with testing.T

test/kubernetes/e2e/features/deployer/suite.go
shashankram: Seems like the naming of the struct doesn't matter as long as we name the suite when invoking t.Run() from the outer suite. So feel free to just use `type suite struct` here and pass the suite name as `t.run("Deployer", ...`

test/kubernetes/e2e/features/route_options/suite.go
shashankram: We can even drop the "Suite", so we would have Go test ID them as `TestBasicInstall/RouteOptions/TestSomething`

test/kubernetes/e2e/test.go
```diff
@@ -3,21 +3,40 @@ package e2e
...
 	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
 	"github.com/solo-io/gloo/test/kubernetes/testutils/runtime"
 
 	"github.com/solo-io/gloo/test/kubernetes/testutils/assertions"
-	"github.com/solo-io/gloo/test/kubernetes/testutils/operations"
 )
 
-func NewTestCluster() *TestCluster {
+var (
+	SkipGlooInstall = os.Getenv("SKIP_GLOO_INSTALL") == "true"
```
sam-heilbron, line 23: nit: I'm good with this work for now. I would propose that we make a couple of changes though:
1. I'd like this, and all env variables that are used for testing, be defined here: https://github.com/solo-io/gloo/blob/main/test/testutils/env.go. This provides a central place for devs to go to know what env variables are available for testing. Or we can do one just for kube tests
2. I think this should live on the runtime.Context, since it's details that a user/caller can set at runtime

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -179,3 +187,53 @@ func (c *Cli) CurlFromEphemeralPod(ctx context.Context, podMeta types.Namespaced
...
+	stderr := new(strings.Builder)
+
+	err := cmdutils.Command(ctx, "kubectl", args...).
+		// For convenience, we set the stdout and stderr to the receiver
+		// This can still be overwritten by consumers who use the commands
+		WithStdout(stdout).
+		WithStderr(stderr).Run().Cause()
+
+	return stdout.String(), stderr.String(), err
+}
```
sam-heilbron, line 239: These are duplicates of code above. Not critical for solving right now, but I think we should consolidate

test/kubernetes/e2e/test.go
shashankram: Feel free to update. I just want to be able to run the tests against an existing cluster, including examples, without installing Gloo for a quick run.

pkg/utils/kubeutils/kubectl/cli.go
shashankram: There are some issues with the existing Command APIs. stdout and stderr should be returned and not be set on a common Cli pointer via its receiver

test/kube2e/util.go
```diff
@@ -200,9 +200,15 @@ func GetTestHelper(ctx context.Context, namespace string) (*helper.SoloTestHelpe
 	if err != nil {
 		return nil, err
 	}
+
+	rootDir := filepath.Join(cwd, "../../..")
```
shashankram, line 204: should use absolute repo path in the future

shashankram (approved):

test/kube2e/util.go
sam-heilbron: agreed. I was looking for a utility that me must have somewhere. Couldn't find it, but I imagine it exists or we should create it

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -179,3 +187,53 @@ func (c *Cli) CurlFromEphemeralPod(ctx context.Context, podMeta types.Namespaced
 		podMeta.Name,
 		curlArgs...)
 }
+
+// CurlFromPod executes a Curl request from the given pod for the given options
```
npolshakova, line 191: [nit] It's tricky to tell when to use `CurlFromEphemeralPod` vs. `CurlFromPod` with the current comments. Can we add a comment to `CurlFromEphemeralPod` that it attaches the curl container?

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -179,3 +187,53 @@ func (c *Cli) CurlFromEphemeralPod(ctx context.Context, podMeta types.Namespaced
...
+	// The e2e test assertions rely on the transforms.WithCurlHttpResponse to validate the response is what
+	// we would expect
+	// For this transform to behave appropriately, we must execute the request with verbose=true
+	appendOption(curl.VerboseOutput())
+
+	curlArgs := curl.BuildArgs(options...)
+
+	args := append([]string{
+		"exec",
+		"--container=curl",
```
npolshakova, line 206: How is this working if the curl container isn't present? For example, if I was using `netshoot` instead of the `curl` container the route options tests are currently spinning up this would fail. I think we also want to make the curl container configurable here? Or add a comment that a `curl` container needs to be present

test/kubernetes/e2e/example/basic_installation_test.go
```diff
@@ -20,31 +23,43 @@ func TestBasicInstallation(t *testing.T) {
 	RegisterFailHandler(Fail)
 
 	ctx := context.Background()
-	testCluster := e2e.NewTestCluster()
+	testCluster := e2e.MustTestCluster()
```
npolshakova, line 26: Why the rename here for `MustTestCluster` and `MustTestHelper`?

test/kubernetes/e2e/example/basic_installation_test.go
```diff
@@ -20,31 +23,43 @@ func TestBasicInstallation(t *testing.T) {
...
 	// We register the cleanup function _before_ we actually perform the installation.
 	// This allows us to uninstall Gloo Gateway, in case the original installation only completed partially
 	t.Cleanup(func() {
 		if t.Failed() {
-			testInstallation.PreFailHandler()
+			testInstallation.PreFailHandler(ctx)
 		}
 
-		testInstallation.UninstallGlooGateway(ctx, testInstallation.Actions.Glooctl().NewTestHelperUninstallAction())
+		testInstallation.UninstallGlooGateway(ctx, func(ctx context.Context) error {
```
npolshakova, line 45: Are there other clean up functions that can be called here instead of `UninstallGlooAll()`? Why not just pass the testHelper here?

test/kubernetes/e2e/example/complex_installation_test.go
```diff
@@ -21,31 +24,38 @@ func TestComplexInstallation(t *testing.T) {
...
-		testInstallation.InstallGlooGateway(ctx, testInstallation.Actions.Glooctl().NewTestHelperInstallAction())
+	t.Run("InstallGateway", func(t *testing.T) {
+		testInstallation.InstallGlooGateway(ctx, func(ctx context.Context) error {
+			return testHelper.InstallGloo(ctx, helper.GATEWAY, 5*time.Minute, helper.ExtraArgs("--values", testInstallation.Metadata.ValuesManifestFile))
+		})
 	})
 
-	t.Run("example feature", func(t *testing.T) {
-		suite.Run(t, example.NewFeatureSuite(ctx, testInstallation))
+	t.Run("Example", func(t *testing.T) {
```
npolshakova, line 58: [nit] I know this is just an example, but do want the names for the complex vs. basic example to be different here?

test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
+	return &testingSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+	}
+}
+
+func (s *testingSuite) SetupSuite() {
```
npolshakova, line 34: Do these (SetupSuite, TearDownSuite, BeforeTest, AfterTest) need to be present even if there are no setup/teardown/before/after operations?

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/stretchr/testify/suite"
+
+	"github.com/solo-io/gloo/pkg/utils/kubeutils"
+	"github.com/solo-io/gloo/pkg/utils/requestutils/curl"
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+)
+
+// testingSuite is the entire Suite of tests for the "Route Options" feature
+// RouteOptions
+type testingSuite struct {
```
npolshakova, line 20: So every feature needs to define it's own testingSuite struct?

test/kubernetes/testutils/assertions/status.go
```diff
@@ -16,23 +16,20 @@ import (
...
-			g.Expect(*status.GetStatuses()[installNamespace]).To(statusStateMatcher)
-		}, currentTimeout, pollingInterval).Should(gomega.Succeed())
-	}
+func (p *Provider) EventuallyResourceStatusMatchesState(
+	_ context.Context,
+	getter helpers.InputResourceGetter,
+	statusMatcher types.GomegaMatcher,
+	timeout ...time.Duration,
+) {
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
```
npolshakova, line 25: Where is the namespace assertion here?

test/kubernetes/testutils/clients/clients.go
```diff
@@ -1,63 +1,62 @@
...
+	mustNotError(err)
 
 	err = v1.AddToScheme(clientScheme)
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	mustNotError(err)
 
 	return clientScheme
 }
+
+func mustNotError(err error) {
```
npolshakova, line 58: Is there a way to run all the tests without triggering the panic? This seems like the tests only run as `failFast` now?

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -2,81 +2,41 @@ package assertions
...
-			expectedResponseMatcher := WithTransform(transforms.WithCurlHttpResponse, matchers.HaveHttpResponse(expectedResponse))
-			g.Expect(res).To(expectedResponseMatcher)
-			ginkgo.GinkgoWriter.Printf("success: %v", res)
-		}).
-			WithTimeout(currentTimeout).
-			WithPolling(pollingInterval).
-			WithContext(ctx).
-			Should(Succeed())
-	}
+func (p *Provider) AssertEventualCurlResponse(
```
npolshakova, line 17:
```suggestion
func (p *Provider) AssertEventuallyCurlResponse(
```

test/kubernetes/e2e/features/route_options/suite.go
shashankram: Yes, because this would be specific to a feature and you could customize it based on the testing requirements.

It's necessary to allow creating feature specific suites.

test/kubernetes/testutils/assertions/curl.go
shashankram: EventualCurlResponse reads better? Eventually is the underlying Ginkgo API that the client doesn't need to be aware of as we could swap this with a different assertion lib if necessary

test/kubernetes/testutils/clients/clients.go
shashankram: We could, but if this step fails there's no point running other tests as they would fail too

test/kubernetes/e2e/features/route_options/suite.go
sam-heilbron: I was also overly verbose to show what a testingSuite _can_ define. But utlimately, I don't think this is that much overhead compared to a set of e2e.Test functions. In essence each of those e2e.Tests are just methods on this struct, and now we can also have a standard, per feature

test/kubernetes/e2e/features/deployer/suite.go
shashankram: They don't need to be defined if you don't need them.

sam-heilbron: Nope! I was overly verbose to show the functionality available, but in a follow-up I'm happy to remove these from this feature, and leave them on the example feature

test/kubernetes/e2e/example/basic_installation_test.go
shashankram: Must will panic if it can't create the cluster, so a naming convention

test/kubernetes/testutils/clients/clients.go
npolshakova: There are instances where you want to see all test failures- especially if they're failing for a common reason. This isn't a blocker, but a new limitation with the changes. It's probably easy to fix though if we add an env `RunAllTests` (or some better name) that doesn't fail fast with a panic and just logs the error and continues,

pkg/utils/kubeutils/kubectl/cli.go
shashankram: Makes sense to default to `curl` and allow clients to override the container. This isn't different from the existing API though, but worth fixing.

test/kubernetes/e2e/example/complex_installation_test.go
sam-heilbron: Fair point! Would it be possible to handle this in a follow-up? Any suggestions? The goal was just to show that 2 installations, with different values, can run the same test

test/kubernetes/e2e/features/route_options/suite.go
npolshakova: Hm, can the examples be overly verbose, but we limit what the route options test uses to what it needs? Also any thoughts on renaming this from `testingSuite`? That sounds too generic if it's actually scoped only to route options.

test/kubernetes/testutils/assertions/status.go
sam-heilbron: Which one? https://github.com/solo-io/gloo/blob/sam/testify-e2e-part2/test/kubernetes/testutils/assertions/status.go#L29:
- We get the status from the object
- We inspect the particular status for this installation
- We assert it matches what the caller provided

I believe this matches the previous behavior, no?

test/kubernetes/e2e/features/deployer/suite.go
npolshakova: I think it would be good to remove them from the route options/deployer suites. I'm fine keeping them in the example.

test/kubernetes/testutils/assertions/status.go
npolshakova: We don't check if it's nil here? `nsStatus := status.GetStatuses()[p.glooGatewayContext.InstallNamespace]`. I think that's fine, but it'll fail the `g.Expect(nsStatus).ToNot(gomega.BeNil())` and I'm not sure it's clear that it's missing the namespace vs. not the status on the namespace is nil? 

```
status:
  statuses:
    gloo-system: {}
```

versus:

```
status:
  statuses: {}
```

Is there a difference between `installation.Metadata.InstallNamespace` and the namespace on the glooGatewayContext (`p.glooGatewayContext.InstallNamespace`).

test/kubernetes/e2e/example/basic_installation_test.go
npolshakova: But then shouldn't the uninstall assertions also be passed in by the user instead of using `EventuallyUninstallationSucceeded`? Or is the uninstall assertion always guaranteed to be the same across the tool being used to uninstall?

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/stretchr/testify/suite"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/solo-io/gloo/test/kubernetes/e2e/features/deployer"
+	"github.com/solo-io/gloo/test/kubernetes/e2e/features/route_options"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
+)
+
+// TestK8sGateway is the function which executes a series of tests against a given installation
+func TestK8sGateway(t *testing.T) {
```
npolshakova, line 23: Is this also going to move to the new framework?

test/kubernetes/e2e/example/complex_installation_test.go
shashankram: No, the tests will be qualified as `TestBasic/Example/TestFoo` and `TestComplex/Example/TestFoo` respectively for a TestFoo within a feature suite.

test/kubernetes/testutils/clients/clients.go
shashankram: These are pretty fixable. We can definitely iterate and avoid panics and let tests handle them

shashankram: FWIW these were panicking before as well, just via Gomega assertions instead

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
npolshakova: So this is the top level test that runs the features (route options, deployer, etc.). Are these tests run in order then? It feels strange having the gateway installation as a separate test instead of a before suite? 

```
	t.Run("InstallGateway", func(t *testing.T) {
		testInstallation.InstallGlooGateway(ctx, func(ctx context.Context) error {
			return testHelper.InstallGloo(ctx, helper.GATEWAY, 5*time.Minute, helper.ExtraArgs("--values", testInstallation.Metadata.ValuesManifestFile))
		})
	})
```

test/kubernetes/testutils/clients/clients.go
npolshakova: But now we're using `mustNotError` instead of `gomega.Expect(err).NotTo(gomega.HaveOccurred())`. The Expect() would fail, but then the next test would run.

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
npolshakova: Addressed in https://github.com/solo-io/gloo/pull/9401/commits/5928f4ac5477a5330f1f047af4b1f99b60d571d5

npolshakova (approved):

test/kubernetes/testutils/clients/clients.go
shashankram: I don't think that was the case because we were using the top level `TestK8sGateway testing.T` to initialize Ginkgo, so it would fail the whole test vs a subtest invoked by t.Run()

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
shashankram: removed, good call

shashankram (approved):

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
sam-heilbron: This is the "new" framework. In essence all we did was swap out the e2e.Test construct, for suite.Suite in testify.

test/kubernetes/testutils/assertions/status.go
sam-heilbron: Good point! Ill update in a follow-up

test/kubernetes/testutils/clients/clients.go
sam-heilbron: We will explore the right balance of failfast and failimmediately in follow-up

test/kubernetes/e2e/example/basic_installation_test.go
sam-heilbron: EventuallyUninstallationSucceeded just checks that the install ns is removed. I agree, if we find that each install has a requisite uninstall, then we'll need to expand this functionality

PR #9402, opened 2024-04-25
Add istio and sds values to GatewayParameters

npolshakova:
# Description 

Currently, the istio integration helm value is still set as an env and passed around. A user sets istioSDS.enabled which then configures the GG_EXPERIMENTAL_ISTIO_MTLS_SDS_ENABLED env on the [gloo control plane deployment](https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/1-gloo-deployment.yaml#L243). 
The gateway v2 deployer then uses the [IstioValues](https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/bootstrap/opts.go#L67) which are filled by the env values on the control plane deployment to inject the istio-proxy and sds sidecars.

Other than istioSDS.enabled none of the other Istio integration or SDS helm values are supported for the Gloo Gateway deployer. This is a current limitation in the k8s Gateway deployer and has caused issues related to running the Istio integration with gloo-ee.

Example GatewayParameters with all config set:
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
    sds:
      sdsContainer:
        sdsBootstrap:
          logLevel: debug
        image:
          image:
            registry: docker.io/sds
            repository: sds
            tag: 1.0.1-dev # this is a placeholder example
      istioIntegration:
        istioDiscoveryAddress: istiod.istio-system.svc.cluster.local:15012
        istioMetaMeshId: mesh
        istioMetaClusterId: cluster
        istioContainer:
          logLevel: debug
          image:
            image:
              registry: docker.io/istio
              repository: proxyv2
              tag: 1.19.2
          resources:
            requests:
              cpu: 100m
              memory: 128Mi
            limits:
              cpu: 200m
              memory: 256Mi
          securityContext:
            runAsUser: 10101
            runAsNonRoot: true
```

Example GatewayParameters config that only sets sds with default values:
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
    sds: {}
```

Example GatewayParameters config that sets Istio integration with default values:
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
    sds:
      istioIntegration: {}
```

Relies on timeout override for pod matches assertion added in https://github.com/solo-io/gloo/pull/9463 

## Code changes
- Adds Istio and SDS values to GatewayParameters 
- Adds deployer changes to use GatewayParameters overrides
- Adds e2e test to create k8s Gateway with Istio and SDS image overrides 

## Notes for reviewers 

Once https://github.com/solo-io/gloo/pull/9493 goes in, the old `IstioSDS` field in the deployer and env value that is set when `global.istioSDS.enabled=true` is set in the helm chart can be removed. 

I opened a separate issue to confirming the gloomtls behavior and convert the old gloomtls regression test into the new framework: https://github.com/solo-io/solo-projects/issues/6210 

Design discussion: https://docs.google.com/document/d/1oF8KpoIEEcux_R8fBrLkdKF0NaonINVedhdCHMuf2bg/edit?usp=sharing

# Checklist:


projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -109,6 +112,93 @@ message EnvoyBootstrap {
...
+  // Istio integration configuration
+  // If not provided, the default values will be used.
+  IstioIntegration istio_integration = 7;
+
+}
+
+// Configuration for the Istio integration settings used by a Gloo Gateway's data plane (Envoy proxy instance)
+message IstioIntegration {
+  // Enables SDS cert-rotator sidecar for istio mTLS cert rotation
+  bool enabled = 1;
```
jenshu, line 131: can we make this a BoolValue so we can differentiate between false and unset?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -109,6 +112,93 @@ message EnvoyBootstrap {
...
+
+// Configuration for the container running Gloo sds.
+message SdsContainer {
+  // The sds container image. See
+  // https://kubernetes.io/docs/concepts/containers/images
+  // for details.
+  //
+  // Default values, which may be overridden individually:
+  //   registry: quay.io/solo-io
+  //   repository: sds
```
jenshu, line 154: sds-ee for enterprise

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -109,6 +112,93 @@ message EnvoyBootstrap {
...
+  k8s.io.api.core.v1.SecurityContext security_context = 2;
+
+  // The compute resources required by this container. See
+  // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+  // for details.
+  kube.gateway.gloo.solo.io.ResourceRequirements resources = 3;
+
+  // Log level for istio-proxy. Options include "info", "debug", "warning", and "error".
+  // Default level is info Default is "warning".
+  string log_level = 4;
```
jenshu, line 199: (same question for log_level above too) - do you think there will be additional non-k8s-specific config that we would want to extract into a reusable message (similar to how EnvoyBootstrap is used in the envoy container above)?

jenshu (commented):
some API comments

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: Hm, the name `EnvoyBootstrap` implies it's setting the bootstrap config, not the container config. Are these container args here what that "bootstrap" config is responsible for setting?

```
        args:
        - proxy
        - sidecar
        - --domain
        - $(POD_NAMESPACE).svc.cluster.local
        - --configPath
        - /etc/istio/proxy
        - --serviceCluster
        - istio-proxy-prometheus
        - --drainDuration
        - 45s
        - --parentShutdownDuration
        - 1m0s
        - --proxyLogLevel={{ $global.glooMtls.istioProxy.logLevel }}
        - --proxyComponentLogLevel=misc:error
        - --connectTimeout
        - 10s
        - --controlPlaneAuthPolicy
        - NONE
        - --dnsRefreshRate
        - 300s
        - --trust-domain=cluster.local
        - --controlPlaneBootstrap=false
```

Then, yes- we could have an `IstioProxyBootstrap` grouping. But, right now there's nothing in the template that's directly configurable other than the logs. We probably couldn't reuse the `EnvoyBootstrap` directly though since there's no `component_log_levels` equivalent and the istio-proxy values could differ from the gloo-ee container values.

For the sds log level- that's set as an env. I'm not sure if we want to make customEnvs configurable on the sds container. They currently aren't and I'd prefer to not add any net new configurations in the first pass.

jenshu: oh, to clarify i didn't mean you should reuse the EnvoyBootstrap message here, but rather if you intend on adding more configurable parameters it might be good to wrap them in a message so they can be imported in multiple places (in case we later allow configuring gatewayparams on non-k8s environments). also perhaps the 'bootstrap' name was poorly chosen, but was meant to indicate any args that envoy starts with (in additional to actual envoy bootstrap config, if we decide to add that later)

npolshakova: Okay the more I think about it, I don't think this would work in non-k8s environments. Because all of these values are already specific to building the k8s container config:
```
  // The sds container image. See
  // https://kubernetes.io/docs/concepts/containers/images
  // for details.
  //
  // Default values for Edge OSS, which may be overridden individually:
  //   registry: quay.io/solo-io
  //   repository: sds / sds-ee
  //   tag: <gloo version> (OSS) / <gloo-ee version> (EE)
  //   pullPolicy: IfNotPresent
  kube.gateway.gloo.solo.io.Image image = 1;

  // The security context for this container. See
  // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#securitycontext-v1-core
  // for details.
  k8s.io.api.core.v1.SecurityContext security_context = 2;

  // The compute resources required by this container. See
  // https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
  // for details.
  kube.gateway.gloo.solo.io.ResourceRequirements resources = 3;

  // Log level for sds.  Options include "info", "debug", "warn", "error", "panic" and "fatal".
  // Default level is "info".
  string log_level = 4;
```

And the istio-proxy container assumes Istio is deployed in a cluster (it won't be able to reach out to Istio otherwise to get the cert). So the only non-kube field would be the log level. I can move it into a separate config for sds, but I don't think it's worth making it live in a separate message for the istio-proxy container.

projects/gateway2/deployer/values_helpers.go
```diff
@@ -94,6 +96,100 @@ func getServiceValues(svcConfig *v1alpha1kube.Service) *helmService {
...
+	} else {
+		var bootstrap *sdsBootstrap
+		if sdsConfig.GetSdsContainer() != nil {
+			bootstrap = &sdsBootstrap{
+				LogLevel: ptr.To(sdsConfig.GetSdsContainer().GetBootstrap().GetLogLevel()),
+			}
+		}
+
+		// convert the service type enum to its string representation;
+		// if type is not set, it will default to 0 ("ClusterIP")
```
jenshu, line 120: comment doesn't seem relevant

npolshakova: Yep- it looks like I copied it from the svc helm values. Removing

npolshakova:
Converting to a draft so it doesn't merge into `[npolshak/add-istio-upstream-tests](https://github.com/solo-io/gloo/tree/npolshak/add-istio-upstream-tests)`

projects/gateway2/deployer/deployer.go
```diff
@@ -192,11 +193,20 @@ func (d *Deployer) getValues(ctx context.Context, gw *api.Gateway) (*helmConfig,
...
 			Image: getDefaultEnvoyImageValues(d.inputs.Extensions.GetEnvoyImage()),
-			IstioSDS: &helmIstioSds{
+			// TODO(npolshak): Remove once default GatewayParameters are supported
+			IstioSDS: &istioSDS{
 				Enabled: &d.inputs.IstioValues.SDSEnabled,
 			},
 		},
 	}
+	contextutils.LoggerFrom(ctx).Errorf("SDS %s", d.inputs.Extensions.GetSdsImage().Tag)
+	contextutils.LoggerFrom(ctx).Errorf("ENVOY %s", d.inputs.Extensions.GetSdsImage().Tag)
```
jbohanon, line 203: was this dev logging?

npolshakova: Yep, will remove!

jbohanon (commented):
Looking good!

projects/gateway2/deployer/values_helpers.go
```diff
@@ -94,6 +96,120 @@ func getServiceValues(svcConfig *v1alpha1kube.Service) *helmService {
...
+}
+
+// Convert istio values from GatewayParameters into helm values to be used by the deployer.
+func getIstioValues(istioConfig *v1alpha1.IstioIntegration) *helmIstioSds {
+	var istioVals *helmIstioSds
+
+	defaultIstioImage := helmImage{
+		Registry:   ptr.To("docker.io/istio"),
+		Repository: ptr.To("proxyv2"),
+		Tag:        ptr.To("1.19.0"),
```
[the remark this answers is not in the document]
jenshu: i think that this is temporary until we have default GatewayParameters populated by the helm values.
once we have default gwparams, we should probably remove any defaulting from the code

projects/gateway2/status/status_syncer.go
```diff
@@ -105,7 +105,7 @@ func (f *statusSyncerFactory) HandleProxyReports(ctx context.Context, proxiesWit
 			newStatusSyncer(plugins).applyStatusPlugins(ctx, proxies)
 		} else {
 			// this should never happen
-			contextutils.LoggerFrom(ctx).DPanicf("no registry found for proxy sync count %d", syncCount)
+			//contextutils.LoggerFrom(ctx).DPanicf("no registry found for proxy sync count %d", syncCount)
```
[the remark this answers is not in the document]
npolshakova: Yep- hit it when running the control plane locally alongside the control plane in the kind cluster.

projects/gateway2/deployer/values_helpers.go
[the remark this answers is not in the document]
npolshakova: Yep, this is temporary, but we can make it match the version in the values-template!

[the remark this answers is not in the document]
npolshakova: Okay it actually looks like for Istio we can just use the `[gloo/install/helm/gloo/values-template.yaml](https://github.com/solo-io/gloo/blob/fd331e502a7513376ac15bafb3621de70af98efb/install/helm/gloo/values-template.yaml#L241)` and not set the default version in code! 🎉

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -98,6 +98,8 @@ spec:
                 fieldPath: metadata.namespace
           - name: ISTIO_MTLS_SDS_ENABLED
             value: "true"
+          - name: LOG_LEVEL
+            value: {{ $gateway.sds.sdsBootstrap.logLevel }}
```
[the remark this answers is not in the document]
npolshakova: It looks like we don't actually need to set the default log level for sds or istio in code if it's set in [gloo/install/helm/gloo/values-template.yaml](https://github.com/solo-io/gloo/blob/fd331e502a7513376ac15bafb3621de70af98efb/install/helm/gloo/values-template.yaml#L241). Going to remove the references.

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -109,6 +112,99 @@ message EnvoyBootstrap {
...
+
+  // Istio integration configuration
+  // If not provided, the default values will be used.
+  IstioIntegration istio_integration = 2;
+
+}
+
+// Configuration for the Istio integration settings used by a Gloo Gateway's data plane (Envoy proxy instance)
+message IstioIntegration {
+  // Enables SDS cert-rotator sidecar for istio mTLS cert rotation
```
sam-heilbron, line 130: When we had the design review, I had thought we were going to remove "enabled" flag and have it be implicit. Was that a different field or am I missing something?

projects/gateway2/deployer/deployer.go
```diff
@@ -192,12 +192,19 @@ func (d *Deployer) getValues(ctx context.Context, gw *api.Gateway) (*helmConfig,
 				Port: &d.inputs.ControlPlane.Kube.XdsPort,
 			},
 			Image: getDefaultEnvoyImageValues(d.inputs.Extensions.GetEnvoyImage()),
-			IstioSDS: &helmIstioSds{
+			// TODO(npolshak): Remove once default GatewayParameters are supported
```
sam-heilbron, line 195: nit: If this is an explicit piece of work to be tackled, personally I like linking the GitHub issue

projects/gateway2/extensions/extensions.go
```diff
@@ -84,6 +87,14 @@ func (e *k8sGatewayExtensions) GetEnvoyImage() Image {
 	}
 }
 
+// GetSdsImage returns the image repo and tag to use for the sds container image
```
sam-heilbron, line 90: This seems viable, though I wonder if/why we need to define this as a an extension. It feels like the sds repo and tag are available at build time (baked into the oss and ee charts) and I imagine we could extract those values from that source, rather than needing to define these runtime extensions

test/kubernetes/e2e/tests/istio_test.go
```diff
@@ -77,4 +78,8 @@ func TestK8sGatewayIstio(t *testing.T) {
 	t.Run("IstioIntegration", func(t *testing.T) {
 		suite.Run(t, istio.NewTestingSuite(ctx, testInstallation))
 	})
+
+	t.Run("IstioGatewayParameters", func(t *testing.T) {
```
sam-heilbron, line 82: I don't think this is running without modifying the GHA to include the name

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: I think we agreed on having sds implicitly enabled:
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
    sds: {} # sds would be enabled with default, Istio would be disabled
```

But I thought we still wanted Istio to have a value for enabled/disabled. I could remove the enable/disable for Istio and have it work the same as sds:

```
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    podTemplate:
      extraLabels:
        pod-label-key: pod-label-val
      extraAnnotations:
        pod-anno-key: pod-anno-val
    envoyContainer:
      bootstrap:
        logLevel: debug
        componentLogLevels:
          upstream: debug
          connection: trace
    sds: # sds would be enabled with default
	istio: {} # istio would be enabled by default
```
```

projects/gateway2/deployer/deployer.go
npolshakova: We currently don't have a github issue for the follow up but https://github.com/solo-io/solo-projects/issues/6107 highlights the steps. I can link to that!

projects/gateway2/extensions/extensions.go
npolshakova: We have the same thing for the Envoy oss vs. envoy-ee image, so we need to do the same thing for sds and sds-ee. @jenshu might have more insight why we needed it originally for the envoy image.

test/kubernetes/e2e/tests/istio_test.go
npolshakova: I thought it should run as part of `TestK8sGatewayIstio`? It's a subset similar to `IstioIntegration`, not a separate suite.

sam-heilbron: ha you're totally right: https://github.com/solo-io/gloo/blob/main/.github/workflows/pr-kubernetes-tests.yaml#L92

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: Discussed offline- will go with `istio: {}` to enable and remove the enabled field.

sam-heilbron (dismissed):

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -48,6 +48,9 @@ message KubernetesProxyConfig {
 
   // Autoscaling configuration.
   kube.gateway.gloo.solo.io.Autoscaling autoscaling = 5;
+
+  // Sds configuration
+  SdsIntegration sds = 6;
```
jbohanon, line 53: nit adding a comment here to indicate that a non-nil value will enable sds integration might be useful. It might just be implied htough

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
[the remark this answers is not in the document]
jbohanon: It doesn't look like this was removed, and I think the original concern still stands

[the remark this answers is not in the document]
jbohanon: Nope, re-read the deployer code and if `istioSDS.enabled` is true, we should always be populating `sds`

jbohanon (dismissed):

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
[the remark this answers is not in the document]
npolshakova: Yep- I needed to change it back to remove the values-templates defaults in order to allow the nil sds value to skip the sds container. The defaults now have to be defined in code.

projects/gateway2/deployer/values.go
```diff
@@ -34,6 +34,7 @@ type helmGateway struct {
 	NodeSelector        map[string]string                 `json:"nodeSelector,omitempty"`
 	Affinity            *extcorev1.Affinity               `json:"affinity,omitempty"`
 	Tolerations         []*extcorev1.Toleration           `json:"tolerations,omitempty"`
+	Sds                 *helmSds                          `json:"sds,omitempty"`
```
jenshu, line 37: nit: for readability, can you group this in a new section e.g. "sds values"? currently they are with the pod template values

projects/gateway2/deployer/values_helpers.go
```diff
@@ -94,6 +97,123 @@ func getServiceValues(svcConfig *v1alpha1kube.Service) *helmService {
...
+	      registry: quay.io/solo-io
+	      repository: sds
+	      pullPolicy: IfNotPresent
+	      # Overrides the image tag whose default is the chart appVersion.
+	      tag: ""
+	*/
+	defaultSdsImage := helmImage{
+		Repository: ptr.To(defaultSds.Repository),
+		Tag:        ptr.To(defaultSds.Tag),
+		Registry:   ptr.To("quay.io/solo-io"),
```
jenshu, line 115: just to confirm, this is another temporary value that will be removed once we implement default Gateway params, right?

projects/gateway2/extensions/extensions.go
jenshu: this is a workaround to be able to pass through the appropriate image which differs between oss and ee. once we have default gateway params that are populated by helm values, we can remove both the image funcs from k8sgatewayextensions

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
jenshu, line 102: should we only render this if `gateway.sds.sdsBootstrap.logLevel` is non empty?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -136,7 +139,7 @@ spec:
           - 45s
           - --parentShutdownDuration
           - 1m0s
-          - --proxyLogLevel=warning
+          - --proxyLogLevel={{ $gateway.sds.istioIntegration.logLevel }}
```
jenshu, line 142: same with this, can logLevel be empty?
can we add a check similar to https://github.com/solo-io/gloo/blob/35db22a46777576adc243bc7ae54067d138d1200/projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml#L43-L46

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -109,6 +112,96 @@ message EnvoyBootstrap {
...
+// Configuration for the sds integration settings used by a Gloo Gateway's data plane (Envoy proxy instance)
+// If not set, internal mtls authentication will not be used.
+message SdsIntegration {
+  // Configuration for the container running Gloo sds.
+  // If not provided, the default values will be used.
+  SdsContainer sds_container = 1;
+
+  // Istio integration configuration
+  // If not provided, the default values will be used.
+  IstioIntegration istio_integration = 2;
```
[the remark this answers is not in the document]
npolshakova: Nope! This is was one of the design decisions from the design discussion that we landed on: https://docs.google.com/document/d/1oF8KpoIEEcux_R8fBrLkdKF0NaonINVedhdCHMuf2bg/edit?usp=sharing The istio integration implies sds must be enabled.

projects/gateway2/deployer/values_helpers.go
npolshakova: No- now that we check the sds and integration values are not nil, we can't have the values-template set the default values. So they need to be set in code: https://github.com/solo-io/gloo/pull/9402#discussion_r1606792181

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
npolshakova: The top level condition should control this: `{{- if or $gateway.istioSDS.enabled $gateway.sds }}` and if `$gateway.istioSDS.enabled` is set, we currently fill in `gateway.sds` if the defaults.

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
npolshakova: Same thing as sds- https://github.com/solo-io/gloo/pull/9402#discussion_r1606924148 This is only set when `{{- if or $gateway.istioSDS.enabled (and $gateway.sds $gateway.sds.istioIntegration) }}` and we fill in the default Istio values in code now.

projects/gateway2/deployer/values_helpers.go
jenshu: hmm, now there are 2 sources of truth for the registry.
if the user sets the helm value `global.image.registry` to something other than quay, will we be able to pass it through to here?

jbohanon (approved):

projects/gateway2/deployer/values_helpers.go
jenshu: sounds good as long as we have a way of passing it from helm instead of defining the default here

npolshakova: How is the envoy image setting the `Registry` is it just defined in the default values-template file?

jbohanon (approved):

projects/gateway2/deployer/values_helpers.go
jenshu: yep that's currently defined in the deployer values [here](https://github.com/solo-io/gloo/blob/35db22a46777576adc243bc7ae54067d138d1200/projects/gateway2/helm/gloo-gateway/values.yaml#L31) although ideally i think the flow would be:
helm values from `gloo` / `gloo-ee` chart populate `spec.kube.envoyContainer.image` in the default GatewayParameters (which is linked to the default GatewayClass via parametersRef), and when the deployer runs it can grab the default envoy image from the default GatewayParameters

npolshakova: Yeah that makes sense- the `values-template` won't work for the sds/istio images since we have the non-nil value check. But the default GatewayParameters should get rid of needing to have the registry hard coded.

jenshu (approved):

PR #9404, opened 2024-04-26
Install yq with install-go-tools

jbohanon:
# Description

Remove need for dev to install yq on their `$PATH` by adding it to `make install-go-tools`


nfuden (dismissed):

jbohanon:
oof, looks like install-go-tools needs `LINTER_VERSION` definition and `LINTER_VERSION` definition needs install-go-tools

PR #9409, opened 2024-04-26
Turn off kubernetes gateway api integration by default

jenshu:
To match GG enterprise behavior, make k8s gateway api integration opt-in for now.
We don't want to break users who upgrade to 1.17+ since 1.17 will require extra steps before installing (i.e. applying the k8s gateway api crds)


install/test/helm_test.go
```diff
@@ -74,12 +73,7 @@ func GetValidationEnvVar() corev1.EnvVar {
 		Value: "true",
 	}
 }
-func GetK8sGwControllerEnvVar() corev1.EnvVar {
-	return corev1.EnvVar{
-		Name:  constants.GlooGatewayEnableK8sGwControllerEnv,
```
sam-heilbron, line 79: Given that we are removing this reference in our tests, are we safe to remove it in our codebase altogether?

install/helm/gloo/generate/values.go
```diff
@@ -295,7 +295,7 @@ type Gloo struct {
 }
 
 type KubeGateway struct {
-	Enabled *bool `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller. Defaults to true."`
+	Enabled *bool `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller."`
```
sam-heilbron, line 298: nit: We are disabling it intentionally, as enabling it requires other steps for users (like applying CRDs). Is it worth it to point users who want to use this field towards the guide that instructs them how to?

changelog/v1.17.0-beta24/disable-kube-gateway.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
```
sam-heilbron, line 2: This feels more like a HELM change to me, where we are turning the default value from true -> false

install/helm/gloo/values-template.yaml
```diff
@@ -5,7 +5,7 @@ k8s:
 
 # Default values for the Gloo Gateway - Kubernetes Gateway API integration
 kubeGateway:
-  enabled: true
+  enabled: false
```
sam-heilbron, line 8: nit: Should we clarify to devs why we chose this to be false?

sam-heilbron (approved):

install/test/helm_test.go
jenshu: we still use the env constant `GlooGatewayEnableK8sGwControllerEnv` in setup syncer to determine whether k8s gw integration runs, and in other tests

install/helm/gloo/generate/values.go
jenshu: probably; is there already a guide somewhere?

install/helm/gloo/values-template.yaml
jenshu: added a comment in this file 39cbaf7

changelog/v1.17.0-beta24/disable-kube-gateway.yaml
jenshu: changed to helm https://github.com/solo-io/gloo/commit/39cbaf701f1daf386294f740e6b2c13305658b86

sam-heilbron (approved):

jbohanon (approved):

PR #9413, opened 2024-04-29
gateway v2: implement route delegation capability

shashankram:
# Description

This change implements capability to delegate
HTTP routes using the Gateway API.
Refer to the detailed [proposal](https://docs.google.com/document/d/1daRDGwWDP3FPuEFKKNd-EvZsONK7kg2WxpGTckWOUdc/edit?usp=sharing) for more info.

## Code changes

- Translator changes to recursively translate child (delegatee) routes selected by a parent HTTPRoute.

- Translator tests for various delegation scenarios.

- Status changes to infer parentRefs of child/delegatee routes that must be inferred when the child doesn't specify parentRefs.

- Status builder tests for delegated routes.

- E2e suite for route delegation and some miscellaneous framework changes.

## Testing steps
Checkout the branch
```
git fetch origin
git checkout origin/shashankram/delegation-v3
```

Install Gloo
```
kind create cluster --name dev

./projects/gateway2/kind.sh

make VERSION=1.0.0-dev CLUSTER_NAME=dev IMAGE_VARIANT=standard kind-build-and-load -B
make VERSION=1.0.0-dev build-test-chart -B

# apply gw api crds
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml

# install gloo
helm upgrade -i -n gloo-system gloo ./_test/gloo-1.0.0-dev.tgz --create-namespace
```

Run the `RouteDelegation` suite in `test/kubernetes/e2e/k8sgateway/k8s_gw_test.go`. Alternatively, manually apply the configuration for various scenarios in `test/kubernetes/e2e/features/route_delegation/inputs` and test traffic.

# Checklist:

- [ ] I have made corresponding changes to the documentation: TODO


projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -245,17 +314,22 @@ func setRouteAction(
 				Weight:  weight,
 				Options: nil,
 			})
-		} else {
+
+		case backendref.RefIsHTTPRoute(backendRef.BackendObjectReference):
+			// This is handled at the beginning of the loop
```
ilackarms, line 319: how difficult would it be to move that code down into the switch statement?

shashankram: I think we can leave the code at the beginning as-is as this case statement is a no-op and should never be executed. I only put it there so someone doesn't think it's a bug where delegated refs are not handled here. We could remove this case entirely.

projects/gateway2/query/delegation.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		if matched {
+			selected = append(selected, child)
+		}
+	}
+
+	return selected, nil
+}
+
+func isDelegatedRouteMatch(
```
npolshakova, line 90: Do we have any unit tests for the `isDelegatedRouteMatch` logic? We could also break it up into smaller checks (matchesParentRefs, matchesPath, matchesHeaders, etc.) and test them individually.

projects/gateway2/reports/reporter.go
```diff
@@ -160,6 +161,10 @@ func (r *reporter) Route(route *gwv1.HTTPRoute) HTTPRouteReporter {
 
 // TODO: flesh out
 func getParentRefKey(parentRef *gwv1.ParentReference) ParentRefKey {
```
npolshakova, line 163: Is this only getting used for statuses? It looks like it has the same fields as `gwv1.ParentReference`?

projects/gateway2/translator/gateway_translator_test.go
```diff
@@ -89,4 +89,249 @@ var _ = Describe("GatewayTranslator", func() {
...
+						Namespace: "infra",
+						Name:      "example-gateway",
+					}: {
+						Proxy: dir + "/testutils/outputs/delegation/cyclic1.yaml",
+						// Reports:     nil,
+					},
+				},
+			}.Run(ctx)
+
+			Expect(err).NotTo(HaveOccurred())
```
npolshakova, line 271: Is there an error reported in the logs/status when a cyclic delegated route is detected?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -99,15 +136,20 @@ func translateGatewayHTTPRouteRule(
...
 					Status: http.StatusInternalServerError,
 				},
 			}
 		}
-		outputRoute.Matchers = []*matchers.Matcher{translateGlooMatcher(match)}
-		routes[idx] = outputRoute
+
+		// A parent route that delegates to a child route should not have a route
+		// as the routes are derived from the child route
+		if outputRoute.GetAction() != nil {
```
npolshakova, line 149: Should an error/warning get logged if a parent route that delegates to a child route also has a route?

test/kubernetes/e2e/features/route_delegation/suite.go
```diff
@@ -0,0 +1,171 @@
...
+	s.ti.Assertions.AssertEventuallyConsistentCurlResponse(s.ctx, defaults.CurlPodExecOpt,
+		[]curl.Option{
+			curl.WithHostPort(proxyHostPort), curl.WithPath(pathTeam2),
+			curl.WithHeader("headerX", "valX"),
+			curl.WithQueryParameters(map[string]string{"queryX": "valX"}),
+		},
+		&testmatchers.HttpResponse{StatusCode: http.StatusNotFound})
+}
+
+func routeStatusContainsMsg(route *gwv1.HTTPRoute, msg string) bool {
```
npolshakova, line 162: Can we make this a shared util?

projects/gateway2/query/delegation.go
shashankram: Lots of translator tests are added to cover these scenarios. Not opposed to adding unit tests for unexported helpers

projects/gateway2/reports/reporter.go
shashankram: This is invoked from `func (r *RouteReport) parentRef`. It flattens the pointers in gwv1.ParentReference to value types to be used as a map key:
```
type ParentRefKey struct {
	Group string
	Kind  string
	types.NamespacedName
}
```

projects/gateway2/translator/gateway_translator_test.go
shashankram: Yes, we verify this in e2e tests

projects/gateway2/translator/httproute/gateway_http_route_translator.go
shashankram: It's ok for a parent route to specify routes to backends while also delegating certain other routes. The below example is valid:
```
spec:
  parentRefs:
  - name: example-gateway
  hostnames:
  - "example.com"
  rules:
  - backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: /a
    backendRefs:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
      name: "*"
      namespace: a
```

Were you asking if this should be an error?

projects/gateway2/translator/gateway_translator_test.go
shashankram: TestCyclic in the newly added e2e suite

projects/gateway2/translator/httproute/gateway_http_route_translator.go
npolshakova: Okay then it looks like the comment might be off? 
```
// A parent route that delegates to a child route should not have a route
		// as the routes are derived from the child route
```

projects/gateway2/query/query.go
```diff
@@ -157,7 +159,6 @@ type gatewayQueries struct {
 }
 
 func (r *gatewayQueries) referenceAllowed(ctx context.Context, from metav1.GroupKind, fromns string, to metav1.GroupKind, tons, toname string) (bool, error) {
```
npolshakova, line 161: How do ReferenceGrants work with a delegated HTTPRoute? If the ReferenceGrant selects the parent, do the children inherit the grant?

projects/gateway2/translator/testutils/inputs/delegation/basic.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: example-gateway
+  namespace: infra
+spec:
+  gatewayClassName: example-gateway-class
```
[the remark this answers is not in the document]
shashankram: allowedRoutes is for Gateway->HTTPRoute binding. Has no bearing on route delegation per the design. Route delegation is decoupled from this configuration.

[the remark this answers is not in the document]
npolshakova: Makes sense! So setting the `allowedRoutes` field to Same namespaces would work the same as not setting this field? Do we have a test for that behavior?

projects/gateway2/query/query.go
npolshakova: But in order to pass conformance, we still need ReferenceGrant to work with HTTPRoutes. So is this how it works right now:
1. A parent HTTPRoute with ReferenceGrant and selected child routes will only have the reference grant applied for the parent.
2. A parent HTTPRoute with ReferenceGrant that is selected by a child route will also only have the reference grant applied for the parent.
3. A child HTTPRoute with ReferenceGrant that is selected by a parent will only have the reference grant applied for itself.
4.  A child HTTPRoute with ReferenceGrant that binds to a parent will only have the reference grant applied for itself.

npolshakova: And we'll put off supporting inherited ReferenceGrants until there's a real use-case?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
shashankram: yes that's confusing, it should be `should not have an outputRoute`

projects/gateway2/translator/testutils/inputs/delegation/basic.yaml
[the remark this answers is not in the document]
shashankram: The field is still required for the gateway to select the parent route, but the testing of that should be unrelated to this feature and the code changes made here. As long as the Gateway selects the parent route, delegation should work from parent->child with the correct config on the parent.

projects/gateway2/query/query.go
shashankram: ReferenceGrants will work the same way for child routes, but don't control the delegation of routes from parent->child. Since a ReferenceGrant can select HTTPRoutes, it could target either the parent or child. 

What do you mean by `ReferenceGrant that is selected`? Isn't it the other way around where the `ReferenceGrant` selects the routes?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
shashankram: reworded it

projects/gateway2/query/query.go
npolshakova: Yep, I meant if you have a ReferenceGrant that selects the parent/child route:

```
kind: ReferenceGrant
metadata:
  name: bar
  namespace: bar
spec:
  from:
  - group: gateway.networking.k8s.io
    kind: HTTPRoute
    namespace: foo # if this is the parent, does the child inherit the grant? 
  to:
  - group: multicluster.x-k8s.io
    kind: ServiceImport
```

I think not having inheritance for ReferenceGrants makes sense for now, but it would be good to check that `getRef` is working as expected since that still checks `referenceAllowed` in the delegated HTTPRoute case.

shashankram: There is no code that allows the child to inherit the ReferenceGrant. Child routes are only ever evaluated while traversing the parent->child graph, and while we evaluate the child we evaluate it exactly as we would if it were a parent route as long as it is a leaf route. 

In other words, ReferenceGrants can be applied to child routes just like they could on the normal non-delegated routes, but that's orthogonal to delegation.

projects/gateway2/query/delegation.go
shashankram: `projects/gateway2/translator/testutils/inputs/delegation/child_rule_matcher.yaml` is the test for testing different combination of route matching

shashankram: I added a unit test nevertheless

npolshakova (approved):
🎉

ilackarms (approved):

ilackarms (approved):

ilackarms (approved):

PR #9418, opened 2024-04-29
Add Istio strict/permissive auth e2e test

npolshakova:
# Description

Transfers [Istio regression tests ](https://github.com/solo-io/gloo/blob/main/test/kube2e/istio/istio_integration_test.go#L682) to new testing framework. This includes the strict and permissive peer auth tests.

There are two types of Istio integration tests added as part of this PR:
- k8s Gateway API with automtls enabled
- k8s Gateway API with automtls disabled

The follow up PR (https://github.com/solo-io/gloo/pull/9436) will add the same tests for the classic edge apis. 

## Testing steps

This PR adds the Istio integrations tests for k8s gateways which can be run via:
```
 TEST_PKG=test/kubernetes/e2e/k8sgateway make go-test
```

This PR also reworks the headless svc tests into two separate suites- one for the k8s gateway controller enabled, one for the k8s gateway controller disabled and only classic edge apis applied. 

# Checklist:


test/kubernetes/e2e/features/istio/suite.go
```diff
@@ -0,0 +1,168 @@
...
+			curl.WithPath("/headers"),
+		},
+		expectedMtlsResponse)
+
+	// Disable automtls
+	upgradeOpts := s.helmOptions
+	upgradeOpts.ExtraArgs = []string{
+		"--set", "global.istioIntegration.enableAutoMtls=false",
+		// TODO: why is this getting overwritten?
+		"--set", "kubeGateway.enabled=true",
```
sam-heilbron, line 147: Perhaps I'm oversimplifying but it feels like this should live again in a separate test, so we have a suite that says "here are all the tests that we perform again a cluster without auto-mtls"

test/kubernetes/e2e/features/istio/suite.go
```diff
@@ -0,0 +1,168 @@
...
+	err = helm.HelmUpgradeInstallGloo(upgradeOpts)
+	s.NoError(err, "can upgrade gloo with automtls disabled")
+
+	gomega.Eventually(func(g gomega.Gomega) {
+		settings, err := s.testInstallation.ResourceClients.SettingsClient().Read(s.testInstallation.Metadata.InstallNamespace, "default", clients.ReadOpts{})
+		g.Expect(err).NotTo(gomega.HaveOccurred(), "can read settings")
+		g.Expect(settings.GetGloo().GetIstioOptions().GetEnableAutoMtls().GetValue()).To(gomega.BeFalse(), "settings have auto mtls disabled")
+	}).Should(gomega.Succeed(), "settings have auto mtls disabled")
+
+	// With auto mtls disabled in the mesh, the response should not contain the X-Forwarded-Client-Cert header
```
sam-heilbron, line 158: One idea would be: we have some istio tests defined in the features/istio package. A testInstallation object has glooGateway.Context with metadata about the installation. We can add enableAutoMtls, and then you can have logic about the behavior baked directly in here. 

Other idea is to have 2 separate tests, and just run one for features/istioAutoMtls and another for features/IstioClassicMtls

test/gomega/matchers/have_http_response.go
```diff
@@ -71,6 +71,22 @@ type HttpResponse struct {
...
+	switch bodyMatcher := r.Body.(type) {
+	case string:
+		bodyString = bodyMatcher
+	case []byte:
+		bodyString = string(bodyMatcher)
+	case types.GomegaMatcher:
+		bodyString = fmt.Sprintf("%#v", bodyMatcher)
+	}
+
+	return fmt.Sprintf("HttpResponse{StatusCode: %d, Body: %s, Headers: %v, Custom: %v}",
```
npolshakova, line 85: Fixes the failure match to print: 
```
 Body: &matchers.NotMatcher{Matcher:(*matchers.ContainSubstringMatcher)(0x14000e22720)} 
```

Instead of: 
```
&{StatusCode:200 Body:0x140006b6f30 Headers:map[] Custom:<nil>}
```

test/kubernetes/testutils/assertions/glooctl.go
```diff
@@ -34,7 +34,7 @@ func (p *Provider) EventuallyCheckResourcesOk(ctx context.Context) {
 		// These are some basic defaults that we expect to work in most cases
 		// We can make these configurable if need be, though most installations
 		// Should be able to become healthy within this window
-		WithTimeout(time.Second * 90).
+		WithTimeout(time.Second * 120).
```
npolshakova, line 37: The istio-proxy/sds containers can take a little longer to startup.

test/kubernetes/e2e/classicedge/classic_api_test.go
```diff
@@ -45,12 +45,11 @@ func TestClassicEdgeGateway(t *testing.T) {
...
 
 	// Install Gloo Gateway with only classic APIs enabled
-	// If the env var SKIP_GLOO_INSTALL=true, installation will be skipped
 	testInstallation.InstallGlooGateway(ctx, func(ctx context.Context) error {
 		return testHelper.InstallGloo(ctx, helper.GATEWAY, 5*time.Minute, helper.ExtraArgs("--values", testInstallation.Metadata.ValuesManifestFile))
 	})
 
 	t.Run("HeadlessSvc", func(t *testing.T) {
-		suite.Run(t, headless_svc.NewHeadlessSvcTestingSuite(ctx, testInstallation, false))
+		suite.Run(t, headless_svc.NewClassicHeadlessSvcSuite(ctx, testInstallation))
```
sam-heilbron, line 53: nit: Can be done in a follow-up. Per @kcbabo we're going to use "Edge Gateway API" and "Kubernetes Gateway API"

test/kubernetes/e2e/test.go
```diff
@@ -35,11 +39,13 @@ func MustTestHelper(ctx context.Context, installation *TestInstallation) *helper
 		panic(err)
 	}
 
+	testHelper.DeployTestServer = false
```
sam-heilbron, line 42: yessss

test/kubernetes/e2e/test.go
```diff
@@ -115,6 +121,9 @@ type TestInstallation struct {
 
 	// Assertions is the entity that creates assertions that can be executed by the Operator
 	Assertions *assertions.Provider
+
+	// IstioctlBinary is the path to the istioctl binary that can be used to interact with Istio
+	IstioctlBinary string
```
sam-heilbron, line 126: nit: I'm good with this here for now. I was initially hoping to try to keep a TestInstallation as generic as possible, and place specific details about "how" an installation should be configured onto either the cluster.Context or glooGateway.Context. But I'm sure we'll find other tests cases like this that require other resources and we can see what makes sense

test/kubernetes/e2e/test.go
```diff
@@ -148,3 +157,155 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
 	logsCmd := i.Actions.Kubectl().Command(ctx, "logs", "-n", i.Metadata.InstallNamespace, "deployments/gloo")
 	logsCmd.Run()
 }
+
+const (
+	IstioctlVersionEnv  = "ISTIOCTL_VERSION"
```
sam-heilbron, line 162: nit: I kind of like having all relevant test env variables in a single place (https://github.com/solo-io/gloo/blob/main/test/testutils/env.go) so as a dev and I go there and see what are my options, instead of needing comb through files. Perhaps we have something like that _just_ for these new e2e tests? Either way, happy with it for now

test/kubernetes/e2e/test.go
```diff
@@ -148,3 +157,155 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
...
+func getIstioctlVersionOrDefault() string {
+	if version := os.Getenv(IstioctlVersionEnv); version != "" {
+		return version
+	} else {
+		return defaultIstioVersion
+	}
+}
+
+// Download istioctl binary from istio.io/downloadIstio and returns the path to the binary
+func DownloadIstio(ctx context.Context, version string) (string, error) {
```
sam-heilbron, line 216: nit: It feels like this could live in a file that just owns istio concepts/details and we can keep it distinct from this central test.go that all tests will use

test/kubernetes/e2e/test.go
```diff
@@ -148,3 +157,155 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
...
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("yes | %s uninstall --purge --context %s", i.IstioctlBinary, i.TestCluster.ClusterContext.KubeContext))
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("istioctl uninstall failed: %w", err)
+	}
+	return nil
+}
+
+func GlooDirectory() string {
```
sam-heilbron, line 305: nit: I think this coudl be a base utility that is then usable by other packages

sam-heilbron (approved):
🚀 

Small nits that can be ignored, handled, or handled in a follow-up

test/kubernetes/e2e/classicedge/classic_api_test.go
npolshakova: Sounds good! This will be a good follow up for https://github.com/solo-io/gloo/pull/9436 which has everything named as "classic"

test/kubernetes/e2e/test.go
npolshakova: I think we can move this to TestCluster- there shouldn't be anything `TestInstallation` specific. We just need the cluster context.

test/kubernetes/e2e/test.go
npolshakova: Yeah this is one is a little weird since it's Istioctl specific, which not all tests will use. But I'm fine moving it to `testutils/env.go`. Let's do that in the follow up- maybe even move the whole `DownloadIstio`/`UninstallIstio` into it's own helper function.

test/kubernetes/e2e/test.go
npolshakova: Agreed- https://github.com/solo-io/gloo/pull/9418#discussion_r1588553876.

test/kubernetes/e2e/test.go
npolshakova: Yeah that makes sense, especially if the Istio install/uninstall is going into a separate util, we can move this to the testutils.

shashankram (approved):

PR #9420, opened 2024-04-30
Adding tilt for enhanced development

davidjumani:
# Description

This PR adds the ability to bring up, run and debug edge via tilt.
It is as simple as running `tilt up` once your cluster is configured.
The full guide is found in devel/tilt

# Checklist:


devel/tilt/tilt.md
```diff
@@ -0,0 +1,64 @@
+## Development with Tilt
+This section describes how to use [kind](https://kind.sigs.k8s.io) and [Tilt](https://tilt.dev) for a simplified
+workflow that offers easy deployments and rapid iterative builds.
+
+This allows
```
ashishb-solo, line 5: looks like an incomplete sentence here

davidjumani: Updated!

ashishb-solo (approved):

tilt-settings.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+providers :
+  gloo:
+    context: _output/projects/gloo
+    image: quay.io/solo-io/gloo
+    live_reload_deps:
+      - projects/gloo
+    label: gloo
+    build_binary: make -B gloo
+    binary_name: gloo-linux-arm64
```
bewebi, line 18: Not all devs will be using this binary
If there's a way to dynamically determine which should be used, that would be great, though I'm not sure if that's reasonable

As an alternative we can add a commented line that calls out that some devs may need to use the AMD binary
```suggestion
    binary_name: gloo-linux-arm64
    # binary_name: gloo-linux-amd64 # devs on AMD machines will need to use this binary instead
```

bewebi (changes requested):
It mainly seems to work as intended however updates seem to fail with the following error when set up using the steps from `devel/tilt/tilt/md`:
```
/app/start.sh: 34: cannot create process.txt: Read-only file system
```
This seems to stem from the Gloo deployment having a `securityContext` with `readOnlyRootFilesystem: true` and Tilt having to write to the pod's filesystem on startup
Manually triggering update from the Tilt UI seems to override the `securityContext`, enabling updates to occur without this error

If there is a way to avoid having the `securityContext` contain `readOnlyRootFilesystem: true` on the initial `tilt up` command that would be ideal, perhaps by ensuring that `.test/kube2e/helm/artifacts/helm.yaml` specifies `securityContext` without `readOnlyRootFilesystem: true`
If this is not reasonable, and perhaps even if it is, the instructions should identify that manually triggering update is a useful troubleshooting step

tilt-settings.yaml
davidjumani: Thanks, I've updated the docs as well as made provision for the architecture

bewebi (approved):
Awesome changes, works like a charm!

ashishb-solo (approved):

Issue #7147, opened 2022-09-14
Introduce Tests to Identify Goroutine Leaks


sam-heilbron:
### Is your feature request related to a problem? Please describe.

A mechanism to identify in tests whether goroutines are being leaked.

### Describe the solution you'd like

In https://github.com/solo-io/gloo/pull/7017, we introduced a utility and used it in one test suite. We should extend this usage (or find a better solution) across all of our tests suites.

PR #9426, opened 2024-05-01
test: improve leak detection

sam-heilbron:
# Description

Attempt to reduce test inconsistencies by improving the retry logic for identifying safe ports, and expanding the usage of our leak detector

# Context
## Leak Detector
Per https://github.com/solo-io/gloo/issues/7147 we have a way to identify goroutine leaks. I think this would be more useful in our tests, and so I did the following:
- Allow it to support dependency injection so that if users need to customize the behavior, they can
- Invoke it in our e2e test suite

**UPDATE**: I backed out the change to our e2e suite as it kept identifying leaks after the tests succeeded. While this is great to know about, it was going to involve a larger change, and this work was primarily motivated for other reasons.

See https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=fad08c2a-57f7-4057-b058-4eed4126296f for an example:
```
[38;5;10m[1mSUCCESS![0m -- [38;5;10m[1m290 Passed[0m | [38;5;9m[1m0 Failed[0m | [38;5;11m[1m0 Pending[0m | [38;5;14m[1m0 Skipped[0m
Step #1 - "run-tests": Step #8 - "run-e2e-tests": --- FAIL: TestE2e (818.34s)
Step #1 - "run-tests": Step #8 - "run-e2e-tests":     debug.go:32: found unexpected goroutines:
Step #1 - "run-tests": Step #8 - "run-e2e-tests":         [Goroutine 13562 in state chan send, 10 minutes, with github.com/solo-io/gloo/test/e2e_test.init.func36.4.3.1 on top of the stack:
Step #1 - "run-tests": Step #8 - "run-e2e-tests":         goroutine 13562 [chan send, 10 minutes]:
Step #1 - "run-tests": Step #8 - "run-e2e-tests":         github.com/solo-io/gloo/test/e2e_test.init.func36.4.3.1({0x9143f00, 0xc000074a80}, 0xc002404b40)
Step #1 - "run-tests": Step #8 - "run-e2e-tests":         	/workspace/gloo/test/e2e/zipkin_test.go:181 +0x28e
....
```

## AdvancePortSafe
Per https://github.com/solo-io/solo-projects/issues/6134, we have seen some confusing behavior around port selection for e2e tests. This relies on a utility defined in open source.

What is strange, is that there are many other places this utility is used (envoy instance) in our e2e tests where we do not see an issue.
Regardless, I added some more retry delay, retry attempts, and logging to try to make debugging failures easier. Also I added some more unit tests.

## Race Flags with Kube Tests
PR before race was enabled: https://github.com/solo-io/gloo/actions/runs/8909351624/job/24466584626 (15min, 14sec)
With race enabled: https://github.com/solo-io/gloo/actions/runs/8910799268/job/24470777788 (15 min, 47 sec)

## Setup Loop Logging Changes
Before this change to error handling, i would see: 
```
{"level":"info","ts":"2024-05-01T12:27:13.085-0600","logger":"setup","caller":"setup/setup_syncer.go:1046","msg":"Gloo setup completed successfully"}
{"level":"error","ts":"2024-05-01T12:27:13.085-0600","logger":"setup","caller":"setup/setup_syncer.go:1037","msg":"gloo main event loop","stacktrace":"github.com/solo-io/gloo/projects/gloo/pkg/syncer/setup.RunGlooWithExtensions.func11\n\t/Users/samheilbron/go/src/github.com/solo-io/gloo/projects/gloo/pkg/syncer/setup/setup_syncer.go:1037"}
```
This error log always through me off because nothing was wrong but we were reporting it to the users. 

With my change, you can focus the example_test, and run
```
TEST_PKG=test/e2e make test
```
the logs will contain:
```
{"level":"info","ts":"2024-05-01T12:40:58.704-0600","logger":"setup","caller":"setup/setup_syncer.go:1055","msg":"Gloo setup completed successfully"}
```
## Interesting decisions
 -

## Testing steps
- I updated the tests for our advance port logic to be doubly sure that it works. It was working before, but I think the tests are more assertive now because they require multiple retries

## Notes for reviewers
I have follow-up work https://github.com/solo-io/solo-projects/pull/6136 to build off of this

# Checklist:


test/ginkgo/parallel/ports_test.go
```diff
@@ -29,8 +31,18 @@ var _ = Describe("Ports", func() {
 			portInDenylistMinusOffset := portInDenylist - advanceAmount
 
 			selectedPort := parallel.AdvancePortSafe(&portInDenylistMinusOffset, portInUseDenylist)
-			Expect(selectedPort).NotTo(Equal(portInDenylist), "should have skipped the port in the denylist")
-			Expect(selectedPort).To(Equal(portInDenylist+1), "should have selected the next port")
+			Expect([]uint32{10010, 10011, 10012}).NotTo(ContainElement(selectedPort), "should have skipped the ports in the denylist")
+			Expect(selectedPort).To(Equal(uint32(10013)), "should have selected the next port")
+		})
+
+		It("exhausts all retries, and returns last attempt, even if retry was unsuccessful", func() {
```
nfuden, line 38: Given that we expect this to always fail should the retrybackoff time be something that is configurable in case soemone attempts to increase from the current 100 ms and dont realize that it could impact the standard success timing of this part?

sam-heilbron: Yeah it's a good point. I had opted to try to make this function as simple as possible. Happy to expand it to make it more configurable

sheidkamp (commented):
Overall LGTM, one nit.

Have you run this repeatedly in the `solo-projects` PR you linked with no flakes? My theory is that these port conflicts are a result of ExtAuth ports colliding with Redis ports, and if for some reason Redis isn't letting go of ports properly, I'm not sure how effective these changes will be.

These changes are a step forward, and maybe the leak detection will help, so I am fine with these changes going in even if they don't end up fixing the flake by themselves.

sam-heilbron:
> Have you run this repeatedly in the `solo-projects` PR you linked with no flakes?

Totally agreed with your point. I don't think that this will solve the flakes, but I figured it's a step towards them, by increasing the retries and adding some simple logging on error

test/ginkgo/parallel/ports_test.go
sam-heilbron: https://github.com/solo-io/gloo/pull/9426/commits/c151fefc065b1e216aa4760a3ddcf09d319fa6f0

test/ginkgo/parallel/ports.go
```diff
@@ -30,15 +33,21 @@ func GetPortOffset() int {
...
 		newPort = AdvancePort(p)
 		return errIfPortInUse(newPort)
 	},
 		retry.RetryIf(func(err error) bool {
 			return err != nil
 		}),
-		retry.Attempts(3),
-		retry.Delay(time.Millisecond*0))
+		// While this is a large number of retries,
+		// this is only run in tests, and we would rather be safe
```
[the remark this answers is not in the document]
sam-heilbron: https://github.com/solo-io/gloo/pull/9426/commits/71db22cc65d080024bcfd594576c8e604d210d2d

Makefile
```diff
@@ -269,7 +269,7 @@ run-kube-e2e-tests: test
 # Go Tests
 #----------------------------------------------------------------------------------
 GO_TEST_ENV ?=
-GO_TEST_ARGS ?=
+GO_TEST_ARGS ?= -cpu=4 -race
```
bewebi, line 272: Adding race detection is good, but comes at significant performance cost
From [Go docs](https://go.dev/doc/articles/race_detector):
```
The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.
```

The length for CI to complete is already a pain for developers, and this would also increase costs significantly, so perhaps there's a way to add race detection to our process without enabling it by default
This could look like any combination of:
- Adding Make targets for running tests with race detection
- Adding a step to our PR checklist encouraging authors to test at least once with race detection
- Adding a git directive (similar to `/kick`) to run CI with race detection on-demand
- Adding a e2e/cloud build to weekly/nightly tests with race detection enabled

If we do want to add this as a default arg we'll want full team buy-in, and perhaps get input from other stakeholders who may want to weigh in on cost and/or insight from other engineering teams around their approach to race detection

test/ginkgo/parallel/ports_test.go
bewebi: Is this behavior we want? If we exhaust all retries shouldn't we fail here rather than downstream, to make it easier to diagnose

bewebi (changes requested):

test/ginkgo/parallel/ports_test.go
sam-heilbron: It's the behavior we have and I wanted to minimize the surface area of these changes and not change the fucntion signature. As it is, if an error occurred, we will see a test fail, so I don't see a large difference between returning an error/not, with respect to tests

Makefile
bewebi: Performance is a reasonable concern in CI as well, no? If we saw even a 2x increase in execution time, yet alone 20x, there would be a meaningful downside in developer experience that's worth taking into account
Lately succesful builds have been taking ~20 minutes
If we find that this grows to >25 minutes with race detection I'd want to regroup as a team to get buy-in and/or align on an alternative paradigm for race testing

Additionally we seem to be failing all cloudbuild runs due to races being detected (for example [here](https://console.cloud.google.com/cloud-build/builds;region=global/1d982837-9e59-4541-93c0-24bc25e87f46;step=1?authuser=1&project=solo-public), so we'll need to actually address those failures before we can make this a default argument (and assess performance impacts)

test/ginkgo/parallel/ports_test.go
```diff
@@ -30,18 +30,19 @@ var _ = Describe("Ports", func() {
...
 
-			selectedPort := parallel.AdvancePortSafe(&portInDenylistMinusOffset, portInUseDenylist, retry.Delay(0))
+			selectedPort, err := parallel.AdvancePortSafe(&portInDenylistMinusOffset, portInUseDenylist, retry.Delay(0))
+			Expect(err).NotTo(HaveOccurred())
 			Expect([]uint32{10010, 10011, 10012}).NotTo(ContainElement(selectedPort), "should have skipped the ports in the denylist")
 			Expect(selectedPort).To(Equal(uint32(10013)), "should have selected the next port")
 		})
 
-		It("exhausts all retries, and returns last attempt, even if retry was unsuccessful", func() {
+		It("exhausts all retries, and returns last attempt and error, even if retry was unsuccessful", func() {
```
bewebi, line 39:
```suggestion
		It("exhausts all retries, and returns last attempt and error, if retry was unsuccessful", func() {
```

Makefile
sam-heilbron: I've update the PR body to show which tests this impact, and their runtime before this flag and after. It's 30 seconds, and that is with a small sample size (ie it might not all be due to the race flag)

sam-heilbron: >Performance is a reasonable concern in CI as well, no? If we saw even a 2x increase in execution time, yet alone 20x, there would be a meaningful downside in developer experience that's worth taking into account

Agreed, in the PR body I showed the (albeit minimal) data to show that the execution time of these tests (it's just one suite of tests that use this flag) has changed. by 30 seconds

> Lately successful builds have been taking ~20 minutes
If we find that this grows to >25 minutes with race detection I'd want to regroup as a team to get buy-in and/or align on an alternative paradigm for race testing
This change only impacts running our new kube test suite. this does not impact our existing kube test suites (which use the race flag already) or our builds

> Additionally we seem to be failing all cloudbuild runs due to races being detected (for example [here](https://console.cloud.google.com/cloud-build/builds;region=global/1d982837-9e59-4541-93c0-24bc25e87f46;step=1?authuser=1&project=solo-public)), so we'll need to actually address those failures before we can make this a default argument (and assess performance impacts)
Wouldn't that be another reason to add this? Ie it helps us identify in tests, when races exist in code?

bewebi: > Wouldn't that be another reason to add this? Ie it helps us identify in tests, when races exist in code?

Yes, definitely something we should be doing regularly
My concern would be that if there's a significant performance impact we may not want to do it _every time_
My initial comments had some suggestions for how we might integrate race detection in our process without having it on every single run

If there isn't a meaningful performance impact observed that's great and we may as well include it all the time

test/ginkgo/parallel/ports_test.go
sam-heilbron: Updated to return error in one method, and panic in another so devs have choice

bewebi (approved):

sheidkamp (approved):

PR #9431, opened 2024-05-01
gateway2/delegation: update translator test names and use WithHostHeader

shashankram:
# Description

update translator test names and use WithHostHeader

### DIFF
--- a/projects/gateway2/translator/gateway_translator_test.go
+++ b/projects/gateway2/translator/gateway_translator_test.go
@@ -120,7 +120,7 @@ var _ = Describe("GatewayTranslator", func() {
 
 		It("should translate a basic configuration with child specifying matching parentRefs", func() {
 			results, err := TestCase{
-				Name:       "delegation-basic",
+				Name:       "delegation-parentRefs-match",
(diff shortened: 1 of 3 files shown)


sam-heilbron (approved):

projects/gateway2/translator/gateway_translator_test.go
```diff
@@ -120,7 +120,7 @@ var _ = Describe("GatewayTranslator", func() {
 
 		It("should translate a basic configuration with child specifying matching parentRefs", func() {
 			results, err := TestCase{
-				Name:       "delegation-basic",
+				Name:       "delegation-parentRefs-match",
```
jenshu, line 123: nit: doesn't have to be done in this PR, but to reduce duplication and chance of typos, i wonder if TestCase can take in the name as an arg, then we can construct the input / output files from that, e.g. `/testutils/inputs/delegation/<testName>.yaml`

jenshu (approved):

projects/gateway2/translator/gateway_translator_test.go
shashankram: Something like that is possible, but I like the flexibility with the current definition. Adding additional abstractions inside the test framework internals makes it harder to debug misconfigurations

PR #9434, opened 2024-05-02
Add VirtualHostOptions e2e tests

jbohanon:
# Description
Testing added to validate the behavior of the VirtualHostOptions plugin for Kubernetes Gateway API. 

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6085


test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -55,14 +53,12 @@ func (s *testingSuite) TestConfigureRouteOptionsWithTargetRef() {
...
-		s.ctx,
 		func() (resources.InputResource, error) {
 			return s.testInstallation.ResourceClients.RouteOptionClient().Read(routeOptionMeta.GetNamespace(), routeOptionMeta.GetName(), clients.ReadOpts{})
 		},
-		gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
-			"State":      gomega.Equal(core.Status_Accepted),
-			"ReportedBy": gomega.Equal("gloo-kube-gateway"), // TODO: this should likely be a constant
-		}))
+		core.Status_Accepted,
+		"gloo-kube-gateway",
```
npolshakova, line 60: [nit] Do we want to make this a const somewhere on the reporter? The route options e2e test currently also has this hard coded.

test/kubernetes/e2e/features/route_options/types.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 var (
-	targetRefManifest      = filepath.Join(util.MustGetThisDir(), "inputs/fault-injection-targetref.yaml")
-	filterExtensioManifest = filepath.Join(util.MustGetThisDir(), "inputs/fault-injection-filter-extension.yaml")
+	targetRefManifest      = filepath.Join(util.MustGetThisDir(), "testdata", "fault-injection-targetref.yaml")
```
npolshakova, line 19: 🎉

devel/tools/changelog.sh
```diff
@@ -9,14 +9,28 @@ NEXT_VERSION=$(echo -n $CURRENT_VERSION | sed -E "s/$OLD_VERSION$/$NEW_VERSION/"
 BRANCH_NAME=$(git symbolic-ref -q HEAD | sed 's#^.*/##')
 DESCRIPTION=${DESCRIPTION:=""}
 
-mkdir -p "changelog/$NEXT_VERSION"
+CHANGELOG_DIR="changelog/$NEXT_VERSION"
+mkdir -p CHANGELOG_DIR
```
jbohanon, line 13:
```suggestion
mkdir -p "$CHANGELOG_DIR"
```

changelog/v1.17.0-beta25/vhopts-status-e2e-tests.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/solo-projects/issues/6085
+    resolvesIssue: true
+    description: >-
+      "Add end-to-end testing for the VirtualHostOptions feature as used by the Kubernetes Gateway API integration"
```
jbohanon, line 6:
```suggestion
      Add end-to-end testing for the VirtualHostOptions feature as used by the Kubernetes Gateway API integration
```

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,161 @@
...
+	// Check status is accepted on VirtualHostOption with section name
+	s.testInstallation.Assertions.EventuallyResourceStatusMatchesState(
+		s.getterForMeta(&sectionNameVirtualHostOptionMeta),
+		core.Status_Accepted,
+		"gloo-kube-gateway",
+	)
+	// Check status is warning on VirtualHostOptions not selected for attachment
+	s.testInstallation.Assertions.EventuallyResourceStatusMatchesWarningReasons(
+		s.getterForMeta(&virtualHostOptionMeta),
+		[]string{"conflict with more-specific or older VirtualHostOption"},
```
jbohanon, line 104: This string is used multiple times, so it could be pulled out into a const. 

What I really need to do is make an exported base error from the vho plugin and check for that error's string

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,161 @@
...
+
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, targetRefManifest)
+	s.NoError(err, "can apply targetRefManifest")
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, extraVhOManifest)
+	s.NoError(err, "can apply extraVhOManifest")
+
+	// Check resources are created for Gateway
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+
+	// Check header manipulation is applied
```
npolshakova, line 130: [nit] I think the comment here is off since it's testing `expectedResponseWithoutContentLength` 
```suggestion
	// Check healthy response with no content-length header
```

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,161 @@
...
+	s.T().Cleanup(func() {
+		err := s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, targetRefManifest)
+		s.NoError(err, "can delete targetRefManifest")
+		err = s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, extraVhOManifest)
+		s.NoError(err, "can delete extraVhOManifest")
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, proxyService, proxyDeployment)
+	})
+
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, targetRefManifest)
+	s.NoError(err, "can apply targetRefManifest")
```
npolshakova, line 123: Do we want to assert the older VirtualHostOption was created here before the newer, conflicting one?

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,161 @@
...
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, extraVhOManifest)
+	s.NoError(err, "can apply extraVhOManifest")
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, sectionNameVhOManifest)
+	s.NoError(err, "can apply sectionNameVhOManifest")
+
+	// Check resources are created for Gateway
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+
+	// Check header manipulation is applied
```
jbohanon, line 85:
```suggestion
	// Check healthy response with added foo header
```

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,161 @@
...
+		s.NoError(err, "can delete manifest")
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, proxyService, proxyDeployment)
+	})
+
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, targetRefManifest)
+	s.NoError(err, "can apply targetRefManifest")
+
+	// Check resources are created for Gateway
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+	s.testInstallation.Assertions.AssertEventualCurlResponse(
```
jbohanon, line 48:
```suggestion
	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
	
	// Check healthy response with no content-length header
	s.testInstallation.Assertions.AssertEventualCurlResponse(
```

npolshakova (dismissed):

test/kubernetes/e2e/features/virtualhost_options/suite.go
jbohanon: I like this idea, but we don't currently have `creationTimestamp` as a field on `solo-kit` generated resources. Do you know of another way to accomplish this?

npolshakova: I think the VirtualService, Upstream, etc. clients won't give you the `creationTimestamp` because those are not client.Objects, but the RouteOptions do seem to have the createTimestamp since we're sorting them [here](https://github.com/solo-io/gloo/blob/23f90b6e5a4b3eb32d268def79b6b4fc1c2bbf3b/projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go#L188) by timestamp. The VirtualHostOptions are also defined here as a client object: https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/api/v1/kube/apis/gateway.solo.io/v1/types.go#L338

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,212 @@
...
+		s.NoError(err, "can apply "+k)
+	}
+}
+func (s *testingSuite) TestConfigureVirtualHostOptions() {
+	resources := map[string]string{
+		"setupManifest":    setupManifest,
+		"basicVhOManifest": basicVhOManifest,
+	}
+
+	s.T().Cleanup(s.cleanupFunc(resources))
```
sam-heilbron, line 59: Given that there is a BeforeTest and AfterTest function in the framework, is there a reason we are defining these functions ourselves? See https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/features/route_delegation/suite.go#L77 for an example. If there is a reason, it may be worthwhile to have a comment clarifying

test/kubernetes/e2e/features/virtualhost_options/testdata/bad-vho.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: gateway.solo.io/v1
+kind: VirtualHostOption
+metadata:
+  name: bad-retries
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: gw
+    namespace: default
```
sam-heilbron, line 10: Is the namespace necessary or relevant to the tests passing?

test/kubernetes/e2e/features/virtualhost_options/types.go
```diff
@@ -0,0 +1,81 @@
...
+	// When we apply the setup file, we expect resources to be created with this metadata
+	glooProxyObjectMeta = metav1.ObjectMeta{
+		Name:      "gloo-proxy-gw",
+		Namespace: "default",
+	}
+	proxyDeployment = &appsv1.Deployment{ObjectMeta: glooProxyObjectMeta}
+	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+
+	// curlPod is the Pod that will be used to execute curl requests, and is defined in the fault injection manifest files
+	curlPodExecOpt = kubectl.PodExecOptions{
```
sam-heilbron, line 33: We have a default for this: https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/defaults/defaults.go#L11

sam-heilbron (commented):
I like the changes!

On the package name changes, I had opened a separate thread to figure out how we want to break up the tests into separate packages to run more efficiently. Given that that will happen asap, could we backout this naming change and we can incorporate it when we re-balance the test suite

test/kubernetes/e2e/features/virtualhost_options/suite.go
jbohanon: I'm not sure how to get these objects cast from what we get out of the resource client though (which is a protoc-gen-go object generated [here](https://github.com/solo-io/gloo/blob/main/projects/gateway/pkg/api/v1/external_options.pb.go)). Do you have any ideas? I thought we had controller runtime resource clients on the test framework at some point, but I can't seem to find them now

npolshakova: Hm, I think it's an issue with our current resource client then for vhosts options and route options. I'm not super familiar with how Edge usually does this, so maybe we can just assert the resources exist after applying the config separately with `EventuallyObjectsExists` after each `s.testInstallation.Actions.Kubectl().ApplyFile`?

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -0,0 +1,212 @@
...
+		"basicVhOManifest": basicVhOManifest,
+		"badVhOManifest":   badVhOManifest,
+	}
+
+	s.T().Cleanup(s.cleanupFunc(resources))
+
+	s.setup(resources)
+
+	// Check resources are created for Gateway
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
```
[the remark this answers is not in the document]
jbohanon: I think it could be helpful visibility into why these tests failed. That said, other tests should be failing if the deployer weren't working as expected so probably not needed

projects/gloo/pkg/defaults/reporter.go
```diff
@@ -0,0 +1,7 @@
+package defaults
+
+const (
+	KubeGatewayReporter = "gloo-kube-gateway"
+	GlooReporter        = "gloo"
+	GlooGatewayReporter = "gateway"
```
sam-heilbron, line 6: nit: I think calling this `EdgeGatewayReporter` would better mirror the naming convention for the two types of gw APIs are supporting, and provide symmetry with the KubeGatewayReporter naming

projects/gloo/pkg/translator/route_config.go
```diff
@@ -164,10 +164,22 @@ func (h *httpRouteConfigurationTranslator) computeVirtualHost(
...
 		if err := plugin.ProcessVirtualHost(params, virtualHost, out); err != nil {
-			validation.AppendVirtualHostError(
-				vhostReport,
+			// Check if the incoming v1.VirtualHost has metadata to track the 'source' object
+			// that created it. If we do have this metadata, include it with
+			// the virtual host error, so that other components can easily find it
+			if staticMetadata := virtualHost.GetMetadataStatic(); staticMetadata != nil {
+				validation.AppendVirtualHostErrorWithMetadata(vhostReport,
+					validationapi.VirtualHostReport_Error_ProcessingError,
+					fmt.Sprintf("invalid virtual host [%s] while processing plugin %T: %s", virtualHost.GetName(), plugin, err.Error()),
```
sam-heilbron, line 173: nit: Given that plugins all have a name, https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/plugins.go#L20, any reason we wouldn't want to just that, instead of the type?

test/kubernetes/e2e/features/upstreams/suite.go
```diff
@@ -56,12 +54,10 @@ func (s *testingSuite) TestConfigureBackingDestinationsWithUpstream() {
...
-		s.ctx,
 		func() (resources.InputResource, error) {
 			return s.testInstallation.ResourceClients.UpstreamClient().Read(upstreamMeta.GetNamespace(), upstreamMeta.GetName(), clients.ReadOpts{})
 		},
-		gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
-			"State":      gomega.Equal(core.Status_Accepted),
-			"ReportedBy": gomega.Equal("gloo"),
-		}))
+		core.Status_Accepted,
+		"gloo",
```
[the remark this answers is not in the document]
sam-heilbron: Should this use the new constant that was created?

test/kubernetes/e2e/features/virtualhost_options/testdata/bad-vho.yaml
```diff
@@ -0,0 +1,18 @@
...
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: gw
+    sectionName: http
+  options:
+    retries:
+      retryOn: "5xx"
+      retryBackOff:
+        # We expect to receive an error when the base interval is less than max interval. 
+        # ref projects/gloo/pkg/plugins/basicroute/plugin.go
```
sam-heilbron, line 16: I really like this ref in the comment here, makes it easy to track the source

test/kubernetes/testutils/assertions/status.go
```diff
@@ -12,41 +11,43 @@ import (
...
 // Checks GetNamespacedStatuses status for gloo installation namespace
-func (p *Provider) EventuallyResourceStatusMatchesWarningReasons(installNamespace string, getter helpers.InputResourceGetter, desiredStatusReasons []string, desiredReporter string, timeout ...time.Duration) {
+func (p *Provider) EventuallyResourceStatusMatchesWarningReasons(getter helpers.InputResourceGetter, desiredStatusReasons []string, desiredReporter string, timeout ...time.Duration) {
 	ginkgo.GinkgoHelper()
 
 	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
 	gomega.Eventually(func(g gomega.Gomega) {
 		statusWarningsMatcher := matchers.MatchStatusInNamespace(
-			installNamespace,
+			p.glooGatewayContext.InstallNamespace,
```
sam-heilbron, line 24: 🎉

sam-heilbron (dismissed):
LGTM! Small nits/questions, but not blockers

test/kubernetes/e2e/features/virtualhost_options/suite.go
jbohanon: Nope, I just didn't know about that structure. I got it moved to the framework way

sam-heilbron (approved):

npolshakova (approved):

sam-heilbron:
^bulldozer has been finicky, so hoping enabling auto-merge will help

PR #9435, opened 2024-05-02
code cleanup: remove glooctl v2 references

jenshu:
# Description

Remove v2 references from glooctl code (rename to kubegateway).
This should have no user-facing changes. The kubegateway glooctl integration is not currently functional.


npolshakova (approved):

davidjumani (commented):
Does this also need to be changed ?
https://github.com/solo-io/gloo/blob/f8ea746ba9ca8ffed51cdb98654f2499c9d0e8cd/projects/gloo/cli/install.ps1#L23

jenshu:
> Does this also need to be changed ?
> 
> https://github.com/solo-io/gloo/blob/f8ea746ba9ca8ffed51cdb98654f2499c9d0e8cd/projects/gloo/cli/install.ps1#L23

looks like that was added earlier and not related to GG v2 https://github.com/solo-io/gloo/pull/8856

davidjumani (approved):

PR #9437, opened 2024-05-03
Secret Delete Flake Fix

sheidkamp:
# Description
Moving the `verifyGlooValidationWorks` call immediately before the secret is attempted to be deleted.

There are two validation steps at the beginning of the test, `EventuallyResourceAccepted` and `verifyGlooValidationWorks`

I don't have a good explanation of why this works, but here is the empirical evidence:
Before the update I was able to reproduce the error locally by focusing the context and running the tests with `-until-it-fails`

The tests failed on attempts 21, 111, 18 and 68, which amount to 4 failures in 218 tests, with a mean time to failure of ~55 runs.

After updating and re-running the tests, I was unable to replicate with runs of 84, 173, and 150 tests (all manually stopped), for a total of X runs without failure. If the MTF is 55 runs, it is extremely unlikely to have over 400 runs succeed without a flake.

### DIFF
--- a/changelog/v1.17.0-beta26/secret-delete-flake-fix-attempt.yaml
+++ b/changelog/v1.17.0-beta26/secret-delete-flake-fix-attempt.yaml
@@ -1,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/gloo/issues/8826
+    resolvesIssue: false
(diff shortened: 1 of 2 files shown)


sam-heilbron (approved):
Love the empirical evidence analysis!

davidjumani (approved):
Approving based on the evidence

PR #9436, opened 2024-05-03
Add Istio strict/permissive auth e2e tests for Gloo Edge Gateway apis using new testing framework

npolshakova:
# Description

Should go in after https://github.com/solo-io/gloo/pull/9460 which enables CI to run Gloo Gateway API tests in ci. 

This PR adds:
- the Gloo Edge Gateway API setup for automtls enabled/disabled for strict/permissive peer auth
- Glooctl "istio inject"/"istio uninject" command added as a separate setup 

Follows up:
- Add "upgrade" path (Replicate old Istio regression test with first Upstream sslConfig configured, then remove Upstream sslConfig) with Gloo Edge Gateway API
- Add "downgrade" path (Automtls enabled, add sslConfig with old config, do a helm upgrade -i to update settings) with Gloo Edge Gateway API 
- Add Upstream sslConfig overwrite test for Gloo Edge Gateway API setup with disabled peer auth + automtls disabled on Upstream
- Add Upstream sslConfig overwrite test for k8s Gateway API setup with disabled peer auth + automtls disabled on Upstream

## Code changes

New Tests:
- Gloo Edge Gateway API suite with automtls enabled 
    - Headless svc test 
    - Strict/Permissive peerauth
- Gloo Edge Gateway API suite with automtls disabled 
    - Headless svc test 
    - Strict/Permissive peerauth
- Gloo Edge Gateway API suite with no Istio integration enabled. Istio integration is added via `glooctl inject/uninject` 
    - Strict/Permissive peerauth

New Assertions:
- `EventuallyPodsMatches` checks pod(s) match the input gomega matchers. A new matcher `PodHasContainersMatcher` is also added to check the gateway has the `sds` and `istio-proxy` after running `glooctl istio inject`. The inverse assertion is made after `glooctl istio uninject` is run. 

## CI Changes

Adds the following tests to cluster_two ci:

- TestGlooGatewayEdgeGateway
- TestGlooctlIstioInjectEdgeApiGateway
- TestAutomtlsIstioEdgeApisGateway
- TestIstioEdgeApiGateway

```
   go-test-run-regex: '(^TestK8sGatewayIstioAutoMtls$$|^TestK8sGatewayIstio$$|^TestGlooGatewayEdgeGateway$$|^TestGlooctlIstioInjectEdgeApiGateway$$|^TestAutomtlsIstioEdgeApisGateway$$|^TestIstioEdgeApiGateway$$)'
```

[...]

PR #9448, opened 2024-05-06
E2E Test framework updates

jmhbh:
# Description
- While writing the E2E test for the enterprise code I made some modifications to our testing framework to allow for the following features
  1. Add additional schemes to the kube client scheme. This is important because we have schemes that are not present in the OSS version that we need to add to the kube client in the enterprise version.
  2. Expose Clients struct so it can be extended in enterprise

## Code changes
- Add additional schemes to client
- Expose Clients struct so it can be extended in enterprise
## CI changes
- n/a

## Docs changes
- n/a

# Context
- Framework updates for supporting E2E testing in enterprise

## Interesting decisions
- n/a 

## Testing steps
- Verified that it works in enterprise e2e test

## Notes for reviewers
- n/a

# Checklist:


test/kubernetes/e2e/test.go
```diff
@@ -130,17 +132,27 @@ func (i *TestInstallation) String() string {
...
 		err := installFn(ctx)
 		i.Assertions.Require.NoError(err)
 		i.Assertions.EventuallyInstallationSucceeded(ctx)
 	}
 
-	// We can only create the ResourceClients after the CRDs exist in the Cluster
-	clients, err := gloogateway.NewResourceClients(ctx, i.TestCluster.ClusterContext)
-	i.Assertions.Require.NoError(err)
-	i.ResourceClients = clients
+	if generateClientsFn != nil {
```
sam-heilbron, line 146: If this is required, it feels like we have one of two cases:
- A different implementation of the ResourceClient interface. If that is the case, I'd be curious what is unique about it
- A different set of clients (ie additional enterprise clients)

If the second case is what we have, then it feels like we need a distinct EnterpriseTestInstallation, with its unique set of clients and other resources that only it is aware of

test/kubernetes/testutils/clients/clients.go
```diff
@@ -25,7 +25,7 @@ func MustClientset() *kubernetes.Clientset {
 	return clientset
 }
 
-func MustClientScheme() *runtime.Scheme {
+func MustClientScheme(additionalSchemes func(scheme *runtime.Scheme) error) *runtime.Scheme {
```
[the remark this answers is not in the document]
jmhbh: `MustClientScheme` is invoked in the function `MustKindContext` which is defined in OSS and creates the clients using the defined schemes. Personally, i'm in favor of this approach because my other option would be to force the caller of `MustTestCluster` which invokes `MustKindContext` to ingest the client instead which seems worse than defining a function that adds additional schemes to an existing scheme that will be used in the creation of the client. Having said that, i'm open to suggestions and pairing on this

test/kubernetes/testutils/assertions/provider.go
```diff
@@ -23,19 +24,31 @@ type Provider struct {
...
+type AssertionOptions struct {
+	Timeout time.Duration
+	Polling time.Duration
 }
 
 // NewProvider returns a Provider that will provide Assertions that can be executed against an
 // installation of Gloo Gateway
 func NewProvider(t *testing.T) *Provider {
 	gomega.RegisterTestingT(t)
+	assertionOptions := &AssertionOptions{
```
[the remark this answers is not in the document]
jmhbh: I agree, I think we should allow the functions themselves to ingest timeout/polling options instead.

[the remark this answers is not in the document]
jmhbh: This tries to solve the case where api doc discovery does not occur until after the containers that are being targeted are ready. The time it takes these containers to be ready could exceed the default timeout.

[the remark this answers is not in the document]
jmhbh: This tries to solve the case where api doc discovery could exceed the default timeout due to api discovery needing to wait for its target containers to become ready. However, after more thought I think we should default back to the original implementation and account for these cases individually within the tests instead of trying to compensate for these cases here.

test/kubernetes/e2e/test.go
```diff
@@ -130,17 +132,27 @@ func (i *TestInstallation) String() string {
 	return i.Metadata.InstallNamespace
 }
 
-func (i *TestInstallation) InstallGlooGateway(ctx context.Context, installFn func(ctx context.Context) error) {
+func (i *TestInstallation) InstallGlooGateway(
+	ctx context.Context,
+	installFn func(ctx context.Context) error,
+	generateClientsFn func(ctx context.Context, clusterCtx *cluster.Context) (gloogateway.ResourceClients, error),
```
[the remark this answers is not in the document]
jmhbh: Yeah one option is to extend the `TestInstallation` interface in enterprise and define our own `InstallTestEnterprise` method which is probably cleaner.

test/kubernetes/e2e/test.go
jmhbh: IMO the new `ResourceClient` interface is an extension of the original interface + additional methods for the clients where the new interface is used only in enterprise. I'm going to try implementing the new `EnterpriseTestInstallation` method and see if that cleans things up here.

jmhbh: Am trying this out but its not actually working very well since we can only embed the OSS `TestInstallation` in a new struct in enterprise. We can't really override the `ResourceClient` in OSS unless we make OSS aware of the enterprise clients which is not a good idea. In my tests I'm currently casting the OSS interface back to the interface with the enterprise clients as seen here - https://github.com/solo-io/solo-projects/pull/6155/files#diff-497b2426b44c884bb66fad179cc20985fcaa22fcb0348a2e8695e563d78e0737R60 which may be something that we want to consider.

test/kubernetes/testutils/assertions/pods.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/solo-io/k8s-utils/testutils/kube"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func (p *Provider) EventuallyPodsAreReady(ctx context.Context, namespace string, timeout time.Duration, podNames ...string) error {
```
sam-heilbron, line 11: Is there a reason that the namespace would be different than the install namespace?

sam-heilbron (approved):

test/kubernetes/testutils/assertions/pods.go
sam-heilbron, line 11: I think the pattern that other assertions take is that they assert a behavior, instead of returning an error and expecting the caller to perform that assertion. In this case, I imagine we could do the following:
1. Use the existing EventuallyRunningReplicas?
2. Assert that noError is encountered instead of returning an error

jmhbh: yeah, in my test I create my pods in a separate namespace so we can use reference grants and what not so we model a possible user setup.

test/kubernetes/testutils/assertions/pods.go
```diff
@@ -0,0 +1,17 @@
...
+	return kube.WaitUntilPodsRunning(ctx, timeout, namespace, podNames...)
+}
+
+func (p *Provider) FindPodNameByLabel(ctx context.Context, namespace string, labelSelector map[string]string) string {
```
sam-heilbron, line 15: This feels like an action (doing/getting something from the cluster) rather than an assertion (asserting that a cluster has a certain state)

npolshakova: I have a similar pod assertion here: https://github.com/solo-io/gloo/pull/9436/files#r1595906558

npolshakova: Where is this getting used? Is it just to check a pod is created?

test/kubernetes/testutils/clients/clients.go
jenshu, line 28: nit: since this is a function, can we rename additionalSchemes to an action e.g. registerSchemes (which is what you're using in the s-p PR)

test/kubernetes/testutils/assertions/pods.go
jmhbh: good callout I overlooked `EventuallyRunningReplicas` I think its exactly what I need.

test/kubernetes/testutils/clients/clients.go
jmhbh: agreed, `registerAdditionalSchemes` sounds good 👍

test/kubernetes/testutils/assertions/pods.go
jenshu: the k8s-utils func that it calls does do assertions; perhaps we can rename this wrapper func accordingly, e.g. something like MustPodNameByLabel

jmhbh: removed this helper as I'm able to use `EventuallyRunningReplicas` instead.

test/kubernetes/testutils/assertions/pods.go
jmhbh: confirmed this works. Thanks Sam!

npolshakova (approved):

sam-heilbron (approved):

jenshu (approved):

PR #9450, opened 2024-05-06
test/kube/e2e: migrate glooctl tests

sam-heilbron:
# Description

Introduce some new e2e tests, mirroring existing glooctl functionality in our kube2e/glooctl suite

# Context
## Glooctl Tests
Our goal is to progressively migrate end-to-end tests to follow our new framework. To do this effectively, we need the proper foundation to be built out.

This PR attempts to make using `glooctl` easier to do in end-to-end tests, and does so by using it in some tests itself. This does not completely mirror the https://github.com/solo-io/gloo/tree/main/test/kube2e/glooctl tests, because the goal is to do this in small chunks.

## Test TmpDir
As part of the `glooctl debug logs` test, I needed to write to a tmp file. We had hit an issue historically when a "feature" owns the location of the generated files. This resulted in https://github.com/solo-io/solo-projects/blob/main/test/kubernetes/e2e/k8sgateway/k8s_gw_test.go#L69.

I propose an alternate solution, where the tmpDir is owned by the TestInstallation, and it allows us to have isolation between tests.

## Default Cluster Name
Based on some Slack feedback from @lgadbank, it improves the usability in an IDE if we can default the cluster name

## Interesting decisions
- 

## Testing steps
- These new tests run in CI, and they pass
- These changes are imported into https://github.com/solo-io/solo-projects/pull/6173 and run in CI

## Notes for reviewers
-

# Checklist:


test/kubernetes/e2e/migrating_tests.md
```diff
@@ -4,4 +4,14 @@ This new [framework](./test.go) provides a standard way to validate behavior end
...
-...We will be adding details here in follow-up work...
\ No newline at end of file
+## 1. Identify the feature
+As per the [README](./README.md#features), we group our test cases by feature. If the test you are migrating is part of already tested feature, just add the test to that feature package directly. If the test you are migrating is part of a new feature, add a new feature package.
+
+## 2. Identify the TestInstallation(s)
+Features in the product are expected to work under a variety of conditions. Therefore, it may be that you need to invoke the feature test suite from different `TestInstallation`. As per the [README](README.md#testinstallation), we group these TestInstallation by file, so that each file runs the set of tests for a single installation.
+
+## 3. Identify the TestCluster
+We [load balance](./load_balancing_tests.md) our tests across multiple clusters. Since you have added new tests, it may be worthwhile to re-balance the tests. Most importantly, if you added a new file for a `TestInstallation`, assign it a test cluster.
```
sam-heilbron, line 14: This references a document that is being created in https://github.com/solo-io/gloo/pull/9460

lgadban: It looks like there is an outstanding question on the implementation (https://github.com/solo-io/gloo/pull/9460/files#r1594147638) -- should this PR still be reviewed as-is?

sam-heilbron: I believe so yeah. This document around load balancing will always exist, the contents in that doc will vary depending on the implementation that is chosen

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -152,6 +153,20 @@ func (c *Cli) DeploymentRolloutStatus(ctx context.Context, deployment string, ex
 	return c.RunCommand(ctx, rolloutArgs...)
 }
 
+// ScaleDeployment scales the deployment to the desired number of replicas
+func (c *Cli) ScaleDeployment(ctx context.Context, deploymentMeta types.NamespacedName, replicas int) error {
```
npolshakova, line 157: Do we want this to also check all replicas are in the `READY` state after running?

npolshakova: Also is this getting used anywhere currently?

projects/gloo/cli/pkg/testutils/cli.go
```diff
@@ -0,0 +1,60 @@
...
+	cli "github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd"
+)
+
+// NewCli returns an implementation of the Cli
+func NewCli() *Cli {
+	return &Cli{}
+}
+
+// Cli is way to execute glooctl commands consistently
+type Cli struct{}
```
lgadban, line 18: nit: Could/should this be named `GlooCli` ?

projects/gloo/cli/pkg/testutils/cli.go
```diff
@@ -0,0 +1,60 @@
...
+	// Under the hood we call the cobra.Command directly so that we re-use whatever functionality is available to users
+	return cli.CommandWithContext(ctx)
+}
+
+// Execute executes an arbitrary glooctl command
+func (c *Cli) Execute(ctx context.Context, argStr string) error {
+	return ExecuteCommandWithArgs(c.NewCommand(ctx), strings.Split(argStr, " ")...)
+}
+
+// ExecuteOut executes an arbitrary glooctl command
```
lgadban, line 31: nit: Should this godoc have an indication of what `Out` means? The main docs are on `Exec...ArgsOut` but as a user-facing method, would be nice to have a simple notice that the stdout/stderr is returned in the string

projects/gloo/cli/pkg/testutils/utils.go
```diff
@@ -2,38 +2,54 @@ package testutils
...
-
 	errors "github.com/rotisserie/eris"
-
-	"github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd"
 )
 
-func Glooctl(args string) error {
-	app := cmd.GlooCli()
-	return ExecuteCli(app, args)
+func Glooctl(argStr string) error {
```
lgadban, line 20: nit: godoc for this; intention is to pass through some args directly to `glooctl`?

pkg/utils/kubeutils/kubectl/cli.go
sam-heilbron: ha good point. It was done because i saw tests I wanted to migrate that relied on this, and then I kept these changes smaller, and never migrated that test. I will back out this change for now just to keep it tightly scoped.

test/kubernetes/e2e/features/glooctl/suite.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"fmt"
+
+	"github.com/onsi/gomega"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/stretchr/testify/suite"
+)
+
+type testingSuite struct {
```
lgadban, line 13: so this e2e test is purely checking `glooctl check` output against a running, empty cluster?

test/kubernetes/e2e/features/glooctl/check.go
```diff
@@ -0,0 +1,92 @@
+package glooctl
```
npolshakova, line 1: How do we want to handle `glooctl` features that are also related to functional features (`glooctl istio inject`, `glooctl istio uninject`, `glooctl istio disable-mtls --upstream httpbin-upstream`, etc.) 

For example, I have the `glooctl istio inject` as part of the Istio test in this [PR](https://github.com/solo-io/gloo/pull/9436) because it does the inject/uninject, but then needs to run the Istio peer auth tests. After this goes in I think the `glooctl istio inject` should live here as well, wdyt?

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
```diff
@@ -89,10 +91,13 @@ func TestK8sGateway(t *testing.T) {
...
 	t.Run("PortRouting", func(t *testing.T) {
 		suite.Run(t, port_routing.NewTestingSuite(ctx, testInstallation))
-
 	})
 
 	t.Run("RouteDelegation", func(t *testing.T) {
 		suite.Run(t, route_delegation.NewTestingSuite(ctx, testInstallation))
 	})
+
+	t.Run("Glooctl", func(t *testing.T) {
```
npolshakova, line 100: [nit] Can we rename this `GlooctlCheck`? Then I'd add `GlooctlIstio` as a separate suite.

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
```diff
@@ -89,10 +91,13 @@ func TestK8sGateway(t *testing.T) {
 
 	t.Run("PortRouting", func(t *testing.T) {
 		suite.Run(t, port_routing.NewTestingSuite(ctx, testInstallation))
-
 	})
 
 	t.Run("RouteDelegation", func(t *testing.T) {
 		suite.Run(t, route_delegation.NewTestingSuite(ctx, testInstallation))
 	})
+
```
npolshakova, line 99: How are we setting the `GlooctlExecName` path for the test setup? I ended up needing to pass it from the testHelper to the suite [here](https://github.com/solo-io/gloo/blob/bea5c92c023c92960e18bab908bd0c4e89237e72/test/kubernetes/e2e/features/glooctl/istio_inject_suite.go#L29) to use the locally built glooctl.

sam-heilbron: I don't think we actually need it. A nice side effect of the NewCli is that we are using the cobra.Command directly, and not relying on the CLI binary itself.

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
sam-heilbron: 100% I was literally just thinking the same thing

test/kubernetes/e2e/features/glooctl/check.go
sam-heilbron: Yeah I think we split each logical grouping of glooctl into its own suite and then just invoke the suites that way? I'm not sure we need to put each suite in it's own sub-folder, but certainly open to that

test/kubernetes/e2e/features/glooctl/suite.go
sam-heilbron: Yeah this mirrors, I believe, our existing tests. BUt good point, would be useful to add some resources just for more accurate tests

test/kubernetes/e2e/features/glooctl/check.go
npolshakova: Yep, that sounds good! I think separate suites should be good enough (so the istio features would live in `features/glooctl/istio-inject.go`, `features/glooctl/istio-enable-mtls.go`)

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
npolshakova: Sounds good! Then I can rewrite the glooctl command once this goes in to use the cobra.Command as well!

test/kubernetes/e2e/features/glooctl/suite.go
lgadban: If it matches the existing behavior makes sense to me, was just trying to understand the intended test

projects/gloo/cli/pkg/testutils/cli.go
sam-heilbron: I like it. I had originally assumed the `gloo` was implied because we're in the projects/gloo directory. But I think this is a good adjustment

projects/gloo/cli/pkg/testutils/cli.go
sam-heilbron: Good call. I updated the go doc to try to clarify. let me know if there is more you'd like to see

projects/gloo/cli/pkg/testutils/utils.go
sam-heilbron: I like it! I'll admit I had gotten lazy when I saw these without the go-doc but it's a great point to add. Let me know if what I put is clear

test/kubernetes/e2e/k8sgateway/k8s_gw_test.go
sam-heilbron: I updated this to have a broader Glooctl, and then some sub-tests. Let me know if you think that makes sense

test/kubernetes/e2e/features/glooctl/suite.go
sam-heilbron: Yeah I tried to keep the reference to the tests we're mirroring at the top of the file: https://github.com/solo-io/gloo/blob/main/test/kube2e/glooctl/check_test.go. I agree we could and should improve these so that there are actual resources to check.

test/kubernetes/e2e/migrating_tests.md
```diff
@@ -4,4 +4,14 @@ This new [framework](./test.go) provides a standard way to validate behavior end
 
 _This document outlines the ways that developers can migrate tests from that old paradigm into the new framework._
 
-...We will be adding details here in follow-up work...
\ No newline at end of file
+## 1. Identify the feature
```
npolshakova, line 7: Can we add this in one of the testing READMEs? https://github.com/solo-io/gloo/pull/9461#discussion_r1594459103

test/kubernetes/e2e/test.go
```diff
@@ -92,13 +86,22 @@ func (c *TestCluster) RegisterTestInstallation(t *testing.T, glooGatewayContext
...
+		// By creating a unique location, per TestInstallation, we guarantee isolation between TestInstallation
+		GeneratedFiles: MustGeneratedFiles(glooGatewayContext.InstallNamespace),
 	}
 	c.activeInstallations[installation.String()] = installation
 
 	return installation
 }
 
 func (c *TestCluster) UnregisterTestInstallation(installation *TestInstallation) {
+	if err := os.RemoveAll(installation.GeneratedFiles.TempDir); err != nil {
```
npolshakova, line 101: Do we want to add an env to persist the generated files?

npolshakova: Discussed offline- won't add until we have a specific use case for persisting

projects/gloo/cli/pkg/testutils/utils.go
lgadban: This is more than I was expecting, looks great!

test/kubernetes/e2e/migrating_tests.md
sam-heilbron: Updated in the debugging.md file, since that is where we describe other uses cases for developers

Makefile
```diff
@@ -284,6 +290,10 @@ go-test: ## Run all tests, or only run the test package at {TEST_PKG} if it is s
 	$(GO_TEST_ARGS) $(GO_TEST_USER_ARGS) \
 	$(TEST_PKG)
 
+.PHONY: go-test-clean
```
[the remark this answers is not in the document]
sam-heilbron: Yeah this actually can be removed because it's not used. This is a side effect of how we invoke tests in CI: https://github.com/solo-io/gloo/blob/main/.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml#L39. 
Ideally when we invoke go-test we reset the directory, but because we call it 3 times in CI, i was afraid that we would cleanup the directory that may contain relevant data. on second thought, if any of those steps failed, we would short-circuit so I don't think its an issue. I will remove

test/kubernetes/e2e/test.go
```diff
@@ -204,6 +213,34 @@ func (i *TestInstallation) InstallIstioOperator(
 	return ctx.Err()
 }
 
+// GeneratedFiles is a collection of files that are generated during the execution of a set of tests
```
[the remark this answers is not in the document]
sam-heilbron: Good call! This was taking existing code which generated files and centralized it, but im happy to add documentation

npolshakova (approved):

npolshakova (approved):

PR #9455, opened 2024-05-06
Lambda non-string header tests

bewebi:
# Description

Bump the default ENVOY_GLOO_IMAGE to `1.29.3-patch2`, bringing in a fix for a bug that caused Envoy to give 500 responses when a Lambda response contained non-string values in a multi-value header and add e2e tests for the fixed behavior

Also update the AWS test guide to reflect steps taken to get set up locally with AWS tests

## Code changes
- Add e2e test for case where Lambda returns non-string headers with `unwrapAsApiGateway` set to `true` on the proxy
- Add e2e test for case where Lambda returns malformed headers with `unwrapAsApiGateway` set to `true` on the proxy
  - Add a `ContainHeaderKeysExact` matcher to be able to assert that no unexpected headers are included 

## CI changes
- Bump the default value of the `ENVOY_GLOO_IMAGE` env var in the Makefile to `1.29.3-patch2`

## Docs changes
- Update the developer guide for getting set up with AWS tests to reflect steps that were required for me to get set up
  - Remove `Temporary Credentials` section which may be out of date, and which misled me when I was trying to get set up 

# Context

The lambdas used in the new test can be found [here](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/non-string-headers-test) and [here](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/malformed-headers-test)

Slack discussion [here](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1715035855464459)

## Interesting decisions
 
I've followed the existing pattern for AWS Lambda tests in the suite
It seems like there is an opportunity to refactor and/or clean these up somewhat, but that is out of scope for this PR, and not something I'd like to attempt the first time that I'm dealing with this suite

We are asserting the current behavior for malformed headers even though it is not explicitly documented
Should this behavior change we will have to also change the test, but this ensures that the behavior doesn't change inadvertantly

The Envoy PR, solo-io/envoy-gloo#328 has more comprehensive unit tests
It doesn't seem necessary to reproduce all of those here, rather we just want to confirm that we're building with a version of Envoy that contains the fix
If reviewers find that additional test cases would be useful they can be added

The original issue #8867 refers repeatedly to a 302 response/error
This refers to a JSON parsing/reference error, and not an HTTP status code
I did not include references to this code in this PR as I found it to be more confusing than helpful

## Testing steps

In addition to running these tests against the bumped Envoy version, I verified that the tests fail with the previous buggy behavior when run against the prior Envoy version

## Notes for reviewers

It may be worthwhile for reviewers to set up and manually run these tests locally, in addition to checking that CI passes, both to ensure they understand the space, and to provide feedback on the test setup guide changes included in this PR

This will need to be backported to 1.15 and 1.16

# Checklist:


ashishb-solo (changes requested):
Looks great! Bernie and I met on a huddle and demonstrated that the envoy bump fixes the issue. With the bump, the test behaves as expected, and without the bump, the test returns an error.

The last ask is for an additional test to ensure that we are testing an additional failure case. At that point, I'm happy to approve

changelog/v1.17.0-beta26/envoy-bump-and-test-multivalueheader-fix.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: DEPENDENCY_BUMP
+    dependencyOwner: solo-io
+    dependencyRepo: envoy-gloo
+    dependencyTag: v1.29.3-patch2
+    issueLink: https://github.com/solo-io/gloo/issues/8867
+    resolvesIssue: false
+    description: >-
+      Bump envoy-gloo to bring us onto 1.29.3-patch2, fixing a bug where Lambdas returning multiValueHeaders with non-string
+      type with `unwrapAsApiGateway` enabled would lead to a 500 response
```
ashishb-solo, line 10: I found the changelog description a little hard to understand. What do you think of this? (Note that I left out the part about the version bump since that should be captured by the `dependency*` fields.)

```suggestion
      Fix a bug when Lambdas that return non-string types with `unwrapAsApiGateway` enabled would lead to a 500 response
```

test/e2e/aws_test.go
```diff
@@ -345,6 +346,58 @@ var _ = Describe("AWS Lambda", func() {
...
+		createProxy(true, false, false, "non-string-headers-test")
+		validateLambda(lambdaValidationParams{
+			offset:             1,
+			envoyPort:          envoyInstance.HttpPort,
+			expectedSubstrings: []string{`"test body"`},
+			expectedHeaders:    http.Header{"Foo": []string{"null,bar,123"}},
+		})
+	}
+
+	// this tests the case where a lambda returns malformed multiValueHeaders, a case which previously caused Envoy to
```
ashishb-solo, line 374: could you explain exactly _why_ we consider the headers to be malformed ?

test/gomega/matchers/contain_headers.go
```diff
@@ -56,3 +56,19 @@ func ContainHeaderKeys(keys []string) types.GomegaMatcher {
 	matcher := gomega.WithTransform(transforms.WithHeaderKeys(), gomega.ContainElements(keys))
 	return gomega.And(matcher)
 }
+
+// ContainHeaderKeysExact produces a matcher that will only match if all provided header keys exist and no others.
+func ContainHeaderKeysExact(keys []string) types.GomegaMatcher {
```
ashishb-solo, line 61: i would personally prefer calling this `OnlyContainHeaderKeys` (or `OnlyContainHeaderKeysAndValues`), but happy to leave this as it is. Whichever you like the most!

test/e2e/aws_test.go
```diff
@@ -144,7 +145,7 @@ var _ = Describe("AWS Lambda", func() {
 			g.Expect(res).Should(testmatchers.HaveHttpResponse(&testmatchers.HttpResponse{
 				StatusCode: expectedStatus,
 				Body:       testmatchers.ContainSubstrings(params.expectedSubstrings),
-				Custom:     testmatchers.ContainHeaders(params.expectedHeaders),
+				Custom:     And(testmatchers.ContainHeaders(params.expectedHeaders), testmatchers.ContainHeaderKeysExact(params.exactExpectedHeaderKeys)),
```
ashishb-solo, line 148: Do we need both `ContainHeaders` and `ContainHeaderKeysExact`? Can we get rid of `ContainHeaders`? If I understand correctly, `ContainHeaders` can never fail unless `ContainHeaderKeysExact` also fails (but I could be wrong here)

ashishb-solo (commented):
Looks great! A few comments, but hopefully not too difficult to address

changelog/v1.17.0-beta26/envoy-bump-and-test-multivalueheader-fix.yaml
nfuden: Should we escalate this to a fix to make sure it shows up in the changelog? or are we putting that in the EE?

test/gomega/matchers/contain_headers.go
nfuden: nit: for the above comment I would consider prepending a `nit:` tag

test/e2e/aws_test.go
bewebi: `ContainHeaders` checks the _values_ of the headers, but doesn't require that those specified constitute all headers
`ContainHeaderKeysExact` checks that the specified keys constitute the exact set of header keys

In theory either could fail but not the other (eg we check one header via `ContainHeaders` and it's value is wrong, but `ContainHeaderKeysExact` shows that we have all of the expected keys and no others)

In practice I don't expect devs to set both
If we want to formalize that we can only test one or the other that can be implemented and it might read a little cleaner
We could also opt to not strictly assert that we have no extra headers in the malformed case, which would remove the need for all of the new logic that checks that

changelog/v1.17.0-beta26/envoy-bump-and-test-multivalueheader-fix.yaml
bewebi: https://github.com/solo-io/gloo/pull/9455/commits/90c197c58e05afeefcf4056651f7a13d4ae4fb6b

test/e2e/aws_test.go
bewebi: https://github.com/solo-io/gloo/pull/9455/commits/e36b6ea1667a1530b4bdd753544a3da44136ef54

nfuden (approved):

test/e2e/aws_test.go
ashishb-solo: Ah, I see now - I misread `ContainHeaderKeysExact` before and thought it was checking values also. Thanks for explaining!

test/gomega/matchers/contain_headers.go
ashishb-solo: in retrospect, i think the current name is better, so i'm closing this

ashishb-solo (approved):

bewebi:
bulldozer?

PR #9456, opened 2024-05-07
Kube Gateway Policy Validation

lgadban:
# Description

Implements webhook validation for Gloo Gateway Policies in the context of Kubernetes Gateway API.
RouteOption and VirtualHostOption CRs will be hermetically validated to prevent admission for semantically incorrect resources. Referential errors (i.e. "warnings" in the Gloo Edge API) will not be checked.

# Context

Validation is implemented in the existing webhook but does NOT do a full translation like the original Gloo Edge API validation. There are many challenges associated with performing a full translation (see design doc for more analysis).
This approach gives us semantic validation/rejection via a. webhook without needing a snapshot-based full translation.

## Testing steps

Manually verified with resources such as:
```
kind: RouteOption
metadata:
  name: policy-attached
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: HTTPRoute
    name: example-route
  options:
    faults:
      abort:
        percentage: 100
        # httpStatus: 500 <-- commented out to break plugin, this causes error
```
```
Error from server: error when creating "projects/gateway2/examples/example-http-route-with-attached-options.yaml": admission webhook "gloo.gloo-system.svc" denied the request: resource incompatible with current Gloo snapshot: [Validating *v1.RouteOption failed: 1 error occurred:
	* Validating *v1.RouteOption failed: validating *v1.RouteOption name:"policy-attached" namespace:"default": 1 error occurred:
	* Route Error: ProcessingError. Reason: *faultinjection.plugin: invalid abort status code '0', must be in range of [200,600). Route Name:
```

```
kind: VirtualHostOption
metadata:
  name: policy-attached
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: gw
  options:
    bufferPerRoute:
      buffer: {}
```
```
Error from server: error when creating "projects/gateway2/examples/example-http-route-with-attached-vhost-options.yaml": admission webhook "gloo.gloo-system.svc" denied the request: resource incompatible with current Gloo snapshot: [Validating *v1.VirtualHostOption failed: 1 error occurred:
	* Validating *v1.VirtualHostOption failed: validating *v1.VirtualHostOption name:"policy-attached" namespace:"default": 1 error occurred:
	* VirtualHost Error: ProcessingError. Reason: invalid virtual host [vhost]: invalid BufferPerRoute.Buffer: embedded message failed validation | caused by: invalid Buffer.MaxRequestBytes: value is required and must not be nil.
```

Still exploring how we can test this in the new e2e testing framework

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6063


projects/gateway/pkg/validation/validator.go
```diff
@@ -378,6 +372,61 @@ func (v *validator) validateProxiesAndExtensions(ctx context.Context, snapshot *
...
+
+	return proxies, errs
+}
+
+// isK8sGatewayProxy returns true if we are evaluating a Policy resource in the context of K8s Gateway API support.
+// Currently the only signal needed to make this decision is if the resource being evaluated is a RouteOption
+// or a VirthalHostOption, as we only directly evaluate them in the validator to support K8s Gateway API mode.
+func isK8sGatewayProxy(res resources.Resource) bool {
+	switch res.(type) {
+	case *sologatewayv1.RouteOption, *sologatewayv1.VirtualHostOption:
```
sam-heilbron, line 423: Is this a safe assumption if/once we enable support for these resources to be validated with Edge APIs? I guess we only make it into here IFF kubeGateway is enabled and the resource in question is a policy api

projects/gateway2/controller/controller.go
```diff
@@ -261,6 +261,7 @@ func (c *controllerBuilder) watchNamespaces(ctx context.Context) error {
 
 func (c *controllerBuilder) watchRouteOptions(ctx context.Context) error {
 	err := ctrl.NewControllerManagedBy(c.cfg.Mgr).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
sam-heilbron, line 264: Can you help me understand what this does and the impact to the product/translation without it?
My understanding is that this is a predicate (filter) that in essence allows us to skip events where the metadata.Generated is unchanged. This would allow us to avoid re-running translation when status updates occur on these resources. Is that accurate?

Given that this slipped past review before, is it worth it to have an explicit integration test that proves each of our resources can update their status without impacting translation or do you feel that is overkill?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -441,6 +467,11 @@ func (wh *gatewayValidationWebhook) validateGvk(ctx context.Context, gvk schema.
 	var reports *validation.Reports
 	newResourceFunc := gloosnapshot.ApiGvkToHashableResource[gvk]
 
+	// check to see if we should perform validation for this GVK
+	if !wh.shouldValidateGvk(gvk) {
```
sam-heilbron, line 471: We have an existing shouldValidateResource method, which has logic to return early under certain conditions. Could/should we pull this functionality into that function?

projects/gateway/pkg/validation/validator.go
```diff
@@ -337,6 +321,17 @@ func (v *validator) validateProxiesAndExtensions(ctx context.Context, snapshot *
 			continue
 		}
 
+		// if we are validating K8s Gateway Policy resources, we only want the errors resulting from
```
sam-heilbron, line 324: At first I was thinking that I don't like that we have multiple places where we gate the logic for this `validatingK8sGateway` but on second thought I think I like it. We know that this is a short-term solution, so it leaves the code as untouched as possible, and clearly isolates the behavior we're introducing

projects/gateway/pkg/validation/validator.go
```diff
@@ -378,6 +372,61 @@ func (v *validator) validateProxiesAndExtensions(ctx context.Context, snapshot *
 	}
 }
 
+// the returned error is a multierror that may contain errors from several Edge Proxies, although this depends on collectAllErrors
```
sam-heilbron, line 375: My understanding is that this is a lift/shift of code, with no changes. Is that correct?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -973,6 +973,7 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 				}
 			}
 
+			kubeGatewayEnabled := envutils.IsEnvTruthy(constants.GlooGatewayEnableK8sGwControllerEnv)
```
sam-heilbron, line 976: Should we just reuse: `if opts.GlooGateway.EnableK8sGatewayController` from above?

sam-heilbron (commented):
Looking great! I like the direction, accepting it as a viable solution in the short-term

projects/gateway/pkg/validation/validator.go
lgadban: Yes exactly. I could leave this block in the parent function and just branch out for the k8s gateway logic as well.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -24,6 +24,12 @@ webhooks:
       path: "/validation"
     caBundle: "" # update manually or use certgen job or cert-manager's ca-injector
   rules:
+{{- if .Values.kubeGateway.enabled }}
+  - operations: [ "CREATE", "UPDATE" ]
+    apiGroups: ["gateway.solo.io"]
+    apiVersions: ["v1"]
+    resources: ["routeoptions", "virtualhostoptions"]
```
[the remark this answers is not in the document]
lgadban: my inclination was to not consider DELETEs ever, as deleting a RouteOption or VirtualHostOption can never result in a semantic validation failure (at least as far as I can tell). Is there a situation you are thinking would be caught here?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
```diff
@@ -475,6 +506,17 @@ func (wh *gatewayValidationWebhook) validateList(ctx context.Context, rawJson []
 	return reports, nil
 }
 
+func (wh *gatewayValidationWebhook) shouldValidateGvk(gvk schema.GroupVersionKind) bool {
+	if gvk == gwv1.RouteOptionGVK || gvk == gwv1.VirtualHostOptionGVK {
+		// only validate RouteOption and VirtualHostOption resources if K8s Gateway is enabled
+		return wh.kubeGatewayEnabled
+	}
```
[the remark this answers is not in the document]
lgadban: The goal here was to enforce that the webhook implementation will never validate a VHostOpt or RouteOpt unless KubeGW is enabled, irrespective of the webhook k8s config, as they are not tightly coupled.

I don't feel too strongly about this; we definitely could rely solely on the k8s config and remove this check in code.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
```diff
@@ -24,6 +24,12 @@ webhooks:
       path: "/validation"
     caBundle: "" # update manually or use certgen job or cert-manager's ca-injector
   rules:
+{{- if .Values.kubeGateway.enabled }}
```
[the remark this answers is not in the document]
lgadban: Your point is valid, we are definitely relying on the fact that RtOpts and VHostOpts are not validated in Edge today, and using this as a signal is not guaranteed if we ever need to validate RtOpts or VHostOpts in the Edge context.

The flip side is that adding RtOpts and VHostOpts to Edge validation would require a new implementation (more than likely?) as the Edge validation is based on the old snapshot-based full translation pass. I think, in most cases, we would rather add hermetic CEL-validation to these resources, rather than continue adding functionality to the existing Edge model.

I see both sides -- what do you think?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
lgadban: I had initially tried that (`shouldValidateGvk` was inspired by) but `shouldValidateResource` seemed more closely associated with the resource itself.
In my mind the logic is `validateGvk` and `shouldValidateGvk` happen up front before we process the individual resource and unmarshal it etc.
I don't feel strongly about this though, so if you prefer the logic in `shouldVlidateResource` I can definitely make that change.

projects/gateway/pkg/validation/validator.go
lgadban: I also am a fan as it essentially acts as a feature flag within the code. Let me know if you feel otherwise.

install/helm/gloo/templates/5-gateway-validation-webhook-configuration.yaml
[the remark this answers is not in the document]
sam-heilbron: No that's a great point. It might make sense for us to explicitly call that out in a comment though? I imagine I (or others) may come across it in the future, see the asymmetry and lose the context around it being intentional

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook.go
sam-heilbron: Not that's a fair point! If you evaluated it, and came forward with this solution, it works for me

projects/gateway/pkg/validation/validator.go
lgadban: No I don't think this is a safe assumption if we want to validate them with Edge APIs.
However, if that happens, there's no real signal we can use WITHOUT doing a translation pass of some sort.
The easiest way would be to check if the resource has a `targetRef`, and if so it's K8s GW, but even that falls short, as if the policy is used as an ExtensionRef, it will not contain a `targetRef` and will look identical to a delegated [Route/VHost]Option in Edge

projects/gateway2/controller/controller.go
lgadban: You're correct, that is exactly what it will do.

I added this just as a quick fix that I found while working on this but should probably be a standalone change, possibly with tests added as you mentioned. I can write up and issue and remove from this PR.
RE: testing, we would essentially be testing controller-runtime, but I'm not sure that matters, as codifying our expectation should catch regressions, etc.

projects/gloo/pkg/syncer/setup/setup_syncer.go
lgadban: yes, good call

projects/gateway2/controller/controller.go
sam-heilbron: I added this just as a quick fix that I found while working on this but should probably be a standalone change, possibly with tests added as you mentioned. I can write up and issue and remove from this PR. - Up to you! It's a good win, so happy to have it as part of this PR

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -85,6 +85,23 @@ func (c *Cli) Apply(ctx context.Context, content []byte, extraArgs ...string) er
 		Cause()
 }
 
+// ApplyFileOut applies the resources defined in a file, and returns an error if one occurred
+func (c *Cli) ApplyFileWithRunError(ctx context.Context, fileName string, extraArgs ...string) error {
```
npolshakova, line 89: Do we need both this and `ApplyFile`?

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook_test.go
```diff
@@ -431,6 +458,80 @@ var _ = Describe("ValidatingAdmissionWebhook", func() {
...
+					} else {
+						Expect(review.Response.Allowed).To(BeTrue())
+					}
+					Expect(review.Proxies).To(BeEmpty())
+				},
+				Entry("RouteOption, CREATE, auto-accept = accepted", true, v1.RouteOptionCrd, v1.RouteOptionCrd.GroupVersionKind(), v1beta1.Create, routeOption),
+				Entry("VirtualHostOption CREATE, auto-accept = accepted", true, v1.VirtualHostOptionCrd, v1.VirtualHostOptionCrd.GroupVersionKind(), v1beta1.Create, vHostOption),
+				Entry("RouteOption, CREATE, auto-fail = fail", false, v1.RouteOptionCrd, v1.RouteOptionCrd.GroupVersionKind(), v1beta1.Create, routeOption),
+				Entry("VirtualHostOption CREATE, auto-fail = fail", false, v1.VirtualHostOptionCrd, v1.VirtualHostOptionCrd.GroupVersionKind(), v1beta1.Create, vHostOption),
+				// TODO(Law): add UPDATE tests here, need to handle oldObject similar to the status update tests
```
npolshakova, line 527: Will these be added back in a follow up?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -984,6 +984,7 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 					gwOpts.Validation.AlwaysAcceptResources,
 					gwOpts.ReadGatewaysFromAllNamespaces,
 					gwOpts.GlooNamespace,
+					opts.GlooGateway.EnableK8sGatewayController, // ontrols validation of KubeGateway policies (e.g. RouteOption, VirtualHostOption)
```
npolshakova, line 987: [nit] 
```suggestion
					opts.GlooGateway.EnableK8sGatewayController, // controls validation of KubeGateway policies (e.g. RouteOption, VirtualHostOption)
```

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -74,6 +86,10 @@ func (s *testingSuite) AfterTest(suiteName, testName string) {
 	}
 
 	for _, manifest := range manifests {
+		if strings.Contains(manifest, "bad") {
```
npolshakova, line 89: Can we have this be something more specific to the webhook/validation (like `webook-reject` or `invalid-config`)? We have other e2e tests that cause status errors and invalid behavior (the port routing tests are named things like `invalid-port-and-valid-targetport.yaml`, but don't cause webhook rejections).

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -99,12 +115,16 @@ func (s *testingSuite) TestConfigureVirtualHostOptions() {
...
 
 func (s *testingSuite) TestConfigureInvalidVirtualHostOptions() {
-	// Check status is rejected on bad VirtualHostOption
-	s.testInstallation.Assertions.EventuallyResourceStatusMatchesState(
-		s.getterForMeta(&badVirtualHostOptionMeta),
-		core.Status_Rejected,
-		defaults.KubeGatewayReporter,
-	)
+	if s.validationEnabled {
+		// TODO: assert that resource doesn't exist in cluster
```
npolshakova, line 119: We can add a new assertion similar to `s.testInstallation.Assertions.EventuallyObjectsNotExist` but without the eventually?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-no-webhook-validation-test-helm.yaml
```diff
@@ -0,0 +1,40 @@
...
+    invalidRouteResponseCode: 404
+    invalidRouteResponseBody: Gloo Gateway has invalid configuration.
+gateway:
+  persistProxySpec: true
+  logLevel: info
+  validation:
+    # this is not relevant for k8s GW
+    allowWarnings: true
+    # always accept regardless of validation result
+    alwaysAcceptResources: true
```
npolshakova, line 25: Do we want a separate manifest for this? Since it's reusing the virtualhosts options tests, maybe we should just have the validation config as part of `gloo-gateway-test-helm.yaml`?

test/kubernetes/e2e/k8sgateway/k8s_gw_test_no_validation_test.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// TestK8sGatewayNoValidation executes tests against a K8s Gateway gloo install with validation disabled
+func TestK8sGatewayNoValidation(t *testing.T) {
+	ctx := context.Background()
+	testCluster := e2e.MustTestCluster()
+	testInstallation := testCluster.RegisterTestInstallation(
+		t,
+		&gloogateway.Context{
+			InstallNamespace:   "k8s-gw-test",
```
npolshakova, line 28: [nit] Unique namespace? `k8s-gw-test-no-validation`?

test/kubernetes/e2e/k8sgateway/manifests/k8s-gateway-no-webhook-validation-test-helm.yaml
npolshakova: Ah- nevermind, this is the `no-validation` case!

test/kubernetes/e2e/features/virtualhost_options/suite.go
jbohanon: I agree with @npolshakova that we should have a more explicit name filter for the specific cases of webhook validation. I like `webhook-reject`

pkg/utils/kubeutils/kubectl/cli.go
lgadban: No definitely not, this was just a quick example of what was needed to be able to assert based on the full error.

I will consolidate these methods now

test/kubernetes/testutils/runtime/context.go
```diff
@@ -28,9 +28,15 @@ func NewContext() Context {
 		runSource = PullRequest
 	}
 
+	// get clusterName from env var, default to "kind" if not set
+	clusterName := os.Getenv(testutils.ClusterName)
```
sam-heilbron, line 32: I added this fallback here: https://github.com/solo-io/gloo/blob/main/test/kubernetes/testutils/cluster/kind.go#L27. I did it in the cluster context specific to KinD clusters

projects/gateway/pkg/services/k8sadmission/validating_admission_webhook_test.go
lgadban: yeah this is the intent, just requires some dedicated time as the way this is handled in other tests is clumsy

projects/gateway2/controller/controller.go
lgadban: Let's leave it in for now if you agree

test/kubernetes/e2e/features/virtualhost_options/suite.go
lgadban: Implemented here: https://github.com/solo-io/gloo/pull/9456/files#diff-1ed56c43076f4f626d0132fe14d52766bcfee733d00ec6c51b1499ef1bbe3151R51

Went with a more native Gloo Edge API-object style approach to reuse what is already in place in the VHostOptions suite. Let me know what you think

test/kubernetes/e2e/k8sgateway/k8s_gw_test_no_validation_test.go
lgadban: Yes good call!

https://github.com/solo-io/gloo/pull/9456/files#diff-0e297257871d3ecd0413a6cb165abf79ce520b9409b8e28dddb40611a30b94ffR26

test/kubernetes/testutils/runtime/context.go
lgadban: Removed the fallback added in this PR

projects/gateway2/controller/controller.go
sam-heilbron: I'm all for leaving it

sam-heilbron (approved):
LGTM!

projects/gateway/pkg/validation/validator.go
```diff
@@ -378,6 +372,61 @@ func (v *validator) validateProxiesAndExtensions(ctx context.Context, snapshot *
...
+	for _, gatewayAndProxyName := range gatewaysByProxy {
+		proxyName := gatewayAndProxyName.Name
+		gatewayList := gatewayAndProxyName.Gateways
+
+		// Translate the proxy and process the errors
+		proxy, reports := v.translator.Translate(ctx, proxyName, snapshot, gatewayList)
+
+		err := v.getErrorsFromResourceReports(reports)
+
+		if err != nil {
```
npolshakova, line 398: So the errs returned from translation aren't associated with a specific proxy? This looks like what we were doing before, but now there can be multiple k8s gateway proxies returned from [TranslateK8sGatewayProxies](https://github.com/solo-io/gloo/blob/664ef7a102c4629720e573875bc3533ad400dedd/projects/gateway/pkg/validation/validator.go#L291). 

It looks like we group all the errors and return them as part of the [validationOutput](https://github.com/solo-io/gloo/blob/664ef7a102c4629720e573875bc3533ad400dedd/projects/gateway/pkg/validation/validator.go#L371) without associating an error with the proxy the caused it?

npolshakova: I'm guessing this is part of the `skip the more sophisticated error aggregation below` comment?

npolshakova (approved):

test/kubernetes/testutils/assertions/objects.go
```diff
@@ -3,11 +3,14 @@ package assertions
...
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	. "github.com/onsi/gomega"
+	"github.com/solo-io/gloo/test/helpers"
+	"github.com/solo-io/solo-kit/pkg/errors"
```
jbohanon, line 13: nit: aliasing this as `skerrors` or as I've commonly seen `apierrors` or something would help readablitiy

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -99,12 +102,20 @@ func (s *testingSuite) TestConfigureVirtualHostOptions() {
 }
 
 func (s *testingSuite) TestConfigureInvalidVirtualHostOptions() {
-	// Check status is rejected on bad VirtualHostOption
-	s.testInstallation.Assertions.EventuallyResourceStatusMatchesState(
-		s.getterForMeta(&badVirtualHostOptionMeta),
-		core.Status_Rejected,
-		defaults.KubeGatewayReporter,
-	)
+	if !s.testInstallation.Metadata.ValidationAlwaysAccept {
```
jbohanon, line 105: nit: 
1. this conditional doesn't need to be negated
2. we could return early from the conditional and eliminate the else clause, reducing nesting

jbohanon (approved):

sam-heilbron:
Per standup, @jbohanon @npolshakova @sam-heilbron are happy to have this merge as is.

Sam is going to resolve the comments to allow the PR to merge

PR #9458, opened 2024-05-07
Add istio bug report on failure

npolshakova:
# Description

- Checks `httpbin` is in a healthy state after applying the setup manifest. This ensures the Istio sidecar has been successfully injected 
- Bumps the timeout for curl requests as part of the Istio strict/permissive peer auth tests 
- The Istio e2e tests will upload the [bug report ](https://istio.io/latest/docs/reference/commands/istioctl/#istioctl-bug-report)on failure as an artifact. 

# Context

This will help debug why the we needed to switch from the `ubuntu-22.04` runner to the `gme-runner` in solo-projects:
https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1714752477497519

Solo-projects PR: https://github.com/solo-io/solo-projects/pull/6154/files 

Example upload: https://github.com/solo-io/gloo/actions/runs/9022110605/job/24790897959?pr=9458 

# Checklist:


test/kubernetes/e2e/k8sgateway/automtls_istio_test.go
```diff
@@ -40,14 +40,17 @@ func TestK8sGatewayIstioAutoMtls(t *testing.T) {
 	t.Cleanup(func() {
 		if t.Failed() {
 			testInstallation.PreFailHandler(ctx)
+
+			// Generate istioctl bug report
```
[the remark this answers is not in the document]
npolshakova: Yep! I think that's a good next step, especially because the `istioctl bug report` can only be run for the Istio tests. I think the Istio bug report is still nice to have as part of these tests because it collects a lot of useful Istio logs and the state of the cluster:
image

test/kubernetes/e2e/features/istio/no_mtls_suite.go
```diff
@@ -65,7 +71,9 @@ func (s *istioTestingSuite) TestStrictPeerAuth() {
 			curl.WithHostHeader("httpbin"),
 			curl.WithPath("/headers"),
 		},
-		expectedServiceUnavailableResponse)
+		expectedServiceUnavailableResponse,
+		60*time.Second, // Allow for a longer timeout for the request to complete to ensure Istio has propagated the config
```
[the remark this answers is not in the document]
npolshakova: The PeerAuth config needs to be applied to get the correct behavior (especially for the strict peer auth policy case where we expect a `503` but if the config is not applied we'd get a `200`). Based on the bug report is doesn't look like we're resource-constrained, but the curl was failing to get a `503` so it looks like an issue with either 1) httpbin not in the mesh yet, 2) the PeerAuth policy hadn't been applied yet.

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return ctx.Err()
+}
+
+func getIstioctlVersionOrDefault() string {
+	if version := os.Getenv(IstioctlVersionEnv); version != "" {
+		return version
+	} else {
+		return defaultIstioVersion
```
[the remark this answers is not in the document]
npolshakova: So before we would set the default Istio version in the kind script here: https://github.com/solo-io/gloo/blob/main/ci/kind/setup-kind.sh#L20 We could have ci set it, but I was thinking if we have it as a required env it will be an extra thing for devs running the Istio suite to have to set. 

I'm fine making the env required and erroring loudly if it's not set, but maybe add a description of the env to the README?

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -62,4 +62,10 @@ jobs:
     - id: auto-succeed-tests
       if: needs.prepare_env.outputs.should-auto-succeed-kube-tests == 'true'
       run: |
-        echo "Regression tests auto-succeeded"
\ No newline at end of file
+        echo "Regression tests auto-succeeded"
+    - name: Archive istioctl bug report on failure
```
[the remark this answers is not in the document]
npolshakova: Actually renaming the file makes it a little hard to read, I'll create a new output directory `_output/test_failure_report` and upload the bug-report there! As we add more artifacts they can live in `_output/test_failure_report` as well.

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+// Download istioctl binary from istio.io/downloadIstio and returns the path to the binary
+func downloadIstio(ctx context.Context, version string) (string, error) {
```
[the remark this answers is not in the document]
npolshakova: So we are currently doing the same in our kind setup script for the Istio regression tests and for the GME upgrade tests. I think eventually caching this would be ideal, but this can be a follow up as part of testing across Istio versions.

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	goModFile, err := modutils.GetCurrentModPackageFile()
+	if err != nil {
+		// This should never happen
+		contextutils.LoggerFrom(ctx).Panic(err)
+	}
+
+	rootDir := filepath.Dir(goModFile)
+	artifactOutput := filepath.Join(rootDir, "_output")
```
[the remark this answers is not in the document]
npolshakova: Updated this as described here: https://github.com/solo-io/gloo/pull/9458#discussion_r1594583760 Let me know if you want to change the output location. The output path is currently _not_ specific to Istio, but the workflow does check if the bug_report is present because only the Istio report is currently being output. When we add other reports we can change the workflow to be more generic.

test/kubernetes/testutils/cluster/istio.go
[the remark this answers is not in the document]
npolshakova: Updated to fail loudly as part of https://github.com/solo-io/gloo/pull/9458/commits/e4e93294f8bd1e4938b52f028f9662b5aae72e2c

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -95,4 +99,9 @@ jobs:
         cluster-name: kind
         test-args: ${{ matrix.test.go-test-args }}
         run-regex: ${{ matrix.test.go-test-run-regex }}
-
+    - name: Archive bug report directory on failure
```
sam-heilbron, line 102: 🎉

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -95,4 +99,9 @@ jobs:
...
+      if: ${{ failure() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: bug-report
+        path: ./_test/bug_report
```
sam-heilbron, line 107: Is this the path local to the root directory, or to the current file definition?

test/kubernetes/e2e/features/port_routing/testdata/setup.yaml
```diff
@@ -47,4 +47,9 @@ spec:
       command:
         - "tail"
         - "-f"
-        - "/dev/null"
\ No newline at end of file
+        - "/dev/null"
+      resources:
```
sam-heilbron, line 51: nit: Given that we re-use the curl pod so widely, i wonder if we just had a shared manifest to define it once, that other tests could re-use would be useful? Not a blocker by any means

test/kubernetes/testutils/runtime/repo.go
```diff
@@ -0,0 +1,16 @@
+package runtime
+
+import (
+	"os/exec"
+	"strings"
+)
+
+// GlooDirectory returns the root directory of the Gloo project
+// This is determined by running `git rev-parse --show-toplevel` and will vary between different clones of the repository (solo-projects vs. solo-io/gloo)
+func GlooDirectory() string {
```
sam-heilbron, line 10: nit: In my earlier PR I had added this to a base testutils package: https://github.com/solo-io/gloo/blob/main/test/testutils/project.go

test/kubernetes/e2e/test.go
```diff
@@ -104,6 +101,27 @@ func (c *TestCluster) UnregisterTestInstallation(installation *TestInstallation)
 	delete(c.activeInstallations, installation.String())
 }
 
+func (c *TestCluster) AddIstioctl(ctx context.Context) error {
```
sam-heilbron, line 104: nit: I realize I may have pushed you in this direction and I'm more than happy to move forward with it. In https://github.com/solo-io/gloo/pull/9469 I tried to remove the concept of the TestCluster altogether, but I'm happy to back out some of those changes if you think it would be useful

sam-heilbron (approved):

.github/workflows/pr-kubernetes-tests.yaml
npolshakova: Root directory! It uploads the directory: https://github.com/solo-io/gloo/actions/runs/9022110605/job/24790897959?pr=9458

test/kubernetes/e2e/features/port_routing/testdata/setup.yaml
npolshakova: Yeah I've wondering that- I don't really like the idea of having a test helper like `testserver`, but it does feel like we should just define this in a shared file once and have the tests reuse it instead of having each feature responsible for owning the curl pod creation. Maybe a discussion for the larger group?

test/kubernetes/testutils/runtime/repo.go
npolshakova: Oh nice- lemme delete this and use `GitRootDirectory`

test/kubernetes/e2e/test.go
npolshakova: Okay I think then it's worth reverting it to be on TestInstallation!

npolshakova: Updated! https://github.com/solo-io/gloo/pull/9458/commits/7a7c00bb8b515bb8881ec0d223cc12ed92589040

test/kubernetes/testutils/runtime/repo.go
npolshakova: Updated! https://github.com/solo-io/gloo/pull/9458/commits/7a7c00bb8b515bb8881ec0d223cc12ed92589040

sam-heilbron (approved):

test/kubernetes/e2e/features/istio/auto_mtls_suite.go
```diff
@@ -64,7 +69,9 @@ func (s *istioAutoMtlsTestingSuite) TestMtlsStrictPeerAuth() {
 			curl.WithHostHeader("httpbin"),
 			curl.WithPath("/headers"),
 		},
-		expectedMtlsResponse)
+		expectedMtlsResponse,
+		60*time.Second, // Allow for a longer timeout for the request to complete to ensure Istio has propagated the config
```
jenshu, line 73: don't know if this has already been discussed, but if this longer timeout is only needed in enterprise, should we  use a shorter timeout here, allow it to be overridable/configurable, and user a longer timeout in EE?

test/kubernetes/e2e/features/istio/no_mtls_suite.go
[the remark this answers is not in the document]
jenshu: is there some way we can check that the PeerAuth policy has been applied, before attempting the curl?

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -0,0 +1,176 @@
...
+	} else {
+		// Fail loudly if ISTIO_VERSION is not set
+		panic(fmt.Sprintf("%s environment variable must be specified to run", glooruntime.IstioVersionEnv))
+	}
+}
+
+// Download istioctl binary from istio.io/downloadIstio and returns the path to the binary
+func downloadIstio(ctx context.Context, version string) (string, error) {
+	if version == "" {
+		contextutils.LoggerFrom(ctx).Infof("ISTIO_VERSION not specified, using istioctl from PATH")
```
jenshu, line 72: should we require that ISTIO_VERSION be specified, so we don't accidentally use some unsupported/untested istio version that exists in the user's path?

test/kubernetes/e2e/features/istio/no_mtls_suite.go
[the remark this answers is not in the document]
npolshakova: So it looks like the PeerAuth policy/curl timeout isn't the root cause of the solo-projects failure: https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1715181891330739 It's being resource-constrained on the smaller runner. I think we can even revert this curl timeout back since it was passing in oss.

test/kubernetes/e2e/features/istio/auto_mtls_suite.go
npolshakova: I'm going to rever the curl timeout since it wasn't the issue: https://github.com/solo-io/gloo/pull/9458#discussion_r1595935846

test/kubernetes/testutils/cluster/istio.go
npolshakova: I think we want to be able to use untested version. I like setting the `ISTIO_VERSION` and istioctl path in the bash env personally so it's easier to use.

sam-heilbron (approved):

jenshu (approved):

PR #9463, opened 2024-05-08
Add Istio upstream tests

npolshakova:
# Description

Adds Istio integration tests with Upstreams. 

Upgrade/downgrade doc for Istio automtls integration: https://docs.google.com/document/d/1QmapbTCVb6BIrgkmgeL60lSwL5KLLIn_MN2KiRLW-is/edit?usp=sharing 

## Code changes

This introduces the following net new tests:

- Override test for Istio integration with Upstream and automtls disabled for k8s Gateway resources (validated with disabled peer auth policy)
- Override test for Istio integration with Upstream and automtls disabled for Edge Gateway resources  (validated with disabled peer auth policy)
- sslConfig based strict peer auth test for Gloo Gateway API resources
- Upgrade test for Istio Integration with Upstream with sslConfig -> automtls for Edge Gateway resources
- Downgrade test Istio Integration with Upstream with automtls -> sslConfig on Upstream for Edge Gateway resources

These can be added if we decide they are worthwhile: 
- Adding the SSL config via cli with `glooctl istio enable-mtls --upstream httpbin-upstream`, removing the SSL config with `glooctl istio disable-mtls --upstream httpbin-upstream`. Currently the tests directly apply the Upstream with sslConfig since discovery is disabled on the Istio tests. The original Istio regression tests were using discovery and editing the Upstream through the cli. This can cause flakes when updating the Upstream resource. 

Manual validation:
- Sending constant traffic with a pod for the Upgrade tests. [Recording of demo](https://solo.zoom.us/rec/share/SxZWWsjOt4jdnROC0hM6HKsnVRXIqsj9UWFTXezo0K7UMEIunLesoHK5PK3XTFXO.Q-Rg7Lj7rKoTmZ0X?pwd=mzUA6McFxa_FZ3hb416l2nzLFcYVFhcd)

1. Install gloo with automtls enabled. Setup initial resources that use sslConfig on Upstream:
```
kind: VirtualService
metadata:
  name: httpbin-vs
  namespace: automtls-istio-edge-api-test
spec:
  virtualHost:
    domains:
    - httpbin
    routes:
    - matchers:
      - prefix: /
      routeAction:
        single:
          upstream:
            name: httpbin-upstream
            namespace: automtls-istio-edge-api-test
---
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: httpbin-upstream
  namespace: automtls-istio-edge-api-test
spec:
  kube:
    selector:
      app: httpbin
    serviceName: httpbin
    serviceNamespace: httpbin
    servicePort: 8000
  sslConfig:
    alpnProtocols:
    - istio
    sds:
      certificatesSecretName: istio_server_cert
      clusterName: gateway_proxy_sds
      targetUri: 127.0.0.1:8234
      validationContextName: istio_validation_context
```

2. Port-forward gateway:
```
k port-forward deploy/gateway-proxy -n automtls-istio-edge-api-test 8080:8080
```

3. Send constant traffic to check headers
```
for i in {1..1000}
do
  curl localhost:8080/headers -H "host: httpbin" -v
  echo "Request $i sent"
done
```

I checked the `X-Forwarded-Client-Cert` was present by grepping the response:

```
for i in {1..1000}; do echo "Request $i:"; response=$(curl -s -H "host: httpbin" "http://localhost:8080/headers"); echo "$response" | grep -q "X-Forwarded-Client-Cert" && echo "Header found" || echo "Error: Header not found"; done
```

4. Update to Upstream to remove sslConfig

```
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: httpbin-upstream
  namespace: automtls-istio-edge-api-test
spec:
  kube:
    selector:
      app: httpbin
    serviceName: httpbin
    serviceNamespace: httpbin
    servicePort: 8000
```

# Checklist:


test/kubernetes/e2e/test.go
```diff
@@ -202,7 +202,10 @@ func MustGeneratedFiles(tmpDirId string) GeneratedFiles {
 		panic(err)
 	}
 
-	failureDir := filepath.Join(testruntime.PathToBugReport(), tmpDirId)
+	// Get cluster id if it is set in the env to avoid conflicts when uploading artifacts on failed workflow runs
```
sam-heilbron, line 205: nit: We call this function when we create a testInstallation, which has access to the cluster.Context and runtime.Context. It seems like we could have the clusterId on one of those objects and pass it into this function

Makefile
```diff
@@ -81,6 +81,11 @@ TEST_ASSET_DIR ?= $(ROOTDIR)/_test
 BUG_REPORT_DIR := $(TEST_ASSET_DIR)/bug_report
 $(BUG_REPORT_DIR):
 	mkdir -p $(BUG_REPORT_DIR)
+	# Create bug report directory for each cluster in ci (cluster_one, cluster_two, etc.)
```
sam-heilbron, line 84: Is this necessary? I would have imagined that a test would create the necessary directory, and that these changes would be unnecessary

sam-heilbron (approved):

Makefile
npolshakova: So the test creates the `tmp` directory, but the `bug_report` is created here at the top level. If we don't create the cluster id directory, the github workflow can't upload the same directory multiple times: https://github.com/solo-io/gloo/actions/runs/9089758950/job/24981435472 

We could potentially have the github workflow create this though. I can add it as a change in the Istio GatewayParameters PR since that will also need a debug report!

test/kubernetes/e2e/test.go
npolshakova: Good call- I'll add this in the next [follow up PR to add GatewayParameters ](https://github.com/solo-io/gloo/pull/9402) unless there are more requested changes!

test/kubernetes/e2e/features/route_options/types.go
```diff
@@ -30,6 +31,15 @@ var (
 	proxyService = &corev1.Service{
 		ObjectMeta: objectMetaInDefault("gloo-proxy-gw"),
 	}
+	proxyDeployment = &appsv1.Deployment{
+		ObjectMeta: objectMetaInDefault("gloo-proxy-gw"),
+	}
+	ngnixPod = &corev1.Pod{
```
jbohanon, line 37: nit: 
```suggestion
	nginxPod = &corev1.Pod{
```

test/kubernetes/e2e/features/virtualhost_options/types.go
```diff
@@ -24,7 +25,25 @@ var (
...
 	}
-	proxyService = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+	proxyService    = &corev1.Service{ObjectMeta: glooProxyObjectMeta}
+	proxyDeployment = &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "gloo-proxy-gw",
+			Namespace: "default",
+		},
+	}
+	ngnixPod = &corev1.Pod{
```
jbohanon, line 35: nit: 
```suggestion
	nginxPod = &corev1.Pod{
```

test/kubernetes/testutils/assertions/pods.go
```diff
@@ -6,11 +6,32 @@ import (
...
+	ctx context.Context,
+	podNamespace string,
+	listOpt metav1.ListOptions,
+	matcher types.GomegaMatcher,
+	timeout ...time.Duration,
+) {
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+
 	p.Gomega.Eventually(func(g gomega.Gomega) {
 		proxyPods, err := p.clusterContext.Clientset.CoreV1().Pods(podNamespace).List(ctx, listOpt)
```
jbohanon, line 36: nit: probably a holdover from this assertion being copied from a specific test, but this var name `proxyPods` doesn't seem totally accurate anymore

test/kubernetes/e2e/features/virtualhost_options/suite.go
```diff
@@ -40,12 +40,24 @@ func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.
 }
 
 func (s *testingSuite) SetupSuite() {
+	// Check that the common setup manifest is applied
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, setupManifest)
+	s.NoError(err, "can apply "+setupManifest)
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment, exampleSvc, ngnixPod)
+	// Check that test resources are running
+	s.testInstallation.Assertions.EventuallyPodsRunning(s.ctx, ngnixPod.ObjectMeta.GetNamespace(), metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/name=nginx",
```
jbohanon, line 49: would it be feasible to condense these checks to one with a set label selector?
`app.kubernetes.io/name in (nginx,gloo-proxy-gw)`

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -41,15 +41,27 @@ func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.
 }
 
 func (s *testingSuite) SetupSuite() {
+	// Check that the common setup manifest is applied
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, setupManifest)
+	s.NoError(err, "can apply "+setupManifest)
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment, exampleSvc, ngnixPod)
+	// Check that test resources are running
+	s.testInstallation.Assertions.EventuallyPodsRunning(s.ctx, ngnixPod.ObjectMeta.GetNamespace(), metav1.ListOptions{
```
jbohanon, line 49: would it be feasible to condense these checks to one with a set label selector?
`app.kubernetes.io/name in (nginx,gloo-proxy-gw)`

test/kubernetes/e2e/features/istio/gloo_gateway_auto_mtls_suite.go
```diff
@@ -101,5 +136,190 @@ func (s *glooIstioAutoMtlsTestingSuite) TestMtlsPermissivePeerAuth() {
...
+		[]curl.Option{
+			curl.WithHost(kubeutils.ServiceFQDN(metav1.ObjectMeta{Name: defaults.GatewayProxyName, Namespace: s.testInstallation.Metadata.InstallNamespace})),
+			curl.WithHostHeader("httpbin"),
+			curl.WithPath("/headers"),
+			curl.WithPort(80),
+		},
+		expectedMtlsResponse, time.Minute)
+}
+
+func (s *glooIstioAutoMtlsTestingSuite) DisableAutomtlsOverridesSSLConfig() {
```
jbohanon, line 268: Does this need to be prefixed with `Test` in order to be run by the suite?

test/kubernetes/e2e/features/virtualhost_options/suite.go
npolshakova: I think this is the `metav1.ListOptions` requirement and it needs to be in the namespace of the pod so I'd rather keep them separate!

test/kubernetes/e2e/features/route_options/suite.go
npolshakova: https://github.com/solo-io/gloo/pull/9463#discussion_r1603492741

test/kubernetes/e2e/features/virtualhost_options/suite.go
jbohanon: Yeah I missed the namespace distinction. To be clear though, is there something about `metav1.ListOptions` that would restrict the use of set-based label selectors as laid out [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering)?

Makefile
npolshakova: Won't need this anymore- I'm combining the cluster id and the test installation namespace as the output file name

npolshakova: Ah, wait that won't work- it'll run into the same directory upload issue. I'll need add the cluster id to the file path instead.

jbohanon (approved):

test/kubernetes/e2e/features/virtualhost_options/suite.go
npolshakova: Nope, it's just the namespace distinction- if there were more things in the namespace I think it makes sense to use the set-based but for a single item I'd prefer to explicitly select it.

sam-heilbron (approved):

jbohanon (approved):

PR #9464, opened 2024-05-08
ggv2: gateway.networking.k8s.io/gateway-name

stevenctl:
In [GEP-1762](https://gateway-api.sigs.k8s.io/geps/gep-1762/)

```
MUST label all generated resources (Service, Deployment, etc) with gateway.networking.k8s.io/gateway-name: my-gateway (where my-gateway is the name of the Gateway resource).
```

While this is GEP is still marked experimental on the site, this is adopted by Istio: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#resource-attachment-and-scaling.


sam-heilbron:
/kick

Logs: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=75eb3b9f-a196-46d4-ad52-eb31a65ec9b2
Error:
```
Step #1 - "run-tests": ERROR: (gcloud.builds.submit) RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Build and Operation Get requests' and limit 'Build and Operation Get requests per minute' of service 'cloudbuild.googleapis.com' for consumer 'project_number:907026730415'.
Step #1 - "run-tests": - '@type': type.googleapis.com/google.rpc.ErrorInfo
Step #1 - "run-tests":   domain: googleapis.com
Step #1 - "run-tests":   metadata:
Step #1 - "run-tests":     consumer: projects/907026730415
Step #1 - "run-tests":     quota_limit: GetRequestsPerMinutePerProject
Step #1 - "run-tests":     quota_limit_value: '900'
Step #1 - "run-tests":     quota_location: global
Step #1 - "run-tests":     quota_metric: cloudbuild.googleapis.com/get_requests
Step #1 - "run-tests":     service: cloudbuild.googleapis.com
Step #1 - "run-tests":   reason: RATE_LIMIT_EXCEEDED
Step #1 - "run-tests": - '@type': type.googleapis.com/google.rpc.Help
Step #1 - "run-tests":   links:
Step #1 - "run-tests":   - description: Request a higher quota limit.
Step #1 - "run-tests":     url: https://cloud.google.com/docs/quota#requesting_higher_quota
```

sam-heilbron (dismissed):

projects/gateway2/helm/gloo-gateway/templates/_helpers.tpl
```diff
@@ -62,6 +62,7 @@ Selector labels
 {{- define "gloo-gateway.gateway.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "gloo-gateway.gateway.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+gateway.networking.k8s.io/gateway-name: {{ .Release.Name }}
```
jenshu, line 65: seems like this should be moved to `gloo-gateway.gateway.labels` instead, which is used in every proxy resource.

if it's also needed on the created pods, then we can keep it here in selectorLabels too

sam-heilbron:
> can you also add tests to make sure the label gets propagated?

I had a similar thought @jenshu though I couldn't find any existing tests for this code, so I figured add all new tests for this change wasn't worth it. Could you link to the tests that should be amended?

projects/gateway2/helm/gloo-gateway/templates/_helpers.tpl
jenshu: @stevenctl can you confirm that the pods don't need this label? (if they do, then we should have it in both labels and selectorLabels)

jenshu (commented):
[deployer_test](https://github.com/solo-io/gloo/blob/main/projects/gateway2/deployer/deployer_test.go) would probably be a good place to put them, but if it's too heavy a lift we can do that later

sam-heilbron:
/kick build-bot was rate-limited and we have increased the quota

sam-heilbron (approved):

projects/gateway2/helm/gloo-gateway/templates/_helpers.tpl
stevenctl: Oh, the pods do need this label. I just put it back into selector labels which looks to be included in common labels.

jenshu (dismissed):

jenshu (approved):

sam-heilbron (approved):

PR #9465, opened 2024-05-09
refactor: Introduce a validator to validate envoy configs

davidjumani:
# Description

Refactor `bootstrap.ValidateBootstrap()` by moving the `DisableTransformationValidation` check into the transformation plugin. This way the `bootstrap.ValidateBootstrap()` can be used in other areas independent of this setting.
As of now it is called only in the transformation plugin so it is safe to make this change
Required for [this enterprise PR](https://github.com/solo-io/solo-projects/pull/6137)

# Checklist:


sheidkamp (approved):
This makes sense to me

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -560,12 +560,14 @@ func (p *Plugin) validateTransformation(
...
-	p.validationLruCache.Add(transformHash, err)
-	if err != nil {
-		return err
+	// If the user has disabled transformation validation, then always return nil
+	if p.settings.GetGateway().GetValidation().GetDisableTransformationValidation().GetValue() {
+		return nil
 	}
-	return nil
+
+	err = bootstrap.ValidateBootstrap(ctx, FilterName, transformations)
```
sheidkamp, line 568: Suggestion: Since we will be doing this same thing in the WAF plugin, would it make sense to create a "ValidateBootstrapCached" that would take these parameters plus the cache and the metrics and wrap up some of that logic?

sheidkamp (commented):
As I'm thinking about the solo-projects PR a couple thoughts here.

projects/gloo/pkg/plugins/transformation/plugin.go
davidjumani: Good idea. Added a new validator

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -560,12 +560,14 @@ func (p *Plugin) validateTransformation(
 		)
 	}
 
-	err = bootstrap.ValidateBootstrap(ctx, p.settings, FilterName, transformations)
-	p.validationLruCache.Add(transformHash, err)
-	if err != nil {
-		return err
+	// If the user has disabled transformation validation, then always return nil
+	if p.settings.GetGateway().GetValidation().GetDisableTransformationValidation().GetValue() {
```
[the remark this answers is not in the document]
davidjumani: Updated to use a validator as suggested

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -91,6 +83,16 @@ func (p *Plugin) Init(params plugins.InitParams) {
...
 	p.logRequestResponseInfo = params.Settings.GetGloo().GetLogTransformationRequestResponseInfo().GetValue()
+	// Skip validation if disabled
+	if !p.settings.GetGateway().GetValidation().GetDisableTransformationValidation().GetValue() {
+		mCacheHits := utils.MakeSumCounter("gloo.solo.io/transformation_validation_cache_hits", "The number of cache hits while validating transformation config")
+		mCacheMisses := utils.MakeSumCounter("gloo.solo.io/transformation_validation_cache_misses", "The number of cache misses while validating transformation config")
+
+		p.validator = validator.New(ExtensionName, FilterName,
+			validator.WithCacheHitCounter(mCacheHits),
+			validator.WithCacheMissCounter(mCacheMisses))
+	}
```
nfuden, line 94: I think that by moving this we are no longer persisting the cache between runs.
Can you assuage that fear?

projects/gloo/pkg/plugins/utils/validator/validator.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/solo-io/gloo/pkg/utils"
+	"github.com/solo-io/gloo/projects/gloo/pkg/bootstrap"
+	"github.com/solo-io/go-utils/contextutils"
+	"go.opencensus.io/stats"
+	"google.golang.org/protobuf/runtime/protoiface"
+	"k8s.io/utils/lru"
+)
+
+const DefaultCacheSize int = 1024
```
nfuden, line 15: I dont think this needs to be exposed.

projects/gloo/pkg/plugins/utils/validator/validator.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (v validator) ValidateConfig(ctx context.Context, config HashableProtoMessage) error {
+	hash, err := config.Hash(nil)
+	if err != nil {
+		contextutils.LoggerFrom(ctx).DPanicf("error hashing the config, should never happen: %v", err)
+		return err
+	}
+
+	// This transformation has already been validated, return the result
```
nfuden, line 59:
```suggestion
	// This proto has already been validated, return the result
```

projects/gloo/pkg/plugins/utils/validator/validator.go
```diff
@@ -0,0 +1,79 @@
...
+		// Error may be nil here since it's just the cached result
+		// so return it as a nil err after cast worst case.
+		errCasted, _ := err.(error)
+		return errCasted
+	} else {
+		utils.MeasureOne(
+			ctx,
+			v.cacheMisses,
+		)
+	}
```
nfuden, line 74:
```suggestion
	} 
	utils.MeasureOne(
		ctx,
		v.cacheMisses,
	)
```

Since we are already in the code lets simplify this

projects/gloo/pkg/plugins/utils/validator/validator.go
davidjumani: I think it is valuable if users want to either view the default or use a multiple of it. No strong opinions and happy to remove it if needed

projects/gloo/pkg/plugins/transformation/plugin.go
davidjumani: Good catch! Moved it to the NewPlugin() method

projects/gloo/pkg/plugins/utils/validator/validator.go
davidjumani: Updated, thanks

projects/gloo/pkg/plugins/utils/validator/validator.go
davidjumani: Good idea. updated

projects/gloo/pkg/plugins/utils/validator/validator.go
nfuden: Lets add a godoc for this then and keep it exposed

projects/gloo/pkg/plugins/transformation/plugin.go
```diff
@@ -66,15 +62,18 @@ type Plugin struct {
...
 
 func NewPlugin() *Plugin {
+	mCacheHits := utils.MakeSumCounter("gloo.solo.io/transformation_validation_cache_hits", "The number of cache hits while validating transformation config")
+	mCacheMisses := utils.MakeSumCounter("gloo.solo.io/transformation_validation_cache_misses", "The number of cache misses while validating transformation config")
+
 	return &Plugin{
-		validationLruCache: lru.New(1024),
+		validator: validator.New(ExtensionName, FilterName,
+			validator.WithCacheHitCounter(mCacheHits),
+			validator.WithCacheMissCounter(mCacheMisses)),
```
nfuden, line 76: Should we add a test making sure that multiple runs dont blow away our cache? 

Asking given our first approach of moving it over to init

projects/gloo/pkg/plugins/utils/validator/validator.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// HashableProtoMessage defines a ProtoMessage that can be hashed. Useful when passing different ProtoMessages objects that need to be hashed.
+type HashableProtoMessage interface {
+	protoiface.MessageV1
+	Hash(hasher hash.Hash64) (uint64, error)
+}
+
+func (v validator) ValidateConfig(ctx context.Context, config HashableProtoMessage) error {
```
sheidkamp, line 52: Would it be possible to write a few tests for this? Maybe check that the cache is populated after a Validation and that if the cache is seeded with a different value than returned by the real validator, that values gets used? (Plus metric behavior)

projects/gloo/pkg/plugins/utils/validator/validator.go
davidjumani: Added

projects/gloo/pkg/plugins/utils/validator/validator.go
davidjumani: Added

projects/gloo/pkg/plugins/transformation/plugin.go
davidjumani: Added

changelog/v1.17.0-beta27/refactor-transformation-validation.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+- type: NON_USER_FACING
+  description: >
```
nfuden, line 3: nit: If using a multiline description try to format it like >| 

Approving anyways as that only matters for changelog rendering and this is non-user facing.

nfuden (approved):

davidjumani:
kick bulldozer

PR #9466, opened 2024-05-09
gateway2/delegation: implement RouteOptions inheritance

shashankram:
# Description
This change implements policy inheritance, specifically in the context of RouteOptions, such that a child route in a delegation chain may inherit RouteOptions from its ancestors. Parents higher in the hierarchy are given preference when merging the RouteOptions down the delegation chain, i.e. for a tree A->B->C the RouteOptions of A are prioritized first, then the RouteOptions of B augment that of A without overriding the options in A, then the options of C augment the merge of A and B without overriding fields.

Additionally, it does the following:

- Refactors the RouteOption query API to perform merging and prioritization within the query.

- Translator tests for the many scenarios of policy inheritance.

- Converts delegation translator test to a table-driven test.

- E2e tests to verify the inheritance and merge functionality.

## Testing steps

Run the e2e test or manually test policy inheritance with the test manifests
```
go test -timeout 600s -run ^TestK8sGateway$/^RouteDelegation$ ./test/kubernetes/e2e/k8sgateway
```

# Checklist:

- [ ] I have made corresponding changes to the documentation `TODO`


ilackarms (approved):

projects/gloo/pkg/utils/merge.go
```diff
@@ -40,3 +44,25 @@ func isEmptyValue(v reflect.Value) bool {
...
+// ShallowMergeRouteOptions merges the top-level fields of src into dst.
+// The fields in dst that have non-zero values will not be overwritten.
+// It performs a shallow merge of top-level fields only.
+func ShallowMergeRouteOptions(dst, src *v1.RouteOptions) *v1.RouteOptions {
+	if src == nil {
+		return dst
+	}
+
+	if dst == nil {
+		return proto.Clone(src).(*v1.RouteOptions)
```
EItanya, line 57: prefer `src.Clone()`. We generate the code and it's much faster.

projects/gateway2/translator/plugins/routeoptions/query/query.go
```diff
@@ -25,15 +34,59 @@ func NewQuery(c client.Client) RouteOptionQueries {
...
+
+	if len(list.Items) == 0 {
+		return nil, nil
+	}
+
+	out := make([]*solokubev1.RouteOption, len(list.Items))
+	for i := range list.Items {
+		out[i] = &list.Items[i]
+	}
+	utils.SortByCreationTime(out)
```
EItanya, line 89: Why do we have to make another list just to sort and then throw away the list?

shashankram: I moved this logic from route_options_plugin.go. It seems because `SortByCreationTime` requires a client.Object type
```
  input slice contains element of unexpected type v1.RouteOption; all elements must implement client.Object

  Full Stack Trace
    github.com/solo-io/gloo/projects/gateway2/translator/plugins/utils.SortByCreationTime({0x1e1cf20, 0xc00097b7a0})
        /src/code/gloo/projects/gateway2/translator/plugins/utils/sort_objects.go:20 +0x6f
```

projects/gloo/pkg/utils/merge.go
shashankram: fixed

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -85,20 +88,26 @@ func (p *plugin) ApplyRoutePlugin(
...
-	// allow for ExtensionRef filter override
-	filterRouteOptions, err := p.handleFilter(ctx, routeCtx)
-	if err != nil {
-		return err
-	}
-	if filterRouteOptions != nil {
-		routeOptions = filterRouteOptions
+	// allow for ExtensionRef filter override only when this route is not
+	// a child in a delegation tree. With delegation, only targetRef based
+	// policy attachment to a route is supported.
```
EItanya, line 93: Just out of curiosity why is this the case?

projects/gateway2/translator/plugins/routeoptions/query/query.go
EItanya: Ok, not a huge deal for now, just such a waste of allocs.

EItanya (dismissed):
Changes look sound to me. The first use I've ever seen of `container.List` 😆

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
shashankram: Since ExtensionRef is used to override TargetRef based attachment, and we have a different override behavior for child routes in a delegation tree, I decided to keep the impl simple such that we don't support different permutations of overriding the policy. So the only difference for child routes as that ExtensionRef based overrides are not applicable. Should be easy to relax this constraint if we need it.

ilackarms (dismissed):

EItanya (approved):

ilackarms (approved):

PR #9469, opened 2024-05-09
test/kube/e2e: improved dependency injection

sam-heilbron:
# Description
Simplify our test framework further, and enable more dependency injection so we can re-use more of it in our Enterprise Portal e2e tests

# Context
## Background
After speaking with @jmhbh we decided to try to merge https://github.com/solo-io/gloo/pull/9448 first, as it will unblock portal efforts. That work will require some code duplication in Enterprise, and this PR will allow us to delete that duplication in a follow-up.

## Remove TestCluster
Originally our `TestCluster` had state on it, and it "managed" the TestInstallations running on it. We have inverted this relationship and TestInstallations just select a cluster and run against it. This means that we do not need an object that tracks the active installations on a cluster.

A side effect of this is that we remove the register/unregister functions. These were in essence constructors and destructors for a TestInstallation. We now expose two constructors (one simple one for oss, and a generic one that enterprise will use), and we build the destructor into the TestInstallation as a finalizer.

## Collapse `RuntimeContext` and `ClusterContext` on TestInstallation
These properties always existed on a TestInstallation. However, they lived on the TestCluster object which has now been removed.

To support this, we also needed to update the few references in the feature suites.

## Testing steps
- CI continues to pass

## Notes for reviewers
-

# Checklist:

### DIFF
--- a/test/kubernetes/testutils/clients/clients.go
+++ b/test/kubernetes/testutils/clients/clients.go
@@ -25,7 +25,7 @@ func MustClientset() *kubernetes.Clientset {
 	return clientset
 }
 
-func MustClientScheme(registerAdditionalSchemes func(scheme *runtime.Scheme) error) *runtime.Scheme {
+func MustClientScheme() *runtime.Scheme {
 	clientScheme := runtime.NewScheme()
 
 	// K8s API resources
@@ -51,11 +51,6 @@ func MustClientScheme(registerAdditionalSchemes func(scheme *runtime.Scheme) err
 
 	err = v1.AddToScheme(clientScheme)
 	mustNotError(err)
-
-	if registerAdditionalSchemes != nil {
-		err = registerAdditionalSchemes(clientScheme)
-		mustNotError(err)
-	}
 
 	return clientScheme
 }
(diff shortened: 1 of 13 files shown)


test/kubernetes/testutils/clients/clients.go
```diff
@@ -25,7 +25,7 @@ func MustClientset() *kubernetes.Clientset {
 	return clientset
 }
 
-func MustClientScheme(registerAdditionalSchemes func(scheme *runtime.Scheme) error) *runtime.Scheme {
```
jenshu, line 28: what is the replacement for this in enterprise? will it just define its own func to add schemes, instead of calling this one?

jenshu (approved):

test/kubernetes/testutils/clients/clients.go
sam-heilbron: Yeah that's what I was thinking:
```go
func MustEnterpriseClientScheme() *runtime.Scheme {
  scheme := MustClientScheme() // build the oss one
  // do the additional adjustments
 return scheme
}
```

jmhbh (approved):
lgtm

PR #9470, opened 2024-05-09
RouteOptions e2e tests

jbohanon:
# Description

Rewrite the route options e2e tests for the k8s gateway plugin to match the vhopts patterns.

Additionally, new test cases were added to cover different edge cases.

TODO: open issue to track the open questions left by this effort:
- [ ] [Statuses are not tracked for RouteOptions attached via filter extension reference](https://github.com/solo-io/solo-projects/issues/6188)
- [ ] [Currently the RouteOptions query makes opinionated filtering decisions; this should be deferred to the plugin so conflicting resources can have statuses indicating such](https://github.com/solo-io/solo-projects/issues/6189)

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6082


sam-heilbron (commented):
Looking like a great start!

test/kubernetes/e2e/features/route_options/testdata/httproute1-bad-extension.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: gateway.networking.k8s.io/v1
```
npolshakova, line 1: Will any of these tests be rejected by the webhook Lawerence is working on? https://github.com/solo-io/gloo/pull/9456 I'm just thinking the name of this file might cause issues with the rejection logic here: https://github.com/solo-io/gloo/pull/9456#discussion_r1596947976

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -32,55 +40,207 @@ func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.
...
+	}
+
+	manifests, ok := s.manifests[testName]
+	if !ok {
+		s.FailNow("no manifests found for %s, manifest map contents: %v", testName, s.manifests)
+	}
+
+	for _, manifest := range manifests {
+		err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, manifest)
+		s.NoError(err, "can apply "+manifest)
```
npolshakova, line 70: Do we still assert `s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)` in the valid manifest cases?

jbohanon: Sam brought up a good point in the vhopts pr that we are not testing the deployer here. So I removed them there and didn't add them here

test/kubernetes/e2e/features/route_options/testdata/httproute1-bad-extension.yaml
jbohanon: Yeah I agree with your assessment there that the rejection filter needs to be much more explicit. It looks like this would currently be an issue with that PR as written

sam-heilbron (approved):

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -32,55 +40,213 @@ func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.
...
+// 	)
+// }
 
-	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+// will fail until manual setup added
+func (s *testingSuite) TestConfigureRouteOptionsWithMultipleTargetRefManualSetup() {
+	// Manually apply our manifests so we can assert that basic rto exists before applying extra rto.
+	// This is needed because our solo-kit clients currently do not return creationTimestamp
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, setupManifest)
+	s.NoError(err, "can apply "+setupManifest)
```
npolshakova, line 127: I think the `EventuallyObjects` exist check is still valid after we apply the `test/kubernetes/e2e/features/route_options/testdata/setup.yaml`. We don't need it for any of the other manifests, but it would be good to assert the gateway is created by the deployer.

test/kubernetes/e2e/features/route_options/suite.go
npolshakova: I think that makes sense for the other manifests! Do we still want to assert the setup created the gateway? https://github.com/solo-io/gloo/pull/9470/files#r1599069776

test/kubernetes/e2e/features/route_options/suite.go
```diff
@@ -32,55 +40,213 @@ func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.
...
 
-func (s *testingSuite) TestConfigureRouteOptionsWithTargetRef() {
-	s.T().Cleanup(func() {
-		err := s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, targetRefManifest)
-		s.NoError(err, "can delete manifest")
-		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, proxyService, proxyDeployment)
-	})
+func (s *testingSuite) SetupSuite() {
+	// We include tests with manual setup here because the cleanup is still automated via AfterTest
+	s.manifests = map[string][]string{
```
npolshakova, line 45: Since all of these are using the same `setupManifest`, do you think it makes sense to apply if in the `SetupSuite` and then remove it in the `TearDownSuite`?

npolshakova (approved):

sam-heilbron (approved):

test/kubernetes/e2e/features/route_options/suite.go
jbohanon: But none of the features of the route options plugin affect the behavior of the deployer, nor are they contingent upon the deployer's success. Given that, I think it makes more sense to let the routing fail and depend on the deployer tests to catch issues that may arise with that feature.

test/kubernetes/e2e/features/route_options/suite.go
jbohanon: I think that might be a bit cleaner, but I don't think the time lost from applying the setup manifest before each is that huge, and it allows for future tests which chop up or use a different setup for whatever reason

test/kubernetes/e2e/features/route_options/suite.go
npolshakova: I think it's more similar to the post installation checks were we assert the gloo gateway is present. If we remove this assertion, then if a setup file is missing a gateway or something is rejecting the gateway, the test will continue and fail on the longer curl assertion. If we assert here, it fails earlier.

PR #9477, opened 2024-05-13
kube e2e test cleanup

jenshu:
# Description

Did some code cleanup of the kubernetes e2e tests:
- Consolidated the k8s / edge gateway e2e tests into `test/kubernetes/e2e/tests`
- Consolidated test helper functions into `test/kubernetes/testutils`
- Moved example tests into `test/kubernetes/e2e/example` (not run in CI)
- Default the test istio version so it doesn't panic when you don't set the env var
- Some updates to debugging doc

# Checklist:


test/kubernetes/e2e/tests/manifests/edge-gateway-test-helm.yaml
npolshakova, line 1: Do we want these to live in `testdata` instead of `manifest`?

jenshu: the helm values were previously all in `manifests`, and the resources to apply are in `testdata`, do we want to keep them separate?

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -15,9 +15,13 @@ import (
 	"github.com/solo-io/go-utils/contextutils"
 )
 
+const (
+	defaultIstioVersion = "1.19.9"
```
npolshakova, line 19: So the default was intentionally removed after this conversation: https://github.com/solo-io/gloo/pull/9458#discussion_r1594091933 I'm not sure we want a hardcoded default instead of failing loudly if it's not set in the env. Not having a default also makes it easier to have the Istio version set as an env to test across Istio versions and make sure we're not using a default in those tests.

test/kubernetes/testutils/assertions/status.go
```diff
@@ -67,3 +67,23 @@ func getResourceNamespacedStatus(getter helpers.InputResourceGetter) (*core.Name
 
 	return namespacedStatuses, nil
 }
+
+// AssertHTTPRouteStatusContainsSubstring asserts that at least one of the HTTPRoute's route parent statuses contains
+// the given substring in its message. A description can optionally be provided to the assertion.
+func (p *Provider) AssertHTTPRouteStatusContainsSubstring(
```
npolshakova, line 73: Can we create a matcher on the status instead of just passing the `substr` and `description`? Like we do for the HTTPRequest match: https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_http_response.go#L57

test/kubernetes/e2e/tests/edge_gw_test.go
```diff
@@ -1,4 +1,4 @@
-package gloo_gateway_edge_test
+package tests_test
```
npolshakova, line 1: Why are we renaming the package to `tests_test`? Does that match the solo-projects structure? I think the old package name was more useful, but maybe we could rename this to `e2e_test`?

test/kubernetes/testutils/cluster/istio.go
jenshu: added this because whenever running locally, we need to always add ISTIO_VERSION=... otherwise the tests will panic. i started updating all the readmes to add in the ISTIO_VERSION but realized it would be cleaner to just default it in one place (i.e. i agree with your reply in the linked comment above).

we already are defaulting it in some places e.g. setup-kind `ISTIO_VERSION="${ISTIO_VERSION:-1.19.9}"` so i don't think it's unreasonable to default it here in the tests too, to the same version

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
           kubectl-version: 'v1.28.4'
           helm-version: 'v3.13.2'
           go-test-args: '-v -timeout=25m'
-          go-test-run-regex: '(^TestK8sGatewayIstioAutoMtls$$|^TestK8sGatewayIstio$$|^TestClassicEdgeGateway$$)'
+          go-test-run-regex: '(^TestK8sGatewayIstioAutoMtls$$|^TestK8sGatewayIstio$$|^TestGlooGatewayEdgeGateway$$)'
```
npolshakova, line 74: This is (and the other new edge gloo gateway api tests) are getting enabled in https://github.com/solo-io/gloo/pull/9436 as well

test/kubernetes/e2e/tests/edge_gw_test.go
jenshu: yep it matches solo-projects
because the pkg name is `tests`

test/kubernetes/testutils/cluster/istio.go
npolshakova: I think we do want it to loudly panic if the `ISTIO_VERSION` is not set in the env. 

The setup-kind script is run from ci versus this would be hard coded in the test code. This adds an additional thing to upgrade as we change supported Istio version and is more likely to get out of sync.

The follow up to the Istio tests is to add a test matrix across Istio versions. If one of those versions is not correctly set in the github workflow matrix, it will still pass the test because we have this default hard coded instead of failing loudly. 

GME also doesn't hard code the version in the test setup and only relies on the env: https://github.com/solo-io/gloo-mesh-enterprise/blob/main/test/scripts/setup-funcs.sh#L334

test/kubernetes/e2e/tests/edge_gw_test.go
npolshakova: Do we need the `tests` directory? We already have it twice in the path then with `test/kubernetes/e2e/tests/`. The solo-projects path doesn't seem to have the second test directory: https://github.com/solo-io/solo-projects/tree/main/test/kubernetes/e2e

jenshu: i've mirrored the structure from this (not yet merged) PR https://github.com/solo-io/solo-projects/pull/6155
under `test/kubernetes/e2e` we have:
- `examples` (not run in ci)
- `features` (reusable across tests)
- `tests` (the main entrypoint to running the tests)

test/kubernetes/testutils/assertions/status.go
jbohanon: Would probably make sense to extend the pattern here to include kube statuses https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_status.go

test/kubernetes/testutils/cluster/istio.go
```diff
@@ -57,12 +61,14 @@ func installIstioOperator(
...
 }
 
-func getIstioVersion() string {
+func getIstioVersion(ctx context.Context) string {
 	if version := os.Getenv(glooruntime.IstioVersionEnv); version != "" {
 		return version
 	} else {
-		// Fail loudly if ISTIO_VERSION is not set
-		panic(fmt.Sprintf("%s environment variable must be specified to run", glooruntime.IstioVersionEnv))
+		contextutils.LoggerFrom(ctx).Infof("%s environment variable is not set; using default version %s",
```
jbohanon, line 68: If this defaulting stays, should we consider this an error? default log level is info so I'm not sure it matters much

.github/workflows/pr-kubernetes-tests.yaml
jenshu: reverted 7d370f1

test/kubernetes/testutils/cluster/istio.go
jenshu: reverted 7d370f1

test/gomega/matchers/have_kubegateway_status.go
```diff
@@ -0,0 +1,84 @@
...
+type KubeGatewayRouteStatus struct {
+
+	// Custom is a generic matcher that can be applied to validate any other properties of a k8s gateway RouteStatus
+	// Optional: If not provided, does not perform additional validation
+	Custom types.GomegaMatcher
+}
+
+// HaveKubeGatewayRouteStatus produces a matcher that will match if the provided status matches the
+// actual status
+func HaveKubeGatewayRouteStatus(expected *KubeGatewayRouteStatus) types.GomegaMatcher {
```
npolshakova, line 23: 🎉

npolshakova (approved):

jmhbh (approved):

npolshakova (approved):

PR #9471, opened 2024-05-09
k8s gateway extensions: allow customizing translator

stevenctl:
# Description

## API changes

The only API is an internal code API to allow GME to take control of the Gateway translation. 

Changed `CreatePluginRegistry` to `GetPluginRegistry`. AFAICT there is no need to recreate the registry each time but I can rever tthis if needed.  

```go
	// GetPluginRegistry exposes the plugins supported by this implementation.
	GetPluginRegistry() registry.PluginRegistry
```

Added: `GetTranslator` to allow conditionally switching the translation logic based on the type of gateway. 

```go
	// GetTranslator allows an extnsion to provide custom translation for
	// different gateway classes.
	GetTranslator(*apiv1.Gateway) translator.K8sGwTranslator

```

# Context

See slack conversation [here](https://solo-io-corp.slack.com/archives/some/post)

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [] I have added tests that prove my fix is effective or that my feature works

PR #9482, opened 2024-05-14
ggv2: allow extra gateway classes

stevenctl:
# Description

When using gloo as a module, an additional list of gateway classes can be passed via `bootstrap.Opts`. 
These gateway classes will get deployed using the same logic as `gloo-gateway`. 

Used in conjunction with https://github.com/solo-io/gloo/pull/9471 and GatewayParameters, it's possible to have very custom types of gateways that still use gloo plugins and partially re-use gloo logic. 

## API changes

Code/Internal API: Added `ExtraGatewayClasses` to bootstrap opts.

## Code changes
- `GWClass` is now a set `GWClasses`

# Context

Gloo waypoints. 

## Interesting decisions
 
## Testing steps

I manually verified behavior by hardcoding an entry in the list. 

# Checklist:

- [  I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/changelog/v1.17.0-beta28/extra-gwclass.yaml
+++ b/changelog/v1.17.0-beta28/extra-gwclass.yaml
@@ -1,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
+    description: >- 
+      Allow configuring additional gatewayclasses for controller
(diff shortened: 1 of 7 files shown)


stevenctl:
wip, adding ut

stevenctl:
combined with #9471

PR #9485, opened 2024-05-14
proxy api: support destination ip/port matching

stevenctl:
# Description

Allow filter chain matches to match on destination IP/port.

## API changes
In `Matcher` added:

- uint32 destination_port
- CidrRange[] prefix_ranges

(continuing to duplicate what envoy has)

## Code changes

Implement the translation from Proxy to Envoy xDS for the new fields.
This was only done in the same places that SourcePrefixRanges were already uses. 

# Context

Related to #9471 

## Interesting decisions
 
## Testing steps

I manually verified behavior by relying on it in an experimental fork. 

# Checklist:


sam-heilbron (commented):
Changes LGTM! Could we add some basic unit tests to demonstrate the new matching capabilities and perhaps an e2e test as well to show that the traffic is routed appropriately? https://github.com/solo-io/gloo/blob/main/test/e2e/hybrid_test.go#L172 Might be a good set of tests to build off of

stevenctl:
@sam-heilbron should be good to go

projects/gateway/api/v1/gateway.proto
```diff
@@ -183,6 +183,9 @@ message Matcher {
     // CidrRange specifies an IP Address and a prefix length to construct the subnet mask for a CIDR range.
     // See https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-cidrrange
     repeated .solo.io.envoy.config.core.v3.CidrRange source_prefix_ranges = 2;
+    repeated .solo.io.envoy.config.core.v3.CidrRange prefix_ranges = 4;
+
+    uint32 destination_port = 5;
```
jenshu, line 188: can you add comments for the new fields?

jenshu, line 188: should we use `UInt32Value` here to differentiate between set and not set? or is 0 equivalent to "not set"?

projects/gateway/api/v1/gateway.proto
```diff
@@ -183,6 +183,9 @@ message Matcher {
     // CidrRange specifies an IP Address and a prefix length to construct the subnet mask for a CIDR range.
     // See https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/address.proto#envoy-v3-api-msg-config-core-v3-cidrrange
     repeated .solo.io.envoy.config.core.v3.CidrRange source_prefix_ranges = 2;
+    repeated .solo.io.envoy.config.core.v3.CidrRange prefix_ranges = 4;
```
jenshu, line 186: to support DelegatedHttpGateways / DelegatedTcpGateways, we may also need to add the new fields to [matchable_http_gateway](https://github.com/solo-io/gloo/blob/main/projects/gateway/api/v1/matchable_http_gateway.proto) and [matchable_tcp_gateway](https://github.com/solo-io/gloo/blob/main/projects/gateway/api/v1/matchable_tcp_gateway.proto) (depending on whether it's applicable to http and/or tcp gateways)

projects/gateway/api/v1/gateway.proto
stevenctl: 0 treated as not set, but we can do that for consistency

projects/gateway/api/v1/gateway.proto
stevenctl: tbh, I only added things to `gateway.proto` because it was easier to get test coverage from existing tests :/ maybe I should back this out and write a fresh testing setup that takes Proxy directly.

stevenctl: Is there any requirement for parity between Proxy and Gateway?

stevenctl: Adding new tests from scratch for this is a little painful for me. If someone comes along with a usecase for this outside of Waypoints, they should add it separately.

EItanya: I agree with Steven on this one, let's focus on the current use-case and we can add them elsewhere if need be.

projects/gateway/pkg/translator/hybrid_translator.go
```diff
@@ -18,9 +18,7 @@ var _ ListenerTranslator = new(HybridTranslator)
 
 const HybridTranslatorName = "hybrid"
 
-var (
-	EmptyHybridGatewayMessage = "hybrid gateway does not have any populated matched gateways"
-)
+var EmptyHybridGatewayMessage = "hybrid gateway does not have any populated matched gateways"
```
EItanya, line 21: nit: why is this `var` and not `const`. I know this isn't your change but a listing rule of some kind

projects/gateway/api/v1/gateway.proto
EItanya: Actually, do we need them here at all?I don't think we need to modify the gloo-edge `Gateway` at all. Is there any reason we need classic edge to support this feature for waypoints to work?

EItanya (commented):
Changes make sense to me, except for the addition of changes to the gloo-edge `Gateway` resource as this is meant for `ggv2`

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -284,21 +320,48 @@ func applySourcePrefixRangesToFilterChain(
...
+	filterChain *plugins.ExtendedFilterChain,
+	prefixRanges []*v3.CidrRange,
+) {
+	if filterChain == nil || len(prefixRanges) == 0 {
+		// nothing to do
+		return
+	}
+
+	ensureMatch(filterChain)
+	filterChain.GetFilterChainMatch().PrefixRanges = buildEnvoyRanges(prefixRanges)
```
EItanya, line 337: Is `FilterChainMatch` guaranteed to be non-nil at this point?

stevenctl: Yeah, that's what `ensureMatch` does.

stevenctl: Could make it more like  `filterChain := ensureMatch(filterChain)` if that's a nicer style?

EItanya: No just double checking

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -40,8 +45,7 @@ type ResourceAssertionHandler func(
 type ReportAssertionHandler func(
 	proxyReport *validation.ProxyReport)
 
-var _ = Describe("Listener Subsystem", func() {
-
+var _ = FDescribe("Listener Subsystem", func() {
```
EItanya, line 48: Unfocus

EItanya (dismissed):

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -148,6 +150,7 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
 		s.syncStatus(ctx, rm, gwl)
 		s.syncRouteStatus(ctx, rm)
 		s.reconcileProxies(ctx, proxies)
+		contextutils.LoggerFrom(ctx).Debugf("resynced %d proxies in %v", len(proxies), time.Since(startTime))
```
lgadban, line 153: this may be a bit misleading; `reconcileProxies` does not synchronously update the proxies in question, so we won't have an accurate measurement here of the time to sync the proxies

stevenctl, line 153:
```suggestion
		contextutils.LoggerFrom(ctx).Debugf("translated and wrote %d proxies in %v", len(proxies), time.Since(startTime))
```

stevenctl:
```suggestion
		contextutils.LoggerFrom(ctx).Debugf("translated and wrote %d proxies in %v", len(proxies), time.Since(startTime))
```

stevenctl: How is this?

Rachael-Graham (commented):
docs

projects/gloo/api/v1/proxy.proto
```diff
@@ -220,9 +220,23 @@ message MatchedTcpListener {
 }
 
 message Matcher {
```
lgadban, line 222: thanks for adding docs to all these fields, really useful!

lgadban (approved):
LGTM but looks like codegen needs to run (again for some reason).
I'll approve now but can reapprove when CI passes

EItanya:
I'm also happy to approve once dogs are fixed

lgadban:
@EItanya 
>I'm also happy to approve once dogs are fixed
which dogs are you referring to? :)

lgadban:
buildbot?

EItanya (approved):

lgadban (approved):

PR #9486, opened 2024-05-14
gateway2/route-options: merge ExtensionRef override with targetRef attachment

shashankram:
# Description

Implements merging of RouteOptions attached via ExtensionRef filter and targetRef based policy attachment. ExtensionRef based options have higher priority than targetRef based options, such that if the ExtensionRef based option sets a field, it cannot be overridden by a policy using targetRef. However, unset fields on a policy selected by ExtensionRef may be overridden by targetRef based options.

The merging semantics is already followed when policies are merged along a delegation chain, so this change is required for consistency.

The core of this change to merge policies in order of their priority within the query resolver itself because the resolver is responsible for recursively resolving the policies for delegated routes. Hence, existing lookup of ExtensionRef based options are moved into the query API that also handles the merge.

## Testing steps

Apply RouteOptions to an HTTPRoute using both ExtensionRef filter and targetRef

# Checklist:

# Future work
https://github.com/solo-io/solo-projects/issues/6204

### DIFF
--- a/changelog/v1.17.0-beta28/deleg-extref.yaml
+++ b/changelog/v1.17.0-beta28/deleg-extref.yaml
@@ -1,0 +1,22 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/solo-projects/issues/6161
+    resolvesIssue: false
+    description: |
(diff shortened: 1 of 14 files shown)


EItanya (approved):
LGTM

ilackarms (approved):

PR #9195, opened 2024-02-26
add_additionalLabels_to_chart

TheCubicleJockey:
# Description
This new feature is to add additionalLabels to helm chart

# Context
See issue [here](https://github.com/solo-io/gloo/issues/9035)

## Testing steps
I manually verified behavior by templating the helm chart with different values

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

Issue #9035, opened 2024-01-05
Add custom label to all objects in kubernetes chart


TheCubicleJockey:
### Do you have a suggestion for code improvement or tracking existing technical debt? Please describe.

This chart improvement would allow the user to include a defined label in the value.yaml, and that label would be included in the kubernetes objects in metadata.label where allowed by kubernetes. I can write the PR and it would be a short PR and clear PR.

### Describe the solution you'd like

Using something like commonLabels at the top level of the values file, that can be optionally included with all the kubernetes object that allow for labeling.

### Additional Context

Labels are good in kubernetes for organization and for some custom tools that read labels as a part of rule sets.

TheCubicleJockey:
Am I good to make the PR?

TheCubicleJockey:
My use case is for the clusterrole to have a label. So, I could make one specifically for the clusterrole, but I figured most charts have a values.commonLabels or a values.additionalLabels as a more generally acceptable version that would fit most need and my needs.

If i am doing the work, I would like to help improve the chart in a way that you or your team might like it improved.

TheCubicleJockey:
PR for this work https://github.com/solo-io/gloo/pull/9195

PR #9491, opened 2024-05-15
helm: Add additional labels to charts

davidjumani:
Continuation of https://github.com/solo-io/gloo/pull/9195

# Description
This new feature is to add additionalLabels to helm chart

# Context
See issue [here](https://github.com/solo-io/gloo/issues/9035)

## Testing steps
I manually verified behavior by templating the helm chart with different values

# Checklist:


Rachael-Graham (dismissed):

install/test/helm_test.go
```diff
@@ -6757,6 +6767,21 @@ metadata:
 				Expect(resources.NumResources()).To(Equal(2))
 			})
 
+			// As every resource created should have the additional labels, testing everyone of them can be tedious.
+			// Instead just validating that one of them has it is sufficient to ensure that the chart renders with the additional labels
```
davidjumani, line 6771: Any thoughts on this ?

sheidkamp: I think there's 2 things to test:
* that the include template renders correctly
* that the template is included everywhere it need to be

This test addresses the first point. I agree that tests in this format for the second point would be cumbersome. Maybe we can address this with some kind of grep/shell command that can look for manual applications of `app: gloo` and `labels` that don't use the include? The command could then be added to the PR description and this comment for later reference.

install/helm/gloo/generate/values.go
```diff
@@ -38,6 +38,7 @@ type Global struct {
 	Graphql              interface{}        `json:"graphql,omitempty" desc:"(Enterprise Only): GraphQL configuration options."`
 	ConfigMaps           []*GlobalConfigMap `json:"configMaps,omitempty" desc:"Config used to create ConfigMaps at install time to store arbitrary data."`
 	ExtraCustomResources *bool              `json:"extraCustomResources,omitempty" desc:"Add additional custom resources to create, as defined by a helm partial. Defaults to false in open source, and true in enterprise."`
+	AdditionalLabels     map[string]string  `json:"additionalLabels,omitempty" desc:"Additional labels to add to gloo resources."`
```
sheidkamp, line 41:
```suggestion
	AdditionalLabels     map[string]string  `json:"additionalLabels,omitempty" desc:"Additional labels to add to all gloo resources."`
```

sheidkamp (commented):
Some thoughts on the tests

install/test/helm_test.go
davidjumani: Thanks. Found a way to expect the labels on all resources

sheidkamp: I like this test, but it only covers the resources that happen to get rendered with the default helm.

Can we still add command line type of check to examine every template?

sheidkamp (changes requested):

install/test/helm_test.go
davidjumani: The output of `helm template` is converted and checked for the required labels (just like a grep)
I've added custom helm settings to enable components that generate their own resources

sheidkamp: Even better

sheidkamp (approved):

Issue #8906, opened 2023-11-15
Change default repository from Quay to gcr


nfuden:
### Gloo Edge Product

Open Source

### Gloo Edge Version

1.16.0-beta

### Is your feature request related to a problem? Please describe.

Today we push to both quay and gcr on publish of an image. 
Our default in our helm charts is quay
Quay keeps having issues meaning anyone who has not put a private registery in place has issues pulling issues
Quay keeps having problems so our ci can have issues for its upgrade tests

### Describe the solution you'd like

Change the default from quay to gcr
Emphasize in our production hardening guidelines that users should have a private artifact repo that they pull the images to.

### Describe alternatives you've considered

n/a

sam-heilbron:
If we do this, we should update our internal tests to rely on this as well. I had tried this on a Portal PR recently and some images seemed to work and others didn't, and this varied across versions of Edge that were installed (https://github.com/solo-io/dev-portal/actions/runs/6870753093/job/18705485422)

PR #9493, opened 2024-05-16
Default GatewayParameters

jbohanon:
# Description

TODO:
- e2e tests
- 

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6113


pkg/utils/utils.go
jbohanon, line 1: remove and use k8s util https://github.com/kubernetes/utils/blob/fe8a2dddb1d0/ptr/ptr.go#L50

projects/gateway2/deployer/deployer.go
```diff
@@ -159,24 +159,92 @@ func (d *Deployer) getGatewayParametersForGateway(ctx context.Context, gw *api.G
...
+	logger := log.FromContext(ctx)
+
+	// check for a gateway params annotation on the Gateway
+	paramRef := gwc.Spec.ParametersRef
+	if paramRef == nil {
+		return nil, eris.Errorf("no default GatewayParameters associated with GatewayClass %s/%s", gwc.GetNamespace(), gwc.GetName())
+	}
+	gwpName := paramRef.Name
+	if gwpName == "" {
+		// there is no custom GatewayParameters; use default values.
```
sam-heilbron, line 207: I'm a bit confused. What are the default values? I thought that they are defined on the GatewayParameters that we were unable to look up?

projects/gateway2/deployer/values_helpers.go
```diff
@@ -110,14 +110,14 @@ func getDefaultSdsValues(defaultSds extensions.Image) *helmSds {
 	      tag: ""
 	*/
 	defaultSdsImage := helmImage{
-		Repository: ptr.To(defaultSds.Repository),
-		Tag:        ptr.To(defaultSds.Tag),
-		Registry:   ptr.To("quay.io/solo-io"),
+		Repository: &defaultSds.Repository,
```
sam-heilbron, line 113: nit: Any reason we don't use the PointerTo utility from this PR?

projects/gateway2/deployer/values_helpers.go
```diff
@@ -135,13 +135,14 @@ func getSdsValues(sdsConfig *v1alpha1.SdsIntegration, defaultSds extensions.Imag
 	// if sdsConfig is not nil, but unset use defaults for sds config
 	if sdsConfig.GetSdsContainer() == nil {
 		sds = getDefaultSdsValues(defaultSds)
-		sds.Istio = getIstioValues(sdsConfig.GetIstioIntegration())
 	} else {
 		// Use GatewayParameter overrides if provided
 		var bootstrap *sdsBootstrap
-		if sdsConfig.GetSdsContainer() != nil {
+		if sdsContainer := sdsConfig.GetSdsContainer(); sdsContainer != nil {
```
sam-heilbron, line 141: nit: Is this guard necessary? We are in the else condition of `GetSdsContainer == nil` so I just assumed that this is implied

projects/gateway2/extensions/extensions.go
```diff
@@ -82,6 +82,7 @@ func (e *k8sGatewayExtensions) CreatePluginRegistry(_ context.Context) registry.
 // in the proxy deployment.
 func (e *k8sGatewayExtensions) GetEnvoyImage() Image {
 	return Image{
+		Registry:   "quay.io/solo-io",
```
sam-heilbron, line 85: nit: Feels like it could be a constant somewhere. Given that we have https://github.com/solo-io/gloo/issues/8906, I could see a world in which we may want to swap the default to GCR and having a centrally defined default would make this easier

projects/gateway2/deployer/merge.go
```diff
@@ -0,0 +1,616 @@
+package deployer
```
sam-heilbron, line 1: What is the performance impact of this deep merge logic? Should we have a basic benchmark test to demonstrate it?

sam-heilbron (commented):
Looking good!

install/helm/gloo/values-template.yaml
```diff
@@ -7,6 +7,30 @@ k8s:
 kubeGateway:
   # disable the integration by default, since it requires extra steps to set up (i.e. applying Gateway API CRDs)
   enabled: false
+  defaultGatewayParameters:
```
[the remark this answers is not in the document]
jbohanon: In flight from review meeting yesterday, changing to `map[string]GatewayParameters` keyed on a camelcase of the gateway class

projects/gateway2/deployer/deployer.go
jbohanon: comment is inaccurate, we'll fail if we don't find a default GatewayParameters, so there are no meaningful "default values"

projects/gateway2/deployer/values_helpers.go
jbohanon: PointerTo util is being removed in favor of the k8s util

jbohanon:
TODO: 
- helm test to validate that version.Version is propagated as default
- e2e tests

projects/gateway2/extensions/extensions.go
jbohanon: These default values have been removed from the codebase with the introduction of the default gwp, where it's inherited from the helm values

projects/gateway2/deployer/values_helpers.go
jbohanon: Good catch; removed

Makefile
```diff
@@ -572,10 +572,16 @@ discovery-distroless-docker: $(DISCOVERY_OUTPUT_DIR)/discovery-linux-$(GOARCH) $
 #----------------------------------------------------------------------------------
 
 GLOO_DIR=projects/gloo
+GATEWAY_DIR=projects/gateway
```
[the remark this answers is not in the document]
jbohanon: Yep, changed it!

projects/gateway2/deployer/merge.go
jbohanon: This is a good point. I don't have the performance impact quantified. Would you consider this blocking to this work merging?

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -29,6 +29,7 @@ runs:
     run: |
       helm upgrade -i -n gloo-system gloo ./_test/gloo-1.0.0-ci1.tgz --create-namespace \
         --set kubeGateway.enabled=true \
+        --set global.istioSDS.enabled=false \
```
npolshakova, line 32: Do we need this to be explicitly set to false?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -81,9 +81,9 @@ spec:
           failureThreshold: 3
         resources:
           {{- toYaml $gateway.resources | nindent 12 }}
-{{- if or $gateway.istioSDS.enabled $gateway.sds }}
+{{- if $gateway.istio.enabled }}
```
npolshakova, line 84: 🎉

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
jbohanon: Ah probably not. Was throwing spaghet at the wall trying to get conformance working; I think the issue was elsewhere. will revert

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
npolshakova: Eventually this will also be controlled by (or glooMtls.enabled istioSDS.enabled) but this is much cleaner for now!

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
+{{- if .Values.kubeGateway.enabled -}}
```
npolshakova, line 1: Let's remove https://github.com/solo-io/gloo/blob/9aabb7de642383053f6b9efae8d422e469b04582/install/helm/gloo/templates/1-gloo-deployment.yaml#L243 as well? I don't think it should be used anywhere anymore

projects/gateway2/controller/controller.go
```diff
@@ -139,7 +139,6 @@ func (c *controllerBuilder) watchGw(ctx context.Context) error {
 	d, err := deployer.NewDeployer(c.cfg.Mgr.GetClient(), &deployer.Inputs{
 		ControllerName: c.cfg.ControllerName,
 		Dev:            c.cfg.Dev,
-		IstioValues:    c.cfg.IstioValues,
```
npolshakova, line 142: I don't think we need `IstioValues` on the cfg anymore either

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: This is used in the plugin construction for istio_automtls as well as in the translator tests. If we need to remove it can we pair to get it done?

changelog/v1.17.0-beta32/default-gatewayparameters.yaml
```diff
@@ -0,0 +1,21 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6113
+    resolvesIssue: true
+    description: >-
+      Added default service type of LoadBalancer to default GatewayParameters. This value will be inherited by each deployed
```
jenshu, line 6: i think that we should also change the [ServiceType](https://github.com/solo-io/gloo/blob/46e504427d15757b63698cd164c2e566b2b43355/projects/gateway2/api/v1alpha1/kube/service.proto#L33) definition to have a different default value (e.g. Unspecified=0, ClusterIP=1, LoadBalancer=2) similar to [PullPolicy](https://github.com/solo-io/gloo/blob/46e504427d15757b63698cd164c2e566b2b43355/projects/gateway2/api/v1alpha1/kube/container.proto#L30) so we can differentiate between undefined (in which case a child gwp would inherit parent values) vs explicitly defined as ClusterIP

install/helm/gloo/generate/values.go
```diff
@@ -308,7 +308,22 @@ type Gloo struct {
 }
 
 type KubeGateway struct {
-	Enabled *bool `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller."`
+	Enabled           *bool                         `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller."`
+	GatewayParameters map[string]*GatewayParameters `json:"gatewayParameters,omitempty" desc:"Map of GatewayClasses to default GatewayParameters"`
```
jenshu, line 312: if this is a map, it renders in the helm docs like
```
kubeGateway.gatewayParameters.NAME.image.tag
```
which might indicate to a user that they can use arbitrary gatewayparams names in the map, and that we would iterate through them all (similar to gatewayProxies).

since we only support `glooGateway` for now, i think this should be a struct with an explicit field named `glooGateway`, instead of a map

install/helm/gloo/generate/values.go
```diff
@@ -308,7 +308,22 @@ type Gloo struct {
...
+}
+
+type GatewayParameters struct {
+	Image      *Image                 `json:"image,omitempty" desc:""`
+	Deployment *ProvisionedDeployment `json:"deployment,omitempty" desc:"Options specific to the deployment of the dynamically provisioned gateway proxy"`
+	Service    *ProvisionedService    `json:"service,omitempty" desc:"Options specific to the service of the dynamically provisioned gateway proxy"`
+}
+
+type ProvisionedDeployment struct {
+	Replicas *int64 `json:"replicas,omitempty" desc:"number of instances to deploy"`
```
jenshu, line 322:
```suggestion
	Replicas *int32 `json:"replicas,omitempty" desc:"number of instances to deploy"`
```

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+kind: GatewayParameters
+apiVersion: gateway.gloo.solo.io/v1alpha1
+metadata:
+  labels:
+    {{- include "gloo-gateway.constLabels" . | nindent 4 }}
+  name: default
```
jenshu, line 7: can we set this to the name of the GatewayClass?
```suggestion
  name: gloo-gateway
```

install/helm/gloo/templates/44-gatewayclass.yaml
```diff
@@ -7,4 +7,9 @@ metadata:
   name: gloo-gateway
 spec:
   controllerName: solo.io/gloo-gateway
-{{- end }}{{/* if .Values.kubeGateway.enabled */}}
\ No newline at end of file
+  parametersRef:
+    group: gateway.gloo.solo.io
+    kind: GatewayParameters
+    name: default
```
jenshu, line 13: also update name here

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: Yeah... this is messier than I thought. The plugin only uses this for the warning if there is no sds container and auto mtls is enabled: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/istio_automtls/plugin.go#L66 

We don't create the plugins per gateway, so if a user creates an overwrite GatewayParameters that disables Istio, the automtls config will still be translated, even though sds is not there. Maybe @lgadban has some ideas on how to change the plugin?

install/helm/gloo/values-template.yaml
```diff
@@ -7,6 +7,15 @@ k8s:
 kubeGateway:
   # disable the integration by default, since it requires extra steps to set up (i.e. applying Gateway API CRDs)
   enabled: false
+  gatewayParameters:
+    glooGateway:
+      image:
+        registry: quay.io/solo-io
```
jenshu, line 13: we don't need to redefine the image registry here since it's the same as the global image registry

install/helm/gloo/generate/values.go
```diff
@@ -308,7 +308,22 @@ type Gloo struct {
 }
 
 type KubeGateway struct {
-	Enabled *bool `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller."`
+	Enabled           *bool                         `json:"enabled,omitempty" desc:"Enable the Gloo Gateway Kubernetes Gateway API controller."`
+	GatewayParameters map[string]*GatewayParameters `json:"gatewayParameters,omitempty" desc:"Map of GatewayClasses to default GatewayParameters"`
+}
+
+type GatewayParameters struct {
+	Image      *Image                 `json:"image,omitempty" desc:""`
```
npolshakova, line 316: Can we add a description here? Is this the top level image override for the GatewayParameters?

projects/gateway2/controller/controller_suite_test.go
```diff
@@ -119,6 +124,32 @@ var _ = BeforeSuite(func() {
 		},
 		Spec: api.GatewayClassSpec{
 			ControllerName: api.GatewayController(gatewayControllerName),
+			ParametersRef: &api.ParametersReference{
+				Group:     "gateway.gloo.solo.io",
+				Kind:      "GatewayParameters",
```
jenshu, line 129: can use the constants from https://github.com/solo-io/gloo/blob/46e504427d15757b63698cd164c2e566b2b43355/projects/gateway2/pkg/api/gateway.gloo.solo.io/v1alpha1/types.go#L17

install/helm/gloo/generate/values.go
```diff
@@ -308,7 +308,22 @@ type Gloo struct {
...
+	Deployment *ProvisionedDeployment `json:"deployment,omitempty" desc:"Options specific to the deployment of the dynamically provisioned gateway proxy"`
+	Service    *ProvisionedService    `json:"service,omitempty" desc:"Options specific to the service of the dynamically provisioned gateway proxy"`
+}
+
+type ProvisionedDeployment struct {
+	Replicas *int64 `json:"replicas,omitempty" desc:"number of instances to deploy"`
+}
+
+type ProvisionedService struct {
+	Type *string `json:"type,omitempty" desc:"K8s service type"`
```
npolshakova, line 326: Is this supposed to match the K8s Service type? https://github.com/solo-io/gloo/blob/main/projects/gateway2/api/v1alpha1/gateway_parameters.proto#L47 

Or are we adding fields as needed here to the helm values, but if the user is defining a manual GatewayParameters resource they will have access to the full Service type? If we're not exposing the whole service here, let's change the description.

install/helm/gloo/generate/values.go
npolshakova, line 322: Let's add the default here?
```suggestion
	Replicas *int64 `json:"replicas,omitempty" desc:"The number of desired instances to deploy. Defaults to 1."`
```

jenshu: we shouldn't add defaults to the descriptions in values.go.
the helm docs generation gets the default values from values-template.yaml

install/helm/gloo/generate/values.go
jbohanon: I think it makes sense to add them as needed. I will update the comment to reflect such

install/helm/gloo/generate/values.go
```diff
@@ -308,7 +308,26 @@ type Gloo struct {
...
+	GatewayParameters *GatewayParametersForGatewayClasses `json:"gatewayParameters,omitempty" desc:"Maps GatewayClasses to default GatewayParameters"`
+}
+
+type GatewayParametersForGatewayClasses struct {
+	GlooGateway *GatewayParameters `json:"glooGateway" desc:"Default GatewayParameters for gloo-gateway GatewayClass."`
+}
+
+type GatewayParameters struct {
+	Image      *Image                 `json:"image,omitempty" desc:"Image options for the dynamically provisioned gateway proxy"`
+	Deployment *ProvisionedDeployment `json:"deployment,omitempty" desc:"Options specific to the deployment of the dynamically provisioned gateway proxy. Only a subset of all possible options is available. See \"ProvisionedDeployment\" for which are configurable via helm."`
```
npolshakova, line 321: Do we want this to match the GatewayParameters proto? (ProxyDeployment vs. Deployment)? I think having the generic `ProvisionedDeployment` struct makes sense.

jbohanon: yeah good call

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
+{{- if .Values.kubeGateway.enabled -}}
+kind: GatewayParameters
+apiVersion: gateway.gloo.solo.io/v1alpha1
+metadata:
+  labels:
+    {{- include "gloo-gateway.constLabels" . | nindent 4 }}
```
npolshakova, line 6: Should this be `glooGateway.constLabels`

jbohanon: I copied it from the `GatewayClass` template and it seems to render correctly

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+    service:
+      type: {{ $gg.service.type }}
+{{- end -}}{{/* if $gg.service */}}
+    envoyContainer:
+      image:
+{{- toYaml $image | nindent 8 }}
+    podTemplate:
+      extraLabels:
+{{- include "gloo-gateway.constLabels" . | nindent 8 }}
+    istioContainer:
```
npolshakova, line 28: Are we rendering the `istioContainer` field every time? Or only if `{{- if $global.glooMtls.istioProxy -}}` is set with an override?

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+      image:
+{{- merge $global.glooMtls.istioProxy.image $global.image | toYaml | nindent 8 }}
+{{- if $global.glooMtls.istioProxy -}}
```
npolshakova, line 31: I think this should go before the `istioContainer`

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+{{- if $global.glooMtls.istioProxy.securityContext }}
+      securityContext:
+{{- toYaml $global.glooMtls.istioProxy.securityContext | nindent 8 }}
+{{- end -}}{{/* if $global.glooMtls.istioProxy.securityContext */}}
+{{- if $global.glooMtls.istioProxy.logLevel }}
+      logLevel: {{ $global.glooMtls.istioProxy.logLevel }}
+{{- end -}}{{/* if $global.glooMtls.istioProxy.logLevel */}}
+{{- end -}}{{/* if $global.glooMtls.istioProxy */}}
+    sdsContainer:
```
npolshakova, line 40: Are we rendering the `sdsContainer` container everytime? Or only when `$gateway.istioSDS.enabled` is set?

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: And I think there should be a top-level check if the Istio integration is enabled (`$gateway.istioSDS.enabled`)

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+{{- end -}}{{/* if $global.glooMtls.sds.securityContext */}}
+{{- if $global.glooMtls.sds.logLevel }}
+      bootstrap:
+        logLevel: {{ $global.glooMtls.sds.logLevel }}
+{{- end -}}{{/* if $global.glooMtls.sds.logLevel */}}
+{{- if $global.glooMtls.sdsResources -}}
+      resources:
+{{- toYaml $global.glooMtls.sdsResources | nindent 8 }}
+{{- end -}}{{/* if $global.glooMtls.sdsResources */}}
+    istio:
```
npolshakova, line 55: Wait, why are `istio` and `istioContainer` separate? I think we discussed having sds top level, but all `istio` config should still be grouped together. Why would you want to set the Istio image if Istio is disabled?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -39,6 +39,14 @@ message KubernetesProxyConfig {
   // Configuration for the container running Envoy.
   EnvoyContainer envoy_container = 2;
 
+  // Configuration for the container running SDS. If an integration requiring SDS is not enabled, this
+  // is not rendered
+  SdsContainer sds_container = 7;
+
+  // Configuration for the container running istio-proxy. If the Istio integration is not enabled, this
+  // is not rendered
+  IstioContainer istio_container = 8;
```
npolshakova, line 48: Why is this separate from `IstioIntegration`? I don't think there is a case where you would want to set the `IstioContainer` but have `IstioIntegration` disabled. I think we discussed having `SdsContainer` as a top level field like `EnvoyContainer` because both Istio integration and Gloo Mtls depend on it. The Istio container is always dependant on Istio integration being enabled. Maybe I missed something from an earlier discussion? @lgadban

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: Always. I think that's one of the upsides of separating the container definition from the integration definition. In this way, a user can enable istio integration from gwp and the container definition is already set

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jbohanon: By separating and rendering the container definitions regardless, a user can flip istio integration on/off without having to go back to helm, by adding flag in their override gateway parameters. The istio and sds container definition are there for them, or they can override if they need

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: So rendered out it would something like this _even if sds is not present:_
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
```

I think that might be more confusing because it implies SDS should be present, but because Istio/glooMtls isn't enabled it won't be there.

jbohanon: The rendered gateway params that include the container definitions will be the default gwp, which a user should never be overriding anyway since helm changes will blow that out

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: But you should be able to do that without having a separate container definition and Istio integration definition. If you disable it, it will still cause the deployer to re-render the gateway. I still don't understand the downside of having them together. But currently it looks like this in my understand:

```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
    istioContainer: #why should this exist?
       image: my-default-image
    istio: 
       enabled: false 
```

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: This will be the rendered default:
```yaml
kind: GatewayParameters
metadata:
  annotations:
    meta.helm.sh/release-name: gloo
    meta.helm.sh/release-namespace: gloo-system
  creationTimestamp: "2024-05-23T13:45:33Z"
  generation: 1
  labels:
    app.kubernetes.io/managed-by: Helm
    gloo: kube-gateway
  name: gloo-gateway
  namespace: gloo-system
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer:
      image:
        pullPolicy: IfNotPresent
        registry: quay.io/solo-io
        repository: gloo-envoy-wrapper
        tag: 1.0.0-ci1
    istio:
      enabled: false
      istioDiscoveryAddress: istiod.istio-system.svc:15012
      istioMetaClusterId: Kubernetes
      istioMetaMeshId: cluster.local
    istioContainer:
      image:
        pullPolicy: IfNotPresent
        registry: docker.io/istio
        repository: proxyv2
        tag: 1.18.2
      logLevel: warning
    podTemplate:
      extraLabels:
        gloo: kube-gateway
    sdsContainer:
      bootstrap:
        logLevel: info
      image:
        pullPolicy: IfNotPresent
        registry: quay.io/solo-io
        repository: sds
        tag: 1.0.0-ci1
    service:
      type: LoadBalancer
```

This will be a user's override:
```yaml
kind: GatewayParameters
metadata:
  name: gw-params
  namespace: same-as-gateway
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
```

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: Versus: 

```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
    istioIntegration:  # this won't be rendered in the defaults if `istioSDS.enabled=true`
       enabled: true
       istioContainer:
           image: my-default-image
```

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: But they're not using SDS or Istio. If I was a user looking at the GatewayParameters for my installation, I think I would be confused why those fields were present if I wasn't enabling the integration.

projects/gateway2/deployer/deployer.go
```diff
@@ -112,6 +112,10 @@ func (d *Deployer) GetGvksToWatch(ctx context.Context) ([]schema.GroupVersionKin
 			"serviceAccount": map[string]any{
 				"create": true,
 			},
+			"istio": map[string]any{
```
npolshakova, line 115: Do we need this if we set https://github.com/solo-io/gloo/blob/63a4c8bb3419736527b2af2c3d3480afed990079/projects/gateway2/deployer/values_helpers.go#L146

changelog/v1.17.0-beta32/default-gatewayparameters.yaml
EItanya: Agreed, but there should still be a default if UNSET, likely `LOAD_BALANCER`

projects/gateway2/api/v1alpha1/gateway_parameters.proto
EItanya: > a user can flip istio integration on/off without having to go back to helm
Why would a user want to do that? In reality if a user is using helm, any change should be made using helm.

npolshakova: So this wouldn't work:
```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
    istioContainer: #why should this exist?
       image: my-default-image
    istio: 
       enabled: true
```

npolshakova: If the default was missing the Istio config, but then the user creates this override, Istio won't work?

```
kind: GatewayParameters
metadata:
  name: gw-params
spec:
  kube:
    deployment:
      replicas: 1
    envoyContainer: 
       image: my-default-image
    sdsContainer: 
       image: my-default-image
    istioContainer: #why should this exist?
       image: my-default-image
    istio: 
       enabled: true
```

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: I think this should be a follow up, I created an issue: https://github.com/solo-io/solo-projects/issues/6243

changelog/v1.17.0-beta32/default-gatewayparameters.yaml
jenshu: the default will differ per GatewayClass; for gloo-gateway it's LoadBalancer as defined in values-template.yaml

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: Discussed offline: the default GatewayParameters will be rendered with all the default values.

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: https://github.com/solo-io/gloo/pull/9493#discussion_r1612108536

projects/gateway2/api/v1alpha1/gateway_parameters.proto
npolshakova: https://github.com/solo-io/gloo/pull/9493#discussion_r1612108536

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: Thank you!!

install/helm/gloo/templates/43-gatewayparameters.yaml
lgadban: >We should add documentation stating that modifying the default GatewayParameters directly (i.e. not via helm) is NOT supported

Can you elaborate on what this means? Are we saying that there is a problem if the user were to e.g. `kubectl edit` the default `GatewayParams` object?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -39,6 +39,11 @@ message KubernetesProxyConfig {
   // Configuration for the container running Envoy.
   EnvoyContainer envoy_container = 2;
 
+  // Configuration for the container running SDS. This field is rendered with defaults by 
+  // helm in the default GatewayParameters for a GatewayClass, and may be overridden in 
+  // override GatewayParameters attached to a Gateway resource.
```
lgadban, line 44: Should we also add a note explaining that this container will not be rendered by the deployer unless an integration that requires it, e.g. IstioIntegration, is enabled?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -49,11 +54,16 @@ message KubernetesProxyConfig {
...
   kube.gateway.gloo.solo.io.Autoscaling autoscaling = 5;
 
-  // Sds configuration.
-  // A non-nil value will enable sds integration.  If empty config is provided, the default values will be used.
-  SdsIntegration sds = 6;
+  // Istio integration configuration.
+  IstioIntegration istio = 6;
+
+  // NOT IMPLEMENTED
+  GlooMtls gloo_mtls = 8;
```
lgadban, line 61: wondering if we should we even add this yet if it's not going to be implemented?

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -113,22 +123,16 @@ message EnvoyBootstrap {
...
-  // A non-nil value will enable sds integration. If empty config is provided, the default values will be used.
-  IstioIntegration istio_integration = 2;
-
-}
-
 // Configuration for the Istio integration settings used by a Gloo Gateway's data plane (Envoy proxy instance)
 message IstioIntegration {
-  // Configuration for the container running istio-proxy.
+  // If true, the Istio integration is enabled for this Gateway. Defaults to nil, which indicates the value should be inherited
+  // from the default GatewayParameters rendered from the helm chart. Set to true to override the default values with enabled; set to
```
lgadban, line 129: what happens if the default GatewayParams has this set to nil?

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: 1. Any changes will be blown out by helm
2. Since the intended pattern is for users to write per-Gateway GatewayParameters which then merge into and inherit from the defaults, if a user were to say delete the envoy container image stanza, there would be nothing to inherit and override gwps will be broken

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -0,0 +1,62 @@
...
+{{- if $global.glooMtls.istioProxy -}}
+{{- if $global.glooMtls.istioProxy.securityContext }}
+        securityContext:
+{{- toYaml $global.glooMtls.istioProxy.securityContext | nindent 10 }}
+{{- end -}}{{/* if $global.glooMtls.istioProxy.securityContext */}}
+{{- if $global.glooMtls.istioProxy.logLevel }}
+        logLevel: {{ $global.glooMtls.istioProxy.logLevel }}
+{{- end -}}{{/* if $global.glooMtls.istioProxy.logLevel */}}
+{{- end -}}{{/* if $global.glooMtls.istioProxy */}}
+      enabled: {{ or $global.istioSDS.enabled $global.glooMtls.enabled }}
```
lgadban, line 56: nit: this should be the first field in this object, i.e. above `istioContainer`

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jbohanon: This feels like an implementation detail to me, but I can add it

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jbohanon: Happy to remove it

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: I think it will be rearranged by kube anyway but I can move it!

projects/gateway2/api/v1alpha1/gateway_parameters.proto
lgadban: I think it's fine either way.
I just wonder if it would be confusing if users see the config in GatewayParams but it doesn't translate to a running container.

projects/gateway2/deployer/deployer.go
jbohanon: Yes because the helper functions aren't called on this codepath. `GetObjsToDeploy` is what calls `getValues` but this GetGVKs func calls render directly

projects/gateway2/api/v1alpha1/gateway_parameters.proto
jbohanon: We discussed the issue of user modifying default gateway parameters, and decided that in the near term it will be a good solution to open an issue tracking the work of adding validation to ensure that a default gatewayparameters cannot be modified to have missing default values. @lgadban

npolshakova (approved):

projects/gateway2/deployer/merge.go
jbohanon: Per @sam-heilbron offline, this should not block this PR.

npolshakova (approved):

projects/gateway2/deployer/values_helpers.go
```diff
@@ -97,249 +95,78 @@ func getServiceValues(svcConfig *v1alpha1kube.Service) *helmService {
...
+		Enabled:               ptr.To(istioConfig.GetEnabled().GetValue()),
+		IstioDiscoveryAddress: ptr.To(istioConfig.GetIstioDiscoveryAddress()),
+		IstioMetaMeshId:       ptr.To(istioConfig.GetIstioMetaMeshId()),
+		IstioMetaClusterId:    ptr.To(istioConfig.GetIstioMetaClusterId()),
 	}
 }
 
 // Get the image values for the envoy container in the proxy deployment. This is done by:
 // 1. getting the image values from a GatewayParameter
 // 2. for values not provided, fall back to the defaults (if any) from the k8s gw extensions
```
jenshu, line 162: nit: not sure if some of these comments are still accurate

projects/gateway2/deployer/deployer_test.go
```diff
@@ -299,15 +365,32 @@ var _ = Describe("Deployer", func() {
...
+				return &api.GatewayClass{
+					TypeMeta: metav1.TypeMeta{},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: wellknown.GatewayClassName,
+					},
+					Spec: api.GatewayClassSpec{
+						ControllerName: wellknown.GatewayControllerName,
+						ParametersRef: &api.ParametersReference{
+							Group:     "gateway.gloo.solo.io",
+							Kind:      "GatewayParameters",
```
jenshu, line 378: nit: constants

jenshu (approved):
LGTM, will you add e2e tests in a follow-up PR?

PR #9498, opened 2024-05-17
gateway2: simplify how plugins handle delegated routes

shashankram:
# Description
This change simplifies how plugins may perform merging of policies in a delegation chain, particularly in the case of RouteOptions. It performs an in-place merge such that the policy on a child route may be overridden by by a subsequent call to the plugin with a different route context.

Further, it accurately tracks the source RouteOptions involved in the merge so that the statuses on them can be correctly reported.

## Testing steps

Apply RouteOptions to routes similar to the configs in `projects/gateway2/translator/testutils/inputs/delegation/route_options*` and confirm the merging of policies works as expected.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6204

### DIFF
--- a/changelog/v1.17.0-beta29/deleg-plugin.yaml
+++ b/changelog/v1.17.0-beta29/deleg-plugin.yaml
@@ -1,0 +1,17 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6204
+    resolvesIssue: true
+    description: |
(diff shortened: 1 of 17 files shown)


EItanya (commented):
I don't have all of the context here, but the changes LGTM

ilackarms (approved):

EItanya (approved):

PR #9502, opened 2024-05-20
reconcile upstreams in gateway controller

ilackarms:
# Description

currently a change to a k8s upstream might not get reconciled because the gateway controller is not set to resync on an Upstream change

this PR fixes that


jenshu (approved):

ilackarms:
closing as this actually should not be needed; the gateway translator does not actually query upstreams during translation, it simply passes refs along to the inner `proxy` translator. this corresponds to the behavior we have seen (routes return a 500 when the upstream is removed)

however this means that we will not get a status update when an upstream is added or removed on an existing route. the proper way to address this is to ensure that the status reporter `HandleProxyReports` callback can be invoked multiple times for a given proxy, (e.g.: because an upstream is updated).

PR #9506, opened 2024-05-21
Glooctl get proxy fixes

jenshu:
# Description

Fixed bug with `glooctl get proxy` returning duplicate results.

## Problem:
After the edge and kube xds syncers were [combined](https://github.com/solo-io/gloo/pull/9310), we were registering the same proxy reader twice, causing duplicates of proxies to be returned from the proxy debug endpoint and glooctl.

## Summary of changes:
- Since the kube and edge translators share the same proxy client now, the proxy debug endpoint doesn't maintain a mapping of proxy "source" to proxy reader anymore. Instead, to query for a specific proxy type (kube or edge), callers must use a selector (e.g. `created_by: gloo-kube-gateway-api` or expression selector (e.g. `created_by in (gloo-kube-gateway-api, gloo-gateway)`).
    - Note: filtering via the CLI tool (`glooctl get proxy`), still uses the flags `--kube` and `--edge`
- Added expression selector support to the proxy debug endpoint to allow set-based queries (e.g. filtering by kube OR edge)
- Added e2e test suite for `glooctl get proxy`

# Checklist:


projects/gloo/cli/pkg/common/proxy.go
```diff
@@ -70,52 +65,55 @@ func computeProxyEndpointPort(settings *gloov1.Settings) (string, error) {
...
-	proxySource := "" // empty string is considered "all"
-	if proxy.EdgeGatewaySource && !proxy.K8sGatewaySource {
-		proxySource = gloodebug.EdgeGatewaySourceName
-	}
-	if !proxy.EdgeGatewaySource && proxy.K8sGatewaySource {
-		proxySource = gloodebug.K8sGatewaySourceName
+// getSelectorFromOpts returns either an equality-based selector (map[string]string) or set-based selector (string),
+// based on the filtering specified by the GetProxy options. If no filtering is specified, then both the returned
+// selectors will be empty.
+func getSelectorFromOpts(opts options.GetProxy) (map[string]string, string) {
```
davidjumani, line 97: nit: Since this is an either / or / none could there be a better way to return just one selector ?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -212,17 +217,12 @@ func (s *ProxySyncer) reconcileProxies(ctx context.Context, proxyList gloo_solo_
...
-				return false, nil
-			}
-			// otherwise always update
+			// always update
 			return true, nil
 		},
 		clients.ListOpts{
-			Ctx: ctx,
+			Ctx:      ctx,
+			Selector: kubeGatewayProxyLabels,
```
jenshu, line 225: don't think this was causing any issues, but cleaned it up to do the filtering via selector

projects/gloo/pkg/syncer/setup/start_func.go
```diff
@@ -61,13 +60,6 @@ func K8sGatewayControllerStartFunc(
 	statusClient resources.StatusClient,
 ) StartFunc {
 	return func(ctx context.Context, opts bootstrap.Opts, extensions Extensions) error {
-		if opts.ProxyDebugServer.Server != nil {
-			// If we have a debug server running, let's register the proxy client used by
-			// the k8s gateway translation. This will enable operators to query the debug endpoint
-			// and inspect the proxies that are stored in memory
-			opts.ProxyDebugServer.Server.RegisterProxyReader(debug.K8sGatewayTranslation, proxyClient)
```
jenshu, line 68: already registered in setup_syncer, don't need it here

test/kubernetes/e2e/features/glooctl/testdata/backend.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
jenshu: in `test/kubernetes/e2e/features/glooctl/get_proxy_suite.go`

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -125,6 +125,17 @@ func (c *Cli) DeleteFile(ctx context.Context, fileName string, extraArgs ...stri
 	return err
 }
 
+// DeleteFileWithRetries deletes the resources defined in a file, and returns an error if one occurred.
+// It will retry up to 8 times with backoff delay.
+func (c *Cli) DeleteFileWithRetries(ctx context.Context, fileName string, extraArgs ...string) error {
```
npolshakova, line 130: Why does the delete fail? Is there a way to print the error on failure?

test/gomega/matchers/object.go
```diff
@@ -0,0 +1,53 @@
...
+	// Name is the object name.
+	Name string
+
+	// Namespace is the object namespace.
+	Namespace string
+}
+
+// ObjectMatches returns a GomegaMatcher that checks whether an object matches
+// the specified fields (currently only name and namespace).
+func ObjectMatches(object ExpectedObject) types.GomegaMatcher {
```
npolshakova, line 21: So this matcher would pass if the objects have different specs but the same name/namespace? Is it just used to check there are no duplicates/missing objects?

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
```diff
@@ -0,0 +1,214 @@
...
+	s.Require().NoError(err, "can apply edge gateways manifest")
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, getEdgeGateway1(ns), getEdgeGateway2(ns))
+
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, kubeGatewaysManifestFile)
+	s.Require().NoError(err, "can apply kube gateways manifest")
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, kubeGateway1, kubeRoute1, kubeGateway2, kubeRoute2)
+
+	// TODO come up with a better way to do this
+	// wait for a bit for translation to run and for the proxies to get created
+	time.Sleep(10 * time.Second)
```
npolshakova, line 84: Can't we just use an `Eventually` here instead of a sleep?

jenshu: yeah but i didn't know what to put in the Eventually.
maybe could keep running one of the `glooctl get proxy` tests until i get 5 proxies back. but it would kind of be repeating one of the actual tests. that seems like the most viable option though unless you can think of something else that tells us translation is done and the proxies are available to query?

test/gomega/matchers/object.go
jenshu: yep it's only to check Name and Namespace for now

pkg/utils/kubeutils/kubectl/cli.go
jenshu: i think it takes awhile for gloo's input snapshot to get updated after deleting a gateway, so when we go to delete the virtualservices, it thinks there are still gateways in the input snapshot that are referencing the virtual services. that's why it takes multiple retries to successfully delete the virtualservices (there's also a comment about this in get_proxy_suite where this func is called)

jenshu: and yes it does currently print the error each time it fails (validation webhook error)

sam-heilbron: On second though, I'm thinking that we want these tools to be as reflective of user behaviors as possible. What if instead of building retry logic into this client, we add the retry logic on the test itself, since it's only under certain circumstances that we need to do it?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -64,6 +62,13 @@ func NewGatewayInputChannels() *GatewayInputChannels {
 	}
 }
 
+var (
+	// labels used to uniquely identify Proxies that are managed by the kube gateway controller
+	kubeGatewayProxyLabels = map[string]string{
```
sam-heilbron, line 67: I don't think this is safe to define here. https://github.com/solo-io/gloo/blob/main/projects/gateway2/translator/gateway_translator.go#L97 has a different definition and so these two will inevitably drift.

I assume this doesn't cause a regression because these labels are a subset of the generated labels so it continues to select resources correctly?

projects/gloo/pkg/debug/proxy_endpoint.go
```diff
@@ -24,116 +20,82 @@ type ProxyEndpointServer interface {
...
+	proxyList, err := p.getMany(ctx, req.GetNamespace(), req.GetSelector(), req.GetExpressionSelector())
 	return &debug.ProxyEndpointResponse{
 		Proxies: proxyList,
 	}, err
 }
 
-func (p *proxyEndpointServer) getOne(ctx context.Context, namespace, name, source string) (*v1.Proxy, error) {
-	proxyReaders, err := p.getProxyReadersForSource(source)
+func (p *proxyEndpointServer) getOne(ctx context.Context, namespace, name string) (*v1.Proxy, error) {
+	reader, err := p.mustProxyReader()
```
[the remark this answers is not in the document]
jenshu: yeah the helper was unnecessary. removed it 92f83470f70dad20fbe09767ab9e488052489c6d

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
sam-heilbron: I agree that we should lean on eventually instead of sleep. If we are testing that after performing some condition, that eventually another condition exists, I think it's the perfect case for an "eventually ... get proxy...returns expected value"

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
```diff
@@ -0,0 +1,214 @@
...
+
+func NewGetProxySuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
+	return &getProxySuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+	}
+}
+
+func (s *getProxySuite) TestGetProxy() {
+	s.T().Cleanup(func() {
```
sam-heilbron, line 42: Given that we do a fair bit of setup and cleanup, and then have different conditions that we are checking, it was a bit confusing to follow the thread of operations. I found https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/features/route_options/suite.go to be a very clear example of using the BeforeTest/AfterTest paradigm to configure resources, and have separate test cases just defined in their own function.

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -212,17 +217,12 @@ func (s *ProxySyncer) reconcileProxies(ctx context.Context, proxyList gloo_solo_
 		s.writeNamespace,
 		proxyList,
 		func(original, desired *gloo_solo_io.Proxy) (bool, error) {
-			// ignore proxies that do not have our owner label
```
sam-heilbron, line 215: Why is this safe to delete? I forget the context around why we wanted to ignore these proxies in the first place. Perhaps @npolshakova remembers?

sam-heilbron (commented):
Looking good! I haven't tested end-to-end yet (but thanks for adding the automated tests for it)

projects/gloo/cli/pkg/common/proxy.go
jenshu: simplified it here f95a72a5b2f4be3a3eca649e0d91368db60dee81

pkg/utils/kubeutils/kubectl/cli.go
jenshu: moved the retries into the test b90a21c206ffda777864c9d52f76473ef3f3793e

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: I think the selector in the client.ListOptions is doing the same thing using the `utils.ProxyTypeKey: utils.GatewayApiProxyValue` selector

npolshakova (approved):

test/gomega/matchers/object.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	"github.com/onsi/gomega/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ExpectedObject is a struct that represents the expected object.
+type ExpectedObject struct {
```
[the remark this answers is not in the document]
jenshu: ah i realized it hadn't been working for me b/c they can't be strucct pointers. updated in 6bd2558b575011491c945a52a6adbd73b47ef1fc and 2a08e1e6745e520048172a3a3f31feaa5bd3677b

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
```diff
@@ -0,0 +1,214 @@
...
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, kubeGateway1, kubeRoute1, kubeGateway2, kubeRoute2)
+
+		ns := s.testInstallation.Metadata.InstallNamespace
+		err = s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, edgeGatewaysManifestFile, "-n", ns)
+		s.NoError(err, "can delete edge gateways manifest")
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, getEdgeGateway1(ns), getEdgeGateway2(ns))
+
+		// we are calling delete with retries here because there is some delay between Gateways being deleted and
+		// gloo picking up the updates in its input snapshot, causing VS deletion to fail in the meantime (gloo
+		// thinks there are still Gateways referencing the VS)
```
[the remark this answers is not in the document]
jenshu: we already set allowWarnings to true in the test's [helm values](https://github.com/solo-io/gloo/blob/3b54e27aea1384f081f97eada3d616e1587c6abd/test/kubernetes/e2e/tests/manifests/k8s-gateway-test-helm.yaml#L22)

projects/gateway2/proxy_syncer/proxy_syncer.go
jenshu: they both use `utils.ProxyTypeKey: utils.GatewayApiProxyValue` to identify that it's a kube proxy.

the additional label `utils.GatewayNamespaceKey: gateway.GetNamespace()` applied by the translator is to identify the specific gateway's namespace and so it isn't used as a selector in terms of filtering by proxy type

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
jenshu: moved into setup/teardown 3334e3bc9f5e8421d94a5f15523bbd692225bec0

test/kubernetes/e2e/features/glooctl/get_proxy_suite.go
jenshu: wait for proxies 3334e3bc9f5e8421d94a5f15523bbd692225bec0

projects/gateway2/proxy_syncer/proxy_syncer.go
sam-heilbron: Agreeed. More calling out that having these two definitions in separate places seems fragile and it will be possible for them to drift without engineers realizing. Could we somehow link the two?

sam-heilbron (approved):
Nice work!

davidjumani (approved):

npolshakova (approved):

projects/gateway2/proxy_syncer/proxy_syncer.go
jenshu: yeah maybe the map could be defined in one file and imported into the other file, or maybe adding a comment on those lines about them needing to stay in sync would be sufficient (i'll do that in a follow up)

PR #9513, opened 2024-05-21
gateway2/delegation: enable optional parent matcher inheritance

shashankram:
# Description
In the current design, a child route must have route matchers that match the parent's to be considered in the delegation chain.

There's a request to enable matcher inheritance from the parent such that instead of child routes needing to specify paths that include the parent's path as prefix, they instead inherit the parent prefix, headers, query params, etc. The result is that the paths specified by child routes are relative to the parent paths. Headers and query params are merged from the parent into the child while giving preference to parent matchers in case of conflicts.

## Testing steps

Test route matcher inheritance by using the annotation `delegation.gateway.solo.io/inherit-parent-matcher: "true"` on child routes so that they can inherit the parent route matcher without needing to specify the fully qualified path (and other matchers).

# Checklist:


projects/gateway2/query/delegation.go
```diff
@@ -2,18 +2,29 @@ package query
...
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// inheritMatcherAnnotation is the annotation used on an child HTTPRoute that
+// participates in a delegation chain to indicate that child route should inherit
+// the route matcher from the parent route.
+const inheritMatcherAnnotation = "delegation.gateway.solo.io/inherit-parent-matcher"
```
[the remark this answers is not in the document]
shashankram: Sure. cc @kcbabo @DuncanDoyle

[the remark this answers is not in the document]
shashankram: Discussed with @DuncanDoyle @ilackarms @yuval-k . We are okay with this annotation and its allowed values.

projects/gateway2/query/delegation.go
```diff
@@ -156,5 +192,104 @@ func isDelegatedRouteMatch(
...
+		}
+	}
+	var result []gwv1.HTTPQueryParamMatch
+	for _, h := range merged {
+		result = append(result, h)
+	}
+	// Sort for deterministic ordering
+	slices.SortStableFunc(result, func(a, b gwv1.HTTPQueryParamMatch) int {
+		return strings.Compare(string(a.Name), string(b.Name))
+	})
```
EItanya, line 293: Can there be multiple query param matches with the same name?

projects/gateway2/query/delegation.go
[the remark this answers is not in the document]
EItanya: Wanted to flag @solo-io/solo-docs so that this behavior gets documented for users.

projects/gateway2/query/delegation.go
```diff
@@ -66,18 +78,42 @@ func (r *gatewayQueries) GetDelegatedRoutes(
...
+				// Non-inherited matcher delegation requires matching child matcher to parent matcher
+				// to delegate from the parent route to the child.
 				if ok := isDelegatedRouteMatch(parentMatch, parentRef, match, child.Namespace, child.Spec.ParentRefs); ok {
 					validMatches = append(validMatches, match)
 					matched = true
 				}
 			}
-			rule.Matches = validMatches
+
+			// Update the rule on the child instead of on a copy of the rule
```
EItanya, line 115: You got piked?

projects/gateway2/query/delegation.go
```diff
@@ -156,5 +192,104 @@ func isDelegatedRouteMatch(
...
+		}
+	}
+	var result []gwv1.HTTPHeaderMatch
+	for _, h := range merged {
+		result = append(result, h)
+	}
+	// Sort for deterministic ordering
+	slices.SortStableFunc(result, func(a, b gwv1.HTTPHeaderMatch) int {
+		return strings.Compare(string(a.Name), string(b.Name))
+	})
```
EItanya, line 268: Same comment about multiple with the same name

EItanya (commented):
LGTM just nits

projects/gateway2/query/delegation.go
shashankram: I thought it could but it cannot due to kubebuilder constraints:
```
	// +listType=map
	// +listMapKey=name
```

projects/gateway2/query/delegation.go
shashankram: I had overlooked that the loop variable `rule` is a copy and not a pointer. Would have been "piked" had it been a pointer.
Is "piked" an official term? I've only seen it in code added by you and Lawrence 😆

EItanya (dismissed):
Comments addressed

projects/gateway2/query/delegation.go
EItanya: You got sort-of piked, you set to a copy, which is almost as bad 😆. Piked is as official as we believe it to be

EItanya (approved):

ilackarms (approved):

Issue #9508, opened 2024-05-21
Gloo Gateway: Deletion of HTTPRoute does not update proxy


lgadban:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.17-beta

### Kubernetes Version

n/a

### Describe the bug

When deleting an HTTPRoute, the proxies are not updated to remove the deleted route

### Expected Behavior

All routes in the deleted HTTPRoute should be removed from the proxy

### Steps to reproduce the bug

1. delete HTTPRoute
2. make a request to the route

PR #9515, opened 2024-05-22
Fix HTTPRoute deletion

lgadban:
# Description

Fixes a bug that prevented deleted `HTTPRoutes` from being fully reconciled to proxies.

Previously, when `HTTPRoutes` were reconciled (regardless of the event that triggered it), the reconcile function did a GET on that route, presumably to allow for intelligent reconciling.

This logic is broken for DELETEs as the GET will never succeed because the route no longer exists.
So the reconcile was short-circuiting on all deletes, preventing the deleted HTTPRoutes from being removed in the proxy until another update of some sort happened.

This fix simply removes the unnecessary GET on `HTTPRoute` reconciliation, allowing deletes to work as any other reconciliation.

## Interesting decisions
 
Behavior for a proxy with no routes seems a bit clumsy and should be investigated in a follow-up issue.

## Testing steps

```
kubectl apply -f projects/gateway2/examples/example-gw.yml
kubectl apply -f projects/gateway2/examples/example-http-route.yaml
kubectl delete httproutes.gateway.networking.k8s.io example-route
```

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9508


Makefile
```diff
@@ -169,7 +169,7 @@ install-go-tools: mod-download ## Download and install Go dependencies
 	# This version must stay in sync with the version used in CI: .github/workflows/static-analysis.yaml
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(LINTER_VERSION)
 	go install github.com/quasilyte/go-ruleguard/cmd/ruleguard@v0.3.16
-
+	go install github.com/onsi/ginkgo/v2/ginkgo
```
jenshu, line 172: this is already in `install-test-tools`

projects/gateway2/controller/controller.go
```diff
@@ -306,13 +306,7 @@ func (r *controllerReconciler) ReconcileNamespaces(ctx context.Context, req ctrl
 }
 
 func (r *controllerReconciler) ReconcileHttpRoutes(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	// find impacted gateways and queue them
-	hr := apiv1.HTTPRoute{}
-	err := r.cli.Get(ctx, req.NamespacedName, &hr)
```
jenshu, line 311: if we wanted to still do the Get, we could check the error type before returning, and do the kick as long as IgnoreNotFound(err)==nil

Makefile
lgadban: thank you!

projects/gateway2/controller/controller.go
lgadban: we could, but there's no point right now as it's not being used.

AFAICT, the original intent was to GET and calculate the Gateways necessary to reconcile.
I am hesitant to try and add that logic in this PR; there is a larger conversation to be had about ownership/dependency tracking.

projects/gateway2/controller/controller.go
```diff
@@ -306,13 +306,7 @@ func (r *controllerReconciler) ReconcileNamespaces(ctx context.Context, req ctrl
...
-	if err != nil {
```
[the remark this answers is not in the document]
lgadban: this is a good call out; the GWClass reconciliation does look a bit different as the retrieved object is used later in the flow, notably for updating status on it.

You are right the same behavior would apply for GWClass deletion -- although I don't know if the semantics of what deleting a GWClass means are well understood.

I'm tempted to not handle it in this PR, but I will add a comment there for posterity.
What do you think?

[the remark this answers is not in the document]
lgadban: On second thought, I think the existing logic is correct, at least for now.
Comment added here: https://github.com/solo-io/gloo/pull/9515/files#diff-2fcaa4047a46112fad4487b4113aa9eb31fd21fe1622746c3bf80ba97e3ff7caR332-R335

sam-heilbron (approved):

sam-heilbron:
bulldozer pls

Issue #6494, opened 2022-05-23
Gloo Administration Interface


sam-heilbron:
### Version

1.12.x (beta)

### Is your feature request related to a problem? Please describe.

Debugging the control plane can be challenging, since there is state that is not easily inspected.

#### Background
Envoy supports an administration interface (https://www.envoyproxy.io/docs/envoy/latest/operations/admin#operations-admin-interface) that is useful for querying various states of the server. This is especially useful for developers who are debugging various scenarios and want to inspect various elements that are not exposed by default. 

Debugging the Control Plane would be significantly streamlined, if Gloo exposed a similar interface.

#### Existing Solution
At the moment, we have solved these requests with one-off solutions:
- We support a DEBUG API th expose `/xds` and `/api`endpoints: https://github.com/solo-io/gloo/blob/bc380b36d42fdad7c83ab8dc4f055258b326aeac/projects/gloo/pkg/syncer/envoy_translator_syncer.go#L181
- We support a Proxy endpoint to enable viewing the state of the Proxy, even if it is stored in memory (as opposed to etcd): https://github.com/solo-io/gloo/blob/0eec04dc0486976fc89bac314b0fd9eccd5261f5/projects/gloo/api/v1/settings.proto#L586

#### Challenges with Existing Solution
Both of these features are extremely valuable and demonstrate that users need the ability to inspect state in the Control Plane, either for development or triaging an ongoing issue. However, since they are not aggregated under a single admin interface, there are two challenges:
1. It is difficult to understand what debugging capabilities the control plane offers without inspecting all areas of the code
2. Expanding this functionality is difficult because there is not a single pattern to follow

#### Proposed Solution
*I think this work can be done in 3 phases:*
1. Understand what administrative operations (queries or mutations) would be valuable (est 2 days)
2. Consolidate the existing operations under a single administrative interface, updating all internal and external documentation around how it should be used (est 4 days)
3. Extend the interface to support some new operations that are identified as part of Step 1 (est 2 days)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1207635280052738/1207993921169832) by [Unito](https://www.unito.io)

sam-heilbron:
In the attached PR, you can see that I begin this work. After that merges, the following work will be necesary:
- Flush out the input/output endpoints for inspecting the snapshot
- Update the user-facing documentation
- Mark old endpoints for deprecation in 1.17
- Remove old endpoints in 1.18

[...]

PR #9517, opened 2024-05-23
gloo/setup: expand stats server to support snapshot api

sam-heilbron:
# Description

Expose new endpoints on our stats server

# Context
## Standard Admin Interface
Per https://github.com/solo-io/gloo/issues/6494 we have somewhat of a sprawled out admin interface. There are multiple servers running to handle debugging requests from Gloo. In this PR, I attempt to standardize this a bit by demonstrating how we can extend the stats server (already used for debugging) to add new endpoints.

## Easy Introspection of Snapshots
Per https://github.com/solo-io/solo-projects/issues/6130, it would be ideal if we had an easy mechanism to capture the state of the running system.

This work is the first step to achieve this.

After this merges, I intend to have @djannot play with it, provide feedback and then I will iterate on it. This is intentionally not feature complete as it is more important to get feedback.

## Interesting decisions
 -

## Testing steps
Follow the steps outlined in https://docs.solo.io/gloo-edge/latest/operations/debugging_gloo/#changing-logging-levels-and-more to open a port-forward to the stats server.

Invoke requests to:
- [x] `/snapshots/input`
- [x] `/xds`

## Notes for reviewers
There are user-facing documents that I will update, but I plan on doing that after this PR merges, and I incorporate any feedback from the field.

# Checklist:


projects/gloo/cmd/main.go
```diff
@@ -6,13 +6,14 @@ import (
...
 	"github.com/solo-io/go-utils/log"
-	"github.com/solo-io/go-utils/stats"
 )
 
 func main() {
 	ctx := context.Background()
+
+	// Start a server which is responsible for responding to liveness probes
 	probes.StartLivenessProbeServer(ctx)
-	stats.ConditionallyStartStatsServer()
```
jbohanon, line 15: what else was served by this server? I see for example `/logging` in the code; do we lose the ability to hit `:9091/logging?level=debug`? I am testing this now

jbohanon (commented):
Looks awesome! Super excited for this

projects/gloo/cmd/main.go
jbohanon: Yeah I'm worried that we're losing all of this https://docs.solo.io/gloo-edge/latest/operations/debugging_gloo/#changing-logging-levels-and-more

jbohanon: never mind, in manual testing it still seems to work though I'm not sure why

sam-heilbron: I can clarify in a comment how this works, because i agree it's not immediately clear

jbohanon: We call `StartCancellableStatsServerWithPort` in projects/gloo/pkg/syncer/setup/start_func.go :man_facepalming:

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,112 @@
...
+func NewHistory() History {
+	return &history{
+		latestInput: map[string]json.Marshaler{},
+		xdsCache:    nil,
+	}
+}
+
+type history struct {
+	sync.RWMutex
+	latestInput map[string]json.Marshaler
```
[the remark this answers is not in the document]
sam-heilbron: I was attempting to mirror https://github.com/solo-io/skv2/blob/main/pkg/stats/io_snapshots.go#L43 which is used for GME. I figured some consistency in nomenclature would make it easier for devs/field to communicate. I do agree with your point though, and found the terminology a little confusing. While there is no concept of "history" at this point, I could see some instances in which this component is updated over the course of translation and can keep some historical measurements (though ideally anything that needs to be persisted would be exported to some telemetry consumer). 

tldr - I don't have strong feelings and am happy to rename if there's a name you think is more apt

go.mod
```diff
@@ -97,6 +97,7 @@ require (
 	golang.org/x/exp v0.0.0-20231006140011-7918f672742d
 	google.golang.org/genproto/googleapis/api v0.0.0-20230822172742-b8732ec3820d
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230822172742-b8732ec3820d
+	gopkg.in/yaml.v2 v2.4.0
```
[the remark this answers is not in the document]
sam-heilbron: Yeah good point. tbh, at this point this dependency is not even used. I am going to remove the invocation and save this for later if/when we need to add yaml support.

projects/gloo/cmd/main.go
sam-heilbron: I tried to clarify this where the server is started now: https://github.com/solo-io/gloo/blob/sam/input-output-snapshot-api/projects/gloo/pkg/syncer/setup/start_func.go#L95. Let me know if that's useful or if there are other places you think the documenation could exist

projects/gloo/pkg/servers/admin/server.go
```diff
@@ -0,0 +1,55 @@
...
+	"net/http"
+
+	"github.com/solo-io/gloo/projects/gloo/pkg/servers/iosnapshot"
+)
+
+// ServerHandlers returns the custom handlers for the Admin Server, which will be bound to the http.ServeMux
+// These endpoints serve as the basis for an Admin Interface for the Control Plane (https://github.com/solo-io/gloo/issues/6494)
+func ServerHandlers(history iosnapshot.History) func(mux *http.ServeMux, profiles map[string]string) {
+	return func(m *http.ServeMux, profiles map[string]string) {
+		m.HandleFunc("/snapshots/input", func(w http.ResponseWriter, request *http.Request) {
```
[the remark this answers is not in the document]
jbohanon: but... I guess we'd have to descend into api-snapshot which isn't really available to us since the iface takes `json.Marshaler` 

the wire size isn't helped, but post processing can always be done in code or with e.g. `jq`

projects/gloo/pkg/syncer/setup/start_func.go
```diff
@@ -87,3 +91,25 @@ func K8sGatewayControllerStartFunc(
...
+	// serverHandlers defines the custom handlers that the Admin Server will support
+	serverHandlers := admin.ServerHandlers(history)
+
+	return func(ctx context.Context, opts bootstrap.Opts, extensions Extensions) error {
+		// The Stats Server is used as the running server for our admin endpoints
+		//
+		// NOTE: There is a slight difference in how we run this server -vs- how we used to run it
+		// In the past, we would start the server once, at the beginning of the running container
+		// Now, we start a new server each time we invoke a StartFunc.
+		stats.StartCancellableStatsServerWithPort(ctx, stats.DefaultStartupOptions(), serverHandlers)
```
nfuden, line 111: Are there any current debug cases where a long running connection would be needed while a start function update occurs.

Such as a case where we try to capture a profile?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -908,12 +911,22 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
...
+
 	translationSync := syncer.NewTranslatorSyncer(
 		watchOpts.Ctx,
 		sharedTranslator,
 		opts.ControlPlane.SnapshotCache,
 		xdsSanitizers,
 		rpt,
+		// opts.DevMode should be deprecated
+		// https://github.com/solo-io/gloo/issues/6494
+		// We are starting to build out a true Admin Server, and enhancements should be added to that server
```
nfuden, line 929: Whats the scope of responsibility between Admin server and our UI?

What security implications does this have if we are always exposing these endpoints?

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -81,7 +81,9 @@ func (s *translatorSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapsh
 	ctx, span := trace.StartSpan(ctx, "gloo.syncer.Sync")
 	defer span.End()
 
+	s.snapshotHistory.SetApiSnapshot(snap)
```
nfuden, line 84: This should be a goroutine. 
Admin interface should never starve the main translation via lock contention.

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,112 @@
...
+func (h *history) GetInput() ([]byte, error) {
+	input, err := h.GetInputCopy()
+	if err != nil {
+		return nil, err
+	}
+
+	return formatMap("json_compact", input)
+}
+
+// GetInputCopy gets an in-memory copy of the output snapshot for all components.
```
nfuden, line 61:
```suggestion
// GetInputCopy gets an in-memory copy of the output snapshot for all components.
// Note that this may contain sensitive data and secrets.
```

projects/gloo/pkg/syncer/setup/start_func.go
sam-heilbron: I don't think so. I try to call out the subtle difference between this implementation and our previous one, but when the setup loop runs, we re-initialize everything, so I think we lose any active state anyway. As it stands, I don't think that users do/should modify the Settings resource directly, and instead would perform a rollout of resources which would restartt all containers anyway

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,132 @@
...
+		} else {
+			cacheEntries[k] = xdsSnapshot
+		}
+	}
+
+	return formatMap("json_compact", cacheEntries)
+}
+
+// apiSnapshotJsonMarshaller is a temporary solution to provide a MarshalJSON interface to the History interface
+// Help Wanted: It would be preferable to support a MarshalJSON directly on the ApiSnapshot type
```
jbohanon, line 123: PRs that could fix/obviate this shim
https://github.com/solo-io/gloo/pull/9530
https://github.com/solo-io/solo-kit/pull/554

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/solo-io/solo-kit/pkg/api/v1/control-plane/cache"
+)
+
+// History represents an object that maintains state about the running system
+// The ControlPlane will use the Setters to update the last known state,
+// and the Getters will be used by the Admin Server
+type History interface {
+	// SetApiSnapshot sets the latest input ApiSnapshot
+	SetApiSnapshot(latestInput *v1snap.ApiSnapshot)
+	// GetInputCopy gets an in-memory copy of the output snapshot for all components.
```
lgadban, line 17:
```suggestion
	// GetInputCopy gets an in-memory copy of the input snapshot for all components.
```

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,132 @@
...
+	// Note that this may contain sensitive data and secrets.
+	GetInputCopy() (map[string]interface{}, error)
+	// GetInput gets the input snapshot for all components.
+	GetInput() ([]byte, error)
```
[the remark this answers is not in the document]
lgadban: also would maybe impact `GetInputCopy`

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: I ended up moving in a direction that sidesteps this for now, and I removed the shim

sam-heilbron:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=3b6efbf7-87f9-4462-83bd-b2b25a66e616

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,159 @@
...
+// The ControlPlane will use the Setters to update the last known state,
+// and the Getters will be used by the Admin Server
+type History interface {
+	// SetApiSnapshot sets the latest ApiSnapshot
+	SetApiSnapshot(latestInput *v1snap.ApiSnapshot)
+	// GetRedactedApiSnapshot gets an in-memory copy of the ApiSnapshot
+	// Any sensitive data contained in the Snapshot will either be explicitly redacted
+	// or entirely excluded
+	GetRedactedApiSnapshot() *v1snap.ApiSnapshot
+	// GetInputSnapshot gets the input snapshot for all components.
```
lgadban, line 20: should we call out here that GetInputSnapshot will be redacted (as it called redactedApiSnapshot under the covers)?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,159 @@
...
+
+// getApiSnapshotSafe gets a clone of the latest ApiSnapshot
+func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
+	h.RLock()
+	defer h.RUnlock()
+	if h.latestApiSnapshot == nil {
+		return &v1snap.ApiSnapshot{}
+	}
+
+	// This clone is critical!!
```
lgadban, line 138: This scares me a bit; seems like the abstraction is off.

Seems like the `iosnapshot` package is intended to be used ONLY for debugging tooling? I.e. in the syncers, we will set snapshots as they are updated, but this history will never be used for any "real" control plane functionality?

If that's the case, and since we are setting in a goroutine (so no blocking the caller), should we clone on the way in, via the setter?

sam-heilbron: heh I actually had done that originally, that way we never stored a reference to the original object. But then I noticed that each request to the history would also mutate the object (to redact information, or extract pieces of the snapshot). So I found myself cloning on the way in and out. I opted to just do it once here, but I'm happy to add back the original code to also do it on ingestion. wdyt?

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Do you mean "call out" as in a comment on the function or by changing the name?

lgadban: Was thinking the comment, but now not so sure :)
Whatever you think!

projects/gloo/pkg/servers/iosnapshot/history.go
lgadban: Ahh that's a good point. I'm fine leaving as is in that case.

lgadban (dismissed):
LGTM!

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -0,0 +1,159 @@
...
+	// We do this to ensure the following cases:
+	//	1. Modifications to this snapshot, by the admin server, DO NOT impact the Control Plane
+	//	2. Modifications to this snapshot by a single request, DO NOT interfere with other requests
+	clone := h.latestApiSnapshot.Clone()
+	return &clone
+}
+
+// redactApiSnapshot accepts an ApiSnapshot, and returns a cloned representation of that Snapshot,
```
jbohanon, line 146: this comment is no longer accurate; the snapshot that is passed in _is_ mutated

sam-heilbron: updated

lgadban (approved):

nfuden (approved):

jbohanon (approved):

sam-heilbron:
An e2e test failed on this PR, and will be solved by: https://github.com/solo-io/gloo/pull/9554

sam-heilbron:
/kick opened a convo in team channel: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1717096987055319

sam-heilbron:
bulldozer pls

PR #9511, opened 2024-05-21
chore: Bump k8s to 1.29

davidjumani:
# Description

Bump the following to support k8s v1.29 :
- k8s
- k8s-utils
- skv2
- solo-kit
- solo-apis

## CI changes
- CI now runs on a kind k8s 1.29.2 cluster
- Ingress will be deprecated on 1.17+. Ref: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1716389614777799
- Add support for on-demand regression and performance tests. Copied over from [Enterprise](https://github.com/solo-io/solo-projects/pull/5869)

# Context
https://github.com/solo-io/gloo/issues/8978

## Testing steps
CI passing
Nightly tests [passing](https://github.com/solo-io/gloo/actions/runs/9210119536/job/25336344692) on 1.25 and 1.29

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #9526, opened 2024-05-27
.github/workflows: introduce nightly-kubernetes-tests, fix slack notifications

sam-heilbron:
# Description

Update our GitHub actions to run the Kubernetes e2e tests during our scheduled (nightly) cadence

# Context
## Scheduled Tests
It is important that we execute our kubernetes e2e tests on a scheduled cadence.  This allows us to have higher confidence in the tests, identify flakes proactively, and find ways in the future to not run tests on every PR, for features that rarely change.

We have added these scheduled tests for our regression tests, and it is important for this new kubernetes e2e test

## Slack Notification
Previously, our Slack notifications would work by emitting artifacts on failure for each workflow, and then the "publish results" workflow would zip those files together. If there were 0 files present, it assumed that meant that no workflows had failed. However, this was an issue because if a workflow failed earlier, or failed in a way that exited early, the job would fail and the artifact would never get created. Also, this meant that this only worked if workflows emitted this artifact on failure. https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1716789197999039 is an example of when this was not done.

My approach was to pass the `workflow.needs` object, and then let the publish workflow make the decision of how to report. This makes the Slack notification less informative (we no longer list the jobs that failed), but I would argue that the accuracy of success/failure is more important than the message. As it is, on a failure, a developer must step into the logs to triage the failure.

## Remove Cluster-Id
Cluster ID was a unique identifier for a Kubernetes cluster. It was responsible for having a unique location, per cluster, where artifacts could be stored for that cluster. This ID is an implementation detail of running multiple clusters in CI, and not really relevant to developers. Plus, it caused confusion around why we had a cluster name and a cluster ID.

My solution was to remove the concept of Cluster ID from the code altogether, and create a requirement that cluster name is unique and thus can be the source of truth for any artifacts that are unique to that cluster. 

In our GitHub workflow, we still need to assign an ID to each cluster in the matrix. This is used to create branch protection rules, but we prevent propagating that to the code by using the ID to create a unique cluster name.

## Istio Version in Tests
At the moment we hard code the istio version used in our tests.

When I forgot to add this to the nightly tests, I saw the following ([logs](https://github.com/solo-io/gloo/actions/runs/9273730628/job/25514460258)):
```
=== RUN   TestAutomtlsIstioEdgeApisGateway
"Tue, 28 May 2024 17:32:13 UTC: /home/runner/work/gloo/gloo/test/kube2e/helper/install.go:294"	found Helm index file at: "/home/runner/work/gloo/gloo/_test/index.yaml"
"Tue, 28 May 2024 17:32:13 UTC: /home/runner/work/gloo/gloo/test/kube2e/helper/install.go:303"	version of ["gloo"] Helm chart is: "1.0.0-ci1"
--- FAIL: TestAutomtlsIstioEdgeApisGateway (0.02s)
panic: ISTIO_VERSION environment variable must be specified to run [recovered]
	panic: ISTIO_VERSION environment variable must be specified to run
```

For now, I follow the same pattern used in our PR worfklow, to hardcode the istio version used. @sheidkamp will follow up and convert this single value into a matrix with upper/lower bounds

## Interesting decisions
I opted to re-use our existing workflow for nightly tests, instead of building a new one

## Testing steps
- [x] I ran the workflow using a manual trigger (https://github.com/solo-io/gloo/actions/runs/9273520408/job/25513799238) and confirmed that it initiates tests agains the min and max versions
- [x] The message on failure includes a mention of the job that failed: https://solo-io-corp.slack.com/archives/C0314KESVNV/p1716915293334769
- [x] The message on success is accurate: https://solo-io-corp.slack.com/archives/C0314KESVNV/p1716916638718969
- [x] When triggering the workflow without specifying kubernetes run regex, the default is to run all tests (https://github.com/solo-io/gloo/actions/runs/9274446944/job/25516816832)

## Notes for reviewers
- After this merges, I will monitor the next scheduled run to ensure that the tests are running and notifying properly

# Checklist:


.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -7,6 +7,9 @@ inputs:
   cluster-name:
     required: true
     description: The name of the KinD cluster
+  cluster-id:
```
sam-heilbron, line 10: Without this, you would see warnings like: 
```
Warning: Unexpected input(s) 'cluster-id', valid inputs are ['cluster-name', 'test-args', 'run-regex']
```

.github/workflows/composite-actions/istio-tests/action.yaml
```diff
@@ -37,11 +37,6 @@ runs:
           export ISTIO_AUTO_MTLS=true
         fi
         make install-test-tools run-kube-e2e-tests
-    - name: save results
```
npolshakova, line 40: Are we moving this somewhere or just deleting it?

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
npolshakova: Jenny left a good point here: https://github.com/solo-io/solo-projects/pull/6212#discussion_r1611713253

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,74 @@ on:
       run-regression:
         description: "Run regression tests"
         type: boolean
+        default: false
       run-performance:
         description: "Run performance tests"
         type: boolean
+        default: false
+      run-kubernetes-end-to-end:
+        # /tests/kubernetes/e2e/...
```
npolshakova, line 35: [nit] Maybe add a comment `Runs all tests in /tests/kubernetes/e2e/...`

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,74 @@ on:
...
+        description: "Run Kubernetes e2e tests"
+        type: boolean
+        default: false
+      kubernetes-end-to-end-run-regex:
+        description: "Kubernetes e2e tests -run regex"
+        type: string
+        required: false
+        default: '^Test'
```
npolshakova, line 43: Did this work instead of `^Test$$'?

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,74 @@ on:
...
+    strategy:
+      # Since we are running these on a schedule, there is no value in failing fast
+      # In fact, we want to ensure that all tests run, so that we have a clearer picture of which tests are prone to flaking
+      fail-fast: false
+      matrix:
+        # When running the tests at night, there is no value in splitting the tests across multiple clusters and running them in parallel.
+        # As a result, we increase the threshold for the tests, since they all run serially on a single cluster
+        test:
+          - cluster-id: 'cluster_one'
+            cluster-node-version: 'v1.28.0@sha256:b7a4cad12c197af3ba43202d3efe03246b3f0793f162afb40a33c923952d5b31'
```
npolshakova, line 60: Can we add a min/max version file to the nightlys? https://github.com/solo-io/solo-projects/tree/main/.github/workflows/.env/nightly-tests

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,74 @@ on:
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: main
+      - name: Prep Go Runner
+        uses: ./.github/workflows/composite-actions/prep-go-runner
+      - id: setup-kind-cluster
+        name: Setup KinD Cluster
+        uses: ./.github/workflows/composite-actions/setup-kind-cluster
```
npolshakova, line 75: Do we want the setup cluster to run as part of `.github/workflows/nightly-tests.yaml` or as part of the composite action (`./.github/workflows/composite-actions/kubernetes-e2e-tests`)? Other e2e tests like the regression tests handle the (kind cluster setup)[https://github.com/solo-io/gloo/blob/866015c574c1b6f3fccce8027fe508d2d0fce51d/.github/workflows/composite-actions/regression-tests/action.yaml#L22] which makes it easier to reuse in both the nightly tests and in the workflows that run for PR merging into main.

.github/workflows/nightly-tests.yaml
```diff
@@ -187,13 +251,10 @@ jobs:
...
         env:
           PARENT_JOB_URL: https://github.com/solo-io/gloo/actions/runs/${{github.run_id}} # parent job hyperlink
           PREAMBLE: ${{ steps.compute-preamble.outputs.preamble }}  # text to hyperlink at start of slack message
           SLACKBOT_BEARER: ${{ secrets.SLACKBOT_BEARER }}
         run: |
-          test_results="$(cat */test-out.json | jq -c --slurp .)"
-          echo $test_results
-          go run .github/workflows/helpers/notify-from-json.go $test_results
\ No newline at end of file
+          go run .github/workflows/helpers/notify/slack.go '${{ toJSON(needs) }}'
```
nfuden, line 260: This seems like throwing out the baby with the bath water. 

Previously we had talked about using [github.event.workflow_run.conclusion ](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow) to augment these results.
Could we perhaps do something that still gives us sub failed jobs but also contains info as to whether any of our jobs in general failed? 

If there are any failures then its sufficient to triage and its a bit easier to have context of what failed directly in slack. It might still be leaky as to how much broke but it seems

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
sam-heilbron: I like it! As I think about it more, cluster-name and cluster-id are very similar. cluster-id is just our logical grouping and is some unique ID that we can run for a matrix. cluster-name is the name of the kind cluster, and we just hardcode it to `kind`. What if we just made `cluster-name` unique in that in CI we use `kind-cluster_1` and `kind-cluster_2`? Then we can remove the cluster-id concept altogether since it's baked into the name

.github/workflows/composite-actions/istio-tests/action.yaml
sam-heilbron: As part of this I just removed this. Tests would save results, and then another action would consume these results. In my changes, I removed this handshake because it was fragile (see: https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1716876298887489 as an example. we had 4 failed tests, but none were reported).

.github/workflows/nightly-tests.yaml
sam-heilbron: I like it!

.github/workflows/nightly-tests.yaml
sam-heilbron: My goal was to run all tests by default, so I should certainly confirm this. My fear with `^Test$$` was that it would only select tests with that name (ie starting and ending in TEst)

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,74 @@ on:
...
+      kubernetes-end-to-end-run-regex:
+        description: "Kubernetes e2e tests -run regex"
+        type: string
+        required: false
+        default: '^Test'
 
 jobs:
+  end_to_end_tests_main:
+    name: End-to-End (branch=main, cluster=${{ matrix.test.cluster-id }})
+    if: ${{ (github.event_name == 'workflow_dispatch' && inputs.run-kubernetes-end-to-end && inputs.branch == 'main') || github.event.schedule == '0 5 * * 1-5' }}
```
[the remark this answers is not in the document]
sam-heilbron: There are 2 "branches" that are at play here:
1. The branch to run the workflow from
2. The branch to use for tests

In this case, `inputs.branch` is the latter, and defines the branch to run tests for. You can select main, 1.16 or 1.15. This workflow will only run if you select the main branch to run the tests

.github/workflows/nightly-tests.yaml
sam-heilbron: I like it. i can give it a shot. we haven't done that in oss, so I held off on making that change, but happy to try to have consistency between our repos

.github/workflows/nightly-tests.yaml
sam-heilbron: I decided to split these into 2 actions, because it felt like they had separate concerns  (one example: we were passing credentials to authenticate to gcloud, for a job that executes tests). It felt just as easy to call 2 smaller actions (both on PRs and nightlies) than one larger one.

Can you help me understand how a single larger action is easier for re-use?

.github/workflows/nightly-tests.yaml
sam-heilbron: https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1716876298887489 is a perfect example I think of just how fragile the existing notification logic is

nfuden: Right its super fragile right now but having some sense of which tests failed is what Im interested in.

It does look like as you linkit will report that so thats good enough as long as there is one link to the main job

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -26,8 +26,12 @@ runs:
     env:
       GO_TEST_USER_ARGS: ${{ inputs.test-args }} -run "${{ inputs.run-regex }}"
       CLUSTER_NAME: ${{ inputs.cluster-name }}
-      CLUSTER_ID: ${{ inputs.cluster-id }}
       TEST_PKG: ./test/kubernetes/e2e/tests
     shell: bash
     run: make go-test
-  # TODO: Emit artifacts to shared location on failure
\ No newline at end of file
+  - name: Archive bug report directory on failure
```
sam-heilbron, line 32: This was moving the existing logic, which was defined in all places this action was called (PR workflow, nightly workflow) and instead define it with the code that produces the artifact (aka the tests themselves)

.github/workflows/nightly-tests.yaml
```diff
@@ -26,11 +26,82 @@ on:
...
+    timeout-minutes: 90
+    strategy:
+      # Since we are running these on a schedule, there is no value in failing fast
+      # In fact, we want to ensure that all tests run, so that we have a clearer picture of which tests are prone to flaking
+      fail-fast: false
+      matrix:
+        test:
+          # When running the tests at night, there is no value in splitting the tests across multiple clusters and running them in parallel.
+          # As a result, we increase the threshold for the tests, since they all run serially on a single cluster
+          - cluster-id: 'cluster-one'
```
sam-heilbron, line 61: Since the cluster-id is used to create the cluster name, the id needs to follow the requirements of a KinD cluster name. The previous format `cluster_one` resulted in an error since `_` is not allwed in the name

.github/workflows/nightly-tests.yaml
npolshakova: Decided on smaller composite actions and making it consistent for the other nightly/PR workflows: https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1716910010015199

.github/workflows/nightly-tests.yaml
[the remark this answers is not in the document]
npolshakova: Yep- I agree, then let's allow the specification of the branch to run workflow against as well as the regex? 
image

npolshakova (dismissed):

.github/workflows/nightly-tests.yaml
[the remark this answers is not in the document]
sam-heilbron: I think I solved this (at least for manual workflow) as a side effect of making changes to support the .env files. Basically when we run the "main" tests, we checkout the latest code. So I think we would achieve the same result that you are suggesting above, but the user would just select main.
I admit it's a bit weird...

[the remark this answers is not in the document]
npolshakova: Oh weird! So even if `branch to run tests against` was main, it would still run the latest tests on the branch?

nfuden (approved):

.github/workflows/helpers/notify/slack.go
```diff
@@ -0,0 +1,124 @@
+// Helper script designed to send a cohesive Slack notification about the result of a collection of GitHub workflows
+// This is a simpler variant of notify-from-json.go, due to some issues we
```
jenshu, line 2: nit: should we remove the reference to notify-from-json.go, since that file no longer exists?

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
       if: needs.prepare_env.outputs.should-auto-succeed-kube-tests != 'true'
       uses: ./.github/workflows/composite-actions/setup-kind-cluster
       with:
-        cluster-name: kind
+        cluster-name: kind-${{ matrix.test.cluster-id }}
```
jenshu, line 108: nit: don't need the kind- prefix on the names since kind will add it

.github/workflows/helpers/notify/slack.go
sam-heilbron: I decided to keep it here so that if you inspect the commit history, you know which file to look at to compare to. But I don't have strong feelings on it

.github/workflows/pr-kubernetes-tests.yaml
sam-heilbron: `nit: don't need the kind- prefix on the names since kind will add it`: yes and no. I agree that the kind prefix isn't necessary, but kind doesn't prefix the name, it prefixes the kube context which is a little more opaque. I figured that having the kind prefix in the name _could_ allow us in the future to run tests against a kind cluster and some other infrastructure provider, and thus differentiate the tests by the cluster-name.

.github/workflows/nightly-tests.yaml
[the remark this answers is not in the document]
sam-heilbron: looks like https://github.com/solo-io/gloo/pull/9511 may have introduced this for us, so ill merge that in

npolshakova (dismissed):

.github/workflows/.env/nightly-tests/max_versions.env
```diff
@@ -0,0 +1,5 @@
+node_version='v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245'
+kubectl_version='v1.28.4'
+kind_version='v0.20.0'
+helm_version='v3.13.2'
```
davidjumani, line 4:
```suggestion
helm_version='v3.14.4'
```

.github/workflows/.env/nightly-tests/max_versions.env
```diff
@@ -0,0 +1,5 @@
+node_version='v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245'
+kubectl_version='v1.28.4'
```
davidjumani, line 2:
```suggestion
kubectl_version='v1.29.2'
```

davidjumani (commented):
Changes related to the k8s bump LGTM

jbohanon (approved):

jenshu (approved):

PR #9531, opened 2024-05-28
Default GatewayParameters use only valid fields for image definitions

jbohanon:
# Description

Address flakes in CI setup due to disparity between fields available in main helm chart `Image` ([ref](https://github.com/solo-io/gloo/blob/53f2eb8ef4bc8f71525827a68a658dbd4e0384bc/install/helm/gloo/generate/values.go#L65)) and k8s gw deployer `helmImage` ([ref](https://github.com/solo-io/gloo/blob/53f2eb8ef4bc8f71525827a68a658dbd4e0384bc/projects/gateway2/deployer/values.go#L67))

The efficacy of this change can be seen on the solo-projects PR here https://github.com/solo-io/solo-projects/pull/6256


sam-heilbron (changes requested):
Can you add a test that demonstrates this working and would fail without your template changes?

sam-heilbron:
The new tests are failing:
```
Step #1 - "run-tests": Step #7 - "run-tests":     <string>: 1.17.0-beta33-9531
Step #1 - "run-tests": Step #7 - "run-tests": to equal
Step #1 - "run-tests": Step #7 - "run-tests":     <string>: 1.0.1-dev
Step #1 - "run-tests": Step #7 - "run-tests": [38;5;243m------------------------------[0m
Step #1 - "run-tests": Step #7 - "run-tests": [38;5;9m• [FAILED] [0.393 seconds][0m
Step #1 - "run-tests": Step #7 - "run-tests": [0mKubernetes Gateway API integration [38;5;243mwhen kube gateway integration is enabled [38;5;9m[1m[It] renders default GatewayParameters[0m
Step #1 - "run-tests": Step #7 - "run-tests": [38;5;243m/workspace/gloo/install/test/k8sgateway_test.go:57[0m
```

I will try to investigate later today if I have time, though I likely won't

install/helm/gloo/templates/_gg-helpers.tpl
```diff
@@ -36,3 +36,26 @@ We want this label to remain constant as it is used in glooctl version.
 {{- define "gloo-gateway.constLabels" -}}
 gloo: kube-gateway
 {{- end }}
+
+{{/*
+Images valid for the deployer
+ref struct helmImage in projects/gateway2/deployer/values.go
+*/}}
+{{- define "gloo-gateway.deployerImage" -}}
```
jenshu, line 44: nit re: the func name and comment here - these don't necessarily need to match the deployer's internal helm values, but they need to match the fields supported in the [image proto](https://github.com/solo-io/gloo/blob/6122c73612de3bed0cfd4f1f960f33e62c4b814f/projects/gateway2/api/v1alpha1/kube/container.proto#L13) used by the GatewayParameters

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -21,13 +21,13 @@ spec:
...
       image:
-{{- toYaml $image | nindent 8 }}
+{{- include "gloo-gateway.deployerImage" $image | nindent 8 }}
     podTemplate:
       extraLabels:
 {{- include "gloo-gateway.constLabels" . | nindent 8 }}
     sdsContainer:
       image:
-{{- merge $global.glooMtls.sds.image $global.image | toYaml | nindent 8 }}
+{{- merge $global.glooMtls.sds.image $global.image | include "gloo-gateway.deployerImage" | nindent 8 }}
```
npolshakova, line 30: Can we use the image from: `{{- $image := merge $gg.image $global.image }}` instead of `global.image`?

install/test/k8sgateway_test.go
```diff
@@ -1,73 +1,129 @@
...
 			ExpectWithOffset(1, err).NotTo(HaveOccurred(), "Failed to render manifest")
 			testManifest = tm
 		}
 
-		It("when kube gateway integration is enabled, relevant resources are rendered", func() {
-			prepareMakefile(namespace, helmValues{
-				valuesArgs: []string{
-					"kubeGateway.enabled=true",
-				},
+		JustBeforeEach(func() {
```
npolshakova, line 30: I was thinking a helm test like this: https://github.com/solo-io/gloo/blob/main/install/test/helm_test.go#L1001 (But for checking if the GatewayParameters were correctly rendered)

npolshakova: I think these new tests are also useful- but having a helm unit test is a quicker way to sanity check the render looks correct for the default values.

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: That image (`$gg.image`) is intended to be the envoy image. I need to clean that up but that seems like not the scope of this PR.

install/test/k8sgateway_test.go
jbohanon: It's not structured exactly the same but let me know if the test added in https://github.com/solo-io/gloo/pull/9531/commits/bb2b93595a3291eb2553d432ecceca72e9a6aa0a covers what you were thinking

install/helm/gloo/templates/_gg-helpers.tpl
jbohanon: I updated the comment and helper name

jenshu (approved):

install/helm/gloo/templates/43-gatewayparameters.yaml
npolshakova: Ah- so this value is always the envoy image? https://github.com/solo-io/gloo/blob/main/install/helm/gloo/values-template.yaml#L12 Can we rename the field to envoy image in a follow up?

npolshakova (approved):

install/helm/gloo/templates/43-gatewayparameters.yaml
jbohanon: Yes, I understand the confusion

sam-heilbron (approved):

Issue #9532, opened 2024-05-28
Support per-LTS branch .trivyignore files


sheidkamp:
### Gloo Edge Product

Open Source

### Gloo Edge Version

current

### Is your feature request related to a problem? Please describe.

When addressing CVEs on older LTS branches, we sometimes have to leave CVEs unadressed or introduce new CVEs as we fix older ones due to lack of support for a full set of fixes constrained by the kubernetes version.

For example: https://github.com/solo-io/gloo/pull/9452  and https://github.com/solo-io/gloo/pull/9451

The mechanism we have to handle this currently is the `.trivyignore` file. When the scheduled security scan runs in CI, it uses the same `.trivyignore` file from the main branch for all security scans. This means that by adding a CVE to the `.trivyignore` file, it will be ignored in scans for all versions, including versions where the CVE has been addressed, potentially masking regressions.

What we would like is the ability to define the CVEs to ignore on a per-version basis.

### Describe the solution you'd like

There are two approached to this to be considered:
- Maintain the CVE lists/`.trivyignore` files in each branch
   - Similar to how we use min/max versions in `solo-projects` test matrix
   - One CI job/step per LTS which checks out that branch's code
   - Clean to maintain, each branch is documenting
   - UX - aligns with development task of "build this image and scan for vulnerabilities" on single branch.
- Maintain a CVE list per LTS branch in main
   - Update jobs to use a different `.trivyignore` file per version, all stored on main
   - UX - aligns with command line task of "scan multiple versions at once"
   - Have to deprecate versions, stale files persist in LTS branches

### Additional Context
Both the OSS and EE scans are run from the OSS ci and use the OSS `.trivyignore` file.

#### CI from https://github.com/solo-io/gloo/blob/main/.github/workflows/trivy-analysis-scheduled.yaml
* Runs on main
* Calls `make run-security-scan` and `make publish-security-scan` with MIN_SCANNED_VERSION and other ENV variables set
* Calls one job, which does all scans

#### Makefile https://github.com/solo-io/gloo/blob/main/Makefile#L1203
* run-security-scan
   * Sets MIN_SCANNED_VERSION and calls `go run docs/cmd/generate_docs.go run-security-scan`
   * Runs for gloo and glooe image repositories.
* publish-security-scan
   * Copies files generated by run-security-scan to bucket
* scan-version https://github.com/solo-io/gloo/blob/main/Makefile#L1210
   * unused in the current process, but worth noting 

[...]

PR #9534, opened 2024-05-28
1.13.x - Clear .trivyignore file

sheidkamp:
# Description
Security scans are run from the main branch, so `.trivyignore` files are not used in the LTS branches.

We are removing their contents to avoid confusion when addressing CVEs


.trivyignore
```diff
@@ -1,37 +1 @@
...
-CVE-2023-45283
-
-# https://github.com/advisories/GHSA-4v7x-pqxf-cx7m
-# This vulnerability is resolved on all images except kubectl.
-# This HTTP2 vulnerability is not exploitable in Gloo Edge as the kubectl image is only used for jobs, not serving traffic.
-# A fix is not available in the supported Kubernetes version
-# https://github.com/solo-io/gloo/issues/9443
-CVE-2023-45288
\ No newline at end of file
+# This file is only meaningful in the `main` branch
```
sam-heilbron, line 1: On second thought, do you think it makes sense to keep the file as is, since our intention in the long run is to support a trivyignore per lts branch (https://github.com/solo-io/gloo/issues/9532)? This will also allow devs to run trivy scans on the branch they are testing, instead of having to do it from main.

I know I had proposed this in the first place so sorry for the flip-flop.

I still think a VERY LOUD comment at the top of this file to clarify its usage in CI is a big win

sheidkamp: It's a close call for me

It is useful to let the devs run the trivy scans from any branch, but it will give inconsistent results if different from main.

The command I used locally was: `for service in gloo gloo-envoy-wrapper discovery ingress sds certgen access-logger kubectl; do trivy image --severity HIGH,CRITICAL "quay.io/solo-io/${service}:1.13.37"; done`, which I took from an earlier PR.  The `trivy image` command takes an `--ignorefile` argument, so all they would have to do is copy the file from main to a non-git-controlled location and use that with the command, which I think would be a safer workflow until we figure out the per-branch approach.

If you have strong feelings, happy to revert and just add the comment at the top.

sam-heilbron (approved):

bewebi (approved):

PR #9538, opened 2024-05-29
.github/workflows: further iterations to running e2e tests consistently

sam-heilbron:
# Description

Improvements to our GitHub actions used to invoke e2e tests

# Context

## .env support for PRs
In https://github.com/solo-io/gloo/pull/9526 I had added support to rely on variables defined in `.env` files to dictate the versions of tools that we use in our nightly tests. I mirror this for our PR workflow, to reduce the chance of user-error when updating this in multiple places. 

I opted to define a standalone PR env file, instead of relying on the nightly-tests/max file. Even though they hold the same values, I figured they are different conceptually and may change. 

## Remove USE_SILENCE_REDIRECT

Internal slack ref: https://solo-io-corp.slack.com/archives/C04CJMXAH7A/p1716961972928579

We had a failure in our build step:
```
GO111MODULE=on CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -ldflags="-X github.com/solo-io/gloo/pkg/version.Version=1.0.0-ci1" -gcflags= -o /home/runner/work/gloo/gloo/_output/projects/envoyinit/cmd/envoyinit-linux-amd64 projects/envoyinit/cmd/main.go 2> /dev/null
cp projects/envoyinit/cmd/Dockerfile.envoyinit /home/runner/work/gloo/gloo/_output/projects/envoyinit/cmd/Dockerfile.envoyinit
cp projects/envoyinit/cmd/docker-entrypoint.sh /home/runner/work/gloo/gloo/_output/projects/envoyinit/cmd/docker-entrypoint.sh
docker buildx build --load  /home/runner/work/gloo/gloo/_output/projects/envoyinit/cmd -f /home/runner/work/gloo/gloo/_output/projects/envoyinit/cmd/Dockerfile.envoyinit \
	--build-arg GOARCH=amd64 \
	--build-arg ENVOY_IMAGE=quay.io/solo-io/envoy-gloo:1.30.1-patch1 \
	-t quay.io/solo-io/gloo-envoy-wrapper:1.0.0-ci1 --label quay.expires-after=3w 2> /dev/null
make: *** [Makefile:712: gloo-envoy-wrapper-docker] Error 1
```

But the logs were suppressed so it was difficult to debug. We originally suppressed these logs because they made it challenging to debug tests, but now tests and builds run in separate steps so the logs are distinct.

## Slack notifications
From a previous run ([logs](https://github.com/solo-io/gloo/actions/runs/9280537914/job/25536079545)), I noticed the following results:
```
{
  "end_to_end_tests_main": {
    "result": "cancelled",
    "outputs": {}
  },
  "regression_tests_main": {
    "result": "failure",
    "outputs": {}
  },
  "regression_tests_15": {
    "result": "skipped",
    "outputs": {}
  },
  "regression_tests_16": {
    "result": "skipped",
    "outputs": {}
  },
  "performance_tests_main": {
    "result": "cancelled",
    "outputs": {}
  },
  "performance_tests_15": {
    "result": "skipped",
    "outputs": {}
  },
  "performance_tests_16": {
    "result": "skipped",
    "outputs": {}
  },
  "kube_gateway_api_conformance_tests_main": {
    "result": "success",
    "outputs": {}
  }
}
```

We previously only notified on "result=failure". I updated this to include "cancelled" as well.

## Remove cluster-id in Github Action
In https://github.com/solo-io/gloo/pull/9526 we removed cluster-id as a concept from the code. @jenshu raised a good point here: https://github.com/solo-io/gloo/pull/9526#discussion_r1617794281 and I just updated the cluster-name to be used as the id.

## Interesting decisions
 -

## Testing steps
-

## Notes for reviewers
-

# Checklist:

- [ x] I have commented my code, particularly in hard-to-understand areas


.github/workflows/helpers/notify/slack.go
```diff
@@ -52,42 +57,65 @@ func main() {
...
+	var failureReasons []string
+
+	for reason, jobs := range unsuccessfulJobs {
+		if len(jobs) == 0 {
+			continue
+		}
+		reasonBuilder := strings.Builder{}
+		reasonBuilder.Write([]byte(reason))
+		reasonBuilder.Write([]byte(":"))
+		reasonBuilder.Write([]byte(strings.Join(jobs, ",")))
```
jenshu, line 113: i think you can also just use WriteString

.github/workflows/helpers/notify/slack.go
```diff
@@ -52,42 +57,65 @@ func main() {
...
+
+		failureReasons = append(failureReasons, reasonBuilder.String())
+	}
+
+	text := fmt.Sprintf(":red_circle: <$PARENT_JOB_URL|$PREAMBLE> have failed some jobs: %s", strings.Join(failureReasons, ","))
```
jenshu, line 118: `<$PARENT_JOB_URL|$PREAMBLE>` do these variables get resolved when sending the slack message? what's an example of what that would look like?

.github/workflows/nightly-tests.yaml
```diff
@@ -121,9 +121,9 @@ jobs:
         test:
           # When running the tests at night, there is no value in splitting the tests across multiple clusters and running them in parallel.
           # As a result, we increase the threshold for the tests, since they all run serially on a single cluster
-          - cluster-id: 'cluster-one'
+          - cluster-name: 'cluster-one'
             go-test-args: '-v -timeout=90m'
-            go-test-run-regex: ${{ inputs.kubernetes-end-to-end-run-regex }}
+            go-test-run-regex: ""
```
jenshu, line 126: why was the regex removed here?

.github/workflows/nightly-tests.yaml
```diff
@@ -326,6 +326,9 @@ jobs:
       - performance_tests_15
       - performance_tests_16
       - kube_gateway_api_conformance_tests_main
+      - end_to_end_tests_on_demand
+      - regression_tests_on_demand
+      - performance_tests_on_demand
```
jenshu, line 331: does having all these listed in `needs` mean that every job needs to run before the results are published?

sam-heilbron: Correct, though "run" could be: skipped, success, cancelled, failure (https://github.com/solo-io/gloo/blob/sam/ci-iterations/.github/workflows/helpers/notify/slack.go#L35).

https://github.com/solo-io/gloo/actions/runs/9288839885 is an example of a job that uses this. It includes a variety of results

.github/workflows/nightly-tests.yaml
sam-heilbron: It wasn't used. In a nightly run, there is no "input" value, so an empty string was passed. From the logs of the job that ran (https://github.com/solo-io/gloo/actions/runs/9280537914/job/25557214191):
```
GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore go test -ldflags="-X github.com/solo-io/gloo/pkg/version.Version=1.0.1-dev" \
-timeout=25m -cpu=4 -race -v -timeout=90m -run "" \
./test/kubernetes/e2e/tests
```

.github/workflows/helpers/notify/slack.go
sam-heilbron: https://solo-io-corp.slack.com/archives/C0314KESVNV/p1716999161128019 is an example:
```
[Gloo OSS Manual run (workflow_initiating_branch)](https://github.com/solo-io/gloo/actions/runs/9288839885) have failed some jobs: failure:end_to_end_tests_on_demand,cancelled:regression_tests_on_demand
```

davidjumani, line 118: nit
```suggestion
	text := fmt.Sprintf(":red_circle: <$PARENT_JOB_URL|$PREAMBLE> have failed some jobs: %s", strings.Join(failureReasons, "\n"))
```

.github/workflows/helpers/notify/slack.go
```diff
@@ -52,42 +57,65 @@ func main() {
...
+
+func sendFailure(unsuccessfulJobs map[string][]string) {
+	var failureReasons []string
+
+	for reason, jobs := range unsuccessfulJobs {
+		if len(jobs) == 0 {
+			continue
+		}
+		reasonBuilder := strings.Builder{}
+		reasonBuilder.Write([]byte(reason))
```
davidjumani, line 111: nit:
```suggestion
		reasonBuilder.Write([]byte("*"))
		reasonBuilder.Write([]byte(reason))
		reasonBuilder.Write([]byte("*"))
```

.github/workflows/helpers/notify/slack.go
```diff
@@ -52,42 +57,65 @@ func main() {
...
+		reasonBuilder.Write([]byte(":"))
```
davidjumani, line 112: nit:
```suggestion
		reasonBuilder.Write([]byte(": "))
```

davidjumani (dismissed):

.github/workflows/nightly-tests.yaml
jenshu: during a nightly test run, the 'on demand' tests won't run. will they automatically show as 'skipped' or 'cancelled'? i'm just wondering if the publish results wouldn't get called because it's waiting on the on-demand tests

sam-heilbron: I would expect any jobs that are skipped (due to the if condition not being met) will be reported as skipped. In the above logs I linked, we run the on_demand tests and thus skip the nightlies. The publish_results runs after all jobs have completed, including hte tests that were skipped or the tests that ran

jenshu (approved):

sam-heilbron:
bulldozer pls

PR #9540, opened 2024-05-29
gateway2: enable self-managed Gateways

shashankram:
# Description

Adds capability to integrate self-managed gateways that are 
not meant to be auto-provisioned by the
controller. This is required to support use cases
where gateways are statically provisioned, such
as when running the gateways external to k8s on
VMs.

It adds a selfManaged field to the GatewayParameters CRD which 
is used to skip the deployment of Gateway related objects (deployment, service, etc.). 
When a gateway is self-managed, it is expected to be
correctly bootstrapped with an envoy config that
matches the Gateway resource's name and namespace, 
specifically the `node` field must be configured
correctly to link a self-managed gateway to the
Gateway resource.

## API changes
Adds `selfManaged` field to GatewayParameters

## Interesting decisions
Standardized behavior to configure Gateway parameters.

## Testing steps

E2e test

# Checklist:


ilackarms (approved):

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -24,6 +25,9 @@ message GatewayParametersSpec {
   oneof environment_type {
     // The proxy will be deployed on Kubernetes.
     KubernetesProxyConfig kube = 1;
+
+    // The proxy will be self-managed and not auto-provisioned.
+    google.protobuf.Struct self_managed = 2;
```
yuval-k, line 30: why is this a Struct? did you mean to use Empty?

shashankram: Yes, per discussion with @ilackarms it's a convention we use for oneof instead of setting a bool

ilackarms (dismissed):

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -24,6 +25,9 @@ message GatewayParametersSpec {
   oneof environment_type {
     // The proxy will be deployed on Kubernetes.
     KubernetesProxyConfig kube = 1;
+
+    // The proxy will be self-managed and not auto-provisioned.
+    google.protobuf.Empty self_managed = 2;
```
lgadban, line 30: quick drive-by: do we not want to reference static provisioning in the naming here?
"Self managed" does sound completely reasonable but is "statically provisioned" the more accepted term?

shashankram: My personal feeling is that `static` is ambiguous, and `selfManaged` is clearer.

projects/gateway2/api/v1alpha1/gateway_parameters.proto
EItanya: I agree with @yuval-k  that we usually use `google.protobuf.Empty` since there's no content

yuval-k (approved):

PR #9544, opened 2024-05-29
GGv2: Name output routes

ilackarms:
just adds names to output routes sent to envoy. this enables plugins with caches to be aware when they are operating on the same route


jmhbh:
lets keep this open for now but after discussing with Scott i'm not sure if we even need to index the output routes.

timflannagan:
Closing - This was superseded by https://github.com/solo-io/gloo/pull/10010.

Issue #8107, opened 2023-04-20
Kube API unavailability results in a gloo container crash


Ati59:
### Gloo Edge Version

1.13.x (latest stable)

### Kubernetes Version

None

### Describe the bug

A customer is facing regular kube-API outage (on all clouds AWS, Azure and GCP) and when it happens, gloo container is crashing on the gloo pod (because of the election not able to choose the lead).
If the API server is unavailable during a scale-out event (increase of load for instance), the new gateway-proxy won't have the configuration from gloo due to this election problem.

### Steps to reproduce the bug

- Installing GE using this helm values : 
```yaml
global:
  glooMtls:
    enabled: true
  istioSDS:
    enabled: false
```
- Make the API-server unavailable some times (I used iptables rule : `iptables -A INPUT -p tcp --dport 6443 -j DROP`)
- Check the restart count and last container logs : 
```
2023-04-20T11:39:21.664432264Z stderr F E0420 11:39:21.663749       1 leaderelection.go:330] error retrieving resource lock gloo-system/gloo-ee: Get "https://10.6.0.1:443/api/v1/namespaces/gloo-system/configmaps/gloo-ee": context deadline exceeded
2023-04-20T11:39:21.664668472Z stderr F I0420 11:39:21.664461       1 leaderelection.go:283] failed to renew lease gloo-system/gloo-ee: timed out waiting for the condition
2023-04-20T11:39:21.666594222Z stderr F {"level":"error","ts":"2023-04-20T11:39:21.665Z","logger":"gloo-ee","caller":"kube/factory.go:61","msg":"Stopped Leading","version":"1.13.9","stacktrace":"github.com/solo-io/gloo/pkg/bootstrap/leaderelector/kube.(*kubeElectionFactory).StartElection.func2\n\t/go/pkg/mod/github.com/solo-io/gloo@v1.13.8/pkg/bootstrap/leaderelector/kube/factory.go:61\nk8s.io/client-go/tools/leaderelection.(*LeaderElector).Run.func1\n\t/go/pkg/mod/k8s.io/client-go@v0.22.4/tools/leaderelection/leaderelection.go:203\nk8s.io/client-go/tools/leaderelection.(*LeaderElector).Run\n\t/go/pkg/mod/k8s.io/client-go@v0.22.4/tools/leaderelection/leaderelection.go:213"}
[...]

Issue #7346, opened 2022-10-20
When leader election fails, gloo crashes


kevin-shelaga:
### Gloo Edge Version

1.12.x (latest stable)

### Kubernetes Version

1.21.x

####
Compromise: if gloo has good configuration it should not crash if it can not reach apiserver
Namely we should do the following:
`Allow candidates who lose leadership to fallback to a follower gracefully. Previously, we fataled to ensure that we do not have multiple leaders at once. The downside to this, is that lost leadership can occur either as a result of throttling or network failure with the ApiServer, which may occur intermittently in an installation of Gloo Edge. While there are other solutions to reduce the chance of these happening, we will change our leadership code to revert back to a follower (ie not write statuses) instead of crashing.`

### Describe the bug

At this moment we arent clear on the route cause, but when leader election fails gloo will crash. etcd and the masters were healthy, but there were resource limits on gloo during this time.

```
I1019 20:02:47.456285       1 leaderelection.go:248] attempting to acquire leader lease grp-gloo-system/gloo-ee...
I1019 20:03:04.435270       1 leaderelection.go:258] successfully acquired lease grp-gloo-system/gloo-ee
I1019 20:03:12.519136       1 trace.go:205] Trace[120689475]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.22.4/tools/cache/reflector.go:167 (19-Oct-2022 20:02:47.564) (total time: 24954ms):
Trace[120689475]: ---"Objects listed" 24906ms (20:03:12.470)
Trace[120689475]: [24.954889394s] [24.954889394s] END
[...]

PR #9563, opened 2024-05-31
Retry on leader lease renewal failure

davidjumani:
# Description

After a container has become a leader, any Kube API server unavailability results in the container crashing. This happens as the leader is unable to renew the lease. This is by design as outlined [here](https://github.com/solo-io/gloo/issues/8107#issuecomment-1516250478)
> if we allowed gloo to continue to function as a leader during kube apiserver outage, we risk having two leaders in other failure modes. we should remove the panic and allow gloo to continue to serve last-known xds as a follower (effectively having two followers until kube apiserver recovers). this idea is similar to the role xds relay could play for gloo edge

However, crashing the gloo pods can also lead to an outage during scaling as the gateway-proxy pods that come up cannot fetch any configs and all routes result in 404s.

Now instead of crashing, the gloo pod falls back to a follower. This prevents an outage and any other pod can take over as leader if possible
This is fine as a leader only writes reports / statuses over [here](https://github.com/solo-io/gloo/blob/0637825838c285e978c51d6a00f68f4e1d5d08d6/pkg/bootstrap/leaderelector/leader.go#L35) and [here](https://github.com/solo-io/gloo/blob/0637825838c285e978c51d6a00f68f4e1d5d08d6/projects/gloo/pkg/syncer/translator_syncer.go#L231). On any failure, the pod becomes a follower and if elected back as a leader will continue to write reports.

## Code changes
- Introduce a new `Reset` method on the identity implementation that allows an identity to fall back to a follower

## CI changes
- Kind clusters that run in CI now do not come up with their own CNI as kindnet does not support custom network policies.
Instead, cilium is installed as CNI as we need to test Kube API server unavailability

# Context
[Kube API unavailability results in a gloo container crash](https://github.com/solo-io/gloo/issues/8107)
[When leader election fails, gloo crashes](https://github.com/solo-io/gloo/issues/7346#top)
[Design Doc](https://soloio.slab.com/posts/dont-crash-on-failure-to-renew-leader-lease-kjkqn0ry)

## Testing steps
Kube2e tests to verify the following :
- It can recover from a Kube APIServer unavailability
- It falls back to a follower
- Another leader can get elected

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8107


pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -55,39 +61,88 @@ func (f *kubeElectionFactory) StartElection(ctx context.Context, config *leadere
...
+			l.Run(ctx)
+
+			if !recoverFromLeaderElectionFailure {
+				contextutils.LoggerFrom(ctx).Fatalf("lost leadership, quitting app")
+			}
+
+			contextutils.LoggerFrom(ctx).Errorf("Leader election cycle %v lost. Trying again", counter.Load())
+			counter.Add(1)
+			// Sleep for a while as this might be a transient issue
+			time.Sleep(5 * time.Second)
```
davidjumani, line 143: Perhaps sleep as long as it takes the lease to expire? Thought

ashishb-solo: perhaps just make this configurable (also via environment variable)?

at the very least, i think this value should be a const variable instead of hard-coded in here

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -55,39 +61,91 @@ func (f *kubeElectionFactory) StartElection(ctx context.Context, config *leadere
...
-				OnNewLeader: func(identity string) {
-					contextutils.LoggerFrom(ctx).Debugf("New Leader Elected with Identity: %s", identity)
-					config.OnNewLeader(identity)
+	var counter atomic.Uint32
+	var justFailed = false
+	var dontDie func()
+
+	newLeaderElector := func() (*k8sleaderelection.LeaderElector, error) {
+
+		dieIfUnrecoverable := func(ctx context.Context) {
```
ashishb-solo, line 70: some documentation here would be helpful. i gave it a first try, but i'm not sure if it's entirely accurate

```suggestion
                // this function causes gloo to exit after a specified amount of time of trying to recover after leader election failure. we do this to safeguard against ... (to be completed)
		dieIfUnrecoverable := func(ctx context.Context) {
```

projects/gloo/pkg/setup/setup.go
```diff
@@ -41,11 +41,7 @@ func startSetupLoop(ctx context.Context) error {
 				contextutils.LoggerFrom(ctx).Infof("new leader elected with ID: %s", leaderId)
 			},
 			OnStoppedLeading: func() {
-				// Kill app if we lose leadership, we need to be VERY sure we don't continue
-				// any leader election processes.
-				// https://github.com/solo-io/gloo/issues/7346
-				// There is follow-up work to handle lost leadership more gracefully
-				contextutils.LoggerFrom(ctx).Fatalf("lost leadership, quitting app")
+				contextutils.LoggerFrom(ctx).Errorf("lost leadership")
```
ashishb-solo, line 44: question: i may have just missed this, but i don't see where exit occurs if gloo is configured to not recover from failure. could you clarify where/how that happens?

ashishb-solo: additionally, i think it may be helpful to replace the deleted comment with an updated one (ie. we kill the app unless configured to recover from losing leadership)

ashishb-solo (commented):
thanks @davidjumani - i've read it over a few times and have a few questions. i think a little more documentation (especially around the go funcs) would be helpful. also, a major question regarding whether we still have the regular behavior if we're not configured to recover from failure.

happy to talk in person if that would be easier

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+		fmt.Println(out)
+		return strings.Contains(out, "successfully rolled out"), err
+	}, "30s", "1s").Should(BeTrue())
+}
+
+func simulateKubeAPIServerUnavailability() {
+	out, err := testutils.KubectlOut("apply", "-f", testHelper.RootDir+"/test/kube2e/gloo/artifacts/block.yaml")
+	Expect(err).ToNot(HaveOccurred())
+	fmt.Println(out)
+	time.Sleep(15 * time.Second)
```
ashishb-solo, line 507: just a suggestion here, but i think kubernetes has ways to wait until a condition is met: https://stackoverflow.com/questions/53536907/kubectl-wait-for-condition-complete-timeout-30s

perhaps that would be more robust than sleeping like this?

ci/kind/cluster.yaml
```diff
@@ -1,5 +1,7 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+networking:
```
sam-heilbron, line 3: Why do we do this? Perhaps a comment clarifying?

ci/kind/setup-kind.sh
```diff
@@ -36,6 +36,11 @@ function create_kind_cluster_or_skip() {
     --name "$CLUSTER_NAME" \
     --image "kindest/node:$CLUSTER_NODE_VERSION" \
     --config="$SCRIPT_DIR/cluster.yaml"
+  helm repo add cilium https://helm.cilium.io/
```
sam-heilbron, line 39: This adds a dependency into our setup script, that our tests now rely on. In the past our setup-kind script has been a nice utilty that makes it easier for devs to run tests. I think if we move forward with this approach, we need to update our readmes and test that rely on this to clarify what happens if you don't have cilium installed

ci/kind/setup-kind.sh
```diff
@@ -36,6 +36,11 @@ function create_kind_cluster_or_skip() {
...
+  helm install cilium cilium/cilium --version 1.15.5 \
```
sam-heilbron, line 40: nit: I think versions are more easily tracked if they are variables that you can pass into a script

pkg/bootstrap/leaderelector/identity.go
```diff
@@ -37,3 +37,8 @@ func (i identityImpl) IsLeader() bool {
 func (i identityImpl) Elected() <-chan struct{} {
 	return i.elected
 }
+
+// Reset updates the current identity to a follower. It can be promoted to a leader by closing the elected channel
+func (i identityImpl) Reset(elected <-chan struct{}) {
+	i.elected = elected
```
sam-heilbron, line 43: Is this goroutine safe?

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -21,9 +23,12 @@ const (
...
 
-	leaseDurationEnvName = "LEADER_ELECTION_LEASE_DURATION"
-	retryPeriodEnvName   = "LEADER_ELECTION_RETRY_PERIOD"
-	renewPeriodEnvName   = "LEADER_ELECTION_RENEW_PERIOD"
+	recoveryTimeout = 60 * time.Second
+
+	leaseDurationEnvName                    = "LEADER_ELECTION_LEASE_DURATION"
+	retryPeriodEnvName                      = "LEADER_ELECTION_RETRY_PERIOD"
+	renewPeriodEnvName                      = "LEADER_ELECTION_RENEW_PERIOD"
+	recoverFromLeaderElectionFailureEnvName = "RECOVER_FROM_LEADER_ELECTION_FAILURE"
```
sam-heilbron, line 31: How will users discover these values so they can use them? Are there user-facing documentation around the toggles that we expose and the behaviors that are available?

would we want to make this "first class" in our API and make it a helm value?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+				logs := getGlooDeploymentLogs()
+				g.Expect(logs).To(ContainSubstring("Leader election cycle 0 lost. Trying again"))
+				g.Expect(logs).To(ContainSubstring("recovered from lease renewal failure"))
+				g.Expect(logs).NotTo(ContainSubstring("lost leadership, quitting app"))
+			}, "30s", "1s")
+		})
+	})
 })
+
+func ModifyDeploymentEnv(resourceClientset *kube2e.KubeResourceClientSet, deploymentName string, containerIndex int, envVar corev1.EnvVar) {
```
sam-heilbron, line 462: We have a function that attempts to achieve this here: https://github.com/solo-io/solo-projects/blob/main/test/kube2e/test_context.go#L264. I think we should do the following:
- Define this in a more shared place for kube tests to use
- Pass in just the required inputs (ie just the kube client, not the whole resource clientset)
- Perhaps use containerName instead of index?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+
+	WaitForRolloutWithOffset(1, deploymentName, namespace, "60s", "1s")
+}
+
+// WaitForRollout waits for the specified deployment to be rolled out successfully.
+func WaitForRollout(deploymentName string, deploymentNamespace string, intervals ...interface{}) {
+	WaitForRolloutWithOffset(1, deploymentName, deploymentNamespace, intervals...)
+}
+
+func WaitForRolloutWithOffset(offset int, deploymentName string, deploymentNamespace string, intervals ...interface{}) {
```
sam-heilbron, line 495: This feels like a good candidate for a function to live here: https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/kubectl/cli.go. This will make it easier for other tests to benefit from

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+	EventuallyWithOffset(offset+1, func() (bool, error) {
+		out, err := testutils.KubectlOut("rollout", "status", "-n", deploymentNamespace, fmt.Sprintf("deployment/%s", deploymentName))
+		fmt.Println(out)
+		return strings.Contains(out, "successfully rolled out"), err
+	}, "30s", "1s").Should(BeTrue())
+}
+
+func simulateKubeAPIServerUnavailability() {
```
sam-heilbron, line 503: https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/kubectl/cli.go#L96

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+	out, err := testutils.KubectlOut("-n", "gloo-system", "logs", "deploy/gloo")
+	Expect(err).ToNot(HaveOccurred())
+	fmt.Println(out)
+	return out
+}
+
+func waitUntilLeaseAcquired() {
+	Eventually(func(g Gomega) {
+		out := getGlooDeploymentLogs()
+		g.Expect(out).To(ContainSubstring("successfully acquired lease gloo-system/gloo"))
```
sam-heilbron, line 523: Are we sure that this is only getting the "new" logs? Each time we start gloo, this log will appear, so I just want to be sure that this isn't asserting that it happens once, but that we are asserting it happened on the most recent pod

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+	}, "30s", "2s")
```
sam-heilbron, line 524: We need to include an assertion (should succeed), otherwise this is a no-op

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+	}
+	_, err = deploymentClient.Update(ctx, d, metav1.UpdateOptions{})
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	WaitForRolloutWithOffset(1, deploymentName, namespace, "60s", "1s")
+}
+
+// WaitForRollout waits for the specified deployment to be rolled out successfully.
+func WaitForRollout(deploymentName string, deploymentNamespace string, intervals ...interface{}) {
+	WaitForRolloutWithOffset(1, deploymentName, deploymentNamespace, intervals...)
```
sam-heilbron, line 492: I think we want to start using the GinkoHelper: https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure:
```go
func EnsureUserCanRead(book Book, user User) {
  GinkgoHelper()
  if book.Title == "Les Miserables" && user.Age <= 3 {
    Fail("user is too young for this book") //note the optional offset is gone
  }
}
```


Perhaps we update https://github.com/solo-io/gloo/blob/main/devel/testing/writing-tests.md to clarify this convention?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
 			})
 		})
 	})
+
+	FContext("Retry leader election failure", func() {
```
sam-heilbron, line 425: `ginkgo unfocus` will unfocus all specs

pkg/bootstrap/leaderelector/kube/factory.go
sam-heilbron, line 31: What is the recommended value for customers? Can we reflect this explicitly in our kube tests as well? Even if we are setting it to the default value, I think it's helpful to be explicit so devs know the configuration

test/kube2e/gloo/artifacts/block.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: cilium.io/v2
```
sam-heilbron, line 1: I think it might be interesting to write up an RFE around introducing Chaos testing to the product (or expand one if there exists one). This type of cluster operation feels like the perfect candidate for a chaos test, where intermittently we can deny access to the kube apiserver and assert that the control plane can recover

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +421,106 @@ var _ = Describe("Bootstrap Clients", func() {
...
+			ModifyDeploymentEnv(resourceClientset, "gloo", 0, corev1.EnvVar{
+				Name:  "RECOVER_FROM_LEADER_ELECTION_FAILURE",
+				Value: "true",
+			})
+
+			waitUntilLeaseAcquired()
+			simulateKubeAPIServerUnavailability()
+
+			Eventually(func(g Gomega) {
+				logs := getGlooDeploymentLogs()
```
sam-heilbron, line 453: This feels like a good start to an assertion, but I think we want to be more aggressive here with validating behavior. Logging is one way to do it, but on top of that, I think we want to assert some action that the control plane can take (ie perhaps hcange config and confirm that it's propagated to envoy and statuses are updated)

sam-heilbron (commented):
looking good! I mainly focused on how we're testing for now, and once that is updated, I'm happy to take a closer look at the implementation itself

ci/kind/cluster.yaml
ashishb-solo: this is explained in the pull request comment:

```
    Kind clusters that run in CI now do not come up with their own CNI as kindnet does not support custom network policies.
    Instead, cilium is installed as CNI as we need to test Kube API server unavailability
```

but i agree that a comment would be helpful (including a link to this pull request in case this causes issues in other tests)

davidjumani: Added a note

pkg/bootstrap/leaderelector/identity.go
davidjumani: Yes since this struct is only accessed via the interface it implements

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -55,39 +61,88 @@ func (f *kubeElectionFactory) StartElection(ctx context.Context, config *leadere
...
+	_, err = newLeaderElector()
 	if err != nil {
 		return identity, err
 	}
 
-	// Start the leader elector process in a goroutine
-	contextutils.LoggerFrom(ctx).Debug("Starting Kube Leader Election")
-	go l.Run(ctx)
-
+	go func() {
```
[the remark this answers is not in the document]
davidjumani: Added comments above this

projects/gloo/pkg/setup/setup.go
davidjumani: Added comments in both places the changes were made

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: I belive that just a sensible value is better. I wouldn't want to expose this as it is a simple delay

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: Thanks, added comments around this function

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: I did look at it but unfortunately this block of code applies a yaml and then deletes it. Kubectl wait can not do anything useful here

ci/kind/setup-kind.sh
davidjumani: As of now, this is needed only for the gloo suite. Shall I restrict this to run only if the gloo suite is tested? I don't think a note might not be necessary as running the setup-kind script abstracts this step can still be run by devs

ci/kind/setup-kind.sh
davidjumani: Updated to use an env var with a default value

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: Yes, it is mentioned in the changelog. During the design discussion, it was decided that this feature gate would be an env variable, not a settings object.
The reasoning is that as of now, the only function of a leader is to write reports, however this can change in the future. To mitigate the possibility of deprecating a settings configuration, we use an env var

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Copied it over from that file. I've tweaked it to be general purpose and live in the helper module

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Added a comment explaining why we are sure only the most recent logs appear

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Thanks, added

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: Added to the tests

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Added a check to verify that a resource created is accepted and translated

pkg/bootstrap/leaderelector/identity.go
sam-heilbron: I'm not sure I follow exactly. `only accessed via the interface it implements` - Are you saying this is a private method that we control the usage of, and therefore are guaranteed access? If not, how do we control who calls this code and how they use it?

pkg/bootstrap/leaderelector/kube/factory.go
sam-heilbron: `Yes, it is mentioned in the changelog.` - A changelog is useful when customers are upgrading, but I think it will become decreasingly visible overtime when users are looking to use this feature. Do you think it makes more sense to call it out in a user-facing guide around HA?

ci/kind/setup-kind.sh
sam-heilbron: Yeah that's fair. I kind of like it being applied across all suites (as you have it). In our newer style tests (https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/load_balancing_tests.md#what-is-our-current-strategy) we try to make our infrastructure less aware of what is running on that cluster

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -21,9 +23,12 @@ const (
...
 
-	leaseDurationEnvName = "LEADER_ELECTION_LEASE_DURATION"
-	retryPeriodEnvName   = "LEADER_ELECTION_RETRY_PERIOD"
-	renewPeriodEnvName   = "LEADER_ELECTION_RENEW_PERIOD"
+	recoveryTimeout = 60 * time.Second
+
+	leaseDurationEnvName                       = "LEADER_ELECTION_LEASE_DURATION"
+	retryPeriodEnvName                         = "LEADER_ELECTION_RETRY_PERIOD"
+	renewPeriodEnvName                         = "LEADER_ELECTION_RENEW_PERIOD"
+	recoverIfKubeAPIServerIsUnreachableEnvName = "RECOVER_IF_KUBE_API_SERVER_UNREACHABLE"
```
sam-heilbron, line 31: nit: The previous pattern was to have all relevant env variables prefixed with `LEADER_ELECTION`. Any reason we don't continue that pattern?

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -55,39 +61,99 @@ func (f *kubeElectionFactory) StartElection(ctx context.Context, config *leadere
...
+	_, err = newLeaderElector()
 	if err != nil {
 		return identity, err
 	}
 
-	// Start the leader elector process in a goroutine
-	contextutils.LoggerFrom(ctx).Debug("Starting Kube Leader Election")
-	go l.Run(ctx)
-
+	// leaderElector.Run() is a blocking method but we need to return the identity of this container to sub-components so they can
```
sam-heilbron, line 135: ❤️

test/kube2e/gloo/artifacts/helm.yaml
```diff
@@ -17,7 +17,7 @@ gateway:
     alwaysAcceptResources: false
 gloo:
   logLevel: info
-  disableLeaderElection: true
+  disableLeaderElection: false
```
sam-heilbron, line 20: 🚀

sam-heilbron: Given this improvement to retry on lease renewal, I think it would make sense to enable leader election on all of our kube2e tests (even if we just run with a single replica). wdyt?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+		})
+
+		// During this test :
+		// - Scale up the deployment to 2 pods
+		// - Block kube api access to the leader pod
+		// - Ensure the leader pod loses leadership
+		// - Create a resource and verify it has been translated : This verifies that the other pod has become a leader
+		It("concedes leadership to another pod", func() {
+			helper.ModifyDeploymentEnv(ctx, deploymentClient, testHelper.InstallNamespace, "gloo", 0, corev1.EnvVar{
+				Name:  "RECOVER_IF_KUBE_API_SERVER_UNREACHABLE",
```
sam-heilbron, line 545: nit: We could use the constant defined in the source code

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+			// we can be sure that the logs fetched were generated only after this test had begun
+			waitUntilStartsLeading()
+
+			// Get the leader pod. Since there is only one pod it is the leader
+			name, err := testutils.KubectlOut("get", "pods", "-n", testHelper.InstallNamespace, "-l", "gloo=gloo", "-o", "jsonpath='{.items[0].metadata.name}'")
+			Expect(err).ToNot(HaveOccurred())
+			name = strings.ReplaceAll(name, "'", "")
+
+			// Scale the deployment to 2 replicas so the other can take over when the leader is unable to communicate with the Kube API server
+			err = testutils.Kubectl("scale", "-n", testHelper.InstallNamespace, "--replicas=2", "deploy/gloo", "--timeout=300s")
```
sam-heilbron, line 559: nit: could we expand https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/kubectl/cli.go so that other test code can just invoke it?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+
+			// Label the leader so the network policy can block communication to the Kube API server
+			pod, err := resourceClientset.KubeClients().CoreV1().Pods(testHelper.InstallNamespace).Get(ctx, name, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+			pod.Labels["block"] = "this"
+			_, err = resourceClientset.KubeClients().CoreV1().Pods(testHelper.InstallNamespace).Update(ctx, pod, metav1.UpdateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			// Block the leader's communication to the Kube API server
+			blockPodYAML := `apiVersion: cilium.io/v2
```
sam-heilbron, line 574: nit: I think this could live in a standalone manifest. I think the value of having these defined in a file, is that we can more easily use it in local dev, or docs, because we can reference the file directly

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			// Verify that the leader has stopped leading
+			Eventually(func(g Gomega) {
+				logs := helper.GetContainerLogs(testHelper.InstallNamespace, "pod/"+name)
+				g.Expect(logs).To(ContainSubstring("lost leadership"))
+			}, "60s", "1s").Should(Succeed())
+
+			// Ensure that we can still operate. This also validates that the second pod has become the leader
+			verifyTranslation()
```
sam-heilbron, line 605: If this functionality is broken (our leadership handoff) will this check fail? I would have imagined that something status specific would be relevant to catch regressions since status reporting is gated by leadership.

Should we also check out metrics to demonstrate that both behavior and the signals to users have changed?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+
+			// Cleanup the network policy. With connectivity restored, the old leader can become a follower
+			_, err = testutils.KubectlOut("delete", "-f", tmpFile.Name())
```
sam-heilbron, line 608: nit: We have [cli.DeleteFile](https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/kubectl/cli.go#L123)

test/kube2e/helper/kube.go
```diff
@@ -0,0 +1,60 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+
+	. "github.com/onsi/gomega"
+	"github.com/solo-io/go-utils/testutils"
+)
+
+func ModifyDeploymentEnv(ctx context.Context, deploymentClient clientsv1.DeploymentInterface, namespace string, deploymentName string, containerIndex int, envVar corev1.EnvVar) {
```
sam-heilbron, line 16: nit: I think in this function we are combining some logic to perform an action (changing deployment) and some logic to assert a behavior  (that nothing went wrong). 

An alternative is a function that does the action and returns an error, and a separate function that the test uses to assert that we do not error. It would mean that we can reduce the amount of "assert with offset" logic that we need to use, and makes the code more re-usable because we can invoke it from tests that don't use ginkgo.

Not a blocker, just sharing a methodology approach

test/kube2e/helper/kube.go
```diff
@@ -0,0 +1,60 @@
...
+
+// WaitForRollout waits for the specified deployment to be rolled out successfully.
+func WaitForRollout(deploymentName string, deploymentNamespace string, intervals ...interface{}) {
+	WaitForRolloutWithOffset(1, deploymentName, deploymentNamespace, intervals...)
+}
+
+// WaitForRolloutWithOffset waits for the specified deployment to be rolled out successfully with an offset.
+func WaitForRolloutWithOffset(offset int, deploymentName string, deploymentNamespace string, intervals ...interface{}) {
+	EventuallyWithOffset(offset+1, func() (bool, error) {
+		out, err := testutils.KubectlOut("rollout", "status", "-n", deploymentNamespace, fmt.Sprintf("deployment/%s", deploymentName))
```
sam-heilbron, line 50: https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/kubectl/cli.go#L160?

test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -418,4 +425,222 @@ var _ = Describe("Bootstrap Clients", func() {
...
+
+			// Simulate the kube api server is down and bring it up after 15 seconds. While it is down :
+			// - The leader should lose the lease
+			// - Create a VS
+			// When the kube api server is back up :
+			// - The VS should be accepted
+			// - It should be translated into an envoy config
+			restoreKubeAPIServer := simulateKubeAPIServerDown()
+			go func() {
+				time.Sleep(15 * time.Second)
```
sam-heilbron, line 515: Instead of a time.Sleep, which create an inherit race (even if its one that we're confident will succeed), what if we used a channel? And basically we signal the channel when we're ready to restore the kube api server?

ci/kind/setup-kind.sh
davidjumani: Thanks. Then I'll keep it this way with comments above the code

pkg/bootstrap/leaderelector/identity.go
davidjumani: So the components that rely on this struct only access it via the Identity interface it implements. That means that reset() can only be called by a method that creates a new identity, ie. StartElection() over [here](https://github.com/solo-io/gloo/pull/9563/files#diff-106d1d8f33ca97ee869dfe39622ccc60402c33519e8704792a0e15440daef918R49)
This way the components that use IsLeader() to perform leader related tasks cannot reset it to a follower and it can be controlled only in the leader election part of the code

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: That would be a good idea. I can work with the docs team on an HA guide. However for the sake of not deprecating any new helm / settings value that we introduce, I prefer to keep it as just an env variable

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: I would like to however this env var does not solely apply to leader election. It handles kube apiserver failure which would crash the leader but that is not the case with this pr. Am open to other variable names

test/kube2e/gloo/artifacts/helm.yaml
davidjumani: I have no strong opinion, however we do provide the ability to disable leader election so IMO it should be used in some test so we know that it works (or at least when it breaks)

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Good idea, updated

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Good idea, moved

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: It should fail since the resource will not have its status written and this method verifies that the resource is accepted

test/kube2e/helper/kube.go
davidjumani: I've added the kubecli to the testHelper. That should allow the users to write actions and assertions into a place that can be used by test while reusing the kubecli struct

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Reused the kubectl/cli while adding assertions

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Updated

test/kube2e/gloo/artifacts/block.yaml
davidjumani: I completely agree. This is a targetted chaos test

test/kube2e/helper/kube.go
davidjumani: Updated

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Updated

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: Updated

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: The purpose of a time.Sleep is to simulate an intermittent issue. It needs to be down for at least the lease renewal  duration (which is a few seconds). Keeping it 15s just to be safe
The `concedes leadership to another pod` test does follow your idea albeit without channels since we want to control when the connectivity will be restored

davidjumani:
/kick

```
Step #1 - "run-tests": Step #7 - "run-tests":   [38;5;9m[FAILED] HttpRoute should be included in input snap
Step #1 - "run-tests": Step #7 - "run-tests":   Expected
Step #1 - "run-tests": Step #7 - "run-tests":       <map[string]interface {} | len:16>: {
Step #1 - "run-tests": Step #7 - "run-tests":           "Proxies": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Ratelimitconfigs": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "VirtualHostOptions": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "AuthConfigs": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Gateways": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "GraphqlApis": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "RouteTables": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Secrets": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Upstreams": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Artifacts": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "Endpoints": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "HttpGateways": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "RouteOptions": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "TcpGateways": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "UpstreamGroups": nil,
Step #1 - "run-tests": Step #7 - "run-tests":           "VirtualServices": nil,
Step #1 - "run-tests": Step #7 - "run-tests":       }
Step #1 - "run-tests": Step #7 - "run-tests":   to have key
Step #1 - "run-tests": Step #7 - "run-tests":       <string>: gateway.networking.k8s.io.HTTPRoute[0m
Step #1 - "run-tests": Step #7 - "run-tests":   [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/workspace/gloo/projects/gloo/pkg/servers/iosnapshot/history_test.go:157[0m [38;5;243m@ 06/10/24 11:34:28.404[0m
Step #1 - "run-tests": Step #7 - "run-tests":   [38;5;9mFull Stack Trace[0m
Step #1 - "run-tests": Step #7 - "run-tests":     github.com/solo-io/gloo/projects/gloo/pkg/servers/iosnapshot.init.func1.3.2()
Step #1 - "run-tests": Step #7 - "run-tests":     	/workspace/gloo/projects/gloo/pkg/servers/iosnapshot/history_test.go:157 +0x928
Step #1 - "run-tests": Step #7 - "run-tests": [38;5;243m------------------------------[0m
```

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -264,3 +264,13 @@ func (c *Cli) Execute(ctx context.Context, stdin *bytes.Buffer, args ...string)
 
 	return stdout.String(), stderr.String(), err
 }
+
+func (c *Cli) Scale(ctx context.Context, namespace string, resource string, replicas uint) error {
+	return c.RunCommand(ctx, "scale", "-n", namespace, fmt.Sprintf("--replicas=%d", replicas), resource, "--timeout=300s")
+}
+
+// GetContainerLogs retrieves the logs for the specified container
+func (c *Cli) GetContainerLogs(ctx context.Context, namespace string, name string) (string, error) {
```
nfuden, line 273: nit: I dont love that we respect kubecontext in some commands but not in others

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -242,17 +242,17 @@ func (c *Cli) CurlFromPod(ctx context.Context, podOpts PodExecOptions, options .
...
 }
 
-func (c *Cli) ExecuteOn(ctx context.Context, kubeContext string, stdin *bytes.Buffer, args ...string) (string, string, error) {
+func (c *Cli) ExecuteOn(ctx context.Context, kubeContext string, args ...string) (string, string, error) {
 	args = append([]string{"--context", kubeContext}, args...)
-	return c.Execute(ctx, stdin, args...)
+	return c.Execute(ctx, args...)
 }
 
-func (c *Cli) Execute(ctx context.Context, stdin *bytes.Buffer, args ...string) (string, string, error) {
```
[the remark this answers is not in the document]
nfuden: nit: In general we should try to not be too aggressive with deletions of Exported functions in gloo as there may be some other consumer we are now breaking.

test/kube2e/gloo/bootstrap_clients_test.go
nfuden: I think in this case sleep works pretty well. Honestly a time.after would have been cleaner but would functionally be roughly the same

test/kube2e/gloo/artifacts/block.yaml
nfuden: This feels like an expansion in scope. We have some basic chaos testing patterns in other products which we should first consider reusing.

nfuden (commented):
Generally this makes sense to me. 

My main remaining question after trying it out is more of an impact question. Namely should we allow users to determine their risk appetite more granularly. 
In this case we don't have an over arching cleanup strategy. 

Consider the case where a single pod is unable to reach the kube apiserver but other pods do not have the same issue. If the user turns on the new environment variable it will never clean itself up. While this risk is inherent in this strategy I wonder if the UX hit we take would be worth it to make this not just a bool but a configuration knob that allows environments with known apiserver loss characteristics to toggle their retry duration.

nfuden (dismissed):

nfuden (dismissed):

pkg/bootstrap/leaderelector/identity.go
ashishb-solo: it sounds like your claim here is that it is safe from concurrent writes. how about reads? i worry about something like this:

1. component reads `i.Elected()`, gets true
2. pod loses leader election
3. component does something since it believes it's the leader when it isn't

do you think this is a serious potential failure scenario?

pkg/bootstrap/leaderelector/kube/factory.go
```diff
@@ -55,39 +74,98 @@ func (f *kubeElectionFactory) StartElection(ctx context.Context, config *leadere
...
-					config.OnStartedLeading(callbackCtx)
-				},
-				OnStoppedLeading: func() {
-					contextutils.LoggerFrom(ctx).Error("Stopped Leading")
-					config.OnStoppedLeading()
-				},
-				OnNewLeader: func(identity string) {
-					contextutils.LoggerFrom(ctx).Debugf("New Leader Elected with Identity: %s", identity)
-					config.OnNewLeader(identity)
+	var counter atomic.Uint32
```
ashishb-solo, line 77: nit: this value is only used in a single function (the goroutine that runs the leader elector loop) and could be moved internal to it

ashishb-solo (commented):
looks good to me. one question on the comment from Sam regarding whether the goroutine is safe. if so, happy to ignore the nits to approve

pkg/bootstrap/leaderelector/kube/factory.go
davidjumani: Good catch, moved

test/kube2e/helper/kube.go
```diff
@@ -0,0 +1,60 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func (h *SoloTestHelper) ModifyDeploymentEnv(ctx context.Context, deploymentClient clientsv1.DeploymentInterface, namespace string, deploymentName string, containerIndex int, envVar corev1.EnvVar) {
```
ashishb-solo, line 16:
```suggestion
// Modify an environment variable in a deployment
func (h *SoloTestHelper) ModifyDeploymentEnv(ctx context.Context, deploymentClient clientsv1.DeploymentInterface, namespace string, deploymentName string, containerIndex int, envVar corev1.EnvVar) {
```

pkg/bootstrap/leaderelector/identity.go
davidjumani: The leader only writes reports, so overwriting reports should not be a problem. Nonetheless, if it loses connectivity to the kube apiserver at the time it loses leadership, it would be unable to write the reports

ashishb-solo (approved):

nfuden (approved):

test/kube2e/gloo/bootstrap_clients_test.go
davidjumani: True but time.After takes up more resources and will need cleanup if the test errors out / panics before the timer expires

pkg/utils/kubeutils/kubectl/cli.go
davidjumani: This context is the test context and not the kube context. The kube context is still honoured if the cli.kubeContext is set

davidjumani: Will address this in another PR

davidjumani:
kick bulldozer

PR #9572, opened 2024-06-03
iosnapshot: support Kubernetes Gateway resources in input snapshot

sam-heilbron:
# Description

Ensure the requests to the admin server for the input snapshot include the Kubernetes Gateway resources, if they are available

# Context

## Previous Work
This PR builds off of the work that was introduced here: https://github.com/solo-io/gloo/pull/9517

## Interesting decisions
-

## Testing steps
Apply a Kubernetes Gateway resources
```
❯ kubectl apply -n gloo-system -f- <<EOF
kind: Gateway
metadata:
  name: http
spec:
  listeners:
  - protocol: HTTP
    port: 8080
    name: http
    allowedRoutes:
      namespaces:
        from: All
EOF
```

Apply a Kubernetes HTTPRoute
```
kubectl apply -f- <<EOF
kind: HTTPRoute
metadata:
  name: httpbin-cors
  namespace: httpbin
spec:
  parentRefs:
  - name: http
    namespace: gloo-system
  hostnames:
    - cors.example
  rules:
    - filters:
        - type: ExtensionRef
          extensionRef:
            group: gateway.solo.io
            kind: RouteOption
            name: cors
      backendRefs:
        - name: httpbin
          port: 8000
EOF
```

Port-Forward the Admin server
```
kubectl -n gloo-system port-forward deploy/gloo 9091
```

Query the `/snapshots/input`
```
```

- [x] Confirm the returned data includes the Kube Gateway resources

## Notes for reviewers
-

# Checklist:


projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -143,6 +162,41 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
 	return &clone
 }
 
+// getApiSnapshotSafe gets a clone of the latest ApiSnapshot
+func (h *historyImpl) appendKubeGatewayResources(ctx context.Context, resources map[string]interface{}) {
```
nfuden, line 166:
```suggestion
// appendKubeGatewayResources searches for routes and gateways
// setting the returned list onto the current resource map at a given key or replacing the contents with an error
func (h *historyImpl) appendKubeGatewayResources(ctx context.Context, resources map[string]interface{}) {
```

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -143,6 +162,41 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
...
+		resources[httpRoutesKey] = err
+	} else {
+		resources[httpRoutesKey] = routeList
+	}
+
+	var gwList apiv1.GatewayList
+	err = cli.List(ctx, &gwList)
+	gwKey := fmt.Sprintf("%s.%s", wellknown.GatewayGroup, wellknown.GatewayKind)
+	if err != nil {
+		resources[gwKey] = err
```
nfuden, line 187: is there any value in log out put of errors? 
What are the failure modes/cases for the cli.List?

Is it sufficient to have the errors as just part of the returned snap?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -143,6 +162,41 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
 	return &clone
 }
 
+// getApiSnapshotSafe gets a clone of the latest ApiSnapshot
+func (h *historyImpl) appendKubeGatewayResources(ctx context.Context, resources map[string]interface{}) {
+	cli := h.getKubeGatewayClientSafe()
+
+	if cli == nil {
```
nfuden, line 169: Non-blocking thought: Things like this are interesting. We probably should prioritize getting basic otel instrumentation into this code soon so that we have easy ways to see when we do noops or errors here. 

Namely if we were to instrument our http handlers with route tags like ```otelhttp.WithRouteTag```  such as as with https://opentelemetry.io/docs/languages/go/getting-started/

This is likely out of scope for this pr

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: My thought was that this API is intended to provide an accurate snapshot of the resources in the cluster. If an issue occurs gathering some of those resources, we shouldn't block returning the other data. Also, if we log errors, it creates 2 separate places that users would have to go to find information about the snapshot (ie the response has the data, and the logs have the error). 

I imagine as we make this API more robust we can find improvements to error handling, wdyt?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -143,6 +162,42 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
...
+	gwKey := fmt.Sprintf("%s.%s", wellknown.GatewayGroup, wellknown.GatewayKind)
+	if err != nil {
+		resources[gwKey] = err
+	} else {
+		resources[gwKey] = gwList
+	}
+}
+
+// getKubeGatewayClientSafe gets the Kubernetes client used for CRUD operations
+func (h *historyImpl) getKubeGatewayClientSafe() client.Client {
```
jbohanon, line 195: nit: this implies there is an _unsafe_ way to get the client. I think this can just be `getKubeGatewayClient`

jbohanon (approved):

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Wouldn't the unsafe way be to access the client without the lock? Since the client is set and get in separate goroutines right now

jbohanon: Yeah but neither the field nor the getter are exported, so I would think a field-level godoc comment noting that it should not be accessed by field access would be enough, but I have no problem with the extra defensiveness

projects/gloo/pkg/servers/iosnapshot/history.go
nfuden: works for me

nfuden (approved):

sam-heilbron:
bulldozer

PR #9575, opened 2024-06-04
Remove port-forwarding from Kubernetes E2E tests

jbohanon:
# Description

Port forwarding is not a pattern that is used in user deployments, but is a quick hack to get networking into a cluster. Since our intent with Kubernetes E2E tests is to imitate user patterns, we opt to eliminate the port forwards from the Kubernetes E2E tests altogether.

Note: for the specific use case of accessing envoy admin endpoints, this is not a particularly user-facing behavior anyway, so we have chosen to run these as a curl from within the proxy pod

The port-forwarding utility did not by default wait for the port-forward to be available before returning non-nil. This PR adds the dial logic to the utility.

# Checklist:


pkg/utils/kubeutils/portforward/portforwarder.go
```diff
@@ -12,9 +12,15 @@ import (
...
 type PortForwarder interface {
-	// Start runs this PortForwarder.
+	// Start will attempt to open a port forward using the retry options provided. This function
+	// makes no guarantees that the port forward will be accepting requests when returning nil.
+	// It is advised to use StartAndWaitForConn if not handling this in calling code.
 	Start(ctx context.Context, options ...retry.Option) error
 
+	// StartAndWaitForConn behaves similarly to Start, but will use the options passed in to
+	// additionally wait for the port forward to respond to a net.Dial.
+	StartAndWaitForConn(ctx context.Context, options ...retry.Option) error
```
sam-heilbron, line 22: Given that this is a safer function to use, do you see value in supporting Start as well? Any reason we don't just make this the Start function and not have 2 separate ones?

pkg/utils/kubeutils/portforward/portforwarder_test.go
```diff
@@ -0,0 +1,62 @@
...
+			return nil
+		}, time.Minute, time.Millisecond*100).ShouldNot(Succeed())
+	})
+
+	It("Creates a usable portforward with StartAndWaitForConn", func() {
+		err := pf.StartAndWaitForConn(ctx, defaultRetryOptions...)
+		Expect(err).ToNot(HaveOccurred())
+		_, err = net.Dial("tcp4", pf.Address())
+		Expect(err).ToNot(HaveOccurred())
+		pf.Close()
```
sam-heilbron, line 59: Should this be deferred in case the assertions short circuit the test?

pkg/utils/kubeutils/portforward/portforwarder_test.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	)
+
+	AfterEach(func() {
+		cancel()
+	})
+
+	BeforeEach(func() {
+		ctx, cancel = context.WithCancel(context.Background())
+		pf = NewPortForwarder(WithService("kube-dns", "kube-system"), WithPorts(30053, 53))
```
sam-heilbron, line 34: nit: I think comments here about why we're choosing the service/ports could be useful

pkg/utils/kubeutils/portforward/portforwarder_test.go
```diff
@@ -0,0 +1,62 @@
...
+	})
+
+	It("Creates a flaky portforward with Start", func() {
+		Eventually(func(g Gomega) error {
+			err := pf.Start(ctx, defaultRetryOptions...)
+			if err != nil {
+				return err
+			}
```
sam-heilbron, line 42:
```suggestion
g.Expect(err).NotTo(HaveOccurred())
```

pkg/utils/kubeutils/portforward/api_forwarder.go
```diff
@@ -54,12 +54,33 @@ type apiPortForwarder struct {
 	restConfig *rest.Config
 }
 
+// Start will attempt to open a port forward using the retry options provided. This function
+// makes no guarantees that the port forward will be accepting requests when returning nil.
+// It is advised to use StartAndWaitForConn if not handling this in calling code.
```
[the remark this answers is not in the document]
jbohanon: Not really but this way the change doesn't break solo-projects callers immediately, so it's a little safer/less painful if I'm not the one making the next gloo bump. IIRC s-p is where you ran into this months back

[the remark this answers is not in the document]
jbohanon: Reduced the changes to make the Start func safer in lieu of adding a new one

pkg/utils/kubeutils/portforward/portforwarder.go
jbohanon: https://github.com/solo-io/gloo/pull/9575#discussion_r1626357304

pkg/utils/kubeutils/portforward/portforwarder_test.go
jbohanon: I've removed this test since the flaky function no longer exists

sam-heilbron (approved):

jenshu (dismissed):

sam-heilbron (dismissed):

jbohanon:
work in progress label added to investigate CI-only failure in Kubernetes End-to-End cluster-two

Removal criteria:
- [ ] Revert changes in test selection
- [ ] Remove CI skips in changelog
- [ ] Actually passing tests

test/testutils/hack/go-utils/testutils/kubectl/kubectl.go
```diff
@@ -0,0 +1,193 @@
...
+				out = KubeLogs(ctx, label, params)
+				return eris.Errorf("%v in crash loop with logs %v", label, out)
+			}
+			if strings.Contains(out, "ErrImagePull") || strings.Contains(out, "ImagePullBackOff") {
+				out, _ = KubectlOut(ctx, Params{
+					Stdin:  params.Stdin,
+					Stdout: params.Stdout,
+					Stderr: params.Stderr,
+					Env:    params.Env,
+					Args:   []string{"describe", "pod", "-l", label},
```
jbohanon, line 175: add namespace

test/testutils/hack/go-utils/testutils/kubectl/kubectl.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			if finished(out) {
+				return nil
+			}
+		}
+	}
+}
+
+func KubeLogs(ctx context.Context, label string, params Params) string {
+	params.Args = []string{"logs", "-l", label}
```
jbohanon, line 187: add namespace

test/testutils/hack/go-utils/testutils/kubectl/kubectl.go
```diff
@@ -0,0 +1,193 @@
+package kubectl
```
sam-heilbron, line 1: I'm not a huge fan of placing a new "hack" directory into our testutils code. It feels like we're creating a potential to add code that will never leave. Why not add the utilities directly to the places we want, and clarify in the code that we would like to progressively migrate some of these components to a shared library when they are stable?

test/kubernetes/testutils/cmd/envoyadmin/envoyadmin.go
```diff
@@ -0,0 +1,305 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/solo-io/gloo/pkg/utils/cmdutils"
+	"github.com/solo-io/gloo/pkg/utils/envoyutils/admincli"
```
sam-heilbron, line 11: How is this different than the admincli in the envoyutils package? If this is a better place for that code, can we just move it all to live in the same place?

test/kubernetes/testutils/assertions/envoy.go
```diff
@@ -3,54 +3,73 @@ package assertions
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/solo-io/gloo/pkg/utils/envoyutils/admincli"
-	"github.com/solo-io/gloo/pkg/utils/kubeutils/portforward"
+	"github.com/solo-io/gloo/pkg/utils/kubeutils"
 	"github.com/solo-io/gloo/pkg/utils/requestutils/curl"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/cmd/envoyadmin"
 )
 
+type EnvoyAdminAssertion func(ctx context.Context, adminClient *envoyadmin.Client, clientRefresh func(ctx context.Context, adminClient *envoyadmin.Client, expectedReplicas int) *envoyadmin.Client)
```
sam-heilbron, line 17: I like the idea of dependency injection, though do you think the `clientRefresh` will have multiple implementations? If anything, it feels like an implementation detail that only exists because of a subtle behavior. It feels like we'd be better suited in fact to "hide" this, ie not expose it to users, and require they use the single implementation. wdyt?

test/kubernetes/testutils/cmd/remote.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Run runs the command
+// If the returned error is non-nil, it should be of type *RunError
+func (cmd *RemoteCmd) Run() *cmdutils.RunError {
+	var combinedOutput threadsafe.Buffer
+
+	cmd.Stdout = io.MultiWriter(cmd.Stdout, &combinedOutput)
+	cmd.Stderr = io.MultiWriter(cmd.Stderr, &combinedOutput)
+
+	_, err := kube.ExecFromEphemeralPod(context.Background(), cmd.EphemeralPodParams)
```
sam-heilbron, line 112: I'm not sure I follow this entirely. What is the value in making a kube specific command, versus just invoking the kube specific method on the client? 

The original intention of the LocalCmd, was to make it so you could have a shared implementation of cmd.Exec, and clients could use that under the hood. It feels like we've inverted that relationship and now calling code must know a command is being used, and isn't aware that a kube.ExecFromEphemeralPod is being called

test/testutils/hack/k8s-utils/testutils/kube/curl.go
```diff
@@ -0,0 +1,168 @@
...
+	DefaultCurlImage = Image{
+		Registry:   "curlimages",
+		Repository: "curl",
+		Tag:        "",
+		Digest:     "aa45e9d93122a3cfdf8d7de272e2798ea63733eeee6d06bd2ee4f2f8c4027d7c",
+		PullPolicy: "IfNotPresent",
+	}
+)
+
+func CurlWithEphemeralPod(ctx context.Context, logger io.Writer, kubeContext, fromns, frompod string, args ...string) string {
```
sam-heilbron, line 26: Building from my comment about the introduction of the `hack` directory, I think if we want to introduce this into the codebase, we can/should just place this on the kube.Cli that we created and perhaps clarify it was copy/pasted from a go-utils@commit-sha

sam-heilbron (commented):
nice! I haven't looked too in depth at the implementation, but I shared some high level thoughts. Could we make sure the following is completed:
- This sentiment around port-forwarding should be captured in our e2e readme
- The PR body should clarify why we believe this will reduce flakes and what approvers can do to validate that
- There is a lot of new code, how will we prevent developers from calling the wrong code at hte wrong time?

test/testutils/hack/k8s-utils/testutils/kube/curl.go
```diff
@@ -0,0 +1,168 @@
...
+func WaitForRollout(ctx context.Context, logger io.Writer, kubeContext, ns, deployment string) {
+	args := []string{"-n", ns, "rollout", "status", "deployment", deployment}
+	mustExecute(ctx, KubectlParams{
+		KubectlCmdParams: kubectl.NewParams(args...),
+		KubeContext:      kubeContext,
+		Logger:           logger,
+	})
+}
+
+func Curl(ctx context.Context, logger io.Writer, kubeContext, ns, fromDeployment, fromContainer, url string) string {
```
bewebi, line 60: Do we want to add the ability to pass additional curl args?

bewebi (commented):
Overall makes sense and looks good, just a few nits/questions

test/kubernetes/testutils/assertions/envoy.go
jbohanon: The biggest thing that needs to be injected on a per-case basis is the required number of replicas. Since the status reporting for pods that have had a termination event raised against them by the kubelet is garbage (no way to tell without querying events if a pod is "terminating"; it still shows as ready), the way I figured out how to select a truly ready pod was to wait for the number of replicas for the deployment to stabilize on a known value. I agree that it's pretty smelly though; maybe I could look at passing the expected replicas into the assertion.

Though as I am typing this up I remember that the main reason it was injected was so that the asstertion could call the refresh _within_ its Eventually block. Keeping the client stabilization as close to the call that needs a stable client as possible. 

It is truly awful that a pod in Terminating state does not have a status condition stating such

jbohanon: According to https://github.com/kubernetes/kubernetes/issues/102537 this is expected behavior

test/testutils/hack/go-utils/testutils/kubectl/kubectl.go
jbohanon: I feel that having a `hack` directory in the code is a huge smell that creates an underlying tension indicating this shouldn't be here. I would think that this code would be more likely to assimilate itself into the codebase and never end up being extracted back out if we interspersed it with permanent code. I would rather deal with the pain of inter-repo PRs than risk a bespoke `Kubectl` function staying in this repo for no reason other than it hasn't been looked at in months.

In either case I think an issue should be opened to track migrating the code back to the shared util.

test/kubernetes/testutils/cmd/remote.go
jbohanon: I can't argue original intent since you wrote the code, but the shape of the code implies to me that `cmdutils.Cmd` was abstracting _away_ the implementation of `exec.Cmd` as `LocalCmd`. The very existence of `LocalCmd` to me implies a `RemoteCmd` implementation was expected eventually. 

I think the benefit of this abstraction is that devs can use context-specific `Cmd`s in ways that have similar ergonomics. If you're running in-memory e2e or unit tests, a `LocalCmd` is what you need, while if you're running in Kube, a `RemoteCmd` is what you need. The latter requires a slight bit more setup on the `Cmder` but other than that, they're wielded in the same way with regard to setting up the input/output streams and calling `Run`

jbohanon: > What is the value in making a kube specific command, versus just invoking the kube specific method on the client?

A `RemoteCmd`, just like a `LocalCmd`, will I expect be primarily used _by_ clients. This code primarily enables the remote envoy admin client in `test/kubernetes/testutils/cmd/envoyadmin/envoyadmin.go` and the helper wrapper for a remote curl in `test/kubernetes/testutils/cmd/curl.go`, which to me is analogous to the `LocalCmd` being used by the `admincli` and `curl` helper clients

test/kubernetes/testutils/cmd/envoyadmin/envoyadmin.go
jbohanon: This could probably use some rebranding/relocation to indicate that it is a utility for accessing the admin client from inside the cluster. I could possibly refactor the admincli to take a `Cmder` to avoid so much code duplication but that seemed like an even deeper bite than I had already committed to

test/testutils/hack/go-utils/testutils/kubectl/kubectl.go
```diff
@@ -0,0 +1,193 @@
...
+// WaitPodsRunning waits for all pods to be running
+func WaitPodsRunning(ctx context.Context, interval time.Duration, namespace string, params Params, labels ...string) error {
+	finished := func(output string) bool {
+		return strings.Contains(output, "Running") || strings.Contains(output, "ContainerCreating")
+	}
+	for _, label := range labels {
+		if err := WaitPodStatus(ctx, interval, namespace, label, "Running or ContainerCreating", finished, params); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jbohanon: I can't say for certain, and I'm not even sure if the reason this was originally done is still relevant, but this was a deliberate change made >5 years ago by Marco in https://github.com/solo-io/go-utils/pull/65/files#diff-194e17bf80495c27c08438d8dda9585f21023dfeb92a8910adb4294d58d01b39L103. I'd tag him to ask why, but it was over 5 years ago, so it's unlikely he remembers I think. If you think it's worth pursuing though I can

jbohanon:
This PR is being placed on hold in favor of a less intrusive fix to the linked issue. The fundamental issue with the test that was flaking is that it is possible to select a pod in `Terminating` state for port-forward. Adding retries to the port-forward and making sure that we aren't selecting a pod in `Terminating` status should be sufficient to solve the stated problem.

PR #9576, opened 2024-06-04
convert RtOpt, VHostOpt to targetRefs

lgadban:
# Description

Convert `RouteOption` and `VirtualHostOption` policy objects to use a repeated list of targetRefs.
This is a future-looking change to support the evolution of the targetRef API in upstream GwAPI and prevent us from needing to break/augment this API right after releasing.
Note that only the first targetRef is actually used for now, the implementation is tracked in https://github.com/solo-io/solo-projects/issues/6286 (private issue for now unfortunately)

## API changes
`RouteOption` and `VirtualHostOption` now have `targetRefs` instead of `targetRef`


EItanya (approved):
👍

projects/gateway2/translator/plugins/virtualhostoptions/query/indexers.go
```diff
@@ -25,7 +25,15 @@ func virtualHostOptionTargetRefIndexer(obj client.Object) []string {
 	}
 
 	var res []string
-	targetRef := vhOpt.Spec.GetTargetRef()
+	targetRefs := vhOpt.Spec.GetTargetRefs()
+	if len(targetRefs) == 0 {
+		return res
+	}
+
+	// use the first targetRef in the list as we only support one ref but have multiple in API for future-compatbility
```
sam-heilbron, line 33: nit: It feels a bit strange to me that we separate this complexity around handling a single item from a list in 2 places:
- We pick the first item here
- We add a warning log inside of the query.go

is there a reason we can't consolidate those two pieces into a single place?

changelog/v1.17.0-beta36/route-vhost-opts-targetrefs.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+  - type: NON_USER_FACING
```
sam-heilbron, line 2: Are we just considering this non-user-facing because functionally the API behaves the same, and the original API was never released in a GA version?

sam-heilbron (approved):
LGTM. One question about the way that we handle warning users, but non-blocking

lgadban (commented):
@sam-heilbron I responded to both convos. As you mentioned they aren't blocking I will resolve them to get this merged. Let me know if you want to discuss any of them further

PR #9578, opened 2024-06-05
Add typed error for ConfigNotFound in plugins

npolshakova:
# Description

Control Plane Validation Webhook does respecting the eventually consistent nature of Kubernetes resources. This is because there is currently not any way for plugins to return typed errors that the main engine could decide on how to handle based on the error type. 

For example, when the [VirtualHost plugins run](https://github.com/solo-io/gloo/blob/d965ab8502a47dc2d1e57c97afa716ce79887a5a/projects/gloo/pkg/translator/route_config.go#L134)- we return an error regardless of the error type:
```
// run the plugins
	for _, plugin := range h.pluginRegistry.GetVirtualHostPlugins() {
		if err := plugin.ProcessVirtualHost(params, virtualHost, out); err != nil {
			validation.AppendVirtualHostError(
				vhostReport,
				validationapi.VirtualHostReport_Error_ProcessingError,
				fmt.Sprintf("invalid virtual host [%s]: %v", virtualHost.GetName(), err.Error()),
			)
		}
	}
```

## API changes
 None

## Code changes
- Adds new error type `ValidationError`

## CI changes
None

## Docs changes
None

# Context

[Design doc](https://soloio.slab.com/posts/control-plane-validation-webhook-does-not-respect-eventual-consistency-7bczzuy7)

## Interesting decisions
 None

## Testing steps
Tested in solo-projects with ratelimit changes: https://github.com/solo-io/solo-projects/pull/6308 

## Notes for reviewers
Only `could not find RateLimitConfig resource ` will be addressed in this first set of changes 

Follow up work will include:
- Adding support for ProcessRouteAction, ProcessWeightedDestination
- Add support for other validation warnings (Delegation, Headers plugin secret does not exist, etc.)

# Checklist:


projects/gloo/pkg/translator/route_config.go
```diff
@@ -164,6 +164,15 @@ func (h *httpRouteConfigurationTranslator) computeVirtualHost(
 	// run the plugins
 	for _, plugin := range h.pluginRegistry.GetVirtualHostPlugins() {
 		if err := plugin.ProcessVirtualHost(params, virtualHost, out); err != nil {
+			// Check if the error is a ValidationError with a ConfigurationWarning
+			if verr, ok := err.(plugins.ValidationError); ok {
+				if verr.ConfigurationWarning() {
+					// TODO(npolshak): Create a VirtualHostReport_Warning report to add a warning: https://github.com/solo-io/gloo/issues/7357
```
sam-heilbron, line 170: I'm confused, I thought the purpose of this change was to solve this specifically, that is to turn this configuration state into a warning that is reflected on the resource?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -164,6 +164,15 @@ func (h *httpRouteConfigurationTranslator) computeVirtualHost(
...
+					contextutils.LoggerFrom(params.Ctx).Warnf("virtual host plugin %s produced a warning: %v", plugin.Name(), err)
+					continue
+				}
+			}
+
```
sam-heilbron, line 175: In the code below this line, we capture metadata from the VirtualHost to include in the report. Should that also exist for the warning?

projects/gloo/pkg/translator/route_config.go
npolshakova: Do we want to add a new [VirtualHostReport](https://github.com/solo-io/gloo/blob/76a49fddacf8a7d26d4bf8dd3b21525a8efe73bd/projects/gloo/api/grpc/validation/gloo_validation.proto#L160C9-L160C26) Warning type in the API?

The way we're handling this for route reports has a specific warning type (ex. InvalidDestinationWarning):
```
    // warning types for the given route config
    message Warning {
        enum Type {
            InvalidDestinationWarning = 0;
        }

        // the type of the error
        Type type = 1;
        // any extra info as a string
        string reason = 2;
    }
```

Do we want a ConfigNotFound warning type. Or do we want to avoid making changes to the API?

projects/gloo/pkg/translator/route_config.go
npolshakova: It depends if we just want to log the warning and unblock the webhook from rejecting the config OR if we want to make an API change and create a new Warning field on the VirtualHostReport.

sam-heilbron:
This PR is not necessary because https://github.com/solo-io/gloo/pull/9582 incorporates the changes

PR #9581, opened 2024-06-05
Refactor Kubernetes E2E test definition for ease of use

jbohanon:
## Description

Introduces abstractions over the test cases for Ordered and Unordered sets of tests within a suite. This allows us to define the run behavior of the tests separate from the suite that runs them.

Also refactors the tests out into an importable `tests` package (`*_test` packages cannot be imported). This will allow us to run the tests in the same way in EE, and also make sure the test suites in OSS and EE stay in lockstep.

Any tests which did not have a comment about needing to be ordered were assumed to be unordered. There is a non-zero risk that this introduces some flakes if the test order was in fact important.

## Context

Tests added to these suites in OSS had to be added in the same manner in EE. By exporting the entire pack of tests to run, the EE suites can import them to run on their bespoke `e2e.TestInstallation`.

## Code changes
- New types and interface to abstract the concept of Ordered and Unordered groups of tests
- Tests actually run in the suite were separated from the suite. This adds a little unnecessary code distance, but in order to import them they could not be in a `*_test` package and it seemed cleaner to me (and possibly necessary for testify but not sure) to leave the suites in `*_test` package.

## Testing steps

I manually verified behavior by opening https://github.com/solo-io/solo-projects/pull/6314 and running nightlies off of that branch to see that the expected suite ran (at that time just the k8s_gw_tests had been migrated).

## Interesting decisions

In order to reduce verbosity of defining a new test, we strip back some of the injection to the actual `Run` function. This means that we cannot need to reference the three things we need injected (`ctx, t, testInstallation`) when we are registering tests. To reconcile with this, what is registered is actually the feature suite construction function. We have enforced this by defining a new typedef for that signature and making a compiler assertion at the top of each feature suite for its construction function.

A side-effect of this is that nested test structures such as used for Glooctl in `k8s_gw_test.go` no longer work since we don't have access to the injected values at registration time. To resolve this, nested test structures must be defined as suites. An example of this can be found in the same Glooctl test, now located in `k8s_gw_tests.go`.

# Checklist:


test/kubernetes/e2e/tests/automtls_istio_edge_api_tests_to_run.go
```diff
@@ -0,0 +1,29 @@
+package tests
```
shashankram, line 1: Can we omit "to_run" from the filenames? Seems silly as tests are meant to be run implicitly.

test/kubernetes/e2e/tests/automtls_istio_edge_api_tests_to_run.go
```diff
@@ -0,0 +1,29 @@
...
+
+func AutomtlsIstioEdgeApiTests() TestRunner { return automtlsIstioEdgeApiTestsToRun }
+
+var (
+	automtlsIstioEdgeApiTestsToRun = UnorderedTests{
+		"HeadlessSvc": func(ctx context.Context, t *testing.T, testInstallation *e2e.TestInstallation) func(t *testing.T) {
+			return func(t *testing.T) {
+				suite.Run(t, headless_svc.NewEdgeGatewayHeadlessSvcSuite(ctx, testInstallation))
+			}
+		},
```
shashankram, line 22: This adds quite a bit of cognitive overhead. Why can't they be regular `t.Run()` blocks as before?

jbohanon: the generator is returning the func that was run before, it's just wrapped in a way that allows us to inject the parameters that are different at runtime between OSS and EE (ctx, t, testInstallation)

test/kubernetes/e2e/tests/types.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+)
+
+type (
+	NamedTest struct {
```
jbohanon, line 12: nit: this probably doesn't need to be exported. `OrderedTests`, `UnorderedTests`, and `TestRunner` are exported to allow similar definition pattern in EE

test/kubernetes/e2e/tests/glooctl_istio_inject_tests.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/solo-io/gloo/test/kubernetes/e2e/features/glooctl"
+	"github.com/solo-io/gloo/test/kubernetes/e2e/features/istio"
+)
+
+func GlooctlIstioInjectTests() TestRunner {
+	// NOTE: Order of tests is important here because the tests are dependent on each other (e.g. the inject test must run before the istio test)
+	// glooctlIstioInjectTests := make(OrderedTests, 0, 3)
+	// glooctlIstioInjectTests := make(OrderedTests, 3)
```
jbohanon, line 11: delete

jbohanon, line 11:
```suggestion
```

jbohanon:
Had to kick cluster one due to this error which looks like the validation flakiness being worked on elsewhere in this repo now

```log
Error from server: error when creating "/home/runner/work/gloo/gloo/test/kubernetes/e2e/features/glooctl/testdata/edge-gateway-gateways.yaml": admission webhook "gloo.k8s-gw-test.svc" denied the request: resource incompatible with current Gloo snapshot: [Validating *v1.Gateway failed: 1 error occurred:
	* Validating *v1.Gateway failed: validating *v1.Gateway name:"gateway1"  namespace:"k8s-gw-test": 1 error occurred:
	* could not render proxy: 2 errors occurred:
	* invalid resource k8s-gw-test.gateway1
	* invalid virtual service ref name:"vs1"  namespace:"default"
```

test/kubernetes/e2e/tests/types.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+)
+
+var (
+	_ TestRunner = new(OrderedTests)
+	_ TestRunner = new(UnorderedTests)
+)
+
+func (o OrderedTests) Run(ctx context.Context, t *testing.T, testInstallation *e2e.TestInstallation) {
+	for _, namedTest := range o.tests {
```
sam-heilbron, line 37: How will this be impacted when we run tests with the `-shuffle` flag?

test/kubernetes/e2e/tests/types.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	"github.com/stretchr/testify/suite"
+)
+
+type (
+	NamedTest struct {
```
sam-heilbron, line 12: For exported types, I think at a minimum we need a docstring on the type and ideally a write up in the readme, following https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/README.md#testcluster

test/kubernetes/e2e/tests/types.go
```diff
@@ -0,0 +1,70 @@
...
+		Name     string
+		NewSuite e2e.NewSuiteFunc
```
sam-heilbron, line 14: nit: If a NamedTest is a combination of a Name and a suite func, it feels like this would be more aptly referred to as a NamedSuite? Better yet, should names just be baked into the Suite itself?

sam-heilbron (commented):
nice work!

test/kubernetes/e2e/tests/types.go
jbohanon: It's an interesting question. I tested this multiple times by running `TEST_PKG=./test/kubernetes/e2e/tests/k8s_gw_test.go GO_TEST_USER_ARGS='-test.run "^TestK8sGateway$$/^Glooctl$$" -test.v -test.shuffle=on' make go-test` against this branch multiple times and against the main branch, and each run executed all of the tests in the same order each time. I found [this PR](https://github.com/stretchr/testify/pull/1413) in the stretchr/testify repo, so it looks like shuffle is just not supported by testify at this time.

test/kubernetes/e2e/tests/types.go
jbohanon: I moved this code, exported a lot less, and added a writeup in the readme.

test/kubernetes/e2e/tests/types.go
jbohanon: I think it is useful to allow different TestInstallations to name suites they are registering differently, since the context they are running under is baked into the `_tests.go` they are being registered within.

sam-heilbron (approved):
great work!

jenshu (approved):

jbohanon:
Dozer?

PR #9594, opened 2024-06-07
Lambda support

ilackarms:
# Description
adds support for lambda and azure function calls to GGv2
BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6160


projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -402,3 +415,43 @@ func setRouteAction(
...
+			continue
+		}
+		if filter.ExtensionRef.Group != parameters.ParameterGroup {
+			continue
+		}
+		if filter.ExtensionRef.Kind != parameters.ParameterKind {
+			continue
+		}
+		sectionName := string(filter.ExtensionRef.Name)
+		switch upstream.Spec.GetUpstreamType().(type) {
```
sam-heilbron, line 436: Can we add a default case here, instead of dropping it? I think it would be useful to provide some feedback about the behavior of cases that aren't supported. Would this be reported as a status on the HtttpRoute? Or at a minimum a log line indicating that the UpstreamType they are using doesn't yet exist?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -402,3 +415,43 @@ func setRouteAction(
 
 	return delegates
 }
+
+// makeDestinationSpec computes the destination spec for a given upstream based on the type of upstream and the Filters from the backend reference
+func makeDestinationSpec(ctx context.Context, upstream *gloov1.Upstream, filters []gwv1.HTTPRouteFilter) *v1.DestinationSpec {
```
sam-heilbron, line 420: What happens in the case that the destinationSpec is returned nil? Do we need to provide any feedback to the user as it could indicate they have misconfigured something?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -367,6 +379,7 @@ func setRouteAction(
 							Namespace: ns,
 						},
 					},
+					DestinationSpec: spec,
```
sam-heilbron, line 382: Given that we add DesitnationSpec to the Upstream case, and it doesn't exist for other cases, would it be worthwhile to add a comment for those cases indicating why we explicitly don't support it?

projects/gateway2/parameters/parameters.go
```diff
@@ -0,0 +1,6 @@
+package parameters
+
+const (
+	ParameterGroup = "gloo.solo.io"
+	ParameterKind  = "Parameter"
```
EItanya, line 5: Should we call this `FunctionName`?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -359,6 +364,13 @@ func setRouteAction(
 			})
 
 		case backendref.RefIsUpstream(backendRef.BackendObjectReference):
```
EItanya, line 366: Why aren't we just checking the type of the `obj`?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
EItanya: `destinationSpec` is `nil` most of the time. When there's a function type it shouldn't be so maybe it's worth at least a debug log for now if that happens

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -359,6 +364,13 @@ func setRouteAction(
...
+			upstream, ok := obj.(*gloov1.Upstream)
+			if !ok {
+				// should never happen
+				contextutils.LoggerFrom(ctx).Errorf("expected upstream, got %T", obj)
```
nfuden, line 370: this seems insufficient should it not be a full dump of the obj?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -402,3 +415,43 @@ func setRouteAction(
...
 }
+
+// makeDestinationSpec computes the destination spec for a given upstream based on the type of upstream and the Filters from the backend reference
+func makeDestinationSpec(ctx context.Context, upstream *gloov1.Upstream, filters []gwv1.HTTPRouteFilter) *v1.DestinationSpec {
+	for _, filter := range filters {
+		// only look for 'parameters' extensionref filters
+		if filter.Type != gwv1.HTTPRouteFilterExtensionRef {
+			continue
+		}
+		if filter.ExtensionRef == nil {
```
nfuden, line 426: Why are these unrolled? can we have a comment that better explains this series of checks?

projects/gateway2/translator/testutils/inputs/http-with-lambda-destination/manifests.yaml
```diff
@@ -0,0 +1,46 @@
...
+                name: uppercase
+---
+apiVersion: gloo.solo.io/v1
+kind: Upstream
+metadata:
+  name: aws-upstream
+spec:
+  aws:
+    region: us-east-1
+    roleArn: arn:aws:iam::123456789012:role/lambda-role
```
nfuden, line 43: if we are hardcoding aws can we pull out this variable somewhere?

projects/gateway2/translator/gateway_translator_test.go
```diff
@@ -91,6 +91,52 @@ var _ = Describe("GatewayTranslator", func() {
...
+			},
+		}.Run(ctx)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(results).To(HaveLen(1))
+		Expect(results[types.NamespacedName{
+			Namespace: "default",
+			Name:      "gw",
+		}]).To(BeTrue())
+	})
```
nfuden, line 115: Should we be checking for more than happy path here? seems like we are once again doing the most lightweight testing we can

projects/gateway2/parameters/parameters.go
ilackarms: I left it open ended because theoretically this could be any string parameter for a backend, not just function name

projects/gateway2/translator/httproute/gateway_http_route_translator.go
ilackarms: that's a good point, i'm just following the established pattern here
i agree though

projects/gateway2/parameters/parameters.go
EItanya: Wouldn't we want to name those on a case by case basis so it's more clear?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
ilackarms: not sure if we need all that, the type is what we care about
no strong opinion though since it should not ever happen

projects/gateway2/translator/testutils/inputs/http-with-lambda-destination/manifests.yaml
ilackarms: it's a fake role, just used to test translation

ilackarms: technically this part isn't even translated i just copied it in, shouldn't need a vraible

projects/gateway2/parameters/parameters.go
ilackarms: Supporting different `kinds` creates more opportunity for a user to make an error, more conceptual overload, more burden on reading docs. The `kind` is already specified on the `backendRef`, and our system is smart enough to discern the difference. 

What we really want is `sectionName`, which does not have an actual kind associated with it. I fear if we start adding kinds here we will be creating an unbounded API we don't even need, let alone want to support.

projects/gateway2/translator/httproute/gateway_http_route_translator.go
ilackarms: only Upstreams support function calls in Gloo, do you feel this is something worth commenting in the code here?

projects/gateway2/translator/gateway_translator_test.go
ilackarms: added some package level unit tests for `makeDestinationSpec` since it contains all the new logic under test

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -22,6 +28,12 @@ import (
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/core/matchers"
 )
 
+var (
+	awsMissingFuncRefError                = eris.Errorf(" upstreams must have a logical name specified in the backend ref via the parameters extensionref")
+	azureMissingFuncRefError              = eris.Errorf(" upstreams must have a function name specified in the backend ref via the parameters extensionref")
```
EItanya, line 33:
```suggestion
	awsMissingFuncRefError                = eris.New("upstreams must have a logical name specified in the backend ref via the parameters extensionref")
	azureMissingFuncRefError              = eris.New("upstreams must have a function name specified in the backend ref via the parameters extensionref")
```

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -22,6 +28,12 @@ import (
...
+	nonFunctionUpstreamWithParameterError = eris.Errorf("parameters extensionref is only supported for aws and azure upstreams")
```
EItanya, line 34:
```suggestion
	nonFunctionUpstreamWithParameterError = eris.New("parameters extensionref is only supported for aws and azure upstreams")
```

changelog/v1.17.0-rc3/ggv2-function-routing.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/solo-projects/issues/6160
+    description: |
+      Enables routing to AWS Lambda and Azure Function upstreams via the GGv2 API.
```
sam-heilbron, line 5: nit: We're really trying to steer away from "v2". This is the Gloo Gateway product, but it's really using the AWS Lambda and Azure Function upstreams on HttpRoutes, right?

sam-heilbron (approved):
LGTM! I like Eitan's proposal for the error changes

nfuden:
/kick gcloud flake ERROR: (gcloud.builds.submit) The required property [project] is not currently set.

projects/gateway2/translator/httproute/gateway_http_route_translator.go
nfuden: Yeah but when things that never should happen do happen we should make sure we can diagnose what object made the thing happen in the way that it never should have

sam-heilbron (approved):

EItanya (approved):

sam-heilbron:
bulldozer

PR #9612, opened 2024-06-12
bump dependencies

stevenctl:
There is no rush on merging this and I have no idea what the side-effects might be.


stevenctl:
Getting an error in a test, maybe a flake. 

https://github.com/solo-io/solo-kit/blob/1baf6de465942dc5be44e7f28f0f739dcd0b967b/pkg/api/shared/status.go#L100 seems to fail to lookup the namespace from an env

```
1 error occurred:\n\t* failed to write status state:Accepted reported_by:\"gloo\" for resource gateway-proxy: error getting status json patch data: getting status reporter namespace: POD_NAMESPACE environment variable is not set
```

Is this a test that might race between setting that env in a non-kube env?

PR #9616, opened 2024-06-12
docker: apt-get settings to be more robust against bad mirrors

nfuden:
# Description

apt-get is used in several place within our code base.
Today we do not choose a mirror that is particularly fast. This means that when a distro changes to a different source location such as focal now pulling from archive.ubuntu.com we are at risk for the slow connections. In order to not timeout mid build and fail what ever pipeline we are in try setting some larger resiliency guard rails as well as set some basic common sense checks against installing unneeded packages.


sam-heilbron:
What's the plan for this PR? There have been 19 commits since the last change, should we remove the 'keep pr updated' label?

PR #9617, opened 2024-06-13
test: print stdout and stderr of Curl request

shashankram:
Prints the stdout and stderr of the curl command executed by the test. This is required to debug test failures when the curl request fails.

### DIFF
--- a/test/kubernetes/testutils/assertions/curl.go
+++ b/test/kubernetes/testutils/assertions/curl.go
@@ -36,14 +36,14 @@ func (p *Provider) AssertEventualCurlReturnResponse(
 	var curlHttpResponse *http.Response
 	p.Gomega.Eventually(func(g Gomega) {
 		curlResponse, err := p.clusterContext.Cli.CurlFromPod(ctx, podOpts, curlOptions...)
+		fmt.Printf("want:\n%+v\nstdout:\n%s\nstderr:%s\n\n", expectedResponse, curlResponse.StdOut, curlResponse.StdErr)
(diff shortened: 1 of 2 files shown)


jenshu (approved):

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -36,14 +36,14 @@ func (p *Provider) AssertEventualCurlReturnResponse(
 	var curlHttpResponse *http.Response
 	p.Gomega.Eventually(func(g Gomega) {
 		curlResponse, err := p.clusterContext.Cli.CurlFromPod(ctx, podOpts, curlOptions...)
+		fmt.Printf("want:\n%+v\nstdout:\n%s\nstderr:%s\n\n", expectedResponse, curlResponse.StdOut, curlResponse.StdErr)
```
sam-heilbron, line 39: Do you think it makes sense to just log directly in the `CurlFromPod` method _if_  verbose output is used (which it is in that implementation)?

sam-heilbron (approved):

test/kubernetes/testutils/assertions/curl.go
lgadban: does this mean verbose output as a curl option?

shashankram: The only reason I chose to log this here is because we may decide to customize how we print the want vs output returned. Currently it is naive just to unblock debugging flakes. Also would be a bit weird to log this in a lib func and callers might log it again anyway for customization like this, so we may end up with dups.

Issue #9353, opened 2024-04-11
port forward util returns before port forward is available for use


jbohanon:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.17.0

### Kubernetes Version

All

### Describe the bug

The utility to create a port forward, when started, does not wait until the port forward is available for use. This can cause a race condition in tests which expect to be able to make requests against the forwarded port immediately upon `Start` returning.

### Expected Behavior

`Start` should not return without an error until the port is reachable

### Steps to reproduce the bug

Set up a test that creates a port forward using 
```go
portFwd := portforward.NewPortForwarder(
	portforward.WithDeployment(proxyName, namespace),
	portforward.WithPorts(19000, 19000),
)
err = portFwd.Start(ctx)
Expect(err).NotTo(HaveOccurred())
defer func() {
	portFwd.Close()
	portFwd.WaitForStop()
}()
```
then immediately try a request against the forwarded port. It is possible that it will work appropriately, but it will be flaky at best since the port is not guaranteed to be available yet

### Additional Context

https://github.com/solo-io/gloo/blob/3e00d8140f91fe0111955bb46fbc29df8008bf47/pkg/utils/kubeutils/portforward/cli_portforwarder.go#L40

https://github.com/solo-io/solo-projects/pull/5973/files#r1560360903

sheidkamp:
Still relevant

PR #9628, opened 2024-06-17
Ensure port forwards do not select terminating pods and are ready before use

jbohanon:
# Description

This PR aims to fix flakiness in testing code which relies on port-forwards.

## Code changes
- Check that pods are not in terminating status before attempting to port-forward.
  - The cli portforwarder does this by using a util to make sure no pods show Terminating from a kubectl get
  - The api portforwarder does this by making sure the pod resource doesn't have a deletionTimestamp

# Context

Port forwards in Gloo have been historically flaky. There are two main root causes for this that were uncovered in this investigation:
- The function that starts the port-forward could return nil (no error) before the port-forward was fully initialized. This would lead to errors when the calling code tried to dial on the forwarded address immediately after receiving the nil return.
- The pod selection code allowed for pods in Terminating status to be selected as the target for the port-forward. This would lead to errors when the selected Terminating pod was finally killed and the port-forward failed.

## Testing steps

TODO

# Checklist:


jbohanon:
Closing as the proposed changes have caused timeouts in kubernetes e2e tests, and this has been deprioritized due to the relative pain of the flake being low.

The work to get this done is still being tracked by https://github.com/solo-io/gloo/issues/9353

PR #9629, opened 2024-06-17
prevent Proxy namespace being from modified

lgadban:
# Description

Previously, a 'metaKey' was generated for each Proxy being translated and synced to envoy.
This key was used as a 'tag' for tracing of the
translation/sync prcoess. It used a custom function that was almost exactly
matched how the SnapshotCacheKey used for xDS was generated; this function
also modified the actual Proxy's namespace, which broke assumptions of where
the proxy lives in the in-memory client.

By removing the namespace modification we fix various status error logs as
now the Proxy is located where the rest of the system expects (the write
namespace). Also, we can simply use the SnapshotCacheKey(...) to generate
a key with the same logic used for xds.

## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/projects/gloo/pkg/syncer/translator_syncer_test.go
+++ b/projects/gloo/pkg/syncer/translator_syncer_test.go
@@ -7,8 +7,6 @@ import (
 	"github.com/solo-io/gloo/projects/gloo/pkg/servers/iosnapshot"
 
 	"github.com/solo-io/gloo/pkg/bootstrap/leaderelector/singlereplica"
-	gloov1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
-	"github.com/solo-io/gloo/projects/gloo/pkg/utils"
 
 	gloo_translator "github.com/solo-io/gloo/projects/gloo/pkg/translator"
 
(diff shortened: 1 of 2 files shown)


jenshu (approved):

projects/gloo/pkg/syncer/translator_syncer_test.go
```diff
@@ -344,68 +342,6 @@ var _ = Describe("Translate multiple proxies with errors", func() {
 
 		Expect(xdsCache.Called).To(BeTrue())
 	})
-
-	Describe("correctl returns metaKey for proxies", func() {
```
npolshakova, line 348: Do we want to replace these tests for `SnapshotCacheKey`?

npolshakova (approved):

projects/gloo/pkg/syncer/translator_syncer_test.go
lgadban: I don't believe so, as the key building is tested by the xds snapshot cache key tests.
I found nothing that checks/tests the tags used in the tracing capabilities, and frankly it doesn't seem used in general

PR #9632, opened 2024-06-17
helm: support webhook.enablePolicyApi

sam-heilbron:
# Description

Introduce new Helm API: `gateway.validation.webhook.enablePolicyApi`

# Context
See https://github.com/solo-io/solo-projects/issues/6321 and https://github.com/solo-io/solo-projects/issues/6251 for context around a race condition that exists in the validation webhook. 

Due to this race, we need a way for users to explore and use the Kubernetes Gateway Integration, but also opt-out of validation (in case the race condition is preventing resources from being updated)

## Interesting decisions
 -

## Testing steps
- I added new unit tests for our Helm chart.
```
 TEST_PKG=install/test make test
```

## Notes for reviewers
-

# Checklist:


install/helm/gloo/generate/values.go
```diff
@@ -450,6 +450,12 @@ type Webhook struct {
...
 	ExtraAnnotations              map[string]string `json:"extraAnnotations,omitempty" desc:"extra annotations to add to the webhook"`
 	SkipDeleteValidationResources []string          `json:"skipDeleteValidationResources,omitempty" desc:"resource types in this list will not use webhook valdaition for DELETEs. Use '*' to skip validation for all resources. Valid values are 'virtualservices', 'routetables','upstreams', 'secrets', 'ratelimitconfigs', and '*'. Invalid values will be accepted but will not be used."`
+	// EnablePolicyApi provides granular access to users to opt-out of Policy validation
+	// There are some known race conditions in our Gloo Gateway processes resource references,
+	// even when allowWarnings=true: https://github.com/solo-io/solo-projects/issues/6321
+	// As a result, this is intended as a short-term solution to provide users a way to opt-out of Policy API validation.
+	// The desired long-term strategy is that our validation logic is stable, and users can leverage it
+	EnablePolicyApi *bool `json:"enablePolicyApi,omitempty" desc:"enable validation of Policy Api resources (RouteOptions, VirtualHostOptions) (default: true)"`
```
nfuden, line 458: Is there a reason we are not extending our granular approach to validation. Ie building in the ability to set this type of escape hatch for a single one of the policies? 

Validation has historically been a big enough issue that we have increasingly delegated to having configuration at the per resource type level.

Should this be a whole new "skip resource" and advise users to configure both routeoptions and virtualhostoptions or is the future api bloat worth the better ux now?

sam-heilbron: slack reference: https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1718645410140749

sam-heilbron: The idea here was mainly that this hatch we are building, is really due to the linked issues that I mentioned in the PR body, and that we need solve those. Our hope was that we could provide a simple (temporary) API, and fix validation correctly, rather than making the API more complex, and leaving the webhook functionality still broken. If we find that validation will need more granular configuration moving forward, we figured we could add it then. wdyt?

docs/content/reference/values.txt
```diff
@@ -454,6 +454,7 @@
 |gateway.validation.webhook.timeoutSeconds|int||the timeout for the webhook, defaults to 10|
 |gateway.validation.webhook.extraAnnotations.NAME|string||extra annotations to add to the webhook|
 |gateway.validation.webhook.skipDeleteValidationResources[]|string||resource types in this list will not use webhook valdaition for DELETEs. Use '*' to skip validation for all resources. Valid values are 'virtualservices', 'routetables','upstreams', 'secrets', 'ratelimitconfigs', and '*'. Invalid values will be accepted but will not be used.|
+|gateway.validation.webhook.enablePolicyApi|bool|true|enable validation of Policy Api resources (RouteOptions, VirtualHostOptions) (default: true)|
```
lgadban, line 457: Should this mention that policy resource validation is currently tied to kube gateway mode?

lgadban (dismissed):
LGTM other than 1 comment

install/helm/gloo/generate/values.go
lgadban: No real comment other than I see both sides of this; I think the biggest factor is that eventually (not sure how feasible it is for existing, large-scale Edge users) we would want to supersede webhook-style validation entirely with CEL-based schema-style validation

docs/content/reference/values.txt
sam-heilbron: Updated to reflect that. Let me know what you think

nfuden (approved):
Note that this means an invlaid order could break a non-k8s gateway resource apply but it seems that is ok in this case.

Therefore approving

docs/content/reference/values.txt
```diff
@@ -454,6 +454,7 @@
 |gateway.validation.webhook.timeoutSeconds|int||the timeout for the webhook, defaults to 10|
 |gateway.validation.webhook.extraAnnotations.NAME|string||extra annotations to add to the webhook|
 |gateway.validation.webhook.skipDeleteValidationResources[]|string||resource types in this list will not use webhook valdaition for DELETEs. Use '*' to skip validation for all resources. Valid values are 'virtualservices', 'routetables','upstreams', 'secrets', 'ratelimitconfigs', and '*'. Invalid values will be accepted but will not be used.|
+|gateway.validation.webhook.enablePolicyApi|bool|true|enable validation of Policy Api resources (RouteOptions, VirtualHostOptions). NOTE: This only applies if the Kubernetes Gateway Integration is also enabled (kubeGateway.enabled). (default: true)|
```
lgadban, line 457: nit: I think the `default: true` text should go before the NOTE, as it's not clear if the default is true for enablePolicyApi or for kubeGateway.enabled

lgadban (approved):
LGTM but found a nit on the new addition to the doc string

docs/content/reference/values.txt
lgadban: resolve and merge if desired

sam-heilbron: heh i had the same fear, ill update

npolshakova (approved):

lgadban (approved):

Issue #5579, opened 2021-11-16
Support Extended HTTP Methods


guydc:
**Is your feature request related to a problem? Please describe.**
Envoy does not support extended HTTP methods. This limitation is related to the envoy http parser: https://github.com/envoyproxy/envoy/issues/5155#issuecomment-957814532. Several protocols use extended HTTP methods, like WebDAV. The [IANA http method registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) mentions several methods that are not supported by envoy, like `LABEL`, `UPDATE`, `CHECKIN`, etc. 

**Describe the solution you'd like**

* Support for all IANA registry methods
* Support for custom (arbitrary) HTTP methods

**Describe alternatives you've considered**
Exposing HTTP services that use custom HTTP methods with a TCP Gateway. This is not ideal, since HTTP request processing features cannot be applied to such services. 

**Additional context**

* https://github.com/nodejs/http-parser/issues/92
* https://github.com/nodejs/llhttp/issues/11
* list of supported methods in http-parser: https://github.com/nodejs/http-parser/blob/ec8b5ee63f0e51191ea43bb0c6eac7bfbff3141d/http_parser.h#L165

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206768562311555/1207435393311532) by [Unito](https://www.unito.io)

nfuden:
@ashishb-solo  this needs a backport to 1.17

PR #9633, opened 2024-06-17
5579 extended http methods

ashishb-solo:
# Description

Enable extended HTTP methods. Prior to this change, Envoy only allowed a
limited set of HTTP methods, and anything not allowed in a hard-coded list
would be rejected. This change allows users to configure Envoy so that extended
methods, such as `LABEL` or `UPDATE` can be passed.

The code changes are fairly straightforward, but potential compatibility issues with future envoy releases are not straightforward, so I would appreciate if reviewers could focus in particular on this area.

## API changes
- Created a new header validation settings message and exposed it at the HTTP Listener options level.

## Code changes
- Create the `HeaderValidation` plugin to manage settings related to header validation

## CI changes
- Added e2e tests (unit tests still needed)

## Docs changes
- See documentation in .proto files

# Context

https://soloio.slab.com/posts/extended-http-methods-design-doc-40j7pjeu

## Interesting decisions
 
See slab document regarding status of option settings in upstream Envoy.

## Testing steps

TODO

## Notes for reviewers

Please read the design document linked on slab carefully to understand the current state of this feature in upstream Envoy, especially how it pertains to Universal Header Validation and the deprecation status of the features that we're using in upstream Envoy.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/5579


projects/gloo/api/v1/options.proto
```diff
@@ -231,6 +232,12 @@ message HttpListenerOptions {
 
     // Enterprise only: Listener-level stateful session settings
     stateful_session.options.gloo.solo.io.StatefulSession stateful_session = 35;
+
+    // Whether custom HTTP methods should be allowed. Defaults to false (custom
```
bewebi, line 236: Here and on VirtualHostOptions the description is out of date, treating the field like a `bool`
It seems the description on RouteOptions is more up-to-date/reflective of the field having a type that might be extended in the future

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,32 @@
...
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation";
+
+message HeaderValidationSettings {
+    // Whether custom HTTP methods should be allowed. Defaults to false (custom
+    // HTTP methods are not allowed). The list of default-allowed HTTP methods
+    // can be found here:
+    // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
+    // **BREAKING API COMPATIBILITY WARNING**:
+    // Note that right now, this field only changes whether custom header methods are allowed on HTTP/1 connections only.
```
bewebi, line 12:
```suggestion
    // Note that right now, this field only changes whether custom header methods are allowed on HTTP/1 connections.
```

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,32 @@
...
+    // HTTP methods not just on HTTP/1, but also on all other HTTP protocols as
+    // well.
+
+    // As of right now, this field is only supported on HTTP/1 connections.
+    // When Universal Header Validation is enabled in Envoy, this field will
+    // apply to all HTTP protocols.
+    // TODO: let's think a bit more carefully about how we want to design this
+    // for future-proofing purposes. I think we may want to consider having an
+    // oneof here instead that could be something like (allow default
+    // methods/allow all methods/custom allow-list).
```
bewebi, line 23: The current implementation seems reasonable enough to me

I would expect once UHV is enabled we'd add a new type (potentially a oneof as described) for configuring UHV and have `allow_custom_header_methods` take the most similar behavior (ie allow all methods) while also being marked as deprecated, perhaps even in a oneof with the UHV settings

We could allow it to stick around in deprecated state indefinitely, until users have had more than enough time to move off it

I think this approach is what's suggested on L27-30

projects/gloo/pkg/plugins/registry/registry.go
```diff
@@ -112,6 +113,7 @@ func Plugins(opts bootstrap.Opts) []plugins.Plugin {
 		deprecated_cipher_passthrough.NewPlugin(),
 		local_ratelimit.NewPlugin(),
 		istio_automtls.NewPlugin(opts.GlooGateway.IstioValues.SidecarOnGatewayEnabled),
+		header_validation.NewPlugin(),
```
bewebi, line 116: I have no reason to think this can't go last, just want to be sure we're considering ordering intentionally

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,32 @@
...
+    // we may also want to model our API similarly to upstream Envoy's UHV API
+    // WARNING: these options should not be considered stable, and this API is
+    // subject to change in the future.
+    // question: alternately, we could enable this as an HTTP/1-only option,
+    // mark it as deprecated, and then when UHV is enabled, we could remove the
+    // deprecated option and introduce a new option for all HTTP protocols.
+    bool allow_custom_header_methods = 1;
```
ashishb-solo, line 30: the name here is a bit overkill. this is the current configuration:

```
spec:
  bindAddress: '::'
  bindPort: 8080
  httpGateway:
    options:
      headerValidationSettings:
        allowCustomHeaderMethods: true
```

bewebi (commented):
Overall seems like a good approach to me

We may also want to call out the potential breaking changes where we specify the Envoy version (ie [here](https://github.com/solo-io/solo-projects/blob/bf81ebde87416b5f850a43236a6b3cbfd9c6295c/Makefile#L47)) so that there are no surprises, but I think there is/will be enough tests that will fail if/when `allow_custom_methods` is deprecated in Envoy that we aren't at risk of shipping something broken

The real risk is that there will be hidden required work at some point when we bump Envoy versions, but we're calling it out as loudly as possible while also describing the path forward, so I think the pain/surprise will be minimized

projects/gloo/pkg/plugins/hcm/plugin.go
```diff
@@ -137,6 +137,17 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, _ *v1.Listener,
...
+	// 		if out.GetHttpProtocolOptions() == nil {
+	// 			out.HttpProtocolOptions = &envoycore.Http1ProtocolOptions{}
+	// 		}
+	// 		// ALERT: AllowCustomMethods is deprecated from upstream Envoy and
+	// 		// scheduled for removal. When that setting is removed, we must use
+	// 		// Universal Header Validation to support this functionality. See
+	// 		// https://soloio.slab.com/posts/extended-http-methods-design-doc-40j7pjeu
+	// 		out.HttpProtocolOptions.AllowCustomMethods = listener.GetOptions().GetHeaderValidationSettings().GetAllowCustomHeaderMethods()
+	// 	}
+
```
sheidkamp, line 150: This was another approach considered?

projects/gloo/pkg/plugins/header_validation/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, _ *v1.Listener, listener *v1.HttpListener, out *envoyhttp.HttpConnectionManager) error {
+	in := listener.GetOptions().GetHeaderValidationSettings()
+	if allow_custom_methods := in.GetAllowCustomHeaderMethods(); allow_custom_methods {
+		if out.GetHttpProtocolOptions() == nil {
+			out.HttpProtocolOptions = &envoycore.Http1ProtocolOptions{}
+		}
+		out.GetHttpProtocolOptions().AllowCustomMethods = allow_custom_methods
+	}
+	return nil
+}
```
sheidkamp, line 49: Should we have a test for this even if it is trivial?

projects/gloo/api/v1/options/header_validation/header_validation.proto
sheidkamp: I like the idea of a one-of to be prepared for when UHV is enabled, and having `allowCustomHeaderMethods` only apply to HTTP/1 so that it continues to behave the same after that update. (and do we ever really remove deprecated options?)

projects/gloo/pkg/plugins/header_validation/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/solo-io/gloo/projects/gloo/pkg/plugins"
+)
+
+// Plugin for header validation.
+// This plugin is intended to serve as a location for header validation in
+// Envoy. Currently, it is quite bare and only sets a single option in Envoy
+// that disables the HTTP/1 parser from filtering requests with custom HTTP
+// methods. However, when Universal Header Validation is enabled in upstream
+// Envoy, this plugin will serve as a location to configure all potential UHV
+// features (including allowing custom HTTP methods).
```
nfuden, line 16: Why is this a plugin? Is it not part of the core hcm functionality? or is this a preemptive attempt to proof against the UHV?

projects/gloo/pkg/plugins/header_validation/plugin.go
ashishb-solo: i definitely plan on adding unit tests! but i'd like to wait until we have an API solidified first

projects/gloo/pkg/plugins/registry/registry.go
nfuden: If anything it should go early.

If we are to make the uhv a thing its more likely that it has side effects that affect other plugins rather than the other way around.



That being said plugins generally shouldnt have sideeffects so should be safe.

test/e2e/header_validation_test.go
```diff
@@ -0,0 +1,92 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	gatewayv1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1"
+	gatewaydefaults "github.com/solo-io/gloo/projects/gateway/pkg/defaults"
+	gloov1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	header_validation "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation"
+	"github.com/solo-io/gloo/test/e2e"
+)
+
+var _ = Describe("Header Validation", Label(), func() {
```
nfuden, line 19: nit: we should be triyng to use the new testing framework

test/e2e/header_validation_test.go
```diff
@@ -0,0 +1,92 @@
...
+			req := testContext.GetHttpRequestBuilder().Build()
+			result, err := testutils.DefaultHttpClient.Do(req)
+			g.ExpectWithOffset(1, err).NotTo(HaveOccurred())
+			g.ExpectWithOffset(1, result).Should(matchers.HaveOkResponse())
+		}, "5s", ".5s").Should(Succeed(), "GET with valid host returns a 200")
+	}
+
+	buildRequest := func() *http.Request {
+		return testContext.GetHttpRequestBuilder().
+			WithMethod("CUSTOMMETHOD").
```
nfuden, line 56: Its likely important to make a distinction between custom and some of the somewhat supported ones. 
Ie a real method that just is currently denied but is part of an extended rfc

projects/gloo/pkg/plugins/hcm/plugin.go
ashishb-solo: Yes, can't decide right now if I should just toss this into the HCM plugin or create a new plugin to future-proof against UHV. Let's move it to [this conversation](https://github.com/solo-io/gloo/pull/9633/files#r1643370258)

ashishb-solo:
After a bit of discussion yesterday, we made the following decisions:

1. only enable this on the gateway (and possibly in settings) for now. we cannot decide exactly how we want lower level configurations (like route table) to override higher level configurations (ie. which one takes precedence), so let's hold off on making a decision here until we have to.

2. we can just pop all this functionality into the existing gloo plugin for hcm. no need to create a new one.

projects/gloo/pkg/plugins/header_validation/plugin.go
ashishb-solo: deleted the new plugin and reverted to adding this to the existing hcm plugin in [this commit](https://github.com/solo-io/gloo/pull/9633/commits/d66a9ce44457317020c05f8a25fe4b333a327763)

projects/gloo/pkg/plugins/hcm/plugin.go
```diff
@@ -137,6 +137,17 @@ func (p *plugin) ProcessHcmNetworkFilter(params plugins.Params, _ *v1.Listener,
 		out.GetHttpProtocolOptions().EnableTrailers = in.GetEnableTrailers().GetValue()
 	}
 
+	// 	if listener.GetOptions().GetHeaderValidationSettings().GetAllowCustomHeaderMethods() {
+	// 		if out.GetHttpProtocolOptions() == nil {
+	// 			out.HttpProtocolOptions = &envoycore.Http1ProtocolOptions{}
+	// 		}
+	// 		// ALERT: AllowCustomMethods is deprecated from upstream Envoy and
```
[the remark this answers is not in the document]
ashishb-solo: added the version in [this commit](956e2ef7700bb78f1eaa6a838411e5a0c0cc536a)

test/e2e/header_validation_test.go
ashishb-solo: At least as far as the current implementation goes, this should not make a difference, because the Envoy option we're changing simply turns off all checking for the header methods.

test/e2e/header_validation_test.go
ashishb-solo: do we have a new api for e2e tests? i thought the new testing framework was for kube2e tests only. for what it's worth i copied this from `test/e2e/example_test.go`

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,32 @@
...
+package header_validation.options.gloo.solo.io;
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation";
+
+message HeaderValidationSettings {
+    // Whether custom HTTP methods should be allowed. Defaults to false (custom
+    // HTTP methods are not allowed). The list of default-allowed HTTP methods
+    // can be found here:
+    // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
+    // **BREAKING API COMPATIBILITY WARNING**:
```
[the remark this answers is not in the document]
ashishb-solo: the updated documentation can be found [here](https://github.com/solo-io/gloo/pull/9633/files#diff-88223df875d38587ec35033f9d48c9c2f462001c81d18709aa043eb7ed8242f5R8) and should no longer have this message

projects/gloo/pkg/plugins/registry/registry.go
ashishb-solo: this is no longer relevant since we got rid of the new plugin and are now reusing hcm

projects/gloo/api/v1/options/header_validation/header_validation.proto
ashishb-solo: i have updated the proto api to use a oneof. the new proto api can be found [here](https://github.com/solo-io/gloo/pull/9633/files#diff-88223df875d38587ec35033f9d48c9c2f462001c81d18709aa043eb7ed8242f5R8)

ashishb-solo:
The following API should be good to configure the gateway to allow extended HTTP methods:

```
kind: Gateway
metadata:
  name: gateway-proxy
  namespace: gloo-system
spec:
  bindAddress: '::'
  bindPort: 8080
  httpGateway:
    options:
      headerValidationSettings:
        allowCustomMethods: {}
```

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,22 @@
...
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation";
+
+message HeaderValidationSettings {
+	// Whether custom HTTP methods should be allowed. If this field is omitted,
+	// Gloo defaults to returning 400 responses for requests that use custom
+	// header methods. The list of HTTP methods that are allowed by default can
+	// be found here:
+  // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
```
ashishb-solo, line 13: TODO fix this indentation

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+package header_validation.options.gloo.solo.io;
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation";
+
+message HeaderValidationSettings {
+	// Whether custom HTTP methods should be allowed. If this field is omitted,
```
ashishb-solo, line 9: TODO clean up this documentation

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,22 @@
...
+	// Gloo defaults to returning 400 responses for requests that use custom
+	// header methods. The list of HTTP methods that are allowed by default can
+	// be found here:
+  // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
+	// Currently, this option only affects HTTP/1 traffic and not HTTP/2. A
+	// future release of Gloo will expand this option to support all HTTP
+	// protocols.
+	oneof custom_methods {
+		// Allow - all HTTP methods will be allowed
+		google.protobuf.Empty allowCustomMethods = 1;
```
ashishb-solo, line 19: TODO think of a better name here and update the documentation accordingly. i'm particularly not a fan of the term "custom" methods but having a hard time thinking of something better. perhaps something like this?

```
oneof method_filter {
    google.protobuf.Empty allowAllHttpMethods = 1;
}

Comment (ashishb-solo) [reply projects/gloo/api/v1/options/header_validation/header_validation.proto]: Resolved in https://github.com/solo-io/gloo/pull/9633/commits/42b5c721b622bda24b9b32dd10e055a8b2a2a4d0 - I changed the name to `disableMethodValidation`

File: projects/gloo/api/v1/options/header_validation/header_validation.proto (hunk 13 above)
Comment (ashishb-solo) [reply projects/gloo/api/v1/options/header_validation/header_validation.proto]: Resolved in https://github.com/solo-io/gloo/pull/9633/commits/42b5c721b622bda24b9b32dd10e055a8b2a2a4d0

File: projects/gloo/api/v1/options/header_validation/header_validation.proto (hunk 14 above)
Comment (ashishb-solo) [reply projects/gloo/api/v1/options/header_validation/header_validation.proto]: Resolved in https://github.com/solo-io/gloo/pull/9633/commits/42b5c721b622bda24b9b32dd10e055a8b2a2a4d0

File: projects/gloo/pkg/plugins/hcm/plugin_test.go [hunk 16]
```diff
@@ -45,25 +44,22 @@ var _ = Describe("Plugin", func() {
...
 		}
-		settings = &hcm.HttpConnectionManagerSettings{}
 	})
 
 	AfterEach(func() {
 		cancel()
 	})
 
-	processHcmNetworkFilter := func(cfg *envoyhttp.HttpConnectionManager) error {
+	processHcmNetworkFilter := func(cfg *envoyhttp.HttpConnectionManager, glooOptions *v1.HttpListenerOptions) error {
```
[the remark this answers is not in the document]
ashishb-solo: Originally, `settings` was a global variable that was never being used in more than one place, so I figured I might as well remove the global and make it a local. I had to change the type anyway - `settings` was originally a `hcm.HttpConnectionManagerSettings` object, but the header validation settings live outside that struct (in the `v1.HttpListenerOptions` struct instead). So I was going to have to update all of them anyway.

projects/gloo/api/v1/options/header_validation/header_validation.proto
```diff
@@ -0,0 +1,24 @@
...
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/header_validation";
+
+message HeaderValidationSettings {
+  // Whether header methods should be validated. By default, Envoy will
+  // validate that the HTTP method is valid on HTTP/1 connections. The default
+  // allowed methods can be found here:
+  // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
```
sheidkamp, line 12:
```suggestion
  // How header methods will be validated. By default, Envoy will
  // validate HTTP methods for HTTP/1 connections against the default set of allowed methods.
  // The default allowed methods can be found here:
  // https://github.com/envoyproxy/envoy/blob/2970ddbd4ade787dd51dfbe605ae2e8c5d8ffcf7/source/common/http/http1/balsa_parser.cc#L54
```

projects/gloo/pkg/plugins/hcm/plugin_test.go
```diff
@@ -408,4 +398,18 @@ You can force this test to pass by adding the new fields listed below to project
...
+		It("Enables header validation settings", func() {
+			cfg := &envoyhttp.HttpConnectionManager{}
+			options := v1.HttpListenerOptions{
+				HeaderValidationSettings: &header_validation.HeaderValidationSettings{
+					HeaderMethodValidation: &header_validation.HeaderValidationSettings_DisableHttp1MethodValidation{},
+				},
+			}
+			err := processHcmNetworkFilter(cfg, &options)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cfg.GetHttpProtocolOptions().AllowCustomMethods).To(BeTrue())
```
sheidkamp, line 412: Could we stick a negative version of this into the test above?

test/e2e/example_test.go
```diff
@@ -54,6 +54,18 @@ var _ = Describe("Example E2E Test For Developers", Label(), func() {
 		// This means that a Proxy object is dynamically generated, and from there an xDS snapshot is computed
 		// and sent to Envoy to handle traffic
 
+		It("defaults to returning HTTP 400 on requests with custom HTTP methods", func() {
```
sheidkamp, line 57: Is this test in the right place?

sheidkamp (commented):
A couple notes on the tests.

Would like to see a demo and/or testing steps before approval.

test/e2e/example_test.go
ashishb-solo: Oof, thanks for catching that. Removed in c7fe0aa18f1a7feb89b7153912137aeba62c2b79

test/e2e/header_validation_test.go
nfuden: Yeah but future proofing in case we make the cut over and it gets stricter would be nice to have a test failing as a signal

projects/gloo/pkg/plugins/hcm/plugin_test.go
ashishb-solo: I think I addressed this in 072a229e426d413e4d978d4605529379f9d4e57a but I'm not 100% sure if this is exactly what you were asking for - let me know if this is the case

sheidkamp: Yup, looking for something like that somewhere.

sheidkamp (approved):

nfuden (approved):

PR #9634, opened 2024-06-17
tests: Update how kube2e tests report and update the glooctl check tests

nfuden:
# Description

XDS pollution occurs with flakey ordering of xds resources as detected in enterprise nack detector for fallback cluster in the case of invalid routes.

This makes enterprise tests fail and can potentially make cluster load assignment appear out of sync until a new update occurs or a pod rolls.

Given that actual impact on production systems is low and the danger of other failing tests not being caught in the interim this is a patch to clean up our test runs.


nfuden:
https://github.com/solo-io/gloo/actions/runs/9566653960?pr=9637 shows the new reporting structure where multiple failures are shown

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -37,5 +37,5 @@ runs:
     if: ${{ failure() }}
     uses: actions/upload-artifact@v4
     with:
-      name: bug-report
+      name: bug-report-${{ inputs.cluster-name }}
```
sheidkamp, line 40: Nice!

sheidkamp (approved):

jenshu (approved):

PR #9636, opened 2024-06-17
Remove old Istio tests, add new Istio suite with regression values

npolshakova:
# Description
Cleans up old Istio regression tests. Adds regression test with old helm values to new framework. 

## API changes
None

## Code changes
- Removes old Istio regression tests from ci pipelines and code
- Removes old Istio glooctl tests 
- Removes Istio installation from setup script
- Adds a new Istio helm setup to the new framework that uses the deprecated values (enableIstioSidecarOnGateway) to check for regressions 
- Adds missing port routing test with edge apis. 

## CI changes
- Removed Istio regression tests 

## Docs changes
None

# Context

## Interesting decisions
None

## Testing steps
- Validated k8s and gloo edge Istio integration tests run locally

## Notes for reviewers
None

# Checklist:


.github/workflows/regression-tests.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
         # upgrade tests are run on LTS but not on main branch, for main they are run nightly
         # ingress will be deprecated from 1.17. Ref: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1716389614777799
         # this is the github action version of ternary op
-        kube-e2e-test-type: [ 'gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'glooctl', 'upgrade', 'istio' ]
```
sam-heilbron, line 52: 🎉

test/kube2e/istio/README.md
```diff
@@ -1,287 +0,0 @@
-An overview of some key concepts related to this test suite:
```
sam-heilbron, line 1: This seems like valuable information, that likely deserved a different place originally. Are we safe to remove this entirely, or should we be moving some/all of this to https://github.com/solo-io/gloo/tree/main/devel/architecture?

test/kubernetes/e2e/features/port_routing/edge_gateway_suite.go
```diff
@@ -0,0 +1,177 @@
...
+
+Each port routing test:
+ 1. Attach HttpRoute with different port/targetport definition per test
+ 2. Remove HttpRoute, proxy still exists without any routes
+
+TearDownSuite:
+ 1. Deletes the k8s Gateway
+ 2. Proxy de-provisioned
+*/
+func NewEdgeGatewayApiTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
```
sam-heilbron, line 44: nit: EdgeGateway implies those are the APIs we use, but the commentary for this fucntions is about k8s Gateway API

test/kubernetes/e2e/features/port_routing/testdata/edge-routing.yaml
```diff
@@ -0,0 +1,16 @@
+# Don't set namespace, apply to test installation namespace
```
sam-heilbron, line 1: ❤️

sam-heilbron (commented):
looking great! Some small comments

test/kubernetes/e2e/features/port_routing/edge_gateway_suite.go
npolshakova: Good catch! I'll update the comment!

test/kube2e/istio/README.md
npolshakova: Good idea, let's move it to the architecture docs!

npolshakova: Or do you think https://github.com/solo-io/gloo/blob/main/docs/content/guides/integrations/service_mesh/istio.md is better? The devel/architecture feels more developer facing, so I think that makes more sense, but a lot of similar information is in the `guides/integrations` docs.

sam-heilbron: Yeah I figure in general the most important part is that we have some docs. I tend to like the developer documentation because it means we can add a lot of context that is relevant to the code, which users might not care about. Plus, as we build our user-facing docs, the docs team can then leverage that.

sam-heilbron (approved):

jenshu (approved):

PR #9639, opened 2024-06-18
[1.17] Retry on leader lease renewal failure

davidjumani:
# Description

Backport of https://github.com/solo-io/gloo/pull/9563

After a container has become a leader, any Kube API server unavailability results in the container crashing. This happens as the leader is unable to renew the lease. This is by design as outlined [here](https://github.com/solo-io/gloo/issues/8107#issuecomment-1516250478)
> if we allowed gloo to continue to function as a leader during kube apiserver outage, we risk having two leaders in other failure modes. we should remove the panic and allow gloo to continue to serve last-known xds as a follower (effectively having two followers until kube apiserver recovers). this idea is similar to the role xds relay could play for gloo edge

However, crashing the gloo pods can also lead to an outage during scaling as the gateway-proxy pods that come up cannot fetch any configs and all routes result in 404s.

Now instead of crashing, the gloo pod falls back to a follower. This prevents an outage and any other pod can take over as leader if possible
This is fine as a leader only writes reports / statuses over [here](https://github.com/solo-io/gloo/blob/0637825838c285e978c51d6a00f68f4e1d5d08d6/pkg/bootstrap/leaderelector/leader.go#L35) and [here](https://github.com/solo-io/gloo/blob/0637825838c285e978c51d6a00f68f4e1d5d08d6/projects/gloo/pkg/syncer/translator_syncer.go#L231). On any failure, the pod becomes a follower and if elected back as a leader will continue to write reports.

## Code changes
- Introduce a new `Reset` method on the identity implementation that allows an identity to fall back to a follower

## CI changes
- Kind clusters that run in CI now do not come up with their own CNI as kindnet does not support custom network policies.
Instead, cilium is installed as CNI as we need to test Kube API server unavailability

# Context
[Kube API unavailability results in a gloo container crash](https://github.com/solo-io/gloo/issues/8107)
[When leader election fails, gloo crashes](https://github.com/solo-io/gloo/issues/7346#top)
[Design Doc](https://soloio.slab.com/posts/dont-crash-on-failure-to-renew-leader-lease-kjkqn0ry)

## Testing steps
Kube2e tests to verify the following :
- It can recover from a Kube APIServer unavailability
- It falls back to a follower
- Another leader can get elected

## Notes for reviewers

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/8107

### DIFF
--- a/changelog/v1.17.0-rc5/dont-crash-on-failed-lease-renewal.yaml
+++ b/changelog/v1.17.0-rc5/dont-crash-on-failed-lease-renewal.yaml
@@ -1,0 +1,4 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/8107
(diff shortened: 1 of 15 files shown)


nfuden (approved):
Im surprised that new_feature in changelog passed bot but looks good to me

jenshu (approved):

afro-coder:
Hey folks is there a label to add backporting to 1.15.x for this fix

PR #9646, opened 2024-06-19
introduce test coverage reporting, update devel with Gloo Gateway references

sam-heilbron:
# Description

- Introduce test coverage reporting
- Update references from Gloo Edge -> Gloo Gateway in `devel` folder

# Context

## Kube Gateway Unit Tests
I updated the unit tests to run with our common make target. Most importantly, this runs tests with `-race` enabled, which previously it did not. When I did this, I found:
- https://github.com/solo-io/solo-projects/issues/6408

I intend to investigate this failure separate from this PR.

## Test Coverage
The goal of this is to provide _some_ signal about our existing test coverage, but to not make it required. 

I had explored a few options, but ultimately chose the tool that was used in GLoo Platform. I figured consistency across projects would make it easy to share learnings/updates.

My hope is that after this PR, we can expand it's usage in the following steps:
- Migrate more unit tests from build-bot into gha
- Try to find smart ways to run tests for a given project only if code in that project changed
- Find ways to expand our testing so that we can raise the coverage in certain areas
- Establish a way to make the test coverage reporting more visible (right now it's part of the action, but the results are swallowed)
- See if we want to set some minimum (reasonable) standards which will fail if not met on a PR 

_I do not intend to do this myself, more just calling out what I think our plan could be_

## Docs Updates
This is intended to be a first pass to update some nomenclature in our docs

## Interesting decisions
 -

## Notes for reviewers
- After this is approved, please ping me in slack. I need to update branch protection rules to account for the name change to the gha

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #9651, opened 2024-06-20
gateway2/deployer: fix race due to context cancellation

sam-heilbron:
# Description

Avoid a helm race condition in our deployer code.

# Context
In https://github.com/solo-io/gloo/pull/9646 we run our tests with `-race`. In doing so, we encountered a race condition: https://github.com/solo-io/solo-projects/issues/6408.

Upon speaking with @yuval-k ([slack ref](https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1718893489626729)), the proposed plan is to side step the race condition in helm.

## Interesting decisions
-

## Testing steps
- This should not change any existing behavior, so I intentionally did not change tests. 

## Notes for reviewers
-

# Checklist:


projects/gateway2/deployer/deployer.go
```diff
@@ -323,17 +323,37 @@ func (d *Deployer) getValues(gw *api.Gateway, gwParam *v1alpha1.GatewayParameter
...
+
+	// Upstream Helm has a subtle race-condition that occurs around context cancellation of Installations
+	// This manifests itself when running tests with `-race` enabled:
+	// https://github.com/solo-io/solo-projects/issues/6408
+	//
+	// https://github.com/helm/helm/blob/a2a324e51165388448be8caaf8737b7fd2f1a19a/pkg/action/install.go#L229
+	// https://github.com/helm/helm/blob/a2a324e51165388448be8caaf8737b7fd2f1a19a/pkg/action/install.go#L403
+	// https://github.com/helm/helm/blob/a2a324e51165388448be8caaf8737b7fd2f1a19a/pkg/action/install.go#L416
+	//
+	// Since we are not actually relying on the release itself, we can side-step this race condition by not
```
yuval-k, line 352: the reason we don't need the context here is because we use it here in `ClientOnly = true` mode. which means that there is no i/o (i.e. no reads/writes to k8s) that would need to be cancelled. essentially guaranteeing that this function terminates.

lgadban (approved):

sam-heilbron:
/kick build bot immediately reported error, but no failure

sam-heilbron:
dozer

PR #9657, opened 2024-06-21
Fully default gwparams

jbohanon:
# Description

This PR simplifies the way default values are derived when deploying a dynamically generated gateway proxy from the Kube Gateway API. Previously, there were some defaults defined in `projects/gateway2/deployer/values.go` which were not user-facing. With the introduction of a default GatewayParameters resource, we have the ability to move all user-facing default values into the main Gloo Gateway helm chart at `install/helm/gloo/values-template.yaml`. 

This PR moves all default values in order that they may be configurable by the user.
## API changes
- Autoscaling has been removed from GatewayParameters CR since it was non-functional and will need some reworking later anyway. 

## Code changes
- Removed code around translating/merging/setting autoscaling in the deployer and in templates

## CI changes
- Added `^TestK8sGatewayNoDefaultGatewayParameters$` to cluster_two of the Kubernetes End-to-End tests.
  - cluster_two was chosen based on having the lowest runtime in recent successful CI runs

# Context
In order to run on OpenShift, we have to be able to modify/unset the security contexts. Before this work, the default values for the envoy container's security context were in the internal helm chart and inaccessible to the user. Because of the inheritance paradigm selected for GatewayParameters (i.e. nil value == inherit from parent), it was not possible to unset a value that had a default. By making the defaults user-facing, users can set the default to `null` via helm and achieve the desired effect.

## Interesting decisions
TODO

## Testing steps
TODO
1. Install helm with `--set kubeGateway.gatewayParameters.glooGateway.envoyContainer.securityContext=null`
2. apply a gateway
3. see that deployed proxy has no container security context

## Notes for reviewers
TODO

# Checklist:


projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -57,8 +57,9 @@ message KubernetesProxyConfig {
   // the network.
   kube.gateway.gloo.solo.io.Service service = 4;
 
+  // NOTE: Not used currently
   // Autoscaling configuration.
-  kube.gateway.gloo.solo.io.Autoscaling autoscaling = 5;
+  // kube.gateway.gloo.solo.io.Autoscaling autoscaling = 5;
```
jbohanon, line 62: we should reserve 5 or renumber the other fields

projects/gateway2/helm/gloo-gateway/templates/_helpers.tpl
```diff
@@ -65,18 +65,6 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 gateway.networking.k8s.io/gateway-name: {{ .Release.Name }}
 {{- end }}
 
-{{/*
-Create the name of the service account to use
```
npolshakova, line 69: Was this not used anywhere?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -318,9 +318,9 @@ data:
         role: gloo-kube-gateway-api~{{ $gateway.gatewayNamespace }}~{{ $gateway.gatewayNamespace }}-{{ $gateway.gatewayName | default (include "gloo-gateway.gateway.fullname" .) }}
     static_resources:
       listeners:
-      - name: read_config_listener
+      - name: readiness_listener
         address:
-          socket_address: { address: 0.0.0.0, port_value: {{ $gateway.readinessPort }} }
+          socket_address: { address: 0.0.0.0, port_value: 8082 }
```
npolshakova, line 323: Is there a reason this isn't configurable anymore?

install/helm/gloo/templates/43-gatewayparameters.yaml
```diff
@@ -21,6 +21,14 @@ spec:
     envoyContainer:
       image:
         {{- merge $gg.envoyContainer.image $global.image | include "gloo-gateway.gatewayParametersImage" | nindent 8 }}
+{{- if $gg.envoyContainer.securityContext }}
```
npolshakova, line 24: Do we have a helm test for the gatewayparameters template?

install/helm/gloo/generate/values.go
```diff
@@ -351,6 +351,15 @@ type SecurityContext struct {
 	*SecurityOpts
 }
 
+// GatewayParamsSecurityContext is a passthrough struct that provides the corev1.SecurityContext without
+// exposing the SecurityOpts/MergePolicy. MergePolicy is irrelevant to the GatewayParameters case because
+// there is already a default and merge behavior defined. The "default" GatewayParameters are expected to
+// be the base config, which is where a default policy can defined; each gwapi.Gateway can have specific
+// GatewayParameters which can then override/merge into the default policy
+type GatewayParamsSecurityContext struct {
```
npolshakova, line 359: Is there a reason we want to wrap this instead of just using `*corev1.PodSecurityContext` directly on the `EnvoyContainer`? Also do we want to add this to the Istio container and SDS container? It looks like the Edge Gateway template does do this: https://github.com/solo-io/gloo/blob/e308d8fa3281a87c322fe30eb28369f30e0718e5/install/helm/gloo/templates/7-gateway-proxy-deployment.yaml#L303

npolshakova: I.e. use *corev1.PodSecurityContext in `SdsContainer`, `IstioProxyContainer` and `SdsContainer` instead of SecurityContext?

test/kubernetes/e2e/tests/manifests/k8s-gateway-no-default-gatewayparameters-test-helm.yaml
```diff
@@ -0,0 +1,58 @@
...
+kubeGateway:
+  # This is the field that enables the K8s Gateway Integration in Gloo Gateway
+  enabled: true
+  gatewayParameters:
+    # We are testing the behavior of the deployer when users null out fields on this object
+    glooGateway: 
+      # TODO (jbohanon): fix this so we don't fail the helm render with kubeGateway.gatewayParameters.glooGateway.proxyDeployment=null
+      proxyDeployment: 
+        replicas: 1
+      # TODO (jbohanon): fix this so we can't end up with Unspecified
```
npolshakova, line 37: Do we have an issue to track these TODOs?

test/kubernetes/e2e/tests/k8s_gw_no_default_gatewayparameters_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/solo-io/gloo/pkg/utils/env"
+	"github.com/solo-io/gloo/test/kube2e/helper"
+	"github.com/solo-io/gloo/test/kubernetes/e2e"
+	. "github.com/solo-io/gloo/test/kubernetes/e2e/tests"
+	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
+	"github.com/solo-io/gloo/test/testutils"
+	"github.com/solo-io/skv2/codegen/util"
+)
+
+// TestK8sGateway is the function which executes a series of tests against a given installation
```
npolshakova, line 18: [nit] 
```suggestion
// TestK8sGatewayNoDefaultGatewayParameters is the function which executes a series of tests against a given installation with no default GatewayParameters
```

test/kubernetes/e2e/tests/manifests/k8s-gateway-no-default-gatewayparameters-test-helm.yaml
```diff
@@ -0,0 +1,58 @@
...
+    allowWarnings: true
+    alwaysAcceptResources: false
+    # skipping delete validation due to flakes per https://github.com/solo-io/solo-projects/issues/6272
+    webhook:
+      skipDeleteValidationResources:
+        - upstreams
+kubeGateway:
+  # This is the field that enables the K8s Gateway Integration in Gloo Gateway
+  enabled: true
+  gatewayParameters:
```
npolshakova, line 31: Is the test correctly named? We're testing an overwrite of the default GatewayParameters , not having no default parameters, right? Maybe `TestK8sGatewayCustomDefaultGatewayParameters` instead?

jbohanon: The goal is to have the entire gwp block her nulled, meaning we would end up with a default gateway parameters resource that would have very very little in it to make sure we still function. Perhaps `Minimal` is the right name modifier

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
lgadban: it was bloat in the values.yaml and essentially a hidden default, as users have no way to customize this anyway

install/helm/gloo/generate/values.go
lgadban: This is a good call; there are a few parts to this:

>Is there a reason we want to wrap this instead of just using *corev1.PodSecurityContext directly on the EnvoyContainer?

The main reason is so that I can add that comment block within trying to fit it inline as a field on `EnvoyContainer`

>Also do we want to add this to the Istio container and SDS container?
I missed this; yes we definitely need to do this, I'll update now.

>I.e. use *corev1.PodSecurityContext in SdsContainer and IstioProxyContainer instead of SecurityContext?
We need to use corev1.SecurityContext as it goes on a container but yes the sentiment is correct. Where it gets tricky is that `SdsContainer` is used in both contexts so I'll need to redefine it for use within GatewayParams.
The upcoming diff should make it more clear what is happening. 


>It looks like the Edge Gateway template does do this
The way that the rest of the edge containers do this is not compatible with containers in the dynamic proxy, as they have a different mechanism for defaulting/merging etc.
Internal slack ref: https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1718727320795099

projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -43,25 +43,27 @@ message KubernetesProxyConfig {
...
 
   // Istio integration configuration.
   IstioIntegration istio = 6;
+
+  // Configuration for the container running SDS. This field is rendered with defaults by
+  // helm in the default GatewayParameters for a GatewayClass, and may be overridden in
+  // override GatewayParameters attached to a Gateway resource.
+  // Note that if no integration requiring SDS is enabled, the SDS container will not be injected
+  // into the gateway proxy deployment.
+  SdsContainer sds_container = 7;
```
jbohanon, line 66: nit: I think this makes more sense by the `envoy_container` as a container definition separated from implementation options

lgadban: I'll move it back but I disagree as it is a more advanced/esoteric config. Skimming down the list of options, you always will have a proxy and its related components but SDS is much less likely to be configured

jbohanon: Sounds good, I'm not gonna impose that if from the outside it stood out to you.

Moreso wanted to make sure it wasn't just moved because the initial placement was presumed arbitrary

test/kubernetes/e2e/tests/manifests/k8s-gateway-no-default-gatewayparameters-test-helm.yaml
npolshakova: Yeah minimal sounds good! I think the name of the test and the helm manifest should reflect what the test is testing. Without context, I would have assumed this was testing deleting the default GatewayParameters/having no GatewayParameter resource.

test/kubernetes/e2e/tests/manifests/k8s-gateway-no-default-gatewayparameters-test-helm.yaml
jbohanon: I resolved the TODOs :+1:

install/helm/gloo/generate/values.go
lgadban: there's more to this; we'll need to tackle in https://github.com/solo-io/solo-projects/issues/6323

npolshakova (approved):

sam-heilbron (approved):

jbohanon:
Bulldozer

Issue #9660, opened 2024-06-21
Kube Gw reconciles any GatewayClass regardless of controllerName


lgadban:
### Gloo Edge Product

Open Source

### Gloo Edge Version

latest

### Kubernetes Version

n/a

### Describe the bug

Kube Gw reconciles any GatewayClass regardless of controllerName

### Expected Behavior

Only GwClasses with gloo-gateway controllerName should be reconciled

### Steps to reproduce the bug

Create a GwClass with any controllerName and see it get accepted

```
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: gateway-bogus
spec:
  controllerName: multi.io/bogus-gateway
```

PR #9668, opened 2024-06-24
Small Gateway Touch-Ups

lgadban:
# Description

Three small gateway fixes, organized by commit

[fix helm values for GwParams + sidecars](https://github.com/solo-io/gloo/pull/9668/commits/e579d15639d238a6dc21545153da382c5170bc7b)
Follow up to https://github.com/solo-io/gloo/pull/9657
Renders out securityContext for sds and istio sidecar containers to dynamic GG proxy and fixes the helm docs for these containers.

[remove extensions from deployer; no longer needed](https://github.com/solo-io/gloo/pull/9668/commits/75a99fbf148daeb16d445c4c694571ae36cc8c5a)
Remove unused Extensions from Deployer component

[filter irrelevant GwClasses](https://github.com/solo-io/gloo/pull/9668/commits/8fa395dc03531fc984999355df7fe60200c5469c)
Only act on GatewayClass resources that use the Gloo Gateway controller name

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9660


projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -120,6 +120,10 @@ spec:
           timeoutSeconds: 1
         resources:
           {{- toYaml .Values.gateway.resources | nindent 12 }}
+        {{- if $gateway.sdsContainer.securityContext }}
+        securityContext:
+          {{- toYaml $gateway.sdsContainer.securityContext | nindent 10 }}
+        {{- end }}
```
jbohanon, line 126: It would probably be valuable to have a unit test or e2e test that validates all fields we have on the GwP end up on the rendered resources. Big oopsie that this was able to slip by in the first place

lgadban: I completely agree; not sure the best way to tackle it though.
I was considering adding an e2e test for this change that could be a starting point but I don't want to delay these fixes getting into the release.

jbohanon: I think a new entry on the unit test table could be a quick way to add at least the validation for this.

> I don't want to delay these fixes getting into the release.

On the contrary, I think validating that nothing else was missed would be a super useful exercise.

install/test/k8sgateway_test.go
```diff
@@ -21,6 +21,7 @@ var _ = Describe("Kubernetes Gateway API integration", func() {
 			testManifest TestManifest
 			valuesArgs   []string
 		)
+		// TODO: why is this called a Makefile?
```
jenshu, line 24: lol copied from helm_test.go. they should probably all be renamed prepareManifest or renderManifest

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -120,6 +120,10 @@ spec:
           timeoutSeconds: 1
         resources:
           {{- toYaml .Values.gateway.resources | nindent 12 }}
```
jenshu, line 122: while you're here can you double check the 2 instances of `nindent 12` in this file? it looks like they should both be 10

lgadban: this is a good call, I will update the indent, but this whole line is actually incorrect as it should be the `sdsContainer` resources, not `gateway`.

I'll see if I can get this fixed and tested before release...

projects/gateway2/controller/controller.go
```diff
@@ -246,6 +245,13 @@ func shouldIgnoreStatusChild(gvk schema.GroupVersionKind) bool {
 func (c *controllerBuilder) watchGwClass(ctx context.Context) error {
 	return ctrl.NewControllerManagedBy(c.cfg.Mgr).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
```
jbohanon, line 248: I haven't looked at the impl, but specifying this multiple times is additive and not replacing right?

lgadban: yes, additive.

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.4/pkg/builder/controller.go#L193-L196

jbohanon (approved):

jenshu (dismissed):

jbohanon (approved):

jenshu (approved):

PR #9676, opened 2024-06-24
setup kind fix for installing cilium

jbohanon:
# Description
In order to avert helm caching issues with folks who may have already had cilium repo on their machines, we add the repo under a separate name, make sure to update it, then delete the repo when we're done.

Additionally, since the affinity for the operator is set to disallow scheduling multiple pods on the same node, set our replica count to 1 ([ref](https://github.com/cilium/cilium/blob/v1.15.5/install/kubernetes/cilium/values.yaml#L2536))


davidjumani (dismissed):

davidjumani (approved):

sheidkamp (approved):
Nit: update PR description from boiler plate

PR #9692, opened 2024-06-26
Remove glooctl init to not automatically register types to global scheme

jbohanon:
# Description

Having the scheme registration for gloo fed APIs in the init causes issues when importing since importing code requires registering the same APIs from a different source. 

In this PR we revert registration to just before needed for an actual call. We also use a local scheme for the client that requires it, and remove the registration of kube gw APIs since we use the library client directly instead of controller-runtime.

## Code changes
- Do not register Kube GW APIs in scheme
- Do not use global scheme

# Context

See internal slack conversation [here](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1719415728479849)

## Interesting decisions
 
## Testing steps

I manually verified behavior by importing this into the location in solo-projects where it was previously failing ([link](https://github.com/solo-io/solo-projects/pull/6453))

# Checklist:

### DIFF
--- a/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
+++ b/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
@@ -20,7 +20,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


projects/gloo/cli/pkg/cmd/check/gloo_instances.go
```diff
@@ -105,8 +105,12 @@ type unstructuredGlooInstanceReader struct {
 }
 
 func getUnstructuredGlooInstanceReader(cfg *rest.Config) (*unstructuredGlooInstanceReader, error) {
+	scheme := runtime.NewScheme()
```
jbohanon, line 108: by using a `NewScheme()` we don't risk trampling on anything else that wants to register APIs to the global

jenshu (approved):

nfuden (approved):

PR #9694, opened 2024-06-26
allow adding arbitrary filters in Proxy

stevenctl:
# Description

Gloo uses [WellKnownFilterStage](https://github.com/solo-io/gloo/blob/1f745e52d7d921d19419a6c77a474a38ec1dfe47/projects/gloo/pkg/plugins/staged_filters.go#L27) internally to determine the ordering of Envoy config generated by plugins. It’s exposed in ExtProc API as [FilterStage](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/filters/stages.proto) which may be inserted at varying stages. Building on top of this system, we can add a list of staged filters directly on the TCPListenerOptions and HTTPListenerOptions

## API changes
-  Created `CustomEnvoyFilter` 
- Added list of `CustomEnvoyFilter` to `TCPListenerOptions` (for network filters)
- Added two lists of `CustomEnvoyFilter` to `HTTPListenerOptions (for both http and network filters)

## Code changes
-  When building any kind of staged filters, append the ones from these new lists. 

## Docs changes
- Generated proto docs changes
- We do not want to publicly document this "feature". 

## Testing steps

- Mostly already covered by staged filter tests
- Added one additional translator test for this specific API

## Notes for reviewers

Should we gate this with some Settings or an env variable? It's a pretty dangerous API for users to directly interact with. 

# Checklist:


nfuden:
Note that the approach that is being set here means that we will both document this feature and literally have it in our protos.

EItanya:
What if we locate these alongside the options structs so they are accessible to us, but not present on the CRs? For instance here: https://github.com/solo-io/gloo/blob/eitanya/ai-gateway/projects/gloo/api/v1/proxy.proto#L187

projects/gloo/pkg/plugins/custom_filters/plugin.go
EItanya, line 1: This functionality seems like it should be integrated directly into the `translator`, but I'm definitely flexible on that. WDYT @nfuden

nfuden: I think my main question on this is where we see this in the future. 

Namely if we think we are building out more options to wrap in the `customenvoyfilter` then I think a plugin is a fine place otherwise the translator seem to make more sense for this break glass.

This whole thing doesn't feel like its an extension and more of a piece of core functionality as we arent exposing options out to the user but rather using it as an internal dependency.

The other rationale for having it be part of the translator means that we don't add a new inter-plugin dependency as we try to make plugins (emphasis on try) independant.

EItanya: My guess is that we're gonna create these on the route and other levels eventually, which makes me think having a bespoke location in the translator will lead to more consistent behavior, and less inter plugin deps like you mentioned.

@stevenctl would it be difficult to instead add this to the core translator?

stevenctl: Done, PTAL

projects/gloo/pkg/translator/translator_test.go
```diff
@@ -3980,6 +3918,76 @@ var _ = Describe("Translator", func() {
...
+		It("http", func() {
+			hybridListener := proxy.Listeners[2]
+			httpListener := hybridListener.GetHybridListener().MatchedListeners[1].GetHttpListener()
+			httpListener.CustomNetworkFilters = []*v1.CustomEnvoyFilter{{
+				FilterStage: &filters.FilterStage{
+					Stage:     filters.FilterStage_AuthZStage,
+					Predicate: filters.FilterStage_After,
+				},
+				Name:   "my-custom-filter",
+				Config: &anypb.Any{},
```
EItanya, line 3931: Can you add some actual config here and below just to make sure it's passed through exactly as is

EItanya (commented):
1 small testing nit and then I'm good to approve

EItanya (approved):

nfuden (approved):
Eh seems to work well enough

PR #9698, opened 2024-06-27
Remove glooctl init to not automatically register types to global scheme

jbohanon:
Forward-port of https://github.com/solo-io/gloo/pull/9692

# Description

Having the scheme registration for gloo fed APIs in the init causes issues when importing since importing code requires registering the same APIs from a different source. 

In this PR we revert registration to just before needed for an actual call. We also use a local scheme for the client that requires it, and remove the registration of kube gw APIs since we use the library client directly instead of controller-runtime.

## Code changes
- Do not register Kube GW APIs in scheme
- Do not use global scheme

# Context

See internal slack conversation [here](https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1719415728479849)

## Interesting decisions
 
## Testing steps

I manually verified behavior by importing this into the location in solo-projects where it was previously failing ([link](https://github.com/solo-io/solo-projects/pull/6453))

# Checklist:

### DIFF
--- a/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
+++ b/projects/gloo/cli/pkg/cmd/check/gloo_instances.go
@@ -19,8 +19,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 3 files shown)


jbohanon:
/kick 
```log
Step #1 - "run-tests": Step #0 - "prepare-workspace": Cloning into 'gloo'...
Step #1 - "run-tests": Step #0 - "prepare-workspace": fatal: unable to access 'https://www.github.com/solo-io/gloo.git/': Failed to connect to www.github.com port 443 after 129383 ms: Couldn't connect to server
Step #1 - "run-tests": Step #0 - "prepare-workspace": {"level":"fatal","ts":"2024-06-27T01:06:39.292Z","caller":"src/main.go:26","msg":"unable to complete cloud build preparation","error":"exit status 128","stacktrace":"main.main\n\t/src/main.go:26\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271"}
```

jenshu (approved):

projects/gloo/cli/pkg/cmd/check/gloo_instances.go
```diff
@@ -105,8 +105,12 @@ type unstructuredGlooInstanceReader struct {
...
 
 func getUnstructuredGlooInstanceReader(cfg *rest.Config) (*unstructuredGlooInstanceReader, error) {
+	scheme := runtime.NewScheme()
+	if err := glooinstancev1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
 	client, err := client.New(cfg, client.Options{
-		Scheme: scheme.Scheme,
+		Scheme: scheme,
 	})
```
nfuden, line 114: Can we slap this into a once?

jbohanon: Unless we want to make this whole client package-level I don't think we gain anything from this because
1. We create a new scheme
2. add our API to the scheme
3. Create a new client
4. Return the wrapped client

nfuden (approved):

PR #9716, opened 2024-07-02
[wip] allow overriding StartFuncs from Extensions

jenshu:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/projects/gloo/pkg/syncer/setup/setup_syncer.go
+++ b/projects/gloo/pkg/syncer/setup/setup_syncer.go
@@ -1035,6 +1035,12 @@ func RunGlooWithExtensions(opts bootstrap.Opts, extensions Extensions) error {
 		}
 	}
 
+	// merge in any StartFuncs specified by the Extensions (will overwrite exiting funcs with the same name)
+	if len(extensions.StartFuncs) > 0 {
+		for k, v := range extensions.StartFuncs {
+			startFuncs[k] = v
+		}
+	}
 	ExecuteAsynchronousStartFuncs(
 		watchOpts.Ctx,
 		opts,
--- a/projects/gloo/pkg/syncer/setup/extensions.go
+++ b/projects/gloo/pkg/syncer/setup/extensions.go
@@ -34,6 +34,11 @@ type Extensions struct {
 	// K8sGatewayExtensionsFactory is the factory function which will return an extensions.K8sGatewayExtensions
 	// This is responsible for producing the extension points that the K8s Gateway integration requires
 	K8sGatewayExtensionsFactory extensions.K8sGatewayExtensionsFactory
+
+	// StartFuncs are functions that should be executed at Gloo startup, in addition to the ones
+	// defined by OSS gloo. If a StartFunc with the same key is already defined by OSS gloo, the one
+	// specified here will override it.
+	StartFuncs map[string]StartFunc
 }
 
 // Validate returns an error if the Extensions are invalid, nil otherwise
@@ -50,6 +55,9 @@ func (e Extensions) Validate() error {
 	}
 	if e.SyncerExtensions == nil {
 		return ErrNilExtension("SyncerExtensions")
+	}
+	if e.StartFuncs == nil {
+		return ErrNilExtension("StartFuncs")
 	}
 
 	return nil
--- a/projects/gloo/pkg/syncer/setup/extensions_test.go
+++ b/projects/gloo/pkg/syncer/setup/extensions_test.go
@@ -32,7 +32,7 @@ var _ = Describe("Extensions", func() {
 			},
 			ApiEmitterChannel: nil,
 		}, MatchError(ErrNilExtension("ApiEmitterChannel"))),
-		Entry("missing ApiEmitterChannel", Extensions{
+		Entry("missing SyncerExtensions", Extensions{
 			K8sGatewayExtensionsFactory: extensions.NewK8sGatewayExtensions,
 			PluginRegistryFactory: func(ctx context.Context) plugins.PluginRegistry {
 				// non-nil function
@@ -41,6 +41,15 @@ var _ = Describe("Extensions", func() {
 			ApiEmitterChannel: make(chan struct{}),
 			SyncerExtensions:  nil,
 		}, MatchError(ErrNilExtension("SyncerExtensions"))),
+		Entry("missing StartFuncs", Extensions{
+			K8sGatewayExtensionsFactory: extensions.NewK8sGatewayExtensions,
+			PluginRegistryFactory: func(ctx context.Context) plugins.PluginRegistry {
+				// non-nil function
+				return nil
+			},
+			ApiEmitterChannel: make(chan struct{}),
+			SyncerExtensions:  make([]syncer.TranslatorSyncerExtensionFactory, 0),
+		}, MatchError(ErrNilExtension("StartFuncs"))),
 		Entry("missing nothing", Extensions{
 			K8sGatewayExtensionsFactory: extensions.NewK8sGatewayExtensions,
 			PluginRegistryFactory: func(ctx context.Context) plugins.PluginRegistry {
@@ -49,6 +58,7 @@ var _ = Describe("Extensions", func() {
 			},
 			ApiEmitterChannel: make(chan struct{}),
 			SyncerExtensions:  make([]syncer.TranslatorSyncerExtensionFactory, 0),
+			StartFuncs:        map[string]StartFunc{},
 		}, BeNil()),
 	)
 


projects/gateway2/controller/start.go
```diff
@@ -109,6 +107,7 @@ func Start(ctx context.Context, cfg StartConfig) error {
 
 	k8sGwExtensions, err := cfg.ExtensionsFactory(ctx, extensions.K8sGatewayExtensionsFactoryParameters{
 		Mgr:                     mgr,
+		SnapshotHistory:         cfg.SnapshotHistory,
```
sam-heilbron, line 110: This couples the SnapshotHistory (a generic concept used for debugging the control plane) to the K8s Gateway Extensions. Is this what we want to do?

projects/gateway2/extensions/extensions.go
```diff
@@ -22,6 +22,7 @@ type K8sGatewayExtensions interface {
 // K8sGatewayExtensionsFactoryParameters contains the parameters required to start Gloo K8s Gateway Extensions (including Translator Plugins)
 type K8sGatewayExtensionsFactoryParameters struct {
 	Mgr                     controllerruntime.Manager
+	SnapshotHistory         iosnapshot.History
```
sam-heilbron, line 25: Can you help me understand the intended dependency injection model? Does EE create a custom implementation that extends the OpenSource solution, or do we inject specific functionality. It seems like we're taking the latter approach which makes sense for this case, but I wonder if we'll find that there are many more situations that we need enterprise fucntionality to be injected

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -23,28 +18,51 @@ import (
...
+	// SetExtraKubeGvks optionally sets a list of extra GVKs to return in the input snapshot when
+	// Kubernetes Gateway integration is enabled, in addition to the default GVKs specified by
+	// `KubeGatewayDefaultGVKs`.
+	//
+	// For example, this may include additional Gateway API resources, Portal resources, or other
+	// resources specific to the Kubernetes Gateway integration.
+	//
+	// If this is not set, then only the default GVKs specified by `KubeGatewayDefaultGVKs` will
+	// be returned from `GetInputSnapshot` (in addition to Edge ApiSnapshot resources).
+	SetExtraKubeGvks(extraGvks []schema.GroupVersionKind)
```
sam-heilbron, line 39: Did you consider making the KubeGvks something that goes into the construction of the object, and not on the object itself? That is, you constrcut a History object and at that time specifically say "these are the GVKs I care about". This doesn't seem like something that can/should change during the runtime

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -23,28 +18,51 @@ import (
...
 	// GetXdsSnapshot returns the entire cache of xDS snapshots
 	// NOTE: This contains sensitive data, as it is the exact inputs that used by Envoy
 	GetXdsSnapshot(ctx context.Context) ([]byte, error)
 }
 
 // NewHistory returns an implementation of the History interface
 func NewHistory(cache cache.SnapshotCache, settings *gloov1.Settings) History {
+	// initialize kube gvks to contain a copy of the default gvks
+	kubeGvks := []schema.GroupVersionKind{}
+	kubeGvks = append(kubeGvks, KubeGatewayDefaultGVKs...)
```
sam-heilbron, line 58: Inline with my earlier comment, I wonder if we should make this injectable

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -144,6 +182,10 @@ func (h *historyImpl) GetProxySnapshot(ctx context.Context) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// sort all the resources before formatting
+	sortResources(resources)
```
sam-heilbron, line 187: Given that we call sort before formatOutput each time, should we just consolidate those fucntions so developers don't forget to sort the resources?

sam-heilbron (commented):
Looking good!

projects/gloo/pkg/servers/iosnapshot/history.go
jenshu: my first attempt was having enterprise code create its own History object, and have it override `startFuncs["admin-server"]` with its own `AdminServerStartFunc(history)` (i.e. overriding [this](https://github.com/solo-io/gloo/blob/0cbdbd795ae3b7f5f39f87dddb873a3c58411e24/projects/gloo/pkg/syncer/setup/setup_syncer.go#L899)) but i realized that the EE history probably then wouldn't be able to hook into the translator syncer [here](https://github.com/solo-io/gloo/blob/0cbdbd795ae3b7f5f39f87dddb873a3c58411e24/projects/gloo/pkg/syncer/setup/setup_syncer.go#L940) which is needed to get the api snapshot updates.

you're right that it shouldn't need to change during runtime, but i couldn't think of a cleaner solution that reused the same History object that was instantiated in the OSS setup syncer

this way was slightly simpler and followed the pattern that was used for setting the kube gateway client

projects/gateway2/extensions/extensions.go
jenshu: yep we are reusing the existing History created in the OSS code, and just injecting it with different gvks - see corresponding solo-projects PR [here](https://github.com/solo-io/solo-projects/pull/6577/)

projects/gateway2/controller/start.go
jenshu: hmm, i'm not sure. the existing K8sGatewayExtensionsFactoryParameters already had a mix of different things like clients, status reporter, and a kick function. i thought this would be a good place to stuff this extra object, since it's already passed through to the enterprise controller. conceptually, i don't know whether it makes sense or fits in with the original intentions of the extensions/params

projects/gloo/pkg/servers/iosnapshot/history.go
jenshu: c69ce4a

jmhbh (approved):
LGTM

projects/gateway2/extensions/extensions.go
```diff
@@ -42,6 +43,7 @@ func NewK8sGatewayExtensions(
 ) (K8sGatewayExtensions, error) {
 	return &k8sGatewayExtensions{
 		mgr:                     params.Mgr,
+		snapshotHistory:         params.SnapshotHistory,
```
sam-heilbron, line 46: How is the snapshot history used in this extensions object? This is set on the object, but then never used in the CreatePluginRegistry.

It feels like a smell to me, if we have implementation details that must exist in our open source solution, but aren't used, just to support some enterprise customizations.

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: What concerns me, is that this is a function which is not invoked anywhere in the open source code base. To me, this is a signal that our interface is breaking the open/closed principle and/or the Interface segregation principle.

WHat do you think about something like https://github.com/solo-io/gloo/compare/extensions-startfuncs...extensions-startfuncs-sam?expand=1 which attemps to break up the interface into smaller parts

jmhbh (approved):

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -464,8 +464,9 @@ func RunGloo(opts bootstrap.Opts) error {
 			ratelimitExt.NewTranslatorSyncerExtension,
 			extauthExt.NewTranslatorSyncerExtension,
 		},
-		ApiEmitterChannel: make(chan struct{}),
-		XdsCallbacks:      nil,
+		ApiEmitterChannel:      make(chan struct{}),
+		XdsCallbacks:           nil,
+		SnapshotHistoryFactory: iosnapshot.GetHistoryFactory(),
```
sam-heilbron, line 469: Do we need to add this property to our extensions.Validate function? https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/syncer/setup/extensions.go#L40

sam-heilbron (dismissed):
lgtm, one question about validation

projects/gloo/pkg/syncer/setup/setup_syncer.go
jenshu: ahh yes thanks

jenshu: 0514398

jmhbh (approved):

sam-heilbron (approved):

PR #9717, opened 2024-07-02
gateway2: misc fixes for AI extension

shashankram:
# Description

Removes listenAddress from GatewayParameters because it
will be set via aiExtension.env. Adds ports to aiExtension
GatewayParameter field to configure container ports.

# Checklist:


projects/gateway2/api/v1alpha1/gateway_parameters.proto
```diff
@@ -245,19 +245,19 @@ message AiExtension {
   // values_template.yaml with any user overrides
   kube.gateway.gloo.solo.io.Image image = 2;
 
-  // Address the extension will listen on.
-  google.protobuf.StringValue listen_address = 3;
```
jenshu, line 249: to avoid a breaking change, it would be safer if we set this to `reserved 3;` and keep the existing numbers the same

shashankram: Discussed offline. No one is using this and since this was only added in the beta release, should be okay.

EItanya (approved):

jenshu (approved):

PR #9718, opened 2024-07-02
Move glooctl tests to new framework

npolshakova:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:


.github/workflows/nightly-tests.yaml
```diff
@@ -304,7 +304,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kube-e2e-test-type: ['gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'glooctl', 'upgrade']
+        kube-e2e-test-type: ['gateway', 'gloo', 'ingress', 'helm', 'gloomtls', 'upgrade']
```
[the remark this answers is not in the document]
npolshakova: Yep! Once it gets backported!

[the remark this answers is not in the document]
npolshakova: I think we at least want to backport it to 1.17. Good question about other LTS branches. @jbohanon what are you doing for your e2e tests?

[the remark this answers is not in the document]
jbohanon: I don't think we should take any of this further than 1.17. That's as old as the `test/kubernetes/e2e` is.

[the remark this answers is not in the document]
npolshakova: Updated!

jbohanon (changes requested):
Should we leave the existing tests in place for a bit?

davidjumani (dismissed):

test/kubernetes/e2e/features/glooctl/check_crds_suite.go
```diff
@@ -0,0 +1,67 @@
...
+	return &checkCrdsSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+	}
+}
+
+func (s *checkCrdsSuite) SetupSuite() {
+	var err error
+
+	s.tmpDir, err = os.MkdirTemp(s.testInstallation.GeneratedFiles.TempDir, "debug-suite-dir")
```
sam-heilbron, line 35: What's the reason we create a sub temp directory instead of just using the original TempDir?

test/kubernetes/e2e/features/glooctl/check_crds_suite.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (s *checkCrdsSuite) TestValidatesCorrectCrds() {
+	if s.testInstallation.Metadata.ReleasedVersion != "" {
+		err := s.testInstallation.Actions.Glooctl().CheckCrds(s.ctx,
+			"-n", s.testInstallation.Metadata.InstallNamespace,
+			"--kube-context", s.testInstallation.ClusterContext.KubeContext,
+			"--version", s.testInstallation.Metadata.ChartVersion)
+		s.NoError(err)
+	} else {
+		chartUri := filepath.Join(testutils.GitRootDirectory(), s.testInstallation.Metadata.TestAssetDir, s.testInstallation.Metadata.HelmChartName+"-"+s.testInstallation.Metadata.ChartVersion+".tgz")
```
sam-heilbron, line 51: nit: I would love it if this tests/our tests didn't have to know how to construct this file. Could we instead pass a chartArtifact to the testInstallation or bake this into our glooctl test tool itself?

test/kubernetes/e2e/test.go
```diff
@@ -28,6 +28,14 @@ func MustTestHelper(ctx context.Context, installation *TestInstallation) *helper
 
 	testHelper.SetKubeCli(installation.ClusterContext.Cli)
 
+	// TODO(npolshak) We should remove the test helper and have the test installation own this
```
sam-heilbron, line 31: This scares me a little. I agree with the comment and am happy to defer to you if you think this first step is necessary to make progress on our tests. It feels like we're temporarily passing legacy code forward, and my fear is that we will not find time to remove it

test/kubernetes/testutils/gloogateway/context.go
```diff
@@ -6,7 +6,25 @@ type Context struct {
...
+
 	// whether or not the validation webhook is configured to always accept resources,
 	// i.e. if this is set to true, the webhook will accept regardless of errors found during validation
 	ValidationAlwaysAccept bool
+
+	// TestAssetDir is the directory holding the test assets. Must be relative to RootDir.
+	TestAssetDir string
+
+	// Helm chart name
+	HelmChartName string
```
sam-heilbron, line 20: A fear I have here is that we are exposing implementation details (helm) on a structure that is intentionally unaware of the implementation. My original idea was that you could then install using Argo, Flux or some other GitOps tool, and it would be easy to then run our existing tests using different tools and not have to write code. Perhaps it's not feasilble, but how would you feel about at least packaging this into a Helm specfic struct?

npolshakova: I think that's out of scope until we rip out the testHelper. I know @jbohanon  is working on some related tooling changes alongside the upgrade test port, so I think this PR should be limited to getting glooctl functionality in to avoid conflicts with parallel work.

test/kubernetes/e2e/features/glooctl/check_crds_suite.go
npolshakova: Yep, we can have testInstallation have the chartUri baked in. I originally tried to limit the new fields added, but it would hide calculating the chart uri logic from the test itself. I'm not sure this would be something glooctl should keep track of? It seems better to have it either version dependent or require the user to specify the path.

test/kubernetes/e2e/test.go
jbohanon: In accordance with [this comment](https://github.com/solo-io/solo-projects/issues/6303#issuecomment-2211086767), we won't consider this work done until the test helper is factored out of the new test code

jbohanon (approved):

test/kubernetes/e2e/features/glooctl/check_crds_suite.go
npolshakova: Yep, I think this is only used for the debug suite, don't need it here. Removed: https://github.com/solo-io/gloo/pull/9718/commits/8edc7a809b9747461f283fd454c641df4f0efccd

jbohanon (approved):

jbohanon (dismissed):

test/kubernetes/e2e/features/glooctl/check_crds_suite.go
npolshakova: https://github.com/solo-io/gloo/pull/9718/commits/81422151f5e87401e4bae4efef5d9b1a873b6df2

jbohanon (approved):

sam-heilbron (approved):

PR #9720, opened 2024-07-03
Migrate utils from `test/kube2e` to `test/kubernetes`

jbohanon:
# Description

## API changes
- Added x field to y resource
- ...

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:


pkg/utils/fsutils/fsutils.go
```diff
@@ -0,0 +1,32 @@
+package fsutils
+
+import (
+	"fmt"
+	"os"
+)
+
+func ToTempFile(content string) (string, error) {
```
jbohanon, line 8: godoc exported funcs

test/kubernetes/e2e/defaults/testdata/tcp_echo.yaml
```diff
@@ -0,0 +1,37 @@
...
+kind: Pod
+metadata:
+  name: tcp-echo
+  namespace: tcp-echo
+  labels:
+    app.kubernetes.io/name: tcp-echo
+spec:
+  containers:
+  - name: tcp-echo
+    image: soloio/tcp-echo:latest
```
jbohanon, line 29: nit: pin a tag; just because the old way used latest doesn't mean we have to

test/kubernetes/testutils/helper/docker/Dockerfile
jbohanon, line 1: add some comment or something about what this is for. I assume `soloio/tcp-echo` but not totally certain

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,308 @@
...
+	Verbose        bool
+}
+
+func ExtraArgs(args ...string) func(*InstallOptions) {
+	return func(io *InstallOptions) {
+		io.GlooctlCommand = append(io.GlooctlCommand, args...)
+	}
+}
+
+// Installs Gloo (and, optionally, the test server)
```
jbohanon, line 163:
```suggestion
// InstallGloo calls glooctl to install Gloo. This is where image variants are handled as well.
```

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/gloo/pkg/utils/fsutils"
+	"github.com/solo-io/gloo/pkg/utils/kubeutils/kubectl"
+	"github.com/solo-io/go-utils/log"
+	"github.com/solo-io/go-utils/testutils/exec"
+)
+
+const (
+	GATEWAY = "gateway"
+)
```
jbohanon, line 25: can remove this from here and the sig for `InstallGloo` below

test/kubernetes/e2e/defaults/testdata/tcp_echo.yaml
jbohanon: welp, no tag to pin https://hub.docker.com/r/soloio/tcp-echo

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,420 @@
...
+
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/gloo/pkg/utils/fsutils"
+	"github.com/solo-io/gloo/pkg/utils/kubeutils/kubectl"
+	"github.com/solo-io/go-utils/log"
+	"github.com/solo-io/go-utils/testutils/exec"
+)
+
+// Default test configuration
+var defaults = TestConfig{
```
npolshakova, line 26: 🎉 Nice! I need this in the glooctl PR as well.

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+		_, current, err := upgrade.GetUpgradeVersions(ctx, repoName)
+		Expect(err).NotTo(HaveOccurred())
+		return current.String()
+	}
+
+	// Assume that releasedVersion is a valid version, for a previously released version of Gloo Edge
+	return releasedVersion
+}
+
+func GetTestHelperForRootDir(ctx context.Context, rootDir, namespace string) (*SoloTestHelper, error) {
```
npolshakova, line 112: Can we add this as a helper function to get the chart path somewhere: 
```
chartUri = filepath.Join(testHelper.RootDir, testHelper.TestAssetDir, testHelper.HelmChartName+"-"+testHelper.ChartVersion()+".tgz")
		
```

We keep recalculating it, but one testInstallation should have the same chartUri.

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+func GetHttpEchoImage() string {
+	httpEchoImage := "hashicorp/http-echo"
+	if runtime.GOARCH == "arm64" {
+		httpEchoImage = "gcr.io/solo-test-236622/http-echo"
+	}
+	return httpEchoImage
+}
+
+// GlooctlCheckEventuallyHealthy will run up until proved timeoutInterval or until gloo is reported as healthy
+func GlooctlCheckEventuallyHealthy(offset int, testHelper *SoloTestHelper, timeoutInterval string) {
```
npolshakova, line 40: We don't need this anymore for the glooctl tests since we have `checkOutputByKey`

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// This response is given by the nginx pod defined in test/kubernetes/e2e/defaults/
+func TestServerHttpResponse() string {
+	return e2edefaults.NginxResponse
+}
+
+// For nightly runs, we want to install a released version rather than using a locally built chart
+// To do this, set the environment variable RELEASED_VERSION with either a version name or "LATEST" to get the last release
+func GetTestReleasedVersion(ctx context.Context, repoName string) string {
```
npolshakova, line 93: Nice- I also need this for one of the glooctl tests.

Do we use this to set `ReleasedVersion` on the testHelper still?
```
		testHelper, err = helper.NewSoloTestHelper(func(defaults helper.TestConfig) helper.TestConfig {
			defaults.RootDir = filepath.Join(cwd, "../../..")
			defaults.HelmChartName = "gloo"
			defaults.InstallNamespace = namespace
			defaults.Verbose = true
			defaults.ReleasedVersion = targetVersion
			return defaults
		})
```

Otherwise, I can just call GetTestReleasedVersion directly.

test/kubernetes/testutils/assertions/deployments.go
```diff
@@ -25,3 +29,28 @@ func (p *Provider) EventuallyRunningReplicas(ctx context.Context, deploymentMeta
 		WithPolling(time.Millisecond * 200).
 		Should(Succeed())
 }
+
+func (p *Provider) EventuallyGlooReachesConsistentState(installNamespace string) {
+	// We port-forward the Gloo deployment stats port to inspect the metrics and log settings
+	glooStatsForwardConfig := assertions.StatsPortFwd{
```
sam-heilbron, line 35: nit: This stats config predates when we had a better utility for configuration port-forwards (https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/portforward/options.go). If we're in this code, do you think it's worth it to update this implementation to use the new code? Happy to defer to you if you tink we should just leave it as is

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+	ginkgo.By("Gloo eventually reaches a consistent state")
+	offset := 1 // This method is called directly from a TestSuite
+	assertions.EventuallyWithOffsetStatisticsMatchAssertions(offset, glooStatsForwardConfig,
+		logLevelAssertion.WithOffset(offset),
+		emitterMetricAssertion.WithOffset(offset),
+	)
+}
+
+// This response is given by the nginx pod defined in test/kubernetes/e2e/defaults/
+func TestServerHttpResponse() string {
```
npolshakova, line 87: Can we make this name more descriptive to make it clear it's the Nginx response?

test/kubernetes/testutils/assertions/glooctl.go
```diff
@@ -39,6 +40,30 @@ func (p *Provider) EventuallyCheckResourcesOk(ctx context.Context) {
...
 }
 
+// EventuallyMatchesVersion asserts that `glooctl version` eventually responds with the expected server version
+func (p *Provider) EventuallyMatchesVersion(ctx context.Context, serverVersion string) {
+	p.expectGlooGatewayContextDefined()
+
+	p.Gomega.Eventually(func(innerG Gomega) {
+		contextWithCancel, cancel := context.WithCancel(ctx)
+		defer cancel()
+		k := version.NewKube(p.glooGatewayContext.InstallNamespace, p.clusterContext.KubeContext)
```
sam-heilbron, line 50: We do we construct an inner context and call cancel on it?

Could we also just create the kubeClient once outside the eventually?

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+	logLevelAssertion := assertions.LogLevelAssertion(zapcore.InfoLevel)
+
+	// The emitter at some point should stabilize and not continue to increase the number of snapshots produced
+	// We choose 4 here as a bit of a magic number, but we feel comfortable that if 4 consecutive polls of the metrics
+	// endpoint returns that same value, then we have stabilized
+	identicalResultInARow := 4
+	emitterMetricAssertion, _ := assertions.IntStatisticReachesConsistentValueAssertion("api_gloosnapshot_gloo_solo_io_emitter_snap_out", identicalResultInARow)
+
+	ginkgo.By("Gloo eventually reaches a consistent state")
+	offset := 1 // This method is called directly from a TestSuite
```
npolshakova, line 79: Do we still need the offset?

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/solo-io/gloo/test/kube2e/upgrade"
+	e2edefaults "github.com/solo-io/gloo/test/kubernetes/e2e/defaults"
+	"github.com/solo-io/gloo/test/testutils"
+	"github.com/solo-io/go-utils/stats"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// UniqueTestResourceLabel can be assigned to the resources used by kube2e tests
```
sam-heilbron, line 24: I think we should update this comment now that `kube2e` isn't a concept in this space, right?

test/kubernetes/testutils/helper/util.go
jbohanon: Yeah this can be removed entirely. It was left to ease porting code that uses this signature

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+	// This unique label per test run ensures that the generated snapshot is different on subsequent runs
+	// We have previously seen flakes where a resource is deleted and re-created with the same hash and thus
+	// the emitter can miss the update
+	UniqueTestResourceLabel = "gloo-kube2e-test-id"
```
sam-heilbron, line 28: Is this still used? It feels like this was an interesting workaround to some test inconsistency, but does our new model of tests render this useless?

test/kubernetes/testutils/assertions/deployments.go
```diff
@@ -25,3 +29,28 @@ func (p *Provider) EventuallyRunningReplicas(ctx context.Context, deploymentMeta
...
+	// Gloo components are configured to log to the Info level by default
+	logLevelAssertion := assertions.LogLevelAssertion(zapcore.InfoLevel)
+
+	// The emitter at some point should stabilize and not continue to increase the number of snapshots produced
+	// We choose 4 here as a bit of a magic number, but we feel comfortable that if 4 consecutive polls of the metrics
+	// endpoint returns that same value, then we have stabilized
+	identicalResultInARow := 4
+	emitterMetricAssertion, _ := assertions.IntStatisticReachesConsistentValueAssertion("api_gloosnapshot_gloo_solo_io_emitter_snap_out", identicalResultInARow)
+
+	offset := 1 // This method is called directly from a TestSuite
```
npolshakova, line 51: Do we still need the offset here?

test/kubernetes/testutils/helper/util.go
sam-heilbron, line 40: Can/should we use the cli utility that we have? Per https://github.com/solo-io/gloo/blob/main/projects/gloo/cli/pkg/testutils/cli.go#L17 it calls the underlying command itself

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+func GetTestReleasedVersion(ctx context.Context, repoName string) string {
+	releasedVersion := os.Getenv(testutils.ReleasedVersion)
+
+	if releasedVersion == "" {
+		// In the case where the released version is empty, we return an empty string
+		// The function which consumes this value will then use the locally built chart
+		return releasedVersion
+	}
+
+	if releasedVersion == "LATEST" {
```
sam-heilbron, line 102: Is this pattern still used in our tests? Im not sure it is, or if it's necessary

test/gomega/assertions/stats.go
```diff
@@ -77,7 +77,7 @@ func EventuallyWithOffsetStatisticsMatchAssertions(offset int, statsPortFwd Stat
 		portForwarder.WaitForStop()
 	}()
 
-	By("Ensure port-forward is open before performing assertions")
+	// By("Ensure port-forward is open before performing assertions")
```
npolshakova, line 80: Why are these commented out?

sam-heilbron (commented):
Looking good! Happy to defer to the group around migrating existing utilities vs trying to remove any legacy functionality while we're in here.

test/kubernetes/testutils/helper/util.go
jbohanon: Thanks! I will double check if the original version was widely used and delete if not

test/kubernetes/testutils/assertions/deployments.go
davidjumani: Would it be better to port forward to a service if possible? That way there are no glitches if the pod restarts during tests

test/kubernetes/testutils/assertions/glooctl.go
jbohanon: > We do we construct an inner context and call cancel on it?

In case there are any async actions launched by the function we pass the context to. This way they get cleaned up each loop.


> Could we also just create the kubeClient once outside the eventually?

Yeah this is a great idea

test/kubernetes/testutils/assertions/deployments.go
```diff
@@ -25,3 +29,28 @@ func (p *Provider) EventuallyRunningReplicas(ctx context.Context, deploymentMeta
 		WithPolling(time.Millisecond * 200).
 		Should(Succeed())
 }
+
+func (p *Provider) EventuallyGlooReachesConsistentState(installNamespace string) {
+	// We port-forward the Gloo deployment stats port to inspect the metrics and log settings
+	glooStatsForwardConfig := assertions.StatsPortFwd{
+		ResourceName:      "deployment/gloo",
```
davidjumani, line 36: Is it better to port forward to a service? That way there are no glitches if the pod restarts during tests

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,420 @@
...
+	return func(io *UpgradeOptions) {
+		io.HelmCommand = append(io.HelmCommand, args...)
+	}
+}
+
+// InstallGloo calls glooctl to install Gloo. This is where image variants are handled as well.
+func (h *SoloTestHelper) InstallGloo(ctx context.Context, timeout time.Duration, options ...InstallOption) error {
+	deploymentType := "gateway"
+	log.Printf("installing gloo in [%s] mode to namespace [%s]", deploymentType, h.InstallNamespace)
+	glooctlCommand := []string{
```
davidjumani, line 178: Can this be done via helm?

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,420 @@
...
+
+func ExtraArgs(args ...string) func(*InstallOptions) {
+	return func(io *InstallOptions) {
+		io.GlooctlCommand = append(io.GlooctlCommand, args...)
+	}
+}
+
+type UpgradeOption func(*UpgradeOptions)
+
+type UpgradeOptions struct {
```
davidjumani, line 163: Can we use a common Option struct? They both contain fields that serve the same purpose but with different names

jbohanon: Yeah. I pulled in your changes from your other open PR without modification; was there a reason it was done that way to begin with?

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,420 @@
...
+
+	return nil
+}
+
+// Wait for the glooctl install command to respond, err on timeout.
+// The command returns as soon as certgen completes and all other
+// deployments have been applied, which should only be delayed if
+// there's an issue pulling the certgen docker image.
+// Without this timeout, it would just hang indefinitely.
+func glooctlInstallWithTimeout(rootDir string, io *InstallOptions, timeout time.Duration) error {
```
davidjumani, line 224: Can this be combined with the `upgradeGlooWithTimeout` or the implementation extracted out to a common method called by glooctlInstallWithTimeout && upgradeGlooWithTimeout? This essentially runs a command with a timeout

test/kubernetes/testutils/helper/install.go
```diff
@@ -0,0 +1,420 @@
...
+
+// does not pass the --all flag to glooctl uninstall
+func (h *SoloTestHelper) UninstallGloo() error {
+	return h.uninstallGloo(false)
+}
+
+func (h *SoloTestHelper) uninstallGloo(all bool) error {
+	log.Printf("uninstalling gloo...")
+	cmdArgs := []string{
+		filepath.Join(h.BuildAssetDir, h.GlooctlExecName), "uninstall", "-n", h.InstallNamespace, "--delete-namespace", "--release-name", h.HelmChartName,
```
davidjumani, line 355: Can this be done via helm ?

test/kubernetes/testutils/helper/util.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		err := check.CheckResources(contextWithCancel, printers.P{}, opts)
+		if err != nil {
+			return errors.Wrap(err, "glooctl check detected a problem with the installation")
+		}
+		return nil
+	}, timeoutInterval, "5s").Should(BeNil())
+}
+
+func EventuallyReachesConsistentState(installNamespace string) {
```
davidjumani, line 60: Is this a duplicate of https://github.com/solo-io/gloo/pull/9720/files#diff-1c21d0f8b14f7fff7a5e6ba5663d5f59b981aafa8413daa701b6b9db1e083ef4R33 ?

jbohanon: Yes good catch will remove

test/kubernetes/testutils/helper/install.go
davidjumani: Nope, it was just quickly done to show that the tests can be ported to the new framework (since settings needed to be changed via a helm upgrade)

test/gomega/assertions/stats.go
jbohanon: The `By` calls fail when used with `go-test` because they have to be called from within ginkgo nodes


```log
Ginkgo detected an issue with your spec structure
By("Ensure port-forward is open before performing assertions")
/home/jacob/src/github.com/solo-io/gloo/test/gomega/assertions/stats.go:80
  It looks like you are calling By outside of a running spec.  Make sure you
  call By inside a runnable node such as It or BeforeEach and not inside the
  body of a container such as Describe or Context.

  Learn more at: http://onsi.github.io/ginkgo/#documenting-complex-specs-by
```

test/kubernetes/testutils/helper/util.go
jbohanon: This is not needed anymore with the assertion that we have; removed

test/kubernetes/testutils/assertions/deployments.go
jbohanon: Possibly. This port-forward method is outmoded anyway and will be replaced in followup work

test/kubernetes/testutils/helper/util.go
jbohanon: Yes this is called by `GetTestHelperForRootDir` which is used in the testHelper construction in `test.go`.

test/gomega/assertions/stats.go
jbohanon: They've just been removed

test/kubernetes/testutils/assertions/deployments.go
jbohanon: Removed (though the func we now call uses it under the hood but not much to do about that)

test/kubernetes/testutils/helper/util.go
jbohanon: Deleted in favor of existing assertion in the provider

test/kubernetes/testutils/helper/util.go
jbohanon: Added!

test/kubernetes/testutils/helper/util.go
jbohanon: Yep removed the whole const block

test/kubernetes/testutils/helper/util.go
jbohanon: Removed entirely

test/kubernetes/testutils/helper/util.go
jbohanon: I think it is, if not that can be assessed during ports.

test/kubernetes/testutils/helper/install.go
jbohanon: merged them into one struct

test/kubernetes/testutils/helper/install.go
jbohanon: yep, I went with the latter and extracted the impl out to a common func

test/kubernetes/testutils/helper/install.go
```diff
@@ -325,15 +309,15 @@ func (h *SoloTestHelper) RevertGlooUpgrade(ctx context.Context, timeout time.Dur
 		helmCommand = append(helmCommand, "-v")
 	}
 
-	io := &UpgradeOptions{
-		HelmCommand: helmCommand,
-		Verbose:     true,
+	opts := &Options{
```
davidjumani, line 312: nit: This can be extracted into a separate method
```
func generateOpts(options ...OptionsMutator) *Options {
	opts := &Options{
       	Command: helmCommand,
		Verbose: true,
	}
        for _, opt := range options {
		opt(opts)
	}
        return opts
}
```

davidjumani (approved):

sam-heilbron (approved):

PR #9722, opened 2024-07-03
LLM Api

EItanya:
This PR introduces the new API for the LLM focused functionality in Gloo Gateway

Usage examples:
https://github.com/solo-io/ai-gateway-poc/tree/eitanya/new-upstream-api/custom_api


projects/gloo/api/external/envoy/extensions/transformation/transformation.proto
```diff
@@ -293,6 +293,9 @@ message TransformationTemplate {
     // If you want to nest elements inside the body, use dot separator in the
     // extractor name.
     MergeExtractorsToBody merge_extractors_to_body = 6;
+    // A list of keys to merge into the JSON body. If the key already exists in
+    // the body it will be overwritten.
```
shashankram, line 297: What is `keys` referring to here in the context of a request/response body?

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -0,0 +1,200 @@
...
+
+  // Rate limiting configuration to apply to the corresponding routes.
+  // All Rate limiting applied this way will use the input_tokens as the counter
+  // rather than incrementing by 1 for each request.
+  RateLimiting rate_limiting = 3;
+
+  // Retrieval Augmented Generation.
+  RAG rag = 4;
+  // Semantic caching configuration
+  SemanticCaching semantic_caching = 5;
```
shashankram, line 62: s/chaching/cache for consistency across fields

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -0,0 +1,200 @@
...
+  // Field name
+  string field = 1;
+  // Field Value, this can be any valid JSON value
+  google.protobuf.Value value = 2;
+  // Whether or not to override the field if it already exists
+  bool override = 3;
+}
+
+
+message Postgres {
```
shashankram, line 83: Since the collection_name here is specific to where the embeddings are stored for RAG, this should likely be nested under `rag.datastore`. Creating a top-level Postgres Message seems a bit out of place with the collection_name field

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -0,0 +1,200 @@
...
+  string connection_string = 1;
+  string collection_name = 2;
+}
+
+
+message Redis {
```
shashankram, line 89: Similar comment as Postgres, perhaps nest this under semantic_cache.datastore

EItanya: I agree the use-case here is probably only for semantic caching, I will move it

projects/gloo/api/external/envoy/extensions/transformation/transformation.proto
EItanya: This isn't user-facing config, internal to our filter.

projects/gloo/api/v1/enterprise/options/ai/ai.proto
shashankram: You are right that it could be used for both RAG and semantic cache. We could leave it here and iterate later if needed

shashankram (approved):

shashankram (approved):

lgadban (commented):
would be really helpful to see some example yamls if possible

projects/gloo/api/v1/options/transformation/transformation.proto
```diff
@@ -250,6 +250,9 @@ message TransformationTemplate {
     // If you want to nest elements inside the body, use dot separator in the
     // extractor name.
     MergeExtractorsToBody merge_extractors_to_body = 6;
+    // A list of keys to merge into the JSON body. If the key already exists in
```
lgadban, line 253: doesn't https://github.com/solo-io/gloo/pull/9722/files#r1664746194 apply here?

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -0,0 +1,393 @@
...
+import "google/protobuf/struct.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/empty.proto";
+import "extproto/ext.proto";
+option (extproto.equal_all) = true;
+option (extproto.hash_all) = true;
+option (extproto.clone_all) = true;
+
+message SingleAuthToken {
+  message SecretRef {
```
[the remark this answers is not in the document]
EItanya: We do, but not with an optional `key` field. So if we were to re-use `core.ResourceRef` we'd need an additional necessary level of nesting. Not sure what the benefit would be in this case as well.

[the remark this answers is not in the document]
EItanya: I would argue yes, how else does the controller know which key to pick, but we can also just make assumptions at first as well and re-use the same type, no strong thought from me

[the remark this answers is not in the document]
lgadban: my point is more why would people use a combined secret for this and only this?
i.e. are we planning on showing examples etc where we create a single secret with keys for the various APIs?

basically im wondering if there is enough utility here to introduce yet-another-way to do something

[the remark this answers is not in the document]
EItanya: Ok, swapping it

lgadban (dismissed):

lgadban (approved):

shashankram (approved):

PR #9735, opened 2024-07-09
Fix/enforce regex inutils

nfuden:
# Description
Just like how we do it for routes https://github.com/solo-io/gloo/blob/fix/enforce-regex-inutils/projects/gateway/pkg/translator/virtual_service_translator.go#L339 we add regexp compiles to CORS.

Transformations, routes and cors are the three places we are building out regex in oss. Routes are vetted as seen above, Transformations get caught by envoy validation and now with this pr we safely check for CORS.

This prevents a potential NACK

## Code changes
- Added new error returning regex formatter that makes sure that the strings it ingests are valid
- Uses new function for cors
- Fixes test that previously tested cors with what is an invalid configuration.

## Docs changes
-  Updated a transformation doc that was hit during testing out all regexs

# Context
Config gets nacked with things like "*" for cors

## Testing steps

I manually verified behavior by installing old version running through the following steps and then updating the gloo pod with the image from this. 
1.  install gloo with hello world.
2. Edit default virtual service to look like the following 
```yaml
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
kind: VirtualService
metadata:
  creationTimestamp: "2024-07-08T21:48:46Z"
  generation: 86
  name: default
  namespace: gloo-system
  resourceVersion: "126451"
  uid: 1ed06b39-f275-4dea-970a-cfac42b46617
spec:
  virtualHost:
    domains:
    - '*'
    options:
      cors:
        allowOriginRegex:
        - '*'
    routes:
    - matchers:
      - prefix: /all-pets
      options:
        prefixRewrite: /api/pets
      routeAction:
        single:
          upstream:
            name: default-petstore-8080
            namespace: gloo-system
```
This adds cors with a bad repeating operator. This previously would get nacked and now should be rejected.

# Checklist:


pkg/utils/regexutils/regex.go
```diff
@@ -11,11 +12,35 @@ import (
...
+	// Envoy uses the RE2 library for regex matching in google's owned c++ impl.
+	// go has https://pkg.go.dev/regexp which implements RE2 with a single caveat.
+	_, err := regexp.Compile(candidateRegex)
+	return err
+}
+
+// NewRegex creates a new regex matcher with the given regex.
+// It is tightly coupled to envoy's implementation of regex.
+// Wraps NewRegexFromSettings which wraps NewRegexWithProgramSize which leads to the tight coupling.
+// NOTE: Call this after having checked regex with CheckRegexString.
```
sheidkamp, line 36: I like the note, but there's a bunch of places where `NewRegex` is called that didn't get updated to call `CheckRegexString` first. Assuming those are out of scope,  should the note be updated with "if regex validtion is needed?"

sheidkamp (changes requested):
Validated and overall looks good, but we should have some tests too.

pkg/utils/regexutils/regex.go
nfuden: I showed that each of those in oss already have checks in place. Namely we already do this for route validation and use envoy validation for transforms

sheidkamp (approved):

pkg/utils/regexutils/regex.go
```diff
@@ -26,7 +51,10 @@ func NewRegexFromSettings(settings *v1.Settings, regex string) *envoy_type_match
 	return NewRegexWithProgramSize(regex, programsize)
 }
 
-func NewRegexWithProgramSize(regex string, programsize *uint32) *envoy_type_matcher_v3.RegexMatcher {
+// NewRegexWithProgramSize creates a new regex matcher with the given program size.
+// This means its tightly coupled to envoy's implementation of regex.
+// NOTE: Call this after having checked regex with CheckRegexString.
+func NewRegexWithProgramSize(candidateRegex string, programsize *uint32) *envoy_type_matcher_v3.RegexMatcher {
```
bewebi, line 57: nit: I'm not sure this name change is necessary, but if we are going to change it here, we should also change it in `NewRegexFromSettings()`

bewebi (commented):
Looks good, one nit

sheidkamp (approved):

bewebi (approved):

PR #9744, opened 2024-07-10
tests: Set helm history-max to zero (unlimited) and other fixes for enterprise tests

davidjumani:
# Description

Sets the max history for helm upgrades and rollbacks to 0 (no limit) from the default of 10. This is used by tests that quickly reach the max history when changing and later reverting settings.
This is useful in enterprise as there can be over 30 revisions with the need to revert back to revision 1

# Checklist:


sam-heilbron (approved):
LGTM.

Would an alternate solution have been to allow defining arbitrary options, so that the tests that needed to defined a custom `history-max` could do so?

jenshu (dismissed):

jenshu (dismissed):

davidjumani:
> Would an alternate solution have been to allow defining arbitrary options, so that the tests that needed to defined a custom history-max could do so?

I don't believe it will be necessary as setting a wrong max history can result in failing to revert to a specific version
Better to be safe and set it to the maximum

test/kube2e/util.go
```diff
@@ -46,13 +46,13 @@ func GetHttpEchoImage() string {
 }
 
 // GlooctlCheckEventuallyHealthy will run up until proved timeoutInterval or until gloo is reported as healthy
-func GlooctlCheckEventuallyHealthy(offset int, testHelper *helper.SoloTestHelper, timeoutInterval string) {
+func GlooctlCheckEventuallyHealthy(offset int, namespace string, timeoutInterval string) {
```
sam-heilbron, line 49: I like this change! Better to have the parameters that this function accepts be as minimal as possible.

Given that this is a breaking change to an exported signature, it's worth it to open a PR in enterprise that pulls this in. That way if another developer tries to do the update before you, we have the code/commit that shows what they need to do

sam-heilbron (approved):

changelog/v1.18.0-beta11/set-max-history.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+- type: NON_USER_FACING
+  description: Sets the max history for helm upgrades and rollbacks to 0 (no limit) from the default of 10. This is used by tests that quickly reach the max history when changing and later reverting settings.
```
nfuden, line 3:
```suggestion
  description: Configures tests to set the max history for helm upgrades and rollbacks to 0 (no limit) from the default of 10. This is used by tests that quickly reach the max history when changing and later reverting settings.
```

test/kube2e/util.go
davidjumani: [Enterprise PR](https://github.com/solo-io/solo-projects/pull/6479) with changes incorporated

sam-heilbron (approved):

jenshu (approved):

PR #9746, opened 2024-07-10
Enhance debug info for failed kubernetes e2e tests

jbohanon:
# Description

Currently we output only logs from the gloo pod on failure, and subsequent failures can overwrite the prior logs.

Here we hope to expand this failure logging

### DIFF
--- a/test/kubernetes/e2e/test.go
+++ b/test/kubernetes/e2e/test.go
@@ -195,8 +195,27 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context, testNamespace str
 	i.Assertions.Require.NoError(err)
 	defer glooLogFile.Close()
 
-	logsCmd := i.Actions.Kubectl().Command(ctx, "logs", "-n", i.Metadata.InstallNamespace, "deployments/gloo")
-	_ = logsCmd.WithStdout(glooLogFile).WithStderr(glooLogFile).Run()
+	glooLogsCmd := i.Actions.Kubectl().Command(ctx, "logs", "-n", i.Metadata.InstallNamespace, "deployments/gloo")
+	_ = glooLogsCmd.WithStdout(glooLogFile).WithStderr(glooLogFile).Run()
+
+	edgeGatewayLogFilePath := filepath.Join(failureDir, "edge_gateway.log")
+	edgeGatewayLogFile, err := os.Create(edgeGatewayLogFilePath)
+	i.Assertions.Require.NoError(err)
+	defer edgeGatewayLogFile.Close()
+
+	// TODO figure out namespaces? probably safe to use provided namespace for this one
+	edgeGatewayLogsCmd := i.Actions.Kubectl().Command(ctx, "logs", "--all-containers", "-n", testNamespace, "--prefix", "-l", "gloo=gateway-proxy")
+	_ = edgeGatewayLogsCmd.WithStdout(edgeGatewayLogFile).WithStderr(edgeGatewayLogFile).Run()
+
+	kubeGatewayLogFilePath := filepath.Join(failureDir, "edge_gateway.log")
+	kubeGatewayLogFile, err := os.Create(kubeGatewayLogFilePath)
+	i.Assertions.Require.NoError(err)
+	defer kubeGatewayLogFile.Close()
+
+	// TODO figure out namespaces? for now we can expect all kube gateways to be in default
+	// since all currently written are
+	kubeGatewayLogsCmd := i.Actions.Kubectl().Command(ctx, "logs", "--all-containers", "--prefix", "-l", "gloo=kube-gateway")
+	_ = kubeGatewayLogsCmd.WithStdout(kubeGatewayLogFile).WithStderr(kubeGatewayLogFile).Run()
 
 	clusterStateFilePath := filepath.Join(failureDir, "cluster_state.log")
 	clusterStateFile, err := os.Create(clusterStateFilePath)
@@ -224,6 +243,8 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context, testNamespace str
 	kubectlGetResourcesCmd := i.Actions.Kubectl().Command(ctx, "get", strings.Join(resourcesToGet, ","), "-A")
 	_ = kubectlGetResourcesCmd.WithStdout(clusterStateFile).WithStderr(clusterStateFile).Run()
 	clusterStateFile.Write([]byte{'\n'})
+
+	edgeGwLogsFile
 }
 
 // GeneratedFiles is a collection of files that are generated during the execution of a set of tests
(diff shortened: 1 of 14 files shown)


test/kubernetes/e2e/test.go
```diff
@@ -180,18 +181,70 @@ func (i *TestInstallation) UpgradeGlooGateway(ctx context.Context, serverVersion
...
-	_ = logsCmd.WithStdout(glooLogFile).WithStderr(glooLogFile).Run()
+	glooLogsCmd := i.Actions.Kubectl().Command(ctx, "logs", "-n", i.Metadata.InstallNamespace, "deployments/gloo")
+	_ = glooLogsCmd.WithStdout(glooLogFile).WithStderr(glooLogFile).Run()
+
+	edgeGatewayLogFilePath := filepath.Join(failureDir, "edge_gateway.log")
+	edgeGatewayLogFile, err := os.Create(edgeGatewayLogFilePath)
+	i.Assertions.Require.NoError(err)
+	defer edgeGatewayLogFile.Close()
+
+	// TODO figure out namespaces? probably safe to use provided namespace for this one
```
nfuden, line 206: something like https://github.com/solo-io/gloo/pull/9747 
?

jbohanon: Yeah that's a great start thanks! Will iterate on this tonight

nfuden: Let me know if you want that merged into this or if it should just be closed

jbohanon: I merged the PR and made some further changes

test/kubernetes/e2e/test.go
```diff
@@ -185,13 +186,68 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
...
+		"upstreamgroups",
+		"authconfigs",
+		"ratelimitconfigs",
+		"virtualservices",
+		"httproutes",
+		"secrets",
+	}
+	kubectlGetResourcesCmd := i.Actions.Kubectl().Command(ctx, "get", strings.Join(resourcesToGet, ","), "-A")
+	_ = kubectlGetResourcesCmd.WithStdout(clusterStateFile).WithStderr(clusterStateFile).Run()
+	clusterStateFile.Write([]byte{'\n'})
```
jbohanon, line 250: nit:
```suggestion
	clusterStateFile.WriteString("\n")
```

test/kubernetes/e2e/test.go
```diff
@@ -185,13 +186,68 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
 	// The idea here is we want to accumulate ALL information about this TestInstallation into a single directory
 	// That way we can upload it in CI, or inspect it locally
 
-	glooLogFilePath := filepath.Join(i.GeneratedFiles.FailureDir, "gloo.log")
-	glooLogFile, err := os.Create(glooLogFilePath)
+	failureDir := i.GeneratedFiles.FailureDir
+	err := os.Mkdir(failureDir, os.ModePerm)
+	i.Assertions.Require.NoError(err)
```
jbohanon, line 195: This should be checked against `os.IsExist` before asserting so we don't fail on just the directory already being there. This could happen if subsequent suites within the same test installation on the same cluster failed

npolshakova:
Looks good! Do you have an example failure run?

jbohanon:
> Looks good! Do you have an example failure run?

Yes! The debug can be downloaded from here https://github.com/solo-io/gloo/actions/runs/9892716982

test/kubernetes/e2e/test.go
```diff
@@ -189,13 +190,68 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
...
+	clusterStateFilePath := filepath.Join(failureDir, "cluster_state.log")
+	clusterStateFile, err := os.OpenFile(clusterStateFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, os.ModePerm)
+	i.Assertions.Require.NoError(err)
+	defer clusterStateFile.Close()
+
+	kubectlGetAllCmd := i.Actions.Kubectl().Command(ctx, "get", "all", "-A")
+	_ = kubectlGetAllCmd.WithStdout(clusterStateFile).WithStderr(clusterStateFile).Run()
+	clusterStateFile.WriteString("\n")
+
+	resourcesToGet := []string{
```
npolshakova, line 245: Does `glooctl debug` only output the logs? It does seem like this would be something useful to create a debug report with all the resources on the cluster. But out of scope for this- maybe we can create an issue and reference this test as an example?

test/kubernetes/e2e/test.go
```diff
@@ -189,13 +190,68 @@ func (i *TestInstallation) PreFailHandler(ctx context.Context) {
...
+		"gatewayclasses",
+		"gatewayparameters",
+		"routeoptions",
+		"virtualhostoptions",
+		"upstreams",
+		"upstreamgroups",
+		"authconfigs",
+		"ratelimitconfigs",
+		"virtualservices",
+		"httproutes",
```
npolshakova, line 256: [nit] Can we group these by categories (k8s gateway apis, solo apis for k8s gateways, edge apis)

test/kubernetes/e2e/test.go
jbohanon: A generic debug report has been a goal for a while but as you say that is out of scope for this. I really just needed some more insight to see what is going on in failures that we're swarming right now

pkg/utils/kubeutils/kubectl/cli.go
```diff
@@ -84,6 +84,15 @@ func (c *Cli) RunCommand(ctx context.Context, args ...string) error {
 	return c.Command(ctx, args...).Run().Cause()
 }
 
+// Namespaces returns a list of namespaces or an error if one occurred
+func (c *Cli) Namespaces(ctx context.Context) ([]string, error) {
+	stdout, _, err := c.Execute(ctx, "get", "namespaces", "-o", "jsonpath={.items[*].metadata.name}")
+	if err != nil {
+		return nil, err
+	}
+	return strings.Fields(stdout), nil
```
nfuden, line 93: Is this guaranteed to be sorted? If not perhaps we throw in a sort?

jbohanon: Good call. Added the sort to the kubectl call.

test/kubernetes/e2e/test.go
jbohanon: Yeah, good call. Added some more (all of our CRDs), used fully qualified names for explicitness, and sorted them by reasonable categories.

jenshu (approved):

jenshu:
merge

PR #9760, opened 2024-07-15
Flatten resources returned by Admin endpoint

jenshu:
# Description

Part 1 of some updates to the Admin server:
- Convert all resources in the input snapshot / proxy snapshot to kubernetes format (with apiVersion, kind, spec, status)
- Display all resources in a flat list (sorted by gvk, namespace, name)
- Add k8s GatewayClasses and ReferenceGrants to the input snapshot resources

## Testing steps

```
# set up cluster
ci/kind/setup-kind.sh

# install local gloo
helm install -n gloo-system gloo _test/gloo-1.0.0-ci1.tgz --create-namespace --set kubeGateway.enabled=true

# (create some edge and/or kube gateway resources)

k port-forward -n gloo-system deploy/gloo 9091
curl localhost:9091/snapshots/input | jq
curl localhost:9091/snapshots/proxies | jq
```

# Checklist:

### DIFF
--- a/projects/gloo/pkg/servers/iosnapshot/convert.go
+++ b/projects/gloo/pkg/servers/iosnapshot/convert.go
@@ -1,0 +1,161 @@
+package iosnapshot
+
+import (
+	"cmp"
+	"slices"
+
+	gatewayv1 "github.com/solo-io/gloo/projects/gateway/pkg/api/v1"
+	ratelimitv1alpha1 "github.com/solo-io/gloo/projects/gloo/pkg/api/external/solo/ratelimit"
+	gloov1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+	extauthv1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/enterprise/options/extauth/v1"
+	graphqlv1beta1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/enterprise/options/graphql/v1beta1"
+	v1snap "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/gloosnapshot"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd"
+	crdv1 "github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd/solo.io/v1"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
+	"github.com/solo-io/solo-kit/pkg/utils/kubeutils"
+	"github.com/solo-io/solo-kit/pkg/utils/protoutils"
+)
+
+// convert all the resources in the input snapshot, excluding Artifacts, Secrets, and Proxies, into Kubernetes format
+func snapshotToKubeResources(snap *v1snap.ApiSnapshot) ([]crdv1.Resource, error) {
+	resources := []crdv1.Resource{}
+
+	// gloo.solo.io resources
+	for _, upstream := range snap.Upstreams {
+		kubeUpstream, err := gloov1.UpstreamCrd.KubeResource(upstream)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeUpstream)
+	}
+	for _, upstreamGroup := range snap.UpstreamGroups {
+		kubeUpstreamGroup, err := gloov1.UpstreamGroupCrd.KubeResource(upstreamGroup)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeUpstreamGroup)
+	}
+	// Endpoints are only stored in memory and don't have a Kubernetes resource equivalent,
+	// so we do custom conversion here to make the format consistent with the other resources
+	for _, endpoint := range snap.Endpoints {
+		kubeEndpoint, err := convertToKube(endpoint, gloov1.EndpointCrd)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeEndpoint)
+	}
+
+	// gateway.solo.io resources
+	for _, gw := range snap.Gateways {
+		kubeGw, err := gatewayv1.GatewayCrd.KubeResource(gw)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeGw)
+	}
+
+	for _, vs := range snap.VirtualServices {
+		kubeVs, err := gatewayv1.VirtualServiceCrd.KubeResource(vs)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeVs)
+	}
+	for _, rt := range snap.RouteTables {
+		kubeRt, err := gatewayv1.RouteTableCrd.KubeResource(rt)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeRt)
+	}
+	for _, vho := range snap.VirtualHostOptions {
+		kubeVho, err := gatewayv1.VirtualHostOptionCrd.KubeResource(vho)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeVho)
+	}
+	for _, rto := range snap.RouteOptions {
+		kubeRto, err := gatewayv1.RouteOptionCrd.KubeResource(rto)
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeRto)
+	}
+	for _, hgw := range snap.HttpGateways {
+		kubeHgw, err := gatewayv1.MatchableHttpGatewayCrd.KubeResource(hgw)
+		if err != nil {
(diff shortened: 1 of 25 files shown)


projects/gloo/pkg/servers/iosnapshot/convert.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return nil, err
+		}
+		resources = append(resources, *kubeGqlApi)
+	}
+
+	return resources, nil
+}
+
+func convertToKube(resource resources.VersionedResource, crd crd.Crd) (*crdv1.Resource, error) {
```
sam-heilbron, line 138: How does this function differ from https://github.com/solo-io/solo-kit/blob/main/pkg/api/v1/clients/kube/crd/crd.go#L78? Would it be worth it to link to this or other solo-kit code that this is similar to and clarify any differences /reasoning?

projects/gloo/pkg/servers/iosnapshot/convert.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/solo-io/solo-kit/pkg/utils/kubeutils"
+	"github.com/solo-io/solo-kit/pkg/utils/protoutils"
+)
+
+// convert all the resources in the input snapshot, excluding Artifacts and Secrets, into Kubernetes format
+func snapshotToKubeResources(snap *v1snap.ApiSnapshot) ([]crdv1.Resource, error) {
+	resources := []crdv1.Resource{}
+
+	// gloo.solo.io resources
+	for _, upstream := range snap.Upstreams {
```
sam-heilbron, line 25: This seems like the type of implementation that historically we have used go templates to solve. I don't think we necessarily need to do it for this, thoughI think it would be good to have a strategy to ensure that we update this function when types in our snapshot change. https://github.com/solo-io/solo-kit/pull/554 was a way to add a custom function in the ApiSnapshot that would could explore using

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -162,40 +168,46 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
...
+	for _, gvk := range gvks {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(gvk)
+		err := kubeGatewayClient.List(ctx, list)
+		if err != nil {
+			return nil, err
+		}
+		for _, uns := range list.Items {
+			out := crdv1.Resource{}
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.Object, &out)
```
sam-heilbron, line 195: What would cause to an error here? This implementation seems to change the behavior of the old API, so curious if it was intentional: in the past errors are surfaced to the user and the operation continues in a best-effort attempt. now errors will cause the operation to end early...is there a preferred approach either from a UX or product perspective?

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -275,29 +209,53 @@ func setClientOnHistory(ctx context.Context, history History, builder *fake.Clie
...
+}
+
+// return true if the list of resources contains a resource with the given gvk, namespace, and name
+func containsResource(
+	resources []crdv1.Resource,
+	gvk schema.GroupVersionKind,
+	namespace string,
+	name string) bool {
+	return slices.ContainsFunc(resources, func(res crdv1.Resource) bool {
+		return res.GroupVersionKind().Group == gvk.Group &&
```
sam-heilbron, line 244: nit: Should we re-use containsResourceType here?

jenshu: hmm not sure if i can do that, it seems like it will do 2 full searches through all the resources if i call that from within this func

projects/gloo/pkg/servers/iosnapshot/convert.go
jenshu: good call, i meant to comment here. added in 18cbe9c

projects/gloo/pkg/servers/iosnapshot/convert.go
jenshu: yeah was thinking about that. we could probably generate some code, but for now i just added a test that checks the number of fields 18cbe9c to make sure that if a new resource type is added to the snapshot, the test will fail

projects/gloo/pkg/servers/iosnapshot/history.go
jenshu: looks like it might fail on unknown fields.

yeah i was thinking about what was the best way to return errors. since it's a flat list of resources now, i didn't want to just stick error strings in the list of results. also i'd almost rather fail loudly instead of quietly mixing errors in with the partial results. perhaps we can ask for field/product input if there's a preferred way to return the errors, and update it in a follow-up

projects/gloo/pkg/servers/iosnapshot/history_test.go
jbohanon: I don't think it's worth it to refactor it for two small functions colocated like this, but the predicates "containsResourceType" and "containsResourceWithNamespacedName" could be pulled out and composed maybe

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -275,29 +209,53 @@ func setClientOnHistory(ctx context.Context, history History, builder *fake.Clie
 
 	history.SetKubeGatewayClient(builder.WithObjects(gwSignalObject).Build())
 
+	eventuallyContainsResource(ctx, history, wellknown.GatewayGVK, defaults.GlooSystem, "gw-signal")
+}
+
+// check that the input snapshot eventually contains a resource with the given gvk, namespace, and name
+func eventuallyContainsResource(
```
jbohanon, line 216: Nit: could either rename this `eventuallyInputSnapshotContainsResource` or accept a function which returns a list of resources as a parameter?

jenshu: renamed 6b83570

projects/gloo/pkg/servers/iosnapshot/history_test.go
jenshu: refactored the gvk check into another function 6b83570

projects/gloo/pkg/servers/iosnapshot/convert.go
```diff
@@ -0,0 +1,171 @@
...
+	v1snap "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/gloosnapshot"
+	"github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd"
+	crdv1 "github.com/solo-io/solo-kit/pkg/api/v1/clients/kube/crd/solo.io/v1"
+	"github.com/solo-io/solo-kit/pkg/api/v1/resources"
+	"github.com/solo-io/solo-kit/pkg/utils/kubeutils"
+	"github.com/solo-io/solo-kit/pkg/utils/protoutils"
+)
+
+// convert all the resources in the input snapshot, excluding Artifacts and Secrets, into Kubernetes format
+func snapshotToKubeResources(snap *v1snap.ApiSnapshot) ([]crdv1.Resource, error) {
```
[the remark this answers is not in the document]
jenshu: i have this in https://github.com/solo-io/gloo/pull/9766, can either merge that into this PR or keep them separate so it's easier to review incrementally

projects/gloo/pkg/servers/iosnapshot/convert_test.go
```diff
@@ -0,0 +1,22 @@
...
+	v1snap "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/gloosnapshot"
+)
+
+var _ = Describe("Convert", func() {
+
+	It("will fail if the ApiSnapshot has a new top level field", func() {
+		// This test checks whether the ApiSnapshot has a new top-level field.
+		// If the number of fields changes, it should be used as an indication that the
+		// `snapshotToKubeResources` function might need to be updated to support the new field.
+		Expect(reflect.TypeOf(v1snap.ApiSnapshot{}).NumField()).To(
```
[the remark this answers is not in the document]
jenshu: good idea, done! f4d8b9b0081c2d6dbd27b6660c62f6d5049f9118

projects/gloo/pkg/servers/iosnapshot/convert.go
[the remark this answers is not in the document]
jbohanon: I think it's fine to keep the PRs separate. To answer my prior question (based on your linked PR) these resources do not need to converted here  because they are not in the API snapshot and are watched by kube gw controller

sam-heilbron (approved):

PR #9766, opened 2024-07-16
Admin endpoint: return more resources

jenshu:
# Description

Part 2 of updates to admin server input snapshot endpoint:
- Add Settings (passed from bootstrap opts)
- Add GatewayParameters, ListenerOptions, HttpListenerOptions when kubeGateway is enabled
- When kubeGateway is enabled, return all RouteOptions, VirtualHostOptions, AuthConfigs, RateLimitConfigs from the cluster rather than only the ones from the edge input snapshot (since the kube gateway controller watches all resources on the cluster)

# Checklist:


projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -83,19 +90,44 @@ func (h *historyImpl) GetInputSnapshot(ctx context.Context) ([]byte, error) {
...
+	}
+
+	// get the resources from the edge api snapshot
 	resources, err := snapshotToKubeResources(snap)
 	if err != nil {
 		return nil, err
 	}
 
+	// get settings, which is not part of the api snapshot
+	if h.settings != nil {
```
jbohanon, line 114: Should it be considered an error if we have nil settings? Is there a valid state of the system where we would have passed in a nil to the `NewHistory` constructor?

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -135,16 +351,218 @@ var _ = Describe("History", func() {
...
+								},
+							},
+						},
+					},
+				})
+
+				// k8s resources on the cluster (in reality this would be a superset of the ones
+				// contained in the apisnapshot above, but we use non-overlapping resource names
+				// in this test just to show that we are getting the ones from k8s instead of the
+				// snapshot)
```
jbohanon, line 517: This explanation is so clear and helpful :tada:

jbohanon (approved):
This is awesome! :rocket:

I think there is some opportunity for serving a specific resource type on `/snapshots/input/{kind}` or `/snapshots/input/{groupkind}` but even just with jq magic, this is so useful e.g.
```sh
curl -s localhost:9091/snapshots/input | jq '.[] | select(.kind=="GatewayParameters" or .kind=="RouteOption" or .kind=="Settings") | .kind, .spec, .status'
```

sam-heilbron (approved):

projects/gloo/pkg/servers/iosnapshot/history.go
jenshu: if settings was nil, gloo would have crashed before it got here, but yeah we should be defensive and check for it.
i'll handle this in a follow-up where we improve the way we handle/return errors in the admin server endpoints

PR #9771, opened 2024-07-16
Test/xds snap ctx aware and leak detector

sheidkamp:
# Changes for deflaking
These changes came about investigating test flakes.

- Fix `ServeXdsSnapshots` goroutine leak by switching to using `ContextuallyServeXdsSnapshots` which Closes the server when the context is Done.
- Add GoRoutineMonitor which can be used to detect per-test goroutine leaks. Initially this was created to ensure that processes were shut down before starting the next test, and can still be used in that manner, though more focused approaches are preferred.

## Context
[EE PR](https://github.com/solo-io/solo-projects/pull/6578)


projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -220,7 +225,15 @@ func (s *translatorSyncer) ServeXdsSnapshots() error {
 		_, _ = fmt.Fprintf(w, "%+v", prettify(s.latestSnap))
 	})
 
-	return http.ListenAndServe(fmt.Sprintf(":%d", devModePort), r)
+	server := &http.Server{Addr: fmt.Sprintf(":%d", devModePort), Handler: r}
```
sam-heilbron, line 228: I like it! Just calling out that this functionality has been deprecated in favor of https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/servers/admin/server.go, and something then @jenshu may be removing in 1.18/main if we can to avoid having duplicate implementations

test/helpers/debug.go
```diff
@@ -2,7 +2,11 @@ package helpers
 
 import (
 	"testing"
+	"time"
 
+	. "github.com/onsi/gomega"
```
sam-heilbron, line 7: nit: Since we are introducing a component that is tightly coupled to gomega, does it make sense to add this to https://github.com/solo-io/gloo/tree/main/test/gomega? (no strong feelings, just wanted to raise it as an option)

test/helpers/debug.go
```diff
@@ -32,3 +36,78 @@ func DeferredGoroutineLeakDetector(t *testing.T) func(...goleak.Option) {
 		goleak.VerifyNone(t, append(leakOptions, additionalLeakOptions...)...)
 	}
 }
+
+// GoRoutineMonitor is a helper for monitoring goroutine leaks in tests
```
sam-heilbron, line 40: Love this comment!

test/helpers/debug.go
```diff
@@ -32,3 +36,78 @@ func DeferredGoroutineLeakDetector(t *testing.T) func(...goleak.Option) {
...
+func NewGoRoutineMonitor() *GoRoutineMonitor {
+	// Store the initial goroutines
+	return &GoRoutineMonitor{
+		goroutines: gleak.Goroutines(),
+	}
+}
+
+type ExpectNoLeaksArgs struct {
+	AllowedRoutines []types.GomegaMatcher // Additional allowed goroutines to ignore. See CommonLeakOptions for example.
+	Timeouts        []time.Duration       // Additional arguments to pass to Eventually to control the timeout/polling interval
```
sam-heilbron, line 72: nit: Since this is a default, we may want to clarify what the behavior is when this is not set

test/helpers/debug.go
```diff
@@ -32,3 +36,78 @@ func DeferredGoroutineLeakDetector(t *testing.T) func(...goleak.Option) {
...
+	notLeaks := make([]interface{}, len(args.AllowedRoutines)+1)
+	// First element is the initial goroutines
+	notLeaks[0] = m.goroutines
+	// Cast the rest of the elements to interface{}
+	for i, v := range args.AllowedRoutines {
+		notLeaks[i+1] = v
+	}
+
+	// Determine the time intervals to pass to Eventually
+	var (
```
sam-heilbron, line 87: nit: This code is somewhat similar to https://github.com/solo-io/gloo/blob/main/test/helpers/input_resources.go#L97, perhaps we could have a standard ginkgo timeouts function that can accept user-defined inputs andfall back to sensible defaults? may be overkill too

test/helpers/debug.go
```diff
@@ -32,3 +36,78 @@ func DeferredGoroutineLeakDetector(t *testing.T) func(...goleak.Option) {
...
+
+func NewGoRoutineMonitor() *GoRoutineMonitor {
+	// Store the initial goroutines
+	return &GoRoutineMonitor{
+		goroutines: gleak.Goroutines(),
+	}
+}
+
+type ExpectNoLeaksArgs struct {
+	AllowedRoutines []types.GomegaMatcher // Additional allowed goroutines to ignore. See CommonLeakOptions for example.
```
sam-heilbron, line 71: nit: "Additional" to me implies that we do ignore some goroutines by default.

test/helpers/debug.go
```diff
@@ -32,3 +36,78 @@ func DeferredGoroutineLeakDetector(t *testing.T) func(...goleak.Option) {
...
+	Timeouts        []time.Duration       // Additional arguments to pass to Eventually to control the timeout/polling interval
+}
+
+func (m *GoRoutineMonitor) ExpectNoLeaks(args *ExpectNoLeaksArgs) {
```
sam-heilbron, line 75: Does it make sense to have a simple test to prove that this functioanlity works as expected? Since this is the type of code that should only error if things break, it would be nice to know that when we create a break (goroutine leak) that we are notified

sam-heilbron (commented):
Looking good! Mostly questions/nits that don't need to be addressed

changelog/v1.18.0-beta8/devmode-cancel-xds.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NON_USER_FACING
+    description: >-
+      - Resolves issue of HTTPS listeners not aggregating appropriately when defined using the
```
jbohanon, line 4: Is this a copypasta error?

test/helpers/debug.go
jbohanon: also akin to https://github.com/solo-io/gloo/blob/f4d8b9b0081c2d6dbd27b6660c62f6d5049f9118/test/kube2e/helper/curl.go#L67 though the kube2e is on its way out

jbohanon (approved):
I agree that testing the leak detector would be valueable, but I am good with this improvement as-is.

test/helpers/debug.go
sheidkamp: Added tests

changelog/v1.18.0-beta8/devmode-cancel-xds.yaml
sheidkamp: probably. updated

jbohanon (dismissed):

jbohanon (approved):

test/gomega/assertions/assertions_suite_test.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"testing"
+
+	"github.com/solo-io/go-utils/testutils"
+
+	. "github.com/onsi/ginkgo/v2"
+)
+
+func TestHelpers(t *testing.T) {
```
sam-heilbron, line 11: nit:

```suggestion
func TestAssertions(t *testing.T) {
```

sam-heilbron (approved):

jbohanon (approved):

jbohanon (approved):

test/helpers/util.go
```diff
@@ -54,3 +54,79 @@ func PercentileIndex(length, pct int) int {
 
 	return int(math.Ceil(float64(length)*(float64(pct)/float64(100)))) - 1
 }
+
+var (
+	// Gomega defaults from https://github.com/onsi/gomega/blob/master/internal/duration_bundle.go#L27
```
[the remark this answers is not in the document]
sheidkamp: updated

sam-heilbron (dismissed):

nfuden (approved):

sam-heilbron (approved):

sam-heilbron:
bulldozer

PR #9773, opened 2024-07-16
Docs: 1.17 release notes

Rachael-Graham:
Adds relnotes to the upgrade page for 1.17. This will need to be backported to 1.17, and will eventually need to be updated for 1.18.


docs/content/operations/upgrading/faq.md
```diff
@@ -84,15 +94,25 @@ You can use the changelogs' built-in [comparison tool]({{< versioned_link_path f
...
+* **Secret deletion**: Secrets can now be deleted even when warnings or errors are present. When the deletion of a secret is validated, the validating admission webhook removes the secret from the current snapshot, runs translations, and looks for errors. Previously, the validating webhook did not delete a secret if errors were present, or if warnings were present and the `ignore_warnings` setting was set to `false`. The old behavior could cause issues when attempting to delete secrets that were unrelated to the warnings or errors. Now, the validating webhook collects all the artifacts of the valdiation process, re-runs validation against the original snapshot, and compares the artifacts from that process to the artifacts previously collected. If the artifacts are the same, the secret is deleted. If the artifacts are different, the secret is not deleted and errors are returned. Because this process might slightly degrade performance, you can disable this feature by setting the `gloo.gloo.deployment.customEnv.DISABLE_VALIDATION_AGAINST_PREVIOUS_STATE` environment variable to `true` in the your Gloo deployment.
```
Nadine2016, line 110:
```suggestion
* **Secret deletion**: Secrets can now be deleted even when warnings or errors are present. When the deletion of a secret is validated, the validating admission webhook removes the secret from the current snapshot, runs translations, and looks for errors. Previously, the validating webhook did not delete a secret if errors were present, or if warnings were present and the `ignore_warnings` setting was set to `false`. The old behavior could cause issues when attempting to delete secrets that were unrelated to the warnings or errors. Now, the validating webhook collects all the artifacts of the valdiation process, re-runs validation against the original snapshot, and compares the artifacts from that process to the artifacts previously collected. If the artifacts are the same, the secret is deleted. If the artifacts are different, the secret is not deleted and errors are returned. Because this process might slightly degrade performance, you can disable this feature by setting the `gloo.gloo.deployment.customEnv.DISABLE_VALIDATION_AGAINST_PREVIOUS_STATE` environment variable to `true` in the your Gloo Gateway deployment.
```

docs/content/operations/upgrading/faq.md
```diff
@@ -84,15 +94,25 @@ You can use the changelogs' built-in [comparison tool]({{< versioned_link_path f
...
+* **Kubernetes API server unavailability**: The new `MAX_RECOVERY_DURATION_WITHOUT_KUBE_API_SERVER` environment variable defines the maximum duration that the Gloo pod can run and attempt to reconnect to the Kubernetes API server if it is unreachable. If the duration is exceeded, the Gloo pod quits. This means that when leader election is enabled, the Gloo pod falls back to a follower. Previously, the Gloo pod crashed in this situation, which could cause an outage. To set this environment variable, you can either update the Gloo deployment or update the Helm values by specifying the `gloo.deployment.customEnv[0].Name=MAX_RECOVERY_DURATION_WITHOUT_KUBE_API_SERVER` and `gloo.deployment.customEnv[0].Value=60s` values.
+* **Extended HTTP methods for Envoy**: Envoy can now accept requests with extended HTTP methods, such as LABEL or UPDATE. Previously, requests with these methods would return an HTTP 400 response. Note that currently, this functionality is supported for HTTP/1 only.
```
Nadine2016, line 115:
```suggestion
* **Extended HTTP methods for Envoy**: Envoy can now accept requests with extended HTTP methods, such as LABEL or UPDATE. Previously, requests with these methods returned an HTTP 400 response. Note that currently, this functionality is supported for HTTP/1 only.
```

artberger (dismissed):

Nadine2016 (approved):

artberger (approved):

PR #9790, opened 2024-07-19
Add `AssertEventualCurlError`

sheidkamp:
# Description
Add a new assertion `AssertEventualCurlError` that asserts that eventually the curl command results in an application error, such as
```
curl: (7) Failed to connect to gateway-proxy.extauth-test.svc.cluster.local port 80 after 2 ms: Connection refused
command terminated with exit code 7
```

This assertion can be used to verify that a deletions virtual service has been fully propagated, as seen in https://github.com/solo-io/solo-projects/pull/6461


test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,41 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+		g.Expect(err).To(HaveOccurred())
+		if expectedErrorCode > 0 {
+			g.Expect(err.Error()).To(Equal(fmt.Sprintf("exit status %d", expectedErrorCode)))
+		}
+
+	}).
+		WithTimeout(currentTimeout).
+		WithPolling(pollingInterval).
+		WithContext(ctx).
+		Should(Succeed(), "failed to get a curl error")
```
nfuden, line 171: I think we should report the final response code. 
For example if we have an expected 403 but it ends up being a 401 we should hold onto that info. 
I know we already attempt something like this with the printfs but this output will show up in the final summary so its somewhat quicker to parse.

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,41 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: podOpts.Name, Namespace: podOpts.Namespace,
+		},
+	})
+
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+
+	p.Gomega.Eventually(func(g Gomega) {
+		curlResponse, err := p.clusterContext.Cli.CurlFromPod(ctx, podOpts, curlOptions...)
+		fmt.Printf("wanted error code")
```
nfuden, line 157: remove?

nfuden (approved):

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,52 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+		if err != nil {
+			curlHttpResponse := transforms.WithCurlResponse(curlResponse)
+			testMessage = fmt.Sprintf("failed to get a curl error, got response code: %d", curlHttpResponse.StatusCode)
+			g.Expect(err).To(HaveOccurred())
+		}
+
+		if expectedErrorCode > 0 {
+			expectedCurlError := fmt.Sprintf("exit status %d", expectedErrorCode)
+			testMessage = fmt.Sprintf("got curl error: %s, expected %s", err.Error(), expectedCurlError)
+			g.Expect(err.Error()).To(Equal(expectedCurlError))
```
nfuden, line 173: Maybe also a Expect on err not being nil?

sheidkamp: If we're here, we've already passed:
```
if err == nil {
...
  g.Expect(err).To(HaveOccurred())
}
```
so we shouldn't be able to reach this block with a nil err. (this may have been changed after you made this comment)

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,50 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+	testMessage := fmt.Sprintf("Expected curl error %d", expectedErrorCode)
+
+	p.Gomega.Eventually(func(g Gomega) {
+		curlResponse, err := p.clusterContext.Cli.CurlFromPod(ctx, podOpts, curlOptions...)
+
+		if err == nil {
+			fmt.Printf("wanted curl error, got response:\nstdout:\n%s\nstderr:%s\n", curlResponse.StdOut, curlResponse.StdErr)
+			curlHttpResponse := transforms.WithCurlResponse(curlResponse)
+			curlHttpResponse.Body.Close()
```
inFocus7, line 162: should this be deferred? or can we still get the status code? (i'm guessing this is okay since we close the body but the code is stored in StatusCode, but just verifying)

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,50 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: podOpts.Name, Namespace: podOpts.Namespace,
+		},
+	})
+
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
+	testMessage := fmt.Sprintf("Expected curl error %d", expectedErrorCode)
+
+	p.Gomega.Eventually(func(g Gomega) {
+		curlResponse, err := p.clusterContext.Cli.CurlFromPod(ctx, podOpts, curlOptions...)
```
nfuden, line 157: Do we think it makes sense to add a comment or a check around curlResponse.

Today our curlFrompod is guaranteed to not be nil but that may change in the future

test/kubernetes/testutils/assertions/curl.go
```diff
@@ -132,3 +132,50 @@ func (p *Provider) AssertEventuallyConsistentCurlResponse(
...
+	timeout ...time.Duration,
+) {
+	// We rely on the curlPod to execute a curl, therefore we must assert that it actually exists
+	p.EventuallyObjectsExist(ctx, &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: podOpts.Name, Namespace: podOpts.Namespace,
+		},
+	})
+
+	currentTimeout, pollingInterval := helper.GetTimeouts(timeout...)
```
inFocus7, line 153: I see another function here has defaults set for timeouts if no `timeout...`s are passed. like so
```go
	pollTimeout := 3 * time.Second
	pollInterval := 1 * time.Second
	if len(timeout) > 0 {
		pollTimeout, pollInterval = helper.GetTimeouts(timeout...)
	}
```

Should we do something similar here? If not, what would happen when passing an empty `timeout...`?

nfuden (approved):

PR #9794, opened 2024-07-19
settings: deprecate dev mode

sam-heilbron:
# Description

Deprecate the Settings.devMode API

# Context
Ref: https://github.com/solo-io/gloo/issues/6494

We have a standard Admin API that we are introducing. As part of this, we want to point customers towards that API.

## Interesting decisions
 I opted to deprecate the APIs, instead of removing them outright, to reduce the chance of breaking someones pipelines. We can/should remove these in future releases. 

## Testing steps
- [x] I let CI act as my testing steps. These changes were docs related so I don't expect there to be a functional impact.

## Notes for reviewers
-

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6604

PR #9797, opened 2024-07-19
admin/history: expand implementation and tests

sam-heilbron:
# Description

A first pass at incorporating feedback on the Admin API. 

# Context
After a first pass at the Admin API, we received the following feedback: https://github.com/solo-io/solo-projects/issues/6391#issuecomment-2231720204. This PR attempts to incorporate some of that feedback. There will be a PR after this to finish the remaining pieces.

- Update tests to be more explicit about what is expected to exist on an object. The idea is that we will be modifying some of this response, and we want to have stable tests that will be easy to update. Plus, this enables us to re-use matchers in our EE code.
- Update docs to reflect that devMode is deprecated as of 1.17 (I ended up pulling in the code from https://github.com/solo-io/gloo/pull/9794 into this PR, since there was some overlap in what was being done)
- Update endpoints to return errors as part of the body, so that we can more easily return partial responses, if necessary
- Expose an endpoint to return edge-api-only data
- Sanitize artifacts and secrets

## Enterprise Changes
Although this PR changes the signature of public methods, these objects are not imported in Gloo Gateway Enterprise. Instead, only the NewHistory constructor is used, and that signature is not changed. ref: https://github.com/solo-io/solo-projects/blob/main/projects/gloo/pkg/setup/history.go#L36.

https://github.com/solo-io/solo-projects/pull/6652 is what is used to test these changes.

## Interesting decisions
-

## Testing steps
- [x] Build-bot passes

**To run the tests locally:**
```
TEST_PKG=projects/gloo/pkg/servers/iosnapshot/ make test
```

**To run the admin server locally:**
Create a KinD cluster
```
kind create cluster
```
Build the latest images
```
VERSION=1.0.0-sh ci/kind/setup-kind.sh
```

Create  the gloo-system ns:
```
kubectl create ns gloo-system
```

Install GG in the KinD cluster, using standard install values:
```
helm install -n gloo-system gloo _test/gloo-1.0.0-sh.tgz -f test/kubernetes/e2e/tests/manifests/k8s-gateway-test-helm.yaml
```

Open a port-forward to the admin server
```
kubectl port-forward -n gloo-system deployment/gloo 9091:9091
```

Curl the various snapshot endpoints:
```
```
```
{"data":[{"kind":"GatewayParameters","apiVersion":"gateway.gloo.solo.io/v1alpha1","metadata":{"name":"gloo-gateway","namespace":"gloo-system","uid":"4cbf950d-e767-4dc5-9d07-744cc1e74741","resourceVersion":"1449","generation":1,"creationTimestamp":"2024-07-24T20:21:54Z","labels":{"app.kubernetes.io/managed-by":"Helm","gloo":"kube-gateway"},"annotations":{"meta.helm.sh/release-name":"gloo","meta.helm.sh/release-namespace":"gloo-system"}},"status":null,"spec":{"kube":{"aiExtension":{"enabled":false,"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"gloo-ai-extension","tag":"1.0.0-sh"}},"deployment":{"replicas":1},"envoyContainer":{"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"gloo-envoy-wrapper","tag":"1.0.0-sh"},"securityContext":{"allowPrivilegeEscalation":false,"capabilities":{"add":["NET_BIND_SERVICE"],"drop":["ALL"]},"readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":10101}},"istio":{"istioProxyContainer":{"image":{"pullPolicy":"IfNotPresent","registry":"docker.io/istio","repository":"proxyv2","tag":"1.22.0"},"istioDiscoveryAddress":"istiod.istio-system.svc:15012","istioMetaClusterId":"Kubernetes","istioMetaMeshId":"cluster.local","logLevel":"warning"}},"podTemplate":{"extraLabels":{"gloo":"kube-gateway"}},"sdsContainer":{"bootstrap":{"logLevel":"info"},"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"sds","tag":"1.0.0-sh"}},"service":{"type":"LoadBalancer"},"stats":{"enableStatsRoute":true,"enabled":true,"routePrefixRewrite":"/stats/prometheus","statsRoutePrefixRewrite":"/stats"}}}}],"error":""}
```

Can use `jq` to parse the response
```
curl -X GET "127.0.0.1:9091/snapshots/input" | jq
```

## Notes for reviewers
-

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/6604


projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -148,6 +123,55 @@ var _ = Describe("History", func() {
...
+				matchers.HaveObjectMeta(types.NamespacedName{
+					Namespace: defaults.GlooSystem,
+					Name:      "ep-snap",
+				}),
+				gstruct.Ignore(),
+			), "returned resources include endpoints")
+		})
+
+		It("Excludes Secrets", func() {
+			// TODO: We want to update the implementation to include secrets, but redact the contents of them
```
npolshakova, line 140: Do we have an issue to track including secrets and artifacts with redactions?

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -725,22 +922,36 @@ func eventuallyInputSnapshotContainsResource(
 		Should(Succeed(), fmt.Sprintf("snapshot should eventually contain resource %v %s.%s", gvk, namespace, name))
 }
 
+// Deprecated: Prefer matchers.ContainCustomResource (gomega/matchers/custom_resource.go)
 func expectContainsResource(
```
npolshakova, line 926: Is this still being used by the legacy tests?

test/gomega/matchers/custom_resource.go
```diff
@@ -0,0 +1,38 @@
...
+	k8stypes "k8s.io/apimachinery/pkg/types"
+)
+
+// HaveObjectMeta returns a GomegaMatcher which matches a struct that has the provided name/namespace
+// This should be used when asserting that a CustomResource has a provided name/namespace
+func HaveObjectMeta(namespacedName k8stypes.NamespacedName) types.GomegaMatcher {
+	return gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
+		"Name":          Equal(namespacedName.Name),
+		"Namespace":     Equal(namespacedName.Namespace),
+		"ManagedFields": BeEmpty(),
```
npolshakova, line 18: Does this need to be set to empty? Can we just use gstruct.Ignore()?

projects/gloo/pkg/servers/iosnapshot/history_test.go
sam-heilbron: Good call. https://github.com/solo-io/solo-projects/issues/6600 is the work to track this, and originally Jenny was going to work on this in parallel, so it felt like overkill to reference. BUT, I always ask for others to link github issues, so I should do the same! (thanks for keeping me accountable)

projects/gloo/pkg/servers/iosnapshot/history_test.go
sam-heilbron: Good call, there are other tests in this file that still use this I believe. I will update all references in this file though, so we don't have multiple ways to do the same thing

changelog/v1.18.0-beta10/remove-history-managed-fields.yaml
```diff
@@ -0,0 +1,12 @@
+changelog:
+  - type: NON_USER_FACING
+    issueLink: https://github.com/solo-io/solo-projects/issues/6603
+    resolvesIssue: false
+    description: >-
+      Remove managedFields from API response, and expand tests to more easily inspect the state
```
jenshu, line 6: > Remove managedFields from API response

is this accurate? looks like only test files were updated

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -79,59 +91,22 @@ var _ = Describe("History", func() {
...
-			expectContainsResource(returnedResources, v1.UpstreamGVK, defaults.GlooSystem, "upstream-west")
+			returnedResources := getTypedInputSnapshot(ctx, history)
+			Expect(returnedResources).To(matchers.ContainCustomResource(
+				matchers.HaveTypeMeta(v1.SettingsGVK),
+				matchers.HaveObjectMeta(types.NamespacedName{
+					Namespace: defaults.GlooSystem,
+					// This matches the name of the Settings resource that we construct the History object with
+					Name: "my-settings",
+				}),
+				gstruct.Ignore(),
```
jenshu, line 105: do we want to add spec matchers in this PR?

test/gomega/matchers/custom_resource.go
jenshu, line 18: i don't know if we should bake the managed fields check in here. there may be other cases (non-admin server) where managedFields might not necessarily need to be empty

changelog/v1.18.0-beta10/remove-history-managed-fields.yaml
jenshu: i'm guessing the client objects in the tests don't contain managed fields in the first place, which is why the tests are passing

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -547,12 +732,7 @@ var _ = Describe("History", func() {
...
-				inputSnapshotBytes, err := history.GetInputSnapshot(ctx)
-				Expect(err).NotTo(HaveOccurred())
-
-				returnedResources := []crdv1.Resource{}
-				err = json.Unmarshal(inputSnapshotBytes, &returnedResources)
-				Expect(err).NotTo(HaveOccurred())
+				returnedResources := getTypedInputSnapshot(ctx, history)
 
 				// should contain the kube resources
 				expectContainsResource(returnedResources, gatewayv1.RouteOptionGVK, "h", "kube-rto")
```
jenshu, line 738: do we want to change over all of these to the new format as well?

jenshu (commented):
looks good so far - much cleaner!

projects/gloo/pkg/servers/iosnapshot/history_test.go
sam-heilbron: Good call, I updated all tests in this file

test/gomega/matchers/custom_resource.go
sam-heilbron: Good point! I updated the assertion to expect the name/namespace, and then also accept arbitrary matchers

projects/gloo/pkg/servers/iosnapshot/history_test.go
sam-heilbron: I decided against it. The reason I added it, is that when we add Secrets and Artifacts we will want to explicitly check the Spec has the appropriate data redacted. However, for the other resources, we don't really need to assert the body is the same

changelog/v1.18.0-beta10/remove-history-managed-fields.yaml
sam-heilbron: Updated to remove managedFields, and introduce them on the test objects

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -135,14 +174,22 @@ func (h *historyImpl) GetInputSnapshot(ctx context.Context) ([]byte, error) {
 	// get the resources from the edge api snapshot
 	resources, err := snapshotToKubeResources(snap)
 	if err != nil {
-		return nil, err
+		return SnapshotResponseData{
+			Status: "error",
```
npolshakova, line 178: [nit] Do we want to define these as types (error, complete, etc.) instead of hard coding the status string?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -238,30 +308,45 @@ func (h *historyImpl) getKubeGatewayResources(ctx context.Context) ([]crdv1.Reso
...
-		// convert each Unstructured to a Resource so that the final list can be merged with the
-		// edge resource list
-		for _, uns := range list.Items {
-			out := crdv1.Resource{}
-			err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.Object, &out)
-			if err != nil {
-				return nil, err
-			}
-			resources = append(resources, out)
+			// todo: handle graceful error
```
npolshakova, line 315: [nit] Do we have a follow up issue to link here, or maybe expand the comment a little so we don't loose context?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -238,30 +308,45 @@ func (h *historyImpl) getKubeGatewayResources(ctx context.Context) ([]crdv1.Reso
...
+	if err != nil {
+		return nil, err
+	}
+	// convert each Unstructured to a Resource so that the final list can be merged with the
+	// Edge API resource list
+	for _, uns := range list.Items {
+		out := crdv1.Resource{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.Object, &out)
+		if err != nil {
+			return nil, err
```
npolshakova, line 340: So we don't want to do a multi error- we return immediately if any of the resources fail to get converted?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -238,30 +308,45 @@ func (h *historyImpl) getKubeGatewayResources(ctx context.Context) ([]crdv1.Reso
...
+		}
+
+		// ManagedFields is noise on the object, that is not relevant to the Admin API, so we sanitize it
+		out.ManagedFields = nil
```
npolshakova, line 344: [nit] Do we want to do something similar to `redactSecretData` and have a helper `sanitizeResource` that sets `out.ManagedFields = nil`? (Not sure if we'll ever re-use it anywhere else though)

sam-heilbron: yes i really like this idea

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Good point. As much as we wanted to improve our error handling, i realized that it wasn't easy to know when we should continue vs exist early (and return the error). This feels like a good instance where conversion should accumulate the error, but also continue. i will update

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: ah great catch! I had intended to solve this in code. Ill update, and if it can't be / isn't solved in this PR, ill be sure to link an issue

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: yess 100%. tbh I had considered this and then was lazy and didn't do it. Thanks for pushing for the better solution

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Updated, let me know if this matches what you had in mind

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Updated, let me know if this matches what you expected

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: I updated this error handling, let me know what you think

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Updated, let me know what you think

jenshu (commented):
still need to go through the changes, but fyi it doesn't seem to be outputting valid json

`curl localhost:9091/snapshots/edge | jq` ==> `jq: parse error: Invalid numeric literal at line 1, column 8`

docs/content/operations/debugging_gloo/_index.md
```diff
@@ -362,7 +362,30 @@ observability:
...
+The Control Plane exposes a set of [Administration endpoints](https://github.com/solo-io/gloo/tree/3ad2c1b4e7f0d73f6caeacee461f04315e612d47/projects/gloo/pkg/servers/admin). To access these:
+
+* Enable port forwarding:
+```
+kubectl port-forward -n gloo-system deploy/gloo 9091:9091
+```
+
+The following endpoints are then available:
+* `http://localhost:9091/snapshots/input`: Returns a list of resources, ordered by GVK, that the Control Plane is aware of.
+* `http://localhost:9091/snapshots/edge`: Returns a map of Edge API resources, key by type, that the Control Plane is aware of.
```
jenshu, line 375:
```suggestion
* `http://localhost:9091/snapshots/edge`: Returns a map of Edge API resources, keyed by type, that the Control Plane is aware of.
```

docs/content/operations/debugging_gloo/_index.md
```diff
@@ -362,7 +362,30 @@ observability:
...
+### Gloo Admin Endpoints [*Recommended, Introduced in 1.17*]
+The Control Plane exposes a set of [Administration endpoints](https://github.com/solo-io/gloo/tree/3ad2c1b4e7f0d73f6caeacee461f04315e612d47/projects/gloo/pkg/servers/admin). To access these:
+
+* Enable port forwarding:
+```
+kubectl port-forward -n gloo-system deploy/gloo 9091:9091
+```
+
+The following endpoints are then available:
+* `http://localhost:9091/snapshots/input`: Returns a list of resources, ordered by GVK, that the Control Plane is aware of.
```
jenshu, line 374: i don't know if this would be too noisy, but do we want to list the resource types that are included? or that certain k8s resources (e.g. services) are not included, even though they may be watched by the control plane?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -278,9 +329,37 @@ func (h *historyImpl) getKubeGatewayClientSafe() client.Client {
...
 	// why we redact Artifacts
-	snap.Artifacts = nil
+	snap.Artifacts.Each(func(element *gloov1.Artifact) {
+		redactArtifactData(element)
+	})
+}
+
+// redactSecretData modifies the secret to remove any sensitive information
+// The structure of a Secret in Gloo Gateway does not lend itself to easily redact data in different places.
+// As a result, we perform a primitive redaction method, where we maintain the metadata, and remove the entire spec
```
jenshu, line 345: do you know if the gloo secrets would contain any annotations such as [this](https://github.com/solo-io/skv2/blob/1583cb716c04eb3f8d01ecb179b0deeabaa6e42b/contrib/pkg/snapshot/redact.go#L20-L26) that would need to be removed?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -278,9 +329,37 @@ func (h *historyImpl) getKubeGatewayClientSafe() client.Client {
...
+func redactSecretData(element *gloov1.Secret) {
+	element.Kind = nil
```
jenshu, line 347: doesn't need to be done in the first pass, but i wonder if it would be useful to retain some info about the secret, e.g. what type it is

projects/gloo/pkg/servers/iosnapshot/resources.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	// resource types to return in the kube gateway input snapshot
+	// KubeGatewayDefaultGVKs is the list of resource types to return in the kube gateway input snapshot
 	KubeGatewayDefaultGVKs = []schema.GroupVersionKind{
```
jenshu, line 14: i just realized (unrelated to your changes) that the kube gateway controller also watches [secrets](https://github.com/solo-io/gloo/blob/9f5fe80a8ec1f65257351a5ad0bb49570b0f6c6e/projects/gateway2/secrets/controller.go#L24) and [upstreams](https://github.com/solo-io/gloo/blob/9f5fe80a8ec1f65257351a5ad0bb49570b0f6c6e/projects/gateway2/controller/controller.go#L80). should we add those to the GVKs here (under `resources shared between Edge and Kubernetes Gateway integration` below) and update [this section](https://github.com/solo-io/gloo/blob/9f5fe80a8ec1f65257351a5ad0bb49570b0f6c6e/projects/gloo/pkg/servers/iosnapshot/history.go#L121-L133) to remove those types from the edge snapshot when kubegateway is enabled?

docs/content/operations/debugging_gloo/_index.md
```diff
@@ -362,7 +362,30 @@ observability:
...
+* `http://localhost:9091/snapshots/edge`: Returns a map of Edge API resources, key by type, that the Control Plane is aware of.
+* `http://localhost:9091/snapshots/proxies`: Returns a list of Proxy CRs.
+* `http://localhost:9091/snapshots/xds`: Returns a map of xDS snapshots, keyed by the cache key for each snapshot.
+
+All endpoints return data in the following shape:
+```json
+{
+   "status": "metadata about the response",
+   "data": "the returned data",
+   "error": "error value, if one was encountered"
```
jenshu, line 384: will this be a single error or a list of aggregated errors?

install/helm/gloo/generate/values.go
```diff
@@ -223,8 +223,9 @@ type Settings struct {
...
 	CircuitBreakers               CircuitBreakersSettings `json:"circuitBreakers,omitempty" desc:"Set this to configure the circuit breaker settings for Gloo."`
 	EnableRestEds                 *bool                   `json:"enableRestEds,omitempty" desc:"Whether or not to use rest xds for all EDS by default. Defaults to false."`
-	DevMode                       *bool                   `json:"devMode,omitempty" desc:"Whether or not to enable dev mode. Defaults to false. Setting to true at install time will expose the gloo dev admin endpoint on port 10010. Not recommended for production."`
-	SecretOptions                 SecretOptions           `json:"secretOptions,omitempty" desc:"Options for how Gloo Edge should handle secrets."`
+	// NOTE: DevMode is deprecated. See https://docs.solo.io/gloo-edge/latest/operations/debugging_gloo/#debugging-the-control-plane for more details.
+	DevMode       *bool         `json:"devMode,omitempty" desc:"Whether or not to enable dev mode. Defaults to false. Setting to true at install time will expose the gloo dev admin endpoint on port 10010. Not recommended for production."`
```
jenshu, line 227: can we also put in the description that it's deprecated? (so it shows up in the helm value docs)

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -82,8 +87,7 @@ type historyImpl struct {
 	xdsCache          cache.SnapshotCache
 	settings          *gloov1.Settings
 	kubeGatewayClient client.Client
-	// this will hold all the kube gvks that we want to show in the input snapshot when kube gateway
-	// integration is enabled
+	// kubeGvks is the list of GVKs that the historyImpl will return when GetInputSnapshot is invoked
```
jenshu, line 90: could add note that it's only the kubernetes gateway (not edge) related gvks

projects/gloo/pkg/servers/iosnapshot/format.go
```diff
@@ -42,3 +44,21 @@ func sortResources(resources []crdv1.Resource) {
...
+	}
+
+	jsn, err := json.Marshal(snap)
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(jsn, &genericMap); err != nil {
+		return nil, err
+	}
+	return genericMap, nil
```
jenshu, line 63: does the marshal/unmarshal produce a consistent ordering of resources each time?
i wonder if we should keep them sorted, e.g. first by key (kind) and then within each kind sort by namespace/name, so it's easier for a user to diff snapshots against each other

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -135,14 +155,14 @@ func (h *historyImpl) GetInputSnapshot(ctx context.Context) ([]byte, error) {
 	// get the resources from the edge api snapshot
 	resources, err := snapshotToKubeResources(snap)
 	if err != nil {
-		return nil, err
+		return errorSnapshotResponse(err)
```
jenshu, line 158: if we're supporting partial results, imo we should never return early, and instead just accumulate the errors to return at the end

docs/content/operations/debugging_gloo/_index.md
sam-heilbron: I think we can leave that out for now. It's not a requirement from product excellence team, and I think oftentimes this API will be used to capture state (for us to consume). So if there are resources we need to be included, we should just include them in the API respnose, and not necessarily in the metatdata about hte repsonse. However, if you/field think it would be a valuable addition to our API, I think it could be done in a follow-up, wdyt?

projects/gloo/pkg/servers/iosnapshot/response.go
```diff
@@ -0,0 +1,50 @@
...
+		Data:   string(bytes),
+		Error:  nil,
+	}
+}
+
+func errorSnapshotResponse(err error) SnapshotResponseData {
+	return SnapshotResponseData{
+		Status: Error,
+		Data:   "",
+		Error:  err,
```
jenshu, line 48: could this be a list of errors, for readability? (otherwise i think it just lists all the multierr errors in one string?)

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: We aren't yet supporting partial results. It's too complex with all of our logic to identify what is partial and what is not. So the approach I took is that we accumulate errors and return them if they exist.

projects/gloo/pkg/servers/iosnapshot/response.go
sam-heilbron: Could we keep it as an error, and if we're concerned about how multiple errors are displayed, then we can create a way to format that multierror into a single one?
Also an error should rarely happen and will require developer intervention, so do you suspect that as it is rendered right now, we will not be able to parse it to act on it?

docs/content/operations/debugging_gloo/_index.md
sam-heilbron: Idea: Perhaps update our docs to clarify _what_ is being returned
Decision: Since this is mostly used internally, we're going to leave the solution as is (without documenting the resources in a user-facing doc). If we find consumers of this API (oftentimes Field or customerse) are missing information,we  will handle that in a follow up

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Good call, will update

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Good idea. i'm not sure what this would look like exactly, but I am planning on creating a "Admin Server Follow Up" issue that includes other potential improvements like filtering, and I can make a mention of this

projects/gloo/pkg/servers/iosnapshot/resources.go
sam-heilbron: sounds good, will update

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: will renamed to `kubeGatewayGVKs` and clarify the point in the comment as well

projects/gloo/pkg/servers/iosnapshot/format.go
sam-heilbron: good call. ill update to ensure it does work that way and is tested

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -547,98 +1002,105 @@ var _ = Describe("History", func() {
...
-				expectContainsResource(returnedResources, gatewayv1.RouteOptionGVK, "h", "kube-rto")
-				expectContainsResource(returnedResources, gatewayv1.VirtualHostOptionGVK, "i", "kube-vho")
-				expectContainsResource(returnedResources, extauthv1.AuthConfigGVK, "j", "kube-ac")
-				expectContainsResource(returnedResources, ratelimitv1alpha1.RateLimitConfigGVK, "k", "kube-rlc")
+				Expect(returnedResources).To(matchers.ContainCustomResource(
+					matchers.HaveTypeMeta(gatewayv1.RouteOptionGVK),
+					matchers.HaveObjectMeta(types.NamespacedName{
+						Name:      "kube-rto",
+						Namespace: "h",
+					}, matchers.HaveNilManagedFields()),
```
sam-heilbron, line 1013: todo: remove the HaveNilManagedFields assertion as it's unnecessary for this test, and could lead to incorrect behavior

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -547,98 +1002,105 @@ var _ = Describe("History", func() {
...
-				Expect(err).NotTo(HaveOccurred())
-
-				returnedResources := []crdv1.Resource{}
-				err = json.Unmarshal(inputSnapshotBytes, &returnedResources)
-				Expect(err).NotTo(HaveOccurred())
-
-				// we should now see the deployment in the input snapshot results
-				expectContainsResource(returnedResources, deploymentGvk, "a", "kube-deploy")
-			})
+			snap := getEdgeApiSnapshot(ctx, history)
```
sam-heilbron, line 1093: todo: assert that the returned data includes artifacts and secrets AND that their data is redacted properly

.github/workflows/static-analysis.yaml
```diff
@@ -51,5 +51,6 @@ jobs:
       with:
         version: v1.57.2
         args: --verbose --modules-download-mode=readonly --allow-parallel-runners
-        skip-pkg-cache: true
+        skip-cache: true
```
sam-heilbron, line 54: See logs from previous commits: https://github.com/solo-io/gloo/actions/runs/10115140341/job/27975260033?pr=9797
```
Warning: Unexpected input(s) 'skip-pkg-cache', valid inputs are ['version', 'install-mode', 'working-directory', 'github-token', 'only-new-issues', 'skip-cache', 'skip-save-cache', 'problem-matchers', 'args', 'cache-invalidation-interval']
```

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -110,8 +117,24 @@ func (h *historyImpl) SetKubeGatewayClient(kubeGatewayClient client.Client) {
...
 
+func (h *historyImpl) GetEdgeApiSnapshot(_ context.Context) SnapshotResponseData {
+	snap := h.getRedactedApiSnapshot()
+
+	m, err := apiSnapshotToGenericMap(snap)
+	if err != nil {
+		errorSnapshotResponse(err)
+	}
+
+	data, err := formatOutput("json_compact", m)
```
npolshakova, line 128: [nit] We have `json_compact` defined in a couple places- can we make it a const?

npolshakova: I think it's also used here: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/servers/admin/server.go#L59

npolshakova (approved):

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -278,9 +334,51 @@ func (h *historyImpl) getKubeGatewayClientSafe() client.Client {
...
+
+	redactGlooResourceMetadata(element.GetMetadata())
+}
+
+// redactGlooResourceMetadata modifies the metadata to remove any sensitive information
+// ref: https://github.com/solo-io/skv2/blob/1583cb716c04eb3f8d01ecb179b0deeabaa6e42b/contrib/pkg/snapshot/redact.go#L20-L26
+func redactGlooResourceMetadata(meta *core.Metadata) {
+	for key, _ := range meta.GetAnnotations() {
+		if key == corev1.LastAppliedConfigAnnotation {
+			meta.GetAnnotations()[key] = redactedString
```
jenshu, line 375: nit: if we found the key, can break here

projects/gloo/pkg/servers/iosnapshot/response.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (r SnapshotResponseData) MarshalJSON() ([]byte, error) {
+	// See: https://github.com/golang/go/issues/5161#issuecomment-1750037535
+	var errorMsg string
+	if r.Error != nil {
+		errorMsg = r.Error.Error()
+	}
+	anon := struct {
+		Data  string `json:"data"`
```
jenshu, line 20: i think there's still something wrong with the marshaling

`curl localhost:9091/snapshots/input | jq` is giving me the entire `data` field as a one-line string:

```
{
  "data": "[{\"kind\":....l}]",
  "error": ""
}
```

should `data` be a map instead?

sam-heilbron: No, data has always been a list of resources, each with GVK

projects/gateway2/wellknown/gwapi.go
```diff
@@ -73,4 +74,9 @@ var (
 		Version: "v1beta1",
 		Kind:    ReferenceGrantListKind,
 	}
+	SecretGVK = schema.GroupVersionKind{
```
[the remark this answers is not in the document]
sam-heilbron: I wasn't sure of a better place, and am open to ideas. I put it here because there are other kubernetes types defined in here, and I wanted to make it distinct from the Gloo secret type.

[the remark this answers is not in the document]
sam-heilbron: I put this into the kube section of our api. Let me know what you think

projects/gloo/pkg/servers/iosnapshot/response.go
jenshu: sorry yeah meant list - just not a string

sam-heilbron: good catch, i updateed the code and our testing

projects/gloo/pkg/api/v1/kube/wellknown/types.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	SecretGVK = schema.GroupVersionKind{
+		Group:   corev1.GroupName,
+		Version: "v1",
```
npolshakova, line 11: K8s doesn't define SecretGVK anywhere? I think it does have the version though:
```
schema.GroupVersionKind{
		Kind:    "Secret",
		Group:   corev1.SchemeGroupVersion.Group,
		Version: corev1.SchemeGroupVersion.Version,
}
```

Or you could use `SchemeGroupVersion.WithKind("Secret")`?

npolshakova (dismissed):

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -238,28 +260,47 @@ func (h *historyImpl) getKubeGatewayResources(ctx context.Context) ([]crdv1.Reso
...
-			err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.Object, &out)
-			if err != nil {
-				return nil, err
-			}
-			resources = append(resources, out)
+			// We intentionally aggregate the errors so that we can return a "best effort" set of
+			// resources, and one error doesn't lead to the entire set of GVKs being short-circuited
+			errs = multierror.Append(errs, err)
 		}
+		resources = append(resources, gvkResources...)
```
jenshu, line 272: we should probably only append if there is no error

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -238,28 +260,47 @@ func (h *historyImpl) getKubeGatewayResources(ctx context.Context) ([]crdv1.Reso
...
+	// Edge API resource list
+	for _, uns := range list.Items {
+		out := crdv1.Resource{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.Object, &out)
+		if err != nil {
+			errs = multierror.Append(errs, err)
+		}
+
+		sanitizeResource(&out)
+		resources = append(resources, out)
```
jenshu, line 301: should only do this if there's no error

projects/gloo/pkg/servers/iosnapshot/history_test.go
```diff
@@ -348,131 +662,281 @@ var _ = Describe("History", func() {
...
+				returnedResources := getInputSnapshotResources(ctx, history)
+				Expect(returnedResources).To(matchers.ContainCustomResource(
+					matchers.MatchTypeMeta(wellknown.GatewayClassGVK),
+					matchers.MatchObjectMeta(types.NamespacedName{
+						Name:      "kube-gw-class",
+						Namespace: "c",
+					}, matchers.HaveNilManagedFields()),
+					gstruct.Ignore(),
+				), fmt.Sprintf("results should contain %v %s.%s", wellknown.GatewayClassGVK, "c", "kube-gw-class"))
+			})
```
jenshu, line 733: nit: looks like a lot of these similar checks can be refactored into a reusable func

jenshu (approved):

npolshakova (approved):

jenshu:
IstioIntegration flake? https://github.com/solo-io/gloo/issues/9836

PR #9808, opened 2024-07-23
gateway2/delegation: fix extraneous route arising from invalid child rule

shashankram:
# Description

There's a bug where if a child route contains an invalid rule (rule not matching the parent matcher), then even though the matcher is discarded, the rule with an empty matcher but valid backendRef is returned by GetDelegatedRoutes(). The result is that a `/` route is programmed for such an invalid route rule. A more precise fix is to also prune the rules that do not have a valid matcher so that we do not rely on the translator to interpret a route without a valid matcher as '/', which could be an alternative fix though fragile.

The essence of this fix is to prune both the `rules` and `matches` field on the child route when we process it in the context of the parent matcher, so that:
1. invalid matchers on the child route are discarded
2. invalid rules (no valid child matchers) are also discarded

Previously, 2. was missing so a child route with a rule without a matcher was configured, which results in a `/` route being exposed for the corresponding backendRef.

## Testing steps

Run the same translator test without the fix and verify that a route `/` gets programmed for the backendRef on the route.

# Checklist:

### DIFF
--- a/projects/gateway2/query/delegation.go
+++ b/projects/gateway2/query/delegation.go
@@ -79,10 +79,15 @@ func (r *gatewayQueries) GetDelegatedRoutes(
 	var selected []gwv1.HTTPRoute
 	for _, child := range children {
 		inheritMatcher := shouldInheritMatcher(child)
-		matched := false
 
 		// Check if the child route has a prefix that matches the parent.
 		// Only rules matching the parent prefix are considered.
+		//
+		// We use validRules to store the rules in the child route that are valid
+		// (matches in the rule match the parent route matcher). If a specific rule
+		// in the child is not valid, then we discard it in the final child route
+		// returned by this function.
+		var validRules []gwv1.HTTPRouteRule
 		for i, rule := range child.Spec.Rules {
 			var validMatches []gwv1.HTTPRouteMatch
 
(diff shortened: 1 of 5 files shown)


projects/gateway2/query/delegation.go
```diff
@@ -100,22 +104,26 @@ func (r *gatewayQueries) GetDelegatedRoutes(
...
-					matched = true
 				}
 			}
 
-			// Update the rule on the child instead of on a copy of the rule
-			child.Spec.Rules[i].Matches = validMatches
+			// Matchers in this rule match the parent route matcher, so consider the valid matchers on the child,
+			// and discard rules on the child that do not match the parent route matcher.
+			if len(validMatches) > 0 {
+				validRule := child.Spec.Rules[i]
```
stevenctl, line 120: nit: why use the indexer instead of the loop var `rule`?

stevenctl (approved):

stevenctl (approved):
lgtm, just the one nit

stevenctl (dismissed):

projects/gateway2/query/delegation.go
shashankram: Just for consistency with L84 where we assign validMatches. Makes it more obvious than using the loop var for that reason.

stevenctl (approved):

jenshu (approved):

PR #9809, opened 2024-07-23
gateway2/delegation: fix extraneous route arising from invalid child rule

shashankram:
# Description

There's a bug where if a child route contains an invalid rule (rule not matching the parent matcher), then even though the matcher is discarded, the rule with an empty matcher but valid backendRef is returned by GetDelegatedRoutes(). The result is that a `/` route is programmed for such an invalid route rule. A more precise fix is to also prune the rules that do not have a valid matcher so that we do not rely on the translator to interpret a route without a valid matcher as '/', which could be an alternative fix though fragile.

The essence of this fix is to prune both the `rules` and `matches` field on the child route when we process it in the context of the parent matcher, so that:
1. invalid matchers on the child route are discarded
2. invalid rules (no valid child matchers) are also discarded

Previously, 2. was missing so a child route with a rule without a matcher was configured, which results in a `/` route being exposed for the corresponding backendRef.

## Testing steps

Run the same translator test without the fix and verify that a route `/` gets programmed for the backendRef on the route.

# Checklist:

### DIFF
--- a/projects/gateway2/translator/gateway_translator_test.go
+++ b/projects/gateway2/translator/gateway_translator_test.go
@@ -249,4 +249,5 @@ var _ = DescribeTable("Route Delegation translator",
 	Entry("RouteOptions merge child override on no conflict", "route_options_inheritance_child_override_ok.yaml"),
 	Entry("RouteOptions multi level inheritance with child override", "route_options_multi_level_inheritance_override_ok.yaml"),
 	Entry("RouteOptions filter override merge", "route_options_filter_override_merge.yaml"),
+	Entry("https://github.com/solo-io/solo-projects/issues/6621", "bug-6621.yaml"),
 )
--- a/projects/gateway2/translator/httproute/delegation_helpers.go
+++ b/projects/gateway2/translator/httproute/delegation_helpers.go
@@ -48,10 +48,15 @@ func filterDelegatedChildren(
 		child := child.Clone()
 
 		inheritMatcher := shouldInheritMatcher(&child.HTTPRoute)
-		matched := false
 
 		// Check if the child route has a prefix that matches the parent.
 		// Only rules matching the parent prefix are considered.
+		//
+		// We use validRules to store the rules in the child route that are valid
+		// (matches in the rule match the parent route matcher). If a specific rule
+		// in the child is not valid, then we discard it in the final child route
+		// returned by this function.
+		var validRules []gwv1.HTTPRouteRule
 		for i, rule := range child.Spec.Rules {
 			var validMatches []gwv1.HTTPRouteMatch
 
@@ -59,7 +64,6 @@ func filterDelegatedChildren(
 			// simply inherit the parent's matcher.
 			if inheritMatcher && len(rule.Matches) == 0 {
 				validMatches = append(validMatches, parentMatch)
-				matched = true
 			}
 
 			for _, match := range rule.Matches {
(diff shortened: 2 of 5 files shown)


projects/gateway2/translator/httproute/delegation_helpers.go
```diff
@@ -69,19 +73,24 @@ func filterDelegatedChildren(
...
-					matched = true
 				}
 			}
 
-			// Update the rule on the child instead of on a copy of the rule
+			// Matchers in this rule match the parent route matcher, so consider the valid matchers on the child,
 			child.Spec.Rules[i].Matches = validMatches
+			// and discard rules on the child that do not match the parent route matcher.
+			if len(validMatches) > 0 {
+				validRule := child.Spec.Rules[i]
```
stevenctl, line 87: nit: why use the indexer instead of the loop var rule?

stevenctl (approved):
lgtm, just the one nit

projects/gateway2/translator/httproute/delegation_helpers.go
shashankram: Just for consistency with L84 where we assign validMatches. Makes it more obvious than using the loop var for that reason.

EItanya (approved):

projects/gateway2/translator/gateway_translator_test.go
```diff
@@ -249,4 +249,5 @@ var _ = DescribeTable("Route Delegation translator",
 	Entry("RouteOptions merge child override on no conflict", "route_options_inheritance_child_override_ok.yaml"),
 	Entry("RouteOptions multi level inheritance with child override", "route_options_multi_level_inheritance_override_ok.yaml"),
 	Entry("RouteOptions filter override merge", "route_options_filter_override_merge.yaml"),
+	Entry("https://github.com/solo-io/solo-projects/issues/6621", "bug-6621.yaml"),
```
jenshu, line 252: nit: can we give the test a more descriptive name (e.g. invalid child matcher)

projects/gateway2/translator/httproute/delegation_helpers.go
```diff
@@ -48,18 +48,22 @@ func filterDelegatedChildren(
 		child := child.Clone()
```
jenshu, line 48: i can't add a comment on the line, but i think the description above should be reversed:

> The child route's path must be a prefix of the parent's path

should instead say "The parent route's path must be a prefix of the child's path"

projects/gateway2/translator/testutils/inputs/delegation/bug-6621.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: gateway.networking.k8s.io/v1
```
jenshu, line 1: nit: instead of gh issue number can we give a more descriptive name to this file?

projects/gateway2/translator/testutils/inputs/delegation/bug-6621.yaml
```diff
@@ -0,0 +1,64 @@
...
+kind: HTTPRoute
+metadata:
+  name: route-a
+  namespace: a
+spec:
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /a/1
```
jenshu, line 44: should we also add a test where the child has no routes matching the parent (i.e. they all get filtered out)?

projects/gateway2/translator/httproute/delegation_helpers.go
shashankram: unrelated to this PR, we can fix the comment separately

projects/gateway2/translator/testutils/inputs/delegation/bug-6621.yaml
shashankram: I thought using a GH issue makes sense as the issue contains all the context pertaining to the bug. We can change it if people have a strong opinion on it. It doesn't bother me regardless.

projects/gateway2/translator/testutils/inputs/delegation/bug-6621.yaml
shashankram: We test that in e2e (test/kubernetes/e2e/features/route_delegation/testdata/header_query_match.yaml), but no harm in including it here even though it's a no-op for the changes in this PR. The loop logic is the same regardless of 1 or more rules.

projects/gateway2/translator/gateway_translator_test.go
shashankram: https://github.com/solo-io/gloo/pull/9812

projects/gateway2/translator/httproute/delegation_helpers.go
shashankram: https://github.com/solo-io/gloo/pull/9812

projects/gateway2/translator/testutils/inputs/delegation/bug-6621.yaml
shashankram: https://github.com/solo-io/gloo/pull/9812

PR #9810, opened 2024-07-23
Move validation tests to new framework

npolshakova:
# Description

When the validation webhook is enabled, and `allowWarnings=true`, there are situations where configuration can be applied in the wrong order and be rejected by the webhook. This can cause flakes in e2e tests due to eventual consistency issues. 

Part of this work will update helm tests manifests in gloo oss and enterprise, to configure `allowWarnings=true`. The `allowWarnings=false` functionality will be moved to the new testing framework which should be more stable because
1) Avoid updating the settings CRD as part of the test and instead rely on the helm manifest
2) Avoid patching resources (default VirtualService, etc.)
3) Avoid using discovery for Upstream resources

## API changes
- The `disable_transformation_validation` field should not have been marked as deprecated in favor of `server_enabled`. These fields are responsible for different things, see slack thread: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1721419080718539

## Code changes
- None

## CI changes
- Added new tests to workflow

## Docs changes
We may want to update the docs to clarify that invalid upstreams are only rejected when there is a valid proxy. 

For example, if you apply thing config:
```
kubectl apply --dry-run=server -f- <<EOF
apiVersion: gloo.solo.io/v1
kind: Upstream
metadata:
  name: invalid-upstream
  namespace: gloo-system
spec:
  static:
    hosts:
      - addr: ~
EOF
```

It will be accepted with `allowWarnings=false` unless there is valid setup to start sending traffic to (valid VS, valid Upstream).

See slack thread: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1721658207422249?thread_ts=1704740564.269759&cid=G01EERAK3KJ

# Context
https://github.com/solo-io/solo-projects/issues/6321

## Interesting decisions
None

## Testing steps
- Added new e2e tests for validation

## Notes for reviewers
None

# Checklist:


test/kubernetes/e2e/features/validation/testdata/invalid-resources/vs-typo.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: gateway.solo.io/v1
+kind: VirtualService
+metadata:
+  name: default
+spec:
+  virtualHoost: {}
```
sam-heilbron, line 6: nit: I can quickly look at the resources and remember why they are considered invalid, but perhaps an inline comment above the misconfigured section would help others easily see what is wrong?

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
         # May 14th: ~ minutes execution time (see load_balancing_tests.md)
         - cluster-name: 'cluster-two'
           go-test-args: '-v -timeout=25m'
-          go-test-run-regex: '^TestK8sGatewayIstioRevision$$|^TestRevisionIstioRegression$$|^TestK8sGateway$$/^Deployer$$|^TestK8sGateway$$/^RouteOptions$$|^TestK8sGateway$$/^VirtualHostOptions$$|^TestK8sGateway$$/^Upstreams$$|^TestK8sGateway$$/^HeadlessSvc$$|^TestK8sGateway$$/^PortRouting$$|^TestK8sGatewayMinimalDefaultGatewayParameters$$'
+          go-test-run-regex: '^TestValidationStrict$$|^TestValidationAlwaysAccept$$|^TestTransformationValidationDisabled$$|^TestK8sGatewayIstioRevision$$|^TestRevisionIstioRegression$$|^TestK8sGateway$$/^Deployer$$|^TestK8sGateway$$/^RouteOptions$$|^TestK8sGateway$$/^VirtualHostOptions$$|^TestK8sGateway$$/^Upstreams$$|^TestK8sGateway$$/^HeadlessSvc$$|^TestK8sGateway$$/^PortRouting$$|^TestK8sGatewayMinimalDefaultGatewayParameters$$'
```
sam-heilbron, line 62: nit: I imagine that with the introduction of these validation tests, we may want to re-time the execution time for cluster2?

sam-heilbron (approved):
Love it! I think the separation of the tests for these different validation features will really help. 

Is it possible to run these tests with N=25, or some high number in CI, to prove that we are flake-free before merging?

test/kubernetes/e2e/features/validation/testdata/invalid-resources/vs-typo.yaml
npolshakova: Good idea! Having the yaml files are nice because you can attempt to apply them and see the error immediately, but I think having a comment on the line let's you know where the issue is.

npolshakova: https://github.com/solo-io/gloo/pull/9810/commits/e3a3e291716f0f1ecf6e9b82d46c8275a3fdee00

npolshakova:
> Love it! I think the separation of the tests for these different validation features will really help.
> 
> Is it possible to run these tests with N=25, or some high number in CI, to prove that we are flake-free before merging?

They're taking a while to run (~20), so I'm running with n=10, but here's the first result:
```
❯ TEST_PKG=./test/kubernetes/e2e/tests/  GO_TEST_ARGS="-test.run=TestValidationStrict" make go-test 
rm -rf /Users/ninapolshakova/gloo/_test/bug_report
mkdir -p /Users/ninapolshakova/gloo/_test/bug_report
GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore go test -ldflags="-X github.com/solo-io/gloo/pkg/version.Version=1.0.1-dev" \
-test.run=TestValidationStrict  -timeout=0 -count=10 -failfast \
./test/kubernetes/e2e/tests/
ok      github.com/solo-io/gloo/test/kubernetes/e2e/tests       1142.307s

```

sam-heilbron (dismissed):

test/kubernetes/e2e/features/validation/README.md
```diff
@@ -0,0 +1,32 @@
+# Validation Feature
+
+## Overview
+The validation feature is a suite of tests that validate the behavior of the validation webhook with invalid configurations. 
+The tests are run against a Gloo Edge installation with the validation webhook enabled with different settings.
```
jbohanon, line 5: nit: 
```suggestion
The tests are run against a Gloo Gateway installation with the validation webhook enabled with different settings.
```

test/kubernetes/e2e/features/validation/README.md
```diff
@@ -0,0 +1,32 @@
...
+| `disableTransformationValidation` | bool | By default gloo will attempt to validate transformations by calling out to a local envoy binary in `validate` mode. Calling this local envoy binary can become slow when done many times during a single validation. Setting this to true will stop gloo from calling out to envoy to validate the transformations, which may speed up the validation time considerably, but may also cause the transformation config to fail after being sent to envoy. When disabling this, ensure that your transformations are valid prior to applying them. |
+| `serverEnabled`                   | bool | By providing the validation field (parent of this object) the user is implicitly opting into validation. This field allows the user to opt out of the validation server, while still configuring pre-existing fields such as `warn_route_short_circuiting` and `disable_transformation_validation`. If not included, the validation server will be enabled.                                                                                                                                                                                      |
```
jbohanon, line 16: Do we have tests for the behavior when `disableTransformationValidation=false` and `serverEnabled=true`? Since we are un-deprecating the former, it seems like we should have that behavior well defined

npolshakova: Yep! We have a new `TestTransformationValidationDisabled` test that tests `disableTransformationValidation=true`, but by default `disableTransformationValidation=false`. The strict and allowWarnings tests run the `TestRejectTransformation` test which expects the webhook to reject the invalid VS.

npolshakova: I don't fully understand what `serverEnabled` is supposed to accomplish (you seem to need `serverEnabled` and `validation.enabled=false` in order to get `serverEnabled` to work): https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1721658431900589?thread_ts=1721419080.718539&cid=G01EERAK3KJ

test/kubernetes/e2e/features/validation/testdata/switch-vs/switch-valid-invalid.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: gateway.solo.io/v1
+kind: VirtualService
+metadata:
+  name: i-am-valid
```
jbohanon, line 4: are the names of the two vs resources in this file mismatched?

test/kubernetes/e2e/features/validation/README.md
jbohanon: I misunderstood the interplay between these two options

test/kubernetes/e2e/features/validation/testdata/switch-vs/switch-valid-invalid.yaml
npolshakova: No- the test switches the valid and invalid VS. In the original gateway test this is done with a patch: https://github.com/solo-io/gloo/blob/main/test/kube2e/gateway/gateway_test.go#L2651 Here we just reapply the file but keep the metadata the same.

npolshakova: I think the purpose of the original test is to show that Gloo preserves the valid virtual services in envoy when a virtual service has been made invalid (so the 200 ok response is still returned even though we flip the "valid" VS to an invalid config).

npolshakova: The reason this "switch" work is `alwaysAccept=true`: https://github.com/solo-io/gloo/blob/main/test/kube2e/gateway/gateway_test.go#L2552

sam-heilbron (approved):

jbohanon (approved):

PR #9843, opened 2024-08-01
Improve sync status time

npolshakova:
# Description

This fixes one of the issues from https://github.com/solo-io/solo-projects/issues/6638 so that there are no extraneous status updates. 

This also adds some debug logs to show reconcile time for the proxy syncer. 

Before:
```

{"level":"debug","ts":"2024-08-01T16:45:37.113Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:99","msg":"starting syncer for k8s gateway proxies"}
{"level":"debug","ts":"2024-08-01T16:45:37.220Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:111","msg":"resyncing k8s gateway proxies [1]"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:207","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:221","msg":"synced statuses in 83ns"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:179","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:200","msg":"synced route statuses in 6.125µs"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:159","msg":"translated and wrote 0 proxies in 113.667µs"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:111","msg":"resyncing k8s gateway proxies [2]"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:207","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:221","msg":"synced statuses in 83ns"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:179","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:200","msg":"synced route statuses in 2µs"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:159","msg":"translated and wrote 0 proxies in 46.625µs"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:111","msg":"resyncing k8s gateway proxies [3]"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:207","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:221","msg":"synced statuses in 167ns"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:179","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:45:37.221Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:200","msg":"synced route statuses in 4.667µs"}
{"level":"debug","ts":"2024-08-01T16:45:37.222Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:159","msg":"translated and wrote 0 proxies in 75.292µs"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:111","msg":"resyncing k8s gateway proxies [4]"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:207","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:221","msg":"synced statuses in 458ns"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:179","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:200","msg":"synced route statuses in 7.041µs"}
{"level":"debug","ts":"2024-08-01T16:45:51.626Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:159","msg":"translated and wrote 0 proxies in 117.916µs"}
```

After change:
```
Fix:

{"level":"debug","ts":"2024-08-01T16:34:47.158Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:100","msg":"starting syncer for k8s gateway proxies"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:112","msg":"resyncing k8s gateway proxies [1]"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:208","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:222","msg":"synced statuses in 125ns"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:180","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:201","msg":"synced route statuses in 5.125µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:160","msg":"translated and wrote 0 proxies in 88.708µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:112","msg":"resyncing k8s gateway proxies [2]"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:208","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:222","msg":"synced statuses in 84ns"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:180","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:201","msg":"synced route statuses in 1.625µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:160","msg":"translated and wrote 0 proxies in 37.959µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:112","msg":"resyncing k8s gateway proxies [3]"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:208","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:222","msg":"synced statuses in 125ns"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:180","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:201","msg":"synced route statuses in 2.5µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:160","msg":"translated and wrote 0 proxies in 31.417µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.262Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:112","msg":"resyncing k8s gateway proxies [4]"}
{"level":"debug","ts":"2024-08-01T16:34:47.263Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:208","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:34:47.263Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:222","msg":"synced statuses in 83ns"}
{"level":"debug","ts":"2024-08-01T16:34:47.263Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:180","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:34:47.263Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:201","msg":"synced route statuses in 1.459µs"}
{"level":"debug","ts":"2024-08-01T16:34:47.263Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:160","msg":"translated and wrote 0 proxies in 35.25µs"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:112","msg":"resyncing k8s gateway proxies [5]"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:208","msg":"syncing k8s gateway proxy status"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer.statusSyncer","caller":"proxy_syncer/proxy_syncer.go:222","msg":"synced statuses in 125ns"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:180","msg":"syncing k8s gateway route status"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer.routeStatusSyncer","caller":"proxy_syncer/proxy_syncer.go:201","msg":"synced route statuses in 3.625µs"}
{"level":"debug","ts":"2024-08-01T16:35:01.695Z","logger":"k8s-gw-syncer","caller":"proxy_syncer/proxy_syncer.go:160","msg":"translated and wrote 0 proxies in 48.958µs"}
```

## API changes
None

## Code changes
- Add `Bar()` function
- ...

## CI changes
- Adjusted schedule for x job
- ...

## Docs changes
- ...

# Context

## Interesting decisions
 
## Testing steps

## Notes for reviewers

# Checklist:

### DIFF
--- a/projects/gateway2/proxy_syncer/proxy_syncer.go
+++ b/projects/gateway2/proxy_syncer/proxy_syncer.go
@@ -176,6 +176,9 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
 func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap) {
 	ctx = contextutils.WithLogger(ctx, "routeStatusSyncer")
 	logger := contextutils.LoggerFrom(ctx)
+	logger.Debugf("syncing k8s gateway route status")
+	startTime := time.Now()
+
 	rl := apiv1.HTTPRouteList{}
 	err := s.mgr.GetClient().List(ctx, &rl)
 	if err != nil {
@@ -192,12 +195,16 @@ func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap)
 			}
 		}
 	}
+	contextutils.LoggerFrom(ctx).Debugf("synced route statuses in %v", time.Since(startTime))
 }
 
 // syncStatus updates the status of the Gateway CRs
 func (s *ProxySyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl apiv1.GatewayList) {
 	ctx = contextutils.WithLogger(ctx, "statusSyncer")
 	logger := contextutils.LoggerFrom(ctx)
+	logger.Debugf("syncing k8s gateway proxy status")
+	startTime := time.Now()
+
 	for _, gw := range gwl.Items {
 		gw := gw // pike
 		if status := rm.BuildGWStatus(ctx, gw); status != nil {
@@ -207,6 +214,7 @@ func (s *ProxySyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl 
 			}
 		}
 	}
+	contextutils.LoggerFrom(ctx).Debugf("synced statuses in %v", time.Since(startTime))
 }
 
 // reconcileProxies persists the proxies that were generated during translations and stores them in an in-memory cache
--- a/changelog/v1.17.2/fix-only-update-status-on-change.yaml
+++ b/changelog/v1.17.2/fix-only-update-status-on-change.yaml
@@ -1,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6638
+    resolvesIssue: true
+    description: |
+      Only update the K8s Gateway resource statuses on change to improve HTTPRoute translation time.


projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -207,6 +214,7 @@ func (s *ProxySyncer) syncStatus(ctx context.Context, rm reports.ReportMap, gwl
 			}
 		}
 	}
+	contextutils.LoggerFrom(ctx).Debugf("synced statuses in %v", time.Since(startTime))
```
sam-heilbron, line 217: Would we ever want this to be a metric instead? So you could easily observe overtime how expensive status operations are on your system (as you scale the number of resources)

npolshakova:
> Should we start by introducing this to 1.18 and then backporting?
> 
> I only see timing notes, which I think are helpful for observability, but no change that actually modifies the behavior of status updates...am I missing something?

Yep, the fix is in https://github.com/solo-io/gloo/pull/9843/commits/96702d1f33649e44da9721c5824552ea75fd9300! The original commit didn't have it since I was testing both versions locally.

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -186,27 +190,36 @@ func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap)
 	for _, route := range rl.Items {
 		route := route // pike
 		if status := rm.BuildRouteStatus(ctx, route, s.controllerName); status != nil {
-			route.Status = *status
-			if err := s.mgr.GetClient().Status().Update(ctx, &route); err != nil {
-				logger.Error(err)
+			if !reflect.DeepEqual(route.Status, *status) {
```
lgadban, line 193: I'm not sure we want to use `reflect` for this; is [`slices.Equal`](https://pkg.go.dev/slices#Equal) an option?

I don't remember off the top of my head, are the status objects in consistent order?

lgadban (commented):
While I think this is a solid and necessary fix, I think the more fundamental thing issue we should fix is that the status writes for HTTPRoutes and Gateways are synchronous.

Are we going to make those async in a separate PR?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -186,27 +190,36 @@ func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap)
...
+				route.Status = *status
+				if err := s.mgr.GetClient().Status().Update(ctx, &route); err != nil {
```
sam-heilbron, line 195: It also feels like number of status updates, broken up by resource type, could be an interesting metric. Not a requirement for this work, just calling out what I as an operator may like to have insight to.

sam-heilbron (commented):
Is there a way to add tests for this functionality? It seems simple enough that I'm less concerned about the functionality working today, but more about it drifting overtime, especially since we've identified it as a critical change for the control plane

projects/gateway2/proxy_syncer/proxy_syncer.go
sam-heilbron: Are we comparing a list of statuses here though? I thought this was just comparing a single object (the route status). Regardless, I think it's a good point that @lgadban raises here, which is it may be valuable to abstract the operation that we perform so that we can more easily change the implemtnation and test it. What if we had an equality method for RouteStatuses and that way we can easily prove that the equality works the way we'd expect, and could even measure the performance impact of using refleection

sam-heilbron:
> While I think this is a solid and necessary fix, I think the more fundamental thing issue we should fix is that the status writes for HTTPRoutes and Gateways are synchronous.
> 
> Are we going to make those async in a separate PR?

Yeah this is a good callout. I believe that @npolshakova had looked into this, and decided it was ultimately more complex to solve (and could potentially be done as a follow-up). Though I agree it would be helpful if we're explicit about whether that is going to be done in a follow-up, or document why we're not doing it, so that furture devs dont' do down that rabbit hole as well

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: RouteStatus is a slice of parentRefs (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/shared_types.go#L502).

lgadban:
>Yeah this is a good callout. I believe that @npolshakova had looked into this, and decided it was ultimately more complex to solve (and could potentially be done as a follow-up). Though I agree it would be helpful if we're explicit about whether that is going to be done in a follow-up, or document why we're not doing it, so that furture devs dont' do down that rabbit hole as well

It may be more complex but I don't see how we can not do it.

If we don't, any time status genuinely needs to be updated, we are holding up the main translation loop waiting for network IO to the k8s api-server.

projects/gateway2/proxy_syncer/proxy_syncer.go
EItanya: `slices.Equal` only works with `comparable`. The underlying type has a bunch of pointers: https://github.com/kubernetes-sigs/gateway-api/blob/f6ac5182637eb84efb5a628b663625c171e3edac/apis/v1/shared_types.go#L35

EItanya:
> > Yeah this is a good callout. I believe that @npolshakova had looked into this, and decided it was ultimately more complex to solve (and could potentially be done as a follow-up). Though I agree it would be helpful if we're explicit about whether that is going to be done in a follow-up, or document why we're not doing it, so that furture devs dont' do down that rabbit hole as well
> 
> It may be more complex but I don't see how we can not do it.
> 
> If we don't, any time status genuinely needs to be updated, we are holding up the main translation loop waiting for network IO to the k8s api-server.

In general we will probably have to make statuses `Async`. But given the amount of changes we will be pursuing related to performance in `1.18` it behooves us to do as little as possible on this branch. Given the way statuses are currently written they will have to change when the proxy syncer is no longer async, which will change the `impl`.

projects/gateway2/proxy_syncer/proxy_syncer.go
EItanya: > Are we comparing a list of statuses here though? I thought this was just comparing a single object (the route status). Regardless, I think it's a good point that @lgadban raises here, which is it may be valuable to abstract the operation that we perform so that we can more easily change the implemtnation and test it. What if we had an equality method for RouteStatuses and that way we can easily prove that the equality works the way we'd expect, and could even measure the performance impact of using refleection

It's not our type so we can't give it any functions unfortunately

projects/gateway2/proxy_syncer/proxy_syncer.go
EItanya: 1000% this should be a metric, these are the types of logs that make our logs flooded

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -186,27 +190,36 @@ func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap)
...
+					logger.Error(err)
+				}
 			}
 		}
 	}
+	contextutils.LoggerFrom(ctx).Debugf("synced route statuses in %v", time.Since(startTime))
```
EItanya, line 201: Turn into a metric, potentially a native histogram @krisztianfekete

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: Given the underlying pointers, github.com/google/go-cmp is another option, which is already a dependency.

I guess without benchmarks we can't be sure, so If we are all okay with it reflection seems fine for a first-pass.

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: The addition of a metric is a great idea but should not mean we need remove this debug log line.

Also, we probably need to wait until we bring in the prom library to actually add the metric

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: This is a debug log and a single-liner at that; I am personally a fan of leaving this in

projects/gateway2/proxy_syncer/proxy_syncer.go
EItanya: I believe `github.com/google/go-cmp` is meant for testing, but could definitely be wrong on that

projects/gateway2/proxy_syncer/proxy_syncer.go
krisztianfekete: Yes, that's correct, the library is not yet available.

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: I kept the original proxy-syncer debug line because it contains the reconcile count, but replaced the other log lines with metrics in: https://github.com/solo-io/gloo/pull/9843/commits/f938646b2d8b392f443fb18f42bf85d5ce3f62c8

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: Discussed offline, decided to go with the old-style of metric: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1722894644951389

npolshakova:
> Is there a way to add tests for this functionality? It seems simple enough that I'm less concerned about the functionality working today, but more about it drifting overtime, especially since we've identified it as a critical change for the control plane

We should add benchmark tests as part of a separate issue IMO, but probably after the new metrics are added.

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: Yep, this is separated into metrics type for statuses: https://github.com/solo-io/gloo/pull/9843/commits/f938646b2d8b392f443fb18f42bf85d5ce3f62c8 As part of https://github.com/solo-io/solo-projects/issues/6658 we should add metrics similar to what solo-kit does per k8s gateway resource: 
```
# TYPE api_gloo_solo_io_translator_resources gauge
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway",resource="clusters"} 17
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway",resource="endpoints"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway",resource="listeners"} 2
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway",resource="routes"} 1
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-1",resource="clusters"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-1",resource="endpoints"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-1",resource="listeners"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-1",resource="routes"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-2",resource="clusters"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-2",resource="endpoints"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-2",resource="listeners"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-example-gateway-2",resource="routes"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-gw",resource="clusters"} 14
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-gw",resource="endpoints"} 13
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-gw",resource="listeners"} 1
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~default~default-gw",resource="routes"} 1
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway",resource="clusters"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway",resource="endpoints"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway",resource="listeners"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway",resource="routes"} 0
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway-test",resource="clusters"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway-test",resource="endpoints"} 16
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway-test",resource="listeners"} 1
api_gloo_solo_io_translator_resources{proxy_name="gloo-kube-gateway-api~infra~infra-http-gateway-test",resource="routes"} 1
```

pkg/utils/statsutils/stopwatch.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+	"time"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+)
+
+type StopWatch interface {
```
sam-heilbron, line 11: nit: For an exported interface, that other developers may use, can we add some comments around it?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -247,3 +267,24 @@ func applyPostTranslationPlugins(ctx context.Context, pluginRegistry registry.Pl
...
 }
+
+var opts = cmp.Options{
+	cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
+	cmpopts.IgnoreMapEntries(func(k string, _ any) bool {
+		return k == "lastTransitionTime"
+	}),
+}
+
+func isGatewayStatusEqual(objA, objB *gwv1.GatewayStatus) bool {
```
sam-heilbron, line 278: Should we add some basic unit tests to demonstrate this works as intended?

projects/gateway2/translator/gateway_translator.go
```diff
@@ -47,6 +48,8 @@ func (t *translator) TranslateProxy(
 	writeNamespace string,
 	reporter reports.Reporter,
 ) *v1.Proxy {
+	stopwatch := statsutils.NewTranslatorStopWatch("TranslateProxy")
+	stopwatch.Start()
```
sam-heilbron, line 52: We call start, and then in this function may return early. What is the impact on the stopWatch that we started in that case? To be safe, should we just defer stopping the watch right after calling start?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -176,7 +183,11 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
 func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap) {
 	ctx = contextutils.WithLogger(ctx, "routeStatusSyncer")
 	logger := contextutils.LoggerFrom(ctx)
-	rl := apiv1.HTTPRouteList{}
+	logger.Debugf("syncing k8s gateway route status")
+	stopwatch := statsutils.NewTranslatorStopWatch("HTTPRouteStatusSyncer")
```
sam-heilbron, line 187: Should this be a NewStatusSyncStopWatch?

pkg/utils/statsutils/translation_metrics.go
```diff
@@ -0,0 +1,47 @@
...
+			Description: "how long the status syncer takes in seconds",
+			Aggregation: view.Distribution(0.01, 0.05, 0.1, 0.25, 0.5, 1, 5, 10, 60),
+			TagKeys:     []tag.Key{statusSyncerNameKey},
+		},
+	); err != nil {
+		log.Fatalf("Failed to register views: %v", err)
+	}
+}
+
+func NewTranslatorStopWatch(translatorName string) StopWatch {
```
sam-heilbron, line 41: What do you see as the value of having separate timers for translation_time and status_syncer? (I'm not opposed, just trying to understand). It feels like all of these are just translation measurements, each with their own granularity

pkg/utils/statsutils/stopwatch.go
```diff
@@ -0,0 +1,37 @@
...
+		measure: measure,
+		labels:  labels,
+	}
+}
+
+func (s *stopwatch) Start() {
+	s.startTime = time.Now()
+}
+
+func (s *stopwatch) Stop(ctx context.Context) {
```
sam-heilbron, line 33: What is the impact of calling Stop twice? Is that somethign that we should prevent, or just document for developers?

pkg/utils/statsutils/stopwatch.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"time"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+)
+
+type StopWatch interface {
+	Start()
+	Stop(ctx context.Context)
```
sam-heilbron, line 13: Is there value in exposing a method to return the captured duration?

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: Yep, only ProxySyncer should be NewTranslatorStopWatch

pkg/utils/statsutils/translation_metrics.go
npolshakova: I was originally imagining statuses and translation would have different units, but with the opencensus metrics it now takes in a configurable `measure` option. I'm fine combining both of them into just `translation_time`?

pkg/utils/statsutils/stopwatch.go
npolshakova: Sure! I don't think it could hurt!

npolshakova: This can also clean up the debug log line in resyncProxies!

projects/gateway2/translator/gateway_translator.go
npolshakova: Yeah, I can switch it to `defer stopwatch.Stop(ctx)`!

pkg/utils/statsutils/stopwatch.go
npolshakova: I think it would just result in the duration being recorded twice. But the second time would be pretty much the same since the recorded duration is calculated based on the time elapsed since s.startTime. 

We could add a mutex and `stopped` bool to keep track of if the stopwatch was stopped or not?

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: Switched to one time as per: https://github.com/solo-io/gloo/pull/9843#discussion_r1706167640

pkg/utils/statsutils/stopwatch.go
sam-heilbron: We spoke offline and agreed that documenting the existing behavior would be enough to provide guardrails for future devs.

sam-heilbron (dismissed):

pkg/utils/statsutils/translation_metrics.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func init() {
+	// Register views with OpenCensus
+	if err := view.Register(
+		&view.View{
+			Name:        "gloo_edge/translation_time_sec",
+			Measure:     translationTime,
+			Description: "how long the translator takes in seconds",
+			Aggregation: view.Distribution(0.01, 0.05, 0.1, 0.25, 0.5, 1, 5, 10, 60),
```
krisztianfekete, line 23: What do you think about having more granular buckets at the high end of the layout? People tend to care less if all is fast, but having better visibility into the other end might be beneficial to understand how slow something is, and how that is improving.
(Note, this is something that native histograms can nicely solve, once we migrate to `prometheus/client_golang`)

pkg/utils/statsutils/translation_metrics.go
```diff
@@ -0,0 +1,33 @@
...
+var (
+	translationTime      = stats.Float64("gloo_edge/translation_time_sec", "how long the translator takes in seconds", "s")
+	translatorNameKey, _ = tag.NewKey("translator_name")
+)
+
+func init() {
+	// Register views with OpenCensus
+	if err := view.Register(
+		&view.View{
+			Name:        "gloo_edge/translation_time_sec",
```
krisztianfekete, line 20: I've started an initial discussion on the new naming convention in this ticket: https://github.com/solo-io/solo-projects/issues/6659

If we are fine with that, the `go_client`-version of this metric would be `gloo_gateway_translation_time`. Can we maybe chose another namespace/prefix for this one as `gloo_edge` suggests that this is specific to Edge, while it's already agnostic of the API it describes? For consistency and to enable smooth migration, we can try to match the existing naming of metrics instrumented via OpenCensus.

pkg/utils/statsutils/translation_metrics.go
npolshakova: What granularity do you think we need here? These are based on our other solo-kit generated metrics granularity: https://github.com/solo-io/solo-kit/blob/1baf6de465942dc5be44e7f28f0f739dcd0b967b/pkg/code-generator/codegen/templates/event_loop_template.go#L34

pkg/utils/statsutils/translation_metrics.go
npolshakova: This is technically a new metric that could be used in edge as well, but I agree we can rename it to `gloo_gateway`. When the metric is initialized, it's also given a name which is used in the tag key:
```
	stopwatch := statsutils.NewTranslatorStopWatch("GatewayStatusSyncer")
	stopwatch.Start()
	defer stopwatch.Stop(ctx)
```

npolshakova: Discussed offline- will keep as `gloo_edge` until we move to new prometheus metrics

pkg/utils/statsutils/translation_metrics.go
npolshakova: Discussed offline- these ranges are currently fine for k8s gateway metrics. Gloo edge envoy translator metric will be more granular

sam-heilbron (approved):

nfuden (approved):

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -247,3 +269,24 @@ func applyPostTranslationPlugins(ctx context.Context, pluginRegistry registry.Pl
...
+
+var opts = cmp.Options{
+	cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),
+	cmpopts.IgnoreMapEntries(func(k string, _ any) bool {
+		return k == "lastTransitionTime"
+	}),
+}
+
+func isGatewayStatusEqual(objA, objB *gwv1.GatewayStatus) bool {
+	if cmp.Equal(objA, objB, opts) {
```
lgadban, line 281: does this compare consider ordering of the underlying slice?

IIRC (e.g. https://github.com/solo-io/gloo/blob/main/projects/gateway2/reports/status.go#L29) the underlying status is not currently sorted

npolshakova: Ah good point, we might need to make the status conditions deterministic. Is `gw.Spec.Listeners` defined by the user though?

lgadban: Correct, same with the iteration done for parentRefs in the HTTPRoute.
We should be good here!

lgadban (approved):

projects/gateway2/proxy_syncer/proxy_syncer.go
npolshakova: Removed comment in https://github.com/solo-io/gloo/pull/9843/commits/903220d62e2403000af222eb2b66d4f48a754cb2

lgadban (approved):

sam-heilbron (approved):

PR #9844, opened 2024-08-01
admin/history: Return redacted Artifacts and Secrets in InputSnapshot

sam-heilbron:
# Description

Update the Gloo Admin API to return data in the expected shape, with the proper fields redacted

# Context

## UX
We have had a few iterations around this API, and this PR is intended to complete this cycle. There will of course be future iterations on this work, but for now the expectations of the API are documented in the `admin/server.go`.

## Main (Previous) Limitation
We were taking the resources from our ApiSnapshot, and converting them into a type `crdv1.Resource`. For all CRDs this worked well enough, but it meant that Secrets and ConfigMaps (Artifacts, in Gloo Edge terminology) were unintentionally wiped entirely during the unmarshal phase. **One of the requirements of the input snapshot API is that Secrets include the keys of the data field, but the values are redacted.** This meant that we had to change how we were gathering objects, and converting them into Kubernetes resources.

### Client Construction
In the past, we relied on a client being constructed by the Kube Gateway integration, and then we would set it on the History object at runtime. The impact of this is that if the Kube Gateway integration was not enabled, we would fallback to not having a kube client, and only requesting resources from memory.

The requirements of this API are that we return all kubernetes objects, so we need to have a client constructed whether we're using Edge Gateway APIs or Kube Gateway APIs.

### GVK List
I broke up the GVK list of resources into smaller chunks so it was easier to see which GVKs are supported, and then have unit tests to demonstrate it.

### Breaking Changes
Since this does change some exported types, I opened an EE PR to pull this in, and have the changes ready: https://github.com/solo-io/solo-projects/pull/6708

## Interesting decisions
-

### Run CI Tests,  Locally
**There are unit tests for the history package, those can be run with:**
```
 TEST_PKG=projects/gloo/pkg/servers/iosnapshot/ make test
```
Expected result:
```
Will run 31 of 31 specs
•••••••••••••••••••••••••••••••

Ran 31 of 31 Specs in 0.173 seconds
SUCCESS! -- 31 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

**There are unit tests for the new Gloo Admin client, those can be run with:**
```
TEST_PKG=pkg/utils/glooadminutils/... make test
```
Expected result:
```
Will run 2 of 2 specs
••

Ran 2 of 2 Specs in 0.022 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 2.044289125s
Test Suite Passed
```

**There are e2e tests which verify that the admin server can be accessed when an installation is running. Those can be run with:**
```
go test -v -timeout 600s ./test/kubernetes/e2e/tests -run ^TestGlooGatewayEdgeGateway$/^GlooAdminServer
```
Expected result:
```
--- PASS: TestGlooGatewayEdgeGateway (73.55s)
    --- PASS: TestGlooGatewayEdgeGateway/GlooAdminServer (1.16s)
        --- PASS: TestGlooGatewayEdgeGateway/GlooAdminServer/TestGetInputSnapshot (1.15s)
PASS
ok      github.com/solo-io/gloo/test/kubernetes/e2e/tests       74.705s
```

### Access Admin Server of Running Installation
Create a KinD cluster
```
kind create cluster
```
Build the latest images
```
VERSION=1.0.0-sh ci/kind/setup-kind.sh
```

Create  the gloo-system ns:
```
kubectl create ns gloo-system
```

Install GG in the KinD cluster, using standard install values:
```
helm install -n gloo-system gloo _test/gloo-1.0.0-sh.tgz -f test/kubernetes/e2e/tests/manifests/k8s-gateway-test-helm.yaml
```

Open a port-forward to the admin server
```
kubectl port-forward -n gloo-system deployment/gloo 9091:9091
```

Curl the various snapshot endpoints:
```
```
```
{"data":[{"kind":"GatewayParameters","apiVersion":"gateway.gloo.solo.io/v1alpha1","metadata":{"name":"gloo-gateway","namespace":"gloo-system","uid":"4cbf950d-e767-4dc5-9d07-744cc1e74741","resourceVersion":"1449","generation":1,"creationTimestamp":"2024-07-24T20:21:54Z","labels":{"app.kubernetes.io/managed-by":"Helm","gloo":"kube-gateway"},"annotations":{"meta.helm.sh/release-name":"gloo","meta.helm.sh/release-namespace":"gloo-system"}},"status":null,"spec":{"kube":{"aiExtension":{"enabled":false,"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"gloo-ai-extension","tag":"1.0.0-sh"}},"deployment":{"replicas":1},"envoyContainer":{"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"gloo-envoy-wrapper","tag":"1.0.0-sh"},"securityContext":{"allowPrivilegeEscalation":false,"capabilities":{"add":["NET_BIND_SERVICE"],"drop":["ALL"]},"readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":10101}},"istio":{"istioProxyContainer":{"image":{"pullPolicy":"IfNotPresent","registry":"docker.io/istio","repository":"proxyv2","tag":"1.22.0"},"istioDiscoveryAddress":"istiod.istio-system.svc:15012","istioMetaClusterId":"Kubernetes","istioMetaMeshId":"cluster.local","logLevel":"warning"}},"podTemplate":{"extraLabels":{"gloo":"kube-gateway"}},"sdsContainer":{"bootstrap":{"logLevel":"info"},"image":{"pullPolicy":"IfNotPresent","registry":"quay.io/solo-io","repository":"sds","tag":"1.0.0-sh"}},"service":{"type":"LoadBalancer"},"stats":{"enableStatsRoute":true,"enabled":true,"routePrefixRewrite":"/stats/prometheus","statsRoutePrefixRewrite":"/stats"}}}}],"error":""}
```

Can use `jq` to parse the response
```
curl -X GET "127.0.0.1:9091/snapshots/input" | jq
```
## Notes for reviewers
-

# Checklist:


test/kubernetes/e2e/features/admin_server/suite.go
```diff
@@ -0,0 +1,71 @@
...
+
+// TestGetInputSnapshot verifies that we can query the /snapshots/input API and have it return data without an error
+func (s *testingSuite) TestGetInputSnapshot() {
+	s.testInstallation.Assertions.AssertGlooAdminApi(
+		s.ctx,
+		metav1.ObjectMeta{
+			Name:      kubeutils.GlooDeploymentName,
+			Namespace: s.testInstallation.Metadata.InstallNamespace,
+		},
+		inputSnapshotAssertion(s.testInstallation),
```
npolshakova, line 46: Will there be other snapshot assertions added here? I'm wondering if we should rename the test to `TestGetInputSnapshotWithSettings` if it's just scoped to settings. Or make `inputSnapshotAssertion` a shared assertion that can be used for other resources (make kind and metadata configurable?)

test/gomega/matchers/client_object.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/types"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	k8stypes "k8s.io/apimachinery/pkg/types"
+)
+
+func MatchClientObject(gvk schema.GroupVersionKind, namespacedName k8stypes.NamespacedName, customMatchers ...types.GomegaMatcher) types.GomegaMatcher {
```
npolshakova, line 12: Is this used anywhere?

test/kubernetes/e2e/tests/edge_gw_tests.go
```diff
@@ -15,6 +16,7 @@ func EdgeGwSuiteRunner() e2e.SuiteRunner {
 	edgeGwSuiteRunner.Register("PortRouting", port_routing.NewEdgeGatewayApiTestingSuite)
 	edgeGwSuiteRunner.Register("ValidationRejectInvalid", validation_reject_invalid.NewTestingSuite)
 	edgeGwSuiteRunner.Register("ValidationAllowWarnings", validation_allow_warnings.NewTestingSuite)
+	edgeGwSuiteRunner.Register("GlooAdminServer", admin_server.NewTestingSuite)
```
npolshakova, line 19: Do we have an issue to add GlooAdminServer tests to check for k8s-gateway-only and edge-only resources?

test/gomega/matchers/client_object.go
sam-heilbron: It's used in the history_test.go: https://github.com/solo-io/gloo/blob/sh/remove-history-managed-fields-part3/projects/gloo/pkg/servers/iosnapshot/history_test.go#L111

test/kubernetes/e2e/tests/edge_gw_tests.go
sam-heilbron: We don't, and it's a good idea. Would you rather I tackle that now (in a quick fast-follow PR)?

test/kubernetes/e2e/features/admin_server/suite.go
sam-heilbron: Yeah it's a good point. I wanted to add automated testing to make it easier for future devs making changes to this feature, to rely on/expand the testing. But I think narrowed the scope to just asserting settings. I'm happy to add a test case for a resource that should only work for k8s gateway api, and another for edge api.

test/kubernetes/e2e/tests/edge_gw_tests.go
sam-heilbron: I pusehd up some simple tests that I think cover this, let me know what you think

test/kubernetes/e2e/features/admin_server/suite.go
sam-heilbron: I updated these functions to be more aptly named for what they do

test/kubernetes/e2e/features/admin_server/suite.go
```diff
@@ -0,0 +1,127 @@
...
+			Name:      defaults.SettingsName,
+			Namespace: s.testInstallation.Metadata.InstallNamespace,
+		}),
+	)
+}
+
+// TestGetInputSnapshotIncludesEdgeApiResources verifies that we can query the /snapshots/input API and have it return Edge API
+// resources without an error
+func (s *testingSuite) TestGetInputSnapshotIncludesEdgeApiResources() {
+	s.T().Cleanup(func() {
```
npolshakova, line 59: For edge-only resources (VirtualServices, etc.) does this work the same as TestGetInputSnapshotIncludesK8sGatewayApiResources?

npolshakova (approved):

test/kubernetes/e2e/features/admin_server/suite.go
sam-heilbron: Yeah so edge-only resources will be returned always (so it's a little different than k8s resources). I used the Upstream resource as an example of this, but we could certaintly apply a VS and assert it exists as well. 

The reason edge-only resources are different than k8s resources, is due to the fact that CRDs must be registered for them to return, and on install all edge resources have the CRDs registered. 

I did run into an issue before where I tried to get the input snapshot when k8s gateway was disabled, and there were errors returned while trying to query the HttpRoute...etc. I have fixed the logic to not attempt those GVKs if it's not enabled, and the assertion that no error is included in the inputSnapshot response, I think demonstrates that we're in a good state

projects/gloo/pkg/servers/iosnapshot/redact.go
```diff
@@ -0,0 +1,64 @@
...
+
+// RedactSecretData returns a copy with sensitive information redacted
+func redactKubeSecretData(obj client.Object) {
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		return
+	}
+
+	for k := range secret.Data {
+		secret.Data[k] = []byte(redactedString)
```
jenshu, line 29: when testing this locally, i see that Secret data looks like:
```
      "data": {
        "ca.crt": "PHJlZGFjdGVkPg==",
        "tls.crt": "PHJlZGFjdGVkPg==",
        "tls.key": "PHJlZGFjdGVkPg=="
      },
```
i.e. it's the string `<redacted>` but  base64-encoded. do we want to just leave that string unencoded (if possible)?

test/kubernetes/e2e/features/admin_server/suite.go
```diff
@@ -0,0 +1,127 @@
...
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, gatewayParametersManifest)
+	s.Assert().NoError(err, "can apply gateway.gloo.solo.io GatewayParameters manifest")
+
+	s.testInstallation.Assertions.AssertGlooAdminApi(
+		s.ctx,
+		metav1.ObjectMeta{
+			Name:      kubeutils.GlooDeploymentName,
+			Namespace: s.testInstallation.Metadata.InstallNamespace,
+		},
+		inputSnapshotContainsElement(s.testInstallation, "GatewayParameters", gatewayParametersMeta),
```
jenshu, line 98: when testing locally (`curl localhost:9091/snapshots/input|jq`), i am not seeing GatewayParameters being returned in the results. how does this assertion work - is it checking any element in the results, or only at the top level? 

i do see that GatewayClass has this in its spec, and wondering if this matcher is incorrectly matching on that?
```
      "spec": {
        "controllerName": "solo.io/gloo-gateway",
        "parametersRef": {
          "group": "gateway.gloo.solo.io",
          "kind": "GatewayParameters",
          "name": "gloo-gateway",
          "namespace": "gloo-system"
        }
      },
```


edit: on the other hand, the gwp you are using here has a different name (`gw-params`) than the default, so maybe that's not the issue

test/kubernetes/e2e/features/admin_server/suite.go
```diff
@@ -0,0 +1,127 @@
...
+	)
+}
+
+func inputSnapshotContainsElement(testInstallation *e2e.TestInstallation, kind string, meta metav1.ObjectMeta) func(ctx context.Context, adminClient *admincli.Client) {
```
jenshu, line 102: can we also pass in `group` as an arg, so we can distinguish between duplicate kinds (e.g. Edge Gateway vs k8s Gateways)

projects/gloo/pkg/servers/iosnapshot/redact.go
sam-heilbron: I don't think that's possible. Plus this is how the GME repo handlees it, so I believe it's ok. The only requirement is that we sanitize the data, I doubt Denis cares about what it is

test/kubernetes/e2e/features/admin_server/suite.go
sam-heilbron: This assertion works by doing the following:
- Using the client to query the input snapshot api
- Return a list of objects
- assert that in that list, there is an object that has the kind/metadata fields that match

pkg/schemes/scheme.go
```diff
@@ -0,0 +1,60 @@
...
+	apiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+var defaultScheme *runtime.Scheme
+
+func init() {
+	s := scheme.Scheme
+	_ = AddToScheme(s)
+	defaultScheme = s
+}
```
jenshu, line 24: i don't think we'll be able to put this in `init`

we tried this before and it had to be fixed - see https://github.com/solo-io/gloo/pull/9692

in your s-p PR you can see the fed test is panicking on registering schemes - https://github.com/solo-io/solo-projects/actions/runs/10218837867/job/28275857410?pr=6708

jenshu (changes requested):

pkg/schemes/scheme.go
npolshakova: Ah good catch- I ran into this before as well.

test/kubernetes/e2e/features/admin_server/suite.go
sam-heilbron: updated to pass gvk

pkg/schemes/scheme.go
sam-heilbron: nice! i didn't know we had hit this. I updated the code that sets up the schemes and added a link to this PR

pkg/schemes/scheme.go
```diff
@@ -0,0 +1,56 @@
...
+	sologatewayv1.AddToScheme,
+
+	// Solo Edge Gloo API resources
+	gloov1.AddToScheme,
+
+	// Enterprise Extensions
+	// These are packed in the OSS Helm Chart, and therefore we register the schemes here as well
+	graphqlv1beta1.AddToScheme,
+	extauthkubev1.AddToScheme,
+	v1alpha1.AddToScheme,
```
jenshu, line 41: super-nit, just because all the others have more descriptive names: this could be `ratelimitv1alpha1` or `rlv1alpha1`

pkg/utils/glooadminutils/admincli/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Client
+This is the Go client that should be used whenever communicating with the Gloo Admin API. Within the Gloo project, it is used inside of tests.
+
+### Philosophy
+We expose methods that the client exposes which are [syntactic sugar](https://en.wikipedia.org/wiki/Syntactic_sugar) on top of the command API. These methods tend to follow the naming convention: `GetX`:
+```go
+func GetInputSnapshot(ctx context.Context) (string, error) {}
+```
+_As a general practice, these methods should return a concrete type, whenever possible._
```
jenshu, line 14: nit: we are returning `(interface{}, error)` now. should these couple lines be updated?

projects/gloo/pkg/servers/iosnapshot/resources.go
```diff
@@ -1,36 +1,70 @@
...
+
+	GlooGVKs = []schema.GroupVersionKind{
+		gloov1.SettingsGVK,
+		gloov1.UpstreamGVK,
+		gloov1.UpstreamGroupGVK,
+		gloov1.ProxyGVK,
+	}
+
+	// PolicyGVKs is the set of GVKs that are used by a classic Gloo Gateway installation,
+	// when only the Edge Gateway APIs are being used
```
jenshu, line 28: correction to comment: VirtualHostOption and RouteOption are used by both Edge and k8s gw

projects/gloo/pkg/servers/iosnapshot/resources.go
```diff
@@ -1,36 +1,70 @@
...
+	}
+
+	KubernetesGatewayGVKs = []schema.GroupVersionKind{
+		wellknown.GatewayClassGVK,
+		wellknown.GatewayGVK,
+		wellknown.HTTPRouteGVK,
+		wellknown.ReferenceGrantGVK,
+	}
+
+	KubernetesGatewayIntegrationPolicyGVKs = []schema.GroupVersionKind{
```
jenshu, line 49: no action needed, just noting that some resource types (VirtualHostOption, RouteOption, Upstream, Secret, ConfigMap) only appear under the "Edge" GVKs, but that is ok for now because the "CompleteInputSnapshotGVKs" includes the Edge ones.

In the future if we ever update the admin server to allow _only_ returning k8s gateway-related resources (and NOT edge resources), we'll need to update these groupings so that the "Kubernetes" gvks directly include these gvks

projects/gloo/pkg/servers/iosnapshot/resources.go
```diff
@@ -1,36 +1,70 @@
...
+
+		// While these are in fact Policy APIs, they are only enabled if the Kubernetes Gateway Integration is turned on
 		gatewayv1.ListenerOptionGVK,
 		gatewayv1.HttpListenerOptionGVK,
-
-		// resources shared between Edge and Kubernetes Gateway integration
-		gatewayv1.RouteOptionGVK,
-		gatewayv1.VirtualHostOptionGVK,
-		extauthv1.AuthConfigGVK,
-		ratelimitv1alpha1.RateLimitConfigGVK,
```
jenshu, line 32: are AuthConfigs and RateLimitConfigs no longer returned?
also, should we add GraphqlApis?

projects/gloo/pkg/servers/iosnapshot/redact.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	redactedString = "<redacted>"
+)
+
+// redactClientObject modifies the Object to remove any unwanted fields
+func redactClientObject(object client.Object) {
+	// ManagedFields is noise on the object, that is not relevant to the Admin API, so we sanitize it
+	object.SetManagedFields(nil)
+
+	redactKubeSecretData(object)
```
jenshu, line 18: nit: do we need to call this for every object (even non-Secrets)?

pkg/utils/glooadminutils/admincli/client_test.go
```diff
@@ -0,0 +1,78 @@
...
+			))
+		})
+
+	})
+
+	Context("Integration tests", func() {
+
+		When("Admin API is reachable", func() {
+			// We rely on e2e tests defined in /test/kubernetes/e2e/features/admin_server to verify this behavior
+		})
```
jenshu, line 46: do we want to leave this here?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -106,93 +127,57 @@ func (h *historyImpl) SetApiSnapshot(latestApiSnapshot *v1snap.ApiSnapshot) {
...
-		return errorSnapshotResponse(err)
-	}
-
-	// get settings, which is not part of the api snapshot
-	if h.settings != nil {
-		settings, err := settingsToKubeResource(h.settings)
+	var objects []client.Object
+	var errs *multierror.Error
+	for _, gvk := range h.inputSnapshotGvks {
+		gvkResources, err := h.listObjectsForGvk(ctx, inputSnapshotClient, gvk)
```
jenshu, line 145: are we no longer excluding Proxies from the input snapshot?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -250,66 +235,51 @@ func (h *historyImpl) getApiSnapshotSafe() *v1snap.ApiSnapshot {
...
-	return resources, errs.ErrorOrNil()
+
+	return objects, errs.ErrorOrNil()
 }
 
-// getKubeGatewayClientSafe gets the Kubernetes client used for CRUD operations
+// getInputSnapshotClientSafe gets the Kubernetes client used for CRUD operations
 // on resources used in the Kubernetes Gateway integration
-func (h *historyImpl) getKubeGatewayClientSafe() client.Client {
+func (h *historyImpl) getInputSnapshotClientSafe() client.Client {
```
jenshu, line 279: do we still need this func, if the inputSnapshotClient is only set once in the constructor?

pkg/schemes/scheme.go
```diff
@@ -0,0 +1,55 @@
...
+func AddToScheme(s *runtime.Scheme) error {
+	return SchemeBuilder.AddToScheme(s)
+}
+
+// DefaultScheme returns a scheme with all the types registered for Gloo Gateway
+// We intentionally do not perform this operation in an init!!
+// See https://github.com/solo-io/gloo/pull/9692 for context
+func DefaultScheme() *runtime.Scheme {
+	s := runtime.NewScheme()
+	_ = AddToScheme(s)
```
jenshu, line 53: we are dropping the error here; is that ok?

projects/gloo/pkg/servers/iosnapshot/resources.go
sam-heilbron: COrrect, those are engterprise resources, which shouldn't be included in an open source install

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: Correct, input snapshot is purely passtrhough to kubernets. so if it's persisted in kubernetes, we'll return it

projects/gloo/pkg/servers/iosnapshot/history.go
sam-heilbron: No, good call, i can update

pkg/schemes/scheme.go
sam-heilbron: Yes, if there is an error it will be visible in other areas (since the scheme isn't registered). All other instances of our code do this

pkg/utils/glooadminutils/admincli/client_test.go
sam-heilbron: I left this here, to mirror our envoy admincli which calls this out

projects/gloo/pkg/servers/iosnapshot/redact.go
sam-heilbron: We can either check the object type and call it, or call it for every object and inside that function we short-circuit if it's not a secret. I opted to mirror the GME prior art and just bake the check for type=secret into the function

projects/gloo/pkg/servers/iosnapshot/resources.go
sam-heilbron: Correct, the "only" was an attempt to say when "only the edge api is used" these are the policies you get.

jenshu (approved):

pkg/schemes/scheme.go
sam-heilbron: Happy to update! I'm going to handle in a follow-up PR

pkg/utils/glooadminutils/admincli/README.md
sam-heilbron: good call, ill update the readme in a fast-follow

Issue #6674, opened 2022-07-04
default docker-comspose-file installation fails


huzlak:
### Version

1.11.x (latest)

### Describe the requested changes

following our [docs](https://docs.solo.io/gloo-edge/latest/installation/gateway/development/docker-compose-file/) fails to start gloo with:

```
gloo_1           | {"level":"fatal","ts":"2022-07-04T07:03:55.307Z","logger":"gloo","caller":"setuputils/main_setup.go:73","msg":"error in setup: starting snapshot watch: initial Proxy list: reading file into *v1.Proxy: unknown field \"metadataStatic\" in gloo.solo.io.Route","version":"1.10.0","stacktrace":"github.com/solo-io/gloo/pkg/utils/setuputils.Main\n\t/workspace/gloo/pkg/utils/setuputils/main_setup.go:73\ngithub.com/solo-io/gloo/projects/gloo/pkg/setup.startSetupLoop\n\t/workspace/gloo/projects/gloo/pkg/setup/setup.go:22\ngithub.com/solo-io/gloo/projects/gloo/pkg/setup.Main\n\t/workspace/gloo/projects/gloo/pkg/setup/setup.go:14\nmain.main\n\t/workspace/gloo/projects/gloo/cmd/main.go:14\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:225"}
```
After defining newer than default `1.10.0` image it works. Docs should be corrected so the default installation steps work. also version should be aligned with at least current minor release number so 1.11.x should not point to 1.10 by default. The best would be to point to latest stable release and add a test to make sure docker compose works.

### Link to any relevant existing docs

https://docs.solo.io/gloo-edge/latest/installation/gateway/development/docker-compose-file/

nfuden:
Likely it should not point to newest release it should just point to the glooversion variable that exists in the makefile if at all possible.

That way if one clones an LTS branch for example then the docker-compose will still work as expected.

sam-heilbron:
We have updated this implementation: https://github.com/solo-io/gloo/tree/main/install/docker-compose-file. We now default to a more recent version, and also allow users to inject the custom Gloo Edge version that they want to run, using the [GLOO_VERSION env variable](https://github.com/solo-io/gloo/blob/main/install/docker-compose-file/docker-compose.yaml#L5). I believe this is now resolved so I am closing the issue. 

If you find other gaps, please feel free to re-open

PR #9867, opened 2024-08-06
Append cors settings (ExposeHeaders)

bewebi:
# Description

Add a field to VirtualHost options for configuring merge strategy for CORS settings which may be set at the route and/or VH level
In this PR we specifically allow ExposeHeaders to be appended (ie union) however this is exposed by an API that can easily be extended to include additional CORS settings and/or merge strategies

## API changes
- Add `cors_policy_merge_settings` field to `VirtualHostOptions` of `CorsPolicyMergeSettings` type
  - `CorsPolicyMergeSettings` is added to `cors.proto` and has one exposed field, `expose_headers` of `MergeStrategy` type
    - Additional fields for each CORS setting are commented out and can be added as needed in the future
    - `MergeStrategy` is an enum, with `DEFAULT` and `UNION` options, with additional options commented out which can be added as needed in the future

## Code changes
- Add a step in the CORS plugin's `ProcessRoute()` to compute the merged CORS settings using `mergeCors()`
  - `mergeCors()` takes a `*VirtualHostOptions_CorsMergeSettings` and a vh and a route `*CorsPolicy` and outputs a new `*CorsPolicy` that represents the result of applying the `CorsMergeSettings` to the given `CorsPolicy`s
- Add testing:
  - Unit test new plugin logic 
  -  Add `ExposeHeaders` to the existing CORS e2e test, setting one `ExposeHeaders` header at each of the Route and VH level and expect only the Route level to be exposed in the response
  -  Add a new test that adds `CorsPolicyMergeSettings` to the VH with `ExposeHeaders` set to `UNION`, set `ExposeHeaders` header at each of the Route and VH level and expect both to be exposed in the response
- Add `Dedupe()` generic util function for deduping slices
  - Unit test  

## Docs changes
- Fix description of `virtual_host` field on `VirtualService` proto
- Update Edge CORS guide to include example of applying CORS at Route level and applying `CorsPolicyMergeSettings`

# Context

See [design doc](https://soloio.slab.com/posts/merge-cors-settings-mtu6xxl6) and [Slack channel](https://solo-io-corp.slack.com/archives/C07DY9ZV38A), particularly discussion of customer requirements [here](https://solo-io-corp.slack.com/archives/C07DY9ZV38A/p1722002854529609)

## Interesting decisions
 
We opt for an extensible API in order to future-proof this area
While only one setting and merge strategy is currently necessary, adding additional fields or strategies will be an easy, non-breaking change in the future

~We opt to do the merge at VH computation time rather than at VS to VH translation time (ie [here](https://github.com/solo-io/gloo/pull/9867/commits/42dae66668f5bb7e83a4a510f3b70574b4d302ff#diff-1798680d501fc0422ed766714393275828a28a75b33ccbe86a7c5ce8ed614d76L299)) as we do not want to actually modify the Gloo CR representation of the VH, but rather ensure that the Envoy VH that is output has the desired CORS behavior applied~
We implement this in the CORS plugin, keeping all CORS logic centralized in that one location, rather than leak CORS implementation to the generic translation logic.

~As this is currently a one-off fix with a single setting for a single customer, I have not added a docs guide, but can do so here or as a followup if it would be useful.~

We did not decide on a mechanism for exposing the effective per-Route CORS policy to the user.
Initially the implementation propagated the merged CORS policy to the Routes in Envoy config, regardless of whether it was strictly necessary, in order to provide a single source of truth that users could find. This, along with documentation, was one of the proposed mechanisms for exposing the effective CORS policy outlined in the [design doc](https://soloio.slab.com/posts/merge-cors-settings-mtu6xxl6#h7ui8-how-to-expose-end-result-to-users).
Given that propagating all CORS policy to the Routes has the potential to dramatically grow the size of Envoy config under certain scenarios, without an agreed-upon benefit, we have instead implemented a solution that only makes necessary changes to Route-level CORS policy.
A mechanism for exposing the effective CORS policy can be decided on later if necessary and implemented in a future PR.

## Testing steps

I have added and updated e2e tests that demonstrate the intended behavior

~I have _not_ manually e2e tested in a k8s cluster
We are currently in the process of evaluating what degree of manual/e2e testing should be required by authors and reviewers on #6674
In this case the behavior is simple enough, and the in-memory e2e tests clear and strong enough, that manual kube e2e testing is not strictly necessary in my opinion, but I am open to reviewers requesting this~

The expected behavior has been manually validated on both the Edge API and Gateway API
Here is a demo video on the Edge API: 

https://github.com/user-attachments/assets/078b0d71-a1d0-4b18-b1ab-859926294a1b

## Notes for reviewers

Users should ensure they understand how the e2e tests work and run them

Feedback is also welcome on naming

# Checklist:


projects/gloo/pkg/translator/route_config.go
```diff
@@ -912,3 +920,33 @@ func ValidateRoutePath(s string) error {
...
 	return nil
 }
+
+func mergeCors(mergeSettings *v1.VirtualHostOptions_CorsMergeSettings, vh, route *cors.CorsPolicy) *cors.CorsPolicy {
+	// if either incoming CorsPolicy is nil, use the other
+	if vh == nil {
+		return route
+	}
+	if route == nil {
+		return vh
```
nfuden, line 930: if route is nil shouldnt we not specify a route level configuration at all and therefore return nil?

bewebi: For consistency's sake, I've taken the approach of always specifying the exact CORS settings we want on the Route itself when mergeSettings are non-nil, even if the result is propagating settings from the VH to the Route where the Route didn't previously have one

I try to get at this in [this comment](https://github.com/solo-io/gloo/pull/9867/files#diff-6490a2e321320a6f99f2947a8fdda39bdd809c3cdce94273b9c1d30831d51c40R143-R144)

In addition to consistency/simplicity, this also ensures that the Route is always a single source of truth in Envoy config when there are merge settings, since the CRs themselves are no longer a user-facing source of truth

projects/gloo/api/v1/options.proto
```diff
@@ -384,6 +384,35 @@ message VirtualHostOptions {
...
+        // mergeStrategy allow_headers = 4;
+
+        MergeStrategy expose_headers = 5;
+
+        // mergeStrategy max_age = 6;
+        // mergeStrategy allow_credentials = 7;
+        // mergeStrategy disable_for_route = 8;
+    }
+
+    CorsMergeSettings cors_merge_settings = 20;
```
sheidkamp, line 415: With this change should we update the docs [here](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/options.proto#L276) and [here](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/options.proto#L444) to reference these settings?

projects/gloo/pkg/translator/route_config.go
nfuden: Doesnt that mean that our envoy configuration could grow drastically as a result of this? Is that a trade off we are comfortable with?

projects/gloo/api/v1/options.proto
bewebi: Yes, great catch
IIRC, that documentation was fixed as part of the same effort that led to the creation of this issue

projects/gloo/pkg/translator/route_config.go
bewebi: It could, yes, and some amount of this is inevitable, ie in this case if there are ExposeHeaders on both VH and Route, we have no choice but to copy the VH-level ExposeHeaders to Routes
That said, we can take a minimalist approach, only copying to Route-level when strictly necessary
This wouldn't prevent Envoy config from growing drastically, but would narrow the cases where this would be a concern

I do think this is reasonable
We don't typically rely on Envoy config to be a user-facing source of truth, so perhaps it's not compelling to do so now at the cost of unnecessarily growing the size of the config

No alternative was decided on in terms of exposing the final effective CORS config for routes, so I went with the low-effort option that satisfied that consideration
We can either decide on one and increase the scope of the PR to encompass it, implement in a fast-follow, or, if it's not a key requirement, kick the can until it arises as a genuine need

projects/gloo/api/v1/options.proto
```diff
@@ -384,6 +385,35 @@ message VirtualHostOptions {
     // Enterprise-only: External Processing filter settings for the virtual host. This can be used to
     // override certain HttpListenerOptions settings, and can be overridden by RouteOptions settings.
     extproc.options.gloo.solo.io.RouteSettings ext_proc = 30;
+
+    // Settings for determining how CORS settings are merged when present on both VirtualHost and Route
+    message CorsMergeSettings {
```
sam-heilbron, line 390: This message seems tightly couple (for good reason) to the CorsPolicy defined here: https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/options/cors/cors.proto. 

Instead of needing to have lots of references across files, could we take the following approach:
- Define this message in the same cors.proto file
- Follow the existing terminology (the existing message is CorsPolicy, so I think this one we could call CorsMergePolicy)

projects/gateway/api/v1/virtual_service.proto
```diff
@@ -116,11 +116,8 @@ message VirtualService {
     option (core.solo.io.resource).short_name = "vs";
     option (core.solo.io.resource).plural_name = "virtual_services";
 
-    // The VirtualHost contains the
-    // The list of HTTP routes define routing actions to be taken
-    // for incoming HTTP requests whose host header matches
-    // this virtual host. If the request matches more than one route in the list, the first route matched will be selected.
-    // If the list of routes is empty, the virtual host will be ignored by Gloo.
+    // The VirtualHost contains configuration for serving a list of routes for a set of domains along with options for
+    // configuring traffic
```
sam-heilbron, line 120: Why are we modifying this comment?

projects/gloo/pkg/translator/route_config.go
```diff
@@ -130,14 +132,23 @@ func (h *httpRouteConfigurationTranslator) computeVirtualHost(
...
-
 	sanitizedName := utils.SanitizeForEnvoy(params.Ctx, virtualHost.GetName(), "virtual host")
 	if sanitizedName != virtualHost.GetName() {
 		virtualHost = virtualHost.Clone().(*v1.VirtualHost)
 		virtualHost.Name = sanitizedName
 	}
 	var envoyRoutes []*envoy_config_route_v3.Route
 	for i, route := range virtualHost.GetRoutes() {
+
+		// Envoy overrides vh-level CORS policy with route policy for any fields present on the route
```
sam-heilbron, line 143: Our plugin model of translation is an attempt to abstract the implementation details/complexities of various Envoy filters, from our core translation logic. Is there a reason why we leak cors translation into our top level route_config translator, instead of introducing this into https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/cors/plugin.go

nfuden (approved):

test/e2e/cors_test.go
```diff
@@ -259,6 +267,103 @@ var _ = Describe("CORS", func() {
...
+							ExposeHeaders:    []string{routeHeader},
+						}}).
+					Build()
+
+				testContext.ResourcesToCreate().VirtualServices = gatewayv1.VirtualServiceList{
+					vsWithCors,
+				}
+			})
+
+			It("should respect CORS", func() {
```
sam-heilbron, line 303: How is the corsMergeSettings expected to impact the behavior of the CORS filter overall? Right now we copy/paste tests that existed before, I and don't quite understand what added coverage/verification we get from having the same tests run. Can you help me understand what I'm missing?

test/e2e/cors_test.go
```diff
@@ -259,6 +267,103 @@ var _ = Describe("CORS", func() {
...
+				// shows that vhost field is respected iff route field is not set
+				allowedOriginRequestBuilder := testContext.GetHttpRequestBuilder().
+					WithOptionsMethod().
+					WithPath("cors").
+					WithHeader("Origin", routeAllowedOrigins[0]).
+					WithHeader("Access-Control-Request-Method", http.MethodDelete).
+					WithHeader("Access-Control-Request-Headers", "X-Requested-With")
+				Eventually(func(g Gomega) {
+					g.Expect(testutils.DefaultHttpClient.Do(allowedOriginRequestBuilder.Build())).Should(matchers.HaveOkResponseWithHeaders(map[string]interface{}{
+						requestACHMethods: MatchRegexp(strings.Join(allowedMethods, ",")), // show that methods are still coming through despite being on the vhost only
```
sam-heilbron, line 359: nit: Should we explore using https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/contain_headers.go instead of using regex?

projects/gloo/api/v1/options.proto
```diff
@@ -384,6 +385,35 @@ message VirtualHostOptions {
...
+
+        // Eventually we may want to allow each CORS setting to be configured, so we reserve fields for them
+
+        // mimicking the CorsPolicy message
+        // mergeStrategy allow_origin = 1;
+        // mergeStrategy allow_origin_regex = 2;
+        // mergeStrategy allow_methods = 3;
+        // mergeStrategy allow_headers = 4;
+
+        MergeStrategy expose_headers = 5;
```
sam-heilbron, line 409: In https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/options/cors/cors.proto#L41, a precedent was set where we put fields on a proto that is shared between Route and VirtualHost. With the addition of this new API, we will now have a couple different ways that users can configure CORS. Is that surface area of APIs intentional?  Or would there be value in adding this field directly to the CORS policy

sam-heilbron (commented):
A good start! As part of this can we be sure to do the following:
- Update our user-facing docs for the Edge APIs: https://docs.solo.io/gloo-edge/latest/guides/security/cors/
- Validate that this also works for Gloo Gateway (it should, because we re-use the same policies), and ensure any updates to docs are available in https://docs.solo.io/gateway/latest/security/cors/
- Schedule a demo/enablement of this work with field members 
- Potentially provide a dev build to be used/tested by requestors of this feature

projects/gloo/pkg/translator/route_config.go
```diff
@@ -912,3 +923,28 @@ func ValidateRoutePath(s string) error {
...
+		AllowCredentials: route.GetAllowCredentials(),
+		DisableForRoute:  route.GetDisableForRoute(),
+	}
+
+	// handle merging for ExposeHeaders field
+	// if either is nil, there is nothing to do
+	if vh.GetExposeHeaders() != nil && route.GetExposeHeaders() == nil {
+		switch mergeSettings.GetExposeHeaders() {
+		case v1.VirtualHostOptions_CorsMergeSettings_UNION:
+			out.ExposeHeaders = append(vh.GetExposeHeaders(), route.GetExposeHeaders()...)
```
sam-heilbron, line 945: What is the impact of having duplicated "ExposeHeaders" defined on the filter? I think a comment here could be helpful clarifying why we are ok with duplicates (if we are) or perhaps we deduplicate the headers

projects/gloo/api/v1/options.proto
bewebi: > Define this message in the same cors.proto file

This seems reasonable

> Follow the existing terminology (the existing message is CorsPolicy, so I think this one we could call CorsMergePolicy)

I'm not sure this follows
The new message is settings for merging CorsPolicies not a new type of CorsPolicy
I thought reusing the "Policy" term would confuse it
Perhaps `CorsPolicyMergeSettings` would get the best of both worlds?

projects/gateway/api/v1/virtual_service.proto
bewebi: At one point I was toying with modifying the VS proto and realized that this comment consisted of an incomplete description of the field followed by the description of the `Routes` field from the `VirtualHost` message copied verbatim
I figured I may as well clean it up while I was there, but can revert, or modify the replacement if desired

projects/gloo/pkg/translator/route_config.go
bewebi: This is a good catch

I had looked into this at one point, prior to PTO so it's fuzzy, but I think I didn't realize that the VH was in scope via `RouteParams` and therefore it appeared that the logic had to go earlier in the flow

I do think it would be appropriate to move this logic to the plugin given that all of the necessary pieces are in scope

test/e2e/cors_test.go
bewebi: I'm not sure why `MatchRegexp` is used, but it is already within the `HaveOkResponseWithHeaders` matcher
That matcher allows different assertions to be made for each header, not just equality, so it's probably worth keeping, so that we can make non-equality assertions on header values as needed

projects/gloo/api/v1/options.proto
bewebi: We got the feedback that the merge settings should be per-VS
I think exposing it on the route level as well, even if we documented that it wasn't intended to be used there, would add to confusion

> we will now have a couple different ways that users can configure CORS

There are still the same number of ways to configure CORS (Route and VH) but now some optionality around how to reconcile the two
Also at this point it's really just one option on one field, so we could narrow it to a single bool, similar to `disable_for_route`, but that reduces the future extensibility

projects/gloo/pkg/translator/route_config.go
bewebi: My assumption is that duplicates are not technically incorrect, just redundant, but I haven't really found documentation for that
The straight append was a simpler solution, but there's no particular reason not to de-dupe, the overhead is minimal

test/e2e/cors_test.go
sam-heilbron: `The rest of the assertions in the new test serve to validate that there are no other (unexpected) behavior changes` Yeah this is what I was focused on. Do we believe there is a reason those assertions would fail and therefore having the tests is helpful? I fear that we're duplicating tests that exist, which make it harder to identify exactly what is being tested (ie the expose-headers)

projects/gloo/pkg/translator/route_config.go
sam-heilbron: `My assumption is that duplicates are not technically incorrect, just redundant, but I haven't really found documentation for that` I think we can/should test this manually or with an e2e test to verify. I imagine the CORS spec has some commentary on it as well.

projects/gloo/api/v1/options.proto
sam-heilbron: Fair enough, just wanted to raise that we have a precedent for properties that are available on the Vhost and Route, and having a subset of those just apply at a single level. I'm happy with your direction though

changelog/v1.18.0-beta14/merge-cors-exposeheaders.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/7689
+    resolvesIssue: false
+    description: >-
+      Expose `CorsMergeSettings` on `VirtualHostOptions` which allows users to specify how to reconcile CORS settings
```
sam-heilbron, line 6: This field name/type is now out of date

projects/gateway/api/v1/virtual_service.proto
sam-heilbron: Oh no I'm good with any/all improvements to adjacent code. I  was confused because it felt like we were removing content that was relevant (ie what happens when multiple routes match). Can you link me to the user-facing documentation that expresses this behavior (that we're removing)?

projects/gloo/api/v1/options.proto
```diff
@@ -274,7 +274,8 @@ message VirtualHostOptions {
     headers.options.gloo.solo.io.HeaderManipulation header_manipulation = 2;
     // Defines a CORS policy for the virtual host
     // If a CORS policy is also defined on the route matched by the request, the route policy
-    // overrides the virtual host policy for any configured field.
+    // overrides the virtual host policy for any configured field unless CorsMergeSettings are specified that define an
```
sam-heilbron, line 277: CorsMergeSettings is now out of date

projects/gloo/api/v1/options.proto
```diff
@@ -441,7 +445,8 @@ message RouteOptions {
     // If true and there is a host rewrite, appends the x-forwarded-host header to requests.
     google.protobuf.BoolValue append_x_forwarded_host = 146;
     // Defines a CORS policy for the route
-    // If a CORS policy is also defined on the route's virtual host any fields set here override the virtual host configuration 
+    // If a CORS policy is also defined on the route's virtual host any fields set here override the virtual host configuration
+    // unless an alternate behavior is defined on the virtual host via CorsMergeSettings
```
sam-heilbron, line 449:
```suggestion
    // If a CORS policy is defined on both the route and the virtual host, the merge behavior for these policies is determined by the CorsPolicyMergeSettings defined on the VirtualHost.
```

projects/gloo/api/v1/options/cors/cors.proto
```diff
@@ -39,4 +39,31 @@ message CorsPolicy {
     // Optional, only applies to route-specific CORS Policies, defaults to false.
     // If set, the CORS Policy (specified on the virtual host) will be disabled for this route.
     bool disable_for_route = 8;
+}
+
+// Settings for determining how CORS settings are merged when present on both VirtualHost and Route
```
sam-heilbron, line 44: I think we can/should add some content about _why_ someone would want this. We have context that different personas tend to manage different resources, and in this case a Platform team which manages a VS wants to set properties that are enforced/inherited across all Routes that application teams may manage

projects/gloo/api/v1/options/cors/cors.proto
```diff
@@ -39,4 +39,31 @@ message CorsPolicy {
...
+message CorsPolicyMergeSettings {
+    enum MergeStrategy{
+        DEFAULT = 0; // Do not merge
```
sam-heilbron, line 47: I think that default/do-no-merge can be improved to clarify the behavior that exists...if we are not merging, then what configuration wins?

projects/gloo/pkg/plugins/cors/plugin.go
```diff
@@ -88,6 +88,15 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
 		return nil
 	}
 
+	// Envoy overrides vh-level CORS policy with route policy for any fields present on the route
+	// Therefore we implement our merge logic in-memory and, if necessary, modify the route-level CORS policy to
+	// have the desired merged value(s)
+	// If either is nil, merging will be ineffectual and we skip this step
+	if mergeSettings := params.VirtualHost.GetOptions().GetCorsPolicyMergeSettings(); mergeSettings != nil &&
```
sam-heilbron, line 95: Is this logic accurate? What if I have a VirtualHost that defines CORS, and merge settings that say "UNION". On the route level, there is no CORS policy defined, but because of UNION, the platform team is expecting their CORS headers to be inherited, right?

projects/gloo/pkg/plugins/cors/plugin.go
```diff
@@ -88,6 +88,15 @@ func (p *plugin) ProcessRoute(params plugins.RouteParams, in *v1.Route, out *env
...
+		params.VirtualHost.GetOptions().GetCors() != nil && in.GetOptions().GetCors() != nil {
```
sam-heilbron, line 96: Checking `in.GetOptions.GetCors()` is unnecessary I think, given that we have that guard at the beginning of the function

projects/gloo/pkg/plugins/cors/plugin.go
```diff
@@ -247,3 +256,42 @@ func getCorsResponseHeadersFromPolicy(corsPolicy *cors.CorsPolicy) []*envoy_conf
 		},
 	}
 }
+
+func mergeCors(mergeSettings *cors.CorsPolicyMergeSettings, vh, route *cors.CorsPolicy) *cors.CorsPolicy {
+	// we propagate the route setting by default
```
sam-heilbron, line 261: Is a nil cors policy and an empty one the same? If we pass into this function a nil route policy, what is the expected behavior?

projects/gloo/pkg/plugins/cors/plugin.go
```diff
@@ -247,3 +256,42 @@ func getCorsResponseHeadersFromPolicy(corsPolicy *cors.CorsPolicy) []*envoy_conf
...
+		ExposeHeaders:    route.GetExposeHeaders(),
+		MaxAge:           route.GetMaxAge(),
+		AllowCredentials: route.GetAllowCredentials(),
+		DisableForRoute:  route.GetDisableForRoute(),
+	}
+
+	// handle merging for ExposeHeaders field
+	// if either is nil, there is nothing to do
+	if vh.GetExposeHeaders() != nil && route.GetExposeHeaders() != nil {
+		switch mergeSettings.GetExposeHeaders() {
```
sam-heilbron, line 276: nit: In a case where we have explicitly defined options, I think it's valuable to enumerate them in the switch, even if they all do nothing or fallthrough

projects/gloo/pkg/plugins/cors/plugin.go
```diff
@@ -247,3 +256,42 @@ func getCorsResponseHeadersFromPolicy(corsPolicy *cors.CorsPolicy) []*envoy_conf
...
+		case cors.CorsPolicyMergeSettings_UNION:
+			out.ExposeHeaders = dedupe(append(vh.GetExposeHeaders(), route.GetExposeHeaders()...))
+		}
+	}
+
+	return out
+}
+
+func dedupe(in []string) []string {
```
sam-heilbron, line 285: I think if we are building a dedupe method we should consider the following:
- This feel like a generic function that could/should live in https://github.com/solo-io/gloo/tree/main/pkg/utils/stringutils
- With the introduction of generics could we instead write a generic function that works for types other than strings?
- Are there go libraries (like `slices`) that provide this functionality for us

projects/gateway/api/v1/virtual_service.proto
bewebi: What has been removed is quite literally already the description of the `Routes` field on `VirtualHost`
See [here](https://github.com/solo-io/gloo/blob/42dae66668f5bb7e83a4a510f3b70574b4d302ff/projects/gateway/api/v1/virtual_service.proto#L161-L163)

This is not super relevant when describing the `VirtualHost` field in the `VirtualService`

test/e2e/cors_test.go
bewebi: If the merge logic was buggy or Envoy changed its behavior these assertions could be useful, however there are now unit tests for the merge logic in the plugin and we trust Envoy not to have this significant a breaking change, so I'll remove the extra assertions

projects/gloo/pkg/translator/route_config.go
bewebi: I wasn't able to find documentation around this specifically but did implement dedupe logic (discussed elsehwere)

projects/gloo/pkg/plugins/cors/plugin.go
bewebi: If no CORS policy is defined on Route level, the VH policy will always be enforced

Merge strategies are only relevant if there are settings present on both levels

The only potential exception would be if/when we implement INTERSECTION
Then we might consider an empty field on one level and a populated field on the other to resolve to the empty set
I don't think this is what we'd desire, but not a decision we have to make now

projects/gloo/pkg/plugins/cors/plugin.go
bewebi: For the time being it's effectively the same, but this may change if/when we are evaluating boolean values since we can't distinguish between a `false` and an empty field

Regardless, if one is nil and the other isn't, we should honor the one that is not nil

projects/gloo/pkg/plugins/cors/plugin.go
bewebi: I haven't found an implementation of dedupe in Go libraries (`slices`, `strings`, etc.)

I'm adding an implementation using generics and adding to a new `sliceutils` package

pkg/utils/sliceutils/sliceutils.go
```diff
@@ -0,0 +1,15 @@
+package sliceutils
+
+func Dedupe[E comparable](in []E) []E {
```
sam-heilbron, line 3: 🚀

sam-heilbron (approved):

docs/content/guides/security/cors/_index.md
```diff
@@ -168,12 +168,189 @@ For more information, see the [API docs]({{% versioned_link_path fromRoot="/refe
...
+        - https://[a-zA-Z0-9]*.example.com
+        exposeHeaders:
+        - origin
+        maxAge: 1d
+      prefixRewrite: /api/pets
+...
+```
+
+By default, values that are set on the VirtualHost level are overridden by those set on the Route level. For example, try amending the VS as follows:
+{{< highlight yaml "hl_lines=16 21 27-32" >}}
```
bewebi, line 207: This section isn't rendering

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/cors/cors.proto.sk.md
```diff
@@ -54,6 +56,39 @@ CorsPolicy defines Cross-Origin Resource Sharing for a virtual service.
 
 
 
+---
+### CorsPolicyMergeSettings
+
+ 
+Settings for determining how CORS settings are merged when present on both VirtualHost and Route
+This may be useful if different teams/personas are responsible for managing VirtualService and Route resources and policies
+on VirtualServices set by one team or persona should be enforced or inherited on Routes
```
artberger, line 65: add punctuation; some grammar nits
```suggestion
Settings to determine how to merge CORS settings when present on both the VirtualHost and Route.
This option can be useful when different user personas or teams share ownership of a VirtualService. For example, you might not want CORS settings on each route to override the virtual host settings, but instead merge them with a `UNION` strategy.
```

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/cors/cors.proto.sk.md
```diff
@@ -54,6 +56,39 @@ CorsPolicy defines Cross-Origin Resource Sharing for a virtual service.
...
+
+
+---
+### MergeStrategy
+
+
+
+| Name | Description |
+| ----- | ----------- | 
+| `DEFAULT` | Follow the default Envoy behavior, which is for Route settings to override VH settings if non-nil |
```
artberger, line 86:
```suggestion
| `DEFAULT` | Follow the default Envoy behavior, which is for Route settings to override VirtualHost settings if non-nil. |
```

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options/cors/cors.proto.sk.md
```diff
@@ -54,6 +56,39 @@ CorsPolicy defines Cross-Origin Resource Sharing for a virtual service.
...
+| `UNION` | When a setting is present on both VH and Route CORS policy, merge by concatenating for list fields and by ORing for boolean fields |
```
artberger, line 87:
```suggestion
| `UNION` | When a CORS policy is present on both VirtualHost and Route, merge the settings. The merge is done by concatenating for list fields and by ORing for boolean fields. |
```

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk.md
```diff
@@ -352,7 +354,7 @@ to be usable by Gloo. (plugins currently need to be compiled into Gloo)
...
 | `appendXForwardedHost` | [.google.protobuf.BoolValue](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/bool-value) | If true and there is a host rewrite, appends the x-forwarded-host header to requests. |
-| `cors` | [.cors.options.gloo.solo.io.CorsPolicy](../options/cors/cors.proto.sk/#corspolicy) | Defines a CORS policy for the route If a CORS policy is also defined on the route's virtual host any fields set here override the virtual host configuration. |
+| `cors` | [.cors.options.gloo.solo.io.CorsPolicy](../options/cors/cors.proto.sk/#corspolicy) | Defines a CORS policy for the route If a CORS policy is defined on both the route and the virtual host, the merge behavior for these policies is determined by the CorsPolicyMergeSettings defined on the VirtualHost. |
```
artberger, line 357:
```suggestion
| `cors` | [.cors.options.gloo.solo.io.CorsPolicy](../options/cors/cors.proto.sk/#corspolicy) | Defines a CORS policy for the route. If a CORS policy is defined on both the route and the virtual host, the merge behavior for these policies is determined by the CorsPolicyMergeSettings defined on the VirtualHost. |
```

docs/content/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/options.proto.sk.md
```diff
@@ -280,6 +281,7 @@ to be usable by Gloo. (plugins currently need to be compiled into Gloo)
...
 | `stagedTransformations` | [.transformation.options.gloo.solo.io.TransformationStages](../options/transformation/transformation.proto.sk/#transformationstages) | Early transformations stage. These transformations run before most other options are processed. If the `regular` field is set in here, the `transformations` field is ignored. |
 | `extProc` | [.extproc.options.gloo.solo.io.RouteSettings](../enterprise/options/extproc/extproc.proto.sk/#routesettings) | Enterprise-only: External Processing filter settings for the virtual host. This can be used to override certain HttpListenerOptions settings, and can be overridden by RouteOptions settings. |
+| `corsPolicyMergeSettings` | [.cors.options.gloo.solo.io.CorsPolicyMergeSettings](../options/cors/cors.proto.sk/#corspolicymergesettings) | Settings for determining merge strategy for CORS settings when present at both Route and VH level. |
```
artberger, line 284:
```suggestion
| `corsPolicyMergeSettings` | [.cors.options.gloo.solo.io.CorsPolicyMergeSettings](../options/cors/cors.proto.sk/#corspolicymergesettings) | Settings for determining merge strategy for CORS settings when present at both Route and VirtualHost levels. |
```

sam-heilbron (approved):

sheidkamp (approved):
Approving to confirm requested changes are in. Did not review subsequent changes

bewebi:
/kick
```
Step #0 - "publish-artifacts": BUILD FAILURE: Build step failure: build step 2 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 1
Step #0 - "publish-artifacts": ERROR: (gcloud.builds.submit) build 15e032ae-6c7e-49f6-94b1-a471c5704dcf completed with status "FAILURE"
```
A number of CI actions failed with 502 responses from Quay as well

sam-heilbron (approved):

PR #9875, opened 2024-08-08
Warn on missing TLS secret

jbohanon:
# Description

Updates the condition of a VirtualService referencing a TLS secret that does not exist from an error state to a warning state. This is to allow for eventual consistency with VS creation and TLS secret creation.

## API changes
- Added warnings to Listener validation API

## Code changes
- Report missing tls secret errors as a Listener Warning instead of error

## Docs changes
TODO

# Context

Users ran into this eventual consistency issue when applying a cert-manager `Certificate` resource at the same time as a `VirtualService` resource. Because the `Certificate` does not synchronously create the TLS secret, the `VirtualService` is rejected by validation. 

## Interesting decisions
 
- curl pod added to kube2e test helper code because the image used previously does not support the `--connect-to` flag. this is to be less intrusive than refactoring the curl tool to support _both_ `--connect-to` AND `--resolve`.

## Testing steps

```sh
# if you don't have a cluster, create one 
kind create cluster
```
```sh
# Install gloo
helm repo add gloo https://storage.googleapis.com/solo-public-helm
helm repo update
# we are disabling validation here to allow the async application of secret/VS
helm install -n gloo-system gloo gloo/gloo --create-namespace --set gateway.validation.alwaysAcceptResources=true --set gateway.validation.allowWarnings=true

# Create a VS and TLS secret
k apply -f - << EOF
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
  name: tls-secret-1
  namespace: gloo-system
data:
  tls.crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUVQVENDQWlXZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFiTVFvd0NBWURWUVFEREFFcU1RMHcKQ3dZRFZRUUtEQVJ5YjI5ME1CNFhEVEl6TVRFd09ERTJORFExTjFvWERUTXpNVEV3TlRFMk5EUTFOMW93SGpFSwpNQWdHQTFVRUF3d0JLakVRTUE0R0ExVUVDZ3dIWjJGMFpYZGhlVENDQVNJd0RRWUpLb1pJaHZjTkFRRUJCUUFECmdnRVBBRENDQVFvQ2dnRUJBTit0V2hoa3QvNVFQTUw4UGorZ1JScUM1blp5TG9Sem5iK3hPazdQTVozRndtYUcKNThvbVhPRm16ZmJlK0VaaGE0UlBhK1BpdFhFbitjZkM5allYRU42dGM1WExWUjlKK1dCRXRhSUpoZlh2VzAvbgpraEg0MWFZa2NCQVMyTEh1U3l4WWd3VERMRzI1OUxVdVJFT3VGSVhtWUZJaGVlZTZ6V3dRMXk0Ujk1VzRoVGFzCi9JVk9wYmttbSsyM0ZVQ1Q3RTcvNzN0RFh3Q1dpekc3UnUyZ1p2aS9tK0ZRVUJCZmFPTGxzelQvVHNwNTB3YmUKY0hxY29UbWJNWUJpWDk1RFBYTWtnZ2g5M1R2bnBWb0taYVZhWDNOdHlGRGJOZnEyLzZaT2daNFlNZVgzb0VMUgpiVllpY01rU3lZRHJWbW9jeHZBMWdQQUsxd2NkVE1OcjlnY0F1b0VDQXdFQUFhT0JpRENCaFRBSkJnTlZIUk1FCkFqQUFNQXNHQTFVZER3UUVBd0lGNERBZEJnTlZIU1VFRmpBVUJnZ3JCZ0VGQlFjREFnWUlLd1lCQlFVSEF3RXcKREFZRFZSMFJCQVV3QTRJQktqQWRCZ05WSFE0RUZnUVVxaW40SEhXN08wR3NLbGgwUEJ2SFFZSzRQT013SHdZRApWUjBqQkJnd0ZvQVVZb3l3TXpJN1BpWGtJd3FMSTdkNzA5SmJnVmN3RFFZSktvWklodmNOQVFFTEJRQURnZ0lCCkFCQ3p2NUUrb3hYT3RBUi9UWWE2YWJMVm94WENPWFZVeVFRell5VFJoekJOY0Vjbk1NeDBHc2V1NHRXeUlmem8KNmQ5LzRkOWdmdDlRNnVTS1RZUkhYU0VIQUFsMmlEWGdQTTZoSk0vNmpxQlE2N1ErVkVrUTJWVUMySDZEYjF1Qwo2VGdldk9MdlA1eDhrS2FjZVNnYTdmSHZJcW95OHVmbm1BSzlhd2ZobE9hajBjUWcxQXV6aW14blhzTVQvdkNVCm9Xa2xPZSt0TDA2OEd3LytIMFJQMTJ6N2t6VGJDbXZuRWU4bk1QSXNrU21NZTcvUnZLcUFuYTd0NHFDOVdyRXgKWlFZK0NlOVhrTnI3RGJaZnprTmpqUFV2OEozdHN2dzY5Zm1HcVBEWVpHMm1HQjRzeGFyNy9mZG4rbGd3MDVsRwpBbEhhaWpXTlVGTWtmcXgvUzZnampEL0NPZVpRcVltM1hLY3RCWkFSUjNJRUR0RWNEUkRzNllvczhCUTBzbS9ECkhnNG1XWWR4Y283WlpnUzF6ZWlhdkNwRWxDaXEzWnB5c0EvS1NLS056Y1RIRk5acXBxYWFNdVQvaTBuaUI1MmEKZmpFSDUzdk1wYXhUK1IvcFIxQ1NMREZ6VDI3OTFMaEprZkJWVWwzQkdnOVY2VCt6bkl1Nkk1aVU1RlZZbVl3UApubS9EYitncE9JalEyU0w5Wm1nVkNvdjFvNTBTU3ZLN1RYSFVIaXFoZXlucDAyR1ZYYXpFT295ODBWcG04dW9JCklvS2NFOS9IUldOT0F5Uk41cWtYRDBnalpJOWM1aUtjV25kQ1B1cnBFZENNZUdiems1cWRUNGhZSFFsM2RKeW8KNHJoZm9MN051R1VUa0o1ZCtGdFBwaXRISkNEVGdUb05LTVNGcTcxbzRoMXkKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
  tls.key: LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JSUV1d0lCQURBTkJna3Foa2lHOXcwQkFRRUZBQVNDQktVd2dnU2hBZ0VBQW9JQkFRRGZyVm9ZWkxmK1VEekMKL0Q0L29FVWFndVoyY2k2RWM1Mi9zVHBPenpHZHhjSm1odWZLSmx6aFpzMzIzdmhHWVd1RVQydmo0clZ4Si9uSAp3dlkyRnhEZXJYT1Z5MVVmU2ZsZ1JMV2lDWVgxNzF0UDU1SVIrTldtSkhBUUV0aXg3a3NzV0lNRXd5eHR1ZlMxCkxrUkRyaFNGNW1CU0lYbm51czFzRU5jdUVmZVZ1SVUyclB5RlRxVzVKcHZ0dHhWQWsreE8vKzk3UTE4QWxvc3gKdTBidG9HYjR2NXZoVUZBUVgyamk1Yk0wLzA3S2VkTUczbkI2bktFNW16R0FZbC9lUXoxekpJSUlmZDA3NTZWYQpDbVdsV2w5emJjaFEyelg2dHYrbVRvR2VHREhsOTZCQzBXMVdJbkRKRXNtQTYxWnFITWJ3TllEd0N0Y0hIVXpECmEvWUhBTHFCQWdNQkFBRUNnZjhQVjFhUUwxaEVteTl0amc1WkR0SFFxRGFJZGZ0UUhwVmdCY0ZtMVVyZzBnSzAKWjZHa204V0REZ2ZqYlNlTlZ1RXhlRnFqV2RwWDkyeFhHbGNJdUV4SFgvZStsTXNRdnBWUllMcGhQblRuQU1YVgp0MG9rN1NYTFBVRll4OC8vcUcvWkZHTzA0UHJYNmFMRFNuZ0NBYXhxOFpNbFpFUkMyaUJKaTAwVXhGNHNKR25WCklJeHFLeVRnbWpBcFV6c1BDWXF5ci9aNmJTRkVnbVNxNzJobmt2Rm1PV3NYcldzRlZLY05iUUplWTRMMEJuWUsKWk5xODNmemt6ekpxbVh3OEVncWFPNlhWMmJtVmYrM1hSL3ViMDRGeXRya0Y5bE1JU0pWOUhkQmtRaGV2VzZhVApoSG5pblV6VkgyUlh4M1piYWw3ck1ZL0lOMWNmaFVlWm9BZDQ0ZGtDZ1lFQTh1bHorbXpPWXg0QVNnWUo0eVNICkdmc1VQeXpJUU9vaEVJQmtxMWNrUHJlbVhsdEdoWTcrRmtYYmh5cnVuOTRIREJJZWR0RWt0YjlZSDkvT0xNME8KK056TlhTYnlaQ3ZTa1U1ZlNNeUpDc3E3L01JTlVpNUxGN1FLQVF1ZXk3WDgvUGU3NHhFQ29hN1AvSlJkTnYrdgpjUnkyUEZOSlNGbTJGdkxGeDA4TDFWa0NnWUVBNjdxV3Q5dENpQ24zRXVPM2hEeFptZ3VSWU55Tk9TWkowdjR5Cm9zaXZ0WFl2ZGVYWGxGZWFWSEJtZG5vZUYrVTZ6TXpUQUY0d09jNmZpenpCMjlGVkxjWjJmTE5tWXc0RmxENFEKR0wzMHVJckQ5WXJraGZWZm85TW9aVHo2cWJnc0xQQnNZTXljelpFOThyb3dmZVp2MkUzQ1lRaDhOL1lGbXRmOQptWTViNFdrQ2dZQkQyc2pHQkl6bWpTUGhpYXhMWWhISFJTYlR1dXU1am0xc0VhR05aMHM5cGNsNGhDREFBRUNqCjhpR3ZzV04xRHUyREJyQ3gyaHhhRkxoR054dDkwazVEWUZLUm1lYU42dHZvTVM5V3c2UG9ldGRtZE1LSjJWcXEKcFdWQ0EzLzVRYjRJNEI4QS8raHZSOGpic29vVGFmc1ZLc01ST09hNHFpNitYRlM1SnpDVUNRS0JnRmZLN2tjYgpTZlFjYlFDRC90MG8vTlg2YVBLQ01iYVBJLytJM0tMenl6engvMHNSaHZDZ2o4SFMrdFkxTlBBQlY1emV5OWJmClBXYktKWEZkOTNVK3lWSjdEN1h4dXJnNWlLcGxVdWxrRmJpRk5lWkZERWMzMDU3WURidG1zcFJ6RzBEQmFodkQKR01NV3pOT1J0RzJ2WFFoYUxZS2wvbDE1S3kwNE5DTDBlaFBCQW9HQkFMK2RmaVFmUjhYVGMwMzF6R0xwQm1kdQpOMU81aDM1cUdIT0s4M25MY1VQSnhzc0JTSGVxNTZEVHBic3VMRjY0V1c2bk5KSWRwckdRLzNpOFZldzhDa05ZCmtXY2ZGWkdUTGM4b2g4Tkw4bWpiZGtITCs3M3ZuVy9FbEZRRDlWZTE3WWN0cEZKbUcwcVBSaFVWQzhkSG5RYnAKWXRMTTQxb01qQlQ3NUdjRjBZZ2wKLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQo=
---
kind: VirtualService
metadata:
  name: vs-1
  namespace: gloo-system
spec:
  sslConfig:
    sniDomains:
      - vs-1
    secretRef:
      name: tls-secret-1
      namespace: gloo-system
  virtualHost:
    domains:
    - 'vs-1'
    - 'vs-1:8443'
    routes:
    - matchers:
      - exact: /
      directResponseAction:
        status: 200
        body: "success from vs-1"
EOF
```
```sh
# Port-forward gateway-proxy
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443 &
export PF_PID=$!
```
```sh
# curl to validate that we're getting traffic
curl -k --connect-to vs-1:8443:127.0.0.1 https://vs-1:8443
```
```sh
# Apply a second VS referencing a TLS secret that does not exist
k apply -f - << EOF
kind: VirtualService
metadata:
  name: vs-2
  namespace: gloo-system
spec:
  sslConfig:
    sniDomains:
      - vs-2
    secretRef:
      name: tls-secret-2
      namespace: gloo-system
  virtualHost:
    domains:
    - 'vs-2'
    - 'vs-2:8443'
    routes:
    - matchers:
      - exact: /
      directResponseAction:
        status: 200
        body: "success from vs-2"
EOF
```
```sh
# curl to show we are still receiving traffic
curl -k --connect-to vs-1:8443:127.0.0.1 https://vs-1:8443
```
```sh
# restart gloo deployment to roll the pod
k rollout restart deploy/gloo -n gloo-system
k rollout status deploy/gloo -n gloo-system
```
```sh
# Port-forward gateway-proxy
kill $PF_PID
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443 &
export PF_PID=$!
```
```sh
# curl to show that we are NO LONGER receiving traffic, even on the good VS
curl -k --connect-to vs-1:8443:127.0.0.1 https://vs-1:8443
```
```sh
# build and load this branch
# note that upgrading isn't causing resources to update to the locally built versions
make kind-build-and-load build-test-chart
helm delete -n gloo-system gloo --wait
helm install -n gloo-system gloo _test/gloo-1.0.1-dev.tgz  --set gateway.validation.alwaysAcceptResources=false --set gateway.validation.allowWarnings=true
```
```sh
# force retranslation of gateway resource by deleting and reapplying the VS
k patch vs -n gloo-system vs-1 --type merge --patch="{\"metadata\":{\"annotations\":{\"manually-updated-at\":\"$(date)\"}}}"
```
```sh
# restart gloo deployment to roll the pod
k rollout restart deploy/gloo -n gloo-system
k rollout status deploy/gloo -n gloo-system
```
```sh
# Port-forward gateway-proxy
kill $PF_PID
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443 &
export PF_PID=$!
```
```sh
# curl to show that we are receiving traffic on the good VS, but not on the invalid VS
curl -k --connect-to vs-1:8443:127.0.0.1 https://vs-1:8443
curl -k --connect-to vs-2:8443:127.0.0.1 https://vs-2:8443
```
```sh
# in another terminal, watch the status on the gateway resource
watch -n1 "kubectl get -n gloo-system ggw gateway-proxy-ssl -o jsonpath='{.status}' | jq ."

# apply the secret that was missing
k apply -f - << EOF
apiVersion: v1
kind: Secret
type: kubernetes.io/tls
metadata:
  name: tls-secret-2
  namespace: gloo-system
data:
  tls.crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUVQVENDQWlXZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFiTVFvd0NBWURWUVFEREFFcU1RMHcKQ3dZRFZRUUtEQVJ5YjI5ME1CNFhEVEl6TVRFd09ERTJORFExTjFvWERUTXpNVEV3TlRFMk5EUTFOMW93SGpFSwpNQWdHQTFVRUF3d0JLakVRTUE0R0ExVUVDZ3dIWjJGMFpYZGhlVENDQVNJd0RRWUpLb1pJaHZjTkFRRUJCUUFECmdnRVBBRENDQVFvQ2dnRUJBTit0V2hoa3QvNVFQTUw4UGorZ1JScUM1blp5TG9Sem5iK3hPazdQTVozRndtYUcKNThvbVhPRm16ZmJlK0VaaGE0UlBhK1BpdFhFbitjZkM5allYRU42dGM1WExWUjlKK1dCRXRhSUpoZlh2VzAvbgpraEg0MWFZa2NCQVMyTEh1U3l4WWd3VERMRzI1OUxVdVJFT3VGSVhtWUZJaGVlZTZ6V3dRMXk0Ujk1VzRoVGFzCi9JVk9wYmttbSsyM0ZVQ1Q3RTcvNzN0RFh3Q1dpekc3UnUyZ1p2aS9tK0ZRVUJCZmFPTGxzelQvVHNwNTB3YmUKY0hxY29UbWJNWUJpWDk1RFBYTWtnZ2g5M1R2bnBWb0taYVZhWDNOdHlGRGJOZnEyLzZaT2daNFlNZVgzb0VMUgpiVllpY01rU3lZRHJWbW9jeHZBMWdQQUsxd2NkVE1OcjlnY0F1b0VDQXdFQUFhT0JpRENCaFRBSkJnTlZIUk1FCkFqQUFNQXNHQTFVZER3UUVBd0lGNERBZEJnTlZIU1VFRmpBVUJnZ3JCZ0VGQlFjREFnWUlLd1lCQlFVSEF3RXcKREFZRFZSMFJCQVV3QTRJQktqQWRCZ05WSFE0RUZnUVVxaW40SEhXN08wR3NLbGgwUEJ2SFFZSzRQT013SHdZRApWUjBqQkJnd0ZvQVVZb3l3TXpJN1BpWGtJd3FMSTdkNzA5SmJnVmN3RFFZSktvWklodmNOQVFFTEJRQURnZ0lCCkFCQ3p2NUUrb3hYT3RBUi9UWWE2YWJMVm94WENPWFZVeVFRell5VFJoekJOY0Vjbk1NeDBHc2V1NHRXeUlmem8KNmQ5LzRkOWdmdDlRNnVTS1RZUkhYU0VIQUFsMmlEWGdQTTZoSk0vNmpxQlE2N1ErVkVrUTJWVUMySDZEYjF1Qwo2VGdldk9MdlA1eDhrS2FjZVNnYTdmSHZJcW95OHVmbm1BSzlhd2ZobE9hajBjUWcxQXV6aW14blhzTVQvdkNVCm9Xa2xPZSt0TDA2OEd3LytIMFJQMTJ6N2t6VGJDbXZuRWU4bk1QSXNrU21NZTcvUnZLcUFuYTd0NHFDOVdyRXgKWlFZK0NlOVhrTnI3RGJaZnprTmpqUFV2OEozdHN2dzY5Zm1HcVBEWVpHMm1HQjRzeGFyNy9mZG4rbGd3MDVsRwpBbEhhaWpXTlVGTWtmcXgvUzZnampEL0NPZVpRcVltM1hLY3RCWkFSUjNJRUR0RWNEUkRzNllvczhCUTBzbS9ECkhnNG1XWWR4Y283WlpnUzF6ZWlhdkNwRWxDaXEzWnB5c0EvS1NLS056Y1RIRk5acXBxYWFNdVQvaTBuaUI1MmEKZmpFSDUzdk1wYXhUK1IvcFIxQ1NMREZ6VDI3OTFMaEprZkJWVWwzQkdnOVY2VCt6bkl1Nkk1aVU1RlZZbVl3UApubS9EYitncE9JalEyU0w5Wm1nVkNvdjFvNTBTU3ZLN1RYSFVIaXFoZXlucDAyR1ZYYXpFT295ODBWcG04dW9JCklvS2NFOS9IUldOT0F5Uk41cWtYRDBnalpJOWM1aUtjV25kQ1B1cnBFZENNZUdiems1cWRUNGhZSFFsM2RKeW8KNHJoZm9MN051R1VUa0o1ZCtGdFBwaXRISkNEVGdUb05LTVNGcTcxbzRoMXkKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
  tls.key: LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JSUV1d0lCQURBTkJna3Foa2lHOXcwQkFRRUZBQVNDQktVd2dnU2hBZ0VBQW9JQkFRRGZyVm9ZWkxmK1VEekMKL0Q0L29FVWFndVoyY2k2RWM1Mi9zVHBPenpHZHhjSm1odWZLSmx6aFpzMzIzdmhHWVd1RVQydmo0clZ4Si9uSAp3dlkyRnhEZXJYT1Z5MVVmU2ZsZ1JMV2lDWVgxNzF0UDU1SVIrTldtSkhBUUV0aXg3a3NzV0lNRXd5eHR1ZlMxCkxrUkRyaFNGNW1CU0lYbm51czFzRU5jdUVmZVZ1SVUyclB5RlRxVzVKcHZ0dHhWQWsreE8vKzk3UTE4QWxvc3gKdTBidG9HYjR2NXZoVUZBUVgyamk1Yk0wLzA3S2VkTUczbkI2bktFNW16R0FZbC9lUXoxekpJSUlmZDA3NTZWYQpDbVdsV2w5emJjaFEyelg2dHYrbVRvR2VHREhsOTZCQzBXMVdJbkRKRXNtQTYxWnFITWJ3TllEd0N0Y0hIVXpECmEvWUhBTHFCQWdNQkFBRUNnZjhQVjFhUUwxaEVteTl0amc1WkR0SFFxRGFJZGZ0UUhwVmdCY0ZtMVVyZzBnSzAKWjZHa204V0REZ2ZqYlNlTlZ1RXhlRnFqV2RwWDkyeFhHbGNJdUV4SFgvZStsTXNRdnBWUllMcGhQblRuQU1YVgp0MG9rN1NYTFBVRll4OC8vcUcvWkZHTzA0UHJYNmFMRFNuZ0NBYXhxOFpNbFpFUkMyaUJKaTAwVXhGNHNKR25WCklJeHFLeVRnbWpBcFV6c1BDWXF5ci9aNmJTRkVnbVNxNzJobmt2Rm1PV3NYcldzRlZLY05iUUplWTRMMEJuWUsKWk5xODNmemt6ekpxbVh3OEVncWFPNlhWMmJtVmYrM1hSL3ViMDRGeXRya0Y5bE1JU0pWOUhkQmtRaGV2VzZhVApoSG5pblV6VkgyUlh4M1piYWw3ck1ZL0lOMWNmaFVlWm9BZDQ0ZGtDZ1lFQTh1bHorbXpPWXg0QVNnWUo0eVNICkdmc1VQeXpJUU9vaEVJQmtxMWNrUHJlbVhsdEdoWTcrRmtYYmh5cnVuOTRIREJJZWR0RWt0YjlZSDkvT0xNME8KK056TlhTYnlaQ3ZTa1U1ZlNNeUpDc3E3L01JTlVpNUxGN1FLQVF1ZXk3WDgvUGU3NHhFQ29hN1AvSlJkTnYrdgpjUnkyUEZOSlNGbTJGdkxGeDA4TDFWa0NnWUVBNjdxV3Q5dENpQ24zRXVPM2hEeFptZ3VSWU55Tk9TWkowdjR5Cm9zaXZ0WFl2ZGVYWGxGZWFWSEJtZG5vZUYrVTZ6TXpUQUY0d09jNmZpenpCMjlGVkxjWjJmTE5tWXc0RmxENFEKR0wzMHVJckQ5WXJraGZWZm85TW9aVHo2cWJnc0xQQnNZTXljelpFOThyb3dmZVp2MkUzQ1lRaDhOL1lGbXRmOQptWTViNFdrQ2dZQkQyc2pHQkl6bWpTUGhpYXhMWWhISFJTYlR1dXU1am0xc0VhR05aMHM5cGNsNGhDREFBRUNqCjhpR3ZzV04xRHUyREJyQ3gyaHhhRkxoR054dDkwazVEWUZLUm1lYU42dHZvTVM5V3c2UG9ldGRtZE1LSjJWcXEKcFdWQ0EzLzVRYjRJNEI4QS8raHZSOGpic29vVGFmc1ZLc01ST09hNHFpNitYRlM1SnpDVUNRS0JnRmZLN2tjYgpTZlFjYlFDRC90MG8vTlg2YVBLQ01iYVBJLytJM0tMenl6engvMHNSaHZDZ2o4SFMrdFkxTlBBQlY1emV5OWJmClBXYktKWEZkOTNVK3lWSjdEN1h4dXJnNWlLcGxVdWxrRmJpRk5lWkZERWMzMDU3WURidG1zcFJ6RzBEQmFodkQKR01NV3pOT1J0RzJ2WFFoYUxZS2wvbDE1S3kwNE5DTDBlaFBCQW9HQkFMK2RmaVFmUjhYVGMwMzF6R0xwQm1kdQpOMU81aDM1cUdIT0s4M25MY1VQSnhzc0JTSGVxNTZEVHBic3VMRjY0V1c2bk5KSWRwckdRLzNpOFZldzhDa05ZCmtXY2ZGWkdUTGM4b2g4Tkw4bWpiZGtITCs3M3ZuVy9FbEZRRDlWZTE3WWN0cEZKbUcwcVBSaFVWQzhkSG5RYnAKWXRMTTQxb01qQlQ3NUdjRjBZZ2wKLS0tLS1FTkQgUFJJVkFURSBLRVktLS0tLQo=
EOF
```
```sh
# Port-forward gateway-proxy
kill $PF_PID
kubectl port-forward -n gloo-system deploy/gateway-proxy 8443 &
export PF_PID=$!
```
```sh
# curl to show that we are receiving traffic on both, now valid VS
curl -k --connect-to vs-1:8443:127.0.0.1 https://vs-1:8443
curl -k --connect-to vs-2:8443:127.0.0.1 https://vs-2:8443
```

# Checklist:


.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -32,10 +32,11 @@ runs:
       TEST_PKG: ./test/kubernetes/e2e/tests
       ISTIO_VERSION: ${{ inputs.istio-version }}
     shell: bash
-    run: make go-test
+    # we tee the output into a temp file and then grep through that for an overall summary of the run at the end.
+    run: 'mkdir ./_test/test_output && make go-test | tee ./_test/test_output/${{inputs.cluster-name}} || grep -E "(--- FAIL)|(--- PASS)" ./_test/test_output/${{inputs.cluster-name}}'
```
jbohanon, line 36: Revert this

pkg/utils/requestutils/curl/request.go
```diff
@@ -134,9 +134,9 @@ func (c *requestConfig) generateArgs() []string {
 	var fullAddress string
 
 	if c.sni != "" {
-		sniResolution := fmt.Sprintf("%s:%d:%s", c.sni, c.port, c.host)
+		sniResolution := fmt.Sprintf("%s:%d:%s:%d", c.sni, c.port, c.host, c.port)
```
sam-heilbron, line 137: Are we/developers guaranteed to have the port for the destination service match? If so, then this seems like a valuable adjustment. 

If we go forward with this, I think we need to update https://github.com/solo-io/gloo/blob/main/pkg/utils/requestutils/curl/option.go#L95

projects/gateway/pkg/reporting/add_proxy_validation_result.go
```diff
@@ -180,6 +182,14 @@ func getListenerLevelErrors(listenerReport *validation.ListenerReport) []error {
 
 	return listenerErrs
 }
+func getListenerLevelWarnings(listenerReport *validation.ListenerReport) []string {
+	listenerWarnings := validationutils.GetListenerWarning(listenerReport)
+
+	// TODO(jbohanon) implement warnings on various listener types and account for them here
```
sam-heilbron, line 188: Is this a todo for future work? Or do you consider this to be a blocker for finishing the work (either in this PR, or a fast-follow)?

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -121,9 +121,25 @@ message ListenerReport {
         string reason = 2;
     }
 
+    // warning types for the given listener config
+    message Warning {
+        enum Type {
+            SSLConfigWarning = 0;
```
sam-heilbron, line 127: Should we introduce a default "catch-all" warning? Or will that encourage the practice of relying on that catch-all and not specifying a specific type?

projects/gloo/api/grpc/validation/gloo_validation.proto
```diff
@@ -121,9 +121,25 @@ message ListenerReport {
         string reason = 2;
     }
 
+    // warning types for the given listener config
+    message Warning {
+        enum Type {
```
sam-heilbron, line 126: Instead of "SslConfigWarning", how do you feel about a type  "UnresolvedReference". And it represents a situationt when a ref couldn't be resolved, and the "reason" has details about it. In my view, this would allow us to use this type if/when other instances of resolving references fails, that aren't specific to SslConfig

projects/gloo/pkg/translator/clusters.go
```diff
@@ -122,7 +122,11 @@ func (t *translatorInstance) initializeCluster(
 		applyDefaultsToUpstreamSslConfig(sslConfig, t.settings.GetUpstreamOptions())
 		cfg, err := utils.NewSslConfigTranslator().ResolveUpstreamSslConfig(*secrets, sslConfig)
 		if err != nil {
-			reports.AddError(upstream, err)
+			if errors.Is(err, utils.SslSecretNotFoundError) {
```
sam-heilbron, line 125: Can/should we tie in https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/configuration_error.go#L18 somehow? Basically there is a paradigm in code around distinguishing error/warnings types in our plugin model, that we may want to extend to translators. 

You can see it in use here: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/translator/reporting.go#L173

projects/gloo/pkg/translator/filter_chain.go
```diff
@@ -294,7 +294,16 @@ func (h *httpFilterChainTranslator) createFilterChainsFromSslConfiguration(
 		// get secrets
 		downstreamTlsContext, err := h.sslConfigTranslator.ResolveDownstreamSslConfig(snap.Secrets, sslConfig)
 		if err != nil {
-			validation.AppendListenerError(h.parentReport, validationapi.ListenerReport_Error_SSLConfigError, err.Error())
+			if errors.Is(err, utils.SslSecretNotFoundError) {
```
sam-heilbron, line 297: SImilar to the above comment, it might make more sense to have a generic error/warning check, instead of gating this functionality on a specific type

projects/gloo/pkg/translator/listener_subsystem_test.go
```diff
@@ -489,7 +501,7 @@ var _ = Describe("Listener Subsystem", func() {
 						SslConfig: &ssl.SslConfig{
 							SslSecrets: &ssl.SslConfig_SecretRef{
 								SecretRef: &core.ResourceRef{
-									Name:      "secret-that-is-not-in-snapshot",
```
sam-heilbron, line 492: Is there value in both of these cases: a secret that exists but is invalid AND a secret that does not exist?

test/kubernetes/e2e/features/server_tls/types.go
```diff
@@ -0,0 +1,127 @@
...
+		StatusCode: http.StatusOK,
+		Body:       gomega.ContainSubstring("success from vs-with-oneway"),
+	}
+	// These codes are defined at https://curl.se/libcurl/c/libcurl-errors.html.
+	// These were determined experimentally.
+	expectedFailedResponseCodeInvalidVs = 16
+	expectedFailedResponseCertRequested = 35
+)
+
+func manifestFromFile(fname string) ([]byte, error) {
```
sam-heilbron, line 115: nit: Since this is used in our tests, and we likely never expect an error, we could turn this into a `mustManifestFromFile` that asserts no error is returned

sam-heilbron (commented):
Looking great!  A few questions just around how specific/generic we go with our error apis and code

projects/gloo/api/grpc/validation/gloo_validation.proto
bewebi: I gather that this is meant to be a parallel to the `SSLConfigError`

projects/gloo/api/grpc/validation/gloo_validation.proto
bewebi: We don't have a corresponding catchall in the Error Type enum and I think it's reasonable to maintain consistency with that precedent

bewebi (commented):
Overall seems to do what it claims, and I mainly have minor comments in that regard
I have not followed the manual testing steps myself

I wonder what the customer requirements are and whether we've considered:
- Feature flagging this so that users can decide whether they want to error or warn on missing TLS secrets
- Error when deleting secrets in use while only warning when creating VS without secret (noted on review comment as well)
- Persist warning message to VS status for easier troubleshooting

I think these would be improvements, but I'm not looking to expand the scope beyond what's needed, which is why I ask about what the requirements are

projects/gloo/pkg/utils/ssl.go
```diff
@@ -31,8 +31,9 @@ var (
 		return eris.Errorf("ocsp staple policy %v not a valid policy", p)
 	}
 
-	SslSecretNotFoundError = func(err error) error {
-		return eris.Wrapf(err, "SSL secret not found")
+	SslSecretNotFoundError = eris.New("SSL secret not found")
+	sslSecretNotFoundError = func(err error) error {
+		return eris.Wrapf(err, SslSecretNotFoundError.Error())
```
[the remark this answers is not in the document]
jbohanon: Allows checking `errors.Is(err, SslSecretNotFound)`

pkg/utils/requestutils/curl/request.go
jbohanon: Good call on changing the link in the comment; will do

Wrt the ports, we are not changing the behavior because `--resolve` doesn't allow providing a port (just short-circuits dns), and so it would always have used the same port anyway. As far as I can tell, the only difference here is that `--connect-to` allows us to provide a string host or an ip whereas `--connect` only allows the latter.

projects/gateway/pkg/reporting/add_proxy_validation_result.go
jbohanon: I was leaving that as a nugget for future cases where we may need to add listener subtype warnings, but I think it's out of scope for this PR

projects/gloo/api/grpc/validation/gloo_validation.proto
jbohanon: Yeah I would rather encourage specificity by convention than encourage an unspecific error, which would be devilish to debug in a running system.

projects/gloo/api/grpc/validation/gloo_validation.proto
jbohanon: I really like the idea of an `UnresolvedReferece` warning type. I will have to take a look at if there is a reasonable place to add that without restructuring a bunch of code though, because I think this warning follows reasonable precedent and fulfills the design intent of this work.

projects/gloo/pkg/translator/listener_subsystem_test.go
jbohanon: Yes. They are separate problems that we need to make sure are being handled appropriately. I changed the ListenerError test here to reference an existing invalid secret and added a separate tabke entry for ListenerWarning which handles the missing reference case

test/kubernetes/e2e/features/server_tls/types.go
```diff
@@ -0,0 +1,127 @@
...
+		Body:       gomega.ContainSubstring("success from vs-2"),
+	}
+	expectedHealthyResponseWithOneWay = &matchers.HttpResponse{
+		StatusCode: http.StatusOK,
+		Body:       gomega.ContainSubstring("success from vs-with-oneway"),
+	}
+	// These codes are defined at https://curl.se/libcurl/c/libcurl-errors.html.
+	// These were determined experimentally.
+	expectedFailedResponseCodeInvalidVs = 16
+	expectedFailedResponseCertRequested = 35
```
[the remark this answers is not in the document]
jbohanon: Yeah good catch I intended to link to the manpage for curl and use these constants. I will make that change

test/kubernetes/e2e/features/server_tls/suite.go
```diff
@@ -0,0 +1,299 @@
...
+	s.T().Cleanup(func() {
+		// ordering here matters if strict validation enabled
+		err := s.testInstallation.Actions.Kubectl().Delete(s.ctx, s.vs1, "-n", s.ns)
+		s.NoError(err, "can delete vs1 manifest file")
+		err = s.testInstallation.Actions.Kubectl().Delete(s.ctx, s.vs2, "-n", s.ns)
+		s.NoError(err, "can delete vs2 manifest file")
+		s.testInstallation.Assertions.EventuallyObjectsNotExist(s.ctx, vs1, vs2)
+		err = s.testInstallation.Actions.Kubectl().Delete(s.ctx, s.tlsSecret1)
+		s.NoError(err, "can delete tls secret manifest file")
+		// tlsSecret2 is deleted in the process of the test.
```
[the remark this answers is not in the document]
jbohanon: Good catch. Copypasta error; will fix

test/kubernetes/e2e/features/server_tls/types.go
jbohanon: I returned the error because it is not in a suite that has access to the assertions, but I suppose I could use the testify `require` package directly to achieve this?

test/kubernetes/e2e/features/server_tls/suite.go
```diff
@@ -0,0 +1,299 @@
...
+	return &serverTlsTestingSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+		ns:               testInst.Metadata.InstallNamespace,
+	}
+}
+
+func (s *serverTlsTestingSuite) SetupSuite() {
+	var err error
+	// We are substituting our namespace via os.ExpandEnv in order to place our referenced
```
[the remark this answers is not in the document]
jbohanon: In `./types.go` we define this variable as a function which reads the manifest in from the filesystem and performs `os.ExpandEnv` on it to turn the `${INSTALL_NAMESPACE}` instances in the file into our actual namespace set during test initialization (see `tests/`)

test/kube2e/gateway/gateway_test.go
```diff
@@ -1324,7 +1334,7 @@ var _ = Describe("Kube2e: gateway", func() {
 			It("correctly routes to the service (tcp)", func() {
 				responseString := fmt.Sprintf(`"hostname":"%s"`, gatewayProxy)
 
-				httpEcho.CurlEventuallyShouldRespond(helper.CurlOpts{
```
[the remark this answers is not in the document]
jbohanon: Yeah I think that's probably true but I'm not sure how else httpEcho might be in use. It is deployed very widely, mostly to curl I gather but I am not certain of that.

projects/gloo/pkg/translator/clusters.go
jbohanon: I think that would be a valuable expansion of the configuration error paradigm, but it seems tangential to this work. wdyt?

test/kubernetes/e2e/features/server_tls/types.go
jbohanon: Looking at it a little closer, I thing we would need to plumb through a `testing.T` or otherwise use a gomega or panic directly. I think I am ok with this returning like this and asserting in the test

projects/gloo/pkg/utils/ssl.go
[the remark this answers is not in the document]
bewebi: I believe the first argument to `eris.Wrapf` is the error whose type will be preserved
We're just passing `SslSecretNotFoundError.Error()`, a `string`, so I wouldn't expect the output to preserve the type of `SslSecretNotFoundError`, but that error doesn't appear to have a special type in any event

From the [docs](https://pkg.go.dev/github.com/rotisserie/eris#Wrapf):
```
Wrapf adds additional context to all error types while maintaining the type of the original error.
```
The original error in this case is what's passed to `sslSecretNotFoundError()`, and the functionality is no different than it was previously as far as I can tell

test/kube2e/gateway/gateway_test.go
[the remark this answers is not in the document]
bewebi: Indeed it does look like we request `httpEcho` elsewhere even though we don't curl _from_ that pod

projects/gloo/pkg/utils/ssl.go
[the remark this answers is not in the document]
jbohanon: Without defining a base error we have to either check by calling the error construction func with a bogus, non-nil error, or do string literal matching which is needlessly fragile. 

Put another way, how do you propose we check that an error we have in hand was derived from this var without exporting a base error?

test/kubernetes/e2e/features/server_tls/suite.go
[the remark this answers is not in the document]
bewebi: Gotcha, I follow now
I think the comment could be clearer, something like:
```suggestion
	// These functions each substitute our namespace for placeholders in the given manifest file via os.ExpandEnv in order to place our referenced
```

test/kubernetes/e2e/features/validation/validation_reject_invalid/suite.go
```diff
@@ -114,12 +114,16 @@ func (s *testingSuite) TestVirtualServiceWithSecretDeletion() {
 		gloo_defaults.GlooReporter,
 	)
 
-	// failing to delete a secret that is in use
-	output, err := s.testInstallation.Actions.Kubectl().DeleteFileWithOutput(s.ctx, validation.Secret, "-n", s.testInstallation.Metadata.InstallNamespace)
-	s.Assert().Error(err)
-	s.Assert().Contains(output, fmt.Sprintf(`admission webhook "gloo.%s.svc" denied the request`, s.testInstallation.Metadata.InstallNamespace))
-	s.Assert().Contains(output, fmt.Sprintf("failed validating the deletion of resource"))
-	s.Assert().Contains(output, fmt.Sprintf("SSL secret not found: list did not find secret %s.tls-secret", s.testInstallation.Metadata.InstallNamespace))
+	// attempting to delete a secret that is in use produces a warning but succeeds
```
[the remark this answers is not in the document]
bewebi: That's what I figured

Perhaps we can call out somewhere, changelog if nowhere else, that Secrets referenced by VS are no longer protected from accidental deletion?
I do think this is a compelling reason to feature flag this

projects/gloo/pkg/utils/ssl.go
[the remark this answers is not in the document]
bewebi: I'm just not clear that this is what's happening

Unless I'm missing something, `SslSecretNotFoundError.Error() == "SSL secret not found"`, so I don't think the code is functionally different than before

Perhaps what we want is
```
sslSecretNotFoundError = func(err error) error {
		return eris.Wrapf(SslSecretNotFoundError, err.Error())
}
```

Or maybe I'm totally missing something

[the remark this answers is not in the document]
jbohanon: 1. This is demonstrated to work in https://github.com/solo-io/gloo/pull/9875/files#diff-6889591de58279e09f13775dd29722474f1ae4bddce67eda35916e47436a1fa7R125 and its tests. 
2. Due to how [errors.Is](https://pkg.go.dev/errors#Is) works, it doesn't actually matter which wraps which as long as both are represented. 
3. However, as we are adding context to the underlying raw error I think it makes sense to wrap it the way it's written. 

I see the point that we're basically just making an error var and then ripping the string out of it, but having the var exported as `error` in lieu of `string` makes the external code that uses it cleaner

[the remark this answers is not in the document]
jbohanon: Here is the impl code for eris error types showing that the `Is` methods do a string comparison

https://github.com/rotisserie/eris/blob/master/eris.go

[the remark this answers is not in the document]
bewebi: Gotcha, thanks for clarifying
I was misunderstanding how `Is()` worked, but more importantly missing that we were newly using `Is()`, hence the promotion of the string to an exported Error

It may be worth adding a comment on `SslSecretNotFoundError` to explain that it's declared here so that we can check for it via `Is()` elsewhere, because without that context it's not clear why we bother having an exported Error rather than a string

bewebi (approved):
LGTM! Appreciate the thorough testing updates to go with the new Helm value/Settings field

test/kubernetes/e2e/features/admin_server/suite.go
```diff
@@ -50,7 +44,7 @@ func (s *testingSuite) TestGetInputSnapshotIncludesSettings() {
 			Name:      kubeutils.GlooDeploymentName,
 			Namespace: s.testInstallation.Metadata.InstallNamespace,
 		},
-		inputSnapshotContainsElement(s.testInstallation, v1.SettingsGVK, metav1.ObjectMeta{
+		s.testInstallation.Assertions.InputSnapshotContainsElement(v1.SettingsGVK, metav1.ObjectMeta{
```
sam-heilbron, line 47: 🚀

test/kubernetes/e2e/tests/manifests/validation-always-accept-helm.yaml
```diff
@@ -20,6 +20,9 @@ gateway:
   validation:
     allowWarnings: true
     alwaysAcceptResources: true
+    # this is needed to run the ServerTls tests which verify that with this setting, we
```
sam-heilbron, line 23: really helpful comment! I also like that even though this is the defaut, we are explicit with it being defined.

sam-heilbron (approved):

jbohanon:
dozer

PR #9888, opened 2024-08-12
Ensure namespace override consistency

jbohanon:
# Description

## Namespace pollution
Currently we have some tests which rely on programmatically setting the `INSTALL_NAMESPACE` env var. This was introduced to be able to expand that var into various resource manifests for edge routing resources. The application of this logic was introduced inconsistently across all of the different test installation suites we currently run, and was causing flakes due to ordering if not cleaned up properly. This PR applies the same logic to all test installation suites (except upgrade which is a special case) and makes sure we clean up after ourselves if we set the env var.

## Test utils unit tests
Test utils unit tests at `test/kubernetes/testutils/helper` have been migrated to run on a GHA in lieu of in cloudbuild since they require an authorized github token. This is to unblock community PRs coming in from forks of the repo.

## Nightlies timeout update
We [recently hit a failure in our nightly tests](https://github.com/solo-io/gloo/actions/runs/10363898580) due to a GHA timeout. In this PR we bump the timeout to 120 minutes for the go test command and 150 minutes for the overall action.


sam-heilbron (approved):

test/kubernetes/e2e/tests/zz_all_tests_test.go
```diff
@@ -0,0 +1,39 @@
...
+		"TestK8sGatewayIstio":                           TestK8sGatewayIstio,
+		"TestK8sGatewayMinimalDefaultGatewayParameters": TestK8sGatewayMinimalDefaultGatewayParameters,
+		"TestK8sGatewayNoValidation":                    TestK8sGatewayNoValidation,
+		"TestK8sGateway":                                TestK8sGateway,
+		"TestRevisionIstioRegression":                   TestRevisionIstioRegression,
+		"TestK8sGatewayIstioRevision":                   TestK8sGatewayIstioRevision,
+		"TestUpgradeFromCurrentPatchLatestMinor":        TestUpgradeFromCurrentPatchLatestMinor,
+		"TestUpgradeFromLastPatchPreviousMinor":         TestUpgradeFromLastPatchPreviousMinor,
+		"TestValidationAlwaysAccept":                    TestValidationAlwaysAccept,
+		"TestValidationStrict":                          TestValidationStrict,
```
npolshakova, line 27: Hm, this feels like another place a developer can forget to add a test to. Maybe we can run all of `^TestGlooGateway` in the nightlys to have at least one test that's a catch all? 

Or alternatively rename some of these tests in the ci workflow so they follow a pattern?
- TestValidation -> (TestValidationAlwaysAccept, TestValidationStrict)
- TestK8sGateway
- TestIstio

jbohanon: Yeah I was worried about adding a new place for a dev to forget to add a test...

WDYT about running all of `^Test` in the nightlies and checking for any output that doesn't start with `TestGlooGateway/`

npolshakova: Yeah that sounds like a good check to make sure all the test are being run!

jbohanon:
Nightly action kicked off here to test https://github.com/solo-io/gloo/actions/runs/10376999685

test/kubernetes/e2e/tests/zz_all_tests_test.go
```diff
@@ -0,0 +1,39 @@
+package tests_test
+
+import "testing"
+
+var (
+	allTests = map[string]func(*testing.T){
```
sam-heilbron, line 6: I think when defining a large map like this, it can help to have a comment outlining what the variables is used for (ie what it contains), what keys in the map represent and any considerations developers should have when reading/modifying this map

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -18,6 +18,10 @@ inputs:
   istio-version:
     required: true
     description: The version of Istio to use
+outputs:
+  test-output:
```
sam-heilbron, line 22: nit: test is in the name of the action, so it feels unnecessary. Also, it might be useful to clarify _what_ the output is. Is it a directory, a file, is it guaranteed to always exist?

.github/workflows/pr-unit-tests.yaml
```diff
@@ -36,11 +36,11 @@ jobs:
     - name: Run Tests
       shell: bash
       env:
-        TEST_PKG: ./projects/gateway2/...
+        TEST_PKG: "./projects/gateway2/... ./test/kubernetes/testutils/helper"
```
sam-heilbron, line 39: nit: Would we ever want to run this as a separate step?

test/kubernetes/e2e/tests/disabled_transformation_validation_test.go
```diff
@@ -2,34 +2,46 @@ package tests_test
...
 	"github.com/solo-io/gloo/test/kubernetes/testutils/helper"
+	"github.com/solo-io/gloo/test/testutils"
 	"github.com/solo-io/skv2/codegen/util"
 )
 
 // TestTransformationValidationDisabled is the function which executes a series of tests against a given
 // installation where validation has disableTransformationValidation=true
 func TestTransformationValidationDisabled(t *testing.T) {
 	ctx := context.Background()
+	installNs, overrodeNs := envutils.LookupOrDefault(testutils.InstallNamespace, "disable-transformation-validation-test")
```
sam-heilbron, line 23: In the first iteration of this work, the installNamespace represented the namespace where the Gloo installation was located. It allowed us to theoretically run multiple installations at once, where each one is isolated by its NS. 

With these changes (that pre-date this change), we are _also_ setting a local environment variable to match this value. I imagine this is because we are supporting templating resource manifests, and dynamically providing that variables at runtime. This is certainly a cool/valuable technique, but I wonder if instead of setting an env variable far away from where it is used, we can instead have a transformation function which says "for a given set of manifests AND an installation" generate the resources. That way the transform can set env variables under the hood, but all tied to the single source of truth: the test installation.


not a requirement for this work, since I see you are improving what already exists. I just wanted to share some thoughts while it was fresh

test/kubernetes/e2e/tests/zz_all_tests_test.go
```diff
@@ -0,0 +1,39 @@
...
+		"TestValidationAlwaysAccept":                    TestValidationAlwaysAccept,
+		"TestValidationStrict":                          TestValidationStrict,
+	}
+)
+
+// TestGlooGateway runs all of the defined Kubernetes E2E tests for Gloo Gateway.
+// By wrapping all of the tests in this parent, we ensure that we get a single
+// summary at the end of the test run, and that tests fail fast and do not proceed
+// to run other tests.
+func TestGlooGateway(t *testing.T) {
```
sam-heilbron, line 35: Do you see this as a short-term solution or a long-term one? I like the point about aggregating results if possible, though I wonder if we'd want to put results into a file (or even emit metrics based on the name of the test that failed). If you see this as a long-term solution, I'm good with it. If you see this as purely short-term, we may want to call that out

test/kubernetes/e2e/tests/zz_all_tests_test.go
jbohanon: Grepping the output turns out to be less straightforward in GHA than I anticipated so I am rolling back the summaries part of this PR to get the other incremental improvements in.

test/kubernetes/e2e/tests/zz_all_tests_test.go
jbohanon: I've rolled back these changes to move the other pieces of this PR forward.

.github/workflows/pr-unit-tests.yaml
jbohanon: Possibly but it didn't seem worth it to me for this iteration, especially since the number of tests in the next package is so low. Also, having it this way demonstrates that it _can_ be done this way.

.github/workflows/composite-actions/kubernetes-e2e-tests/action.yaml
jbohanon: Removed this part of the work from this PR

test/kubernetes/e2e/tests/disabled_transformation_validation_test.go
jbohanon: Yeah I think that a utility taking the installation and the resource templates would be a good way to do it. It raises the question of how to handle templating. I think that work needs some deeper thought and this effort at least brings everything to a consistent place with the current paradigm.

sam-heilbron (approved):

npolshakova (approved):

PR #9917, opened 2024-08-19
Add ginkgo to make install-go-tools command

tjons:
# Description

This PR adds `ginkgo` as a installation requirement in the `_output/bin` directory when running `make install-go-tools`. `ginkgo` is a requirement of the test harness, and should be installed similar to other go tools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ install-go-tools: mod-download ## Download and install Go dependencies
 	go install github.com/cratonica/2goarray
 	go install github.com/golang/mock/mockgen
 	go install github.com/saiskee/gettercheck
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.17.0
+	go install github.com/onsi/ginkgo/v2/ginkgo@$(GINKGO_VERSION)
 	# This version must stay in sync with the version used in CI: .github/workflows/static-analysis.yaml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -167,6 +167,7 @@ install-go-tools: mod-download ## Download and install Go dependencies
 	go install github.com/cratonica/2goarray
 	go install github.com/golang/mock/mockgen
 	go install github.com/saiskee/gettercheck
+	go install github.com/onsi/ginkgo/v2/ginkgo@v2.17.0
```
nfuden, line 170: Can we link this to the go mod please

lgadban (commented):
This tripped me up as well, but it already exists in a different target;
see https://github.com/solo-io/gloo/blob/main/Makefile#L214

Makefile
jenshu, line 170: this is already installed as part of `install-test-tools`

tjons:
Closing as this is part of a different target.

Makefile
tjons: 👍 after internal discussion, we will add it here and preserve it in the `install-test-tools` target as CI uses a lighter dependency set.

tjons: 👍

sam-heilbron (approved):
Alternative would be to invoke the `install-test-tools` as part of the `install-go-tools`

PR #9919, opened 2024-08-19
Makefile: Add a generate-changelog target

timflannagan:
# Description

This commit introduces a new `generate-changelog` target that makes it easy to
generate changelog entry stubs. The new target calls the existing bash script
under-the-hood and mirrors the way that target was implemented in the GME
repository. This helps improve devexp as we're interfacing with make targets
directly (vs. calling bash scripts) and the target shows up the the `help` output.

# Checklist:


Makefile
```diff
@@ -421,6 +421,10 @@ generated-code-apis: clean-solo-kit-gen go-generate-apis fmt ## Executes the tar
 .PHONY: generated-code-cleanup
 generated-code-cleanup: getter-check mod-tidy update-licenses fmt ## Executes the targets necessary to cleanup and format code
 
+.PHONY: generate-changelog ## Generate a changelog entry
+generate-changelog:
+	@./ci/changelog.sh
+
```
timflannagan, line 427: I could move this to it's own section:

```make
# Generate <something>
```

But I felt like that was overkill.

Makefile
```diff
@@ -421,6 +421,10 @@ generated-code-apis: clean-solo-kit-gen go-generate-apis fmt ## Executes the tar
 .PHONY: generated-code-cleanup
 generated-code-cleanup: getter-check mod-tidy update-licenses fmt ## Executes the targets necessary to cleanup and format code
 
+.PHONY: generate-changelog
+generate-changelog: ## Generate a changelog entry
+	@./devel/tools/changelog.sh
+
```
timflannagan, line 427: See https://solo-io-corp.slack.com/archives/C06C8RA01NF/p1724096357041519 for some context. Jacob mentioned that the changelog.sh bash script supports a single "edit" parameter, so we could define a CHANGELOG_ARGS-esq variable (e.g. `make generate-changelog CHANGELOG_ARGS=--edit`) that gets passed to the script, but that felt like overkill as the script doesn't support much user input right now.

sam-heilbron (approved):

Makefile
```diff
@@ -421,6 +421,10 @@ generated-code-apis: clean-solo-kit-gen go-generate-apis fmt ## Executes the tar
 .PHONY: generated-code-cleanup
 generated-code-cleanup: getter-check mod-tidy update-licenses fmt ## Executes the targets necessary to cleanup and format code
 
+.PHONY: generate-changelog
+generate-changelog: ## Generate a changelog entry
+	@./devel/tools/changelog.sh
```
jbohanon, line 426:
```suggestion
	@./devel/tools/changelog.sh edit
```

jbohanon (approved):

Makefile
timflannagan: Let's add that as a follow-up if folks feel like it's the better UX.

PR #9922, opened 2024-08-20
fix: fix NPE if an empty params is provided

yuval-k:
- fix NPE if an empty params is provided (added unit test).
- added missing values to values.yaml so that `helm template` works.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/pull/9922


projects/gateway2/helm/gloo-gateway/values.yaml
```diff
@@ -21,6 +21,12 @@ gateway:
   service:
     type: LoadBalancer
 
+  image:
+    pullPolicy: IfNotPresent
+
+  istio:
+    enabled: false
```
jbohanon, line 28: The work done in https://github.com/solo-io/gloo/pull/9657 explicitly removed as many values as possible from these values in favor of helm defaults; what do we gain by adding default values here that are opaque to the user?

yuval-k: I wanted `helm template ./projects/gateway2/helm/gloo-gateway/` to work. It was convenient to be able to render the chart like that  when i was debugging something

jenshu (approved):
we need to figure out the defaulting (or erroring) behavior when default gatewayparams don't exist, but the getter fix looks good to me

changelog/v1.18.0-beta17/fix-empty-gwp.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/pull/9922
```
jenshu, line 3: nit - this is not an issue link, just a link to this PR

jenshu (approved):

changelog/v1.18.0-beta17/fix-empty-gwp.yaml
yuval-k: yep, there was no issue, because i created the PR immediately after finding the bug without filing an issue.. so figured a link to the PR will be a good reference to what was the  issue

projects/gateway2/helm/gloo-gateway/values.yaml
jbohanon: Ah yeah that makes sense. That was the purpose of this file still existing in _at all_ so sounds good.

jbohanon (approved):

PR #9923, opened 2024-08-20
.github,Makefile: Add on-demand support for generating conformance reports

timflannagan:
# Description

Add on-demand support for generating Kubernetes Gateway API conformance reports. This introduces a new
GHA workflow_dispatch workflow that allows users to run the standard or experimental conformance suite
against an arbitrary Gloo release or the latest locally built main branch image. When the "experimental" option
has been enabled, automation will generate a conformance report and that report will be uploaded to the
workload's artifacts when the suite succeeds:

Screenshot 2024-08-20 at 3 12 37 PM

In order to test with a local built off main, you can leave the optional "version" field unset/empty/etc.

# Context

Right now, the conformance report publishing process is completely manual and we haven't updated the gateway-api repository in a while. See https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/reports/v1.0.0/gloo-gateway the latest report we published. We need an automated way to reduce toil and improve reproducibility for this process.

## Testing steps

See https://github.com/timflannagan/gloo/actions/workflows/conformance-tests.yaml for testing results
in my fork.

## Notes for reviewers

# Checklist:


.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,55 @@ runs:
...
+      fi
   - name: Setup test env
     shell: bash
     env:
       CLUSTER_NODE_VERSION: ${{ matrix.kube-version.node }}
       IMAGE_VARIANT: ${{ matrix.image-variant }}
       CONFORMANCE: "true"
     run: ./ci/kind/setup-kind.sh
-  - name: Install Gloo Gateway Enterprise with k8s gateway integration enabled
+  - name: Install Gloo Gateway with the k8s gateway integration enabled
```
timflannagan, line 40: s/Gloo Gateway Enterprise/Gloo Gateway since I wasn't able to find any evidence we were enabling enterprise features.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,55 @@ runs:
...
+    run: |
+      # We want to conditionally set the VERSION variable based on the matrix value
+      if [[ -z "${{ matrix.version }}" ]]; then
+        echo "VERSION=$(make print-VERSION)" >> $GITHUB_ENV
+      else
+        # else, set the VERSION variable to the matrix value and
+        # skip the docker build step so we use real image tags from
+        # the helm repository.
+        echo "VERSION=${{ matrix.version }}" >> $GITHUB_ENV
+        echo "SKIP_DOCKER=true" >> $GITHUB_ENV
```
timflannagan, line 31: Setting SKIP_DOCKER in this fashion is a bit hacky, but I didn't want to refactor the setup-kind.sh bash script to meet this use cases. In the case that a user has provided a version input, we can skip building and loading GG images into the kind cluster being created in that script.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,55 @@ runs:
...
+        # If matrix.version is empty, use the local chart path specified in the Makefile.
+        helm upgrade -i -n gloo-system gloo ./_test/gloo-${VERSION}.tgz --create-namespace \
+          --set kubeGateway.enabled=true \
+          --set global.image.variant=${{ matrix.image-variant }}
+      else
+        # Else, use the provided version to install Gloo from the helm repository.
+        helm upgrade -i -n gloo-system gloo gloo/gloo --version ${VERSION} --create-namespace \
+          --set kubeGateway.enabled=true \
+          --set global.image.variant=${{ matrix.image-variant }}
+      fi
```
timflannagan, line 53: One potential follow-up here is we invest more time into using helm repositories instead of local chart files. We use this pattern in GME quite a bit.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,55 @@ runs:
...
+  - name: Capture debug information when tests fail
     if: ${{ failure() }}
     shell: bash
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
+  - name: Upload conformace reports if profile is experimental
+    if: "${{ matrix.profile == 'experimental' }}"
+    uses: actions/upload-artifact@v4
+    with:
+      name: conformance-gloo-gateway-report
+      path: _test/conformance/report.yaml
```
timflannagan, line 71: One potential follow-up is adding another step -- either in this composite or in the workflow_dispatch workflow -- for a.) producing a slack notification, or b.) automatically creating a PR against k8s-sigs/gateway-api with the report results. For the latter, there's still some pieces missing in the release process & open questions that need to be burned in order to do this smoothly.

Makefile
```diff
@@ -1187,17 +1186,26 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
 # Targets for running Kubernetes Gateway API conformance tests
 #----------------------------------------------------------------------------------
 
+# FIXME(tim): Re-evaluate the --version & --contact flags. I think both are fine, but need to validate.
+# TODO(tim): We should have a single Makefile target to drive all of this...
+# Note(tim): Looks like adding the "TLS" profile breaks everything. We obviously don't support the "MESH" profile either.
+# Note(tim): All of the profiles and suites were refactored in the release-1.1 branch. The profiles became more granular
+# 			 among other things like the experimental_conformance_test.go file being removed.
```
timflannagan, line 1193: Whoops, I'll remove this comment once we get consensus on these changes.

.github/workflows/nightly-tests.yaml
```diff
@@ -404,13 +404,14 @@ jobs:
                         { node: 'v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245', kubectl: 'v1.29.2', kind: 'v0.20.0', helm: 'v3.14.4' }]
         image-variant:
           - standard
+        profile:
+        - standard
     steps:
```
timflannagan, line 409: This is the for the kube_gateway_api_conformance_tests_main nightly. IMO, we could run through both the standard and experimental suites here.

.github/workflows/regression-tests.yaml
```diff
@@ -82,6 +82,8 @@ jobs:
         # (see https://github.com/solo-io/solo-projects/issues/6094)
         image-variant:
           - standard
+        profile:
+        - standard
```
timflannagan, line 86: For pull requests, I think we just want to run through the standard suite even through the runtime for the experimental suite is fairly quick. Any objections? For PRs that introduce experimental features, we could run the on-demand workflow to validate those changes (or wait for nightly results) on top of the regression suite running through the core conformance checks.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
...
+        helm upgrade -i -n gloo-system gloo gloo/gloo --version ${VERSION} --create-namespace \
+          --set kubeGateway.enabled=true \
+          --set global.image.variant=${{ matrix.image-variant }}
+      fi
+  - name: Run the kubernetes gateway API conformance tests
     shell: bash
-    run: make conformance
-  - name: print events
+    run: |
+      # Default to the standard profile if none is provided. Otherwise,
```
jenshu, line 57: do you think there's a need to support both standard and experimental, or should we always use experimental (which includes everything in standard)?

timflannagan: I think in the short term, we could always run the experimental suite, especially since we don't support all features yet or conformance profiles, e.g. TLS, mesh, etc. The main benefit for separating these suites is managing the overall CI runtime for PRs, so if the experimental functionality has a longer runtime than standard, then having this separation could be beneficial.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
...
     run: ./ci/kind/setup-kind.sh
-  - name: Install Gloo Gateway Enterprise with k8s gateway integration enabled
+  - name: Install Gloo Gateway with the k8s gateway integration enabled
     shell: bash
     run: |
-      helm upgrade -i -n gloo-system gloo ./_test/gloo-1.0.0-ci1.tgz --create-namespace \
-        --set kubeGateway.enabled=true \
-        --set global.image.variant=${{ matrix.image-variant }}
-  - name: Run kubernetes gateway api conformance tests
+      if [[ -z "${{ matrix.version }}" ]]; then
```
sam-heilbron, line 43: We are following an existing convention here, so feel free to ignore....

I don't like how this composite-action, which should be a standalone/re-usable component, has references to a 'matrix' in it, because that means we are coupled to using it in a certain way. In https://github.com/solo-io/solo-projects/blob/main/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml I tried to decouple these by supplying inputs to the action, and therefore allow the workflow that invokes it to specify those directly.

Not a requirement, just figured I'd share my perspective so you could have it

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
...
+      fi
+  - name: Run the kubernetes gateway API conformance tests
     shell: bash
-    run: make conformance
-  - name: print events
+    run: |
+      # Default to the standard profile if none is provided. Otherwise,
+      # assume the profile is experimental.
+      if [[ -z ${{ matrix.profile }} ]]; then
+        make conformance
```
sam-heilbron, line 60: This means that a profile must be selected and can only be one or another...would there be value in supplying a list of profiles? And we invoke the corresponding target per-profile? We could have targets like `conformance-{profile}` in a standard naming convention, and the default `conformance` just calls `conformance-standard`

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
...
+      else
+        make conformance-experimental
+      fi
+  - name: Capture debug information when tests fail
     if: ${{ failure() }}
     shell: bash
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
+  - name: Upload conformace reports if profile is experimental
+    if: "${{ matrix.profile == 'experimental' }}"
```
sam-heilbron, line 69: nit: I think a guard like this could use a small comment for future devs. Over time, it will be a lot more challenging to recall why we only upload it for experiemental

.github/workflows/conformance-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          - distroless
+      profile:
+        description: "The conformance profile to run."
+        required: true
+        type: choice
+        options:
+          - standard
+          - experimental
+      version:
+        description: "Optional: Specify an existing Gloo Gateway release tag to deploy and test (e.g., 1.17.3). Leave empty to use the default branch."
```
sam-heilbron, line 22: nit: This comment confuses me a little, since we are referring to image tags and the a 'default branch'.

.github/workflows/conformance-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+  # TODO: Trigger on release events too.
+  workflow_dispatch:
+    inputs:
+      image-variant:
+        description: "The image variant to use."
+        required: true
+        type: choice
+        options:
+          - standard
+          - distroless
```
sam-heilbron, line 13: I'm not sure how you'd prefer to capture this, but https://github.com/solo-io/solo-projects/issues/6094, outlines why only standard images are supported. I would be in favor of leaving this as a breadcrumb for future devs, but I also undersatnd it's non-standard practice to link private issues in oss code

.github/workflows/conformance-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+      profile:
+        description: "The conformance profile to run."
+        required: true
+        type: choice
+        options:
+          - standard
+          - experimental
+      version:
```
sam-heilbron, line 21: If I wanted to build these images from source, and not use the main branch (ie use an LTS branch like 1.17.x), how would I trigger the workflow?

.github/workflows/conformance-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  run-conformance-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO(tim): Avoid hardcoding versions here. It's a bit tricky based on
+        # how this was setup and there's a limited # of dispatch inputs that GH
+        # supports. We can revisit this later.
+        kube-version:
```
sam-heilbron, line 34: Check out https://github.com/solo-io/gloo/blob/3f8a060e072425f8bc10af149a4ca1ac6c632c10/.github/workflows/pr-kubernetes-tests.yaml#L74 for a way to rely on these versions in a similar way as our other kube workflows

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
sam-heilbron: I think it would be good to have a slack notification on start/success/failure. https://github.com/solo-io/gloo/blob/main/.github/workflows/trivy-analysis-scheduled.yaml#L46 is a basic example of a way to at least notify on failure. wdyt?

ci/kind/setup-kind.sh
```diff
@@ -58,7 +58,11 @@ function create_kind_cluster_or_skip() {
 create_kind_cluster_or_skip
 
 if [[ $SKIP_DOCKER == 'true' ]]; then
+  # TODO(tim): refactor the Makefile & CI scripts so we're loading local
+  # charts to real helm repos, and then we can remove this block.
   echo "SKIP_DOCKER=true, not building images or chart"
+  helm repo add gloo https://storage.googleapis.com/solo-public-helm
```
sam-heilbron, line 64: nit: Given that we are leaving the helm repo work as a todo, are these `repo add` and `repo update` commands necessary in this PR?

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
...
+    uses: actions/upload-artifact@v4
+    with:
+      name: conformance-gloo-gateway-report
+      path: _test/conformance/report.yaml
```
sam-heilbron, line 73: If we run conformance tests against multiple supported versions (say kubernetes min version, kubernetes max version), would we have conflict because we would upload reports of the same name?

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
sam-heilbron: Good catch!  Nothing in gloo oss should have Enterprise

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan: Yup, totally agreed. This was something I also noticed and had a local TODO in my backlog to fix this as a potential follow-up. Lemme queue up a quick tech debt issue to re-evaluate / fix this approach.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan: Good point - I think this will get refactored once Gloo supports the 1.1 release. Do you think it's still worth chasing in the short term?

I'm also evaluating some other follow-up changes, e.g. manually vendoring these test suites, that would have implications on this approach too. Unfortunately, we just need to get this workflow introduced to main so it shows up in the "Actions" tab, and then further iteration can be done by overriding the branch being selected for this action.

.github/workflows/conformance-tests.yaml
timflannagan: Another good question, and I thought I covered this in the PR description, but I realize now that we could benefit from some README documentation for this automation.

> If I wanted to build these images from source, and not use the main branch (ie use an LTS branch like 1.17.x), how would I trigger the workflow?

You would leave the optional "version" input field empty. If you wanted to run conformance against a real image tag, you'd populate that field and automation would stamp out that image tags' release in the gloo helm chart registry.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan: Yeah, seems reasonable to me as a potential follow-up item. Do we have a slack channel that we can use, even if it's internal to solo's slack instance?

.github/workflows/conformance-tests.yaml
timflannagan: > I'm not sure how you'd prefer to capture this, but https://github.com/solo-io/solo-projects/issues/6094, outlines why only standard images are supported

Thanks, that's good context to know. 

>  I would be in favor of leaving this as a breadcrumb for future devs, but I also undersatnd it's non-standard practice to link private issues in oss code

Very true. My preference is decoupling OSS from the private enterprise repo. That obviously has largely implications on process, planning, etc. that we'd need to form some consensus with the team. Let's talk internally on this at some point. I have some experience in how to do this and I'd like to get your thoughts as well.

sam-heilbron (approved):
LGTM! We agreed in slack that this is a first pass, which we can iterate on more easily once it's checked into the main branch./

.github/workflows/conformance-tests.yaml
jenshu: side note but i think we can actually get the image variants working now but it does require some code changes first - i added a [comment](https://github.com/solo-io/solo-projects/issues/6094#issuecomment-2302907258) to the existing issue

timflannagan:
Internal slack link for the convo that Sam was mentioning: https://solo-io-corp.slack.com/archives/C03MFATU265/p1724181719824309.

ci/kind/setup-kind.sh
timflannagan: I did fairly extensive testing in my fork, and I didn't capture why this is necessary in my notes, but afaik, this was needed or automation failed to find the helm registry. I'll circle back on this as a follow-up (™️).

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -17,23 +17,57 @@ runs:
   - uses: azure/setup-helm@v4
     with:
       version: ${{ matrix.kube-version.helm }}
+  - name: Set and retrieve environment variables
+    shell: bash
+    run: |
+      # We want to conditionally set the VERSION variable based on the matrix value
```
jenshu, line 23: covered by sam's [comment below](https://github.com/solo-io/gloo/pull/9923/files#r1725651798) but this section could also use a refactor so that it's only using `inputs` instead of reading values from the matrix

Makefile
```diff
@@ -167,7 +167,6 @@ install-go-tools: mod-download ## Download and install Go dependencies
 	go install github.com/cratonica/2goarray
 	go install github.com/golang/mock/mockgen
 	go install github.com/saiskee/gettercheck
-	go install github.com/onsi/ginkgo/v2/ginkgo@$(GINKGO_VERSION)
```
jenshu, line 170: i think this should  be reverted; it was added recently by https://github.com/solo-io/gloo/pull/9917

timflannagan: Oh man, nice catch. I'm not sure too sure how this got deleted, but I totally agree we need to revert it.

Makefile
```diff
@@ -1190,14 +1189,18 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
 # Pull the conformance test suite from the k8s gateway api repo and copy it into the test dir.
 $(TEST_ASSET_DIR)/conformance/conformance_test.go:
 	mkdir -p $(TEST_ASSET_DIR)/conformance
-	echo "//go:build conformance" > $@
```
jenshu, line 1193: if this is removed, will the `-tags conformance` below still work?

timflannagan: Yeah, afaik, the Go toolchain will ignore tags that don't exist. Obviously, there's no need for specifying a non-existent tag, so I'll purge that in the next commit. Do you know offhand why this tag was originally introduced?

timflannagan:
Sorry for the force push - I forked this repo and got my local branch into a weird spot.

Makefile
timflannagan: Updated in https://github.com/solo-io/gloo/pull/9923/commits/14111d0868efae97aed34a33803d7efb3dab4e29.

Makefile
timflannagan: Updated in https://github.com/solo-io/gloo/pull/9923/commits/14111d0868efae97aed34a33803d7efb3dab4e29.

Makefile
jenshu: > Do you know offhand why this tag was originally introduced?

not sure, from git history looks like @yuval-k  added it

Makefile
```diff
@@ -1190,23 +1190,35 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
...
 
 # Run only the specified conformance test. The name must correspond to the ShortName of one of the k8s gateway api
 # conformance tests.
 conformance-%: $(TEST_ASSET_DIR)/conformance/conformance_test.go
-	go test -ldflags=$(LDFLAGS) -tags conformance -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS) \
+	go test -ldflags=$(LDFLAGS) -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS) \
 	-run-test=$*
 
+.PHONY: conformance-experimental ## Run the extended conformance test suite
+conformance-experimental: CONFORMANCE_ARGS += -conformance-profiles=HTTP -report-output=$(TEST_ASSET_DIR)/conformance/$(VERSION)-report.yaml -organization=solo.io -project=gloo -version=$(VERSION) -url=github.com/solo-io/gloo -contact=github.com/solo-io/gloo/issues/new/choose
```
jenshu, line 1215: should project be `gloo-gateway`? (not sure exactly how/where this value is used)
what is `-contact` supposed to point to?

jenshu (approved):
LGTM as a first pass, approving so we can iterate on it!

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan: I'll look into doing this with the #slack-integration-testing channel you added me to a couple of days ago.

Makefile
timflannagan: Fixed in #9932.

PR #9927, opened 2024-08-20
go.*: Remove the stale github.com/Sirupsen/logrus replace directive

timflannagan:
# Description

A quick maintenance PR for removing a stale replace directive. This was previously needed _way_ back in time when the original logrus repo was renamed. See https://github.com/sirupsen/logrus/issues/570 for more details.

# Checklist:


Implausiblyfun:
can you please attach the go mod graph before and after to show that this replace is not in affect?

timflannagan:
@Implausiblyfun Just to quickly clarify: for the mod graph, what do you suggest? Running `go mod why ...` locally and pasting the output or did you have something else in mind? In either case, I'm happy to do that tomorrow and paste the output as a comment when I get back to my comp.

With that said, I'm now wondering whether the current evidence, which is running `go mod tidy` doesn't result in any go.sum changes, and `cat`ing the go.sum file itself  locally doesn't contain this directive could be concluded as sufficient evidence? AFAIK, this used to be a problem when the repository was renamed, but indirect repositories that consume this library hadn't updated their go.mod files, and it was causing wide spread conflicts. Since then, the ecosystem has adjusted as we can remove the replace pin to get some short term maintenance wins.

nfuden:
yeah just something quick to show that the replace has no effect therefore the removal is a net win.

go.mod
```diff
@@ -326,7 +326,6 @@ require (
 )
 
 replace (
-	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.4.2
```
sam-heilbron, line 329: If we are removing this here, can we also be sure to remove it https://github.com/solo-io/solo-projects/blob/main/go.mod#L503?

sam-heilbron (approved):

davidjumani (approved):

timflannagan:
@nfuden I ran this from a different branch, but here's the why output:

```bash
$ dev :: /work/gloo ‹chore/update-contributing› » go mod why github.com/Sirupsen/logrus      
# github.com/Sirupsen/logrus
(main module does not need package github.com/Sirupsen/logrus)
```

I think this is sufficient along with `go mod tidy` producing no go.sum changes, and the go.sum file doesn't contain this entry either.

PR #9931, opened 2024-08-22
[AI] Semantic Cache API Updates, (mode, score_threshold)

EItanya:
# Description

AI Semantic caching API updates

## API changes

- Added a mode field to semantic caching to allow the user to selectively add new entries to the cache
- Added a `score_threshold` field to semantic caching to allow the user to decide how similar a prompt must be in order for caching to trigger.


projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -301,19 +301,35 @@ message SemanticCache {
   message Redis {
     // Connection string to the Redis database
     string connection_string = 1;
+
+    // Similarity score threshold value between 0.0 and 1.0 that determines how similar
+    // two queries need to be in order to return a cached result.
+    // The lower the number, the more similar the queries need to be for a cache hit.
+    //
+    // +kubebuilder:validation:Minimum=0
```
shashankram, line 309: Our code gen via solo-kit likely isn't enabling kubebuilder markers, we may want to do that

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -301,19 +301,35 @@ message SemanticCache {
...
   // Data store from which to cache the request/response pairs
   message DataStore {
     oneof datastore {
       Redis redis = 1;
     }
   }
+  enum Mode {
+    // Read and write to the cache as a part of the request/response lifecycle
+    READ_WRITE = 0;
+    // Only read from the cache, do not write to it. Data will be written to the cache outside the request/response cycle.
```
shashankram, line 322: nit: drop the period or add it to all comments for consistency

projects/gloo/api/v1/enterprise/options/ai/ai.proto
```diff
@@ -301,19 +301,35 @@ message SemanticCache {
...
+    READ_ONLY = 1;
+  }
   // Which data store to use
   DataStore datastore = 1;
   // Model to use to get embeddings for prompt
   Embedding embedding = 2;
   // Time before data in the cache is considered expired
   uint32 ttl = 3;
+  // Cache mode to use: READ_WRITE or READ_ONLY
```
shashankram, line 331: Maybe another comment line on what it does similar to the enum desc

shashankram (approved):

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -249,6 +249,8 @@ spec:
 {{- if (($gateway.aiExtension).enabled) }}
       - name: gloo-ai-extension
         image: "{{ template "gloo-gateway.gateway.image" $gateway.aiExtension.image }}"
+        args:
+        - "extproc"
```
shashankram, line 253: the entire spec is opaque, so perhaps this should be exposed via GatewayParameters as well. However, I would rather not make this change and instead use an env var and let the process use that instead.

EItanya: This is still opaque, essentially I wanted to re-use the program rather than creating a new image for the apiserver. If we make this configurable the user can accidentally not include this, but it's necessary 100% of the time if this is enabled, so it belongs here. Really this entire spec should be located alongside the code rather than in a different repo.

projects/gloo/api/v1/enterprise/options/ai/ai.proto
EItanya: That's a much bigger issue, I agree with you but probably not for right now.

projects/gloo/api/v1/enterprise/options/ai/ai.proto
EItanya: Why? if a user wants more detail they can click into the `enum`, this is just meant as high level.

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
shashankram: It isn't because "extproc" doesn't make sense without knowing about the container being used and its impl. Why not use an env var as it allows to do exactly what you are intending to?

EItanya: What do you mean "use an env var", where would we use one that's opaque to the user?

shashankram: You can default the impl to use a sidecar and we can set an env var when running as an apiservice?

sam-heilbron (approved):

projects/gloo/api/v1/enterprise/options/ai/ai.proto
shashankram: Makes it easier in the doc to now click another link, nothing more

PR #9940, opened 2024-08-23
Docs redirects and version bumps

Nadine2016:
# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


docs/build-docs.sh
```diff
@@ -48,7 +48,239 @@ firebaseJson=$(cat <<EOF
...
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/:version/security/auth/:path",
+        "destination": "/gloo-edge/:version/guides/security/auth/:path/",
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/v1.13/:path",
+        "destination": "/gloo-edge/latest/",
```
artberger, line 120: I personally am fine with this redirect since the old versions are quite outdated. However, I think in theory we could add `:path` to the end, and it might redirect them to the content (if the content is still at the same path, that is).

docs/build-docs.sh
```diff
@@ -48,7 +48,239 @@ firebaseJson=$(cat <<EOF
...
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/:version/dev/writing_auth_plugins/",
+        "destination": "/gloo-edge/:version/guides/dev/writing_auth_plugins/",
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/:version/enterprise/",
+        "destination": "/gloo-edge/:version",
```
artberger, line 150: might not be needed, but sometimes we have the trailing slash at the end of this, sometimes we don't...it looks like if we don't put it in the browser, the URL gets redirected to the trailing slash anyways.

docs/build-docs.sh
```diff
@@ -48,7 +48,239 @@ firebaseJson=$(cat <<EOF
...
+        "destination": "/gloo-edge/:version/guides/security/auth/extauth/:path",
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/:version/guides/traffic_management/destination_types/grpc_to_rest/",
+        "destination": "/gloo-edge/:version/guides/traffic_management/destination_types/grpc/",
+        "type":"301"
+      },
+      {
+        "source": "/gloo-edge/:version/latest/guides/traffic_management/request_processing/transformations/",
```
artberger, line 234: Is this purposeful? the duplicate `/:version/latest/`

(but if it's not, then I don't know how the path is different than what we put in destination, so it might be, just wanted to confirm)

docs/build-docs.sh
Nadine2016: yeah the problem is that if that path does not exist, we will have other broken links that the SEO team wants us to fix. that's why I just use this general path

docs/build-docs.sh
Nadine2016: yes apparently we have links with latest/latest and latest/main haha

artberger (approved):

Rachael-Graham (approved):

PR #9958, opened 2024-08-26
Front Channel Logout

sheidkamp:
# Description
API changes in support of implementing [Front Channel Logout](https://openid.net/specs/openid-connect-frontchannel-1_0-final.html)

## API changes
Added FrontChannelLogout message to OidcAuthorizationCode.

Currently only defines the `Path` field, but was defined as its own message in case we need to configure valdiation options down the road.

## Code changes
Generated code only

## Docs changes
Update OIDC logout docs to include front channel logout

# Context
Clients want to implement front channel logout in their OIDC integrated applications

## Testing steps
Testing steps will be outlined in `solo-projects` PR

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1861,6 +1869,14 @@ message ExtAuthConfig {
             // To turn on Redis user session caching, use the `userSessionConfig` field.
             RedisOptions claims_caching_options = 4;
         }
+
+       // For the moment this is just path, but we may want to configure things like iss/sid validation
+        message FrontChannelLogout {
```
davidjumani, line 1874: Can this be extracted if it grows and won't require changes in multiple places ?

sam-heilbron: When you say extracted, are you imagining that it is a single message, used in both places? If so, I actually think we want to avoid doing that. For re-use purposes it's great. But the two APIs that it's used in are actually separate APIs: one is user-facing and defines the fields availabe in our CRs, one is service-facing and defines the fields available that are sent over xDS to the ext-auth-service. If we create a single message, it ties those two APIs together, and will create a breaking change if we need to introduce content only in one API or the other.

changelog/v1.18.0-beta20/fc_logout.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NEW_FEATURE
```
sam-heilbron, line 2: nit: I tend to create this as non-user-facing since it's just an API, and an API for a feature that is only availabe in enterprise. And then I put the NEW_FEATURE description in the enterprise PR

changelog/v1.18.0-beta20/fc_logout.yaml
```diff
@@ -0,0 +1,6 @@
...
+    issueLink: https://github.com/solo-io/gloo/issues/9510
+    resolvesIssue: false
+    description: >-
+      Add API support for OIDC [Front Channel Logout](https://openid.net/specs/openid-connect-frontchannel-1_0-final.html) with the OidcAuthorizationCode.FrontChannelLogout field
```
sam-heilbron, line 6: nit: if we are going with the new_feature changelog, I think being explicit about the impacted CR and that actual new feild, will help users

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1861,6 +1869,14 @@ message ExtAuthConfig {
...
+            // Path to use for front channel logout. Should not be the same as logout or callback paths.
+            string path=1;
+        }
+
+        FrontChannelLogout front_channel_logout = 28;
```
sam-heilbron, line 1879: nit: Could it be useful clarifying when to use this, and how it relates to the logout field?

sam-heilbron (approved):

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: Also if by "extracted" you mean just put in the path at the top level now and move it later if needed, that ends up being a breaking change to the API.

changelog/v1.18.0-beta20/fc_logout.yaml
sheidkamp: Went with "NON_USER_FACING", will have details in enterprise PR.

changelog/v1.18.0-beta20/fc_logout.yaml
sheidkamp: Updated

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: Updated here and the other FrontChannelLogout field

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -857,6 +857,17 @@ message OidcAuthorizationCode {
...
         RedisOptions claims_caching_options = 4;
     }
+
+    // For the moment this is just path, but we may want to configure things like iss/sid validation
+    message FrontChannelLogout {
+        // Path to use for front channel logout. Should not be the same as logout or callback paths.
+        string path=1;
+    }
+
+    // Configuration for front channel logout. This is used to log out the user from multiple apps/clients associated with one OIDC provider (OP)
```
jbohanon, line 867: nit: common terminology for this would be OIDC Identity Provider (IdP)

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -857,6 +857,17 @@ message OidcAuthorizationCode {
...
+    // When this is configured here and in OP, when the user goes to the logout endpoint they will be directed to a page OP's end_session endpoint which
+    // will render an iframe which calls the FrontChannelLogout.path for each logged in app/client.
```
jbohanon, line 869: This comment is confusingly worded. Maybe it was meant to say something like
```suggestion
    // When this is configured here and in OP, when the user goes to the logout endpoint they will be redirected to the OP's end_session endpoint which
    // will render an iframe which calls the FrontChannelLogout.path for each logged in app/client.
```

jbohanon, line 869: additionally, I think we don't need to mention the iframe implementation detail as that's more the domain of the OIDC IdP. I think it would be sufficient to say 
```suggestion
    // When this is configured here and in OP, when the user goes to the logout endpoint they will be directed to a page OP's end_session endpoint which
    // will attempt to log out each logged in app/client.
```

docs/content/guides/security/auth/extauth/oauth/_index.md
```diff
@@ -170,6 +170,37 @@ You can also override the revocation endpoint through the [DiscoveryOverride fie
 If the authorization server has a service error, Gloo logs out the user, but does not retry revoking the access token. Check the logs and your identity provider for errors, and manually revoke the access token.
 {{% /notice %}}
 
+### Front Channel Logout
+
+Gloo also supports Front Channel Logout [RFC](https://openid.net/specs/openid-connect-frontchannel-1_0-final.html). Front channel logout is used when the OIDC provider is handling multiple logged in sessions and the logout endpoint should log out of all of them. With front channel logout, when the user goes to the `logoutPath` endpoint, the browser is redirected to an iframe that calls the `frontChannelLogout.Path` endpoint for each session before redirecting to the appropriate after-logout URL. The `frontChannelLogout.Path` must be registered with the OIDC provider.
```
jbohanon, line 175: similarly I don't think we need to describe the details of how OIDC IdPs handle front channel logout, but I'm less opposed to it here for explanatory purposes.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
jbohanon: Oooh thanks for the clarification

jbohanon (approved):

PR #9959, opened 2024-08-27
Improved Virtual Service TLS validation

nfuden:
# Description

SSL configurations have several pieces in envoy. This includes things like the basic cert and key as well as pieces like tls directionality, validation contexts and the like. Some of these options are everywhere and some are more tightly scoped.

In particular we have bits of validation in several places with the most of it living on our ResolveCommonSSLConfig in the projects/gloo/pkg/utils/ssl file which covers most upstream/downstream configuration and validation. This does not include configuration on the virtualservice for [server tls](https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls).

One of the few places that previously did not go through this piece was the virtual service sslconfig. We have a decent bit of validation there for things like domain, and overlap but today it doesn't validate certificate veracity. 
Most of the time users will configure ssl via secret refs which will generally have some level of protection against bad certificates but cannot fully be relied on.

## Code changes

1. Expose some ssl util functions for reuse
2. Reject invalid certs on virtualservices which previously would have been nacked by envoy. This is performed by reused to key + cert validation logic found elsewhere. Does not help in sds case.

# Context
Users were able to get envoy to nack configuration via malformed certificates on virtualservices.

This is reproducible by performing the following:
1. Install gloo with validation enabled and alwaysaccept=false
2. Setup https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls/
3. retrieve the crt from the tls secret. base64 decode it
4. Concatenate the original crt with itself and remove the --- End Certificate --- line
5. Reencode the string and set it as the crt in the secret

Old Behavior: Gloo thinks its good but glooctl check shows that its out of sync
New Behavior its rejected

## Interesting decisions
Some of the comonssl configuration is not useful for virtualservice use case so we only reuse a portion of the code.

## Notes for reviewers
As part of this we had several other checks on ssl configuration that we ultimately decided did not need updates and so tried not to add too much extra outside tests as it would add more lines to the diff.

# Checklist:


projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -334,6 +341,44 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
+// skip if you dont have any ssl config
+func validateSSLConfiguration(ctx context.Context, vs *v1.VirtualService,
+	snapshot *gloosnapshot.ApiSnapshot, reports reporter.ResourceReports) {
+
+	sslConfig := vs.GetSslConfig()
+	if sslConfig == nil {
+		return
+	}
+
+	if sslFiles := sslConfig.GetSslFiles(); sslFiles != nil {
```
sheidkamp, line 354: Had to go back to the proto definition to understand this "if else". Might be helpful to have a comment reminding readers that ssl has one of secrets/ssl files/or sds.

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -334,6 +341,44 @@ func VirtualHostName(vs *v1.VirtualService) string {
...
+
+			if settingsutil.MaybeFromContext(ctx).GetGateway().GetValidation().GetWarnMissingTlsSecret().GetValue() &&
+				errors.Is(err, utils.SslSecretNotFoundError) {
+				reports.AddWarning(vs, err.Error())
+			} else {
+				reports.AddError(vs, utils.InvalidTlsSecretError(secretRef, err))
+			}
+		}
+	} else if sslConfig.GetSslSecrets() != nil && sslConfig.GetSds() == nil {
+		contextutils.LoggerFrom(ctx).DPanic(vs.GetDisplayName() + "has unvalidated ssl secret")
```
sheidkamp, line 377: Why is this a panic when the others add to the reports?

projects/gloo/pkg/utils/ssl.go
```diff
@@ -413,7 +415,10 @@ func (s *sslConfigTranslator) ResolveCommonSslConfig(cs CertSource, secrets v1.S
 	return tlsContext, err
 }
 
-func getSslSecrets(ref core.ResourceRef, secrets v1.SecretList) (string, string, string, []byte, error) {
+// GetSslSecrets retrieves the given secret from the snapshot secretlist
+// it returns the cert chain, private key, and root CA from the given secret.
+// If the secret is not found or invalid then it returns
+func GetSslSecrets(ref core.ResourceRef, secrets v1.SecretList) (string, string, string, []byte, error) {
```
sheidkamp, line 421: nit: public facing functions should probably have tests, although it looks like we get coverage from tests of ResolveCommonSslConfig

sheidkamp (changes requested):
Would be great to see more/detailed comments in the validateSSLConfiguration function

projects/gateway/pkg/translator/virtual_service_translator.go
nfuden: This is if its an unhandled version. Today in utils we also handle sds but do no additional validation.
Therefore we should catch this in dev via the dpanic.

projects/gloo/pkg/utils/ssl.go
nfuden: Coverage currently shows 100 percent locally

projects/gateway/pkg/translator/virtual_service_translator.go
nfuden: added comment showing the proto comment

projects/gateway/pkg/translator/virtual_service_translator.go
```diff
@@ -334,6 +341,44 @@ func VirtualHostName(vs *v1.VirtualService) string {
 	return fmt.Sprintf("%v.%v", vs.GetMetadata().GetNamespace(), vs.GetMetadata().GetName())
 }
 
+// validate that the ssl configuration is mostly ok
```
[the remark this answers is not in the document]
nfuden: updated wdyt now

PR #9966, opened 2024-08-28
Increased timeouts for proxies check in glooctl

tkukushkin:
# Description

- Added more attempts to make `ProxyEndpointRequest` in `glooctl check` when checking proxies.
- Increased timeout on getting metrics from proxies in `glooctl check` when checking proxies.

# Context

In our environment we run `glooctl check` every 5 minutes and pretty often we get errors with connection problems like:

``` 
* rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:51350: connect: connection refused"
...
* timed out trying to connect to localhost during port-forward, errors: 8 errors occurred:
* Get "http://localhost:51666/stats/prometheus": dial tcp [::1]:51666: connect: connection refused
...
```

About first error. Debugging `glooctl check` locally I found out that sometimes port-forward starts to really work after more than 1 second. But request to gloo is made right after starting port-forward. It has 5 retries, and sometimes it is not enough for our environment. And as `glooctl check` makes this request for all watched namespaces (we have many) and port-forward is created on every request, chance of problems increases.

About second error. We get it not so often, maybe once a day, but it still annoys. We have around 240_000 metrics on our proxies and probably 30 seconds timeout is not always enough.

## Testing steps

I manually tested `glooctl check` in my environmnent. Everything works fine.

# Checklist:


sheidkamp:
@tkukushkin - we recently made some similar changes that addressed the `proxy.go` issue: https://github.com/solo-io/gloo/pull/10017

If you pull in main, keep only the pkg/cliutil/uri.go timeout change, and add a changelog, we can get this that timeout increase in.

Thanks!

tkukushkin:
@sheidkamp Hello! Do I need to create issue for this PR to add it to changelog?

sheidkamp:
> @sheidkamp Hello! Do I need to create issue for this PR to add it to changelog?

That would be great! A changelog is needed to pass CI. You can use a type of `NON_USER_FACING`, something like https://github.com/solo-io/gloo/blob/TestConfigureNotAttachedHttpListenerOptions-fix/changelog/v1.18.0-beta21/doc-fixes.yaml . 

It should go in the `changelog/v1.18.0-beta22` directory as that is the next release

tkukushkin:
Thanks! Done.

sheidkamp:
/skip-ci build bot doesn't want to run because `Step #0 - "publish-artifacts": Already have image (with digest): gcr.io/cloud-builders/gcloud` (https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=ab4bf19c-e8ae-4410-9ecc-d501ab13bb79)

sheidkamp (approved):

andy-fong (approved):

Issue #8728, opened 2023-09-28
Export allowMissing field for Gloo Edge


edubonifs:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

1.15.x

### Is your feature request related to a problem? Please describe.

Hi, 

We have a customer that wants their clients to navigate through the same Upstreams given that the are authenticated or not, authentication is optional.

If users are authenticated, they would like to perform JWT validation, but if they are not authenticated, this JWT validation shouldn't be done.

The only possible way that I have found of doing this in Gloo Edge is having allowMissing field exported.

As it has been exposed on Gloo Gateway I suppose it would be more or less the same procedure.

### Describe alternatives you've considered

A possible workaround for this would be duplicating every matcher in our RouteTables, checking if the header "Cookie" is present, if it is present continue with JWT Validation, if it is not present, don't do JWT validation.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206768562311555/1208137921423962) by [Unito](https://www.unito.io)

nfuden:
What does https://docs.solo.io/gloo-edge/latest/reference/api/github.com/solo-io/gloo/projects/gloo/api/v1/enterprise/options/jwt/jwt.proto.sk/#vhostextension not cover for this? 
is it some combo check thing that we are looking for?

huzlak:
@nfuden There's only allowMissingOrFailed where we need to not allow failed, only allow missing. Something like a [validation policy](https://docs.solo.io/gloo-mesh-gateway/latest/reference/api/jwt_policy/#jwtpolicyspec-config-validationpolicy) in GMG, or expose allowMissing option next to already available allowMissingOrFailed.

PR #9985, opened 2024-09-04
Separate `allow_missing` option for jwt

npolshakova:
# Description

Expose `allow_missing` field: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-field-extensions-filters-http-jwt-authn-v3-jwtrequirement-allow-missing

This allows the use case where if users are authenticated, they will be able to perform JWT validation, but if they are not authenticated, this JWT validation won't be done.

## API changes
- Added `allow_missing` field to JWT config on VirtualHost/RouteOptions resource (and VirtualService)

## Code changes
- Only API changes in this PR

## CI changes
NONE

## Docs changes
- New `allow_missing` enum is now supported on new validationPolicy field. This is inline with how the Gloo Platform JWTPolicy sets allow_missing/allow_missing_or_failed

# Context

https://github.com/solo-io/gloo/issues/8728
If users are authenticated, they would like to perform JWT validation, but if they are not authenticated, this JWT validation shouldn't be done.

## Interesting decisions
None

## Testing steps
See solo-projects PR: https://github.com/solo-io/solo-projects/pull/6882

## Notes for reviewers
NONE

# Checklist:


projects/gloo/api/v1/enterprise/options/jwt/jwt.proto
```diff
@@ -49,6 +49,10 @@ message VhostExtension {
     // If this is false (default false), requests that fail JWT authentication will fail authorization immediately.
     // For example, if a request requires either JWT auth OR another auth method, this can be enabled to allow a failed JWT auth request to pass through to the other auth method.
     bool allow_missing_or_failed_jwt = 2;
```
timflannagan, line 51: Are we planning on deprecating this field?

projects/gloo/api/v1/enterprise/options/jwt/jwt.proto
```diff
@@ -49,6 +49,10 @@ message VhostExtension {
...
+
+    // Allow pass through of JWT requests for this virtual host, even if JWT token is missing.
+    // If this is false (default false), requests that are missing a JWT token will fail authorization immediately.
+    bool allow_missing_jwt = 3;
```
timflannagan, line 55: We tried to avoid boolean fields as much as possible in GMG's API surface over time. WYDT about an enum option that provides more us with more extensibility over time? Ex: https://docs.solo.io/gloo-mesh-gateway/latest/reference/api/jwt_policy/#jwtpolicyspec-config-validationpolicy.

projects/gloo/api/v1/enterprise/options/jwt/jwt.proto
npolshakova: Nope, I thought we wanted to at first but envoy supports `allow_missing_or_failed_jwt` and `allow_missing_jwt`. There's no separate `allow_failed` field.

projects/gloo/api/v1/enterprise/options/jwt/jwt.proto
npolshakova: The only issue is it's not going to be the same type as `bool allow_missing_or_failed_jwt = 2;`, but that shouldn't have any user facing consequences.

npolshakova: Ah, I see envoy actually has them as a one_of... https://github.com/aryan16/envoy/blob/226d37da07b1f015827b44f1dc093ec81d71b063/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto#L535 I think then we should move to the enum, and mark `allow_missing_or_failed_jwt` as deprecated, then remove it in 1.19.

timflannagan: > which we can't deprecate in this release

Why not? Anything preventing us from adding that allow_missing_or_failed as an enum option?

timflannagan: @npolshakova Another thing to consider here is that kevin had some opinions about whether should even be supporting allow_missing functionality. I can add you to a tmp channel where he outlined his concerns if that context is helpful at all.

npolshakova: We can mark it as deprecated, but the field will still need to be there for another release until we can remove it completely.

timflannagan: Yup, totally agree that we can't deprecate and remove in the same release. I just meant we should start the deprecation process as soon as possible and track the warn/need to remove this functionality in future releases so it doesn't slip through the cracks.

timflannagan (approved):

sam-heilbron (approved):

PR #10010, opened 2024-09-10
Add route name to output envoy config

npolshakova:
# Description

RouteOptions JWT PR added unique name for HTTPRoute routes: https://github.com/solo-io/solo-projects/pull/6788/files#diff-e63ec6ccf19f856877aa52c64a5797245fa30a435e5803180875e5f19ca4cb56 

## API changes
None

## Code changes
None, only changed translator envoy expected output

## CI changes
Checked the error code is now correct by adding a `Fail()` to the gateway tests:
```
❯ TEST_PKG="./projects/gateway2/..." make go-test-with-coverage
FAIL
make: *** [Makefile:309: go-test] Error 1

```
Then check code:
```
❯ echo $?                                                       
2

``` 
https://github.com/solo-io/gloo/actions/runs/10816368488/job/30007394716?pr=10010

## Docs changes
NONE

# Context
https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1725975132629659

# Checklist:


npolshakova:
/kick build bot failed to start: https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=5e3023ae-8e7a-484a-a4cb-469102df0026

timflannagan (approved):

projects/gateway2/translator/testutils/outputs/route-sort.yaml
```diff
@@ -15,6 +15,7 @@ listeners:
           - matchers:
             - exact: /a/1/2
             options: {}
+            name: route-1-infra-0
```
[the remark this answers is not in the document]
npolshakova: This is the config that's generating it:
```
kind: HTTPRoute
metadata:
  name: route-1
  namespace: infra
spec:
  parentRefs:
  - name: example-gateway
  hostnames:
  - "example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /a
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: /a/1
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: RegularExpression
        value: /a/.*
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: RegularExpression
        value: /b/c/.*
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: Exact
        value: /a/1
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: Exact
        value: /a/1/2
    backendRefs:
    - name: example-svc
      port: 80
---
kind: HTTPRoute
metadata:
  name: route-2
  namespace: infra
spec:
  parentRefs:
  - name: example-gateway
  hostnames:
  - "example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /x
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: /x/1
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: RegularExpression
        value: /x/.*
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: RegularExpression
        value: /x/y/.*
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: Exact
        value: /x/1
    backendRefs:
    - name: example-svc
      port: 80
  - matches:
    - path:
        type: Exact
        value: /x/1/2
    backendRefs:
    - name: example-svc
      port: 80
```

The rule to create the unique name is: `name-namespace-id`. So in this case the route is `route-1`, namespace is `infra`, and this is the first route:
```
  - matches:
    - path:
        type: PathPrefix
        value: /a
    backendRefs:
    - name: example-svc
      port: 80
```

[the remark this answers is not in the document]
npolshakova: Hm... but I see what you mean. This one should be id=0:
```
          - matchers:
            - exact: /a/1/2
            options: {}
            name: route-1-infra-0
            routeAction:
              single:
                kube:
                  port: 80
                  ref:
                    name: example-svc
                    namespace: infra
```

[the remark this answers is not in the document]
npolshakova: Ah, I think the rule also needs to be included in the name...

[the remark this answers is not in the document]
npolshakova: Okay updated this in https://github.com/solo-io/gloo/pull/10010/commits/180495c91d68a92f2963719de030b9fe087fb0c2 but might be good to have a discussion offline if that's the best way of handling the uniqueness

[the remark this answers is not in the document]
timflannagan: Totally agree. At a glance, it seems reasonable but I'd recommend starting a thread to get thoughts from other folks.

projects/gateway2/query/httproute.go
```diff
@@ -83,8 +83,8 @@ func (hr *HTTPRouteInfo) Clone() *HTTPRouteInfo {
 }
 
 // UniqueRouteName returns a unique name for the route, based on the index of the route in the parent.
-func (hr *HTTPRouteInfo) UniqueRouteName(index int) string {
-	return fmt.Sprintf("%s-%s-%d", hr.GetName(), hr.GetNamespace(), index)
+func (hr *HTTPRouteInfo) UniqueRouteName(index int, ruleHash uint64) string {
+	return fmt.Sprintf("%s-%s-%d-%d", hr.GetName(), hr.GetNamespace(), ruleHash, index)
```
EItanya, line 87: Is there a length constraint we need to worry about here?

EItanya (changes requested):

projects/gateway2/query/httproute.go
npolshakova: So we can either use the hash here or the index from `for ruleIdx, rule := range route.Spec.Rules`:
```
Alternative approaches:
1. name-namespace-ruleHash-index
2. name-namespace-ruleIndex-matchesIndex
```

The[ route name](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route.proto#config-route-v3-routeconfiguration) is just a string in envoy, so the limit is just the proto limit:
```
A string must always contain UTF-8 encoded[5] or 7-bit ASCII text, and cannot be longer than 232.
```

So we could run into the limit if the name + namespace is extremely long that appending the hash + index puts it over the edge. I can add a check here and error if it's too long. The `ruleIndex-matchIndex` is a little shorter, but would still potentially run into this issue.

EItanya: My bad., I don't see any length constraint [here](https://github.com/envoyproxy/envoy/blob/7b8baff1758f0a584dcc3cb657b5032000bcb3d7/api/envoy/config/route/v3/route_components.proto#L257C1-L258C1) so I think we're good to go.

EItanya (approved):

projects/gateway2/query/httproute.go
```diff
@@ -83,8 +83,8 @@ func (hr *HTTPRouteInfo) Clone() *HTTPRouteInfo {
 }
 
 // UniqueRouteName returns a unique name for the route, based on the index of the route in the parent.
-func (hr *HTTPRouteInfo) UniqueRouteName(index int) string {
-	return fmt.Sprintf("%s-%s-%d", hr.GetName(), hr.GetNamespace(), index)
+func (hr *HTTPRouteInfo) UniqueRouteName(index int, ruleHash uint64) string {
```
lgadban, line 86: godoc comment should be updated explaining what ruleHash is

projects/gateway2/query/httproute.go
lgadban: The rule index makes it actually parse-able by a human -- is there a benefit to using the rulehash?

projects/gateway2/query/httproute.go
npolshakova: Updated comment to describe rule index and match index

sam-heilbron (approved):

Issue #9656, opened 2024-06-20
Expose max_buffered_{rx/tx}_bytes in the Tap filter


sadieleob:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

1.17

### Is your feature request related to a problem? Please describe.

This is not related to an issue. 

### Describe the solution you'd like

The customer is asking to expose:

- **max_buffered_rx_bytes**
- **max_buffered_tx_bytes** 
- **record_headers_received_time**
- **record_downstream_connection**

The customer stated that without the max_buffered_<rx/tx>_bytes fields set, the maximum body size is 1KiB  that can be sent to the tap filter. Unfortunately, this max body size blocks our integration of the tap filter with their "in-house monetization" project since it requires payloads up to 4KB.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/tap/v3/common.proto#config-tap-v3-outputconfig
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/tap/v3/tap.proto#extensions-filters-http-tap-v3-tap

### Describe alternatives you've considered

There are no alternatives

# Related Issues

- https://github.com/solo-io/solo-projects/issues/5455
- https://github.com/solo-io/gloo/issues/9845

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206768562311555/1208137702093564) by [Unito](https://www.unito.io)

nfuden:
https://github.com/solo-io/solo-projects/issues/5455 encompasses this as well. @DuncanDoyle

DuncanDoyle:
@nfuden As discussed, removing the "Streaming support" from this requirement an putting this in a separate issue: https://github.com/solo-io/gloo/issues/9845

ashishb-solo:
The changes in gloo only expose the protobuf changes, but the translation logic lives in solo-projects and that has not yet merged. Reopening.

nfuden:
This also requires a backport to 1.17 and notes on the enterprise releases that it will be released in

ashishb-solo:
This is a duplicate of https://github.com/solo-io/solo-projects/issues/5455

DuncanDoyle:
Released in Gloo Gateway Enterprise 1.17.2

PR #10016, opened 2024-09-11
Expose max_buffered_{rx/tx}_bytes in the Tap filter (#9656)

andy-fong:
# Description

Expose new fields in Tap filter Config

## API changes

- added maxBufferedRxBytes, maxBufferedTxBytes, recordHeadersReceivedTime, recordDownstreamConnection to tap.proto

## Docs changes

Auto-generated doc and crds are updated

# Context

## Testing steps

Added new fields to the protobuf file only. This feature is in enterprise version only. So, no testing steps here. 

## Notes for reviewers

# Checklist:


sam-heilbron (approved):
LGTM!

npolshakova (approved):

ashishb-solo (approved):
Looks good, thanks! Apologies for taking an extra moment to review, but I wanted to confirm whether this statement was correct:

> If truncation occurs, the truncated field will be set.

It does appear that this is set outside the tap sink (for which we have our own custom implementation), so it should still get set in the output message. But I just wanted to double-check to make sure that this was the case.

PR #10025, opened 2024-09-12
1.17/fix/persistproxyspec

nfuden:
# Description
In 1.17 we introduced a regression for instances using persist proxy spec.

Persist proxy spec is mostly non-useful and will be removed in 1.18 but does have some smaller edge cases still supported in 1.17.
Unlike most of our resources proxies were called at a stage where we can accept a resource into our gloo system but fail to write any statuses back out along with failing to update the proxy CR at rest in etcd.

## Code changes
Label selection cleanup

## CI changes
Updated Upgrade tests to check for this step

# Context
On upgrading from a version 1.13+ to 1.17+ with persist proxy spec set to true manual intervention was required prior to this wrk.

## Testing steps

Thanks to sam https://github.com/solo-io/gloo-gateway-shared-resources/tree/main/issues/gloo/9968

Original reproduction was hampered by overly aggressive clean up scripts scheduled by default on local so this simplest set up helps alot.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #10033, opened 2024-09-13
test/kube2e/upgrade: overhaul to actually do validation and make persistproxyspec not break

nfuden:
# Description
In 1.17 we introduced a regression for instances using persist proxy spec.

Persist proxy spec is mostly non-useful and will be removed in 1.18 but does have some smaller edge cases still supported in 1.17.
Unlike most of our resources proxies were called at a stage where we can accept a resource into our gloo system but fail to write any statuses back out along with failing to update the proxy CR at rest in etcd.

## Code changes
Label selection cleanup

## CI changes
Updated Upgrade tests to check for this step

# Context
On upgrading from a version 1.13+ to 1.17+ with persist proxy spec set to true manual intervention was required prior to this wrk.

## Testing steps

Thanks to sam https://github.com/solo-io/gloo-gateway-shared-resources/tree/main/issues/gloo/9968

Original reproduction was hampered by overly aggressive clean up scripts scheduled by default on local so this simplest set up helps alot.

# Checklist:


nfuden:
see https://github.com/solo-io/gloo/pull/10025#issuecomment-2349892751

sam-heilbron (approved):

jenshu (approved):

sam-heilbron (approved):

PR #10047, opened 2024-09-16
projects/utils/ssl: Add a check for ssl secrets that will be rejected by envoy but not go

nfuden:
# Description

In go pem decoding is intentionally spec compliant. This means that it is very permissive with its decoding and will skip un-parsable, possibly usable or just commented data.

This means that when we consume crt data we may pass along info that is rejected by envoy which can stall configuration and is generally something we try to avoid.

## Code changes

Rely on kubernetes cert encoding to check to see if anything was dropped in decoding from pem and then emit an error if anything was dropped.

## Interesting decisions

Originally we attempted to go the route as seen in https://gist.github.com/anitgandhi/58b0618512fdb3caa89e86c8a6a536ab  per https://github.com/golang/go/issues/34069 
While this mostly worked and caught several issues we continued to find more edge cases such as with headers. Given this we decided to lean heavily on kubernetes setup to make sure that we are most compliant with what another big project does.

For example the gist catches the unterminated cert block but did not catch invalid headers being provided for example 
`-----BEGIN HEADERS-----
Header: 1
-----END HEADERS-----` 
is invalid as it does not have a newline like
`-----BEGIN HEADERS-----
Header: 1

-----END HEADERS-----`

## Testing steps

Reproduction:
Follow https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls/
Now update the crt by adding 
`-----BEGIN CERTIFICATE-----
Then update your secret

` kubectl create secret tls animal-certs  --key tls.key \
    --cert tls.crt --namespace gloo-system \
--save-config \
--dry-run=client \
-o yaml | \
kubectl apply -f -`

Next update something to kick translation ideally a route on the accepted vs.

See that envoy nacks the request.

Remove the secret reference. 
Upgrade to this pr's image
Readd the reference and see validation block the update.

# Checklist:


projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -45,5 +47,19 @@ func isValidSslKeyPair(certChain, privateKey, rootCa []byte) error {
 	}
 
 	_, err := tls.X509KeyPair([]byte(certChain), []byte(privateKey))
+	if err != nil {
+		return err
+	}
+	candidateCert, err := cert.ParseCertsPEM([]byte(certChain))
+	if err != nil {
+		// return err rather than overriding certchain to maintain ux with kubernetes gateway
```
sam-heilbron, line 55: Can you help me understand this comment? Perhaps either link to the place in code where this behavior is defined or to the docs where we state this?

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -45,5 +47,19 @@ func isValidSslKeyPair(certChain, privateKey, rootCa []byte) error {
...
+		return err
+	}
+	reencoded, err := cert.EncodeCertificates(candidateCert...)
```
sam-heilbron, line 58: nit: It sounded like during your investigation, there was a difference between certificates validity according to go and envoy.,..Should we add a comment clarifying what this doesn't protect against?

projects/gloo/pkg/utils/ssl.go
```diff
@@ -443,7 +445,28 @@ func isValidSslKeyPair(certChain, privateKey, rootCa string) error {
...
+		return err
+	}
+
+	// validate that the parsed piece is valid
+	// this is still faster than a call out to openssl despite this second parsing pass of the cert
+	// pem parsing in go is permissive while envoy is not
+	// this might not be needed once we have larger envoy validation
+	candidateCert, err := cert.ParseCertsPEM([]byte(certChain))
+	if err != nil {
+		// return err rather than overriding certchain to maintain ux with kubernetes gateway
```
sam-heilbron, line 461: Should we just have a shared utility that both edge gw and kube gw invoke, so that we are guaranteed they stay in sync?

sam-heilbron (approved):
LGTM!

projects/gateway2/translator/sslutils/ssl_utils.go
nfuden: This fixes it. Namely go ignores things encode certs and unmarshal elides all the stuff go jumps over

projects/gloo/pkg/utils/ssl.go
nfuden: For some reason the implementors copied the name but not the functionality. Therefore I think this falls into the general audit needed to reduce deduplication.

As I would rather get this out earlier I would rather not re-litigate their choices in this pr

projects/gateway2/translator/sslutils/ssl_utils.go
nfuden: I dont quite understand whats being asked here?

changelog/v1.18.0-beta22/tls-secret-validation.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6772
+    resolvesIssue: false
+    description: >-
+      Plugs a gap where go would check a secret for validity per spec but envoy is more aggressive.
+      For example a tls secret with a certchain that contains an invalid pem block will be rejected by envoy but not go.
+      Prior to this pr these types of secrets would be accepted by gloo and nacked by envoy.
```
bewebi, line 8: nit: I'm not sure if we have specific team standards around this, but since it's customer facing we should make sure changelog messages have correct capitalization for acronyms and proper nouns

```suggestion
      Plugs a gap where go would check a secret for validity per spec but Envoy is more aggressive.
      For example a TLS secret with a certChain that contains an invalid PEM block will be rejected by Envoy but not Go.
      Prior to this PR these types of secrets would be accepted by Gloo and nacked by Envoy.
```

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -45,5 +47,20 @@ func isValidSslKeyPair(certChain, privateKey, rootCa []byte) error {
 	}
 
 	_, err := tls.X509KeyPair([]byte(certChain), []byte(privateKey))
+	if err != nil {
+		return err
+	}
+	candidateCert, err := cert.ParseCertsPEM([]byte(certChain))
```
bewebi, line 53: nit
```suggestion
	_, err := tls.X509KeyPair(certChain, privateKey)
	if err != nil {
		return err
	}
	candidateCert, err := cert.ParseCertsPEM(certChain)
```

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -45,5 +47,20 @@ func isValidSslKeyPair(certChain, privateKey, rootCa []byte) error {
...
+	if err != nil {
+		// return err rather than sanitize. This is to maintain UX with older versions and to prevent a large refactor
+		// for gatewayv2 secret validation code.
```
bewebi, line 56: If I understand correctly from the linked gist, by "sanitize" we mean "attempt to extract valid blocks even if the certChain is not entirely valid"

I think we can improve the comment here
```suggestion
		// return err rather than attempt to extract and return valid certs from the certChain.
		// while doing so may be desirable in some circumstances, returning when any cert is invalid maintains consistent
		// UX with older versions and to prevent a large refactor for gatewayv2 secret validation code.
```

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -45,5 +47,20 @@ func isValidSslKeyPair(certChain, privateKey, rootCa []byte) error {
...
+		return err
+	}
+	reencoded, err := cert.EncodeCertificates(candidateCert...)
```
bewebi, line 59: We'll eventually return this error if it's non-nil, but first we return our own error if the cert doesn't match expectations

I would expect that we should actually return the error right away, but if not can we justify the behavior with a comment

bewebi (commented):
Some comments around comments

projects/gateway2/translator/sslutils/ssl_utils.go
nfuden: dont love changing other peopes code without thought but I cant see why not 🚢

projects/gateway2/translator/sslutils/ssl_utils.go
nfuden: this isnt locked to cets and we do have a comment that states what does or doesnt work.

This comment is purely around why we dont update our certchain to being good rather than just error out.

projects/gateway2/translator/sslutils/ssl_utils.go
bewebi: Seemed like it was likely copied from [here](https://github.com/solo-io/gloo/blob/de652d6a3a83350f885c1b1c55d722a9860818d1/projects/gloo/pkg/utils/ssl.go#L449) in ssl.go where the casting is actually needed
Only suggested the change since the pattern was persisted in new code

projects/gloo/pkg/utils/ssl.go
```diff
@@ -439,6 +439,9 @@ func getSslSecrets(ref core.ResourceRef, secrets v1.SecretList) (string, string,
 	return certChain, privateKey, rootCa, ocspStaple, nil
 }
 
+// isValidSslKeyPair validates that the cert and key are a valid pair
+// It previously only chekced in go but now also checks that nothing is lost in cert encoding
```
bewebi, line 443:
```suggestion
// It previously only checked in go but now also checks that nothing is lost in cert encoding
```

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -40,6 +40,9 @@ func validatedCertData(sslSecret *corev1.Secret) error {
 	return nil
 }
 
+// isValidSslKeyPair validates that the cert and key are a valid pair
+// It previously only chekced in go but now also checks that nothing is lost in cert encoding
```
bewebi, line 44:
```suggestion
// It previously only checked in go but now also checks that nothing is lost in cert encoding
```

changelog/v1.18.0-beta22/tls-secret-validation.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6772
+    resolvesIssue: false
+    description: >-
+      Plugs a gap where go would check a secret for validity per spec but Envoy is more aggressive.
+      For example a TLS secret with a certChain that contains an invalid PEM block will be rejected by Envoy but not Go.
+      Prior to this PR these types of secrets would be accepted by Gloo and nacked by Envoy.
+
```
nfuden, line 9:
```suggestion
      Prior to this PR these types of secrets would be accepted by Gloo and nacked by Envoy.
```

bewebi (approved):

PR #10048, opened 2024-09-16
1.17: projects/utils/ssl: Add a check for ssl secrets that will be rejected by envoy but not go

nfuden:
# Description

In go pem decoding is intentionally spec compliant. This means that it is very permissive with its decoding and will skip un-parsable, possibly usable or just commented data.

This means that when we consume crt data we may pass along info that is rejected by envoy which can stall configuration and is generally something we try to avoid.

## Code changes

Rely on kubernetes cert encoding to check to see if anything was dropped in decoding from pem and then emit an error if anything was dropped.

## Interesting decisions

Originally we attempted to go the route as seen in https://gist.github.com/anitgandhi/58b0618512fdb3caa89e86c8a6a536ab  from https://github.com/golang/go/issues/34069 
While this mostly worked and caught several issues we continued to find more edge cases such as with headers. Given this we decided to lean heavily on kubernetes setup to make sure that we are most compliant with what another big project does.

For example the gist catches the unterminated cert block but did not catch invalid headers being provided for example 
`-----BEGIN HEADERS-----
Header: 1
-----END HEADERS-----` 
is invalid as it does not have a newline like
`-----BEGIN HEADERS-----
Header: 1

-----END HEADERS-----`

## Testing steps

Reproduction:
Follow https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls/
Now update the crt by adding 
`-----BEGIN CERTIFICATE-----
Then update your secret

` kubectl create secret tls animal-certs  --key tls.key \
    --cert tls.crt --namespace gloo-system \
--save-config \
--dry-run=client \
-o yaml | \
kubectl apply -f -`

Next update something to kick translation ideally a route on the accepted vs.

See that envoy nacks the request.

Remove the secret reference. 
Upgrade to this pr's image
Readd the reference and see validation block the update.

# Checklist:

### DIFF
--- a/changelog/v1.17.8/strict-tls-validation.yaml
+++ b/changelog/v1.17.8/strict-tls-validation.yaml
@@ -1,0 +1,9 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6772
+    resolvesIssue: false
+    description: >-
+      Plugs a gap where go would check a secret for validity per spec but envoy is more aggressive.
+      For example a tls secret with a certchain that contains an invalid pem block will be rejected by envoy but not go.
+      Prior to this pr these types of secrets would be accepted by gloo and nacked by envoy.
+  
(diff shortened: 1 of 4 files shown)


sam-heilbron (approved):
LGTM

PR #10050, opened 2024-09-16
admindcli helper and test framework improvement (#9656)

andy-fong:
# Description

While working on the test for https://github.com/solo-io/solo-projects/issues/5455, found an issue with GetSingleListenerFromDynamicListeners() in admincli where it would crash if the config_dump returns an empty array. 

Also, when running focused tests, the docker containers is left behind. 

## Code changes

- return error in GetSingleListenerFromDynamicListeners() if the config array is empty
- check docker container no longer exists before finishing AfterEach()

## Docs changes

# Context

## Notes for reviewers

# Checklist:


pkg/utils/envoyutils/admincli/client.go
```diff
@@ -239,8 +239,13 @@ func (c *Client) GetSingleListenerFromDynamicListeners(
 		return nil, fmt.Errorf("could not get envoy config_dump from adminClient: %w", err)
 	}
 
+	configs := cfgDump.GetConfigs()
+	if len(configs) == 0 {
+		return nil, fmt.Errorf("could not get config: config is empty")
```
nfuden, line 244: can we add a comment as to where we can get a valid configdump but have no configdump?

test/e2e/test_context.go
```diff
@@ -141,6 +141,9 @@ func (c *TestContext) AfterEach() {
 	ginkgo.By("TestContext.AfterEach: Cancelling test context")
 	// All services connected to the TestContext are tied to the context, so cancelling it will clean those up
 	c.cancel()
+	// This line prevents the process to quit before the envoy docker process is killed when running
+	// Focused Tests
```
nfuden, line 145: how does this work with the environment variable for no tear own? can we show how and why this is safe?

sam-heilbron (approved):

pkg/utils/envoyutils/admincli/client.go
andy-fong: This is the case where I dump the config too soon and there is no dynamic_listener setup yet. Also, with the regex match, if no listener matches, it returns empy object: 
```
Solo-System-AFong:solo-projects andyfong$ curl -v "localhost:20001/config_dump?resource=dynamic_listeners&name_regex=foo"
* Host localhost:20001 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:20001...
* Connected to localhost (::1) port 20001
> GET /config_dump?resource=dynamic_listeners&name_regex=foo HTTP/1.1
> Host: localhost:20001
> User-Agent: curl/8.9.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< content-type: application/json
< cache-control: no-cache, max-age=0
< x-content-type-options: nosniff
< date: Tue, 17 Sep 2024 00:20:48 GMT
< server: envoy
< transfer-encoding: chunked
< 
{}```

I will add a comment for both case.

test/e2e/test_context.go
andy-fong: @nfuden I was not aware of a TEAR_DOWN env. So, thank you for bringing that up. Are you talking about this: https://github.com/solo-io/gloo/blob/bddae6d73f635c55fee089b36fd9db7b3be9fdda/test/testutils/env.go#L10 ?
Looks like that only get used for kube2e test to tear down Gloo in k8s. In e2e test, gloo is in memory or inside the test process itself (I think).

nfuden (approved):

PR #10061, opened 2024-09-17
Set GOMEMLIMIT / GOMAXPROCS [1.18]

jenshu:
# Description

Set the GOMAXPROCS and GOMEMLIMIT env variables in the `gloo` and `discovery` pods when resource limits are set. These will hint to Go how much CPU and memory it can use.

# Checklist:


.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -55,31 +55,36 @@ jobs:
...
           go-test-args: '-v -timeout=25m'
           go-test-run-regex: '(^TestK8sGatewayIstio$$|^TestGlooGatewayEdgeGateway$$|^TestGlooctlIstioInjectEdgeApiGateway$$|^TestK8sGatewayNoValidation$$)'
 
-        # September 16, 2024: 17 minutes
+        # September 18, 2024: 17 minutes
         - cluster-name: 'cluster-five'
           go-test-args: '-v -timeout=25m'
           go-test-run-regex: '^TestValidationStrict$$|^TestValidationAlwaysAccept$$|^TestTransformationValidationDisabled$$|^TestGloomtlsGatewayEdgeGateway$$|^TestHelm$$'
 
+        # September 18, 2024: x minutes
```
sam-heilbron, line 83: Is there a reason we believe that tests would behave differently with and without resource limits set? I would have expected that we would just run all of our tests with resource limits defined as it is the recommended install pattern.

jenshu: Unless I'm missing it, I don't think we explicitly recommend using resource limits in production; https://docs.solo.io/gloo-edge/latest/operations/production_deployment/ only mentions resource requests. Also none of our existing e2e tests had resource limits defined in the helm manifests and those are supposed to mimic prod recommendations.

So as not to affect the existing tests, I decided to do a separate suite that tests some of the basic functionality from both k8s gw and edge, but with resource limits. One caveat of using resource limits is that it needs to be tuned for the specific environment you're running in, and can cause OOM errors if set too low. I'm open to just moving the resource limits into the existing suites instead of adding new ones if that's preferred!

EItanya: Running with `k8s` limits set WITHOUT `go` limits set can lead to `go` improperly scheduling as `go` will use the machine CPU numbers rather than those available to the container. I have seen significant gains setting this properly when the container is running on a large shared node.

jenshu: ok it seems we don't currently run all the TestK8sGateway tests on PRs. the new suite i added tries to run all tests, and is failing on a test that is normally only run on nightlies. rather than spending time debugging that test now, i think i'll just do as sam suggested above and add the resource limits to the existing suites, instead of creating the new suite that runs everything

jenshu:
s-p PR that pulls this in: https://github.com/solo-io/solo-projects/pull/6929

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -216,6 +216,20 @@ spec:
                 name: {{ .Values.license_secret_name }}
                 key: license-key
 {{- end }}
+{{- if (((.Values.gloo.deployment.resources).limits).memory) }}
```
sam-heilbron, line 219: nit: Should we include these values in our production recommendations as well?

sam-heilbron (approved):

nfuden (approved):

install/helm/gloo/templates/1-gloo-deployment.yaml
jenshu: good idea, i added a [comment](https://github.com/solo-io/docs/issues/489#issuecomment-2372042811) to an existing docs issue

PR #10064, opened 2024-09-18
validation/api: Expand envoy validate mode API

jbohanon:
# Description

Expand the use of Envoy validation to validate all translation loops. This will significantly affect performance, but will help catch configuration issues which are not (or not yet) caught as errors by Gloo translation before Gloo attempts to serve the bad config to Envoy and gets nacked.

This PR is only for the API changes; functional changes will be implemented in future PRs.

## API changes

- Added gateway.validation.fullEnvoyValidation field to settings resource and to Helm


sam-heilbron (approved):

sheidkamp (commented):
I want to make sure I understand this - we are just adding the API with no functional changes?

Also would like to see a PR description other than the default one, even if its just a single sentence.

sheidkamp (approved):

PR #10072, opened 2024-09-20
Makes MustKindContextWithScheme() Private

danehans:
# Description

MustKindContextWithScheme() was only being called from the cluster pkg so the function has been updated to be private.

## Testing steps

I successfully ran `go test -v -timeout 600s ./test/kubernetes/e2e/tests -run ^TestK8sGateway$/^Deployer$` which calls this function through `CreateTestInstallation()`.

## Notes for reviewers

# Checklist:

### DIFF
--- a/test/kubernetes/testutils/cluster/kind.go
+++ b/test/kubernetes/testutils/cluster/kind.go
@@ -18,11 +18,11 @@ import (
 
 // MustKindContext returns the Context for a KinD cluster with the given name
 func MustKindContext(clusterName string) *Context {
-	return MustKindContextWithScheme(clusterName, kubetestclients.MustClientScheme())
+	return mustKindContextWithScheme(clusterName, kubetestclients.MustClientScheme())
 }
 
-// MustKindContextWithScheme returns the Context for a KinD cluster with the given name and scheme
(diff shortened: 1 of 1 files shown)


test/kubernetes/testutils/cluster/kind.go
```diff
@@ -18,11 +18,11 @@ import (
 
 // MustKindContext returns the Context for a KinD cluster with the given name
 func MustKindContext(clusterName string) *Context {
-	return MustKindContextWithScheme(clusterName, kubetestclients.MustClientScheme())
+	return mustKindContextWithScheme(clusterName, kubetestclients.MustClientScheme())
 }
 
-// MustKindContextWithScheme returns the Context for a KinD cluster with the given name and scheme
-func MustKindContextWithScheme(clusterName string, scheme *runtime.Scheme) *Context {
```
jenshu, line 25: this is called [here](https://github.com/solo-io/solo-projects/blob/e9342a5d6d60ec12cafbddea621d4813a949f70e/test/kubernetes/e2e/test.go#L81) so we can't remove it. it would be good to add a comment here that it's used outside this repo, so we don't accidentally remove it

danehans: Ah... I didn't expect the function to be used outside this project. Thanks for the feedback!

PR #10079, opened 2024-09-20
projects: Disable k8s generation for graphql, extauth, gloosnapshot

timflannagan:
Remove the solo-kit.json options that generate the hack/update-codegen.sh bash scripts. This isn't needed as we don't generate any k8s clients for these APIs / snapshots.

# Description

    projects: Disable k8s generation for graphql, extauth, gloosnapshot
    
    Remove the solo-kit.json options that generate the hack/update-codegen.sh
    bash scripts. This isn't needed as we don't generate any k8s clients for
    these APIs / snapshots.
    
    Signed-off-by: timflannagan <timflannagan@gmail.com>

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/solo-kit.json
```diff
@@ -1,8 +1,6 @@
 {
   "name": "enterprise.gloo.solo.io",
   "version": "v1",
-  "gen_kube_types": true,
-  "gen_kube_validation_schemas": true,
```
sam-heilbron, line 5: I'm a bit surprised we'd want to delete this line as well. Per https://github.com/solo-io/solo-kit/blob/main/pkg/code-generator/model/project.go#L48 this _should_ control whether we build the validation schemas on CRDs for resources in this project, and the AuthConfig does have schemas built during codegen. Any indication why removing this doesn't also remove the schemas?

sam-heilbron (commented):
LGTM! One question about the AuthConfig CRD schema

projects/gloo/api/v1/enterprise/options/extauth/v1/solo-kit.json
timflannagan: Yeah, I think we have a potential gap in our codegen strategy where we don't "clean" these files before re-generating them. In the case of the hack/*-codegen.sh files, I had to manually delete those, re-run the codegen targets, and verify those files weren't recreated to validate we could remove these files.

I'll look into whether the schema generation is still relevant here locally.

timflannagan: @sam-heilbron Looked into this further before hopping off. What's weird is that manually deleting install/helm/gloo/crds/*.yaml and sequentially running `make clean && make install-go-tools generate-all` does not re-generate those CRDs (i.e. those files are still deleted according to git). I didn't end up finding the root cause but that behavior does seem concerning without having the full context on our codegen story.

sam-heilbron: https://github.com/solo-io/gloo/pull/10081 shows what is happening here, and why this change isn't causing errors on your PR. I tried to put context on my PR. Happy to talk more on Monday to see if there are ways we can improve the known gaps/docs while we're here.

timflannagan: Awesome, thanks for digging further into that. Makes more sense now why CI wasn't catching this. I'll plan on reverting the schema toggle changes. I do think this is surprising generation behavior that's worth addressing though. Otherwise we risk accumulating maintenance debt over time.

sam-heilbron: Sounds good! I also would prefer to set these values to false and add a comment clarifying _why_ they're false, rather than just removing and losing the context of why. wdyt?

timflannagan: @sam-heilbron Do we still need this field for the gloosnapshot snap.proto too? Just looking at that file's comment, it seems like the answer is no?

timflannagan:
Sorry for the force push -- got my branch into a weird working state locally.

projects/gloo/api/v1/enterprise/options/extauth/v1/solo-kit.json
sam-heilbron: Correct, I think that we can disable kube schema generation for the gloosnapshot.proto. That's an wrapper api that defines the set of CRs that make up a snapshot

projects/gloo/api/v1/enterprise/options/extauth/v1/solo-kit.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "enterprise.gloo.solo.io",
   "version": "v1",
-  "gen_kube_types": true,
```
sam-heilbron, line 4: nit: I had mentioned this in another comment, but I kind of like when things are explicitly set to false since it shows its intentional, rather than not defined (which another engineer could assume was just "forgotten")

sam-heilbron (approved):

projects/gloo/api/v1/enterprise/options/extauth/v1/solo-kit.json
timflannagan: Not disagreement here. I was just maintaining consistency with the rest of the solo-kit.json files I was looking at for non-kube facing APIs.

nfuden (approved):

PR #10080, opened 2024-09-20
in mesh upstreams

stevenctl:
# Description

To support gloo-gateway as a mesh ingress, we need a mechanism to reference the VirtualDestination. The hosts field of a static upstream marked for mesh will use ServiceEntry (which in this usecase are expected to be generated by VD) to find the endpoints we want to route to. 

## API changes

- add `inMesh` bool marker to Upstream_Static

## Code changes

- Skip handling mesh static upstreams in static upstream plugin


projects/gloo/api/v1/options/static/static.proto
```diff
@@ -33,6 +33,11 @@ message UpstreamSpec {
     // If both this and host.sni_addr are set, host.sni_addr has priority.
     // defaults to "true".
     google.protobuf.BoolValue auto_sni_rewrite = 6;
+
+    // In mesh indicates that hosts references ServiceEntries, possibly the
+    // ServiceEntries derived from VirtualDestination, to route upstream
+    // traffic.
+    bool in_mesh = 7;
```
artberger, line 40: Does this mean if I set to true, that the static hosts have to be ServiceEntries? maybe something like:

```suggestion
    // To use Gloo Gateway as an ingress gateway in a service mesh, set to true.
    // Then, set this static upstream's hosts to reference ServiceEntries,
    // such as ServiceEntries derived from VirtualDestinations,
    // to route upstream traffic to.
    bool in_mesh = 7;
```

josh-pritchard (approved):

stevenctl:
I think maybe we just don't merge it if we're not going to merge an impl for it.

PR #10090, opened 2024-09-23
validation: enhance translation with full envoy validation

jbohanon:
# Description

Implement a full validation of the translated xds snapshot via Envoy Validate mode. This is a new feature enabled in helm or in the Settings CR via `gateway.validation.fullEnvoyValidation`. Enabling this causes discrete transformation and waf validations to become redundant. 

## Code changes

Add functionality to convert xds code into static bootstrap which can be fed into Envoy validate mode. Also add conditional call to this validation in the translation loop after all translation has occurred, and report an error if Envoy validate returns one.

## CI changes

- Added kubernetes e2e test installation suite for new feature

## Testing steps

Testing can be done by following the added kubernetes e2e testing script.

Testing for the pkg/utils/envoyutils/validation package is handled via e2e testing.

Unit tests added for the new bootstrap conversion utils

# Checklist:


pkg/utils/envoyutils/bootstrap/bootstrap.go
```diff
@@ -102,3 +115,220 @@ func FromFilter(filterName string, msg proto.Message) (string, error) {
 
 	return FromEnvoyResources(&EnvoyResources{Listeners: []*envoy_config_listener_v3.Listener{listener}})
 }
+
+// FromSnapshot accepts an xds Snapshot and converts it into valid bootstrap json.
```
sam-heilbron, line 119: For functions that are this large, and with specific use-cases, I think we'd benefit from having a series of unit tests. wdyt?

pkg/utils/envoyutils/validation/validation.go
```diff
@@ -21,3 +24,24 @@ const (
...
 }
+
+// ValidateSnapshot accepts an xDS snapshot, clones it, and does the necessary
+// conversions to imitate the same config being provided as static bootsrap config to
+// Envoy, then executes Envoy in validate mode to ensure the config is valid.
+func ValidateSnapshot(
+	ctx context.Context,
+	snap envoycache.Snapshot,
+) error {
+	// THIS IS CRITICAL SO WE DO NOT INTERFERE WITH THE CONTROL PLANE.
```
sam-heilbron, line 35: nit: To me, this implies 2 things (which may be accurate and good) - (1) that this function should only be called from the control plane. (2) that this function does mutate the snapshot

are both of those true?

pkg/utils/envoyutils/validation/validation.go
```diff
@@ -21,3 +24,24 @@ const (
 func ValidateBootstrap(ctx context.Context, bootstrap string) error {
 	return runner.RunEnvoyValidate(ctx, envutils.GetOrDefault(constants.EnvoyBinaryEnv, defaultEnvoyPath, false), bootstrap)
 }
+
```
sam-heilbron, line 27: ^ In a slack thread the other day, it was raised that we have some plugin functionality that invokes this code. It was proposed that we adjust our plugins to make this injectable, and thus inject a mocked implementation for easier testing. 

Not necessary to do in this PR, just wanted to callout that we may want to consider how we can effectively port/use this code

test/kubernetes/e2e/features/validation/full_envoy_validation/suite.go
```diff
@@ -0,0 +1,51 @@
...
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, validation.ExampleUpstream, "-n", s.testInstallation.Metadata.InstallNamespace)
+	s.Assert().NoError(err)
+	s.testInstallation.Assertions.EventuallyResourceStatusMatchesState(
+		func() (resources.InputResource, error) {
+			return s.testInstallation.ResourceClients.UpstreamClient().Read(s.testInstallation.Metadata.InstallNamespace, "nginx-upstream", clients.ReadOpts{Ctx: s.ctx})
+		},
+		core.Status_Accepted,
+		gloo_defaults.GlooReporter,
+	)
+	// rejects invalid inja template in transformation
```
sam-heilbron, line 46: Should we have a postitive case to mirror this negative case as well?

test/kubernetes/e2e/features/validation/full_envoy_validation/suite.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
+	return &testingSuite{
+		ctx:              ctx,
+		testInstallation: testInst,
+	}
+}
+
+// TestRejectInvalidTransformation checks webhook rejects invalid transformation when fullEnvoyValidation=true
```
sam-heilbron, line 35: This is a great test to prove parity with an existing set of known invalid configuration. Is the idea to find other test cases before merging or is this the intended scope of our e2e testing?

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
         # September 16, 2024: 17 minutes
         - cluster-name: 'cluster-five'
           go-test-args: '-v -timeout=25m'
-          go-test-run-regex: '^TestValidationStrict$$|^TestValidationAlwaysAccept$$|^TestTransformationValidationDisabled$$|^TestGloomtlsGatewayEdgeGateway$$|^TestHelm$$'
+          go-test-run-regex: '^TestFullEnvoyValidation$$|^TestValidationStrict$$|^TestValidationAlwaysAccept$$|^TestTransformationValidationDisabled$$|^TestGloomtlsGatewayEdgeGateway$$|^TestHelm$$'
```
sam-heilbron, line 81: nit: We're adding new tests, so we may want to update the comment clarifying how long these tests take (which are used for future load-balancing purposes)

pkg/utils/envoyutils/bootstrap/bootstrap.go
```diff
@@ -102,3 +115,220 @@ func FromFilter(filterName string, msg proto.Message) (string, error) {
...
+
+	// Next, we will look through our Snapshot's clusters and delete the ones which are
+	// already routed to.
+	processClusters(routedCluster, resources.Clusters, resources.endpoints)
+
+	// We now need to find clusters which do not exist, even though they are targeted by
+	// a route. In static mode, envoy won't start without these. At this point in the
+	// processing, routedClusters holds this list, so we use it to create blackhole
+	// clusters for these routes to target.
+	addBlackholeClusters(routedCluster, resources.Clusters)
```
nfuden, line 148: Can we add a comment as to why we want unique named blackhole clusters?

Such as: We use unique cluster names rather than having a single routed cluster in all locations as some validation in envoy cares about having distinct clusters such as weighted routing.

I would also be ok with just adding a TODO on this part if thats preferable.

test/kubernetes/e2e/features/validation/full_envoy_validation/suite.go
jbohanon: I think this does a sufficient job of testing the intended behavior is working for the purposes of this PR. I agree that other cases that would rely on this Envoy validation or otherwise would slip through and be nacked would be useful to add to this suite as they crop up.

test/kubernetes/e2e/features/validation/full_envoy_validation/suite.go
jbohanon: I would think it would be more prudent to include a happypath feature suite on the test installation with this flag enabled. WDYT?

pkg/utils/envoyutils/validation/validation.go
jbohanon: I don't think (1) necessarily follows from the comment. The mere fact that this function _is_ called from the control plane is sufficient to require a cloning so that the passed in data is not altered. 

Yes, this function does mutate the snapshot. Mutation side effects are listed in the function comments for each of the helpers.

pkg/utils/envoyutils/bootstrap/bootstrap.go
jbohanon: Definitely agree. I will throw some time at getting the tests up and running.

pkg/utils/envoyutils/validation/validation.go
```diff
@@ -21,3 +24,24 @@ const (
...
+// conversions to imitate the same config being provided as static bootsrap config to
+// Envoy, then executes Envoy in validate mode to ensure the config is valid.
+func ValidateSnapshot(
+	ctx context.Context,
+	snap envoycache.Snapshot,
+) error {
+	// THIS IS CRITICAL SO WE DO NOT INTERFERE WITH THE CONTROL PLANE.
+	snap = snap.Clone()
+
+	logger := contextutils.LoggerFrom(ctx)
```
nfuden, line 38: nit: move into the error case for micro optimization

test/kubernetes/e2e/tests/manifests/full-envoy-validation-helm.yaml
nfuden, line 1: nit: consider renaming folder to testdata https://pkg.go.dev/cmd/go/internal/test

pkg/utils/envoyutils/bootstrap/bootstrap.go
```diff
@@ -102,3 +115,220 @@ func FromFilter(filterName string, msg proto.Message) (string, error) {
...
+				return nil, nil, err
+			}
+			if hcm, ok := hcmAny.(*envoy_extensions_filters_network_http_connection_manager_v3.HttpConnectionManager); ok {
+				return hcm, f, nil
+			} else {
+				return nil, nil, eris.New("hcm config casting to concrete failed; likely wrong type")
+			}
+		}
+	}
+	return nil, nil, errNoHcm
```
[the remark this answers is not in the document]
jbohanon: If there is not an HCM we can treat it as just not an HTTP listener and continue. If there is any other error, we are possibly processing an HTTP listener and have encountered a real exception so we should puke it up

pkg/utils/envoyutils/validation/validation.go
```diff
@@ -21,3 +24,24 @@ const (
...
+
+	bootstrapJson, err := bootstrap.FromSnapshot(ctx, snap)
+	if err != nil {
+		logger.Error(err)
```
sheidkamp, line 42: Is there any more context that should be provided when logging this error?

sheidkamp (commented):
I don't feel comfortable being an approver of the guts of this functionality, so reviewing more to get up to speed. Some small comments.

pkg/utils/envoyutils/bootstrap/bootstrap.go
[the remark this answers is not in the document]
nfuden: I guess what is confusing to me is that we have many other cases where an empty return is still valid just means that there was nothing found. 🤷 i dont really care either way so happy to leave it as is  but I did want to at least ask the question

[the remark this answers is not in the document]
jbohanon: Thanks for opening the discussion! I'm open to following conventions!

pkg/utils/envoyutils/bootstrap/bootstrap.go
jbohanon: Added unit tests

sam-heilbron (approved):

pkg/utils/envoyutils/bootstrap/bootstrap_test.go
```diff
@@ -96,18 +291,199 @@ var _ = Describe("Static bootstrap generation", func() {
...
+						Name: wellknown.HTTPConnectionManager,
+						ConfigType: &envoy_config_listener_v3.Filter_TypedConfig{
+							TypedConfig: hcmAny,
+						},
+					}},
+				}},
+			})
+		})
+
+		JustBeforeEach(func() {
```
nfuden, line 344: Should we add a comment why we dont cover SecretTypeV3? Cause otherwise someone might dig into our pseduo enum for Resource types in xDS v3 and ask the question

pkg/utils/envoyutils/validation/validation.go
```diff
@@ -21,3 +25,25 @@ const (
 func ValidateBootstrap(ctx context.Context, bootstrap string) error {
 	return runner.RunEnvoyValidate(ctx, envutils.GetOrDefault(constants.EnvoyBinaryEnv, defaultEnvoyPath, false), bootstrap)
 }
+
+// ValidateSnapshot accepts an xDS snapshot, clones it, and does the necessary
+// conversions to imitate the same config being provided as static bootsrap config to
+// Envoy, then executes Envoy in validate mode to ensure the config is valid.
```
nfuden, line 31:
```suggestion
// Envoy, then executes Envoy in validate mode to ensure the config is valid.
// This is necessary as some configurations that work in dynamic do not work in static
// and some configurations may require the context of the destination such as mounted files.
```

nfuden (approved):

PR #10091, opened 2024-09-23
add target clusters to parameters api

ilackarms:
reference: https://github.com/solo-io/gloo-mesh-enterprise/issues/18646


EItanya:
I mentioned this when we discussed originally that I don't love the idea of adding concepts to this API that aren't used by `Gloo` itself. Was wondering if any further design discussions had been done about this with other and maybe product? @kcbabo @DuncanDoyle @rvennam

josh-pritchard:
I'll set something up to review with product some decisions that have been made here to achieve this PoC, we can adjust based off product feedback.

ilackarms:
closing as no longer needed

PR #10094, opened 2024-09-24
[WIP] k8s gateway serviceaccount config

jenshu:
# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -341,9 +341,16 @@ apiVersion: v1
...
 metadata:
   name: {{ include "gloo-gateway.gateway.fullname" . }}
+  {{- with (($gateway.serviceAccount).extraAnnotations) }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   labels:
     {{- include "gloo-gateway.gateway.constLabels" . | nindent 4 }}
     {{- include "gloo-gateway.gateway.labels" . | nindent 4 }}
+    {{- with (($gateway.serviceAccount).extraLabels) }}
```
sam-heilbron, line 351: Given that we support `gateway.labels`, why do we need an addditional API for "extra" labels?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -341,9 +341,16 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "gloo-gateway.gateway.fullname" . }}
+  {{- with (($gateway.serviceAccount).extraAnnotations) }}
```
sam-heilbron, line 344: nit: Given that we don't have any defined annotations, should we just call this `annotations` (and remove the extra prefix)?

projects/gateway2/deployer/values.go
```diff
@@ -80,6 +83,11 @@ type helmService struct {
 	ExtraLabels      map[string]string `json:"extraLabels,omitempty"`
 }
 
+type helmServiceAccount struct {
+	ExtraAnnotations map[string]string `json:"extraAnnotations,omitempty"`
+	ExtraLabels      map[string]string `json:"extraLabels,omitempty"`
```
sam-heilbron, line 88: How do we expect the labels on the pod/SA (and any other parts of this proxy) to stay in sync? Should they, or does it make sense to have separate APIs for each

test/kubernetes/e2e/features/deployer/suite.go
```diff
@@ -63,18 +64,29 @@ func (s *testingSuite) TestConfigureProxiesFromGatewayParameters() {
...
 	s.Require().NoError(err, "can apply manifest")
 
 	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, gwParametersManifestFile)
 	s.Require().NoError(err, "can apply manifest")
-	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyDeployment)
+	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, proxyService, proxyServiceAccount, proxyDeployment)
 	s.testInstallation.Assertions.EventuallyObjectsExist(s.ctx, gwParams)
 	s.testInstallation.Assertions.EventuallyRunningReplicas(s.ctx, proxyDeployment.ObjectMeta, gomega.Equal(1))
 
+	// check that the labels and annotations got passed through from GatewayParameters to the ServiceAccount
```
sam-heilbron, line 79: I like the addition of these in the deployer e2e tests!

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
jenshu: `gloo-gateway.gateway.labels` is just a helper func that returns hardcoded [labels](https://github.com/solo-io/gloo/blob/2a290f060d9f8fb1ce1b2585fe9e38f4f25265a6/projects/gateway2/helm/gloo-gateway/templates/_helpers.tpl#L50). We don't currently expose a way for user to set the labels. So these extraLabels will be in addition to the hardcoded ones

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
jenshu: i'd like to keep the 'extra' for naming consistency with the existing helm values we use elsewhere (including Edge ones). i think `extraAnnotations` can/should be interpreted as "in addition to the default annotations" and shouldn't be affected by whether the default annotations are empty or not (also, it doesn't preclude us from adding default annotations in the future)

projects/gateway2/deployer/values.go
jenshu: I think it's more flexible if we don't tie them together; user may want to set them independently.

In the Edge API the labels/annotations are not the same:
- it uses `gatewayProxies.gatewayProxy.podTemplate.extraGatewayProxyLabels` for the pod labels, and doesn't allow customization of the SA labels
- for annotations, it uses different values `gateway.proxyServiceAccount.extraAnnotations` (for SA) vs. `gatewayProxies.gatewayProxy.podTemplate.extraAnnotations` (for pod)

sam-heilbron (approved):

jbohanon (approved):

PR #10096, opened 2024-09-24
tests: Allow running TestWatchNamespaceSelector in Enterprise

davidjumani:
# Description

Allow running TestWatchNamespaceSelector in Enterprise by avoiding upgrades and other fixes

# Checklist:


install/helm/gloo/generate/values.go
```diff
@@ -216,7 +215,7 @@ type KnativeProxyInternal struct {
...
-	WatchNamespaceSelectors       []*v1.LabelSelector     `json:"watchNamespaceSelectors,omitempty" desc:"A list of Kubernetes selectors that specify the set of namespaces to restrict the namespaces that Gloo controllers take into consideration when watching for resources. Elements in the list are disjunctive (OR semantics), i.e. a namespace will be included if it matches any selector. An empty list means all namespaces. If this and WatchNamespaces are specified, WatchNamespaces takes precedence and this is ignored"`
+	WatchNamespaceSelectors       interface{}             `json:"watchNamespaceSelectors,omitempty" desc:"A list of Kubernetes selectors that specify the set of namespaces to restrict the namespaces that Gloo controllers take into consideration when watching for resources. Elements in the list are disjunctive (OR semantics), i.e. a namespace will be included if it matches any selector. An empty list means all namespaces. If this and WatchNamespaces are specified, WatchNamespaces takes precedence and this is ignored"`
```
davidjumani, line 218: This is needed to marshal the helm values yaml into the settings struct as the *v1.LabelSelector.MatchLabels and MatchExpressions do not have the correct json tag

test/kubernetes/e2e/features/watch_namespace_selector/suite.go
```diff
@@ -19,40 +19,40 @@ type testingSuite struct {
 
 func NewTestingSuite(ctx context.Context, testInst *e2e.TestInstallation) suite.TestingSuite {
 	return &testingSuite{
-		base.NewBaseTestingSuite(ctx, testInst, e2e.MustTestHelper(ctx, testInst), setupSuite, testCases),
+		base.NewBaseTestingSuite(ctx, testInst, setupSuite, testCases),
```
davidjumani, line 22: Since upgrades are not required anymore, removing the requirement of creating a testHelper

install/helm/gloo/generate/values.go
sam-heilbron: Good catch! Since we have an e2e test that uses this value it seems we discovered this behavior. Does this signal that a unit test to prove the defining this in helm then leads to this being defined on the settings CR? something like : https://github.com/solo-io/gloo/blob/main/install/test/helm_test.go#L613

sam-heilbron (approved):
LGTM

install/helm/gloo/generate/values.go
jbohanon: Can we recreate the portions of the structs we need with correct json tags here?

jbohanon (approved):

install/helm/gloo/generate/values.go
davidjumani: I don't think that will have any value except for tests as other fields do the same - and if anything changes it will require making changes here as well

PR #10101, opened 2024-09-24
Initial implementation of control plane API

ashishb-solo:
# Description

Implement the Control Plane API for editing Span Names.

## API changes

The control plane already has two locations where tracing can be configured:

* [ListenerTracingSettings](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/projects/gloo/api/v1/options/tracing/tracing.proto#L26) is enabled solely in the [HttpConnectionManagerSettings](https://github.com/solo-io/gloo/blob/e05b5a5a71c1dcfc266cd0acbd883031b5dd9447/projects/gloo/api/v1/options/hcm/hcm.proto#L93), and contains fields for configuring the tracing upstream as well as sampling rates (via the `TracePercentages` message). It also provides a mechanism to add more information in the span attributes via the `environment_variables_for_tags` field, and the `literals_for_tags` fields.
* [RouteTracingSettings](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/projects/gloo/api/v1/options/tracing/tracing.proto#L52-L61) is enabled solely in [RouteOptions](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/projects/gloo/api/v1/options.proto#L420) and used to configure `TracePercentages` (ie. sampling rates), and whether tracing headers should be propagated to upstream.

Note that the virtual host options object does not currently expose any tracing-related fields.

For the span name changes, this pull request proposed to add a new `SpanDecoratorSettings` message that configures the new ["otel_semconvs" filter](https://github.com/solo-io/envoy-gloo-ee/commit/96bd2ba20a36656416de4750281b85a24bb4178b#diff-5aaec8104213c9376f909170a965365b0645aa3f96935845eb5e9817a2f33bd9) (which I am considering renaming to the `SpanDecorator` filter or something like that). The intent is that it should be possible to use this message in `ListenerTracingSettings`, `RouteTracingSettings`, or any virtual host tracing settings object that we may wish to create in the future.

## Docs changes

# Context

## Notes for reviewers

# Checklist:


projects/gloo/api/v1/options/tracing/tracing.proto
```diff
@@ -43,6 +43,32 @@ message ListenerTracingSettings {
     repeated TracingTagEnvironmentVariable environment_variables_for_tags = 6;
     // Optional. If specified, Envoy will include the literals with the given tag as tracing tags.
     repeated TracingTagLiteral literals_for_tags = 7;
+
+    SpanDecoratorSettings span_decorators = 10;
```
ashishb-solo, line 47: i might remove this for the second dev build we send out for testing. it would be preferable not to add support for features until we are asked to implement and support them.

projects/gloo/api/v1/options/tracing/tracing.proto
```diff
@@ -43,6 +43,32 @@ message ListenerTracingSettings {
...
+        //     // TODO define an optional "overwrite" field that can toggle
+        //     // whether this field should be overwritten if set.
+        // }
+        // Custom attributes to set for the span. If any of these spans are
+        // already defined, they will be overwritten.
+        // repeated span_attributes = 2;
+    }
+
+    oneof span_decorator {
+      StaticSpanData static_span_data = 1;
```
ashishb-solo, line 70: we could just make this a `string static_span_name = 1;` field instead, but if we want to enable support for configuring attributes via this filter as well, using a custom message type makes it easier to support.

nfuden: I think a custom message makes more sense for some of the truly advanced decorators. like if we want to add after the fact baggage or other data that are not fully supported by an upstream filter. Most of this I hope we can upstream but in the interest of time if we start here then it will help alot

PR #10106, opened 2024-09-24
test/kubernetes/e2e: introduce profiles for tests to use recommendationed helm values

sam-heilbron:
# Description

Introduce standard recommended helm values for installations, and update our end-to-end tests to use them

# Context
## Helm Profiles
The idea here is that there are different "profiles" who will install Gloo Gateway. Depending on the use-case, the recommended Helm values will vary. We don't need to make these profiles a first-class part of the product yet, but by defining them in tests, it puts us a step closer towards being able to introduce these in our CLI or other install tooling.

## Single Source of Truth Value
I believe there is a lot of value in restructuring our tests to ensure that we define our production recommendations in a single place. Below, I try to clarify.

### Align Recommendations with Tests
We can now have user-facing documentation that is tied directly to a single file, and ensure that what we recommend to users is exactly what our tests use

### Reduce Developer Mistakes
In this PR you can see the number of manifests that exist to support our tests. Many have almost identical configuration. However, as we introduce new recommended values, you must apply that to all files, and a developer may forget that. In addition, for new tests, it is easy for a developer to not have all the context and accidentally run tests against a non-standard set of Helm values. 
By defining our recommended values in a single file, and requiring a "profile" to be defined for all tests, it reduces the chance for developer mistakes.

### Focus Unique Test Values
For a given test installation, it is easier to see what is unique about that installation, since the file now only contains the non-standard values. 

## Interesting decisions
-

## Testing steps
- All existing tests continue to pass

## Notes for reviewers
-

# Checklist:


test/kubernetes/e2e/example/info_logging_test.go
```diff
@@ -21,43 +20,40 @@ import (
 	"github.com/solo-io/gloo/test/kubernetes/testutils/gloogateway"
 )
 
-// TestInstallationWithDebugLogLevel is the function which executes a series of tests against a given installation
-func TestInstallationWithDebugLogLevel(t *testing.T) {
+// TestInstallationWithInfoLogLevel is the function which executes a series of tests against a given installation
+func TestInstallationWithInfoLogLevel(t *testing.T) {
```
nfuden, line 24: Im not sure i like this.

I understand that its more real but when a test fails i really like having debug logs. 

Thoughts on this?

nfuden (commented):
One question other makes sense to me

test/kubernetes/e2e/example/info_logging_test.go
nfuden: Perfect! Ill run the tests and do a full review!

sam-heilbron: With that said, I completely agree that relying on debug logs is really useful, and perhaps we just explore having gloo run with debug level logs by default in our tests?

test/kubernetes/e2e/features/upgrade/suite.go
```diff
@@ -43,6 +43,7 @@ func (s *testingSuite) TearDownSuite() {
 
 func (s *testingSuite) BeforeTest(suiteName, testName string) {
 	err := s.TestHelper.InstallGloo(s.Ctx, 600*time.Second, helper.WithExtraArgs([]string{
+		"--values", s.TestInstallation.Metadata.ProfileValuesManifestFile,
```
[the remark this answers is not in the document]
sam-heilbron: Yeah its a good point. I had considered it, and leaned towards requiring a profile. I did it in part because by making the requirement loud, I could more easily identify tests  that had not been updated to include it. I also did it because of a personal preference (which everyone may not agree with) that explicit tests are valuable for readability. I like the idea of saying a test author needs to explicitly define the profile and under circumstances where a non-standard (ie empty) profile is desired, there should be a rationale for it. wdyt?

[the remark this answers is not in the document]
davidjumani: Yeah makes sense to standardizing this

sam-heilbron:
> One question/comment - it looks like a lot of extra config got dropped because it is redefining defaults (for example `invalidRouteResponseCode: 404`). Do we have a way of comparing before/after configs to either ensure that nothing changed or that we are OK with the changes? Or are we happy with saying "We are using the intended defaults and test are passing - this is our new correct config"?

Yeah this is a great question. I'm kicking myself a little for the approach, since my original idea was to add profiles and keep all the same values the same. Then in a second iteration modify specific tests to reduce duplicates. Partially I took this route since it was more efficient, and partially because it allowed me to identify the profiles only by evaluating and changing all the tests.

As for the question of "how do we know that nothing bad changed", in part we don't. I figure that we are now using some new config that we weren't before (one example is in a lot of tests we had disabled leader election, when we shouldn't have). I believe that if we could disable configuration and tests pass, unfortunately it either means that our tests aren't testing what they should, OR, it means that the configuration we had previously set was unnecessary. I'm inclined to trust the accuracy of the tests and rely on the fact that if CI is passing, it means the necessary values are defined. what do you think @sheidkamp ?

sam-heilbron:
Adding the `work in progress` label so people can approve and it won't merge.

test/kubernetes/e2e/example/info_logging_test.go
nfuden: I guess its not a huge deal and runs faster with info. 
Basically I think the flow is run iwth info. If fails then locally run with debug.
If only failing remotely add debug to specific test and run with until it fails in ci

test/kubernetes/e2e/tests/manifests/istio-revision-helm.yaml
```diff
@@ -10,37 +10,7 @@ global:
...
         tag: 1.22.0 # This tag has to match the version of Istio being used in the test
-  podSecurityStandards:
-    container:
-      enableRestrictedContainerDefaults: true
-gloo:
-  logLevel: info
-  disableLeaderElection: true
-  deployment:
-    #  We have limited GitHub action resources which can cause containers to not create
-    # therefore we lessen the cpu resource requests values from the default (500m) to 100m.
```
nfuden, line 21: Do we no longer hit this failure?

sheidkamp:
> > One question/comment - it looks like a lot of extra config got dropped because it is redefining defaults (for example `invalidRouteResponseCode: 404`). Do we have a way of comparing before/after configs to either ensure that nothing changed or that we are OK with the changes? Or are we happy with saying "We are using the intended defaults and test are passing - this is our new correct config"?
> 
> Yeah this is a great question. I'm kicking myself a little for the approach, since my original idea was to add profiles and keep all the same values the same. Then in a second iteration modify specific tests to reduce duplicates. Partially I took this route since it was more efficient, and partially because it allowed me to identify the profiles only by evaluating and changing all the tests.
> 
> As for the question of "how do we know that nothing bad changed", in part we don't. I figure that we are now using some new config that we weren't before (one example is in a lot of tests we had disabled leader election, when we shouldn't have). I believe that if we could disable configuration and tests pass, unfortunately it either means that our tests aren't testing what they should, OR, it means that the configuration we had previously set was unnecessary. I'm inclined to trust the accuracy of the tests and rely on the fact that if CI is passing, it means the necessary values are defined. what do you think @sheidkamp ?

I think it's OK to rely on CI here. I don't think the existing configs were rigorously vetted (probably a lot of copy and pasting) and were relying on CI passing as the validation in the first place. With this PR, we end up with much more well-defined config, and as you pointed out with the leader election example, we will now be more consistent.

test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
```diff
@@ -0,0 +1,112 @@
...
+# In this file we define any values which are not the default, but are the recommended pattern of installing,
+# and thus testing, Gloo Gateway. Any values in this file should ideally be converted into
+# defaults in our values-template.yaml overtime. That way, users of the product can have the easiest experience,
+# and gain the best value, without custom configuration.
+
+# Configuration that is global to Gloo Gateway
+global:
+  podSecurityStandards:
+    container:
+      enableRestrictedContainerDefaults: true
```
sheidkamp, line 18: I like it!

sheidkamp (approved):

test/kubernetes/e2e/tests/manifests/istio-revision-helm.yaml
sam-heilbron: I opted to set this as a common 
-recommendation since we want people to define limits and we want our tests to have certain limits as well

test/kubernetes/e2e/suite.go
```diff
@@ -50,7 +50,7 @@ func NewSuiteRunner(ordered bool) SuiteRunner {
 	return new(suites)
 }
 
-func (o orderedSuites) Run(ctx context.Context, t *testing.T, testInstallation *TestInstallation) {
+func (o *orderedSuites) Run(ctx context.Context, t *testing.T, testInstallation *TestInstallation) {
```
jbohanon, line 53: nit: what changed that required these to be converted to pointer receivers?

sam-heilbron: My IDE complained that we had a struct that mixed recievers

jbohanon (approved):

Issue #8635, opened 2023-09-01
Invalid resourceVersion error in upstream discovery


jenshu:
### Gloo Edge Product

Enterprise

### Gloo Edge Version

v1.14.x

### Kubernetes Version

1.27

### Describe the bug

A customer reported that gloo discovery is not creating all the upstreams that they expect. Deleting and recreating the services doesn't seem to help.

When checking the `discovery` pod logs, they saw errors like:
`{"level":"error","ts":"2023-08-30T12:43:51.281Z","logger":"uds.v1.event_loop.uds","caller":"syncer/setup_syncer.go:117","msg":"error: event_loop.uds: error in uds plugin : reconciling resource default-xxx-80: updating kube resource default-xxx-80: (want 2398796740): upstreams.gloo.solo.io \"default-xxx-80\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update","version":"undefined","stacktrace":"github.com/solo-io/gloo/projects/discovery/pkg/uds/syncer.RunUDS.func2\n\t/go/pkg/mod/github.com/solo-io/gloo@v1.14.12/projects/discovery/pkg/uds/syncer/setup_syncer.go:117"}`

Deleting the erroring upstreams didn't seem to resolve the issue (upstreams would get recreated again, with same error in logs)

Note:  I have also seen a similar error (`metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`) on a clean install, but in my case it didn't seem to affect upstream creation.

Recommended steps:
- work with customer to reproduce / figure out under what circumstances the error occurs
- confirm whether and how it affects upstream discovery
- if needed, share any short-term workarounds while customer is awaiting a fix

### Expected Behavior

resourceVersion error should not occur and/or should resolve itself with retries, and expected upstreams should be created

### Steps to reproduce the bug

See above

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206768562311555/1208262642635027) by [Unito](https://www.unito.io)

DuncanDoyle:
@jenshu Any update on this issue? Without a reproducer this is not actionable. Did "the customer" ever come back to you? If not, I'm gonna close this ticket (and can create a new one, or re-open if this issue surfaces again).

jenshu:
@DuncanDoyle - responded offline with context

DuncanDoyle:
Closing, as there is no reproducer. Can re-open if this surfaces again.

sam-heilbron:
[...]

Issue #8227, opened 2023-05-16
Discovery update of unmodified Upstreams when K8s service has changed 


bdecoste:
### Version

1.13.x

### Is your feature request related to a problem? Please describe.

A K8s service selector was modified and the corresponding discovered Upstream was not updated. 

### Describe the solution you'd like

Discovery should update discovered Upstreams when changes to the corresponding K8s service are detected if the Upstreams have not been modified. Possibly use `kubectl.kubernetes.io/last-applied-configuration` annotation or the `discovered_by: kubernetesplugin` label to determine if the resource has been changed since discovery. 

### Describe alternatives you've considered

Delete discovered Upstream to force re-discovery

PR #630, opened 2019-03-27
uds only starts once and then gets kicked

ilackarms:
fix race when reconcile called more than once
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/631

PR #10118, opened 2024-09-26
Discovery resource version err

bewebi:
# Description

Fix a bug that caused Resync for discovered Upstreams to fail when `watchLabels` were in use

## Code changes
- Remove watchLabels from the selector used to select Upstreams for Resync
  - Add kube2e test 

# Context

See #8635 for context on issue, as well as [root cause](https://github.com/solo-io/gloo/issues/8635#issuecomment-2374785559) and [reasoning around the solution](https://github.com/solo-io/gloo/issues/8635#issuecomment-2374923454)

I did encounter #8227 as well while working on this but have not investigated that piece further

This bug has been present since at least v1.14 and therefore we expect to backport this to all LTS branches

### Manual testing

I followed the reproduction steps from the issue to reproduce the bug, observing both the error in logs and that the discovered Upstream would not get updates when its parent Service was updated

I built the discovery image from this branch, loaded it into KinD, and modified the `discovery` deployment to use the test image
```
$ VERSION=0.0.0-test make -B kind-build-and-load-discovery
```

Following the same reproduction steps I then observed that the Upstream was, in fact, updated when its parent service was, and that logs no longer contained errors

### Automated testing

There is no testing infrastructure for unit tests of Discovery resync, and any such tests would require extensive mocking of k8s, complicated async setup, etc. and would therefore represent a significant lift
In my estimation this would not be worthwhile as a kube2e test, which has been added, is sufficient to test the buggy behavior

# Checklist:


test/kube2e/gateway/gateway_test.go
```diff
@@ -1195,6 +1195,64 @@ var _ = Describe("Kube2e: gateway", func() {
 			}, "15s", "0.5s").ShouldNot(BeNil())
 		})
 
+		It("Modifies discovered Upstream when Service is modified", func() {
```
jenshu, line 1198: nit: is there a place in the kubernetes/e2e tests where we can insert this new test?

jenshu (approved):

projects/gloo/pkg/discovery/discovery.go
```diff
@@ -136,12 +136,14 @@ func (d *UpstreamDiscovery) Resync(ctx context.Context) error {
...
 		udsName := strings.Replace(reflect.TypeOf(uds).String(), "*", "", -1)
 		udsName = strings.Replace(udsName, ".", "", -1)
+
+		// selecting on the discovery plugin label will get all Upstreams created by Discovery
+		// there is no need to select on watchLabels as these are not present on Upstream metadata, nor are they
+		// necessary to identify Upstreams that may need to be resynced
 		selector := map[string]string{
 			"discovered_by": udsName,
 		}
-		for k, v := range d.extraSelectorLabels {
```
sam-heilbron, line 142: This deletes the only usage of `d.extraSelectorLabels`. I think we need to go through the history (It looks like it was added in https://github.com/solo-io/gloo/pull/630) and understand what this was intended to do, and how it is communicated to users (in our docs or API reference). 

That way we can delete all references to it, if it is in fact safe to delete

projects/gloo/pkg/discovery/discovery.go
```diff
@@ -136,12 +136,14 @@ func (d *UpstreamDiscovery) Resync(ctx context.Context) error {
 	for uds, desiredUpstreams := range d.latestDesiredUpstreams {
 		udsName := strings.Replace(reflect.TypeOf(uds).String(), "*", "", -1)
 		udsName = strings.Replace(udsName, ".", "", -1)
+
+		// selecting on the discovery plugin label will get all Upstreams created by Discovery
+		// there is no need to select on watchLabels as these are not present on Upstream metadata, nor are they
+		// necessary to identify Upstreams that may need to be resynced
```
sam-heilbron, line 142: How does this relate to this API that we expose https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/settings.proto#L281? If a user has changed those values, wouldn't we continue to select and process resources which they wouldn't expect?

test/kube2e/gateway/gateway_test.go
sam-heilbron: Seconded. We have found a lot of value in these new style tests and those are documented here: https://github.com/solo-io/gloo/blob/main/test/kubernetes/README.md#end-to-end-testing

bewebi: I don't believe there are existing tests in the new framework covering Discovery
I do see that there is a guide to migrating tests, but I haven't personally done so for any tests yet
I'm not sure whether it's strictly a requirement for new tests, but it may be worth the effort

test/kube2e/gateway/gateway_test.go
```diff
@@ -1195,6 +1195,64 @@ var _ = Describe("Kube2e: gateway", func() {
...
+			// This tests the fix for a bug whereby Discovery would fail to update discovered Upstreams on Service
+			// updates when watchLabels were enabled
+
+			watchedKey, watchedValue := "watchKey", "watchValue"
+			bonusKey, bonusValue, modifiedBonusValue := "foo", "bar", "baz"
```
sam-heilbron, line 1203: nit: When I was debugging a failure, I saw a set of values that were almost identical, except one had `bar` and the other `baz`. I think we can try to make this easier to signal a difference, beyond a character. One option would be `bar` -> `modified-bar`

test/kube2e/gateway/gateway_test.go
```diff
@@ -1195,6 +1195,64 @@ var _ = Describe("Kube2e: gateway", func() {
...
+
+			// get the service, modify its bonus label, and update it
+			svc, err := resourceClientset.KubeClients().CoreV1().Services(namespace).Get(ctx, svcName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			svc.Labels[bonusKey] = modifiedBonusValue
+
+			_, err = resourceClientset.KubeClients().CoreV1().Services(namespace).Update(ctx, svc, metav1.UpdateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			// expect the discovered Upstream's DiscoveryMetadata to eventually reflect the updated label
```
sam-heilbron, line 1240: nit: To some extent this feels like an integration test of the discovery component and etcd. It seems like what we're trying to test is that updates to services can get picked up, processed, and transmitted to our proxies. Once possible enhancement to this test would be to validate that the product (ie the proxies) behaves a certain way

test/kube2e/gateway/gateway_test.go
sam-heilbron: `I don't believe there are existing tests in the new framework covering Discovery` I agree, though all of our new features (https://github.com/solo-io/gloo/commits/main/test/kubernetes/e2e/features) have been created by developers unfamiliar with the framework. I think it would be valuable to write this as a new-style test as its the pattern that is being adopted by the team, and will make it easier for us to migrate the old-style tests over

projects/gloo/pkg/discovery/discovery.go
bewebi: I did look through the history, #630 was the last time it was modified, it in fact dates to the original commit from this repo, and no additional context was added in PR descriptions, comments, etc.

I surmise that the intent was to ensure we were only selecting the intended Upstreams to resync, however discovered Upstreams have their own unique label and _do not_ inherit labels from their parent Services, so this was unnecessary and, in fact, incorrect

This is not user-facing, it's not even an exported field, and does not impact the watchLabel functionality other than to introduce the bug which we're resolving here

projects/gloo/pkg/discovery/discovery.go
bewebi: That refers to watching Services, and nothing in this PR changes how Services are watched/selected

projects/gloo/pkg/discovery/discovery.go
sam-heilbron: Yeah what threw me off a little is that when I traced back the existing `extraLabelSelectors`, I noticed that it was pulled from:
- startUds watchOpts parameter (https://github.com/solo-io/gloo/blob/2a290f060d9f8fb1ce1b2585fe9e38f4f25265a6/projects/gloo/pkg/discovery/discovery.go#L96)
- which is called from the uds setup (https://github.com/solo-io/gloo/blob/2a290f060d9f8fb1ce1b2585fe9e38f4f25265a6/projects/discovery/pkg/uds/syncer/setup_syncer.go#L93)
- which is passed into RunUDS (https://github.com/solo-io/gloo/blob/2a290f060d9f8fb1ce1b2585fe9e38f4f25265a6/projects/discovery/pkg/uds/syncer/setup_syncer.go#L21)
- which seems to come from https://github.com/solo-io/gloo/blob/2a290f060d9f8fb1ce1b2585fe9e38f4f25265a6/projects/gloo/pkg/syncer/setup/setup_syncer.go#L424


and so while your changes make sense, what confuses me a little is _if_ the selector that we've deleted was always nil, why would deleting it have any impact?

bewebi: [Here](https://github.com/solo-io/gloo/blob/b2932084fa1f23c29b4b332d43bc2a40d300d7ec/projects/discovery/pkg/uds/syncer/setup_syncer.go#L34) is where the watchLabels from Settings are added to `WatchOpts` which are then passed to `StartUds()`

test/kubernetes/e2e/features/discovery_watchlabels/discovery_watchlabels_suite.go
```diff
@@ -0,0 +1,102 @@
...
+	// add one service with labels matching our watchLabels
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-with-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
+	s.Assert().NoError(err, "can apply service")
+
+	// add one service without labels matching our watchLabels
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-without-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
+	s.Assert().NoError(err, "can apply service")
+
+	// eventually an Upstream should be created for the Service with labels
+	labeledUsName := fmt.Sprintf("%s-%s-%d", s.testInstallation.Metadata.InstallNamespace, "example-svc", 8000)
```
sam-heilbron, line 59: Should we use https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/kubernetes/uds_convert.go#L77?

test/kubernetes/e2e/features/discovery_watchlabels/discovery_watchlabels_suite.go
```diff
@@ -0,0 +1,102 @@
...
+	s.T().Cleanup(func() {
+		err := s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-with-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
+		s.Assertions.NoError(err, "can delete service")
+
+		err = s.testInstallation.Actions.Kubectl().DeleteFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-without-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
+		s.Assertions.NoError(err, "can delete service")
+	})
+
+	// add one service with labels matching our watchLabels
+	err := s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-with-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
```
sam-heilbron, line 51: nit: Our other tests define these file paths in a standalone file https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/features/glooctl/types.go#L14

test/kubernetes/e2e/tests/manifests/discovery-watchlabels-test-helm.yaml
```diff
@@ -0,0 +1,4 @@
+discovery:
```
sam-heilbron, line 1: I think we need to explicity enable discovery, yeah?

test/kubernetes/e2e/tests/discovery_watchlabels_test.go
```diff
@@ -0,0 +1,59 @@
...
+			testInstallation.PreFailHandler(ctx)
+		}
+
+		testInstallation.UninstallGlooGateway(ctx, func(ctx context.Context) error {
+			return testHelper.UninstallGlooAll()
+		})
+	})
+
+	// Install Gloo Gateway with only Gloo Edge Gateway APIs enabled
+	testInstallation.InstallGlooGateway(ctx, func(ctx context.Context) error {
```
sam-heilbron, line 54: Lets use https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/test.go#L176

test/kubernetes/e2e/features/discovery_watchlabels/discovery_watchlabels_suite.go
```diff
@@ -0,0 +1,102 @@
...
+			},
+		},
+	)
+
+	// modify the non-watched label on the labeled service
+	err = s.testInstallation.Actions.Kubectl().ApplyFile(s.ctx, filepath.Join(util.MustGetThisDir(), "testdata/service-with-modified-labels.yaml"), "-n", s.testInstallation.Metadata.InstallNamespace)
+	s.Assert().NoError(err, "can re-apply service")
+
+	// expect the Upstream's DiscoveryMeta to eventually match the modified labels from the parent Service
+	time.Sleep(10 * time.Second) // would rather do an Eventually
```
sam-heilbron, line 93: What prevents us from wrapping the below code in an eventually?

test/kubernetes/e2e/tests/manifests/discovery-watchlabels-test-helm.yaml
bewebi: I haven't specified that when reproducing the issue, developing the solution, or developing the tests, but from looking at the Helm templates it does seem like it would be needed
It certainly doesn't hurt

sam-heilbron: My guess is that discovery.enabled defaults to true, which is why it wasn't problematic before. Howevever, now in our common-recommendations we set it to false (https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/tests/manifests/common-recommendations.yaml#L110)

test/kube2e/gateway/gateway_test.go
sam-heilbron, line 1198: Now that we've added the new-style test, I think we can remove this one, wdyt?

Also, do you think it's worth it to try to migrate a few other test cases around this feature?

bewebi: I had assumed I would remove the kube2e test, but on further consideration I do think it's useful since this will be backported to LTS branches that do not have the new kubernetes tests

This will allow the backport PRs to have testing without introducing any "new" code

I'll take a stab at migrating additional tests

bewebi: All tests are now encompassed in the new suite, though four are grouped under one test now
- Discovery creates Upstream for Service with watchLabel
- Discovery can update Upstream for Service with watchLabel when Service is updated (this includes the previous case anyway)
- Discovery does not create an Upstream for a Service without the watchLabel(s)
- Discovery does not create an Upstream for a Service whose watchLabel doesn't have the expect value

I'm not sure if we prefer to keep tests as narrow as possible over minimizing setup time at this point

sam-heilbron: `but on further consideration I do think it's useful since this will be backported to LTS branches that do not have the new kubernetes tests` This is a great idea. one caveat will just be finding ways to communicate to future devs where the overlap is, and where changes should be made. Would there be value in a comment on the main branch that clarifies that we are going to remove the legacy e2e test, and perhaps a comment on the backported branches pointing to where the tests will live in the new-style?

sam-heilbron (approved):

PR #10120, opened 2024-09-26
1.17: Poc/scrubbedsecrets

nfuden:
So hyper strict tls validation is... good.
That being said everyone is right its a misuse and misunderstanding of pem but we need to make sure that Envoy doesnt puke.

That being said extra lines shouldnt mess up a cert but what we SHOULD do is take the actionable pieces like the RFC says.

Now I hear you, but invalid stuff like an unterminated block is still an easy mistake for an operator.

I understand this and propose we eventually have a metric increment. In the meantime this helps us still have a lower chance of envoy nacks and only opens us slightly to the concept of a minimized cert chain due to invalid user input.


changelog/v1.17.10/strict-tls-validation.yaml
```diff
@@ -0,0 +1,11 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/6772
+    resolvesIssue: false
+    description: >-
+      Changes the 1.17.8 tls secret validation strictness.
+      Instead of just performing pure go validation and erroring if we lose things we now will be slightly less aggressive for certs.
+      We still validate several pieces of the cert but also scrub down to usable bits.
+      Flip side here is that we may miss some partially invalid pieces of PEM.
+      This means we are RFC compliant but it may mean that there could exist some cert data in an edge case which we will elide from envoy but previously woul be nacked.
```
ashishb-solo, line 10: If it's possible, it would be great if we could brush this up a bit:

1. "We still validate several pieces of the cert but also scrub down to usable bits." Could we be a bit more specific here? Which pieces do we validate? What do we mean by "usable bits" here?
2. "Flip side here is that we may miss some partially invalid pieces of PEM." Again, could we specify which pieces we are talking about here?
3. "previously woul be nacked" s/woul/would

Or to just be a bit more general, I'm having a hard time understanding from this changelog entry what changes are occurring. I can glean that we are being a little less strict with our cert validation, but I'm still not sure in what set of circumstances I can expect Gloo's behaviour to be different.

Alternately, do we want to consider putting this information in documentation somewhere and linking to it here? I worry that there might be important information here that should live somewhere more prominent than the changelog.

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -19,55 +18,50 @@ var (
...
 	if err != nil {
-		return err
-	}
-	trimmedEncoded := strings.TrimSpace(string(reencoded))
-	if trimmedEncoded != strings.TrimSpace(string(certChain)) {
-		return fmt.Errorf("certificate chain does not match parsed certificate")
+		// return err rather than sanitize. This is to maintain UX with older versions and to keep in line with gateway2 pkg.
+		return
 	}
+	cleanedChainBytes, err := cert.EncodeCertificates(candidateCert...)
```
ashishb-solo, line 63: do we need to check this `err` here?

projects/gloo/pkg/utils/ssl.go
```diff
@@ -449,18 +454,12 @@ func isValidSslKeyPair(certChain, privateKey, rootCa string) error {
...
-	reencoded, err := cert.EncodeCertificates(candidateCert...)
-	if err != nil {
-		return err
-	}
-	trimmedEncoded := strings.TrimSpace(string(reencoded))
-	if trimmedEncoded != strings.TrimSpace(certChain) {
-		return fmt.Errorf("certificate chain does not match parsed certificate")
+		return "", err
 	}
+	cleanedChainBytes, err := cert.EncodeCertificates(candidateCert...)
```
ashishb-solo, line 459: do we need to check `err` here?

projects/gloo/pkg/utils/ssl_test.go
```diff
@@ -189,6 +189,25 @@ MIID6TCCA1ICAQEwDQYJKoZIhvcNAQEFBQAwgYsxCzAJBgNVBAYTAlVTMRMwEQYD`
...
+			func(c func() utils.CertSource) {
+				// unterminated cert in chain is valid for go b ut not envoy
+				tlsSecret.CertChain += `# subject
+
+
+
+
+-----BEGIN HEADERS-----
+Header: 1
+-----END HEADERS----------BEGIN CERTIFICATE-----
```
ashishb-solo, line 202: i'm a bit confused here. is `BEGIN/END HEADERS` just supposed to be junk at the start of the certificate? is the expectation here that we discard the entire certificate because there's no newline between  `END HEADERS` and `BEGIN CERTIFICATE`?

projects/gloo/pkg/utils/ssl.go
```diff
@@ -449,18 +454,12 @@ func isValidSslKeyPair(certChain, privateKey, rootCa string) error {
...
 		// return err rather than sanitize. This is to maintain UX with older versions and to keep in line with gateway2 pkg.
-		return err
-	}
-	reencoded, err := cert.EncodeCertificates(candidateCert...)
-	if err != nil {
-		return err
-	}
-	trimmedEncoded := strings.TrimSpace(string(reencoded))
-	if trimmedEncoded != strings.TrimSpace(certChain) {
-		return fmt.Errorf("certificate chain does not match parsed certificate")
```
ashishb-solo, line 460: can you clarify why we remove `TrimSpace` here? Is this because we no longer consider untrimmed certificates as invalid?

ashishb-solo (commented):
This looks sound to me overall. I'm having a bit of trouble reviewing this because I feel like I'm lacking a bit of context here. If I understand correctly what we're doing here, we're being a little less strict with our secret validation strictness - but an example of exactly what we're changing and how the behaviour of how we handle that example before/after this change is merged would help out a lot (if it's not too much to ask). Thanks!

projects/gloo/pkg/utils/ssl_test.go
nfuden: This would be rejected by envoy due to not having a new line after Header: 1
I have added a comment to the test

projects/gloo/pkg/utils/ssl.go
nfuden: Because openssl doesnt puke on extra spaces. We previously did it as part of our comparison analysis.

changelog/v1.17.10/strict-tls-validation.yaml
nfuden: i like it. pushed

ashishb-solo (commented):
happy to approve as is. do you have any thoughts on whether we need to add more documentation for this functionality? what we allow and what we don't allow seems a bit blurry now.

projects/gloo/pkg/utils/ssl_test.go
```diff
@@ -177,13 +177,32 @@ var _ = Describe("Ssl", func() {
...
 MIID6TCCA1ICAQEwDQYJKoZIhvcNAQEFBQAwgYsxCzAJBgNVBAYTAlVTMRMwEQYD`
 				_, err := resolveCommonSslConfig(c(), secrets)
-				Expect(err).To(HaveOccurred())
+				Expect(err).NotTo(HaveOccurred())
+
+			},
+			Entry("upstreamCfg", func() utils.CertSource { return upstreamCfg }),
+			Entry("downstreamCfg", func() utils.CertSource { return downstreamCfg }),
+		)
+		DescribeTable("should scrub invalid non-certs like bad headers (no space afterdefinitions) if invalid private cert is provided",
```
ashishb-solo, line 192: > if invalid private cert is provided

nit: this test no longer has a private certificate anymore

ashishb-solo (approved):

projects/gateway2/translator/sslutils/ssl_utils.go
```diff
@@ -19,55 +18,44 @@ var (
...
 	if err != nil {
-		return err
-	}
-	trimmedEncoded := strings.TrimSpace(string(reencoded))
-	if trimmedEncoded != strings.TrimSpace(string(certChain)) {
-		return fmt.Errorf("certificate chain does not match parsed certificate")
+		// return err rather than sanitize. This is to maintain UX with older versions and to keep in line with gateway2 pkg.
+		return "", err
 	}
+	cleanedChainBytes, err := cert.EncodeCertificates(candidateCert...)
```
jenshu, line 57: is there a way to tell if we dropped any part of the cert chain, and if so, emit some sort of warning or log?

nfuden: we could but it would slam logs. 
We could do a metric if we really care enough

nfuden: Wdyt?

ashishb-solo: i like the metrics idea

ashishb-solo: but happy to push through as is too. not a blocker for approval from me

jenshu (approved):

Issue #10115, opened 2024-09-25
[gloo/cli]: Uses v1.0.0-rc1 CRDs instead of v1.0.0


danehans:
### Gloo Edge Product

Open Source

### Gloo Edge Version

v1.17.7

### Kubernetes Version

1.28.1

### Describe the bug

`GatewayCrds` is a public variable that is a []byte of Gateway API CRDs:

```
$ grep -ri "v1.0.0" projects/gateway2
projects/gateway2/crds/gateway-crds.yaml:    gateway.networking.k8s.io/bundle-version: v1.0.0-rc1
projects/gateway2/crds/gateway-crds.yaml:    gateway.networking.k8s.io/bundle-version: v1.0.0-rc1
projects/gateway2/crds/gateway-crds.yaml:    gateway.networking.k8s.io/bundle-version: v1.0.0-rc1
projects/gateway2/crds/gateway-crds.yaml:    gateway.networking.k8s.io/bundle-version: v1.0.0-rc1
```

As you can see from the output above, these CRDs are based on rc1 instead of the final v1.0.0 release.

This variable is used throughout the gloo cli codebase:

```
$ grep -ri GatewayCrds .
./projects/gateway2/crds/crds.go:var GatewayCrds []byte
./projects/gloo/cli/pkg/cmd/install/kubegateway/uninstall.go:	crds, err := deployer.ConvertYAMLToObjects(cli.Scheme(), crds.GatewayCrds)
./projects/gloo/cli/pkg/cmd/install/kubegateway/install.go:	crdsExist, err := kubegatewayutils.DetectKubeGatewayCrds(cfg)
./projects/gloo/cli/pkg/cmd/install/kubegateway/install.go:		crds, err := deployer.ConvertYAMLToObjects(cli.Scheme(), crds.GatewayCrds)
./projects/gloo/cli/pkg/cmd/check/kube_gateway.go:	hasCRDs, err := kubegatewayutils.DetectKubeGatewayCrds(cfg)
./projects/gloo/cli/pkg/kubegatewayutils/detect.go:func DetectKubeGatewayCrds(cfg *rest.Config) (bool, error) {
```

### Expected Behavior

That v1.0.0 CRDs are being used by the gloo cli.

### Steps to reproduce the bug

Review the steps above or use gloo cli to install.

### Additional Environment Detail

N/A

### Additional Context

N/A

[...]

PR #10121, opened 2024-09-26
*: Bump the sigs.k8s.io/gateway-api dependency to v1.1.0

timflannagan:
# Description

Bump the sigs.k8s.io/gateway-api client and server versions to v1.1.0. This is the
initial step towards making Gloo conformant with the new(ish) OSS release.

In order to accomplish this, the existing conformance-related make targets needed
to be refactored:
    
- Reduces the length of the CONFORMANCE_ARGS variable. This helps
  make the Makefile easier and more flexible for dev input.
- Removes the conformance-experimental target. This target is no
  longer relevant as the v1.1 GW API release overhauled the
  conformance suite and consolidated everything into a single
  suite entrypoint.

Further work will be needed to add support for EndpointSlices to make Gloo fully
conformant with the 1.1 release. Without it, we're partially conformant as upstream
introduced a standard/core test that enforces different service backend types.

## CI changes

- Updates the ci/kind/setup-kind.sh bash script to
  install the GW API CRDs using kustomize. Right now, there's
  no decent way to install these CRDs without installing the raw
  CRD artifacts from GH release artifacts. I.E. there's no helm
  install equivalent that's provided by the community.
- Remove the "profile" workflow_dispatch input option. This is no longer
  relevant with the 1.) consolidated standard/experimental suites and 2.)
  the "standard" suite supports generating conformance YAML reports.
  Previously, only the "experimental" suite would generate these reports
- Removes the conditional that guarded the store-artifact GHA job. Now,
  all conformance runs will attempt to upload the generated YAML report.
  We may want to tinker with this behavior over time.

# Context

## Testing steps

https://github.com/solo-io/gloo/actions/runs/11061690230

# Checklist:

### DIFF
--- a/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
+++ b/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
@@ -62,5 +62,7 @@ runs:
     if: ${{ success() }}
     uses: actions/upload-artifact@v4
     with:
-      name: conformance-gloo-gateway-report-${{ matrix.kube-version.node }}
-      path: _test/conformance/report.yaml
+      # Name of the path to upload. The VERSION variable refers to the Makefile
+      # VERSION variable.
+      name: conformance-gloo-gateway-report
+      path: _test/conformance/${{ env.VERSION }}-report.yaml
--- a/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
+++ b/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
@@ -53,21 +53,14 @@ runs:
       fi
   - name: Run the kubernetes gateway API conformance tests
     shell: bash
-    run: |
-      # Default to the standard profile if none is provided. Otherwise,
-      # assume the profile is experimental.
-      if [[ -z ${{ matrix.profile }} ]]; then
-        make conformance
-      else
-        make conformance-experimental
-      fi
+    run: make conformance
   - name: Capture debug information when tests fail
     if: ${{ failure() }}
     shell: bash
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
   - name: Upload conformace reports if profile is experimental
-    if: "${{ matrix.profile == 'experimental' }}"
+    if: ${{ success() }}
     uses: actions/upload-artifact@v4
     with:
-      name: conformance-gloo-gateway-report
+      name: conformance-gloo-gateway-report-${{ matrix.kube-version.node }}
       path: _test/conformance/report.yaml
--- a/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
+++ b/.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
@@ -58,7 +58,7 @@ runs:
     if: ${{ failure() }}
     shell: bash
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
-  - name: Upload conformace reports if profile is experimental
+  - name: Upload reports
     if: ${{ success() }}
     uses: actions/upload-artifact@v4
     with:
--- a/Makefile
+++ b/Makefile
@@ -1207,13 +1207,11 @@ $(TEST_ASSET_DIR)/conformance/conformance_test.go:
 	cat $(shell go list -json -m sigs.k8s.io/gateway-api | jq -r '.Dir')/conformance/conformance_test.go >> $@
 	go fmt $@
 
-# Pull the experimental conformance test suite from the k8s gateway api repo and copy it into the test dir.
-$(TEST_ASSET_DIR)/conformance/experimental_conformance_test.go:
-	mkdir -p $(TEST_ASSET_DIR)/conformance
-	cat $(shell go list -json -m sigs.k8s.io/gateway-api | jq -r '.Dir')/conformance/experimental_conformance_test.go >> $@
-	go fmt $@
-
-CONFORMANCE_ARGS := -gateway-class=gloo-gateway -supported-features=Gateway,ReferenceGrant,HTTPRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRouteResponseHeaderModification,HTTPRoutePortRedirect,HTTPRouteHostRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteRequestMirror
+CONFORMANCE_SUPPORTED_FEATURES ?= -supported-features=Gateway,ReferenceGrant,HTTPRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRouteResponseHeaderModification,HTTPRoutePortRedirect,HTTPRouteHostRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteRequestMirror
+CONFORMANCE_SUPPORTED_PROFILES ?= -conformance-profiles=GATEWAY-HTTP
+CONFORMANCE_SKIP_TESTS ?= -skip-tests=HTTPRouteServiceTypes
+CONFORMANCE_REPORT_ARGS ?= -report-output=$(TEST_ASSET_DIR)/conformance/$(VERSION)-report.yaml -organization=solo.io -project=gloo-gateway -version=$(VERSION) -url=github.com/solo-io/gloo -contact=github.com/solo-io/gloo/issues/new/choose
+CONFORMANCE_ARGS := -gateway-class=gloo-gateway $(CONFORMANCE_SKIP_TESTS) $(CONFORMANCE_SUPPORTED_FEATURES) $(CONFORMANCE_SUPPORTED_PROFILES) $(CONFORMANCE_REPORT_ARGS)
 
 .PHONY: conformance ## Run the conformance test suite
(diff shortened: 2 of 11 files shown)


ci/kind/setup-kind.sh
```diff
@@ -75,7 +80,7 @@ fi
 make -s build-cli-local
 
 # 5. Apply the Kubernetes Gateway API CRDs
-kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml
+kubectl apply -k "https://github.com/kubernetes-sigs/gateway-api/config/crd/$CONFORMANCE_CHANNEL?ref=$CONFORMANCE_VERSION"
```
timflannagan, line 83: I mentioned this in one of the commits, but using kustomize makes maintenance a bit easier. It supports reading from a remote GH URL and we can avoid the scenario we need a series of `k apply -f ...` commands to install standard or experimental GW API CRDs.

Makefile
timflannagan, line 1: TODO: Before I forgot, we need to update the nested, embedded CRDs to tackle https://github.com/solo-io/gloo/issues/10115 as well.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan, line 1: Note, this action still has the limitation where it's getting it's inputs from the "matrix" key instead of the "inputs" key. I played around with this locally but wasn't able to get it working perfectly and didn't want to hold up this work while I hacked on that in the background.

timflannagan:
Looks like quay is having a major outage right now.

ci/kind/setup-kind.sh
sam-heilbron: Good callout! 

nit: Should we include that context in a comment here so future devs can easily find it?

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -53,21 +53,16 @@ runs:
...
-        make conformance-experimental
-      fi
+    run: make conformance
   - name: Capture debug information when tests fail
     if: ${{ failure() }}
     shell: bash
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
   - name: Upload conformace reports if profile is experimental
-    if: "${{ matrix.profile == 'experimental' }}"
+    if: ${{ success() }}
```
sam-heilbron, line 62: Let's update the name of this step, since it no longer is accurate. We can probably just do something like:
```
Upload Reports
```

sam-heilbron (approved):

changelog/v1.18.0-beta24/support-gw-api-1.1-conformance.yaml
jbohanon, line 1: can we add a `DEPENDENCY_BUMP` entry to this changelog for the `sigs.k8s.io/gateway-api`?

Makefile
```diff
@@ -1204,35 +1204,26 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
...
-# Pull the experimental conformance test suite from the k8s gateway api repo and copy it into the test dir.
-$(TEST_ASSET_DIR)/conformance/experimental_conformance_test.go:
-	mkdir -p $(TEST_ASSET_DIR)/conformance
-	cat $(shell go list -json -m sigs.k8s.io/gateway-api | jq -r '.Dir')/conformance/experimental_conformance_test.go >> $@
-	go fmt $@
-
-CONFORMANCE_ARGS := -gateway-class=gloo-gateway -supported-features=Gateway,ReferenceGrant,HTTPRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRouteResponseHeaderModification,HTTPRoutePortRedirect,HTTPRouteHostRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteRequestMirror
+CONFORMANCE_SUPPORTED_FEATURES ?= -supported-features=Gateway,ReferenceGrant,HTTPRoute,HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRouteResponseHeaderModification,HTTPRoutePortRedirect,HTTPRouteHostRewrite,HTTPRouteSchemeRedirect,HTTPRoutePathRedirect,HTTPRouteHostRewrite,HTTPRoutePathRewrite,HTTPRouteRequestMirror
+CONFORMANCE_SUPPORTED_PROFILES ?= -conformance-profiles=GATEWAY-HTTP
+CONFORMANCE_SKIP_TESTS ?= -skip-tests=HTTPRouteServiceTypes
```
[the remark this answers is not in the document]
timflannagan: I added some color in the PR description, but I should've made an inline comment here. The tl;dr is that the HTTPRouteServiceTypes test was introduced as a "core" test in the 1.1 release and we need to skip it in order to get green conformance CI runs. As a result of skipping core tests, we're technically partially conformant with the 1.1 release and there's follow-up RFE work that will make gloo fully conformant in 1.18.

ci/kind/setup-kind.sh
```diff
@@ -17,6 +17,11 @@ JUST_KIND="${JUST_KIND:-false}"
 IMAGE_VARIANT="${IMAGE_VARIANT:-standard}"
 # If true, run extra steps to set up k8s gateway api conformance test environment
 CONFORMANCE="${CONFORMANCE:-false}"
+# The version of the k8s gateway api conformance tests to run. Requires CONFORMANCE=true
+CONFORMANCE_VERSION="${CONFORMANCE_VERSION:-v1.1.0}"
```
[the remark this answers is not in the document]
timflannagan: That's an interesting callout. I'll add that as a TODO to follow-up on this once we're closer to the finish line with the GW API 1.1 stream of work.

ci/kind/setup-kind.sh
timflannagan: Sure, I don't have a strong opinion here. I'll add a comment to address sam's concerns and update the command to use the long form kustomize flag.

.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
timflannagan: Updated in the last commit.

Makefile
timflannagan: I'll queue this up as a follow-up since I have other more pressing work that needs to be burned down.

Makefile
```diff
@@ -1204,35 +1204,26 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
...
+CONFORMANCE_REPORT_ARGS ?= -report-output=$(TEST_ASSET_DIR)/conformance/$(VERSION)-report.yaml -organization=solo.io -project=gloo-gateway -version=$(VERSION) -url=github.com/solo-io/gloo -contact=github.com/solo-io/gloo/issues/new/choose
+CONFORMANCE_ARGS := -gateway-class=gloo-gateway $(CONFORMANCE_SKIP_TESTS) $(CONFORMANCE_SUPPORTED_FEATURES) $(CONFORMANCE_SUPPORTED_PROFILES) $(CONFORMANCE_REPORT_ARGS)
 
 .PHONY: conformance ## Run the conformance test suite
 conformance: $(TEST_ASSET_DIR)/conformance/conformance_test.go
-	go test -ldflags=$(LDFLAGS) -run TestConformance -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS)
+	go test -mod=mod -ldflags=$(LDFLAGS) -tags conformance -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS)
```
timflannagan, line 1219: The `-mod=mod` test flag was needed to avoid the failures that were outlined in https://github.com/solo-io/gloo/actions/runs/11061690230. Running `make conformance` without it will result in a non-zero exit that indicates `go mod tidy` is needed.

Makefile
```diff
@@ -1204,35 +1204,26 @@ build-test-chart: ## Build the Helm chart and place it in the _test directory
...
 # Run only the specified conformance test. The name must correspond to the ShortName of one of the k8s gateway api
 # conformance tests.
 conformance-%: $(TEST_ASSET_DIR)/conformance/conformance_test.go
-	go test -ldflags=$(LDFLAGS) -run TestConformance -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS) \
+	go test -mod=mod -ldflags=$(LDFLAGS) -tags conformance -test.v $(TEST_ASSET_DIR)/conformance/... -args $(CONFORMANCE_ARGS) \
 	-run-test=$*
 
-.PHONY: conformance-experimental ## Run the extended conformance test suite
-conformance-experimental: CONFORMANCE_ARGS += -conformance-profiles=HTTP -report-output=$(TEST_ASSET_DIR)/conformance/$(VERSION)-report.yaml -organization=solo.io -project=gloo-gateway -version=$(VERSION) -url=github.com/solo-io/gloo -contact=github.com/solo-io/gloo/issues/new/choose
-conformance-experimental: $(TEST_ASSET_DIR)/conformance/conformance_test.go $(TEST_ASSET_DIR)/conformance/experimental_conformance_test.go
```
timflannagan, line 1230: This target is no longer relevant as upstream consolidated the standard and experimental suites into a single suite. Previously, the experimental suite was the only one that supported generating/output conformance YAML reports which requires providing additional CLI flags (e.g. --report-output, --conformance-profiles, etc.) that were added to the CONFORMANCE_ARGS. Now, the standard suite supports this functionality as well, so I've moved those additional conformance args to the normal CONFORMANCE_ARGS initialization.

jbohanon (approved):

PR #10125, opened 2024-09-27
[1.17] test/kubernetes/e2e: introduce profiles for tests to use recommended helm values

sam-heilbron:
Backport of https://github.com/solo-io/gloo/pull/10106

# Description

Introduce standard recommended helm values for installations, and update our end-to-end tests to use them

# Context
## Backport
I chose to backport this to ensure that our 1.17 and 1.18 branches stay as in sync as possible. We are regularly backporting improvements to 1.17, so having the tests be in sync will make the lives of developers a lot easier.

## Helm Profiles
The idea here is that there are different "profiles" who will install Gloo Gateway. Depending on the use-case, the recommended Helm values will vary. We don't need to make these profiles a first-class part of the product yet, but by defining them in tests, it puts us a step closer towards being able to introduce these in our CLI or other install tooling.

## Single Source of Truth Value
I believe there is a lot of value in restructuring our tests to ensure that we define our production recommendations in a single place. Below, I try to clarify.

### Align Recommendations with Tests
We can now have user-facing documentation that is tied directly to a single file, and ensure that what we recommend to users is exactly what our tests use

### Reduce Developer Mistakes
In this PR you can see the number of manifests that exist to support our tests. Many have almost identical configuration. However, as we introduce new recommended values, you must apply that to all files, and a developer may forget that. In addition, for new tests, it is easy for a developer to not have all the context and accidentally run tests against a non-standard set of Helm values. 
By defining our recommended values in a single file, and requiring a "profile" to be defined for all tests, it reduces the chance for developer mistakes.

### Focus Unique Test Values
For a given test installation, it is easier to see what is unique about that installation, since the file now only contains the non-standard values. 

## Interesting decisions
-

## Testing steps
- All existing tests continue to pass

## Notes for reviewers
-

# Checklist:


jbohanon (approved):

sheidkamp (approved):
Looks the same as the initial PR, minus the helm tests that aren't in this branch.

PR #10160, opened 2024-10-03
tests: Fix flaky TestWatchNamespaceSelector

davidjumani:
# Description

Fixes the flaky TestMatchLabels that could be caused by not removing the label on a namespace after the test.
The presence of the label on the namespace (from the prior test even though the namespace was deleted???) could be present which causes the test to flake.
I've run this test ~20 times without flakes so it appears to be stable now

# Checklist:

### DIFF
--- a/test/kubernetes/e2e/features/watch_namespace_selector/suite.go
+++ b/test/kubernetes/e2e/features/watch_namespace_selector/suite.go
@@ -59,6 +59,11 @@ func (s *testingSuite) testWatchNamespaceSelector(key, value string) {
 
 	// The VS defined in the random namespace should be translated
 	s.TestInstallation.Assertions.CurlEventuallyRespondsWithStatus(s.Ctx, "random/", http.StatusOK)
+
+	s.unwatchNamespace(key)
+
+	// Ensure CRs defined in non watched-namespaces are not translated
+	s.TestInstallation.Assertions.CurlConsistentlyRespondsWithStatus(s.Ctx, "random/", http.StatusNotFound)
 }
 
 func (s *testingSuite) TestUnwatchedNamespaceValidation() {
--- a/changelog/v1.18.0-beta25/fix-watch-ns.yaml
+++ b/changelog/v1.18.0-beta25/fix-watch-ns.yaml
@@ -1,0 +1,4 @@
+changelog:
+- type: NON_USER_FACING
+  description: Fixes the flaky TestMatchLabels that could be caused by not removing the label on a namespace after the test.
+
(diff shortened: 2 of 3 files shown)


test/kubernetes/e2e/features/watch_namespace_selector/suite.go
```diff
@@ -59,6 +59,11 @@ func (s *testingSuite) testWatchNamespaceSelector(key, value string) {
 
 	// The VS defined in the random namespace should be translated
 	s.TestInstallation.Assertions.CurlEventuallyRespondsWithStatus(s.Ctx, "random/", http.StatusOK)
+
+	s.unwatchNamespace(key)
```
sam-heilbron, line 63: Is this a "cleanup" step and therefore we should consider performing it in the AfterTest method, or in a deferred function so it is always executed?

sam-heilbron (approved):

test/kubernetes/e2e/features/watch_namespace_selector/suite.go
davidjumani: It's only for this specific test and it also serves as a way to verify that unlabelling a namespace leads to the resources defined in it to not be translated

Makefile
```diff
@@ -210,7 +210,7 @@ analyze:
 
 GINKGO_VERSION ?= $(shell echo $(shell go list -m github.com/onsi/ginkgo/v2) | cut -d' ' -f2)
 GINKGO_ENV ?= GOLANG_PROTOBUF_REGISTRATION_CONFLICT=ignore ACK_GINKGO_RC=true ACK_GINKGO_DEPRECATIONS=$(GINKGO_VERSION)
-GINKGO_FLAGS ?= -tags=purego --trace -progress -race --fail-fast -fail-on-pending --randomize-all --compilers=5
+GINKGO_FLAGS ?= -tags=purego --trace -progress -race --fail-fast -fail-on-pending --randomize-all --compilers=5 --flake-attempts=3
```
sam-heilbron, line 213: This will affect all times we invoke these tests, on PRs and nightlies. I'm good with this, though initially i had envisioned we would only add this for PRs, so that nightlies still give us a good signal. wdyt?

sam-heilbron (approved):

Makefile
davidjumani: Since I assume the plan is to eventually move away from ginkgo, this should be fine as long as we make strides to migrate tests

danehans:
Fixes #10160

davidjumani:
> @davidjumani do you mind adding Fixes #10160 to the PR description to close out the associated issue?

I'll close out the issue when this is backported to 1.17

jenshu (approved):

Makefile
jenshu, line 213: i'm in favor of reverting this change so we can catch flakes more easily

Issue #10058, opened 2024-09-17
TestK8sGateway/DirectResponse/TestInvalidOverlappingFilters


bewebi:
### Which tests failed?

TestK8sGateway/DirectResponse/TestInvalidOverlappingFilters

### Initial Investigation

https://github.com/solo-io/gloo/actions/runs/10877848909/job/30179850572

Full logs:
```
=== RUN   TestK8sGateway/DirectResponse/TestInvalidOverlappingFilters
gateway.gateway.networking.k8s.io/gw created
httproute.gateway.networking.k8s.io/httpbin created
directresponse.gateway.gloo.solo.io/test-1 created
directresponse.gateway.gloo.solo.io/test-2 created
pod.go:51: expected pod to have status Running, but it was Pending
want:
HttpResponse{StatusCode: 500, Body: , Headers: map[], Custom: <nil>}
stdout:

stderr:  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 10.96.146.136:8080...
* Connected to gloo-proxy-gw.default.svc.cluster.local (10.96.146.136) port 8080 (#0)
> GET / HTTP/1.1
> Host: www.example.com
> User-Agent: curl/7.83.1-DEV
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< date: Mon, 16 Sep 2024 06:24:22 GMT
< server: envoy
< content-length: 0
< 

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
* Connection #0 to host gloo-proxy-gw.default.svc.cluster.local left intact


success: &{StdErr:  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 10.96.146.136:8080...
* Connected to gloo-proxy-gw.default.svc.cluster.local (10.96.146.136) port 8080 (#0)
> GET / HTTP/1.1
> Host: www.example.com
> User-Agent: curl/7.83.1-DEV
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
[...]

Issue #10103, opened 2024-09-24
Not all kubernetes e2e tests are getting run on PRs


jenshu:
### Do you have a suggestion for code improvement or tracking existing technical debt? Please describe.

Our [kubernetes e2e tests](https://github.com/solo-io/gloo/tree/main/test/kubernetes/e2e) consist of a set of top-level test suites that run "test installations" that each tests one or more "features". On PRs, to save CI time we [load-balance tests](https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/load_balancing_tests.md) so that a subset of tests runs on each cluster. To do this, we define [regexes](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/.github/workflows/pr-kubernetes-tests.yaml#L57-L81) for which tests will run on each cluster. This usually involves splitting up the top-level suites (e.g. TestK8sGateway) between different clusters. The problem is that when we register new tests under a top-level suite (e.g. [here](https://github.com/solo-io/gloo/blob/f89eec59130f473fe66b519da91d1f4277a19f00/test/kubernetes/e2e/tests/k8s_gw_tests.go#L22)) we may forget to add the new tests to the GH workflow.

Note that on the nightly tests, we [run all tests](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/.github/workflows/nightly-tests.yaml#L66-L70) (i.e. with no regex), instead of splitting, so we don't run into this problem. Sometimes tests that we forgot to add to the PR workflow show failures in nightlies, but those are less visible unless we monitor them regularly.

As part of this issue, we should do the following:
- [x] fix the DirectResponse test, which we know to be failing, but currently only runs on nightlies https://github.com/solo-io/gloo/issues/10058
- [x] check if we are accidentally skipping any other kubernetes e2e tests in gloo, and add all the missing tests to the [test matrix](https://github.com/solo-io/gloo/blob/58c532b256b4f437b5f98b4480916a9a1f2eb028/.github/workflows/pr-kubernetes-tests.yaml#L57-L81) - https://github.com/solo-io/gloo/pull/10197
    - [x] ~do the same for solo-projects~ - only the OSS suites run on nightlies and not PRs
- [x] add to the gloo e2e test [README](https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/README.md) 
[...]

PR #10161, opened 2024-10-03
Fix DirectResponse e2e tests

jenshu:
# Description

This fixes 2 DirectResponse tests:
- TestInvalidOverlappingFilters - see https://github.com/solo-io/gloo/issues/10058. This was failing consistently due to a typo in the test itself (we were checking for the wrong status `Reason` string). It wasn't caught because the test only ran on nightlies, and we don't triage them as urgently as tests that run on PRs (since they don't hold up PRs from merging).
- TestInvalidMultipleRouteActions - this was flaking, due to race conditions in our controller when updating the HTTPRoute status. Sometimes we wouldn't have the latest copy of the HTTPRoute, and when we tried writing the new status, it would fail. To fix this, added retries to the HTTPRoute status writing (see proxy_syncer.go)

## Other testing changes

I noticed that sometimes the nightly tests would report failure on the `TestK8sGateway` suite even though no individual tests were marked as failed. When scrolling through the test logs you could see that one of the checks failed, but for some reason it didn't fail the test. E.g. see [this test run](https://github.com/solo-io/gloo/actions/runs/11099717687/job/30834386912) where there is an assertion error shown when running `TestInvalidMultipleRouteActions` however the test was reported as
```
--- PASS: TestK8sGateway/DirectResponse/TestInvalidMultipleRouteActions
```
and the suite as a whole was marked failed:
```
--- FAIL: TestK8sGateway
```
I'm not 100% certain but it may have been due to our use of `EventuallyWithT` in some of our tests. I replaced with a regular `Eventually` and it seems to properly mark failing tests as failed now.

## CI changes

The DirectResponse kubernetes e2e tests were previously not being run on PRs (see https://github.com/solo-io/gloo/issues/10103). Added them to the PR GH workflow.

## Testing steps

Run the kubernetes e2e tests locally or in CI and verify they pass.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/10058


jenshu:
blocked on https://github.com/solo-io/gloo/pull/10160

sam-heilbron (approved):

timflannagan (approved):

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -187,23 +190,40 @@ func (s *ProxySyncer) syncRouteStatus(ctx context.Context, rm reports.ReportMap)
...
 
-	for _, route := range rl.Items {
-		route := route // pike
-		if status := rm.BuildRouteStatus(ctx, route, s.controllerName); status != nil {
-			if !isHTTPRouteStatusEqual(&route.Status, status) {
-				route.Status = *status
-				if err := s.mgr.GetClient().Status().Update(ctx, &route); err != nil {
-					logger.Error(err)
+		for _, route := range rl.Items {
+			route := route // pike
```
andy-fong, line 205: nit: I don't know how big the list can get. This retry will redo the whole list even one failed to update but since it does a compare before update, it may not be that bad.

jenshu: true. i had to include the client.List call within the retry block since presumably the update failed due to not having the latest version of the resource

PR #10164, opened 2024-10-03
tests: Fix flaky Bootstrap Clients tests

davidjumani:
# Description

Increase the timeout to resolve the flaky Bootstrap Clients test. This increase is to give additional time for the new pod to come up and assume leadership.

# Checklist:


test/kube2e/gloo/bootstrap_clients_test.go
```diff
@@ -462,7 +458,7 @@ var _ = Describe("Bootstrap Clients", func() {
 				// specify a long enough interval for it to be accepted when the kube api server comes back up
 				helpers.EventuallyResourceAccepted(func() (resources.InputResource, error) {
 					return resourceClientset.VirtualServiceClient().Read(testHelper.InstallNamespace, testVS.Metadata.Name, skclients.ReadOpts{})
-				}, "120s", "10s")
+				}, "300s", "10s")
```
sam-heilbron, line 461: I don't think this is safe. In fact, 120s (existing value) concerns me. Status updates should occur fairly quickly, so if it takes a while, it means one of two things:
- the trigger that leads to a status update takes too long to be processed by our system (ie translation is slow)
- the trigger that leads to a status update is dropped by our system (ie we dont re-run translation for some events)

either way, a status not being updated to me is a signal that something could be broken in our product itself.

PR #10167, opened 2024-10-03
*: Add support for Kubernetes EndpointSlices

timflannagan:
Introduces initial support for the EndpointSlice API in the core gloo translator logic.

TODO: fill out description & add tests. Right now this is in barebones stage and want to get consensus before further investing in changes.

Open Questions:

- How to handle the legacy Endpoints API going forward? Introduce a mutually exclusive feature flag that enables/disables EndpointSlice functionality and promote that feature flag over time? The EndpointSlices API has been stable in core kube for a while now and most third-party controllers/workloads have been updated to avoid reading the legacy Endpoints API like kube documentation recommends.
- Do we need any logic for the targetRef field when iterating over EndpointSlices resources? The logic for handling Endpoints checks whether the targetRef is referencing a backing Pod and has some additional filtering/validation logic that determines whether the underlying endpoint will be added to EDS

Known Limitations:

- Gloo EDS will filter out any EndpointSlices that have endpoints with non-ready conditions. Users that are manually creating EndpointSlices (either by hand or via a third-party controller) will need to ensure that the endpoints are in a ready state before Gloo EDS will consider them for routing
- Support for the legacy Endpoints API may be removed in the future
- Requires that Upstream resources be present (either automatically via discovery, or manually in dense environments). This is a potential barrier of adoption for the GW API integration. Why do I need to create an Upstream CR for my HTTPRoute when it's referencing a Service that has a backend EndpointSlices that isn't managed by kube-controller-manager (or wherever the endpointslice controller is deployed)?

# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


projects/gloo/pkg/plugins/kubernetes/eds.go
```diff
@@ -319,113 +338,190 @@ func getServiceFromUpstreamSpec(spec *kubeplugin.UpstreamSpec, services []*corev
...
 			endpointsMap[key] = append(endpointsMap[key], &copyRef)
 			continue
 		}
 
+		// TODO(tim): can we introduce a feature flag that allows users to opt-in to
+		// using endpoint slices? we'd graduate this to a stable feature in the future.
+
+		// find each matching endpoint slice. we'll then use the address and port
+		// fields from the endpoint slice to create the endpoint(s) needs for EDS.
+		for _, eps := range endpointSlices {
```
yuval-k, line 404: endpoint slices can have duplications, so we may need to account for that. see:
https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints

danehans: You can find a reference implementation for how to perform this endpoint aggregation and deduplication as part of the `EndpointSliceCache` code within kube-proxy.

danehans:
>How to handle the legacy Endpoints API going forward?

Support both Endpoints and EndpointSlices and prefer EndpointSlices when available since this aligns with Kubernetes. You should be able to implement this by checking if EndpointSlices are available for the targetRef and falling back to Endpoints if not. We then deprecate Endpoints and remove the Endpoints code.

danehans:
>Do we need any logic for the targetRef field when iterating over EndpointSlices resources? The logic for handling Endpoints checks whether the targetRef is referencing a backing Pod and has some additional filtering/validation logic that determines whether the underlying endpoint will be added to EDS

I would expect `endpointslices.endpoints[].targetRef` to be handled in the same manner as Endpoints. If the additional logic that you mention was implemented to understand whether the `targetRef` is in a ready state, `endpointslices.endpoints[].conditions` can be used instead.

danehans:
>Gloo EDS will filter out any EndpointSlices that have endpoints with non-ready conditions. Users that are manually creating EndpointSlices (either by hand or via a third-party controller) will need to ensure that the endpoints are in a ready state before Gloo EDS will consider them for routing

Do we have a known use case to support user-defined EndpointSlices? If so, we should document ^. If not, then the controller should ensure the EndpointSlice was created by the Service controller using `ownerReferences`.

timflannagan:
See https://solo-io-corp.slack.com/archives/C03MFATU265/p1727991040023559 for some offline discussion about these changes. I think we'll throw this PR on ice and introduce EndpointSlice support for GW API integration only.

danehans:
>Requires that Upstream resources be present (either automatically via discovery, or manually in dense environments). This is a potential barrier of adoption for the GW API integration. Why do I need to create an Upstream CR for my HTTPRoute when it's referencing a Service that has a backend EndpointSlices that isn't managed by kube-controller-manager (or wherever the endpointslice controller is deployed)?

An Upstream resource should not be required for Gateway API. If an Upstream is needed as a transition strategy, the HTTPRoute controller could create the Upstream if it doesn't exist based on the `backendRef`. We could support non-Kubernetes backends in the future with Upstream as a `backendRef`.

danehans:
xref: https://github.com/solo-io/solo-projects/issues/6910

timflannagan:
Closing - going to throw this PR on ice for now.

PR #10179, opened 2024-10-04
tests: Append k8s version to artifacts uploaded and upload conformance test report only if run failed

davidjumani:
# Description

Upload conformance tests artefact only if the job has failed

# Context

The nightly runs are successful but fail because of a conflict when uploading artefacts with the same name run on different kubernetes versions. This ensures that artifacts are uploaded only if the test fails.
Also append the k8s version to the artifact so that all failures across k8s versions will be uploaded.

## Notes for reviewers

# Checklist:


.github/workflows/composite-actions/kube-gateway-api-conformance-tests/action.yaml
```diff
@@ -59,10 +59,10 @@ runs:
...
     run: kubectl -n gloo-system  get events --sort-by='{.lastTimestamp}'
   - name: Upload reports
-    if: ${{ success() }}
+    if: ${{ failure() }}
     uses: actions/upload-artifact@v4
     with:
       # Name of the path to upload. The VERSION variable refers to the Makefile
       # VERSION variable.
-      name: conformance-gloo-gateway-report
+      name: conformance-gloo-gateway-report@k8s${{ matrix.kube-version.kubectl }}
```
davidjumani, line 67: Fixes runs marked as fail even though the tests pass
https://github.com/solo-io/gloo/actions/runs/11156177944/job/31008336270#step:3:3348

jenshu (approved):

sheidkamp (approved):

Issue #10145, opened 2024-10-02
Update datadog tracing config api to match envoy 1.31


andy-fong:
### Gloo Edge Product
Open Source

### Gloo Edge Version
1.17

### Is your feature request related to a problem? Please describe.
This is in preparation for upgrading envoy to 1.31.x because of the following change:

```
tracing/datadog: Disabled remote configuration by default. To enable this feature, use the [remote_config](https://www.envoyproxy.io/docs/envoy/v1.31.0/api-v3/config/trace/v3/datadog.proto#envoy-v3-api-field-config-trace-v3-datadogconfig-remote-config) field.
```

Here is the envoy documentation for [remote_config](https://www.envoyproxy.io/docs/envoy/v1.31.0/api-v3/config/trace/v3/datadog.proto#envoy-v3-api-field-config-trace-v3-datadogconfig-remote-config)

We are also missing the `collector_hosename` field which is not related but might as well add that with this update.

Details:
remote_config poll the datadog site for config changes to the agent. This was enabled before and according to the ticket can only be changed through ENV var inside dd-trace-cpp. We have a [guide](https://docs.solo.io/gloo-edge/latest/guides/integrations/datadog/) for setting up datadog but is not using the datadog envoy extension. Instead, we use the datadog kubenetes integration which install a datadog agent to scrape metrics endpoints. So, this does not affect our datadog guide but if some customers are using the datadog extension directly (which we do support), we will need to add the `remote_config` field and potentially default that to `{}` (the present of the field even empty means enabled) to preserve the existing behavior.

### Describe the solution you'd like
For a customer using the envoy datadog extension, we should maintain the default behavior but defaulting the `remote_config` field to `{}`. The present of the field (even with an empty object) means the feature is enabled.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1206681874205238/1208513389197261) by [Unito](https://www.unito.io)

andy-fong:
This might be a potential blocking to envoy 1.31 upgrade

andy-fong:
Currently, we use the exact same proto file from envoy go-control-plane for our api and the envoy config. The problem is that if we have a default value to enable remote_config (`remote_config: {}`), customer won't be able to disable it unless we add a `disable` bool field. So, need input for this.

nfuden:
If we are letting our api diverge should we consider just migrating remote_config in ours to a boolwrapper? or is the disabled a cleaner thing so that we can keep the upstream definition as is and just augment side by side?

[...]

PR #10191, opened 2024-10-08
Update datadog tracer config proto and preserve remote_config default behavior for envoy v1.31 (#10145)

andy-fong:
# Description

- Updated datadog tracer config proto to match envoy v1.31
- Updated Makefile to pull in envoy-gloo 1.31.2-patch1
- Updated ratelimit e2e test to not expect the `x-ratelimit-reset` header as it's no longer set in envoy v1.31 (reversible with runtime guard [envoy.reloadable_features.no_timer_based_rate_limit_token_bucket](https://www.envoyproxy.io/docs/envoy/v1.31.0/version_history/v1.31/v1.31.0#july-19-2024:~:text=local_ratelimit%3A%20Updated%20the,to%20false.)

## API changes

Added fields from envoy 1.31:
- collector_hostname
- remote_config

For remote_config, we also added a `disabled` field which is not on the envoy side

## CI changes

None because this plugin only has unit test and does not have e2e test, the tests would pass even with old envoy (because we already imported the new go-control-plane some time ago that already has the new fields)

## Docs changes

# Context

Before envoy 1.31, the remote_config functionality is on by default and cannot be turned off. In order to maintain the behavior for potential customers that might rely on this, we enable the feature by default if it's
not explicitly disable. Customers who are using datadog tracer would not need to change anything when upgrading to gloo/gloo-ee 1.18.

slack-conversation: https://solo-io-corp.slack.com/archives/G01EERAK3KJ/p1728052383874909

## Testing steps

Only rely on unit test on this one because we do not have a datadog setup to actually test this e2e.

## Notes for reviewers

# Checklist:


test/e2e/local_ratelimit_test.go
```diff
@@ -57,14 +57,13 @@ var _ = Describe("Local Rate Limit", func() {
 		expectNotRateLimitedWithOutXRateLimitHeader = func() {
 			response := expectNotRateLimited()
 			// Since the values of the x-rate-limit headers change with time, we only check the presence of these header keys and not match their value
-			ExpectWithOffset(1, response).ToNot(matchers.ContainHeaderKeys([]string{"x-ratelimit-reset", "x-ratelimit-limit", "x-ratelimit-remaining"}),
+			ExpectWithOffset(1, response).ToNot(matchers.ContainHeaderKeys([]string{"x-ratelimit-limit", "x-ratelimit-remaining"}),
```
andy-fong, line 60: envoy 1.31 switch the local ratelimit implementation to AtomicBucket which is no longer timer based, so the `x-ratelimit-reset` is no longer a thing and will not send out that header anymore. Will call this out in docs upgrade faq and the way to revert back to the timer-based bucket implementation.

pkg/utils/api_conversion/trace.go
```diff
@@ -29,9 +29,27 @@ var (
...
-		CollectorCluster: clusterName,
-		ServiceName:      glooDatadogConfig.GetServiceName().GetValue(),
+		CollectorCluster:  clusterName,
+		ServiceName:       glooDatadogConfig.GetServiceName().GetValue(),
+		CollectorHostname: glooDatadogConfig.GetCollectorHostname(),
 	}
+
+	remoteConfig := glooDatadogConfig.GetRemoteConfig()
+	if remoteConfig == nil {
+		// An empty RemoteConfig object on the envoy side means enabling RemoteConfig
```
sam-heilbron, line 39: This is a super helpful comment!

pkg/utils/api_conversion/trace.go
```diff
@@ -29,9 +29,27 @@ var (
...
+		// and not aware that they need to set this object
+		envoyDatadogConfig.RemoteConfig = &envoytrace.DatadogRemoteConfig{}
+	} else if !remoteConfig.GetDisabled().GetValue() {
+		// This Disabled field does not exist on the envoy side. We added it because our default
+		// is to enable when remoteConfig is not set to maintain backward compatibility. So, we need
+		// this field to disable if desire.
+		envoyDatadogConfig.RemoteConfig = &envoytrace.DatadogRemoteConfig{
+			PollingInterval: glooDatadogConfig.GetRemoteConfig().GetPollingInterval(),
+		}
+	} // Leaving RemoteConfig not set means it's disabled on the envoy side.
```
sam-heilbron, line 51: I like this callout. I wonder if we'd benefit from just setting the envoy value explicitly to disabled, instead of relying on the default behavior. In my view, this ensures that any changes to the default behavior won't bite us.

projects/gloo/api/external/envoy/config/trace/v3/datadog.proto
```diff
@@ -19,6 +20,18 @@ option (solo.io.udpa.annotations.file_status).package_version_status = ACTIVE;
...
 // [#protodoc-title: Datadog tracer]
 
+// Configuration for the Remote Configuration feature.
+message DatadogRemoteConfig {
+  // Frequency at which new configuration updates are queried.
+  // If no value is provided, the default value is delegated to the Datadog tracing library.
+  google.protobuf.Duration polling_interval = 1;
+
+  // Disabled remote config.
+  // This field does not exist in envoy's config but allow us to preserve the default behavior
```
sam-heilbron, line 30: nit: Since this is our API, and we are exposing this specifically for backwards compatibility, should we clarify what the default behavior is?

test/e2e/local_ratelimit_test.go
```diff
@@ -57,14 +57,13 @@ var _ = Describe("Local Rate Limit", func() {
...
 				"x-ratelimit headers should not be present for non rate limited requests")
 		}
 
 		expectNotRateLimitedWithXRateLimitHeader = func() {
 			response := expectNotRateLimited()
-			// Since the x-ratelimit-reset header value changes with time, we only check the presence of this header key and not match its value
-			ExpectWithOffset(2, response).To(matchers.ContainHeaderKeys([]string{"x-ratelimit-reset", "x-ratelimit-limit", "x-ratelimit-remaining"}),
+			ExpectWithOffset(2, response).To(matchers.ContainHeaderKeys([]string{"x-ratelimit-limit", "x-ratelimit-remaining"}),
```
sam-heilbron, line 66: I think the above comment about why the reset header will not be present is really useful, and I wonder if it could be even more effective if it were placed in code. By that I mean, what if we added an assertion to confirm that the reset header is not contained in the response headers?

sam-heilbron (approved):
LGTM!

Confirmed in standup that this PR should not merge until a downstream PR to pull in this dependency is passing CI

pkg/utils/api_conversion/trace.go
andy-fong: > I like this callout. I wonder if we'd benefit from just setting the envoy value explicitly to disabled, instead of relying on the default behavior. In my view, this ensures that any changes to the default behavior won't bite us.

disabled does not exist on the envoy side by choice base on PR comments

projects/gloo/api/external/envoy/config/trace/v3/datadog.proto
andy-fong: If have have other updates I need to push, I will clarify that. However, in general, because all the proto field default to false, 0 or "". The implied default behavior is `disabled=false`. That's the reason to pick `disabled` as the field instead of `enabled` (In general, I really don't like anything that's double negative to mean positive coming from the C++ world but I was taught that in go and protobuf, rely on the default value instead.)

pkg/utils/api_conversion/trace.go
sam-heilbron: ah my bad, totally forgot that.

test/e2e/local_ratelimit_test.go
andy-fong: I should put the comment back but just remove the x-ratelimit-reset part. In general, do we ever do tests with the runtime guard option because we will call out in the docs if customer needs this back, they can set the runtime guard setting (for the next 6 months before envoy remove it). So, technically, we are not testing a path that customer might need be using but that might be a whole new can of worms. It's the right things to do but not sure it's worth it as they will be throwaway code in 6 months.

nfuden (approved):

test/e2e/local_ratelimit_test.go
andy-fong: Since I got 2 approval and the Enterprise side tests a mostly passing (re-running some flakes), I will probably add the comment in another PR without possibly a changelog only update to call out the runtime guard.

PR #10193, opened 2024-10-09
KRT for GG

lgadban:
# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
+	xdsCache envoycache.SnapshotCache,
+	settings *gloov1.Settings,
+	syncerExtensions []syncer.TranslatorSyncerExtension,
+	legacySecretClient gloov1.SecretClient,
+	glooReporter reporter.StatusReporter,
 ) *ProxySyncer {
+	restCfg := kube.NewClientConfigForRestConfig(mgr.GetConfig())
+	client, err := kube.NewClient(restCfg, "")
+	if err != nil {
+		// TODO move this init somewhere we can handle the err
```
yuval-k, line 143: note: i plan to move this in one of my follow-up PR, as i need it earlier

projects/gateway2/controller/start.go
```diff
@@ -127,8 +130,14 @@ func Start(ctx context.Context, cfg StartConfig) error {
 		mgr,
 		k8sGwExtensions,
 		cfg.ProxyClient,
-		cfg.QueueStatusForProxies,
+		cfg.Translator,
+		cfg.Opts.ControlPlane.SnapshotCache,
+		cfg.Opts.Settings,
+		cfg.SyncerExtensions,
+		cfg.SecretClient,
+		cfg.StatusReporter,
```
jenshu, line 138: should we just pass in the whole `cfg` object?

projects/gateway2/examples/example-https-gw-and-route-with-attached-options.yaml
```diff
@@ -0,0 +1,114 @@
+kind: Gateway
```
jenshu, line 1: can all these examples be moved to kubernetes/e2e tests?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -39,12 +77,20 @@ type ProxySyncer struct {
 	k8sGwExtensions extensions.K8sGatewayExtensions
 
 	// proxyReconciler wraps the client that writes Proxy resources into an in-memory cache
-	// This cache is utilized by the debug.ProxyEndpointServer
-	proxyReconciler gloo_solo_io.ProxyReconciler
+	// This cache is utilized by RateLimit and the debug.ProxyEndpointServer
+	proxyReconciler gloov1.ProxyReconciler
+
+	proxyTranslator ProxyTranslator
+	istioClient     kube.Client
```
jenshu, line 84: does this mean GG cannot be installed without istio now?

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -74,19 +75,30 @@ func measureResource(ctx context.Context, resource string, length int) {
...
 
-	s.snapshotHistory.SetApiSnapshot(snap)
-	s.latestSnap = snap
+	// filter out Kube Gateway proxies
+	var (
+		edgeProxies v1.ProxyList
+		kubeProxies v1.ProxyList
+	)
+	for _, proxy := range snap.Proxies {
+		if proxy.GetMetadata().GetLabels()[utils.ProxyTypeKey] == utils.GlooEdgeProxyValue {
```
jenshu, line 90: nit: can use this func https://github.com/solo-io/gloo/blob/5cf76a9eabac70c655b8fdf8ffaf905ad269724d/projects/gloo/pkg/utils/translators.go#L46

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -151,7 +167,7 @@ func (s *translatorSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapsh
 
 		sanitizedSnapshot := s.sanitizer.SanitizeSnapshot(ctx, snap, xdsSnapshot, reports)
 		// if the snapshot is not consistent, make it so
-		xdsSnapshot.MakeConsistent()
+		sanitizedSnapshot.MakeConsistent()
```
jenshu, line 170: ahh - good catch. do you know if this was causing any bugs?

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -74,19 +75,30 @@ func measureResource(ctx context.Context, resource string, length int) {
...
+	var (
+		edgeProxies v1.ProxyList
+		kubeProxies v1.ProxyList
+	)
+	for _, proxy := range snap.Proxies {
+		if proxy.GetMetadata().GetLabels()[utils.ProxyTypeKey] == utils.GlooEdgeProxyValue {
+			edgeProxies = append(edgeProxies, proxy)
+		} else if proxy.GetMetadata().GetLabels()[utils.ProxyTypeKey] == utils.GatewayApiProxyValue {
+			kubeProxies = append(kubeProxies, proxy)
+		}
```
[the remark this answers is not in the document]
jenshu: we technically have other proxy types too (knative, ingress) so it's good to do the explicit check

projects/gateway2/proxy_syncer/proxy_syncer.go
yuval-k: no, this just uses istio as a library

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
-			return
+	kubeUpstreams, kubeUsClient := setupCollectionDynamic[glookubev1.Upstream](
+		ctx,
+		s.istioClient,
+		glookubev1.SchemeGroupVersion.WithResource("upstreams"),
+		krt.WithName("KubeUpstreams"),
+	)
+
+	// helper collection to map from the kube Upstream type to the gloov1.Upstream wrapper
+	glooUpstreams := krt.NewCollection(kubeUpstreams, func(kctx krt.HandlerContext, u *glookubev1.Upstream) **upstream {
```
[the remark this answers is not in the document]
lgadban: This was done to work around the generic types in the transformation functions, see: https://github.com/istio/istio/blob/51903838fc216411efd9ad147ed3cfe0c6e258e9/pkg/kube/krt/core.go#L144-L145
i.e. if a single-pointer type is returned here, the collection is effectively `glooUpstreams krt.Collection[upstream]`

I am exploring if there is a simple way to get this to work without the double pointer

cc @stevenctl

projects/gloo/pkg/syncer/envoy_translator_syncer.go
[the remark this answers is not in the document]
yuval-k: makes sense; but how should we handle the "else" block then? should we cover all types here?

projects/gateway2/proxy_syncer/proxy_syncer.go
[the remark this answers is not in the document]
yuval-k: yes, you will get a  `krt.Collection[upstream]`. is that a problem? i think it will work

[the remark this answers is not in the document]
lgadban: fixed via dc4b2dbf8975b8be8ca0b2f98aaad08cb08ee6ba

projects/gateway2/controller/controller.go
```diff
@@ -334,6 +336,28 @@ func (c *controllerBuilder) watchDirectResponses(_ context.Context) error {
 		Complete(reconcile.Func(c.reconciler.ReconcileDirectResponses))
 }
 
+func (c *controllerBuilder) watchPods(ctx context.Context) error {
+	err := ctrl.NewControllerManagedBy(c.cfg.Mgr).
+		// WithEventFilter(predicate.GenerationChangedPredicate{}).
```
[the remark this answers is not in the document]
lgadban: pods and endpoints actually don't have generation metadata; this was commented out as it was a c/p from the other controllers

removing entirely in a follow up

projects/gateway2/controller/controller.go
```diff
@@ -334,6 +336,28 @@ func (c *controllerBuilder) watchDirectResponses(_ context.Context) error {
...
+		Complete(reconcile.Func(c.reconciler.ReconcilePods))
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *controllerBuilder) watchEndpoints(ctx context.Context) error {
+	err := ctrl.NewControllerManagedBy(c.cfg.Mgr).
+		// WithEventFilter(predicate.GenerationChangedPredicate{}).
```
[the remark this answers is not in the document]
lgadban: see comment above

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
+
+func (ep *glooEndpoint) Equals(in *glooEndpoint) bool {
+	return proto.Equal(ep, in)
+}
+
+func (ep *glooEndpoint) ResourceName() string {
+	return ep.Metadata.GetName() + "/" + ep.Metadata.GetNamespace()
+}
+
+var _ krt.ResourceNamer = &upstream{}
```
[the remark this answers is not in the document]
lgadban: fixed via dc4b2dbf8975b8be8ca0b2f98aaad08cb08ee6ba

projects/gloo/pkg/syncer/envoy_translator_syncer.go
lgadban: no bugs to my knowledge as the snapshot always seems to be created in consistent order but I still think it's worth fixing for correctness

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
+	kubeUpstreams, kubeUsClient := setupCollectionDynamic[glookubev1.Upstream](
+		ctx,
+		s.istioClient,
+		glookubev1.SchemeGroupVersion.WithResource("upstreams"),
+		krt.WithName("KubeUpstreams"),
+	)
+
+	// helper collection to map from the kube Upstream type to the gloov1.Upstream wrapper
+	glooUpstreams := krt.NewCollection(kubeUpstreams, func(kctx krt.HandlerContext, u *glookubev1.Upstream) **upstream {
+		// TODO: not cloning, this is already a copy from the underlying cache, right?!
```
howardjohn, line 295: Is there? I don't think there is in the typical istio flow.. maybe from something we are doing in gloo

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
-		applyPostTranslationPlugins(ctx, pluginRegistry, &gwplugins.PostTranslationContext{
-			TranslatedGateways: translatedGateways,
-		})
+		// TODO: handle garbage collection on status plugins
+		var proxiesWithReports []translatorutils.ProxyWithReports
+		proxiesWithReports = append(proxiesWithReports, snap.proxyWithReport)
+		applyStatusPlugins(ctx, proxiesWithReports, snap.pluginRegistry)
+	})
+
+	go s.istioClient.RunAndWait(ctx.Done())
```
howardjohn, line 373: I don't think you need to `go` this. Given you block on sync on L375 anyways

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -78,94 +124,271 @@ var kubeGatewayProxyLabels = map[string]string{
...
+
+	s.istioClient.WaitForCacheSync(
```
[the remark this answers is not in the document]
howardjohn: No but probably you need to check the KRT collections have synced. Istio example: https://github.com/istio/istio/blob/915ac2bf05dbc7a08ae61a8bc39fcd6ea1f3d11a/pilot/pkg/serviceregistry/kube/controller/ambient/ambientindex.go#L531-L536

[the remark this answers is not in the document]
howardjohn: Actually its plausible you might want to depending on the event handlers. Else you will get some incomplete state pushed to handlers. Istio's handle this ok, so they do `Register; Run; Wait`.

Actually thinking about this more, I am not so sure this is true in Istio :thinking:

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -174,10 +397,244 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
...
-			secretsWarmed = true
-			resyncProxies()
+			proxyTrigger.TriggerRecomputation()
+		}
+	}
+}
+
+// ripped from: projects/gloo/pkg/plugins/kubernetes/eds.go#newEndpointsWatcher(...)
+func buildEndpoints(
+	ctx context.Context,
```
howardjohn, line 409: use of context.Context is a code smell. Because it usually implies a deadline, which implies IO, which violates KRT.

This is only used for kubernetes.FilterEndpoints which ignores the ctx input anyways so its not wrong, just a red flag I guess. Maybe could be more clear by dropping it here, passing Background() to FilterEndpoints, and mention its unused?

howardjohn: ~This IS actually violated in `inMemUpstreams` though, which calls `ServiceToUpstream` which calls Create though~ actually it appears the naming there is just confusing. The usage of `ctx` there is for extracting Settings

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -174,10 +397,244 @@ func (s *ProxySyncer) Start(ctx context.Context) error {
...
+	rm := reports.NewReportMap()
+	r := reports.NewReporter(&rm)
+
+	var translatedGateways []gwplugins.TranslatedGateway
+	gatewayTranslator := s.k8sGwExtensions.GetTranslator(ctx, gw, pluginRegistry)
+	if gatewayTranslator == nil {
+		contextutils.LoggerFrom(ctx).Errorf("no translator found for Gateway %s (gatewayClass %s)", gw.Name, gw.Spec.GatewayClassName)
+		return nil
+	}
+	proxy := gatewayTranslator.TranslateProxy(ctx, gw, s.writeNamespace, r)
```
howardjohn, line 468: TranslateProxy calls GetRoutesForGateway which hits the API server (probably actually informer cache, but the point still remains). This is outside of the krt.Fetch which means it will miss changes.

This could be fine IFF `proxyTrigger.MarkDependant` is always called when a change is needed?

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,169 @@
...
+	if ctxWithTags, err := tag.New(ctx, tag.Insert(resourceNameKey, resource)); err == nil {
+		stats.Record(ctxWithTags, envoySnapshotOut.M(int64(length)))
+	}
+}
+
+// buildXdsSnapshot will translate from a gloov1.Proxy to xdsSnapshot using the supplied api snapshot.
+// This method also merges in reports from extension syncing. Note that extensions are NOT actually synced,
+// as use a NoOp snapshot when running the extension syncers; we only care about getting extensions errors and warnings
+// related to the Proxy being processed.
+// The actual syncing of the extension server's AND the status of the extension resources is handled by the legacy syncer.
```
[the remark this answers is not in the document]
lgadban: The only duplicate work being done in here is the extension processing which will also happen in the legacy syncer.
Example flow:
- kube `Gateway` changes so a kube gw gloov1.Proxy needs to be translated and synced
- the resulting gloov1.Proxy for it will be translated to xds here, and NoOp extension syncing will run so we can get correct reports
- the Proxy will be reconciled to the in-memory Proxy client
- legacy syncer will run but will not do anything with this gloov1.Proxy because it is filtered out
- extension syncers will run with this gloov1.Proxy in its snapshot (this is technically the duplicate work)
  - actual xds syncing of extension xds clients will happen
  - status reports for the extension resources will be written

So there's really limited duplication and it is unfortunately necessary, at least for now.

>can we replace the legacy one completely?
In theory we could, but this would mean basically converting to this new KRT-based model for the Edge APIs, which seems like a massive risk to destabalize/regress on a known-working component. Maybe eventually we will be willing to take this on.

I'm updating the commentary here to be more clear on what is happening.

projects/gateway2/examples/example-https-gw-and-route-with-attached-options.yaml
lgadban: yeah they probably can; at least for this PR I won't add a new one

projects/gloo/pkg/syncer/envoy_translator_syncer.go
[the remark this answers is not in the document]
lgadban: took a look, the other proxy types have their own 'gateway pod' equivalent still (e.g. https://github.com/solo-io/gloo/blob/law/xds-syncer-krt/projects/ingress/pkg/translator/translator_syncer.go#L99-L105) that will reconcile Proxies somewhere and expect `gloo` to handle it, so this syncer technically needs to handle these other proxy types.

Side note, the old non-gateway types reconcile to etcd still (https://github.com/solo-io/gloo/blob/law/xds-syncer-krt/projects/ingress/pkg/setup/setup_syncer.go#L67-L70) so we will break them once we finally deprecate Proxies in etcd (unless we update them to use the in-memory client)

Lastly, this code was clumsy and leftover from before I even explored KRT.

tl;dr -- I will filter out kube proxies only and leave all other proxies to be handled normally

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: >Maybe could be more clear by dropping it here, passing Background() to FilterEndpoints, and mention its unused?

Yeah great catch, will update!

lgadban: >This IS actually violated in inMemUpstreams though, which calls ServiceToUpstream which calls Create though actually it appears the naming there is just confusing. The usage of ctx there is for extracting Settings

Correct; that is some legacy code as well that I'm not sure we want to take on refactoring, at least not right now

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: Thanks for flagging this, I was admittedly confused about this myself.
I _think_ at one point I tried this and I was blocked indefinitely, but looking at the impl I don't see how that could be the case.
Will update this

projects/gateway2/proxy_syncer/proxy_syncer.go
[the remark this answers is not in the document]
lgadban: >No but probably you need to check the KRT collections have synced

This definitely seems like a current gap so I am going to add the collection's HasSynced to this Wait(...) --
LMK if that doesn't seems reasonable.

w.r.t. to
>do we want to move the call to register after the wait for cache?

It seems like there's very little benefit to not doing this, so I'll make this change as well.

Thanks!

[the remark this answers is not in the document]
lgadban: updated in 9eec3e22150587729c7ee56b3c994d84d2c63dd7

Note that I no longer wait for any of the informers but now wait for every single krt collection, rather than just the "final" collection; I'm not sure if it is overkill or if we can just wait for the derived collections?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,109 +103,531 @@ var kubeGatewayProxyLabels = map[string]string{
...
+		Proxy: proxy.proxy,
+		Reports: translatorutils.TranslationReports{
+			ProxyReport:     proxyReport,
+			ResourceReports: r,
+		},
+	}
+	envoySnap, ok := xdsSnapshot.(*xds.EnvoySnapshot)
+	if !ok {
+		// fixme
+	}
```
yuval-k, line 578: we can probably remove the `, ok` part so it panics here

projects/gloo/pkg/syncer/envoy_translator_syncer.go
nfuden: i thought we explicitly did this so that our length reports were good but that we dont send empty info to envoy?

test/kubernetes/e2e/features/virtualhost_options/vhost_opt_suite.go
```diff
@@ -285,6 +286,8 @@ func (s *testingSuite) TestOptionsMerge() {
 		defaults.KubeGatewayReporter,
 	)
 
+	time.Sleep(time.Second * 5)
```
nfuden, line 289:
```suggestion

```

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,109 +103,531 @@ var kubeGatewayProxyLabels = map[string]string{
...
 		}
+		return uss
+	}, krt.WithName("InMemoryUpstreams"))
+
+	finalUpstreams := krt.JoinCollection([]krt.Collection[upstream]{glooUpstreams, inMemUpstreams})
 
-		pluginRegistry := s.k8sGwExtensions.CreatePluginRegistry(ctx)
-		rm := reports.NewReportMap()
-		r := reports.NewReporter(&rm)
+	podClient := kclient.New[*corev1.Pod](s.istioClient)
```
stevenctl, line 298:
```suggestion
	podClient := kclient.NewFiltered[*corev1.Pod](s.istioClient, kclient.Filter{
		ObjectTransform: kube.StripPodUnusedFields,
	})
```

This gets rid of ManagedFields which is large and almost definitely unused

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,109 +103,531 @@ var kubeGatewayProxyLabels = map[string]string{
...
-			duration := stopwatch.Stop(ctx)
-			contextutils.LoggerFrom(ctx).Debugf("translated and wrote %d proxies in %s", len(proxies), duration.String())
-		}()
 
-		var gwl gwv1.GatewayList
-		err := s.mgr.GetClient().List(ctx, &gwl)
-		if err != nil {
-			// This should never happen, try again?
-			return
+	kubeUpstreams := setupCollectionDynamic[glookubev1.Upstream](
```
stevenctl, line 266: nit: the name to me indicates `Upstream_Kube` when it means the k8s resource. Would just `upstreams` be ok? Or `rawUpstreams` maybe?

projects/gateway2/proxy_syncer/proxy_syncer.go
stevenctl: I think that's the case

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,109 +103,531 @@ var kubeGatewayProxyLabels = map[string]string{
...
+		// the FilterEndpoints(...) call is being removed in a follow-up so we don't need to mess with it
+		context.Background(),
+		"gloo-system",
+		keps,
+		svcs,
+		pods,
+		upstreamSpecs,
+	)
+	if len(warns) > 0 || len(errs) > 0 {
+		// FIXME
```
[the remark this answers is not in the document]
stevenctl: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/discovery/discovery.go#L208

projects/gloo/pkg/syncer/envoy_translator_syncer.go
lgadban: I'm not sure -- since there's uncertainty I'm going to back this change out as I don't want to hold this PR up

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+) (cache.Snapshot, reporter.ResourceReports, *validation.ProxyReport) {
+	ctx = contextutils.WithLogger(ctx, "kube-gateway-xds-snapshot")
+	logger := contextutils.LoggerFrom(ctx)
+	snapHash := hashutils.MustHash(snap)
+	metaKey := xds.SnapshotCacheKey(proxy)
+	logger.Infof("build xds snapshot for proxy %v (%v upstreams, %v endpoints, %v secrets, %v artifacts, %v auth configs, %v rate limit configs)",
+		metaKey, len(snap.Upstreams), len(snap.Endpoints), len(snap.Secrets), len(snap.Artifacts), len(snap.AuthConfigs), len(snap.Ratelimitconfigs))
+	defer logger.Infof("end sync %v", snapHash)
+
+	proxyCtx := ctx
```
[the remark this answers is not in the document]
lgadban: this is a copy/paste from the edge syncer: https://github.com/solo-io/gloo/blob/law/xds-syncer-krt/projects/gloo/pkg/syncer/envoy_translator_syncer.go#L146-L150

projects/gateway2/proxy_syncer/proxy_syncer.go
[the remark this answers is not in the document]
lgadban: also for ref: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/kubernetes/eds.go#L212-L223

test/kubernetes/e2e/features/virtualhost_options/vhost_opt_suite.go
lgadban: this was a dangling change for locally running e2e tests, thanks!

however, i do think this a potentially (very) flaky test setup so we may need to add a proper wait here eventually

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+// This method returns the generated xdsSnapshot along with a combined report of proxy->xds translation and extension processing on the Proxy.
+// NOTE: Extensions are NOT actually synced here as use a NoOp snapshot when running the extension syncers.
+// The actual syncing of the extensions and the status of the extension resources (e.g. AuthConfigs, RLCs) is still handled by the legacy syncer.
+func (s *ProxyTranslator) buildXdsSnapshot(
+	ctx context.Context,
+	proxy *v1.Proxy,
+	snap *v1snap.ApiSnapshot,
+) (cache.Snapshot, reporter.ResourceReports, *validation.ProxyReport) {
+	ctx = contextutils.WithLogger(ctx, "kube-gateway-xds-snapshot")
+	logger := contextutils.LoggerFrom(ctx)
```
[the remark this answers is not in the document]
lgadban: good idea, thanks!

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+	snapHash := hashutils.MustHash(snap)
+	metaKey := xds.SnapshotCacheKey(proxy)
+	logger.Infof("build xds snapshot for proxy %v (%v upstreams, %v endpoints, %v secrets, %v artifacts, %v auth configs, %v rate limit configs)",
```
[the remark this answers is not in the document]
lgadban: yeah this makes sense; this was also a c/p from the edge syncer: https://github.com/solo-io/gloo/blob/law/xds-syncer-krt/projects/gloo/pkg/syncer/envoy_translator_syncer.go#L102-L103

I'll update in both places

test/kubernetes/e2e/features/virtualhost_options/vhost_opt_suite.go
nfuden: Fair enough. The way we generally have handled that in the past is by checking in an eventually. That way while there are magic wait times you also have a polling interval which you recheck to see if the environment is in the state that you need it to be in.

Helps make things quicker and more sure than a normal wait but are definitely more annoying to write

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,170 @@
...
+
+func init() {
+	_ = view.Register(envoySnapshotOutView)
+}
+
+func measureResource(ctx context.Context, resource string, length int) {
+	if ctxWithTags, err := tag.New(ctx, tag.Insert(resourceNameKey, resource)); err == nil {
+		stats.Record(ctxWithTags, envoySnapshotOut.M(int64(length)))
+	}
+}
```
nfuden, line 46:
```suggestion
```

New format no longer records metrics

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: Following up here, yes this is the case while we are in this first step of adopting krt.
i.e. the controller-runtime 'model' is still responsible for determining when a change to the Proxies is needed and will trigger recompute for everything needed (really, there is overlap, between what is krt.Fetched and what ctrl-rtime controllers are watching for -- so we get some slight duplicate work being done)

one of the next steps here is to bring the existing ctrl-rtime reconcilers/controllers into krt and properly track dependencies entirely

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: good call, fixed

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
lgadban: this was also c/p from the existing syncer and I still see it there in `main`: https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/syncer/envoy_translator_syncer.go#L38-L75

either way, I'll remove

test/kubernetes/e2e/features/virtualhost_options/vhost_opt_suite.go
lgadban: it is currently in an eventually but the order of manifest application matters so the sleep helped enforce the order.

either way, removing as it is not intended to be merged

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+	// // deep copy the reports so we can release the lock
+	// reports := make(reporter.ResourceReports, len(s.latestReports))
+	// for k, v := range s.latestReports {
+	// 	reports[k] = v
+	// }
+	// s.reportsLock.RUnlock()
+
+	logger := contextutils.LoggerFrom(ctx)
+	logger.Debugf("gloo reports to be written: %v", reports)
+	// if s.identity.IsLeader() {
```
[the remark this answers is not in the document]
lgadban: known problem with ggv2 unfortunately: https://github.com/solo-io/solo-projects/issues/6367
would rather not piecemeal it

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+	// snapHash := hashutils.MustHash(snap)
+	// TODO: add versions?
+	logger.Infof("begin kube gw sync for proxy %v (%v listeners, %v clusters, %v routes, %v endpoints)",
+		proxyKey, len(snap.Listeners.Items), len(snap.Clusters.Items), len(snap.Routes.Items), len(snap.Endpoints.Items))
+	// defer logger.Infof("end sync %v", snapHash)
+
+	// stringifying the snapshot may be an expensive operation, so we'd like to avoid building the large
+	// string if we're not even going to log it anyway
+	if contextutils.GetLogLevel() == zapcore.DebugLevel {
+		// logger.Debug(syncutil.StringifySnapshot(snap))
```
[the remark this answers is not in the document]
lgadban: commented out because the debug logs are a nightmare to parse, but adding back in! ty

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// if the snapshot is not consistent, make it so
+	snap.MakeConsistent()
+	s.xdsCache.SetSnapshot(proxyKey, snap)
+
+	return s.syncStatus(ctx, reports)
```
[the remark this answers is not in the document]
lgadban: the structure has changed quite a bit since this question, but as it stands now, no the status syncing will still be done async

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: This is a great call out, was written under assumptions from our prior usage of controller-runtime cache and solo-kit based clients which will automatically DeepCopy in some form or fashion

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,134 +106,645 @@ var kubeGatewayProxyLabels = map[string]string{
...
+			secrets,
+			finalUpstreams,
+			authConfigs,
+			rlConfigs,
+		)
+		return xdsSnap
+	})
+
+	// unused collection, needed for "side-effect" of proxy reconciliation
+	proxiesToReconcile := krt.NewManyFromNothing(func(kctx krt.HandlerContext) []glooProxy {
```
yuval-k, line 380: thinking about this, we want to output the list of proxies as a single entity.. so should this perhaps be `NewSingleton` that returns the list as a single object? (maybe `ProxyList`?)

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,134 +106,645 @@ var kubeGatewayProxyLabels = map[string]string{
...
+		proxies := krt.Fetch(kctx, glooProxies)
+		var proxyList gloov1.ProxyList
+		for _, p := range proxies {
+			proxyList = append(proxyList, p.proxy)
+		}
+		s.reconcileProxies(ctx, proxyList)
```
yuval-k, line 386: i don't remember if this does i/o or not? if so, should this be in `Register`?

lgadban: it _should not_ perform i/o as it only persists to the in-memory cache for Proxies. Some more context is added in the godoc for this method (https://github.com/solo-io/gloo/pull/10193/files#diff-c658efd653154e7eb520a5ea6c646cdd45af7ed0ac56224360affdf481a562c4R813-R823)

Theoretically though, if persistProxy was enabled, then this would indeed perform a write to etcd. While this feature is unsupported in ggv2 (and as we know, ideally removed ASAP) there's no guarantee it's off.

Lastly, what we saw from the previous set of work related to i/o is that we wouldn't want blocking i/o even in a handler (side note: would love @howardjohn 's thoughts on this finding) so if we wanted to be safe we would want to schedule the reconciliation to be async in a Register(...)

What do you think?

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: So that's effectively what we are doing here, but your suggestion is a much cleaner and more obvious approach, good call!

I'll make this change and it will also allow for whatever we decide as a result of https://github.com/solo-io/gloo/pull/10193#discussion_r1809412887 to be done easier.

projects/gateway2/proxy_syncer/proxy_syncer.go
yuval-k: ah makes sense. i think we have somewhere a "queue" that only retains the last value to be synced between go-routines (even though it's called queue, the depth is 1)

go.mod
```diff
@@ -6,14 +6,14 @@ go 1.23.1
...
+	github.com/Masterminds/semver/v3 v3.3.0
 	github.com/Netflix/go-expect v0.0.0-20180928190340-9d1f4485533b
 	github.com/avast/retry-go v2.4.3+incompatible
 	github.com/aws/aws-sdk-go v1.34.9
 	github.com/census-instrumentation/opencensus-proto v0.4.1
-	github.com/cncf/xds/go v0.0.0-20240423153145-555b57ec207b
+	github.com/cncf/xds/go v0.0.0-20240830210341-88aa3b3c978a
 	github.com/cratonica/2goarray v0.0.0-20190331194516-514510793eaa
-	github.com/envoyproxy/go-control-plane v0.13.0
+	github.com/envoyproxy/go-control-plane v0.13.1-0.20240823165802-4363a624d376
```
jenshu, line 16: what go-control-plane version is this? should it be added to the changelog?

go.mod
```diff
@@ -341,7 +369,10 @@ replace (
 	// remove this when we fix https://github.com/solo-io/solo-kit/issues/516
 	github.com/iancoleman/strcase => github.com/iancoleman/strcase v0.1.3
 
-	github.com/opencontainers/go-digest => github.com/opencontainers/go-digest v1.0.0-rc1
+	github.com/imdario/mergo => github.com/imdario/mergo v0.3.16
+
+	// github.com/opencontainers/go-digest => github.com/opencontainers/go-digest v1.0.0-rc1
```
jenshu, line 374: why the pin and removal of other pin?

timflannagan: AFAIK, this might be needed due to the istio.io/istio dependency being introduced here. See https://github.com/istio/istio/blob/master/go.mod#L5-L6 for why they're doing it. @lgadban is that assumption accurate? Can we remove the commented out replace pin in a follow-up to unblock this work?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -75,134 +107,665 @@ var kubeGatewayProxyLabels = map[string]string{
...
+
+	duration := stopwatch.Stop(ctx)
+	contextutils.LoggerFrom(ctx).Debugf("translated proxy %s/%s in %s", proxy.GetMetadata().GetNamespace(), proxy.GetMetadata().GetName(), duration.String())
 
-		s.queueStatusForProxies(ctx, proxies, &pluginRegistry, totalResyncs)
-		s.syncStatus(ctx, rm, gwl)
-		s.syncRouteStatus(ctx, rm)
-		s.reconcileProxies(ctx, proxies)
+	// TODO: these are likely unnecessary and should be removed!
+	applyPostTranslationPlugins(ctx, pluginRegistry, &gwplugins.PostTranslationContext{
```
jenshu, line 586: it's used for portal

projects/gloo/pkg/syncer/envoy_translator_syncer.go
```diff
@@ -189,11 +203,6 @@ func (s *translatorSyncer) syncEnvoy(ctx context.Context, snap *v1snap.ApiSnapsh
 		logger.Debugf("Full snapshot for proxy %v: %+v", proxy.GetMetadata().GetName(), xdsSnapshot)
 	}
 
-	// Call the callback with the proxies and reports
-	if s.onProxyTranslated != nil {
```
jenshu, line 193: was this callback never used?

jenshu (approved):

projects/gateway2/proxy_syncer/proxy_syncer.go
lgadban: technically it is used for portal, but the plugin doesn't actually use anything from the postTranslationContext, see https://github.com/solo-io/solo-projects/blob/main/projects/k8sgateway/pkg/translator/plugins/portal/portal_plugin.go#L463

projects/gloo/pkg/syncer/envoy_translator_syncer.go
lgadban: it was used by the ggv2 status syncer.
the flow was:

- list all kube gw `Gateways` and translate to a list of `Proxy` objects
- reconcile proxies
- legacy edge syncer gets an update and performs Proxy->xDS translation for the full list
  - as a result of this, we get a report per Proxy->xDS translation
- use this callback to "send" the Proxy + translation report to ggv2 status syncer component

It is no longer needed as the kube gw proxy_syncer is responsible for translation of `Proxy->xDS` and therefore has access to all of the reports it needs synchronously (no need for a callback from the edge syncer as it's now irrelevant)

PR #10205, opened 2024-10-11
aws_sts_cluster static cluster config for kubernetes gateway api [old]

jenshu:
# Description

As part of supporting [AWS Lambda with EKS ServiceAccounts](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/destination_types/aws_lambda/eks-service-accounts/) with the Kubernetes Gateway API, new fields `spec.kube.aws.enableServiceAccountCredentials` and `spec.kube.aws.stsCredentialsRegion` have been added to the GatewayParameters API. These values can also be set on the default GatewayParameters during install/upgrade using the existing Helm values `settings.aws.enableServiceAccountCredentials` and `settings.aws.stsCredentialsRegion`.

When `enableServiceAccountCredentials` is set, a `aws_sts_cluster` cluster will automatically be added to dynamically provisioned proxies, so that Envoy can reach AWS to assume the role needed to access Lambdas. `stsCredentialsRegion` can be used to specify the AWS STS endpoint (will default to the global endpoint `sts.amazonaws.com`)

## API changes

New fields added to GatewayParameters:
- `spec.kube.aws.enableServiceAccountCredentials`
- `spec.kube.aws.stsCredentialsRegion`

## Helm changes

The above fields will be populated in the default `gloo-gateway` GatewayParameters using the following Helm values (note: these are the same Helm values already used to configure Edge proxies, as documented in [AWS Lambda with EKS ServiceAccounts](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/destination_types/aws_lambda/eks-service-accounts/)
- `settings.aws.enableServiceAccountCredentials`
- `settings.aws.stsCredentialsRegion`

## Testing

The following tests were added:
- Helm unit tests showing the aws settings are passed to default GatewayParameters
- Deployer unit tests showing that the GatewayParameters values are merged and applied to the Envoy ConfigMap
- e2e deployer test showing that the aws values from GatewayParameters cause the expected cluster to get created in the Envoy ConfigMap

# Checklist:


projects/gateway2/api/v1alpha1/gateway_parameters_types.go
```diff
@@ -726,6 +738,33 @@ func (in *CustomLabel) GetKeyDelimiter() *string {
 	return in.KeyDelimiter
 }
 
+// Configuration for AWS integration.
+type Aws struct {
+	// Use ServiceAccount credentials to authenticate lambda requests.
+	//
+	// +kubebuilder:validation:Optional
+	EnableServiceAccountCredentials *bool `json:"enableServiceAccountCredentials,omitempty"`
```
EItanya, line 746: Why is this a `*bool`? Is the default false or true?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -542,6 +542,34 @@ data:
                         address: 127.0.0.1
                         port_value: 8234
         {{- end }} {{/* if $gateway.istio.enabled */}}
+        {{- if ($gateway.aws).enableServiceAccountCredentials }}
```
EItanya, line 545: Can we figure this out from code and then add it via a plugin similar to how the GCP cluster works?

jenshu: That _could_ be possible, but we would need the plugin to be aware of the GatewayParameters to retrieve the AWS settings (unless we want the aws settings to be global and only stored in Settings, rather than per-k8s Gateway?). Whereas we already have existing plumbing in place to move from user's helm values -> GatewayParameters -> deployer -> ConfigMap helm template. What is the advantage of doing it in a plugin rather than as a static cluster in the helm template?

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
jenshu: it's for merging GWPs. The default for the `gloo-gateway` GatewayClass would be determined by the `settings.aws*` helm values. Let's say that globally they set it to true, but wanted to set it to false for a particular Gateway. It would not be possible for a Gateway's GatewayParameters to override it with `false` unless it was a pointer so we could differentiate between false and unset

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
EItanya: Why does it need to be in the GWP at all, do we already have this information in the Upstream, or rather should we add it there instead?

This may be a bigger conversation than this PR, but I think that model may be simpler

projects/gateway2/api/v1alpha1/gateway_parameters_types.go
```diff
@@ -102,6 +102,11 @@ type KubernetesProxyConfig struct {
 	// +kubebuilder:validation:Optional
 	AiExtension *AiExtension `json:"aiExtension,omitempty"`
 
+	// Configuration for AWS integration.
```
nfuden, line 105: Is this for other aws stuff to or just aws lambda?

nfuden: Also should this be a nested behind a cloud providers? Or are we comfortable with a deprecation strategy in the future?

test/kubernetes/e2e/features/deployer/testdata/aws-gateway-parameters.yaml
nfuden, line 1: I dont understand the test we are doing here.

I see that we enable sts but I dont see where we are hitting an actual lambda to test that we are able to use the sts setup?

jenshu:
i don't think the current configuration via GWP will work because our aws plugin uses some values from Settings and that presumably needs to match values in the proxy configmap. if you override some values via the GWP on a Gateway, the aws config in the proxy configmap may not match what's in settings. so i think, that the deployer needs to use awsOptions directly from Settings, rather than making things further configurable via GWP

jenshu:
closing in favor of https://github.com/solo-io/gloo/pull/10248

PR #10206, opened 2024-10-12
Attempt to change http-echo image

ashishb-solo:
# Description

This is just a test to see if we can change the image used for our tests from kennship/http-echo to jmalloc/echo-server. See changelog for details.

CI passes successfully, so it doesn't seem like this change causes any harm. I was hoping that CI runs would be faster since this image is smaller than the previous one, but it does not appear to make any difference - compare to the times for CI to run on [this pull request](https://github.com/solo-io/gloo/actions/runs/11165938304/job/31041201811) as an example and you can see that the times to run do not change much at all.

However, this change does still provide an improvement in that the new image supports HTTP/2 while the other previous image does not:

```
ashish@laptop5 ~/go/src/github.com/solo-io/gloo/test/kubernetes/e2e/features/tracing/testdata $ kubectl -n curl exec -it curl -- curl --http2 http-echo.http-echo.svc.cluster.local:3000
curl: (52) Empty reply from server
command terminated with exit code 52

ashish@laptop5 ~/go/src/github.com/solo-io/gloo/test/kubernetes/e2e/features/tracing/testdata $ kubectl -n curl exec -it curl -- curl http-echo.http-echo.svc.cluster.local:3000
{"path":"/","headers":{"host":"http-echo.http-echo.svc.cluster.local:3000","user-agent":"curl/7.83.1-DEV","accept":"*/*"},"method":"GET","body":{},"fresh":false,"hostname":"http-echo.http-echo.svc.cluster.local","ip":"::ffff:10.244.0.230","ips":[],"protocol":"http","query":{},"subdomains":["svc","http-echo","http-echo"],"xhr":false}
```

## CI changes

Change http-echo image to use jmalloc/echo-server instead

## Docs changes

# Context

## Notes for reviewers

# Checklist:

### DIFF
--- a/test/kubernetes/e2e/defaults/testdata/http_echo.yaml
+++ b/test/kubernetes/e2e/defaults/testdata/http_echo.yaml
@@ -14,6 +14,7 @@ spec:
     app.kubernetes.io/name: http-echo
   ports:
     - protocol: TCP
+      targetPort: 8080
       port: 3000
 ---
 apiVersion: v1
@@ -26,9 +27,9 @@ metadata:
 spec:
   containers:
   - name: http-echo
-    image: kennship/http-echo@sha256:144322e8e96be2be6675dcf6e3ee15697c5d052d14d240e8914871a2a83990af
+    image: jmalloc/echo-server
     ports:
-      - containerPort: 3000
+      - containerPort: 8080
     resources:
       requests:
         cpu: "100m"
--- a/test/kubernetes/e2e/defaults/testdata/http_echo.yaml
+++ b/test/kubernetes/e2e/defaults/testdata/http_echo.yaml
@@ -27,7 +27,8 @@ metadata:
 spec:
   containers:
   - name: http-echo
-    image: jmalloc/echo-server@sha256:c87f80de8dbb976a59b228fc9ecf257e0574c6f760c2f1c5f05f64c7ac7fbc1e
+    # we avoid using checksums so that we automatically select amd64 or arm64
+    image: jmalloc/echo-server@v0.3.7
     ports:
       - containerPort: 8080
     resources:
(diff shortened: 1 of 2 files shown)


test/kubernetes/e2e/defaults/testdata/http_echo.yaml
```diff
@@ -26,9 +27,9 @@ metadata:
 spec:
   containers:
   - name: http-echo
-    image: kennship/http-echo@sha256:144322e8e96be2be6675dcf6e3ee15697c5d052d14d240e8914871a2a83990af
+    image: jmalloc/echo-server
```
ashishb-solo, line 30: we should probably use an image checksum here to be safe.
```suggestion
    image: jmalloc/echo-server@sha256:c87f80de8dbb976a59b228fc9ecf257e0574c6f760c2f1c5f05f64c7ac7fbc1e
```

test/kubernetes/e2e/defaults/testdata/http_echo.yaml
```diff
@@ -26,9 +27,9 @@ metadata:
 spec:
   containers:
   - name: http-echo
-    image: kennship/http-echo@sha256:144322e8e96be2be6675dcf6e3ee15697c5d052d14d240e8914871a2a83990af
+    image: jmalloc/echo-server@sha256:c87f80de8dbb976a59b228fc9ecf257e0574c6f760c2f1c5f05f64c7ac7fbc1e
```
ashishb-solo, line 30:
```suggestion
    # image checksum is for linux/amd64, but an arm64 image is also available
    image: jmalloc/echo-server@sha256:c87f80de8dbb976a59b228fc9ecf257e0574c6f760c2f1c5f05f64c7ac7fbc1e
```

sam-heilbron (approved):
I like the change! There are a number of other references to this image that we are removing still in the codebase. How should we handle removing those from the codebase?

In our docs we refer to hashicorp/http-echo, and in some of our code (https://github.com/solo-io/gloo/blob/c3337f7343a5b37ce16d7b8e3ffb0bf9e26a4e93/test/kube2e/util.go#L41) we refer to some other images.

I'm good with this change as is, just wanted to raise the potential for additional improvements in this area

PR #10212, opened 2024-10-17
conradhanson/18590 access token validation distributed claims

conradhanson:
# Description

This PR updates the `AuthConfig` proto to enable distributed claims resolution for the oauth access token validation config.

## API changes

To AuthConfig's `oauth.accessTokenValidation` API, added:
1. `Azure` config so that distributed claims can be resolved
2. `claimsToHeaders` to map JWT claims to downstream request headers

## Docs changes

# Context

## Notes for reviewers

# Checklist:


projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1091,6 +1091,13 @@ message AccessTokenValidation {
     // and the structure of the claim value will be preserved in the metadata struct.
     map<string, string> dynamic_metadata_from_claims = 7;
 
+    // A list of claims to be mapped from the JWT token received by ext-auth-service to an upstream destination
+    repeated ClaimToHeader claims_to_headers = 8;
+
+    // For apps in Microsoft Azure, configure Microsoft Entra ID as the OpenID Connect (OIDC) provider.
+    // This way, you can enable distributed claims and caching for when users are members of more than 200 groups.
+    Azure azure = 9;
```
sam-heilbron, line 1099: Our ExtAuth API is a bit convoluted in this file. https://github.com/solo-io/solo-projects/blob/main/devel/architecture/extauth/api.md#extauth-server-api details this a bit, and it calls out that this proto actually has 2 apis within it: the user-facing API (AUthConfig CR), the service-facing API (how gloo<->ext-auth-service communicate over xDS).

The main difference is that the Gloo control plane resolves secrets instead of passing the reference to the ext-auth-service. I think in https://github.com/solo-io/gloo/blob/bcf0e995597704b023096c1cb06ccd29abcba3a9/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto#L1093 we'll need to add the necessary configuration to ensure that our service-facing API support these new fields

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -629,6 +629,29 @@ message ClaimToHeader {
     bool append = 3;
 }
 
+// For apps in Microsoft Azure, configure Microsoft Entra ID as the OpenID Connect (OIDC) provider.
+// This way, you can enable distributed claims and caching for when users are members of more than 200 groups.
+message Azure {
```
sheidkamp, line 634: Should moving this get labelled a breaking change? This will change the names of exported types.

Eg: `func (x *OidcAuthorizationCode) GetAzure() *OidcAuthorizationCode_Azure {` to `func (x *OidcAuthorizationCode) GetAzure() *Azure {`

Also, a similar change of scope will need to be made in the internal API

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -629,6 +629,29 @@ message ClaimToHeader {
...
+    string client_id = 1;
+
+    // The tenant ID represents the MS Entra organization ID where the ExtAuthService app is registered.
+    // This tenant ID may or may not be the same as in the top level `OidcAuthorizationCodeConfig`,
+    // depending on how your Azure account is provisioned.
+    string tenant_id = 2;
+
+    // The client secret of the ExtAuthService app that is registered with MS Entra to communicate with the MS Graph API.
+    // The client secret data must be placed in a k8s secret under a key called 'client-secret'.
+    core.solo.io.ResourceRef client_secret = 3;
```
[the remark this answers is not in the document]
conradhanson: it is required to successfully resolve distributed claims. 
So if it is omitted (while client_id and/or tenant_id are set), then the server will return a translation error since all three need to be set to successfully translate.
If the client secret value is poorly formatted (it is a simple string) then the azure settings will be misconfigured, the distributed claims will fail to resolve since authn to the azure api will fail, and then a warning will be omitted in ext-auth-service logs with the specific error. No authz decision is made due to azure api calls. 

With the failure scenario said, I think it is important to note that usually these resolved claims are used to enforce group-based authz of users through a downstream OPA policy.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -629,6 +629,29 @@ message ClaimToHeader {
...
+
+    // Redis connection details to cache MS Entera claims.
+    // This way, you avoid performance issues of accessing the Microsoft Graph API too many times.
+    // Note that this setting does NOT turn on Redis caching for the user session.
+    // To turn on Redis user session caching, use the `userSessionConfig` field.
+    RedisOptions claims_caching_options = 4;
```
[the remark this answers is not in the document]
conradhanson: Redis is not a requirement, omitting redis simply means no claims can be cached and azure api calls will be made on every authz decision that specifies distributed claims config. 

I would opt on the side of that world does not currently exist. Qualcomm is the only customer i know of that is using this azure-based dist claims feature set (work done specifically for them). I think we would just be introducing premature complexity if we were to add that flexibility since adoption of this feature is already sparse at best.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -835,29 +858,6 @@ message OidcAuthorizationCode {
     // No-op, represents default OIDC behavior
     message Default {}
 
-    // For apps in Microsoft Azure, configure Microsoft Entra ID as the OpenID Connect (OIDC) provider.
```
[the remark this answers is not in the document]
conradhanson: From an external API point im pretty sure it is safe (CRD doesn't change so no customer is impacted), but from an internal API point it is a breaking change since the Azure proto is no longer nested. 

Do you think we need to be concerned with breaking changes to our internal api?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
conradhanson: I think the service-facing api already has the proper Azure config for resolving secrets on its own in general, [here exactly](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto#L1871). Would you agree? Or am I misunderstanding something? 

So the azure config was previously done for oidc authz code, and i am adding it to jwt access token validation, which required me to unnest the azure msg from the oidc authz code msg. I believe to support this on gloo side of things I need to do a similar thing to the internal api, correct? Or would we want to do that in a separate PR for testing purposes?

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp: I agree its a gray area. I think its fine as we're the only consumers of this functionality. Is this getting backported?

I'm not sure we need to make the internal changes now. Maybe an argument  if we're considering the changes breaking so we do all the breaking changes at once. I think if do we need to label them as "not implemented" if we do add them.

conradhanson: It is not getting backported to my knowledge. I will hold off on the internal changes unless there is a consensus.

sheidkamp:
Upgrade validation steps:

```
# Installed glooee
glooctl install gateway enterprise --license-key $GLOO_LICENSE_KEY

# Create oidc ac with azure reference

# in solo-projects branch, pull in dependencies and update requirements-template (version: 1.18.0-beta29-10212)
go get github.com/solo-io/solo-apis@v0.0.0-20241017164533-c713042e37fc
go get github.com/solo-io/gloo@extauth-gme-parity

# build and load images
CLUSTER_NAME=upgrade-test make kind-build-and-load-standard -B

# build test chart
make build-test-chart

# Upgrade via helm
helm upgrade -n gloo-system gloo _test/gloo-ee-1.0.0-sah1.tgz --version 1.0.0-sah1 --set-string license_key=$GLOO_LICENSE_KEY --set gloo-fed.enabled=false --set gloo-fed.glooFedApiserver.enable=false

# Translation succeeds!

# Manually downgrade extauth image to 1.17.3 - still translates!

# restore extauth image and downgrade gloo to 1.17.3 - still translates!

# but - still old CRDs? (does that matter? - don't think so because the existing crd did not change)
```

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
sheidkamp, line 1099: Another question on this... in the `OidcAuthorizationCode` message, we put the Azure config in a "Provider" `oneof`:

```
    oneof Provider {
        Default default = 26;
        Azure azure = 27;
    }
```

Do we need to/should we do something similar here?

conradhanson: yea that is a great suggestion, i'll push this change real quick

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1094,9 +1094,13 @@ message AccessTokenValidation {
...
-    // This way, you can enable distributed claims and caching for when users are members of more than 200 groups.
-    Azure azure = 9;
+    oneof Provider {
+        Default default = 9;
+        Azure azure = 10;
+    }
+
+    // No-op, represents default OIDC distributed claims behavior
+    message Default {}
```
sheidkamp, line 1103: While it would be nice to reuse the existing message, we shouldn't rename it, and having a random "default" message at the top level with no context seems worse.

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
```diff
@@ -1094,9 +1094,13 @@ message AccessTokenValidation {
     // A list of claims to be mapped from the JWT token received by ext-auth-service to an upstream destination
     repeated ClaimToHeader claims_to_headers = 8;
 
-    // For apps in Microsoft Azure, configure Microsoft Entra ID as the OpenID Connect (OIDC) provider.
-    // This way, you can enable distributed claims and caching for when users are members of more than 200 groups.
```
sheidkamp: Can we keep this comment around somewhere?

conradhanson: it is on the Azure message definition already, this was just duplicated. So it should still show

projects/gloo/api/v1/enterprise/options/extauth/v1/extauth.proto
conradhanson: so no `Default`? remove?

sheidkamp: No, I think we leave it as it is.

sheidkamp (approved):

sam-heilbron (approved):

PR #10225, opened 2024-10-22
prep endpoints for dest rules

yuval-k:
- read nodes in addition to pods to provide topology metadata
- add 2 collections:
  - endpoints - reads pod&node info and creates an EndpointsForUpstream IR
  - EnvoyEndpoints - converts the IR to envoy Resources.
- We need the 2 collections as in a follow-up we will convert the IR in different ways for different clients.
- resource hashing happens per resource, so no need to do expensive hashes when constructing envoy snapshot
- tilt - support installing metallb and multi helm values files to make it easy to run conformance tests

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/7036


projects/gateway2/controller/start.go
```diff
@@ -171,3 +179,14 @@ func Start(ctx context.Context, cfg StartConfig) error {
 
 	return mgr.Start(ctx)
 }
+
+func createKubeClient() istiokube.Client {
```
lgadban, line 183: redundant/unnecessary with https://github.com/solo-io/gloo/pull/10225/files#diff-2aba2a58d09c4e941016852249707f79e8cc39d1a3579999f8b5fc4ca39a2b59R233-R242 ?

projects/gloo/pkg/syncer/setup/setup_syncer.go
```diff
@@ -223,6 +230,17 @@ func getAddr(addr string) (*net.TCPAddr, error) {
 	return &net.TCPAddr{IP: ip, Port: port}, nil
 }
 
+func createKubeClient() istiokube.Client {
+	restCfg := istiokube.NewClientConfigForRestConfig(ctrl.GetConfigOrDie())
+	client, err := istiokube.NewClient(restCfg, "")
+	if err != nil {
+		// TODO move this init somewhere we can handle the err
+		panic(err)
+	}
```
lgadban, line 239: can we now return and handle this error in `Setup(...)` ?

projects/gateway2/krtcollections/pods.go
```diff
@@ -0,0 +1,121 @@
...
+	return c.name
+}
+func (c NodeMetadata) Equals(in NodeMetadata) bool {
+	return c.name == in.name && maps.Equal(c.labels, in.labels)
+}
+
+type LocalityPod struct {
+	krt.Named
+	Locality        PodLocality
+	PodLabels       map[string]string
```
lgadban, line 36: why do we need the raw pod labels?
maybe godoc here?

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,412 @@
...
+	Upstreams      krt.Collection[UpstreamWrapper]
+	Endpoints      krt.Collection[*corev1.Endpoints]
+	Pods           krt.Collection[krtcollections.LocalityPod]
+	EnableAutoMtls bool
+	Services       krt.Collection[*corev1.Service]
+}
+
+func NewGlooK8sEndpointInputs(settings *v1.Settings, istioClient kube.Client, pods krt.Collection[krtcollections.LocalityPod], services krt.Collection[*corev1.Service], finalUpstreams krt.Collection[UpstreamWrapper]) EndpointsInputs {
+	epClient := kclient.New[*corev1.Endpoints](istioClient)
+	kubeEndpoints := krt.WrapClient(epClient, krt.WithName("Endpoints"))
```
lgadban, line 43: we already have this client here: https://github.com/solo-io/gloo/pull/10225/files#diff-c658efd653154e7eb520a5ea6c646cdd45af7ed0ac56224360affdf481a562c4R363

any reason to not consolidate?

projects/gateway2/krtcollections/pods.go
```diff
@@ -0,0 +1,121 @@
...
+	labels map[string]string
+}
+
+type PodLocality struct {
+	Region  string
+	Zone    string
+	Subzone string
+}
+
+func (c NodeMetadata) ResourceName() string {
```
stevenctl, line 26: nit add comments or type asserts to indicate that adding new types will need to impl these

projects/gateway2/krtcollections/pods.go
```diff
@@ -0,0 +1,121 @@
...
+	return c.Named == in.Named && c.Locality == in.Locality && maps.Equal(c.PodLabels, in.PodLabels) && maps.Equal(c.AugmentedLabels, in.AugmentedLabels)
+}
+
+func newNodeCollection(istioClient kube.Client) krt.Collection[NodeMetadata] {
+	nodeClient := kclient.New[*corev1.Node](istioClient)
+	nodes := krt.WrapClient(nodeClient, krt.WithName("Nodes"))
+	return NewNodeMetadataCollection(nodes)
+}
+
+func NewNodeMetadataCollection(nodes krt.Collection[*corev1.Node]) krt.Collection[NodeMetadata] {
```
stevenctl, line 50: should this be unexported?

projects/gateway2/krtcollections/pods.go
```diff
@@ -0,0 +1,121 @@
...
+			labels: us.Labels,
+		}
+	})
+}
+
+func NewPodsCollection(ctx context.Context, istioClient kube.Client) krt.Collection[LocalityPod] {
+	podClient := kclient.New[*corev1.Pod](istioClient)
+	pods := krt.WrapClient(podClient, krt.WithName("Pods"))
+	nodes := newNodeCollection(istioClient)
+	pods.Synced().WaitUntilSynced(ctx.Done())
```
stevenctl, line 63: Should we be doing this wait here? I think we can wait on everything after doing the collection setup all at once.

projects/gateway2/krtcollections/pods.go
```diff
@@ -0,0 +1,121 @@
...
+				if region != "" {
+					labels[corev1.LabelTopologyRegion] = region
+				}
+				if zone != "" {
+					labels[corev1.LabelTopologyZone] = zone
+				}
+				if subzone != "" {
+					labels[label.TopologySubzone.Name] = subzone
+				}
+				//	labels[label.TopologyCluster.Name] = clusterID.String()
```
stevenctl, line 107: if we're an istio ingress, it may be good to respect these istio labels

projects/gateway2/proxy_syncer/cla.go
```diff
@@ -0,0 +1,70 @@
...
+		var l *envoy_config_core_v3.Locality
+		if loc != (krtcollections.PodLocality{}) {
+			l = &envoy_config_core_v3.Locality{
+				Region:  loc.Region,
+				Zone:    loc.Zone,
+				SubZone: loc.Subzone,
+			}
+		}
+
+		endpoints := []*envoy_config_endpoint_v3.LocalityLbEndpoints{{
```
stevenctl, line 56: Why the one-item slice and loop?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -241,20 +237,36 @@ func (ep *glooEndpoint) ResourceName() string {
...
+	proto.Message
+	interface {
+		GetMetadata() *core.Metadata
+	}
+}
 
-// upstream provides a keying function for Gloo's `v1.Upstream`
-type upstream struct {
-	*gloov1.Upstream
+type ResourceWrapper[T GlooResource] struct {
```
stevenctl, line 247: should probably move this type to `krtcollections` package

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,412 @@
...
+	enableAutoMtls := settings.GetGloo().GetIstioOptions().GetEnableAutoMtls().GetValue()
```
stevenctl, line 44: can this setting change after startup? we could theoretically make it a krt Singelton to get dynamic updates

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,412 @@
...
+func (c EndpointsForUpstream) ResourceName() string {
+	return c.UpstreamRef.String()
+}
+
+func (c EndpointsForUpstream) Equals(in EndpointsForUpstream) bool {
+	return c.UpstreamRef == in.UpstreamRef && c.lbEpsEqualityHash == in.lbEpsEqualityHash
+}
+
+func NewGlooK8sEndpoints(ctx context.Context, inputs EndpointsInputs) krt.Collection[EndpointsForUpstream] {
+	return krt.NewCollection(inputs.Upstreams, TransformUpstreamsBuilder(ctx, inputs), krt.WithName("K8sClusterLoadAssignment"))
```
stevenctl, line 116: krt.WithName seems like a slight misnomer; I don't know what debug info is like but it could be misleading to call these CLAs even if they're similar in concept.

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+func TransformUpstreamsBuilder(ctx context.Context, inputs EndpointsInputs) func(kctx krt.HandlerContext, us UpstreamWrapper) *EndpointsForUpstream {
+	augmentedPods := inputs.Pods
+	kubeEndpoints := inputs.Endpoints
```
stevenctl, line 121: is there any advantage to doing EndpointSlices?

stevenctl:
LGTM!

projects/gateway2/proxy_syncer/endpoints.go
yuval-k: mainly that i can copy paste the old code from gloo-edge v1 (this file is adapted from https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/plugins/kubernetes/eds.go); endpointslice can be a follow-up

projects/gateway2/proxy_syncer/cla.go
yuval-k: whops  - this was a real loop in my side branch. will fix

projects/gateway2/krtcollections/pods.go
yuval-k: yea, but the problem is that i don't have the clusterId..
left it commented out incase we add support knowing the clusterId in the futurez

projects/gateway2/proxy_syncer/endpoints.go
yuval-k: we should chat about that tomorrow - i like the approach of dynamically updating settings, but it might  collide with how gloo does this today. infact, it might already collide with. 

Today the whole ggv2 logic [starts](https://github.com/solo-io/gloo/blob/94d02de30eb9c150b6bd4be6daa6f3ecaafbedd2/projects/gloo/pkg/syncer/setup/setup_syncer.go#L898) from gloo's SetupSyncer.

That [setup-syncer](https://github.com/solo-io/gloo/blob/94d02de30eb9c150b6bd4be6daa6f3ecaafbedd2/projects/gloo/pkg/syncer/setup/setup_syncer.go#L227C23-L227C28) is driven from a different syncer](https://github.com/solo-io/gloo/blob/94d02de30eb9c150b6bd4be6daa6f3ecaafbedd2/pkg/utils/setuputils/setup_syncer.go#L53), that watches the settings.

whenever the settings change, it cancels the context for gloo's setup-syncer, and restarts it with the new settings. this is effectively an internal re-start.
we can choose to depart from this approach (i agree that we should), and that would require moving the ggv2 proxy syncer init as a [sibling](https://github.com/solo-io/gloo/blob/4538565e65911a0f7d684aa5ab66eccc055774d2/pkg/utils/setuputils/main_setup.go) to the settings syncer.

projects/gateway2/krtcollections/pods.go
yuval-k: yes, because in a follow-up PR i will wire this collection to the xDS server callbacks, that is initialized before we wait for the caches in the proxy syncer

projects/gateway2/krtcollections/pods.go
yuval-k: i don't remember why i did that.. removed them for now :)

projects/gateway2/krtcollections/pods.go
yuval-k: potentially yes; i'll might be using it for tests; will check

yuval-k: yea, i use it in endpoints_test.go

yuval-k:
last commit decouples ggv2 from gloo-edge with the intention of avoid the internal restart that happens on settings changes
Notes:
ec2 plugin not working in ggv2
k8s uds,fds not working in ggv2 - pretty sure that was always the case
gloo-edge failover plugin might not also work; also it doe IO which plugins should not do.

not dynamically set (need restart):
- istio integration enabled won't change in deployer
- xds host and port won't change in deployer
- proxy write namespace 
- QPS on settings doesn't apply for anything ggv2 (even before this PR)

projects/gateway2/proxy_syncer/endpoints.go
timflannagan: I'm wondering whether we should rip this logic out and only support EndpointSlices for GG v2 going forward. We can tackle that question in the follow-up PR though.

projects/gloo/pkg/plugins/kubernetes/plugin.go
```diff
@@ -68,26 +74,41 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *cl
...
 
-	lister := p.kubeCoreCache.NamespacedServiceLister(kube.Kube.GetServiceNamespace())
-	if lister == nil {
-		return errors.Errorf("Upstream %s references the service '%s' which has an invalid ServiceNamespace '%s'.",
-			upstreamRef.String(),
-			kube.Kube.GetServiceName(),
-			kube.Kube.GetServiceNamespace(),
-		)
-	}
+	if p.svcCollection != nil {
```
lgadban, line 77: nit: comment on why we branch and why krt exists here

projects/gloo/pkg/plugins/registry/registry.go
```diff
@@ -56,16 +56,47 @@ import (
...
 
-func Plugins(opts bootstrap.Opts) []plugins.Plugin {
+type PluginOpts struct {
+	Ctx                     context.Context
+	Secrets                 factory.ResourceClientFactory
+	SidecarOnGatewayEnabled bool
+	Consul                  bootstrap.Consul
+	KubeClient              k8skube.Interface
+	KubeCoreCache           corecache.KubeCoreCache
+	SvcCollection           krt.Collection[*corev1.Service]
```
lgadban, line 75: nit: godoc for SvcCollection saying only needed from ggv2 flow to provide services to k8s plugin without the need of legacy svc client

projects/discovery/pkg/fds/syncer/setup_syncer.go
```diff
@@ -24,15 +24,15 @@ type Extensions struct {
 }
 
 func NewSetupFunc() setuputils.SetupFunc {
-	return setup.NewSetupFuncWithRunAndExtensions(RunFDS, nil)
+	return setup.NewSetupFuncWithRunAndExtensions(RunFDS, nil, nil)
```
lgadban, line 27: doesn't this actually need BootstrapOpts? i'm missing how `nil` is okay here

projects/gateway2/extensions/extensions.go
```diff
@@ -26,12 +29,13 @@ type K8sGatewayExtensions interface {
...
 // K8sGatewayExtensionsFactoryParameters contains the parameters required to start Gloo K8s Gateway Extensions (including Translator Plugins)
 type K8sGatewayExtensionsFactoryParameters struct {
-	Mgr                     controllerruntime.Manager
-	AuthConfigClient        v1.AuthConfigClient
-	RouteOptionClient       gatewayv1.RouteOptionClient
-	VirtualHostOptionClient gatewayv1.VirtualHostOptionClient
-	StatusReporter          reporter.StatusReporter
-	KickXds                 func(ctx context.Context)
+	Mgr                         controllerruntime.Manager
+	IstioClient                 istiokube.Client
```
lgadban, line 33: this is future-proofing right? i.e. so extensions can create their own collections?

projects/gateway2/proxy_syncer/proxy_syncer.go
```diff
@@ -72,21 +72,26 @@ type ProxySyncer struct {
...
-	proxyReconciler gloov1.ProxyReconciler
-
 	proxyTranslator ProxyTranslator
 	istioClient     kube.Client
 
 	// secret client needed to use existing kube secret -> gloo secret converters
 	// the only actually use is to do client.NewResource() to get a gloov1.Secret
 	// we can/should probably break this dependency entirely relatively easily
-	legacySecretClient gloov1.SecretClient
+	legacySecretClient *kubesecret.ResourceClient
```
lgadban, line 81: nit: update comment as you broke the dependency :)

projects/discovery/pkg/fds/syncer/setup_syncer.go
yuval-k: i think FDS uses its own run function that won't use the bootstrap opts. i can double check

projects/gateway2/setup/ggv2setup.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func getInitialSettings(ctx context.Context, c istiokube.Client, nns types.NamespacedName) *glookubev1.Settings {
+	// get initial settings
+	logger := contextutils.LoggerFrom(ctx)
+	logger.Infof("getting initial settings. gvr: %v", settingsGVR)
+
+	i, err := c.Dynamic().Resource(settingsGVR).Namespace(nns.Namespace).Get(ctx, nns.Name, metav1.GetOptions{})
+	if err != nil {
+		logger.DPanicf("failed to get initial settings: %v", err)
```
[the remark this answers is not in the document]
yuval-k: we should make it crash, i believe that's what gloo-edge does

projects/gateway2/controller/start.go
```diff
@@ -114,60 +110,108 @@ func Start(ctx context.Context, cfg StartConfig) error {
...
+
+func (c *ControllerBuilder) Start(ctx context.Context) error {
+	logger := contextutils.LoggerFrom(ctx)
+	logger.Info("starting gateway controller")
+	xdsHost, xdsPort := c.cfg.SetupOpts.GetXdsAddress(ctx)
+	if xdsHost == "" {
+		return ctx.Err()
 	}
 
+	logger.Info("got xds address for deployer")
```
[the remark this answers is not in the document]
yuval-k: it's mainly here because GetXdsAddress is blocking, so though a log saying that we passed it will be useful. i'll add the address

projects/gateway2/krtcollections/wrapper.go
```diff
@@ -0,0 +1,38 @@
...
+	core "github.com/solo-io/solo-kit/pkg/api/v1/resources/core"
+	"google.golang.org/protobuf/proto"
+	"istio.io/istio/pkg/kube/krt"
+)
+
+type GlooResource interface {
+	proto.Message
+	interface {
+		GetMetadata() *core.Metadata
+	}
```
lgadban, line 13: this is unfamiliar to me, why a nested interface? wouldn't it be promoted either way?

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,341 @@
...
+
+func NewEndpointsForUpstream(us UpstreamWrapper, logger *zap.Logger) *EndpointsForUpstream {
+	clusterName := translator.UpstreamToClusterName(us.Inner.GetMetadata().Ref())
+	// start with a hash of the cluster name. technically we dont need it for krt, as we can compare the upstream name. but it helps later
+	// to compute the hash we present envoy with.
+	h := fnv.New64()
+	h.Write([]byte(clusterName))
+	lbEpsEqualityHash := h.Sum64()
+
+	// add the upstream hash to the clustername, so that if it changes the envoy cluster will become warm again.
```
lgadban, line 95: this comment is slightly confusing, if upstream name changes looks like the envoy cluster is created with a new name, so is this saying it will be a net new cluster and then go through warming?

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -0,0 +1,341 @@
...
+	return func(kctx krt.HandlerContext, us UpstreamWrapper) *EndpointsForUpstream {
+		var warnsToLog []string
+		defer func() {
+			for _, warn := range warnsToLog {
+				logger.Warn(warn)
+			}
+		}()
+		logger := logger.With(zap.String("upstream", us.Inner.GetMetadata().Ref().Key()))
+
+		logger.Debug("building endpoints")
```
lgadban, line 166: nit: Debugf() with name of upstream building endpoints for

lgadban (commented):
non-blocking comments; looks good to me!

projects/gateway2/krtcollections/wrapper.go
yuval-k: Not sure what you mean by promoted. I did this so that the generic can access the metadata, and i don't need anything else.

projects/gateway2/proxy_syncer/endpoints.go
yuval-k: this is a workaround to an envoy bug. when a cluster changes envoy re-warms it. if eds name stays the same, envoy won't get an update for the cluster, and it will have no endpoints

projects/gateway2/proxy_syncer/endpoints.go
yuval-k: see line above (`logger.With`)- all logs in this function will have the upstream name

projects/gateway2/proxy_syncer/endpoints.go
yuval-k: yes, i agree. but prefer to do it in a follow up

lgadban (approved):

projects/gateway2/proxy_syncer/endpoints.go
lgadban: wasn't paying close enough attention 👍

projects/gateway2/proxy_syncer/endpoints.go
lgadban: could useful to add that context in comment (in next pr) but doesn't seem majorly important

projects/gateway2/krtcollections/wrapper.go
lgadban: promotion as in the fields/methods on an anonymous inner-type will be 'promoted' and can be called directly on the parent; after looking at it again, what you are doing is clear so we can disregard either way!

stevenctl (approved):

PR #10226, opened 2024-10-22
Add option to control Envoy closing connections on host set change

ryanrolds:
# Description

Adds option that allows setting Envoy's `close_connections_on_host_set_change` to true. 

# Checklist:


davidjumani (commented):
Looks great!

changelog/v1.18.0-beta29/lb-close-connections-on-set-change.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/9505
+    description: >-
+      Allow setting the Envoy LB policy to close connections on set change.
```
sam-heilbron, line 5: nit:

Our changelogs are compiled into user-facing documentation here: https://docs.solo.io/gloo-edge/latest/reference/changelog/open_source/. As a result, I've found that when exposing a new API, it's helpful to callout specifically what resource/field we are exposing so that users are then empowered to find find that field in our API reference (https://docs.solo.io/gloo-edge/latest/reference/api/)

projects/gloo/api/v1/load_balancer.proto
```diff
@@ -107,4 +107,7 @@ message LoadBalancerConfig {
         // If not specified, the default is 10%.
         google.protobuf.DoubleValue min_weight_percent = 3;
     }
+
+    // If set to true, the load balancer will close connections when the host set changes.
```
sam-heilbron, line 111: Our users/customers look to us for clarity on the tradeoffs of enabling a given field, and under what circumstances they would want to do it. I think we can expand this comment to clarify some of those tradeoffs or reasons to use it. wdyt?

projects/gloo/pkg/plugins/loadbalancer/plugin_test.go
```diff
@@ -364,6 +364,37 @@ status: {}
...
+	It("should not set close connections on host set change", func() {
+		upstream.LoadBalancerConfig = &v1.LoadBalancerConfig{
+			CloseConnectionsOnHostSetChange: false,
+		}
+		err := plugin.ProcessUpstream(params, upstream, out)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(out.CommonLbConfig).To(BeNil())
+	})
+
+	It("should not set close connections on host set change w/ other options", func() {
```
sam-heilbron, line 376: I really like this test. I was about to comment above aboout the value of having a unit test that handles more of the options, since there is complicated logic to handle it in the plugin, and you beat me to it!

projects/gloo/pkg/plugins/loadbalancer/plugin.go
```diff
@@ -97,7 +97,9 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 		return nil
 	}
 
-	if cfg.GetHealthyPanicThreshold() != nil || cfg.GetUpdateMergeWindow() != nil || cfg.GetLocalityConfig() != nil {
+	if cfg.GetHealthyPanicThreshold() != nil || cfg.GetUpdateMergeWindow() != nil ||
+		cfg.GetLocalityConfig() != nil || cfg.GetCloseConnectionsOnHostSetChange() {
```
sam-heilbron, line 101: nit: At first I was concerned about this change, because in all other cases we are checking for nil. I think this is safe because we are saying, if the CloseConnectionsOnHostSetChange is false, we will leave it unset and defer to the envoy default behavior. It might be useful to add a comment clarifying why this guard is different than the others (perhaps is overkill and easily inferred)

sam-heilbron (commented):
Looking great!

There are a few other types of tests that we tend to add for changes like this, and there is some discretion about what makes sense.

We have integration tests (https://github.com/solo-io/gloo/blob/main/devel/testing/e2e-tests.md) which test the behavior of gloo/envoy.

We also have kubernetes e2e tests which replicate customer scenarios and validate the behavior inside a kubernetes cluster.

For plugin changes, I think it's good at a minimum that we add an integration test. It's also a good opportunity to see the system run end to end.

I don' think we need a kubernetes e2e test since the behavior we're testing isn't specific to kubernetes, but I'm happy to be convinced otherwise

projects/gloo/pkg/plugins/loadbalancer/plugin.go
ryanrolds: I added a comment that should make the purpose of the block clearer. The main difference between is the new a PB field isn't a pointer, which probably isn't worth calling out.

davidjumani (approved):

ryanrolds:
I chatted with @sam-heilbron about the integration tests. We agreed that integration tests are appropriate, even if we are mostly confirming that Envoy is working as expected. I'm working on tests that at a minimum will:
1. Stand up GG
2. Configure the upstream with the newly plumbed option
3. Add multiple Envoy instances
4. Establish multiple long-lived connections (protocol TBD, may use WS to match GH issue case)
5. Add an Envoy instance to the set and confirm that the established connections are closed

projects/gloo/api/v1/load_balancer.proto
nfuden: In addition we may want to call out here or in docs that this doesn't close long lived gRPC connections https://github.com/envoyproxy/envoy/issues/8867

test/e2e/loadbalancer_plugin_test.go
```diff
@@ -0,0 +1,158 @@
...
+		WithRoutePrefixMatcher(e2e.DefaultRouteName, "/endpoint").
+		WithRouteActionToUpstream(e2e.DefaultRouteName, upstream).
+		Build()
+
+	testContext.ResourcesToCreate().VirtualServices = v1.VirtualServiceList{
+		customVS,
+	}
+}
+
+var _ = FDescribe("Load Balancer Plugin", Label(), func() {
```
nfuden, line 35: Unfocus before merge

test/e2e/loadbalancer_plugin_test.go
```diff
@@ -0,0 +1,158 @@
...
+			Eventually(func(g Gomega) {
+				dump, err := testContext.EnvoyInstance().StructuredConfigDump()
+				g.Expect(err).NotTo(HaveOccurred())
+
+				dacs, err := findDynamicActiveClusters(dump)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(dacs).NotTo(BeEmpty())
+				g.Expect(dacs).To(HaveLen(1))
+
+				g.Expect(dacs[0].LbPolicy).To(Equal(envoy_config_cluster_v3.Cluster_MAGLEV))
```
nfuden, line 116: nit:
```suggestion
				g.Expect(dacs[0].LbPolicy).To(Equal(envoy_config_cluster_v3.Cluster_MAGLEV), dacs[0])
```
Might be a cleaner troubleshooting output

changelog/v1.18.0-beta29/lb-close-connections-on-set-change.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/gloo/issues/9505
+    description: >-
+      Allow setting the Envoy LB policy to close connections when the upstream host set changes.
+      This is useful when using Ring Hash or Maglev load balancing policies and helps ensure
+      rebalancing when recovering from disruptions. This feature is disabled by default.
```
nfuden, line 7: Should we mention the associated metric somewhere? upstream_connections_closed_on_host_set_change

test/e2e/loadbalancer_plugin_test.go
ryanrolds: Yes, I definitely will do that when I'm not longer working on the test.

ryanrolds: Thank you. I will make sure to include more detailed error descriptions.

projects/gloo/api/v1/load_balancer.proto
ryanrolds: Interesting, thank you for calling that out. I was about to discover that limitation as I was going to try using a long-lived gRPC connection to test the closing.

changelog/v1.18.0-beta29/lb-close-connections-on-set-change.yaml
ryanrolds: I will add it.

test/e2e/loadbalancer_plugin_test.go
ryanrolds: Thank you. I will make sure to include more detailed assertion failure descriptions.

projects/gloo/api/v1/load_balancer.proto
ryanrolds: I'm doing a quick tour of the handling of this flag in Envoy to better understand why gRPC isn't being closed but WebSockets would (it might not). It's possible plumbing option through (which I think we should) won't resolve their issue and we will find WebSockets has the same issues a gRPC.

ryanrolds: I traced the the code path down to the [logic for draining/closing connection pools ](https://github.com/envoyproxy/envoy/blob/b77f7b6cd04ac8136f22c353c67a1aab2ad8eb42/source/common/upstream/cluster_manager_impl.cc#L1938-L1982). My read is that HTTP connection pool is drained and TCP connection is closed. I'm unsure WS is going to drain or be closed. I'm going to implement the test and see. 

At the very least I'm doing to update the documentation this PR and make the HTTP vs TCP behavior clearer.

nfuden: We can also just link to envoy docs. 
Long lived connections that dont have active connections and no limits on timing wont die as it attempts graceful shutdown. Its in their rfc as well https://docs.google.com/document/d/1yX8qRDXfbcOqNwSpaptFk1ru-bY_yHDLLOVBuLaclR4/edit?tab=t.0#heading=h.cvkycohhpnph

sam-heilbron:
> @sam-heilbron, after looking into the gRPC issue @nfuden pointed out and deep diving into Envoy (see my onboarding channel), it looks like this option is less a "close" and more a "drain". What do you think about dropping the long-lived connection part of the integration tests and instead confirm that the cluster is being told to drain it's connections or not? I suspect trying to find a meaningful difference between how it handles long-lived connections with the option on/off and asserting the effect on the connections is not going to be useful (they likely are not going to be closed until they hit an idle/age timeout in both cases).

I'm for it! I this is a good example of how an integration test helped us uncover the actual behavior of the API.

ryanrolds:
I've created integrations tests that try to confirm the the hosts is drained, but I'm not seeing the number of listeners that I would expect. With the option on or off, sometimes there are draining listeners at the start of a test. Other times the test fails because it never sees a draining listener after adding an envoy instance with the option enabled.

I'm sure I'm misunderstanding something or using the testing harness incorrectly. i would appreciate someone with a better understanding of Envoy looking at this and telling me what I'm doing wrong.

I'm running `TEST_PKG=./test/e2e/... make test` multiple times to see the varying behavior locally.

I've also updated the documentation to better explain the what the feature does and does not do.

changelog/v1.18.0-beta29/lb-close-connections-on-set-change.yaml
ryanrolds: I looked for this metric and didn't see anything. Where are you seeing it? I was thinking it may help confirm the change in behavior when setting this option.

nfuden: I dont see it in latest envoy 🤷 maybe removed when they changed connection pool logic. All good then

ryanrolds:
After determining that we don't have good way to confirm Envoys behavior. I removed the failing tests and refined the changelog/docs.

nfuden (approved):

PR #10228, opened 2024-10-23
KRT syncer follow-up

lgadban:
# Description

`AuthConfigs` were not working as expected after being added to the snapshot due to missing metadata. This PR will fix the metadata to work correctly.

Also cleanup some logs and run the e2e tests with debug logging by default.


projects/gateway2/translator/gateway_translator.go
```diff
@@ -52,10 +54,11 @@ func (t *translator) TranslateProxy(
 	stopwatch.Start()
 	defer stopwatch.Stop(ctx)
 
+	ctx = contextutils.WithLogger(ctx, "k8s-gateway-translator")
+	logger := contextutils.LoggerFrom(ctx)
 	routesForGw, err := t.queries.GetRoutesForGateway(ctx, gateway)
 	if err != nil {
-		// TODO(ilackarms): fill in the specific error / validation
-		// reporter.Gateway(gateway).Err(err.Error())
+		logger.Errorf("err getting routes for gw '%s': %v", client.ObjectKeyFromObject(gateway), err)
```
davidjumani, line 61: Would we want to add this to the report as well ?

lgadban: for context this Report is is used to generate status for k8s GW resources, as expected in this case the `Gateway`.

For now we don't want to add this error to the report because there is no straightforward, corresponding error condition in the spec for Gateway status.

Theoretically we could report something here but would take some thinking on the best way to handle it, so I'll add the TODO back in

yuval-k (approved):

davidjumani (approved):

PR #10230, opened 2024-10-23
fix: Handle errors in ggv2 plugins when resource is not in a watched …

davidjumani:
…namespace

# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:

### DIFF
--- a/projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
+++ b/projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
+	"github.com/rotisserie/eris"
 	"github.com/solo-io/gloo/projects/gloo/pkg/xds"
 	"github.com/solo-io/go-utils/contextutils"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
@@ -133,9 +135,14 @@ func (p *plugin) ApplyStatusPlugin(ctx context.Context, statusCtx *plugins.Statu
 		}
 	}
 	routeOptionReport := make(reporter.ResourceReports)
+	var multierr *multierror.Error
 	for roKey, status := range p.legacyStatusCache {
 		// get the obj by namespacedName
-		roObj, _ := p.routeOptionClient.Read(roKey.Namespace, roKey.Name, clients.ReadOpts{Ctx: ctx})
+		roObj, err := p.routeOptionClient.Read(roKey.Namespace, roKey.Name, clients.ReadOpts{Ctx: ctx})
+		if err != nil {
+			multierr = multierror.Append(multierr, eris.Wrapf(err, "error reading RouteOption %s in namespace %s", roKey.Name, roKey.Namespace))
+			continue
+		}
 
 		// mark this object to be processed
 		routeOptionReport.Accept(roObj)
@@ -146,12 +153,13 @@ func (p *plugin) ApplyStatusPlugin(ctx context.Context, statusCtx *plugins.Statu
 		}
 
 		// actually write out the reports!
-		err := p.statusReporter.WriteReports(ctx, routeOptionReport, status.subresourceStatus)
+		err = p.statusReporter.WriteReports(ctx, routeOptionReport, status.subresourceStatus)
 		if err != nil {
-			return fmt.Errorf("error writing status report from RouteOptionPlugin: %w", err)
+			multierr = multierror.Append(multierr, fmt.Errorf("error writing status report from RouteOptionPlugin: %w", err))
+			continue
 		}
 	}
-	return nil
+	return multierr.ErrorOrNil()
 }
 
 // tracks the attachment of a RouteOption so we know which RouteOptions to report status for
--- a/projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
+++ b/projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/rotisserie/eris"
 	"github.com/solo-io/go-utils/contextutils"
 	"github.com/solo-io/solo-kit/pkg/api/v1/clients"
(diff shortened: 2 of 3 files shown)


sam-heilbron (dismissed):
I think any PR that fixes a segfault in the control plane must have sufficient testing to prove that it will not regress. I think we need to have an e2e test to adequately prove this out

davidjumani:
> I think any PR that fixes a segfault in the control plane must have sufficient testing to prove that it will not regress. I think we need to have an e2e test to adequately prove this out

Yeah agreed this needs testing. I've added unit tests to ensure that there is no npe and an error is returned

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -133,9 +135,14 @@ func (p *plugin) ApplyStatusPlugin(ctx context.Context, statusCtx *plugins.Statu
 		}
 	}
 	routeOptionReport := make(reporter.ResourceReports)
+	var multierr *multierror.Error
 	for roKey, status := range p.legacyStatusCache {
 		// get the obj by namespacedName
-		roObj, _ := p.routeOptionClient.Read(roKey.Namespace, roKey.Name, clients.ReadOpts{Ctx: ctx})
+		roObj, err := p.routeOptionClient.Read(roKey.Namespace, roKey.Name, clients.ReadOpts{Ctx: ctx})
+		if err != nil {
+			multierr = multierror.Append(multierr, eris.Wrapf(err, "error reading RouteOption %s in namespace %s", roKey.Name, roKey.Namespace))
```
npolshakova, line 143: [nit] Instead of inlining we could make a public error here to make it easier to reuse in the tests.

projects/gateway2/translator/plugins/routeoptions/route_options_plugin_test.go
```diff
@@ -259,6 +259,46 @@ var _ = Describe("RouteOptionsPlugin", func() {
 			})
 		})
 
+		When("There is an error reading the RouteOptions", func() {
+			It("errors out", func() {
+				deps := []client.Object{attachedRouteOption()}
```
npolshakova, line 264: This route option has a target ref: https://github.com/solo-io/gloo/blob/5ba215860f8f82cef28c6666ccbfc7ff19faf5b4/projects/gateway2/translator/plugins/routeoptions/route_options_plugin_test.go#L776 Is there a test with the HTTPRoute extension ref?

npolshakova (approved):

jmhbh (approved):

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin_test.go
```diff
@@ -194,6 +196,28 @@ var _ = Describe("VirtualHostOptions Plugin", func() {
 				}
 			})
 		})
+
+		When("There is an error reading the VirtualHostOptions", func() {
```
sam-heilbron, line 200: I like this test! I would argue it would be valuable to test not just that we error out, but that _other_ reports are correctly written. That is an important functional behavior we need to maintain

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin_test.go
```diff
@@ -194,6 +196,28 @@ var _ = Describe("VirtualHostOptions Plugin", func() {
...
+							Reports: translatorutils.TranslationReports{
+								ProxyReport:     &validation.ProxyReport{},
+								ResourceReports: reporter.ResourceReports{},
+							},
+						},
+					},
+				}
+
+				err := plugin.ApplyStatusPlugin(ctx, statusCtx)
+				Expect(err).To(HaveOccurred())
```
sam-heilbron, line 217: nit: Is this implicitly done by the assertion in the next line? (Should we just remove this)

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
```diff
@@ -214,13 +220,14 @@ func (p *plugin) ApplyStatusPlugin(ctx context.Context, statusCtx *plugins.Statu
...
+		err = p.statusReporter.WriteReports(ctx, virtualHostOptionReport, status.subresourceStatus)
 		if err != nil {
-			return eris.Wrap(err, "writing status report from VirtualHostOptionPlugin")
+			multierr = multierror.Append(multierr, eris.Wrap(err, "writing status report from VirtualHostOptionPlugin"))
+			continue
 		}
 
 	}
-	return nil
+	return multierr.ErrorOrNil()
```
sam-heilbron, line 230: nit: I think it could be useful to clarify in the function comment under what circumstances this function returns an error, and the impact on the control plane when one occurs

sam-heilbron (commented):
This issue occurs because we have code to implement status reporting on these objects, is that correct? If so, that means that we _could be_ susceptible to this when we implement status reporting on ListenerOptions and HttpListenerOptions. If so, I think we should capture that somehow so when future devs implement it, they don't forget this subtlety

projects/gateway2/translator/plugins/routeoptions/route_options_plugin_test.go
davidjumani: There are tests but none that test this particular scenario. I'm willing to add it although it might not be of much value since it would still follow the sad path that is being tested here

npolshakova: The targetRef case doesn't get processed if the route option is in a namespace not being watched though, right? Versus the extensionRef is processed, but it can't find the RouteOption because the namespace isn't watched.

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin_test.go
davidjumani: Right now the errors are only [logged](https://github.com/solo-io/gloo/blob/a01d3f41d45fb69b202ba8f3160dd375c6040159/projects/gateway2/proxy_syncer/proxy_syncer.go#L745). Discussed over [slack](https://solo-io-corp.slack.com/archives/C07Q05BDMRD/p1729789692496579?thread_ts=1729784649.663349&cid=C07Q05BDMRD) that status reporting will be tacked in another issue

projects/gateway2/translator/plugins/virtualhostoptions/virtualhost_options_plugin.go
davidjumani: I added comments on the return value but didn't mention anything about the control plane as it isn't germane to this method and should be handled in the caller

projects/gateway2/translator/plugins/routeoptions/route_options_plugin_test.go
davidjumani: Added a test with the HTTPRoute extension ref

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
```diff
@@ -31,6 +33,8 @@ import (
 var (
 	_ plugins.RoutePlugin  = &plugin{}
 	_ plugins.StatusPlugin = &plugin{}
+
+	ReadingRouteOptionErrStr = "error reading RouteOption"
```
npolshakova, line 37: [nit] I was thinking something more like:

```
func ReadingRouteOptionErrStr(name, ns string) error { return eris.Wrapf(err, "failed reading RouteOption %s in namespace %s", roKey.Name, roKey.Namespace) }
```

But this works!

npolshakova (approved):

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
davidjumani: I thought of this as well but since we're extracting the error, it could be a string that's easily reused even outside this method

PR #10238, opened 2024-10-25
gloo/api/options.proto: split options to separate files

sam-heilbron:
# Description
Decompose the `options.proto` file into separate files.

# Context
## Proto Best Practices
https://protobuf.dev/programming-guides/dos-donts/ outlines some best practices. There are a number of learnings that we could take away from that document and apply to the Gloo Gateway codebase. 

We try to apply [Do Define Message Types in Separate Files](https://protobuf.dev/programming-guides/dos-donts/#separate-files ) in this PR. 

As part of this work, I also went through the imports for these proto files, and reduced them to just the required import files. This should hopefully make it easier to understand the code and maintain them. 
_I have found in the past that code which appears unused can be scary to remove because you don't know the unknown consequences of it. Since I am aware that removing these unused imports is safe, I am attempting to help future devs_

## Docs User Experience
The main inspiration for this work is one of user experience. We define our APIs in proto files, and we have a lot of Protobuf messages. However, we have only a handful of user-facing Custom Resources. 

_The impact on users is that it can be challenging to navigate our docs and understand what top level resources exist and what is configurable_

To make it more challenging, we produce a single markdown file for each proto file. In our `options.proto` we had multiple top level custom resources defined in the same file. This meant that users would visit the markdown of our options.proto and be faced with all of our options APIs and get overwhelmed.

## Ideal Future Work
It would be preferable if each user-facing API had a single page to visit, and all the messages in that API were contained in that file.

A challenge is how we manage the docs generation and our proto maintenance. As explained above, proto files are intended to be split by message. Our docs currently have the 1:1 limitation of proto file -> markdown file. We would need a way to keep proto files small and distinct, and allow our docs to be composed of messages from multiple files.

See [the internal slack context](https://solo-io-corp.slack.com/archives/C07QCA4FG0M/p1729881695241449)

## Testing steps
- [x] Confirm that all tests still pass. These changes to protos are not breaking because the generated code remains in the same groups
- [x] Confirm the docs are generated to have all our our Options APIs (ListenerOptions, HttpListenerOptions, TcpListenerOptions, VirtualHostOptions, RouteOptions) in separate files. See https://github.com/solo-io/gloo/pull/10238#issuecomment-2438598684 for the generated docs

# Checklist:


sam-heilbron:
During this PR, I hit a test flake with the Kubernets Gateway API conformance tests. I created https://github.com/solo-io/gloo/issues/10240 to track this

nfuden (approved):

PR #10239, opened 2024-10-25
Add glooctl snapshot cmd

bleggett:
# Description

Adds a `glooctl proxy snapshot` cmd, which will dump everything the gateway has (stats, config, listeners, endpoints) into a timestamped zipfile, to aid diagnosis of gateway (or non-gateway) problems.

Some of these already exist piecemeal in the tool (config dump, stats dump) but this adds a few extra bobs and creates an artifact.

## Code changes

- Added `glooctl proxy snapshot`
- Changed `GetEnvoyAdminData` to take any k8s specifier - previously it always assumed (but did not say) you should give it the name of a deployment. Now, it will work with either a kube-style `deployment/<name>` specifier, or a raw podname. This was added for `glooctl proxy snapshot` but also affects `glooctl proxy dump` and `glooctl proxy stats`.

## Docs changes

TBD

# Context

## Interesting decisions

The existing code for `dump` and `stats` actually would always start hitting the port-forwarded endpoint before the port-forwarding was completed - which would always fail and spam spurious messages. This really needs to be fixed properly by the use of a real Golang kube client, but in the interim we can fix it with stdout scans.

# Checklist:

BOT NOTES: 
resolves https://github.com/solo-io/solo-projects/issues/7131


projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -37,20 +39,78 @@ func dumpCmd(opts *options.Options, optionsFunc ...cliutils.OptionsFunc) *cobra.
...
+//
+// Admin endpoint data will be fetched from `path` using the defined timeout.
 // Note that a `/` will be prepended to path if it does not exist.
-func GetEnvoyAdminData(ctx context.Context, proxyName, namespace, path string, timeout time.Duration) (string, error) {
+func GetEnvoyAdminData(ctx context.Context, proxySelector, namespace, path string, timeout time.Duration) (string, error) {
 	if !strings.HasPrefix(path, "/") {
 		path = "/" + path
 	}
 	adminPort := strconv.Itoa(int(defaults.EnvoyAdminPort))
+	// TODO	this should use a real Go kube client library someday
```
sam-heilbron, line 86: A while ago, I had added https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/portforward/portforwarder.go (and two corresponding implementations). I had tried to migrate some areas where it was used to the new implementation, but avoided making a widespread change. My hope was to see it work and then expand the usage. 
I think we should migrate this implementation to at least use the [Cli PortForwarder](https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/portforward/cli_portforwarder.go), what do you think?

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -108,23 +171,65 @@ func GetEnvoyCfgDump(opts *options.Options) (string, error) {
...
-		RunE: func(cmd *cobra.Command, args []string) error {
-			cfgDump, err := getEnvoyStatsDump(opts)
-			if err != nil {
-				return err
-			}
-			fmt.Printf("%v", cfgDump)
-			return nil
-		},
+func getEnvoyStatsDump(opts *options.Options) (string, error) {
+	return GetEnvoyAdminData(opts.Top.Ctx, opts.Proxy.Name, opts.Metadata.GetNamespace(), "/stats", 30*time.Second)
```
sam-heilbron, line 175: Historically we had a lot of duplicated code around kubernetes and envoy split betweeen our cli, control plane and tests. We've tried to extract some of the envoy particulars into https://github.com/solo-io/gloo/tree/main/pkg/utils/envoyutils. 

Most importantly, we have a CLI with a stats method: https://github.com/solo-io/gloo/blob/main/pkg/utils/envoyutils/admincli/client.go#L97.

Is it possible to use that? If not, could we find a way to capture/consolidate the envoy details in our envoyutils location, so that other areas of the code can benefit?

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -108,23 +171,65 @@ func GetEnvoyCfgDump(opts *options.Options) (string, error) {
...
+	defer proxyOutArchive.Close()
+
+	proxyName := opts.Proxy.Name
+	proxyNamespace := opts.Metadata.GetNamespace()
+	if proxyNamespace == "" {
+		proxyNamespace = defaults.GlooSystem
+	}
+    var errs []error
+
+	errs = append(errs, recordEnvoyAdminData(fileInArchive(proxyOutArchive, "config.log"), opts.Top.Ctx, "/config_dump", proxyName, proxyNamespace))
```
sam-heilbron, line 194: A while ago, I had explored something like this, and captured some progress in https://github.com/solo-io/gloo/pull/9365. It is by no means the right or only way to do it, I'm just sharing so you have the context.

In that, I did attempt to isolate the file specific utility into a shared utils, as it felt like something that could provide broader usage to the repository

sam-heilbron (commented):
Looking great! I think this tool will help the usage of capturing actionable information about the running cluster.
I shared some context about potential areas of the code that may solve parts of this already, or areas where we could expand our functionality for better re-use. 
I know this is just a draft, but I was hoping to share this to help save you some time in case you were unaware of those code areas. Happy to chat further if it helps!

projects/gloo/cli/pkg/cmd/gateway/dump.go
bleggett: ah perfect. yep!

projects/gloo/cli/pkg/cmd/gateway/dump.go
bleggett: Makes sense, ty!

bleggett:
> Deployer Single gwc and gw create and validate objs [It] correct deployment with sds, AI extension, and floatinguUserId enabled

This unit test failure looks spurious, it's not related to any change here.

pkg/utils/glooadminutils/admincli/client.go
```diff
@@ -34,7 +33,7 @@ func NewClient() *Client {
 		curlOptions: []curl.Option{
 			curl.WithScheme("http"),
 			curl.WithHost("127.0.0.1"),
-			curl.WithPort(DefaultAdminPort),
+			curl.WithPort(int(defaults.EnvoyAdminPort)),
```
sam-heilbron, line 36: I don't think this is safe. There are 2 different admin servers/ports that are relevant:
- Gloo pod (accessible by 9091)
- Envoy pod (gateway-proxy, accessibly by the admin_server, https://github.com/solo-io/gloo/blob/main/install/helm/gloo/templates/9-gateway-proxy-configmap.yaml#L345, which is port 19000)

I had tried to co-locate the relevant port with the client, so that consumers of the client didn't have to know about the internal details. However, I'm certainly ok with a shared defaults that captures this.

I think what is unsafe about this change though, is that in this package, we have a client that can talk to the gloo admin server, which is accessible at 9091, but we are changing the client to use port 19000 which will break functionality (and likely is causing end-to-end tests to fail)

pkg/utils/kubeutils/portforward/cli_portforwarder.go
```diff
@@ -64,13 +66,40 @@ func (c *cliPortForwarder) startOnce(ctx context.Context) error {
 		fmt.Sprintf("%s/%s", c.properties.resourceType, c.properties.resourceName),
 		fmt.Sprintf("%d:%d", c.properties.localPort, c.properties.remotePort),
 	)
-	c.cmd.Stdout = c.properties.stdout
-	c.cmd.Stderr = c.properties.stderr
+
+	fwdOut, err := c.cmd.StdoutPipe()
+	fwdErr, err := c.cmd.StderrPipe()
+	if err != nil {
+		// should not happen unless some other thing has futzed
```
sam-heilbron, line 73: Should we have this err guard for both commands?

pkg/utils/kubeutils/portforward/cli_portforwarder.go
```diff
@@ -64,13 +66,40 @@ func (c *cliPortForwarder) startOnce(ctx context.Context) error {
...
 
 	c.errCh = make(chan error, 1)
 
-	return c.cmd.Start()
+	// short circuit error return if we can't even start
+	if err := c.cmd.Start(); err != nil {
+		return err
+	}
+
+	// Because we are not using a real Go-only kube client but are spawning a long-running
```
sam-heilbron, line 87: Love the comment!

test/helpers/kube_dump.go
```diff
@@ -303,26 +302,16 @@ func EnvoyDumpOnFail(_ io.Writer, proxies ...metav1.ObjectMeta) func() {
...
-			recordEnvoyAdminData(fileAtPath(filepath.Join(envoyOutDir, "stats.log")), "/stats", proxyName, proxyNamespace)
-			recordEnvoyAdminData(fileAtPath(filepath.Join(envoyOutDir, "clusters.log")), "/clusters", proxyName, proxyNamespace)
-			recordEnvoyAdminData(fileAtPath(filepath.Join(envoyOutDir, "listeners.log")), "/listeners", proxyName, proxyNamespace)
-		}
-	}
-}
 
-func recordEnvoyAdminData(f *os.File, path, proxyName, namespace string) {
-	defer f.Close()
+			adminCli, shutdown, _ := admincli.NewPortForwardedClient(context.Background(), fmt.Sprintf("deployment/%s", proxyName), proxyNamespace)
```
sam-heilbron, line 306: Yes! I like this way of using our cli code directly in our test setup as a form of dog-feeding

test/kubernetes/testutils/assertions/gloo.go
```diff
@@ -26,7 +27,7 @@ func (p *Provider) AssertGlooAdminApi(
 
 	portForwarder, err := p.clusterContext.Cli.StartPortForward(ctx,
 		portforward.WithDeployment(glooDeployment.GetName(), glooDeployment.GetNamespace()),
-		portforward.WithPorts(admincli.DefaultAdminPort, admincli.DefaultAdminPort),
+		portforward.WithPorts(int(defaults.EnvoyAdminPort), int(defaults.EnvoyAdminPort)),
```
sam-heilbron, line 30: I think this is another exapmle of us mixing up the glooAdminPort and the EnvoyAdminPort (explained in a separate comment, though happy to talk further about it)

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -47,13 +51,54 @@ func NewClient() *Client {
...
 
+// NewPortForwardedClient takes a pod selector like <podname> or `deployment/<podname`,
+// and returns a port-forwarded Envoy admin client pointing at that pod,
+// as well as a deferrable shutdown function.
+//
+// Designed to be used by tests and CLI from outside of a cluster where `kubectl` is present.
+// In all other cases, `NewClient` is preferred
+func NewPortForwardedClient(ctx context.Context, proxySelector, namespace string) (*Client, func(), error) {
+	var selector portforward.Option
+	if sel := strings.Split(proxySelector, "/"); len(sel) == 2 {
```
sam-heilbron, line 69: I like this. Another option would be that we bake this into the [WithResource](https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/portforward/options.go#L42) implementation

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -47,13 +51,54 @@ func NewClient() *Client {
...
+		portForwarder.WaitForStop()
+	}
+
+	// 3. Create a CLI that connects to the Envoy Admin API
+	adminCli := NewClient().
+		WithCurlOptions(
+			curl.WithHostPort(portForwarder.Address()),
+		)
+
+	return adminCli, deferFunc, err
```
sam-heilbron, line 99: Would it make sense to return the PortForwarder, instead of the function to close it? It puts more resopnsibility on the caller, but it allows them to have the forwarder that may drop errors on a channel

sam-heilbron (commented):
Exciting work!

pkg/utils/glooadminutils/admincli/client.go
bleggett: Ah, I didn't mean to change this one - I did notice there were 2 "EnvoyAdminPorts" and was only interested in changing the `19000` one - ty.

pkg/utils/envoyutils/admincli/client.go
bleggett: > Another option would be that we bake this into the [WithResource](https://github.com/solo-io/gloo/blob/main/pkg/utils/kubeutils/portforward/options.go#L42) implementation

Yeah fair enough.

pkg/utils/kubeutils/portforward/cli_portforwarder.go
bleggett: I should TODO this, really.

pkg/utils/envoyutils/admincli/client.go
bleggett: I thought about that - but for the envoy client you don't really care about the state of the port forwarder itself, you don't need to wait for it to terminate, and you don't care much about the Err channel.

Basically it would complicate the current usages, and none of the current usages care about that Error channel.

Consumers that have different needs can always construct their own portforwarder, and skip this convenience func (that's all it is really).

bleggett:
> > Deployer Single gwc and gw create and validate objs [It] correct deployment with sds, AI extension, and floatinguUserId enabled
> 
> This unit test failure looks spurious, it's not related to any change here.

Okay chased this down with @jenshu 's help a bit - it's not related to any change here, but these changes expose a test bug because they indirectly remove an ambient go pkg import that is required for the tests in `./projects/gateway2/deployer/` to pass.

bleggett:
@sam-heilbron @jenshu this should be g2g - some of the e2es in different clusters flake on a per-run basis, but that seems to be an existing issue.

Raised https://github.com/solo-io/gloo/issues/10256 for the import weirdness - the actual dependency impacting this test needs to be chased down.

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -1,130 +1,122 @@
...
+	}
+
+	adminCli, deferFunc, err := admincli.NewPortForwardedClient(opts.Top.Ctx, opts.Proxy.Name, opts.Metadata.GetNamespace())
+	if err != nil {
+		return proxyOutArchiveFile.Name(), err
+	}
+
+	defer deferFunc()
+
+	writeErr := adminCli.WriteEnvoyDumpToZip(opts.Top.Ctx, proxyOutArchive)
```
nfuden, line 104: Consider passing options to writer. 
Note that there will likely be alot of toggles we want to consider.

For example a debug snapshot for when a customer has a profile enabled proxy instance could include calling /heap_dump and grabbing the dump or grabbing existing [watchdog](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/watchdog/profile_action/v3/profile_action.proto) outputs. 

In addition in the short term we may want to have config_dump optionally include eds. Users may not want to include eds incase it makes the command stall but it can be very powerful and sometimes is required for debugging

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -1,130 +1,122 @@
...
+	if proxyNamespace == "" {
+		proxyNamespace = defaults.GlooSystem
+	}
+
+	adminCli, deferFunc, err := admincli.NewPortForwardedClient(opts.Top.Ctx, opts.Proxy.Name, opts.Metadata.GetNamespace())
+	if err != nil {
+		return proxyOutArchiveFile.Name(), err
+	}
+
+	defer deferFunc()
```
nfuden, line 102: nit: deferFunc doesnt tell me what the deffered func does meaning i have to dig into the portforwarder. Looking at how we use it I assume that it cleans up the portforwarder but its nice when the name itself tells me a bit more.

pkg/utils/glooadminutils/admincli/client.go
```diff
@@ -12,9 +12,8 @@ import (
 )
 
 const (
-	InputSnapshotPath = "/snapshots/input"
-
-	DefaultAdminPort = 9091
+	InputSnapshotPath    = "/snapshots/input"
+	DefaultGlooAdminPort = 9091
```
nfuden, line 16: This is already in glooadmin why the rename?
Note that this is an exported option and changing it could break consuming packages

projects/gloo/cli/pkg/cmd/gateway/dump.go
bleggett: > In addition in the short term we may want to have config_dump optionally include eds. 

Yeah, I agree.

> For example a debug snapshot for when a customer has a profile enabled proxy instance could include calling /heap_dump and grabbing the dump or grabbing existing [watchdog](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/watchdog/profile_action/v3/profile_action.proto) outputs.

Yep - I agree we will want to add these things. I'd prefer to do that in a followup, I am intentionally trying to keep this delta small ATM tho, and mostly fixup the existing tests that rely on the CLI bits.

pkg/utils/glooadminutils/admincli/client.go
bleggett: I am fine changing it back tho, if it is an export.

projects/gloo/cli/pkg/cmd/gateway/dump.go
bleggett: I'll add the `eds` flag now as that's probably immediately useful. I'd prefer to get feedback on how useful the basic version as-is in this PR is, before adding more stuff.

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -261,3 +296,34 @@ func (c *Client) GetSingleListenerFromDynamicListeners(
...
 	return &listener, nil
 }
+
+// WriteEnvoyDumpToZip will dump config, stats, clusters and listeners to zipfile in the current directory.
+// Useful for diagnostics or testing
+func (c *Client) WriteEnvoyDumpToZip(ctx context.Context, zip *zip.Writer) error {
+	// zip writer has the benefit of not requiring tmpdirs or file ops (all in mem)
+	// - but it can't support async writes, so do these sequentally
+	// Also don't join errors, we want to fast-fail
+	if err := c.ConfigDumpCmd(ctx, nil).WithStdout(fileInArchive(zip, "config.log")).Run().Cause(); err != nil {
```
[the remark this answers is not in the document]
bleggett: Nope, this is mostly mirroring the filenames used in other spots in the existing code. 

I'll change to `config.txt`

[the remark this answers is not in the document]
bleggett: (and the others too, since the same argument applies)

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -261,3 +296,34 @@ func (c *Client) GetSingleListenerFromDynamicListeners(
 	}
 	return &listener, nil
 }
+
+// WriteEnvoyDumpToZip will dump config, stats, clusters and listeners to zipfile in the current directory.
```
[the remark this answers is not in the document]
bleggett: `server_info` is a very good call, ty

pkg/utils/envoyutils/admincli/client.go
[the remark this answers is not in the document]
bleggett: (actually no, derp - `config` and `server_info` should be `.json`)

bleggett:
@nfuden think I addressed all your comments, PTAL

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -1,130 +1,123 @@
...
+	if writeErr == nil {
+		fmt.Println("proxy snapshot written to " + proxyOutArchiveFile.Name())
+	} else {
+		fmt.Printf("Error writing proxy snapshot: %s", writeErr)
+	}
+
+	return proxyOutArchiveFile.Name(), writeErr
+}
+
+// createArchive forcibly deletes/creates the output directory
```
sam-heilbron, line 116: nit: Let's update this to match the new name and clarify that no deletion takes place, but instead it's the callers responsibility to delete the file

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -261,3 +304,42 @@ func (c *Client) GetSingleListenerFromDynamicListeners(
...
+// Useful for diagnostics or testing
+func (c *Client) WriteEnvoyDumpToZip(ctx context.Context, options DumpOptions, zip *zip.Writer) error {
+	configParams := make(map[string]string)
+	if options.ConfigIncludeEDS {
```
sam-heilbron, line 312: Is it possible for us to make this default to true and have users turn it off instead? It provides really useful content, and something we likely benefit from getting from the user and the fact that it defaults to false in the envoy admin endpoint doesnt mean we should pass that complexity on to the user (i think). what do you think?

projects/gloo/cli/pkg/cmd/gateway/dump.go
```diff
@@ -1,130 +1,123 @@
...
+func createArchiveFile() (*os.File, error) {
+	f, err := os.Create(fmt.Sprintf("glooctl-proxy-snapshot-%s.zip", time.Now().Format("2006-01-02-T15.04.05")))
```
sam-heilbron, line 118: It's possible to have multiple proxies running in a given cluster, attached to a single control plane (https://github.com/solo-io/gloo/blob/main/projects/gloo/pkg/xds/README.md#xds-requests). Given this, how will we ensure that we can create archives for each proxy uniquely? Would it be useful to include some metadata in the name?

test/helpers/kube_dump.go
```diff
@@ -303,26 +303,16 @@ func EnvoyDumpOnFail(_ io.Writer, proxies ...metav1.ObjectMeta) func() {
...
```
sam-heilbron, line 307: This is great, this is the preFailHandler for our old-syle e2e tests in gloo. https://github.com/solo-io/gloo/tree/main/test/kubernetes#end-to-end-testing is where our new-style tests are, and I think they would benefit from calling this code as well.

Could we update https://github.com/solo-io/gloo/blob/main/test/kubernetes/e2e/test.go#L229 to include thiss?

sam-heilbron (commented):
Looking great! A few minor questions

pkg/utils/envoyutils/admincli/client.go
bleggett: I debated that. If we feel it's safe to default to true I am happy to do so.

test/helpers/kube_dump.go
bleggett: I don't wanna scopecreep this too much with testcode refactors - but yes. Opened https://github.com/solo-io/gloo/issues/10289 and I'll do that in a separate PR.

sam-heilbron (approved):

pkg/utils/envoyutils/admincli/client.go
```diff
@@ -25,10 +30,14 @@ const (
 	HealthCheckPath    = "healthcheck"
 	LoggingPath        = "logging"
 	ServerInfoPath     = "server_info"
-
-	DefaultAdminPort = 19000
```
nfuden, line 29: Technically a breaking change?

nfuden (approved):

pkg/utils/envoyutils/admincli/client.go
bleggett: https://github.com/solo-io/gloo/issues/10314 based on slack convo

PR #10244, opened 2024-10-28
Add Helm field to simplify GG Portal enablement

inFocus7:
# Description

Add new field in oss chart (for enterprise use only) to enable portal and sets up the env var automatically. solo-projects https://github.com/solo-io/solo-projects/pull/7127.

## API changes

- Add new helm field for GG Portal `gloo.kubeGateway.portal.enabled`.

# Context

enabling portal was confusing. there are two components, the translation/discovery that runs in gloo and the web server. 

previously to enable both, you need to do
```yaml
gloo:
  kubeGateway:
    enabled: true
  gloo:
    deployment:
      customEnv: # this is usually what people tend to forget to set/unset when enabling/disabling portal.
      - name: GG_PORTAL_PLUGIN
        value: "true"
gateway-portal-web-server:
  enabled: true
  glooPortalServer:
    # this is technically default to true, so we don't need to set it, but doing this to be explicit
    enabled: true
```

now it's like
```yaml
gloo:
  kubeGateway:
    enabled: true
    portal: # because this value is on the oss chart we can simply use it to automatically set that envVar.
      enabled: true
gateway-portal-web-server:
  glooPortalServer:
    # this is technically default to true, so we don't need to set it, but doing this to be explicit
    enabled: true
```

## Testing steps

in ee repo.

# Checklist:


inFocus7:
/kick (flake fail? need it to publish chart to use & verify on solo-projects)

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -241,6 +241,9 @@ spec:
           - name: GG_K8S_GW_CONTROLLER
             value: "true"
         {{- end }}{{/* if .Values.kubeGateway.enabled */}}
+        {{- if .Values.kubeGateway.portal.enabled }}
+          - name: GG_PORTAL_PLUGIN
+            value: "true"
```
jenshu, line 246: missing `end` here

inFocus7: Yeah just noticed, pushing up the fix 🤦🏼

andy-fong (approved):

jmhbh (approved):

andy-fong (approved):

PR #10248, opened 2024-10-29
aws_sts_cluster static cluster config for kubernetes gateway api

jenshu:
# Description

As part of supporting [AWS Lambda with EKS ServiceAccounts](https://docs.solo.io/gloo-edge/latest/guides/traffic_management/destination_types/aws_lambda/eks-service-accounts/) with the Kubernetes Gateway API, the existing Helm values `settings.aws.enableServiceAccountCredentials` and `settings.aws.stsCredentialsRegion` are now respected when using Kubernetes Gateway API. 

The AWS options from Settings CR (which are configured at install time) are passed through to the kubernetes gateway controller. When `enableServiceAccountCredentials` is true, a `aws_sts_cluster` cluster, configured with the STS endpoint specified by `stsCredentialsRegion`, will automatically be added to dynamically provisioned proxies, so that Envoy can reach AWS to assume the role needed to access Lambdas.

## Testing

The following tests were added:
- Deployer unit tests showing that passing aws options to the deployer results in the expected envoy cluster config
- Added new TestK8sGatewayAws kubernetes e2e test suite which installs GG with aws options configured. For now it only runs the Deployer tests, to make sure that the aws options are correctly passed through to the configured proxies.
- For a complete e2e test using aws/lambda, we plan to run our k8s gateway lambda workshop using a dev build with these changes and verify all steps in the workshop succeed

# Checklist:

### DIFF
--- a/install/test/k8sgateway_test.go
+++ b/install/test/k8sgateway_test.go
@@ -1,7 +1,6 @@
 package test
 
 import (
-	"encoding/json"
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -66,14 +65,7 @@ var _ = Describe("Kubernetes Gateway API integration", func() {
 				testManifest.Expect("ClusterRoleBinding", "", deployerRbacName+"-binding").NotTo(BeNil())
 			})
 			It("renders default GatewayParameters", func() {
-				gwpUnstructured := testManifest.ExpectCustomResource("GatewayParameters", namespace, wellknown.DefaultGatewayParametersName)
-				Expect(gwpUnstructured).NotTo(BeNil())
-
-				var gwp v1alpha1.GatewayParameters
-				b, err := gwpUnstructured.MarshalJSON()
-				Expect(err).ToNot(HaveOccurred())
-				err = json.Unmarshal(b, &gwp)
-				Expect(err).ToNot(HaveOccurred())
+				gwp := getDefaultGatewayParameters(testManifest)
 
 				gwpKube := gwp.Spec.Kube
 				Expect(gwpKube).ToNot(BeNil())
@@ -194,14 +186,7 @@ var _ = Describe("Kubernetes Gateway API integration", func() {
 					valuesArgs = append(valuesArgs, extraValuesArgs...)
 				})
 				It("passes overrides to default GatewayParameters with Istio container", func() {
-					gwpUnstructured := testManifest.ExpectCustomResource("GatewayParameters", namespace, wellknown.DefaultGatewayParametersName)
-					Expect(gwpUnstructured).NotTo(BeNil())
-
-					var gwp v1alpha1.GatewayParameters
-					b, err := gwpUnstructured.MarshalJSON()
-					Expect(err).ToNot(HaveOccurred())
-					err = json.Unmarshal(b, &gwp)
-					Expect(err).ToNot(HaveOccurred())
+					gwp := getDefaultGatewayParameters(testManifest)
 
 					gwpKube := gwp.Spec.Kube
 					Expect(gwpKube).ToNot(BeNil())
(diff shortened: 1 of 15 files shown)


install/test/k8sgateway_test.go
```diff
@@ -66,14 +65,7 @@ var _ = Describe("Kubernetes Gateway API integration", func() {
 				testManifest.Expect("ClusterRoleBinding", "", deployerRbacName+"-binding").NotTo(BeNil())
 			})
 			It("renders default GatewayParameters", func() {
-				gwpUnstructured := testManifest.ExpectCustomResource("GatewayParameters", namespace, wellknown.DefaultGatewayParametersName)
```
jenshu, line 69: not relevant to the other changes, but i copied over this refactor from the [original PR](https://github.com/solo-io/gloo/pull/10205) because it cleans things up a bit

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -542,6 +542,26 @@ data:
...
+        {{- if ($gateway.aws).enableServiceAccountCredentials }}
+        - name: {{ $gateway.aws.stsClusterName }}
+          connect_timeout: 5.000s
+          type: LOGICAL_DNS
+          lb_policy: ROUND_ROBIN
+          transport_socket:
+            name: envoy.transport_sockets.tls
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
+              sni: {{ $gateway.aws.stsUri }}
```
andy-fong, line 554: So, the protobuf comment is: 
```
    // The full uri of the aws sts endpoint
    string uri = 2 [ (validate.rules).string.min_bytes = 1 ];
```
Usually, a "full uri" means that it includes the protocol like `https://sts.amazonaws.com`. SNI usually is the CNAME that matches the cert. Should we update the doc to make sure people only put the hostname and nothing else in the uri field?

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
```diff
@@ -542,6 +542,26 @@ data:
...
+          load_assignment:
+            cluster_name: {{ $gateway.aws.stsClusterName }}
+            endpoints:
+            - lb_endpoints:
+              - endpoint:
+                  address:
+                    socket_address:
+                      port_value: 443
+                      address: {{ $gateway.aws.stsUri }}
```
andy-fong, line 563: looks like we do expect the Uri as just a host name. That's quite confusing.

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
andy-fong: looks like in the 9-gateway-proxy-configmap.yaml , sni is set to:
```
sni: sts.{{ .Values.settings.aws.stsCredentialsRegion }}.amazonaws.com
```
which makes more sense.

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
andy-fong: in other template, the address is set to:
```
address: sts.{{ .Values.settings.aws.stsCredentialsRegion }}.amazonaws.com
```

projects/gateway2/helm/gloo-gateway/templates/gateway/proxy-deployment.yaml
jenshu: yeah, we call it uri in our envoy proto so i just kept that naming.
[sts docs](https://docs.aws.amazon.com/general/latest/gr/sts.html) refer to it as an 'endpoint', would that be better? i could rename the values that were added in this PR but the envoy proto would still have `uri` though

andy-fong: If the `AWSLambdaConfig_ServiceAccountCredentials` message is only created and used internally and customer cannot set that directly, I think just fixing or clarifying the comment on the `uri` field will be fine.

andy-fong: so, had a discussion, because the proto file is from envoy-gloo, we will leave that for now.

andy-fong (approved):

nfuden (approved):

PR #10251, opened 2024-10-30
feat: support destination rules for endpoints & upstreams 

yuval-k:
Limited destination rule support for endpoints and upstreams.
- For usptreams, we only support the outlier detection
- For endpoints we only support failover/failOverPriority


projects/gateway2/controller/start.go
```diff
@@ -68,7 +68,9 @@ type StartConfig struct {
 
 	Client istiokube.Client
 
-	Pods            krt.Collection[krtcollections.LocalityPod]
+	Pods krt.Collection[krtcollections.LocalityPod]
+	Ucc  krt.Collection[krtcollections.UniqlyConnectedClient]
```
stevenctl, line 72: nit: rename before merging

projects/gateway2/krtcollections/uniqueclients.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func newConnectedClient(uniqueClientName string) ConnectedClient {
+	return ConnectedClient{
+		uniqueClientName: uniqueClientName,
+	}
+}
+
+type UniqlyConnectedClient struct {
+	// name namespace of gateway
```
stevenctl, line 35: not sure what this comment means

projects/gateway2/proxy_syncer/cla.go
```diff
@@ -69,3 +77,132 @@ func prioritize(ep EndpointsForUpstream) *envoy_config_endpoint_v3.ClusterLoadAs
...
+}
+
+func snapshotPerClient(l *zap.Logger, uccCol krt.Collection[krtcollections.UniqlyConnectedClient],
+	mostXdsSnapshots krt.Collection[xdsSnapWrapper], ie IndexedEndpoints) krt.Collection[xdsSnapWrapper] {
+
+	xdsSnapshotsForUcc := krt.NewCollection(uccCol, func(kctx krt.HandlerContext, ucc krtcollections.UniqlyConnectedClient) *xdsSnapWrapper {
+		mostlySnaps := krt.Fetch(kctx, mostXdsSnapshots, krt.FilterKey(ucc.Role))
+		if len(mostlySnaps) != 1 {
+			return nil
+		}
```
lgadban, line 177: is this replicating `FetchOne` ?

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -104,18 +125,9 @@ func (s *ProxyTranslator) syncXdsAndStatus(
...
-		logger.Debugw(syncutil.StringifySnapshot(snap), "proxyKey", proxyKey)
-	}
-
-	// if the snapshot is not consistent, make it so
-	snap.MakeConsistent()
-	s.xdsCache.SetSnapshot(proxyKey, snap)
-
 	// TODO: only leaders should write status (https://github.com/solo-io/solo-projects/issues/6367)
-	logger.Debugf("gloo reports for proxy %s to be written: %v", proxyKey, reports)
+	// 	logger.Debugf("gloo reports for proxy %s to be written: %v", proxyKey, reports) // QUESTION: this spams the logs. can we disable?
```
lgadban, line 129: i think we can; this was leftover from previous copy/paste and I couldn't make a decision on whether or not it was useful

projects/gateway2/krtcollections/uniqueclients.go
```diff
@@ -0,0 +1,303 @@
...
+
+func (x *callbacksCollection) streamClosed(sid int64) {
+	ucc := x.del(sid)
+	if ucc != nil {
+		x.trigger.TriggerRecomputation()
+	}
+}
+
+func (x *callbacksCollection) del(sid int64) *UniqlyConnectedClient {
+	x.stateLock.Lock()
```
stevenctl, line 146: In a normal KRT collection we would:

1. Get triggered by our input collection
2. Compute our outputs, edit internal state (locked), also collect changed stuff into a list for step 3
3. Call downstream handlers/dependent collection with the _changed_ outputs from step 2

The stuff that we pass to handlers in 3 is a list separate from the collections lock-protected state. 

In this impl:

1. Triggered by xds events
2. Update our internal state
3. Call recompute trigger
4. When the recompute trigger happens, we query our internal state

I think this is fine, as when we `getClients()` that's lock protected. We just need to take care not to call TriggerRecomputation while holding the lock.

projects/gateway2/krtcollections/uniqueclients.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func (x *callbacksCollection) newStream(sid int64, r *envoy_service_discovery_v3.DiscoveryRequest) error {
+	ucc, err := x.add(sid, r)
+	if err != nil {
+		x.logger.Debug("error processing xds client", zap.Error(err))
+		return err
+	}
+	if ucc != "" {
+		nodeMd := r.GetNode().GetMetadata()
```
stevenctl, line 218: If we mutate a reference to something in our state, we should do that holding the lock, but we only hold the lock in `add` here

yuval-k: i don't think i mutate anything that's in our state here?
ucc is a string and r is not part of the state?

projects/gateway2/setup/ggv2setup.go
```diff
@@ -107,6 +109,8 @@ func StartGGv2(ctx context.Context,
 		settingsGVR,
 		krt.WithName("GlooSettings"))
 
+	ucc := uccBuilder(ctx, pods)
```
lgadban, line 112: nit: this is existing code but it's more apparent here, can we rename `pods` collection to `augmentedPods` or something similar to point out how this is not raw k8s pods

projects/gateway2/krtcollections/uniqueclients.go
stevenctl: misunderstood, lg

projects/gateway2/proxy_syncer/destrule.go
```diff
@@ -0,0 +1,88 @@
...
+	"google.golang.org/protobuf/proto"
+	networkingclient "istio.io/client-go/pkg/apis/networking/v1"
+	"istio.io/istio/pkg/config/schema/gvr"
+	"istio.io/istio/pkg/kube"
+	"istio.io/istio/pkg/kube/kclient"
+	"istio.io/istio/pkg/kube/krt"
+	"istio.io/istio/pkg/kube/kubetypes"
+)
+
+type NsWithHostname struct {
```
stevenctl, line 16: super nit: can we just use `krt.Named` and treat `Name` as a hostname?

projects/gateway2/proxy_syncer/cla.go
```diff
@@ -69,3 +77,132 @@ func prioritize(ep EndpointsForUpstream) *envoy_config_endpoint_v3.ClusterLoadAs
...
+type IndexedEndpoints struct {
+	endpoints krt.Collection[uccWithEndpoints]
+	index     krt.Index[string, uccWithEndpoints]
+}
+
+func (ie *IndexedEndpoints) FetchEndpointsForClient(kctx krt.HandlerContext, ucc krtcollections.UniqlyConnectedClient) []uccWithEndpoints {
+	return krt.Fetch(kctx, ie.endpoints, krt.FilterIndex(ie.index, ucc.ResourceName()))
+}
+
+func NewIndexedEndpoints(logger *zap.Logger, uccs krt.Collection[krtcollections.UniqlyConnectedClient],
```
stevenctl, line 105: the word endpoint is starting to get confusing in some spots

```suggestion
func NewPerClientEnvoyEndpoints(logger *zap.Logger, uccs krt.Collection[krtcollections.UniqlyConnectedClient],
```

projects/gateway2/proxy_syncer/cla.go
```diff
@@ -69,3 +77,132 @@ func prioritize(ep EndpointsForUpstream) *envoy_config_endpoint_v3.ClusterLoadAs
...
+	lbInfo := LoadBalancingInfo{
+		PodLabels:    c.Labels,
+		PodLocality:  c.Locality,
+		PriorityInfo: priorityInfo,
+	}
+
+	return prioritizeWithLbInfo(logger, ep, lbInfo)
+}
+
+func fromEndpoint(ep EndpointsForUpstream) string {
```
stevenctl, line 153: I think `EndpointsForUpstream` might need a `Hostnames` field on it. Besides the `cluster.local` thing, `ServiceEntry` will provide (potentially many) hostnames

stevenctl: Can handle that in my PR though

yuval-k: makes sense

projects/gateway2/proxy_syncer/cla.go
yuval-k: we can use fetch one here yea

projects/gateway2/proxy_syncer/cla.go
yuval-k: i can add the hostname cluster local here

projects/gloo/pkg/setup/setup.go
```diff
@@ -30,16 +32,20 @@ func Main(customCtx context.Context) error {
 }
 
 func startSetupLoop(ctx context.Context) error {
+	k8sgw := envutils.IsEnvTruthy(constants.GlooGatewayEnableK8sGwControllerEnv)
 
 	// get settings:
-
-	setupOpts := bootstrap.NewSetupOpts(xds.NewAdsSnapshotCache(ctx))
+	var ucc xdsserver.Callbacks
```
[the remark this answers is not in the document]
yuval-k: ahh yea

projects/gateway2/utils/upstream.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+
+	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
+)
+
+func GetHostnameForUpstream(us *v1.Upstream) string {
+	// get the upstream name and namespace
+	// TODO: suppport other suffixes that are not cluster.local
+
+	switch uptype := us.GetUpstreamType().(type) {
```
stevenctl, line 13: note to myself mostly: support static upstream by using the first host here

stevenctl: https://github.com/stevenctl/gloo/blob/cdf9ad741ab1fb6e1d91eb03bb7a78f36df3bf20/projects/gloo/pkg/plugins/istio_automtls/plugin.go#L178C6-L178C14

This feels vaguely similar to this code

projects/gloo/pkg/translator/clusters.go
```diff
@@ -50,57 +50,86 @@ func (t *translatorInstance) computeClusters(
...
 	}
-	return out
+	return out, errs
 }
 
 func (t *translatorInstance) initializeCluster(
 	upstream *v1.Upstream,
-	upstreamRefKeyToEndpoints map[string][]*v1.Endpoint,
-	reports reporter.ResourceReports,
+	eds bool,
```
stevenctl, line 113: where is `eds` used?

projects/gloo/pkg/translator/clusters.go
```diff
@@ -50,57 +50,86 @@ func (t *translatorInstance) computeClusters(
...
+			}
+		}
 		clusterToUpstreamMap[cluster] = upstream
 		clusters = append(clusters, cluster)
 	}
 
 	return clusters, clusterToUpstreamMap
 }
 
+func (t *translatorInstance) TranslateCluster(
```
stevenctl, line 72: add a comment for this one's intended usage.. also, is `eds` ever going to be true when calling this exported version? Seems like we would want it inferred by plugins when using this path. Setting `eds = false` could set a weird expectation.

yuval-k:
mostly ready for review - just need to add unit tests

projects/gateway2/proxy_syncer/destrule.go
yuval-k: we could, but i think that would be more confusion to a reader without the context

projects/gloo/pkg/translator/clusters.go
yuval-k: yea eds won't ever be true here; i'll remove it

projects/gateway2/krtcollections/uniqueclients.go
yuval-k: not sure either!. i'll write up some comments in general

install/helm/gloo/templates/44-rbac.yaml
```diff
@@ -50,6 +50,11 @@ rules:
   resources:
   - customresourcedefinitions
   verbs: ["get", "list", "watch"]
+- apiGroups:
```
yuval-k, line 53: I think this is the correct place. @jenshu , can you confirm?

jenshu: yes this is the right place

do we need to gate this by some flag, or do we have a dependency on istio resources now?

yuval-k: i don't think there is harm putting it there even if the CRD is not installed (i.e. it will still be accepted). so this is not a strong dependency.

stevenctl (approved):

lgadban (approved):
Amazing! 🎉

projects/gloo/pkg/translator/translator.go
```diff
@@ -68,12 +78,22 @@ func NewTranslatorWithHasher(
 	pluginRegistry plugins.PluginRegistry,
 	hasher func(resources []envoycache.Resource) (uint64, error),
 ) *translatorInstance {
+	shouldEnforceStr := os.Getenv(api_conversion.MatchingNamespaceEnv)
+	shouldEnforceNamespaceMatch := false
+	if shouldEnforceStr != "" {
+		var err error
+		shouldEnforceNamespaceMatch, err = strconv.ParseBool(shouldEnforceStr)
+		if err != nil {
+			// TODO: what to do here?
```
nfuden, line 87: We should log it minimally.

projects/gloo/pkg/syncer/setup/callbacks.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+// OnStreamRequest is called once a request is received on a stream.
+// Returning an error will end processing and close the stream. OnStreamClosed will still be called.
+func (mc MutltiCallbacks) OnStreamRequest(sid int64, dr *envoy_service_discovery_v3.DiscoveryRequest) error {
+	for _, c := range mc {
+		err := c.OnStreamRequest(sid, dr)
+		if err != nil {
+			return err
```
nfuden, line 37: Can we call out that multicallbacks are implicitly ordered due to this error behavior?

Issue #10250, opened 2024-10-29
TestDiscoveryWatchlabels/Discovery/TestDiscoverySpecPreserved


bewebi:
### Which tests failed?

TestDiscoveryWatchlabels/Discovery/TestDiscoverySpecPreserved

Logs:
```
=== RUN   TestDiscoveryWatchlabels/Discovery/TestDiscoverySpecPreserved
service/example-svc created
=== NAME  TestDiscoveryWatchlabels
    status.go:71: 
        Timed out after 20.001s.
        The function passed to Eventually failed at /home/runner/work/gloo/gloo/test/kubernetes/testutils/assertions/status.go:70 with:
        have matcher for namespace discovery-watchlabels-test which is not found
service "example-svc" deleted
```

### Initial Investigation

This has been reproducible locally, though infrequently, whereas it seems to occur much more frequently in CI

This flake occurs when an Upstream created by Discovery does not have its status populated

The failure occurs [here](https://github.com/solo-io/gloo/blob/bc01f4dc028bf0d6724b6607a2420259e748ef6d/test/kubernetes/e2e/features/discovery_watchlabels/discovery_watchlabels_suite.go#L133), on the assertion that eventually an Upstream, which we expect Discovery to create, has an "Accepted" namespaced status

When debugging I found that occasionally the Upstream is created and never gets a status, which leads that assertion to fail

I'm not sure what would cause an Upstream not to get a status, but it may be notable that this failure seems to exclusively occur on the second test, `TestDiscoverySpecPreserved`, and not `TestDiscoverUpstreamMatchingWatchLabels`, suggesting that there may be some form of test pollution

### Additional Information

[One example](https://github.com/solo-io/gloo/actions/runs/11547696038/job/32137980538) of a failure on nightlies

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1207635280052738/1208757360394486) by [Unito](https://www.unito.io)

sam-heilbron:
"I'm not sure what would cause an Upstream not to get a status" - The gloo pod is responsible for writing statuses, and to me a status not getting written is a sign of a potential bug. Even if there is test pollution, I would expect a resource to be processed. 

I would recommend looking to see if we apply another resource (ie trigger translation) does a status appear on this object? If so, it's a sign that the ControlPlane may have "dropped" an event. If not, it's a sign that there's potentially a more fundamental bug (I would look at debug logs of the controlplane to see why that upstream isn't receiving updates)

[...]

PR #10253, opened 2024-10-30
Test discovery spec preserved flake

bewebi:
# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


bewebi:
This PR is/was an example solution for working around a flake (#10250), albeit without fixing the underlying bug (solo-io/solo-projects#7153)
There is no intention to merge this PR in particular, even if we want to do a quick workaround rather than fixing the root cause a better mechanism has been identified, namely not reusing resource names across specs

PR #10255, opened 2024-10-30
ggv2: make cluster names parseable

stevenctl:
# Description

https://github.com/solo-io/solo-projects/issues/7105

This will make cluster names for kube services look like `kube-svc_name_namespace_port` or `kube-svc_reviews_bookinfo_8080`. If we add other resources besides a kube service, this also lets us get things like `istio-se_helloworld_backend-ns_5000`. 

This will only happen for Gateway v2 translation. We can disable it using the env var `GG_K8S_GW_LEGACY_CLUSTER_NAMES`.

## API changes

We can disable it using the env var `GG_K8S_GW_LEGACY_CLUSTER_NAMES`.

## Code changes

* XDS translator now takes options that can customize this new behavior. We instantiate translator separately for v2, so that is the only path to enable it. 
* Added another function to customize the generated name rather than modifying the existing cluster name function (it has over 100 usages! mostly in unit tests)
* Use internal labels to propagate the info in a structured way. 

## Docs changes

We should document this as a breaking change; users who rely on the old format should use the env var or update their EnvoyFilters or whatever config relies on XDS names... 

# Context

https://github.com/solo-io/solo-projects/issues/7105

## Interesting decisions

* Use internal labels to propagate the info in a structured way. This was done so we can support more than just Upstream_Kube upstreams. 
* Functional options for translator constructor. I didn't feel like updating the many references to this constructor, so varargs help. In hindsight, a second constructor could have also solved this. 

## Testing steps

TODO!

### DIFF
--- a/projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
+++ b/projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
@@ -8,10 +8,8 @@ import (
 	"github.com/solo-io/solo-kit/pkg/api/v2/reporter"
 	"istio.io/istio/pkg/kube/krt"
 
-	"github.com/solo-io/gloo/pkg/utils/envutils"
 	"github.com/solo-io/gloo/pkg/utils/settingsutil"
 	"github.com/solo-io/gloo/pkg/utils/statsutils"
-	"github.com/solo-io/gloo/projects/gloo/constants"
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/grpc/validation"
 	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
 	v1snap "github.com/solo-io/gloo/projects/gloo/pkg/api/v1/gloosnapshot"
@@ -66,12 +64,10 @@ func (s *ProxyTranslator) buildXdsSnapshot(
 		Messages: map[*core.ResourceRef][]string{},
 	}
 
-	legacyClusterNames := envutils.IsEnvTruthy(constants.GlooGatewayKubeStyleClusterNames)
 	tx := s.translator.NewTranslator(
 		ctx,
 		settings,
-		translator.ForGatewayV2(),
-		translator.WithParseableClusterNames(!legacyClusterNames),
+		translator.ForKubeGatewayAPI(),
 	)
 	xdsSnapshot, reports, proxyReport := tx.Translate(params, proxy)
 
--- a/projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
+++ b/projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
@@ -8,12 +8,15 @@ import (
 	"github.com/solo-io/solo-kit/pkg/api/v2/reporter"
 	"istio.io/istio/pkg/kube/krt"
 
+	"github.com/solo-io/gloo/pkg/utils/envutils"
 	"github.com/solo-io/gloo/pkg/utils/settingsutil"
 	"github.com/solo-io/gloo/pkg/utils/statsutils"
+	"github.com/solo-io/gloo/projects/gloo/constants"
 	"github.com/solo-io/gloo/projects/gloo/pkg/api/grpc/validation"
 	v1 "github.com/solo-io/gloo/projects/gloo/pkg/api/v1"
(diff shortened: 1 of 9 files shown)


projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -64,7 +67,14 @@ func (s *ProxyTranslator) buildXdsSnapshot(
 		Messages: map[*core.ResourceRef][]string{},
 	}
 
-	xdsSnapshot, reports, proxyReport := s.translator.NewTranslator(ctx, settings).Translate(params, proxy)
+	legacyClusterNames := envutils.IsEnvTruthy(constants.GlooGatewayKubeStyleClusterNames)
```
yuval-k, line 70: i dont think we need config here - IIUC for ggv2 this should always be true. cc @josh-pritchard

stevenctl:
#10251  conflicts with this a little - https://github.com/solo-io/gloo/pull/10251/files#diff-1e41d85739cdfae37c6b9d7d40477a5030a000c07d7694c9fd3f0148df966aa2R321

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
krisztianfekete: Maybe it would still make sense to enable users to upgrade from earlier versions, and configure it independently? E.g. enabling this by default for ggv2, but keeping it toggle-able.

projects/gloo/pkg/translator/translator.go
```diff
@@ -45,35 +45,60 @@ type Translator interface {
...
+func WithParseableClusterNames(enabled bool) Option {
+	return func(o *translatorOpts) {
+		o.parseableClusterNames = enabled
+	}
+}
+
+type translatorOpts struct {
+	// ggv2 should only be set for translating ggv2 proxies.
+	ggv2 bool
+	// parseableClusterNames only applies when GGv2 is also true.
```
jenshu, line 78: let's not use "ggv2" anywhere.  "kube gateway" or k8s gateway" is preferred (to distinguish between the kubernetes gateway vs. edge integration)

projects/gloo/pkg/plugins/kubernetes/uds_convert.go
```diff
@@ -20,6 +21,37 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+// these labels are used to propagate internal data
+// on synthetic Gloo resources generated from other Kubernetes
+// resources (generally Service).
+const (
+	// KubeSourceResourceLabel indicates the kind of resource that the synthetic
+	// resource is based on.
+	KubeSourceResourceLabel = "~internal.solo.io/kubernetes-source-resource"
```
jenshu, line 30: why do these all begin with `~`? are we purposely using an invalid label character?

projects/gloo/pkg/plugins/kubernetes/uds_convert.go
```diff
@@ -20,6 +21,37 @@ import (
...
+	// that the synthetic resource is based on.
+	KubeNamespaceLabel = "~internal.solo.io/kubernetes-namespace"
+	// KubeSourceResourceLabel indicates the service port when applicable.
+	KubeServicePortLabel = "~internal.solo.io/kubernetes-service-port"
+)
+
+// ClusterNameForKube builds the cluster name based on _internal_ labels.
+// All of the kind, name, namespace and port must be provided.
+func ClusterNameForKube(us *v1.Upstream) (string, bool) {
+	labels := us.Metadata.GetLabels()
```
nfuden, line 44:
```suggestion
	labels := us.GetMetadata().GetLabels()
```

Our ci will block this via a getterchecker requirement. Ie if there is a getter use that as a nil safe way of chaining. 🤷

nfuden:
I know that its a bit painful but we dont have that many examples of remote services in oss that are populated in helm bootstrap. We have several in enterprise. 
Would you mind putting up a pr in sp and just showing that with this change ci isnt broken?

projects/gloo/pkg/plugins/kubernetes/uds_convert.go
stevenctl: I can add that to the comment about synthetic resources.

jenshu:
closing in favor of https://github.com/solo-io/gloo/pull/10286 (non-fork version)

PR #10265, opened 2024-11-01
project/gateway2: Replaces Endpoints with EndpointSlice for Creating Upstream Endpoints

danehans:
# Description

Updates the gateway2 project to use EndpointSlice instead of Endpoints for creating upstream endpoints.

Fixes: https://github.com/solo-io/solo-projects/issues/6910

## API changes

N/A

## Code changes

The primary change is updating `proxy_syncer.TransformUpstreamsBuilder()` to transform an EndpointSlice into into a `envoy_config_endpoint_v3.LbEndpoint` instead of using an Endpoint.

## CI changes

Removes skipping k8s GW API conformance tests.

## Docs changes

N/A (internal change)

# Context

Required for passing k8s gw api conformance tests.

## Testing steps

Updates existing tests.

## Notes for reviewers

N/A

# Checklist:


danehans:
Converted to draft as I troubleshoot e2e failures.

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -172,68 +171,79 @@ func TransformUpstreamsBuilder(ctx context.Context, inputs EndpointsInputs) func
...
 			logger.Debug("findPortForService - not found.", zap.Uint32("port", spec.GetServicePort()), zap.String("svcName", spec.GetServiceName()), zap.String("svcNamespace", spec.GetServiceNamespace()))
 			return nil
 		}
 
-		maybeEps := krt.FetchOne(kctx, kubeEndpoints, krt.FilterObjectName(types.NamespacedName{
-			Namespace: spec.GetServiceNamespace(),
-			Name:      spec.GetServiceName(),
+		// Fetch all EndpointSlices for the service
+		endpointSlices := krt.Fetch(kctx, kubeEndpointSlices, krt.FilterGeneric(func(o any) bool {
+			es, ok := o.(*discoveryv1.EndpointSlice)
```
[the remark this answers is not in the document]
yuval-k: see index creation example here:
https://github.com/solo-io/gloo/blob/c2593423ee9578fe1c61e87d3a6984897ff596a5/projects/gateway2/proxy_syncer/destrule.go#L66
and usage example here:
https://github.com/solo-io/gloo/blob/c2593423ee9578fe1c61e87d3a6984897ff596a5/projects/gateway2/proxy_syncer/destrule.go#L91

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -172,68 +171,79 @@ func TransformUpstreamsBuilder(ctx context.Context, inputs EndpointsInputs) func
...
-				}
-				ep := createLbEndpoint(addr.IP, port, podLabels, enableAutoMtls)
 
-				ret.Add(l, EndpointWithMd{
-					LbEndpoint: ep,
-					EndpointMd: EndpointMetadata{
-						Labels: augmentedLabels,
-					},
-				})
+					for _, address := range endpoint.Addresses {
```
yuval-k, line 215: note that you can have the same address twice multiple endpoint slices, so we need logic to account for that.
see: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints

nfuden:
Likely there are some lookups that currently hard code to endpoints which may be causing the breaks

projects/gateway2/proxy_syncer/endpoints.go
danehans: @yuval-k thanks for catching this. Commit https://github.com/solo-io/gloo/pull/10265/commits/3f3954517104dcc38166263d27dcd30180308410 adds ep dedupe support, PTAL.

projects/gateway2/proxy_syncer/endpoints.go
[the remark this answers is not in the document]
danehans: @yuval-k  commit https://github.com/solo-io/gloo/pull/10265/commits/571ed2bfaeec79996af06bf572dbb61db18ad4b0 implemented an endpointslice index keyed by service ns/name and updated `TransformUpstreamsBuilder()` to use the index, PTAL.

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -45,38 +47,68 @@ func (p EndpointsSettings) ResourceName() string {
...
+		Services:                services,
 	}
 }
 
 type EndpointWithMd struct {
 	*envoy_config_endpoint_v3.LbEndpoint
 	EndpointMd EndpointMetadata
 }
+
+func (e EndpointWithMd) Equals(o EndpointWithMd) bool {
```
yuval-k, line 95: why is the reason for adding Equals?
I intentionally did not implement equals here, instead using the lbEpsEqualityHash.
there are two reasons for that:
- refelction has caused performance issues in the past
- i wanted the equality to work even if the order of the endpoints changes.

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -106,41 +138,73 @@ func NewEndpointsForUpstream(us UpstreamWrapper, logger *zap.Logger) *EndpointsF
...
+	if e.clusterName != o.clusterName {
+		return false
+	}
+	if e.UpstreamRef != o.UpstreamRef {
+		return false
+	}
+	if len(e.LbEps) != len(o.LbEps) {
+		return false
+	}
+	for locKey, eEndpoints := range e.LbEps {
```
yuval-k, line 183: similar to above - why was the change in the logic here needed?

projects/gateway2/proxy_syncer/cla.go
```diff
@@ -55,9 +54,14 @@ func prioritize(ep EndpointsForUpstream) *envoy_config_endpoint_v3.ClusterLoadAs
 				SubZone: loc.Subzone,
 			}
 		}
+		// Collect the LbEndpoints from the eps map
```
yuval-k, line 57: why this change? i think the logic is the same here?

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -45,38 +47,68 @@ func (p EndpointsSettings) ResourceName() string {
...
+	// Compare Endpoint Metadata Labels
+	if !reflect.DeepEqual(e.EndpointMd.Labels, o.EndpointMd.Labels) {
+		return false
+	}
+	return true
+}
+
 type EndpointsForUpstream struct {
-	LbEps       map[krtcollections.PodLocality][]EndpointWithMd
+	LbEps       map[krtcollections.PodLocality]map[string]EndpointWithMd
```
yuval-k, line 111: see my other comments re equals. i dont think this needs to be a map

nfuden: I do like the map for deduplication [here](https://github.com/solo-io/gloo/pull/10265/files#diff-1e41d85739cdfae37c6b9d7d40477a5030a000c07d7694c9fd3f0148df966aa2R148) that being said we can have the interstitial map be stored at rest as a slice to keep order more cleanly. 

Or do you think its cleaner to dedup later rather than at this point?

yuval-k: ah i see; yea that's ok then

projects/gateway2/proxy_syncer/endpoints.go
```diff
@@ -45,38 +47,68 @@ func (p EndpointsSettings) ResourceName() string {
...
 	endpointSettings := krt.NewSingleton(func(ctx krt.HandlerContext) *EndpointsSettings {
 		settings := krt.FetchOne(ctx, settings.AsCollection())
 		return &EndpointsSettings{
 			EnableAutoMtls: settings.Spec.GetGloo().GetIstioOptions().GetEnableAutoMtls().GetValue(),
 		}
 	})
 
+	// Create index on EndpointSlices by service name and namespace
+	endpointSlicesByService := krt.NewIndex(kubeEpSlices, func(es *discoveryv1.EndpointSlice) []types.NamespacedName {
+		svcName, ok := es.Labels["kubernetes.io/service-name"]
```
timflannagan, line 70: nit: The [discoveryv1.LabelServiceName](https://pkg.go.dev/k8s.io/api/discovery/v1#pkg-constants) constant is also available for us to use.

projects/gateway2/proxy_syncer/endpoints_test.go
timflannagan, line 1: Do we need a test that filters out EndpointSlices that have unready endpoints?

danehans: Added ^ test in commit d640df9.

projects/gateway2/proxy_syncer/endpoints.go
danehans: Commit d640df9 uses ^ constant.

projects/gateway2/proxy_syncer/cla.go
danehans: Commit d640df9 removes this change.

projects/gateway2/proxy_syncer/endpoints.go
danehans: Commit d640df9 removes this change.

projects/gateway2/proxy_syncer/endpoints.go
danehans: Commit d640df9 removes this change.

projects/gateway2/krtcollections/endpoints.go
```diff
@@ -192,66 +206,95 @@ func transformK8sEndpoints(ctx context.Context, inputs EndpointsInputs) func(kct
...
-				podNamespace := eps.Namespace
-				targetRef := addr.TargetRef
-				if targetRef != nil {
-					if targetRef.Kind == "Pod" {
-						podName = targetRef.Name
-						if targetRef.Namespace != "" {
-							podNamespace = targetRef.Namespace
+			for _, endpoint := range endpointSlice.Endpoints {
+				// Skip endpoints that are not ready
+				if endpoint.Conditions.Ready != nil && !*endpoint.Conditions.Ready {
```
yuval-k, line 254: note for future yuval: i think the previous impl allowed for non ready endpoints. i think this change is fine now; but may want to rethink and propagate readiness info to envoy instead, it as it might have impact on failover and/or draining behavior

yuval-k (approved):
lgtm; left a note to myself

nfuden (approved):

projects/gateway2/krtcollections/endpoints.go
danehans: @yuval-k I created https://github.com/solo-io/gloo/issues/10320 to track ^

PR #10271, opened 2024-11-04
Remove portal enablement in favor of general env setup for gloo deployment

inFocus7:
https://github.com/solo-io/solo-projects/pull/7127
WIP. need to verify through EE.

# Description

Remove unused portal enable field.

## API changes

- Removed (now) unused portal enablement helm value that was introduced in last merge.
- Added general environment variable definition used for gloo deployment. This will be used in the EE side to setup the GG_PORTAL_PLUGIN env.

# Context

See slack conversation [here](https://solo-io-corp.slack.com/archives/C07QCA4FG0M/p1730497880405319?thread_ts=1730481614.607899&cid=C07QCA4FG0M).

TL;DR
- We don't want Enterprise-only fields declared in OSS chart.
- This follows similar setup like done in settings where we have a defined variable/template that is set in EE in order to inject it into OSS templates.
- In EE we set an **unexported** automatically-defined `global.<>` value which is injected into this env vars template so that Gloo OSS knows when portal is enabled.
    - It is important that the `global` variable is undefined so users don't know about it and wrongfully modified, as it is for internal-use only.
- using this generic env vars, instead of doing a `if global.portaEnabled` here, because we don't want to expose that field's existence to users as it doesn't concern them. more context above and in the thread linked.

## Testing steps

Verifying on EE branch.

# Checklist:


inFocus7:
Helm doesn't allow including non-existent template. So had a failure setting up chart when defining the `gloo.extraDeploymentEnvVars` in Enterprise and not here.

I made it an empty field in OSS, as (at least for now) it's meant for Enterprise to be able to inject the Portal envVar. Need to see/hope that it's possible for the helper in EE to **overwrite** this one's setting...

install/helm/gloo/templates/1-gloo-deployment.yaml
```diff
@@ -273,6 +269,10 @@ spec:
           - name: HEADER_SECRET_REF_NS_MATCHES_US
             value: "true"
         {{- end}}
+{{- $extraEnvVars := include "gloo.extraDeploymentEnvVars" . | trim -}}
+{{- if $extraEnvVars -}}
+{{ $extraEnvVars | nindent 10 }}
+{{- end }}
```
inFocus7, line 275: doing this here, and in EE:
```tpl
{{- $_ := set .Values.global "gatewayPortalWebServerEnabled" (index .Values "gateway-portal-web-server" "enabled") }}
{{ define "gloo.extraDeploymentEnvVars" -}}
{{- if .Values.global.gatewayPortalWebServerEnabled -}}
- name: GG_PORTAL_PLUGIN
  value: "true"
{{- end -}}
{{- end -}}
```

so that it's generic enough if needed + users don't directly know about the .Values.global.gatewayPortalWebServerEnabled` if looking into the template and think they need to mess with that one.

---

If we want to focus on a better template setup, but technically exposing that undocumented global, we could simply directly just do the enablement of the exact env var in this template
```yaml
{{- if .Values.global.gatewayPortalWebServerEnabled -}}
- name: GG_PORTAL_PLUGIN
  value: "true"
{{- end -}}
```

PR #10272, opened 2024-11-04
extensibility for KRT and custom backend types

stevenctl:
this is built on top of  #10251  

# Description

Allows extensions to our core KRT collections, adding arbitrary Upstreams or Endpoints via KRTExtension.

Because these extensions could introduce new backend types, we also need a way to do custom backendRefs. 

## Code changes

* Add KRTExtensions interface to allow adding arbitrary new Upstreams or Endpoints
* Move some shared code out of `proxy_syncer` into `krtcollections` to avoid cyclic imports. 
* Add BackendRefResolver extensions to GatewayQueries for backend resolution
* Add BackendRefPlugins to translation to allow converting these custom backend types into Destinations
* Make it so the Istio transportsocket sets SNI in a way that Istio expects to match. 

## Testing steps

TODO


projects/gloo/pkg/plugins/edsupstream/us_plugin.go
```diff
@@ -0,0 +1,51 @@
...
+		Type: cluster_v3.Cluster_EDS,
+	}
+	// tell envoy to use ADS to resolve Endpoints
+	out.EdsClusterConfig = &cluster_v3.Cluster_EdsClusterConfig{
+		EdsConfig: &envoycore.ConfigSource{
+			ConfigSourceSpecifier: &envoycore.ConfigSource_Ads{
+				Ads: &envoycore.AggregatedConfigSource{},
+			},
+		},
+	}
```
yuval-k, line 49: we have a helper for that `SetEdsOnCluster`. the reason being that we had an envoy bug where we had to use REST for eds. it may not be relevant today though; @EItanya do you remember?

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -353,6 +353,21 @@ func setRouteAction(
 			}
 		}
 
+		fromPlugin := false
+		for _, bp := range pluginRegistry.GetBackendPlugins() {
+			if dest, ok := bp.ApplyBackendPlugin(obj, backendRef.BackendObjectReference); ok {
```
yuval-k, line 358: do we want to `break` if `ok`?

projects/gloo/pkg/plugins/edsupstream/us_plugin.go
stevenctl: added that

projects/gloo/pkg/plugins/static/plugin.go
```diff
@@ -75,6 +75,11 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 		return nil
 	}
 
+	// something else converted this to an EDS cluster
```
lgadban, line 78:
```suggestion
	// something else converted this to an EDS cluster so return early and let them handle this Upstream
```

lgadban: just to elaborate on the goal of this for people reading static plugin but unfamiliar with how EDS could be set on this. we may even want to explicitly mention the plugin name that could do this

lgadban (commented):
mostly LGTM, few comments from 1st pass

projects/gateway2/translator/httproute/gateway_http_route_translator.go
```diff
@@ -353,6 +353,22 @@ func setRouteAction(
 			}
 		}
 
+		fromPlugin := false
+		for _, bp := range pluginRegistry.GetBackendPlugins() {
```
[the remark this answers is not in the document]
stevenctl: Not sure what you mean? I don't expect many to exist period. WHen they do exist, they shouldn't overlap (at most 1 should return ok)

projects/gateway2/extensions/extensions.go
```diff
@@ -46,7 +54,7 @@ type K8sGatewayExtensionsFactory func(
 
 // NewK8sGatewayExtensions returns the Open Source implementation of K8sGatewayExtensions
 func NewK8sGatewayExtensions(
-	_ context.Context,
+	ctx context.Context,
```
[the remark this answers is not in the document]
stevenctl: holdover from when I was actually initializing a sample extension here, will revert

projects/gateway2/translator/httproute/gateway_http_route_translator.go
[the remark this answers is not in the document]
lgadban: it looks like we will immediately return as soon as one of these plugins can handle the backend so the first one will win; if they don't overlap then yeah it's no problem

projects/gloo/pkg/plugins/static/plugin.go
```diff
@@ -75,6 +76,11 @@ func (p *plugin) ProcessUpstream(params plugins.Params, in *v1.Upstream, out *en
 		return nil
 	}
 
+	// explicitly requested to be treated as an EDS cluster
+	if _, ok := in.GetMetadata().GetLabels()[edsupstream.InternalEDSLabel]; ok {
```
stevenctl, line 80: We cannot do `if cluster.isEds()  { return }` because at this point in the code many if not all static upstream WILL have been converted to EDS in `initializeCluster` b/c we see an Endpoint for them in the upstreamsToEndpoints map.

yuval-k (approved):

lgadban (approved):

PR #10282, opened 2024-11-06
docs/api/reference: generate api reference for Kubebuilder APIs

sam-heilbron:
# Description

Add automated Markdown generation for the following kubebuilder-based APIs:
- GatewayParameters
- DirectResponseAction

# Context
There will be a follow up to this PR to better format the generated code, but we want to introduce it first so it's visible to users.

## Interesting decisions
-

## Testing steps
-

## Notes for reviewers
-

# Checklist:


Makefile
```diff
@@ -178,6 +178,8 @@ install-go-tools: mod-download ## Download and install Go dependencies
 	# This version must stay in sync with the version used in CI: .github/workflows/static-analysis.yaml
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(LINTER_VERSION)
 	go install github.com/quasilyte/go-ruleguard/cmd/ruleguard@v0.3.16
+	# Kubebuilder docs generation
+	go install fybrik.io/crdoc@latest
```
jenshu, line 182: can we pin to a specific version?

Nadine2016 (approved):

Makefile
sam-heilbron: the reason I didn't do this is because the latest release is from august 2023, but the latest commit is from january.im still happy to pin it though, good call

Nadine2016 (approved):

jenshu (approved):

PR #10286, opened 2024-11-06
Feat/stevenctl/cls name

nfuden:
** this is the PR to use **

Originally https://github.com/solo-io/gloo/pull/10255

# Description

https://github.com/solo-io/solo-projects/issues/7105

This will make cluster names for kube services look like `kube-svc_name_namespace_port` or `kube-svc_reviews_bookinfo_8080`. If we add other resources besides a kube service, this also lets us get things like `istio-se_helloworld_backend-ns_5000`. 

This will only happen for Gateway v2 translation. We can disable it using the env var `GG_K8S_GW_LEGACY_CLUSTER_NAMES`.

## API changes

We can disable it using the env var `GG_K8S_GW_LEGACY_CLUSTER_NAMES`.

## Code changes

* XDS translator now takes options that can customize this new behavior. We instantiate translator separately for v2, so that is the only path to enable it. 
* Added another function to customize the generated name rather than modifying the existing cluster name function (it has over 100 usages! mostly in unit tests)
* Use internal labels to propagate the info in a structured way. 

## Docs changes

We should document this as a breaking change; users who rely on the old format should use the env var or update their EnvoyFilters or whatever config relies on XDS names... 

# Context

https://github.com/solo-io/solo-projects/issues/7105

## Interesting decisions

* Use internal labels to propagate the info in a structured way. This was done so we can support more than just Upstream_Kube upstreams. 
* Functional options for translator constructor. I didn't feel like updating the many references to this constructor, so varargs help. In hindsight, a second constructor could have also solved this. 

## Testing steps

TODO!


jenshu:
closing this and moving the PR back to gloo https://github.com/solo-io/gloo/pull/10357

Issue #10186, opened 2024-10-07
Gateway API conformance test matrix


jenshu:
### Do you have a suggestion for code improvement or tracking existing technical debt? Please describe.

Once we determine the Gateway API version support matrix for GG 1.17 and 1.18, we should run the nightly/weekly conformance tests with a matrix (of at least min/max supported GW API versions, or maybe all versions if feasible)

Some work is required to support the newer GW API versions:
1.1 https://github.com/solo-io/solo-projects/issues/6334
1.2 https://github.com/solo-io/gloo/issues/10175

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1207635280052738/1208493868718948) by [Unito](https://www.unito.io)

jenshu:
we can close this one since the linked issue above is still open

PR #10294, opened 2024-11-07
add krt debug endpoint

yuval-k:
This PR adds /snapshots/krt admin endpoint to 'see' krt collections.
Also 'fixes' the /snapshots/xds endpoint to overcome a panic in the solokit generic snapshot.


yuval-k:
/skip-changelog

non-user facing changes (debug interfaces)
and bug fixed on unreleased version; no need for changelog

yuval-k:
saw gw conformance fail with
```
2024-11-08T03:40:18.8730408Z         	Error:      	Received unexpected error:
2024-11-08T03:40:18.8731800Z         	            	the installed CRDs version is different from the suite version
```
looks like a flake ? some race installing the CRDs?

changelog/v1.18.0-beta34/krtdebug.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Add debug interfaces to KRT collections.
```
nfuden, line 3: I know this says interfaces but it looks like if you can enable everything the same way as the StartOptions Dev flag?

yuval-k: not sure i understand the question?

nfuden: Never mind I misread a comment and didnt run through functionality at the time

projects/gateway2/setup/ggv2setup.go
```diff
@@ -159,7 +159,8 @@ func StartGGv2WithConfig(ctx context.Context,
 		InitialSettings: initialSettings,
 		Settings:        settingsSingle,
 		// Useful for development purposes; not currently tied to any user-facing API
```
nfuden, line 161:
```suggestion
		// Dev flag may be useful for development purposes; not currently tied to any user-facing API
```

nfuden: basically when reading originally i was lazy and assumed that this applied to both dev and debugger

ci/kind/setup-kind.sh
```diff
@@ -18,7 +18,7 @@ IMAGE_VARIANT="${IMAGE_VARIANT:-standard}"
 # If true, run extra steps to set up k8s gateway api conformance test environment
 CONFORMANCE="${CONFORMANCE:-false}"
 # The version of the k8s gateway api conformance tests to run. Requires CONFORMANCE=true
-CONFORMANCE_VERSION="${CONFORMANCE_VERSION:-v1.1.0}"
+CONFORMANCE_VERSION="${CONFORMANCE_VERSION:-v1.2.0}"
```
sam-heilbron, line 21: With this change, do we need to consider running conformance tests with earlier versions (1.1) in our nightly tests? If so, I don't think we need to make that change in this PR, but it would be great to capture the work in an issue so we don't lose sight of it

pkg/version/enterprise_version_test.go
```diff
@@ -30,12 +30,21 @@ entries:
   gloo-ee:
   - apiVersion: v1
     version: 0.19.1
+    name: test
```
sam-heilbron, line 33: Is this change due to one of the dependency bumps?

projects/gateway2/krtcollections/endpoints.go
```diff
@@ -140,11 +158,11 @@ func (c EndpointsForUpstream) ResourceName() string {
 }
 
 func (c EndpointsForUpstream) Equals(in EndpointsForUpstream) bool {
-	return c.UpstreamRef == in.UpstreamRef && c.Port == in.Port && c.LbEpsEqualityHash == in.LbEpsEqualityHash && c.Hostname == in.Hostname
+	return c.UpstreamRef == in.UpstreamRef && c.ClusterName == in.ClusterName && c.Port == in.Port && c.LbEpsEqualityHash == in.LbEpsEqualityHash && c.Hostname == in.Hostname
```
sam-heilbron, line 161: Why is clusterName required if upstreamRef is already there? It feels like those two should capture the same information

projects/gateway2/krtcollections/pods.go
```diff
@@ -73,17 +78,17 @@ func NewNodeMetadataCollection(nodes krt.Collection[*corev1.Node]) krt.Collectio
 	})
 }
 
-func NewPodsCollection(ctx context.Context, istioClient kube.Client) krt.Collection[LocalityPod] {
+func NewPodsCollection(ctx context.Context, istioClient kube.Client, dbg *krt.DebugHandler) krt.Collection[LocalityPod] {
```
sam-heilbron, line 81: What is the behavior of this code if a nil debughandler is passed in?

projects/gateway2/proxy_syncer/upstreams.go
```diff
@@ -49,10 +49,11 @@ func (iu *PerClientEnvoyClusters) FetchClustersForClient(kctx krt.HandlerContext
 
 func NewPerClientEnvoyClusters(
 	ctx context.Context,
+	dbg *krt.DebugHandler,
```
sam-heilbron, line 52: We pass the DebugHandler to a lot of components, which seems quite valuable for being transparent about what is being passed around. Alternatively, we could pass CollectionOptions (one of with is the WithDebugging option). Is that something you considered or think is worth exploring?

projects/gloo/pkg/servers/admin/server.go
```diff
@@ -60,3 +65,30 @@ func getContentType(format string) string {
 		return "application/json"
 	}
 }
+
+// writeJSON writes a json payload, handling content type, marshaling, and errors
+func writeJSON(w http.ResponseWriter, obj any, req *http.Request) {
```
sam-heilbron, line 70: Does it make sense to use this function for our other snapshot endpoints and get ride of the respondJson method?

projects/gloo/pkg/servers/iosnapshot/history.go
```diff
@@ -197,6 +197,15 @@ func (h *historyImpl) GetXdsSnapshot(_ context.Context) SnapshotResponseData {
 	return completeSnapshotResponse(cacheEntries)
 }
 
+func (h *historyImpl) getXdsSnapshot(k string) (cache cache.Snapshot, err error) {
+	defer func() {
```
sam-heilbron, line 201: Do you think it's worth it to wrap this reocver operation around the entire server, instead of just this method? That way if there are future issues that arise, the control plane is better protected?

sam-heilbron (approved):
LGTM!

I didn't test this out yet, and left a few questions, but I'm happy to have this merge and we can iterate

projects/gateway2/krtcollections/endpoints.go
```diff
@@ -140,11 +158,11 @@ func (c EndpointsForUpstream) ResourceName() string {
...
 }
 
 func NewGlooK8sEndpoints(ctx context.Context, inputs EndpointsInputs) krt.Collection[EndpointsForUpstream] {
-	return krt.NewCollection(inputs.Upstreams, transformK8sEndpoints(ctx, inputs), krt.WithName("GlooK8sEndpoints"))
+	return krt.NewCollection(inputs.Upstreams, transformK8sEndpoints(ctx, inputs), krt.WithName("GlooK8sEndpoints"), krt.WithDebugging(inputs.Debugger))
```
sam-heilbron, line 165: Do you see any risk that developers now need to understand to use the WithDebugging option for constructing the krt client? It seems like it would be possible to forget, and then unintentionally not make certain components debuggable.

ci/kind/setup-kind.sh
jenshu: i had filed https://github.com/solo-io/gloo/issues/10186 awhile back

yuval-k: pulling latest istio pulled in gwapi 1.2 go code.
conformance tests then refused to run with older CRDs.. that's why i had to make this change. see my comment here: https://github.com/solo-io/gloo/pull/10294#issuecomment-2463703749

pkg/version/enterprise_version_test.go
yuval-k: yes, helm dep bump

projects/gateway2/krtcollections/endpoints.go
yuval-k: they should (and that's why i didn't originally include it here)! except for one thing - there was (and maybe still is) an envoy issue that made us change the cluster service name to include the upstream hash.

projects/gateway2/krtcollections/pods.go
yuval-k: nothing really. i call it with nil in the tests. if it is not nil, the collection will register itself to the debug handler.

projects/gloo/pkg/servers/admin/server.go
yuval-k: i don't mind it, but it seems that the output snapshot also support yaml outputs.. i'm not sure where that's used and don't want to break anything

projects/gloo/pkg/servers/iosnapshot/history.go
yuval-k: i believe go already has a recover for http handlers. i need it specifically on this level. otherwise this function won't work from gloo-enterprise, as the rate limit and extauth snapshots panic.

projects/gateway2/krtcollections/endpoints.go
yuval-k: it is possible to forget, but it's an easy fix. and i prefer not to abstract things too early. once of the annoyances i had in my last few PRs is over-abstracted code making it hard to make changes...

projects/gateway2/krtcollections/pods.go
sam-heilbron: Yeah that's why i was curious. I saw it in the tests, and just wanted to understand what could happen in live code if a dev accidentally used a nil

PR #10300, opened 2024-11-08
Add generic http upstream generator

ryanrolds:
# Description

* Add `v1helpers.NewTestHttpUpstreamWithHandler` which takes an `http.Handler` and run it behind Envoy
* Refactored HTTPBin generator to use new generic logic
* Added tests/examples for HTTPBin and httpHandler generators

# Context

Testing complex logic requires build mocks with complex state. This PR makes it easier to leverage Go packages that use the common http.Handler to implement servers. Also allows defining our own microservices that run inside of the testing framework.

# Checklist:


test/e2e/example_test.go
```diff
@@ -164,4 +167,75 @@ var _ = Describe("Example E2E Test For Developers", Label(), func() {
...
+
+			testContext = testContextFactory.NewTestContext()
+			testContext.SetUpstreamGenerator(v1helpers.NewTestHttpUpstreamWithHandler(mux))
+			testContext.BeforeEach()
+		})
+
+		It("Can count using custom handler", func() {
+			Eventually(func(g Gomega) {
+				requestBuilder := testContext.GetHttpRequestBuilder().
+					WithPath("test").
```
ashishb-solo, line 228: nit: can we make `test` a const?

ashishb-solo (approved):
looks good to me. one nit, feel free to ignore it

test/e2e/example_test.go
```diff
@@ -164,4 +167,75 @@ var _ = Describe("Example E2E Test For Developers", Label(), func() {
...
+				defer response.Body.Close()
+				body, err := io.ReadAll(response.Body)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(response.StatusCode).Should(Equal(http.StatusOK))
+
+				var headersResponse HeadersResponse
+				err = json.Unmarshal(body, &headersResponse)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				g.Expect(headersResponse.Headers).Should(HaveKeyWithValue("X-Test-Header", ConsistOf("test-value")))
```
sam-heilbron, line 200: Is it possible to use https://github.com/solo-io/gloo/blob/main/test/gomega/matchers/have_http_response.go#L48 instead?

test/e2e/example_test.go
```diff
@@ -164,4 +167,75 @@ var _ = Describe("Example E2E Test For Developers", Label(), func() {
...
 	})
 
+	Context("Using httpbin upstream generator to test", func() {
+		type HeadersResponse struct {
+			Headers map[string]interface{} `json:"headers"`
+		}
+
+		BeforeEach(func() {
+			testContext = testContextFactory.NewTestContext()
+			testContext.SetUpstreamGenerator(v1helpers.NewTestHttpUpstreamWithHttpbin)
```
sam-heilbron, line 177: I really like this as an example! Perhaps it's too wide, of a change, but would we ever change this upstreamGenerator to be our default type that all e2e tests use?

sam-heilbron (approved):
LGTM!

test/e2e/example_test.go
ryanrolds: Let me see how much breaks. One pain point is the http.handler generator writes all requests/responses to the channels while the generators that use `runTestServerWithHealthReply` don't write requests `/health`. I have no idea how many tests make asserts that would be affected by that difference.

test/e2e/example_test.go
ryanrolds: The matcher would need to consumer the body (which contains the request headers) instead of the headers. Would adding a new matcher for the httpbin /headers response be better? When we make assertions against headers we probably want to be specific about which set of headers we expect.

ryanrolds: I think I'm hearing you want things DRYed out a little bit more.

ryanrolds: @sam-heilbron, I add some matchers that allows what I think you were wanting:
* `ContainsDeepMapElements` - The expected nested structure is present in the actual value
* `JSONContains` - Given an expected and actual `[]byte`, it unmarshalls the JSON and confirms that the actual value contains the expected value 
* `HaveOKResponseWithJSONContaining` - Given a JSON string it asserts that the values in the string are present in the JSON body and the status is OK

Please let know if I missed the mark.

sam-heilbron (approved):

PR #10305, opened 2024-11-09
Makes the TCPRoute controller optional

### LINKED ISSUE #10304: Make TCPRoute Optional
### Gloo Edge Product

Open Source

### Gloo Edge Version

main

### Is your feature request related to a problem? Please describe.

https://github.com/solo-io/gloo/pull/10232 added TCPRoute support. Users may want to use Gateway API without TCPRoute. In this scenario, the controller manager will not start without the TCPRoute CRD installed.

### Describe the solution you'd like

Make the TCPRoute controller optional.

### Describe alternatives you've considered

_No response_

### Additional Context

_No response_

danehans:
# Description

Fixes #10304

## API changes

None

## Code changes

- Conditionally adds the Gateway API v1beta2 scheme.
- Adds a testing scheme that unconditionally includes the Gateway API v1beta2 scheme and updates tests throughout to use  it.
- Adds `CRDs sets.Set[string]` to the `GatewayConfig` type.  `CRDs` defines the set of discovered Gateway API CRDs.
- Updates `addIndexes()` to conditionally add the TCPRoute index based on the presence of the tcproutes.gateway.networking.k8s.io CRD.
- Updates `watchTcpRoute()` to conditionally watch TCPRoute kinds based on the presence of the tcproutes.gateway.networking.k8s.io CRD.
- Updates `NewControllerBuilder()` to conditionally add the TCPRoute controller based on the presence of the tcproutes.gateway.networking.k8s.io CRD.
- Removes `requiredCRDsExist` from the NewData type since it's no longer needed (not making an API call to check for the .tcproutes.gateway.networking.k8s.io CRD). Removes all references to this field.

## CI changes

Updates the scheme used by tests.

## Docs changes

N/A

# Context

## Testing steps

- Create a kind cluster using the CI script.
- Run e2e tests.
- Tests pass
- Remove the tcproutes.gateway.networking.k8s.io CRD (installed by default).
- Run e2e tests.
- All tests except TCPRoute pass.

# Checklist:


pkg/schemes/scheme.go
```diff
@@ -22,7 +24,6 @@ import (
 var SchemeBuilder = runtime.SchemeBuilder{
 	// K8s Gateway API resources
 	gwv1.Install,
-	gwv1a2.Install,
```
yuval-k, line 25: whats the harm in added these apis to the schema?

projects/gateway2/controller/start.go
```diff
@@ -98,9 +99,16 @@ func NewControllerBuilder(ctx context.Context, cfg StartConfig) (*ControllerBuil
 	}
 	ctrl.SetLogger(zap.New(opts...))
 
+	scheme := glooschemes.DefaultScheme()
+
+	// Extend the scheme if the TCPRoute CRD exists.
+	if err := glooschemes.AddGatewayV1A2Scheme(cfg.RestConfig, scheme); err != nil {
```
nfuden, line 105: It feels like this is something we will hit with every iteration of gateway api right? 

Should we do something where we try to discover the current gatewayapi version via CRDs, emit that as a log and then add appropriate pieces to the scheme?

Basically exactly whats being done here except changing names a bit. 
So TCPRoute's presence would indicate that we have 1.1 installed but in the future we might start with some other indicator to see if say we are on the future 1.3 or something

pkg/schemes/scheme.go
danehans: @yuval-k thanks for the review. From my understanding, the TCPRoute type in the scheme causes the controller-runtime manager's cache to attempt to create an informer for it. If the TCPRoute CRD is not installed in the cluster, it will lead to the controller failing to start with errors like:
```sh
"error":"no matches for kind \"TCPRoute\" in version \"gateway.networking.k8s.io/v1alpha2\""
```

danehans: @yuval-k and I discussed this conversation and agreed to proceed as-is. The gwv1a2 scheme is used to signal whether to manage TCPRoute resources instead of making an API call to check for the CRD (and caching the response). @stevenctl is working on refactoring the query pkg to use krt which provides a mechanism to dynamically add informers, e.g. TCPRoute, based on the presence of CRDs that will supersede this work.

projects/gateway2/controller/start.go
danehans: >It feels like this is something we will hit with every iteration of gateway api right?

By "iteration", I'm assuming you mean every release. If so, resources do not graduate for every release. This function would need to be updated based on resources graduating. For example, when TCPRoute graduates to the gwv1 scheme, this function is no longer needed unless we want to conditionally add the gw1v1a2 scheme based on the presence of some other gw1v1a2 resource.

> Should we do something where we try to discover the current gatewayapi version via CRDs, emit that as a log and then add appropriate pieces to the scheme?

We currently assume the presence of the gwv1 resources, e.g. GatewayClass, HTTPRoute, etc. and unconditionally add informers, caches, etc. to the controller-runtime manager for these resources. I focused the approach on TCPRoute since the associated issue is for making the TCPRoute controller optional. If you have a strong opinion, do you mind creating an issue and linking it here?

>So TCPRoute's presence would indicate that we have 1.1 installed

Slightly different, TCPRoute's presence indicates the TCPRoute CRD is installed. A user may only want HTTPRoute with GW API 1.1 so (currently) they would `kubectl delete crd/tcproutes.gateway.networking.k8s.io` after installing the GW API experimental CRDs (`kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.1.0/experimental-install.yaml`). As the kind script does, we should consider providing users the option of installing GW API CRDs from either the `standard` or `experimental` release channel. __Note:__ The `experimental` release channel includes all the resources, fields, etc. of the `standard` channel.

danehans: @nfuden I created https://github.com/solo-io/gloo/issues/10312 to track the need for providing GW API CRD installation optionality, PTAL.

nfuden: ah yeah its a great point about experimental apis

pkg/schemes/extended_scheme.go
```diff
@@ -0,0 +1,54 @@
...
+		if err := gwv1a2.Install(scheme); err != nil {
+			return fmt.Errorf("error adding Gateway API v1alpha2 to scheme: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// Helper function to check if a CRD exists
+func CRDExists(restConfig *rest.Config, group, version, kind string) (bool, error) {
```
nfuden, line 32: future thought: persist a map?
The memory footprint is minimal and would make future calls ez

danehans: `CRDs sets.Set[string]` was added to the GatewayConfig type for this purpose. To your point, this field should be added to the ControllerBuilder type and initialized in `NewControllerBuilder()`. I created https://github.com/solo-io/gloo/issues/10321 to track this tech debt.

nfuden (approved):

nfuden (approved):

sam-heilbron (approved):

jenshu (approved):

danehans:
The failing CI test is passing when run locally:

```sh
$ go test -v ./projects/gateway2/setup
--- PASS: TestScenarios (32.05s)
    --- PASS: TestScenarios/failover-default-diffns (1.10s)
    --- PASS: TestScenarios/failover-default (1.15s)
    --- PASS: TestScenarios/failover (1.13s)
    --- PASS: TestScenarios/happypath (1.13s)
    --- PASS: TestScenarios/listeneropts (1.06s)
PASS
ok  	github.com/solo-io/gloo/projects/gateway2/setup	32.117s
```

Rebasing to try resolving the CI failure.

PR #10315, opened 2024-11-11
Validate config using STDIN and /dev/fd/0

ryanrolds:
# Description

To address an issue with large configs passed as arguments causing a `arguments to long` error during validation; This PR adjusts how the config is provided to Envoy. The config is now fed to the process via STDIN and the `--config-yaml` is set to the file descriptor for STDIN.

## Code changes

* Switch to STDIN for injecting the config during validation
* Pass CLUSTER_NAME to kube2e test running
* Adjusted instructions for running kube3e tests locally.

# Context

A [customer with a large config](https://github.com/solo-io/solo-projects/issues/7089) reported the error.

## Interesting decisions

Initially there were discussions about saving a file inside of the container, but concerns about read-only root filesystems were raised. A volume would address that issue, but is a more complex solution. To avoid the need for a volume in some cases, I opted to use STDIN to provide the config to the program and read the STDIN FD to the config.

## Testing steps

I've applied the large config yaml in the test with and without the fix confirming that is was broken and is now fixed.

```
% ./ci/kind/setup-kind.sh
% helm upgrade --install -n gloo-system --create-namespace gloo ./_test/gloo-1.0.0-ci1.tgz --values ./test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
% make -B kind-reload-gloo
% kubectl apply -f test/kube2e/gloo/artifacts/large-configuration.yaml
```

## Notes for reviewers

# Checklist:


Makefile
```diff
@@ -253,7 +253,7 @@ install-protoc:
 
 .PHONY: test
 test: ## Run all tests, or only run the test package at {TEST_PKG} if it is specified
-	$(GINKGO_ENV) $(DEPSGOBIN)/ginkgo -ldflags=$(LDFLAGS) \
+	CLUSTER_NAME=${CLUSTER_NAME} $(GINKGO_ENV) $(DEPSGOBIN)/ginkgo -ldflags=$(LDFLAGS) \
```
andy-fong, line 256: nit: This seems unnecessary because if CLUSTER_NAME is already an env variable, ginkgo should already picked it up.

andy-fong (approved):

changelog/v1.18.0-beta34/validate-large-configs.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/7089
+    resolvesIssue: false
+    description: >-
+      Fix the validation of large configurations. 
+      Previously, the configuration was being passed as an argument and we were 
+      seeing `argument list too long`. 
+      This change passes the configuration as a file instead, using STDIN (/dev/fd/0)
+      avoiding the need to store the file on disk.
```
nfuden, line 10: This is written very much for engineers in the format I would think of for the pr description itself. 
However this is customer facing.

```suggestion
      Fix the validation of large configurations when using envoy validation.
      Previously if the configuration grew too large translation would be blocked unless Transformation Validation was disabled.
      This change passes the configuration as a file instead, using STDIN (/dev/fd/0)
      avoiding the need to store the file on disk.
```

Makefile
ryanrolds: I confirmed you are correct. Removed.

test/kube2e/gateway/gateway_suite_test.go
```diff
@@ -72,8 +72,11 @@ func StartTestHelper() {
 		installGloo()
 	}
 
+	clusterName := os.Getenv("CLUSTER_NAME")
+	Expect(clusterName).NotTo(BeEmpty(), "CLUSTER_NAME must be set")
+
 	// We rely on the "new" kubernetes/e2e setup code, since it incorporates controller-runtime logging setup
-	clusterContext := cluster.MustKindContext("kind")
+	clusterContext := cluster.MustKindContext(clusterName)
```
nfuden, line 79: definitely an improvement. How would you feel about migrating to an approach like https://github.com/solo-io/gloo/blob/feat/stevenctl/cls-name/test/kubernetes/e2e/test.go#L54 so we dont have the os env stashed in this file

nfuden (approved):

test/kube2e/gloo/large_configuration_test.go
```diff
@@ -0,0 +1,43 @@
+package gloo_test
```
sam-heilbron, line 1: This is an "old-style" e2e test. Is that intentional? We should be able to migrate to the new style fairly easily https://github.com/solo-io/gloo/tree/main/test/kubernetes/e2e.

In fact this could live in: https://github.com/solo-io/gloo/tree/main/test/kubernetes/e2e/features/validation/full_envoy_validation

ryanrolds: It wasn't intentional, I thought that was the current tests. Where can I read about what tests suites are current?

I will move it.

ryanrolds: Should I add a deprecation notice to the kube2e/README.md?

sam-heilbron: Yeah great callout, and something we should have done from the beginning. In that notice, could you link https://github.com/solo-io/solo-projects/issues/6303? It's strange to have the cross-repo link so we could also just migrate that issue to gloo oss and link it. It was created in the private repo since that was where we were creating a lot of our "tasks"

timflannagan:
Nice changelog entry description!

test/kube2e/gloo/large_configuration_test.go
ryanrolds: I assume you want the issue transferred. Tried and it's not showing the `gloo` repo, only another private repo. I assume I don't have permissions or private to public issue transfer isn't allowed.

Issue #10323, opened 2024-11-12
Remove Endpoints Resource from Helm RBAC Manifest


danehans:
### Do you have a suggestion for code improvement or tracking existing technical debt? Please describe.

When ".Values.kubeGateway.enabled" is set, the "gloo-gateway.fullname" ClusterRole includes the Endpoints resource even though https://github.com/solo-io/gloo/pull/10265 replaces Endpoints with EndpointSlice.

### Describe the solution you'd like

Remove the Endpoints resource from Helm RBAC manifest since https://github.com/solo-io/gloo/pull/10265 replaces Endpoints with EndpointSlice.

danehans:
This issue is specific to GGv2 configured with `.Values.kubeGateway.enabled`. @sam-heilbron I see this issue in other Helm RBAC templates.

PR #10324, opened 2024-11-12
Removes endpoints from the 44-rbac.yaml Helm template

danehans:
Removes endpoints from the `44-rbac.yaml` Helm template since the gw API controller uses EndpointSlices instead of Endpoints due to https://github.com/solo-io/gloo/pull/10265.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/10323


sam-heilbron (approved):

timflannagan (approved):

jenshu:
> dont we hit endpoints in classic?
this rbac file is only used for k8s gw

PR #10335, opened 2024-11-13
Remove hard-coded gateway namespace

ashishb-solo:
# Description

We need the gateway's namespace to be configured dynamically in the test using the "-n" flag in suite.go. To do this, we move that to a separate file and apply it separately using that flag. All the other resources are still preserved in their original namespaces.

## API changes

None

## Code changes

Tests only

## CI changes

Fixed some tests

## Docs changes

None

# Context

Solo-projects tests were failing because they use a different namespace, so the hard-coded values in this test were wrong.

## Interesting decisions

None

## Testing steps

So far tests are working on gloo. Need to test in solo-projects as well. In the mean time, I can try to change the namespace in the test that's run and see if things still work.

## Notes for reviewers

None

# Checklist:


danehans:
@ashishb-solo a changelog is needed (`make generate-changelog`).

jenshu:
@ashishb-solo are you still working on this or can it be closed?

PR #10364, opened 2024-11-19
Add note about CNCF donation

yuval-k:


nfuden (approved):

README.md
```diff
@@ -8,6 +8,10 @@
   An Envoy-Powered Kubernetes-Native API Gateway
 </h1>
 
+
+# 🚧 NOTE: This project is in the process of being donated to the CNCF. It is being re-factored as part of the process. 🚧
```
danehans, line 12: @yuval-k does it make sense to state that APIs, even GA APIs, may have breaking changes?

yuval-k: APIs wouldn't break; they may be deprecated, and there will be a migration path

PR #10397, opened 2024-12-10
feat: initial impl of removing usage gloo proxy

yuval-k:
this is part of https://github.com/k8sgateway/k8sgateway/issues/10363

To enable opaque extendibility, we can't use  protobuf based IR.

This pr changes the plugin model:
- Each plugin can contribute CRDs that will be translated opaquely
- Each plugin translates its own CRDs opaquely to the rest of the system.
- The system handles policy attachment for the plugins

2 plugin stages:
- Gateway api to gateway IR (uses krt); this creates a gateway IR with all the policies attached
- gateway IR to xDS (pure function, no IO)

Added some example CRDs, by no means in final form.

still WIP, there's a bunch of intentional panics in the code; created draft PR for early feedback


projects/gateway2/translator/httproute/delegation_helpers.go
```diff
@@ -47,10 +48,11 @@ func filterDelegatedChildren(
 		// make a copy; multiple parents can delegate to the same child so we can't modify a shared reference
 		clone := c.Clone()
 
-		child, ok := clone.Object.(*gwv1.HTTPRoute)
+		origChild, ok := clone.Object.(*ir.HttpRouteIR)
```
shashankram, line 51: s/origChild/childIR

projects/gateway2/translator/irtranslator/delegation_helpers.go
```diff
@@ -0,0 +1,277 @@
+package irtranslator
```
shashankram, line 1: Hard to know how many files were moved and edited vs reimplemeted. Did you use `git mv` to preserve history?

yuval-k: remove these files they are duplicates

projects/gateway2/krtcollections/policy.go
```diff
@@ -0,0 +1,717 @@
...
+	h := &RoutesIndex{policies: policies, refgrants: refgrants, upstreams: upstreams}
+	h.httpRoutes = krt.NewCollection(httproutes, h.transformHttpRoute, krtopts.ToOptions("http-routes-with-policy")...)
+	hr := krt.NewCollection(h.httpRoutes, func(kctx krt.HandlerContext, i ir.HttpRouteIR) *RouteWrapper {
+		return &RouteWrapper{Route: &i}
+	}, krtopts.ToOptions("routes-http-routes-with-policy")...)
+	tr := krt.NewCollection(tcproutes, func(kctx krt.HandlerContext, i *gwv1a2.TCPRoute) *RouteWrapper {
+		t := h.transformTcpRoute(kctx, i)
+		return &RouteWrapper{Route: t}
+	}, krtopts.ToOptions("routes-tcp-routes-with-policy")...)
+	h.routes = krt.JoinCollection([]krt.Collection[RouteWrapper]{hr, tr})
```
ilrudie, line 388: missed adding our opts and a name here
```suggestion
	h.routes = krt.JoinCollection([]krt.Collection[RouteWrapper]{hr, tr}, krtopts.ToOptions(name: "all-routes-with-policy")...)
```

projects/gateway2/controller/start.go
```diff
@@ -132,39 +132,49 @@ func NewControllerBuilder(ctx context.Context, cfg StartConfig) (*ControllerBuil
...
+	k8sSecrets := krt.NewCollection(k8sSecretsRaw, func(kctx krt.HandlerContext, i *corev1.Secret) *ir.Secret {
+		res := ir.Secret{
+			ObjectSource: ir.ObjectSource{
+				Group:     "",
+				Kind:      "Secret",
+				Namespace: i.Namespace,
+				Name:      i.Name,
+			},
+			Obj:  i,
+			Data: i.Data,
```
ilrudie, line 163: Why would we store the entire object and its data? Could we just embed the object instead?

projects/gateway2/controller/start.go
```diff
@@ -132,39 +132,49 @@ func NewControllerBuilder(ctx context.Context, cfg StartConfig) (*ControllerBuil
...
+	}
+	commoncol := common.CommonCollections{
+		Client:   cfg.Client,
+		KrtOpts:  cfg.KrtOptions,
+		Secrets:  krtcollections.NewSecretIndex(secrets),
+		Pods:     cfg.AugmentedPods,
+		Settings: cfg.Settings,
+	}
+	k8sGwExtensions := registry.AllPlugins(ctx, commoncol)
 	if err != nil {
```
ilrudie, line 178: The code which generated this err got removed. It is now referencing an error created way back around ln 125 which is already checked

projects/gateway2/controller/start.go
yuval-k: no big reason; mainly follow the pattern with the rest of the system where the objects themselves are opaque. I agree that it is less important with secrets.

projects/gateway2/extensions2/plugin/plugin.go
```diff
@@ -0,0 +1,103 @@
...
+	envoy_config_endpoint_v3 "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	"github.com/solo-io/gloo/projects/gateway2/ir"
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/solo-io/gloo/projects/gateway2/reports"
+	"istio.io/istio/pkg/kube/krt"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type AttachmentPoints uint
```
nfuden, line 16:
```suggestion
type AttachmentPoint uint
```

What do you think of singular?

projects/gateway2/extensions2/plugins/directresponse/direct_response_plugin.go
```diff
@@ -0,0 +1,152 @@
...
+	"istio.io/istio/pkg/kube/krt"
+)
+
+type directResponse struct {
+	ct   time.Time
+	spec v1alpha1.DirectResponseSpec
+}
+
+// in case multiple policies attached to the same resouce, we sort by policy creation time.
+func (d *directResponse) CreationTime() time.Time {
```
nfuden, line 29: It feels like this will be very repeated and build some spaghetti. Is there a way we can make this not a policy requirement or have a standard load of it into a policy?

projects/gateway2/api/applyconfiguration/utils.go
```diff
@@ -0,0 +1,98 @@
...
+	internal "github.com/solo-io/gloo/projects/gateway2/api/applyconfiguration/internal"
+	v1alpha1 "github.com/solo-io/gloo/projects/gateway2/api/v1alpha1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	schema "k8s.io/apimachinery/pkg/runtime/schema"
+	testing "k8s.io/client-go/testing"
+)
+
+// ForKind returns an apply configuration type for the given GroupVersionKind, or nil if no
+// apply configuration type exists for the given GroupVersionKind.
+func ForKind(kind schema.GroupVersionKind) interface{} {
```
nfuden, line 16: How would someone extend this?

yuval-k: this whole folder is generated code;
i created the client generated code so i can use fake client in the translation yaml tests

yuval-k: also see https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/update-codegen.sh

projects/gateway2/extensions2/plugin/plugin.go
yuval-k: this is an bit flags that may have info on more than one point

projects/gateway2/krtcollections/builtin.go
```diff
@@ -0,0 +1,236 @@
...
+		route.RequestMirrorPolicies = append(route.RequestMirrorPolicies, &envoy_config_route_v3.RouteAction_RequestMirrorPolicy{
+			Cluster:         up.ClusterName(),
+			RuntimeFraction: fraction,
+		})
+
+		return nil
+	}
+}
+
+func getFactionPercent(f gwv1.HTTPRequestMirrorFilter) *envoy_config_core_v3.RuntimeFractionalPercent {
```
ilrudie, line 148:
```suggestion
func getFractionPercent(f gwv1.HTTPRequestMirrorFilter) *envoy_config_core_v3.RuntimeFractionalPercent {r
```

projects/gateway2/krtcollections/setup.go
```diff
@@ -0,0 +1,90 @@
...
+	gwv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gwv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+func InitCollectionsWithGateways(ctx context.Context,
+	kubeRawGateways krt.Collection[*gwv1.Gateway],
+	httpRoutes krt.Collection[*gwv1.HTTPRoute],
+	tcproutes krt.Collection[*gwv1a2.TCPRoute],
+	refgrants *RefGrantIndex,
+	extensions extensionsplug.Plugin, istioClient kube.Client, krtopts krtutil.KrtOptions) (*GatweayIndex, *RoutesIndex, krt.Collection[ir.Upstream], krt.Collection[ir.EndpointsForUpstream]) {
```
ilrudie, line 22: nit: formatting

ilrudie (commented):
couple of comments from a partial review pass

projects/gateway2/extensions2/plugins/upstream/plugin.go
```diff
@@ -80,14 +85,25 @@ type plugin2 struct {
...
+func registerTypes(ourCli versioned.Interface) {
+	skubeclient.Register[*v1alpha1.Upstream](
+		v1alpha1.UpstreamGVK.GroupVersion().WithResource("upstreams"),
+		v1alpha1.UpstreamGVK,
+		func(c skubeclient.ClientGetter, namespace string, o metav1.ListOptions) (runtime.Object, error) {
+			return ourCli.GatewayV1alpha1().Upstreams(namespace).List(context.Background(), o)
+		},
+		func(c skubeclient.ClientGetter, namespace string, o metav1.ListOptions) (watch.Interface, error) {
+			return ourCli.GatewayV1alpha1().Upstreams(namespace).Watch(context.Background(), o)
+		},
```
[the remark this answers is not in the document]
yuval-k: the motivation here is that i can use fake client to test these;
i think client getter basically does the same thing using the istio client.. cc @stevenctl WDYT?

projects/gateway2/krtcollections/builtin.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (d *builtinPlugin) CreationTime() time.Time {
+	return time.Time{}
+}
+
+func (d *builtinPlugin) Equals(in any) bool {
+	// we don't really need equality check here, because this policy is embedded in the httproute,
+	// and we have generation based equality checks for that already.
+	return true
```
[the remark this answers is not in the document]
yuval-k: it should be true, false will cause excessive jitter.
in this specific case, it is not a problem because a built in type is embdded in the httproute CR, so we will catch it in the httproute equals (that is based off of the generation).

i'll try to see if its easy to add equals here for readability sake

projects/gateway2/api/v1alpha1/route_policy_types.go
```diff
@@ -0,0 +1,185 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:metadata:labels={app=gateway,app.kubernetes.io/name=gateway}
+// +kubebuilder:resource:categories=gateway,shortName=hlp
+// +kubebuilder:subresource:status
+// +kubebuilder:metadata:labels="gateway.networking.k8s.io/policy=Direct"
+type HttpListenerPolicy struct {
```
jenshu, line 94: can we put HttpListenerPolicy / ListenerPolicy / Upstream each in its own file? this file seems to be for route policies

jenshu (commented):
review part 1

projects/gateway2/krtcollections/builtin.go
yuval-k: the suggestion has an `r` in the end so i can't apply it. will push this in shortly

projects/gateway2/api/applyconfiguration/utils.go
yuval-k: this specific folder is related to server-side-apply and managed fields; its not the regular client.

projects/gateway2/krtcollections/builtin.go
ilrudie: updated, but you probably already pushed, good either way

projects/gateway2/translator/httproute/delegation_helpers.go
```diff
@@ -72,30 +78,31 @@ func filterDelegatedChildren(
 			}
 
 			for _, match := range rule.Matches {
+				match := *match.DeepCopy()
```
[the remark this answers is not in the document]
yuval-k: to make sure we are not changing what's in the krt cache. 
I believe controller runtime used to do that automatically for us

projects/gateway2/krtcollections/builtin.go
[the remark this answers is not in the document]
yuval-k: it's here because policy attachment happens in the collections in policy.go..
I do agree that it might worth moving policy.go to translator package; but prefer to do it in a follow-up

projects/gateway2/translator/plugins/routeoptions/route_options_plugin.go
lgadban, line 1: do we want to completely delete these plugins already?
alternatively we can keep them around until the necessary functionality is ported into the new framework

nfuden (approved):
A great start that we can start reimplementing alot such as the required checks that we havent quite gotten passing

PR #10413, opened 2024-12-22
more cleanup

yuval-k:
bring in control plane to gateway
don't start gloo anymore.
use upstream go control plane
remove gloov1 objects from gateway (settings still remain)

Note : now that gloo doesn't start anymore we lose the gloo features that are not the control plane. these come to mind:
- stats \ stats page. Gloo uses deprecated `opencensus`. so this is an opportunity to use `otel` instead. Stats should be served on their own port.
- admin page (for snapshots, pprof, changing log level). I don't recommend using the go-utils admin page as it mixes stats and admin page. In addition i don't think admin page should be in an external dependency. it should be part of this project where it is easy to change.
- leader election - we didn't have it before in gateway, but have it even less now. relevant for status writing.
- webhook validation
- snapshot history (not sure what this one is; might not be needed with the krt and xds snapshots available)
- snapshot sanitizers (we can bring these back, though i'm not sure we need to with the changes to envoy over the years)


lgadban:
>snapshot history (not sure what this one is; might not be needed with the krt and xds snapshots available)

provides GME-style snapshots for debugging etc but is dependent on the gloov1 apiSnapshot schema.
We should replace it with an equivalent solution that is backed by krt and/or xds snapshot (probably both, as we want to capture the input resources which feed krt)

projects/gateway2/setup/ggv2setup.go
```diff
@@ -128,23 +146,18 @@ func StartGGv2WithConfig(ctx context.Context,
 	}, krt.WithName("GlooSettingsSingleton"))
 
 	logger.Info("creating reporter")
```
lgadban, line 148:
```suggestion
```

projects/gloo/pkg/setup/setup.go
```diff
@@ -28,48 +19,7 @@ func Main(customCtx context.Context) error {
...
-				// Don't die if we fall from grace. Instead we can retry leader election
-				// Ref: https://github.com/solo-io/gloo/issues/7346
-				contextutils.LoggerFrom(ctx).Errorf("lost leadership")
-			},
-		},
-	})
+	return ggv2setup.StartGGv2(ctx, nil, nil)
 }
 
 func newSetupFunc(setupOpts *bootstrap.SetupOpts) setuputils.SetupFunc {
```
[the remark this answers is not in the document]
lgadban: discussed offline; yes this is unnecessary!

lgadban (approved):
LGTM, good iteration on cleanup, will continue to be an ongoing process

tests are failing so can reapprove if larger changes are needed to get things passing

yuval-k:
re failing tests:
- regression gloo & (old) gateway fail because gloo no longer turned on.
- the Kubernetes E2E will need to be fixed one by one, as they have both new gateway and old gloo test, so i figure that can be done in a follow-up.

projects/gateway2/translator/gateway/gateway_translator.go
```diff
@@ -1,4 +1,4 @@
-package translator
+package gatewaytranslator
```
jenshu, line 1: can we name the package `gateway` (same as directory name) otherwise we will get confused when importing it

projects/gateway2/translator/gateway/gateway_translator_test.go
```diff
@@ -1,4 +1,4 @@
-package translator_test
+package gatewaytranslator_test
```
jenshu, line 1: same comment - `gateway_test`

projects/gateway2/translator/gateway/translator_case_test.go
```diff
@@ -1,4 +1,4 @@
-package translator_test
+package gatewaytranslator_test
```
jenshu, line 1:
```suggestion
package gateway_test
```
same comment

projects/gateway2/krtcollections/policy_test_utils.go
```diff
@@ -0,0 +1,7 @@
+package krtcollections
+
+import "github.com/solo-io/gloo/projects/gateway2/ir"
+
+func (h *RoutesIndex) TEST() []ir.HttpRouteIR {
```
jenshu, line 5: what is this file for, do we need this?

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
```diff
@@ -33,20 +32,7 @@ func (s *ProxyTranslator) syncXds(
 	// TODO: me may need to copy this to not change krt cache.
 	// TODO: this is also may not be needed now that envoy has
 	// a default initial fetch timeout
-	snap.MakeConsistent()
-	s.xdsCache.SetSnapshot(proxyKey, snap)
+	// snap.MakeConsistent()
```
jenshu, line 35: do we need to add this back later?

projects/gateway2/proxy_syncer/xdswrapper.go
```diff
@@ -5,61 +5,80 @@ import (
...
 	return p
 }
 
 var _ krt.ResourceNamer = XdsSnapWrapper{}
 
 func (p XdsSnapWrapper) Equals(in XdsSnapWrapper) bool {
-	return p.snap.Equal(in.snap)
+	// check that all the versions are the equal
+	for i, r := range p.snap.Resources {
+		if r.Version != in.snap.Resources[i].Version {
```
jenshu, line 41: are the snapshot resources guaranteed to be sorted?

jenshu (approved):
i'm good with merging so we can iterate

projects/gateway2/proxy_syncer/xdswrapper.go
yuval-k: yes, its an array where the index is the resource type.

projects/gateway2/proxy_syncer/kube_gw_translator_syncer.go
yuval-k: i think not, but we can discuss.
i think we can build the code such that, we no longer need sanitizers

projects/gateway2/krtcollections/policy_test_utils.go
yuval-k: i can deleted that, i used it temporary for a test

projects/gateway/pkg/validation/validator.go
```diff
@@ -288,8 +289,8 @@ func (v *validator) validateProxiesAndExtensions(ctx context.Context, snapshot *
 
 	validatingK8sGateway := isK8sGatewayProxy(opts.Resource)
 	if validatingK8sGateway {
```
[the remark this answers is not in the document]
yuval-k: this code will get deleted anyway (its from ggv2 validation where a gateway is translated to a proxy); just didn't want to do this in this PR.

lgadban (approved):

PR #10451, opened 2025-01-14
install: Remove unsupported legacy sub-directories

timflannagan:
# Description

Removes several install/ directories that are no longer maintained. Related to https://github.com/k8sgateway/k8sgateway/issues/10423 for the new chart.

# Checklist:


install/.gitignore
timflannagan, line 1: This was an empty .gitignore file. I think we might introduce this back into source over time if it's necessary:

```
charts/
requirements.lock
Chart.yaml
Chart.lock
...
```

timflannagan:
I was tempted to remove https://github.com/k8sgateway/k8sgateway/tree/main/install/test & the install/utils package if we're redesigning the control plane chart from the ground up. Thoughts? I can add a line item that introduces helm tests over time in the tracker.

jenshu (approved):

jenshu:
> I was tempted to remove https://github.com/k8sgateway/k8sgateway/tree/main/install/test 

hmm, maybe keep that for now, until we have a replacement helm chart; i still use the existing helm chart to test things locally

PR #10458, opened 2025-01-15
remove legacy Settings type

lgadban:
The controller and deployer had a few dependencies on the legacy Gloo Settings type. This commit removes the usage of the Settings type and KRT collection and replaces it with a 1:1 copy that is non-KRT driven and purely derived from environment variables.

These can be set with "KGW_" prefixing the setting name (from the struct)
e.g.
```
KGW_ISTIOINTEGRATION: true
KGW_ENABLEAUTOMTLS: true
```

### Testing

As e2e testing is currently broken, this can be manually tested by setting the appropriate env vars on the deployment
```
        - name: KGW_ISTIOINTEGRATION
          value: "true"
        - name: KGW_ENABLEAUTOMTLS
          value: "true"
```

While e2e testing is disable, the ggv2test envtests DO test with istio features enabled, so we do have some proof the feature still works


projects/gateway2/extensions2/settings/settings.go
```diff
@@ -0,0 +1,21 @@
+package settings
+
+import (
+	"github.com/kelseyhightower/envconfig"
+)
+
+type Settings struct {
```
jenshu, line 7: question: do we not want this to be a new CR? what are the pros/cons of passing through env vars vs. using a CR to hold these values?

lgadban: eventually we may want a new CR, but if we do so, it will need to be a kgw specific CRD.

this is a quick (possibly long) fix to break the last remaining Gloo-type dependency, which will allow us to move forward with the removal of all the legacy gloo code.

the Settings feature overall will be tracked in a separate follow on issue, which I can link here shortly when I create it

lgadban: see: https://github.com/k8sgateway/k8sgateway/issues/10465

projects/gateway2/extensions2/settings/settings.go
```diff
@@ -0,0 +1,21 @@
...
+	IstioIntegration bool
```
jenshu, line 8:
```suggestion
	EnableIstioIntegration bool
```

lgadban:
fixes https://github.com/k8sgateway/k8sgateway/issues/10445

timflannagan (approved):
An envvar-based approach SGTM. I think we'll ideally want the new chart to enable/disable this once that API has been fleshed out more. WYDT? Should we add that as a line item to the new chart tracker.

lgadban:
>An envvar-based approach SGTM. I think we'll ideally want the new chart to enable/disable this once that API has been fleshed out more. WYDT? Should we add that as a line item to the new chart tracker.

Yes definitely. There is an existing helm API for this, so I would imagine as we look for which features to port over, this needs to be one of them. A specific line-item sounds like a good idea!

yuval-k (approved):

PR #10460, opened 2025-01-15
Admin server - part 1

jenshu:
First pass at adding back the admin server:
- start admin server from "ggv2" setup func instead of the old setup loop
- with this change, the paths/functions currently available are:
    - port 9091 (will be refined in a follow-up PR)
        - /debug/pprof
        - /logging
        - /metrics
        - /zpages
    - port 9095 (will add back additional funcs in follow-up PR)
        - /snapshots/krt
        - /snapshots/xds
- removed old admin server and "history" code

Local testing steps:
```
# set up cluster, build images & helm chart
kind delete cluster --name kind
rm -rf _test _output
ci/kind/setup-kind.sh

# install local chart
helm upgrade -i -n gloo-system gloo ./_test/gloo-1.0.0-ci1.tgz --create-namespace -f - <<EOF
kubeGateway:
  enabled: true
gloo:
  logLevel: debug
EOF

# add some resources e.g. gateway / httproute
k apply -f https://raw.githubusercontent.com/solo-io/gloo/v1.13.x/example/petstore/petstore.yaml
k apply -f - <<EOF
kind: Gateway
metadata:
  name: gw1
spec:
  listeners:
    - protocol: HTTP
      port: 8080
      name: http
      allowedRoutes:
        namespaces:
          from: All
---
kind: HTTPRoute
metadata:
  name: example-route1
spec:
  parentRefs:
    - name: gw1
  hostnames:
    - "example1.com"
  rules:
    - backendRefs:
        - name: petstore
          port: 8080
EOF

k port-forward -n gloo-system deploy/gloo 9091
# then go to http://localhost:9091 in a browser

k port-forward -n gloo-system deploy/gloo 9095
# then go to http://localhost:9095 in a browser
```


projects/gateway2/setup/ggv2setup.go
```diff
@@ -171,6 +175,59 @@ func StartGGv2WithConfig(ctx context.Context, setupOpts *controller.SetupOpts,
...
 	setting.Synced().WaitUntilSynced(ctx.Done())
 
+	logger.Info("running start funcs")
+	runStartFuncs(ctx, setupOpts)
+
 	logger.Info("starting controller")
 	return c.Start(ctx)
 }
+
+func runStartFuncs(ctx context.Context, setupOpts *controller.SetupOpts) {
```
yuval-k, line 185: I think we can just do this for now:
```go
go admin.AdminServerStartFunc(setupOpts)
```

and bring back the complexity of startFuncs only when we need it (hopefully never)?

projects/gateway2/admin/server.go
```diff
@@ -0,0 +1,222 @@
...
+	return func(ctx context.Context) error {
+		// serverHandlers defines the custom handlers that the Admin Server will support
+		serverHandlers := getServerHandlers(ctx, setupOpts.KrtDebugger, setupOpts.Cache)
+
+		// The Stats Server is used as the running server for our admin endpoints
+		//
+		// NOTE: There is a slight difference in how we run this server -vs- how we used to run it
+		// In the past, we would start the server once, at the beginning of the running container
+		// Now, we start a new server each time we invoke a StartFunc.
+		stats.StartCancellableStatsServerWithPort(ctx, stats.DefaultStartupOptions(), func(mux *http.ServeMux, profiles map[string]string) {
```
yuval-k, line 41: will this change to just include just the handlers we need on the correct port? is that part 2?

projects/gateway2/setup/ggv2setup.go
jenshu: yeah i had something simple like that at first, then brought back all the startFunc stuff because it was handling error collection /  logging (and in case we wanted to add other servers / goroutines later). do you think we'll have a need for that? if not, i can simplify and remove it

projects/gateway2/admin/server.go
jenshu: yep, in the next PR i plan to move this stuff onto port 9095 and stop calling go-utils

projects/gateway2/setup/ggv2setup.go
yuval-k: i think it's worth simplifying so the code is easier to read.
do we expect to have many of these sub-systems? i.e. do you think we will very quickly need this machinery back?

jenshu: not sure. removed for now! 4dbe3de

projects/gateway2/admin/server.go
```diff
@@ -8,50 +8,60 @@ import (
...
+	envoycache "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"github.com/rotisserie/eris"
+	"github.com/solo-io/gloo/projects/gateway2/controller"
 	"github.com/solo-io/go-utils/contextutils"
+	"github.com/solo-io/go-utils/stats"
 	"istio.io/istio/pkg/kube/krt"
 )
 
 const (
 	AdminPort = 9095
```
lgadban, line 21: I would be in favor of keeping the true admin endpoints on 9091, but don't feel strongly (and don't think it needs to be in this PR either)

lgadban (commented):
current state of changes LGTM, but general question (echoing @yuval-k a bit):
* when the new handlers and ports are refined in the follow-up, we will only have 1 port for now to serve all admin functions? (as we will metrics separately, in a later phase, correct?)

projects/gateway2/admin/server.go
jenshu: would it be better to reserve 9091 for when we add metrics back?
and keep the pod-internal endpoints on 9095?

jenshu:
> current state of changes LGTM, but general question (echoing @yuval-k a bit):
> 
> * when the new handlers and ports are refined in the follow-up, we will only have 1 port for now to serve all admin functions? (as we will metrics separately, in a later phase, correct?)

yep, that's the plan. i filed some sub-issues for https://github.com/k8sgateway/k8sgateway/issues/10416 (will fill out the details shortly)

yuval-k (approved):

lgadban (approved):

PR #10466, opened 2025-01-16
Feat/action linter

nfuden:
Add a github linter.

This linter (also github itself's in ui view) doesnt really handle composite actions so also move composite to a higher order folder.


.github/workflows/action_lint.yaml
```diff
@@ -0,0 +1,21 @@
...
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@4
+        with:
+          sparse-checkout: .github
+      - uses: actions/setup-go@5
+        with:
+          go-version-file: 'go.mod'
+      - run: go install github.com/rhysd/actionlint/cmd/actionlint@latest
```
yuval-k, line 20: we can do instead
```
go run github.com/rhysd/actionlint/cmd/actionlint@latest
```
and skip the next step.

also, should we set a specific version (instead of latest)?

.github/workflows/action_lint.yaml
```diff
@@ -0,0 +1,21 @@
+name: pr-github-workflow-lint
+
+on:
+  pull_request:
+    paths:
+      - .github/*.yml #
```
jenshu, line 6: do we want to check both `yaml` and `yml`?

.github/workflows/action_lint.yaml
```diff
@@ -0,0 +1,21 @@
...
+  contents: read
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@4
+        with:
+          sparse-checkout: .github
+      - uses: actions/setup-go@5
```
jenshu, line 17:
```suggestion
      - uses: actions/checkout@v4
        with:
          sparse-checkout: .github
      - uses: actions/setup-go@v5
```

.github/workflows/action_lint.yaml
timflannagan: I think we'll need to overhaul how we manage tools in this repo. There's some caching issues (IIRC) with `go get ...` compared to `go install ...` and it could impact CI runtime. Not a huge issue right now though.

+1 on pinning to a specific version though.

jenshu: +1 for pinning to specific version

nfuden: yeah we likely need go cache but Ive recently seen some issues with go mod cache recently so I am willing to kick that down the line

timflannagan (approved):

jenshu (approved):

PR #10482, opened 2025-01-21
Admin server - part 2

jenshu:
- Move all of the internal debugging endpoints to port 9091 (and retire 9095).
- Add back logging and pprof endpoints
    - also fixed the logging handler so that it accurately reflects the current log level when you load the admin page
- The `/zpages` endpoint is removed.
- The `/metrics` endpoint is removed, but will be added back later (https://github.com/kgateway-dev/kgateway/issues/10452)

Port 9091 now contains:
- /debug/pprof/
- /logging
- /snapshots/krt
- /snapshots/xds


projects/gateway2/admin/server.go
```diff
@@ -6,38 +6,41 @@ import (
...
 	// serverHandlers defines the custom handlers that the Admin Server will support
 	serverHandlers := getServerHandlers(ctx, setupOpts.KrtDebugger, setupOpts.Cache)
 
-	stats.StartCancellableStatsServerWithPort(ctx, stats.DefaultStartupOptions(), func(mux *http.ServeMux, profiles map[string]string) {
-		// let people know these moved
-		profiles[fmt.Sprintf("http://localhost:%d/snapshots/", AdminPort)] = fmt.Sprintf("To see snapshots, port forward to port %d", AdminPort)
-	})
+	// initialize the atomic log level
+	if envLogLevel := os.Getenv(contextutils.LogLevelEnvName); envLogLevel != "" {
+		contextutils.SetLogLevelFromString(envLogLevel)
```
nfuden, line 28: this is definitely a good reminder that we should pull contextutils into this repo to work towards not needing the go-utils dependency

jenshu: yeah i noticed we use it everywhere for logging. do we no longer want the go-utils dep, even for logging stuff?
if so, i can work on removing it completely from the repo (in a separate PR)

projects/gateway2/admin/logging.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, level := range supportedLogLevels {
+		if level == currentLogLevel {
+			selectorText += fmt.Sprintf(`<option value="%s" selected>%s</option>`, level, level)
+		} else {
+			selectorText += fmt.Sprintf(`<option value="%s">%s</option>`, level, level)
+		}
+	}
+	selectorText += `</select>`
+
+	return `View or change the log level of the program.<br/>
```
nfuden, line 37:
```suggestion
	return `View or change the log level of the program. Note does not persist across restarts.<br/>
```

Wdyt?

projects/gateway2/admin/server.go
lgadban: we should remove the contextutils/go-utils dep entirely but i dont think it should factor into this work

lgadban (commented):
LGTM but will defer approval

projects/gateway2/admin/server.go
nfuden: agreed. This is more of an interesting point that we still want something that allows this toggle to change logging holistically from a single end point

nfuden (approved):

projects/gateway2/admin/server.go
yuval-k: i agree that this is out of scope for this change, but for context,  we probably want to change they way we do logging - having them attached to the context is annoying:
- we have to pass context even to util functions that dont do IO
- we can't have log topics and enable debug on specific topics

PR #10470, opened 2025-01-17
Bootstrap initial kgateway helm chart

timflannagan:
# Description

This commit introduces an initial install/helm/kgateway chart via the `helm create` bootstrap command and updates the gateway2 codegen bash script to generate CRDs and RBAC for this chart along with the legacy chart. Note: this chart is not functional at this point. Future PRs will iterate on this effort, expand this chart values & templates, make sure it can be deployed correctly, passes the conformance suite, etc.

In the short term, the plan is to migrate away from the legacy chart + publish this chart with goreleaser to GHCR.io before we cut our first release in this new project. 

Related to https://github.com/kgateway-dev/kgateway/issues/10443. 

Initial task for the https://github.com/kgateway-dev/kgateway/issues/10423 tracker.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #10484, opened 2025-01-21
*: Migrate container registry from quay.io to ghcr.io

timflannagan:
# Description

This commit is a follow-up to #10443 and #10470 and migrates away from the quay.io/solo-io container image registry in favor of the ghcr.io repository in the project.

Additional work around renaming the gloo component, binaries, etc. will be handled once we're closing the finish line with the https://github.com/kgateway-dev/kgateway/issues/10444 cleanup work.

Introducing automation to replace the quay container image expiration label is being tracked in the https://github.com/kgateway-dev/kgateway/issues/10441 epic.

# Checklist:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,8 +9,7 @@ on:
       - main
 
 env:
-  # IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
-  IMAGE_REGISTRY: quay.io/solo-io
+  IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
 
 permissions:
   contents: write
--- a/test/makefile/make_test.go
+++ b/test/makefile/make_test.go
@@ -15,8 +15,7 @@ const (
 
 // Makefile vars
 const (
-	HelmBucket          = "HELM_BUCKET"
-	QuayExpirationLabel = "QUAY_EXPIRATION_LABEL"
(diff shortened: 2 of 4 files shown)


test/makefile/make_test.go
timflannagan, line 1: IMO this test case can be removed. I held off for now until the release pipeline is closer to the finish line.

.github/workflows/release.yaml
```diff
@@ -9,8 +9,7 @@ on:
       - main
 
 env:
-  # IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
-  IMAGE_REGISTRY: quay.io/solo-io
+  IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
```
jenshu, line 12: does repository_owner == kgateway?

nfuden: I actually like it this way. Makes forks easier to maintain

jenshu (approved):

.github/workflows/release.yaml
timflannagan: Yep! We could hardcode the image registry here, but this approach allows us to easily test the workflow in forks.

jenshu: oh didn't mean we should hardcode; i was just confirming if the value of repository_owner is indeed `kgateway`, as opposed to `kgateway-dev` (org name). before, it seemed to be the org name

PR #10488, opened 2025-01-22
[DNM] gloo cleanup part 2: standalone kgateway build

lgadban:
# Description

## Docs changes

# Context

## Notes for reviewers

# Checklist:


timflannagan:
@lgadban Looks like there's some merge conflicts.

projects/envoyinit/cmd/main.go
yuval-k, line 1: it might be not too bad to fix envoy init; if that's the case we can keep it so we wont looe git history

lgadban: good timing, I had to solve it in a way for `deployer_test.go` so I agree, at least the minimal fix to get it working

PR #10493, opened 2025-01-23
Add support for AI APIs 

npolshakova:
# Description
Adds support for AI APIs: https://github.com/kgateway-dev/kgateway/issues/10494

## API changes
Introduces AI APIs in kgateway. This is based on the [Gloo AI APIs](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/enterprise/options/ai/ai.proto). 

The HTTPRoute selects an Upstream via a backendRef:

```
kind: HTTPRoute
metadata:
  name: route-to-upstream
  namespace: gwtest
spec:
  parentRefs:
    - name: http-gw-for-test
  hostnames:
    - "test"
  rules:
    - matches:
        - path:
            type: Exact
            value: /v1/chat/completions
    - backendRefs:
        - name: openai
          kind: Upstream
          group: gateway.kgateway.dev
```

Where an Upstream specifies a provider and auth type:
```
apiVersion: gateway.kgateway.dev/v1alpha1
kind: Upstream
metadata:
  labels:
    app: kgateway
  name: anthropic
  namespace: gwtest
spec:
  ai:
    llm:
      provider:
        anthropic:
          authToken:
            kind: "Passthrough"
```

To select a RoutePolicy you can use extensionRef on the AI Upstream route:
```
kind: HTTPRoute
metadata:
  name: route-to-upstream
  namespace: gwtest
spec:
  parentRefs:
    - name: http-gw-for-test
  hostnames:
    - "test"
  rules:
    - matches:
        - path:
            type: Exact
            value: /v1/chat/completions
    - backendRefs:
        - name: openai
          kind: Upstream
          group: gateway.kgateway.dev
      filters:
      - type: ExtensionRef
        extensionRef:
           group: gateway.kgateway.dev/v1alpha1
           kind: RoutePolicy
           name: open-ai-opt
---
apiVersion: gateway.kgateway.dev/v1alpha1
kind: RoutePolicy
metadata:
  name: route-test
  namespace: gwtest
spec:
  ai:
    promptEnrichment:
      prepend:
        - role: SYSTEM
          content: "respond to all questions in French"
        - role: SYSTEM
          content: "Say hello before each response"
    promptGuard:
      request:
        customResponse:
          message: "Rejected due to inappropriate content"
          statusCode: 400
        regex:
          matches:
          - pattern: "credit card"
      response:
        regex:
          builtins:
            - PHONE_NUMBER
            - EMAIL
            - SSN
            - CREDIT_CARD
```

The APIs follow the required vs. optional API conventions defined in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required: 

Optional fields:
- Have omitempty set
- Are pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or have a built-in nil value (e.g. maps and slices).

Required fields:
- Do not have omitempty set
- They are not a pointer type in the Go definition (e.g. AnotherFlag SomeFlag).

## Code changes
Plugin and extauth changes will be implemented in a follow up. This PR only introduces the AI APIs. 

## CI changes
NONE

## Docs changes
NONE

# Context
Enhancement Proposal: https://github.com/kgateway-dev/kgateway/pull/10495

## Testing steps
Manually validated CR validation by install helm chart and applying example configs.

# Checklist:


projects/gateway2/api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,307 @@
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/runtime"
+
+// AI config is used to configure the behavior of the LLM provider
+// on the level of individual routes. These route settings, such as prompt enrichment,
+// retrieval augmented generation (RAG), and semantic caching, are applicable only
+// for routes that send requests to an LLM provider backend.
+type AI struct {
```
EItanya, line 9: Since the upstream section is `AIUpstream`, should this be `AIRoutePolicy`?

projects/gateway2/api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// Enrich requests sent to the LLM provider by appending and prepending system prompts.
+	// This can be configured only for LLM providers that use the `CHAT` API route type.
+	PromptEnrichment AIPromptEnrichment `json:"prompt_enrichment,omitempty"`
+
+	// Set up prompt guards to block unwanted requests to the LLM provider and mask sensitive data.
+	// Prompt guards can be used to reject requests based on the content of the prompt, as well as
+	// mask responses based on the content of the response.
+	PromptGuard AIPromptGuard `json:"prompt_guard,omitempty"`
```
EItanya, line 18: Just out of curiosity, why aren't these pointers? Is there default behavior? If the user leaves these out I think we should do nothing?

projects/gateway2/api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,307 @@
...
+	// A list of messages to be appended to the prompt sent by the client.
+	Append []Message `json:"append,omitempty"`
+}
+
+// RouteType is the type of route to the LLM provider API.
+type RouteType int
+
+const (
+	// The LLM generates the full response before responding to a client.
+	CHAT RouteType = iota
```
EItanya, line 76: How do `iota` types work when rendering into `json`. I believe they turn into numbers. IIRC kube APIs typically use `type RouteType string` and then define each option. For example const RouteType_CHAT RouteType =  "CHAT"` or something like that. And then they add enum validation via kube builder.

projects/gateway2/api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,230 @@
...
+
+import corev1 "k8s.io/api/core/v1"
+
+type AIUpstream struct {
+	// Send requests to a custom host and port, such as to proxy the request,
+	// or to use a different backend that is API-compliant with the upstream version.
+	CustomHost Host `json:"customHost,omitempty"`
+
+	// The LLM provider backend to use.
+	LLM LLMBackend `json:"llm,omitempty"`
```
EItanya, line 11: Now that we're not using `proto` we can actually flatten this if we want. We can move. For example this struct could be 

```
type AIUpstream struct {
 	// Send requests to a custom host and port, such as to proxy the request,
 	// or to use a different backend that is API-compliant with the upstream version.
 	CustomHost Host `json:"customHost,omitempty"`

 	// The LLM provider backend to use.


	LLMProviders `json:",inline"` // Embeds all fields from LLMProviders directly into LLMBackend as inline fields.
 	MultiPool    *MultiPoolConfig `json:"multipool,omitempty"`
}
```

I definitely like the idea of reusing `LLMProviders` for multi pool, that will make the code much much cleaner, but I don't think we need this top level.

projects/gateway2/api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,230 @@
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+type AIUpstream struct {
+	// Send requests to a custom host and port, such as to proxy the request,
+	// or to use a different backend that is API-compliant with the upstream version.
+	CustomHost Host `json:"customHost,omitempty"`
```
EItanya, line 8: Why is this at the top level and how is it different from the one located on the individual backend.

projects/gateway2/api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,230 @@
...
+	Gemini      *GeminiConfig      `json:"gemini,omitempty"`
+	VertexAI    *VertexAIConfig    `json:"vertexai,omitempty"`
+	Mistral     *MistralConfig     `json:"mistral,omitempty"`
+}
+
+type SingleAuthTokenKind string
+
+const (
+	// Inline provides the token directly in the configuration for the Upstream.
+	Inline SingleAuthTokenKind = "INLINE"
```
EItanya, line 35: This is the type of enum we need in `AIRouteOptions`

EItanya (commented):
Overall this make sense to me. I have some nits about structure, but my big takeaways are:

1. What's the decision process between pointer and non-pointer types throughout the API, it's not clear to me.
2. Why are we using `iota` for enums in some place, and `string` aliases in others?
3. The `AIRouteOptions` need more validation akin to the upstream.

projects/gateway2/api/v1alpha1/ai_policy.go
npolshakova: Ah okay that makes sense- the iota types were based on the old protoc-gen-go generated APIs. If it doesn't render them as strings in the yaml, we should use strings here instead.

projects/gateway2/api/v1alpha1/ai_upstream.go
npolshakova: The top level one is on the static upstream, this is a different upstream type but reuses the Host field here. 

This was just from: https://github.com/solo-io/gloo/blob/4a5953a94de16f26e0284e8aec18d6f997227671/projects/gloo/api/v1/enterprise/options/ai/ai.proto#L62

projects/gateway2/api/v1alpha1/ai_policy.go
npolshakova: Yeah good call!

api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,253 @@
...
+
+// LLMProviders configures the AIRoutePolicy gateway to use a single LLM provider backend.
+// +kubebuilder:validation:XValidation:message="There must one and only one LLMProviders type set",rule="1 == (self.openai != null?1:0) + (self.azureopenai != null?1:0) + (self.anthropic != null?1:0) + (self.gemini != null?1:0) + (self.vertexai != null?1:0) + (self.mistral != null?1:0)"
+type LLMProviders struct {
+	OpenAI      *OpenAIConfig      `json:"openai,omitempty"`
+	AzureOpenAI *AzureOpenAIConfig `json:"azureopenai,omitempty"`
+	Anthropic   *AnthropicConfig   `json:"anthropic,omitempty"`
+	Gemini      *GeminiConfig      `json:"gemini,omitempty"`
+	VertexAI    *VertexAIConfig    `json:"vertexai,omitempty"`
+	Mistral     *MistralConfig     `json:"mistral,omitempty"`
```
EItanya, line 25: Should we just remove mistral since it's identical to `openai`?

npolshakova: Yep, good call!

projects/gateway2/api/v1alpha1/ai_policy.go
npolshakova: Updated the APIs to follow the guide: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,298 @@
...
+// moderation model. If the model identifies the content as harmful based on its rules,
+// the request is automatically rejected.
+//
+// You can configure a moderation endpoint either as a standalone prompt guard setting
+// or alongside other request and response guard settings.
+type Moderation struct {
+	// Pass prompt data through an external moderation model endpoint,
+	// which compares the request prompt input to predefined content rules.
+	// Configure an OpenAI moderation endpoint.
+	OpenAIModeration *OpenAIModeration `json:"openAIModeration"`
```
andy-fong, line 193: Can this use Upstream->OpenAIConfig instead?

api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,253 @@
...
+
+import corev1 "k8s.io/api/core/v1"
+
+// +kubebuilder:validation:XValidation:message="There must one and only one LLM or MultiPool can be set",rule="(has(self.llm) && !has(self.multipool)) || (!has(self.llm) && has(self.multipool))"
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
+type AIUpstream struct {
+	// Send requests to a custom host and port, such as to proxy the request,
+	// or to use a different backend that is API-compliant with the upstream version.
+	CustomHost *Host `json:"customHost,omitempty"`
```
andy-fong, line 11: I am not seeing where Host is defined. Was going to say should we add the `HostName` field now or later?

api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,253 @@
...
+	// Optional: The model path to route to. Defaults to the Gemini model path, `generateContent`.
+	ModelPath *string `json:"modelPath,omitempty"`
+
+	// The type of publisher model to use. Currently, only Google is supported.
+	// +kubebuilder:validation:Enum=GOOGLE
+	Publisher Publisher `json:"publisher"`
+}
+
+// MistralConfig configures the settings for the [Mistral AIRoutePolicy](https://docs.mistral.ai/getting-started/quickstart/) LLM provider.
+type MistralConfig struct {
```
andy-fong, line 173: Is this still used?

api/v1alpha1/gateway_parameters_types.go
```diff
@@ -101,7 +101,7 @@ type KubernetesProxyConfig struct {
 	// +kubebuilder:validation:Optional
 	Stats *StatsConfig `json:"stats,omitempty"`
 
-	// Configuration for the AI extension.
+	// Configuration for the AIRoutePolicy extension.
```
andy-fong, line 104: Do you really mean to change this to AIRoutePolicy extension? This is configuration for the AI Extension pod. There are more below.

api/v1alpha1/route_policy_types.go
```diff
@@ -31,5 +31,6 @@ type RoutePolicyList struct {
 type RoutePolicySpec struct {
 	TargetRef LocalPolicyTargetReference `json:"targetRef,omitempty"`
 	// +kubebuilder:validation:Minimum=1
-	Timeout int `json:"timeout,omitempty"`
+	Timeout int            `json:"timeout,omitempty"`
```
andy-fong, line 34: During the discussion, I think we decide to remove this?

internal/kgateway/deployer/deployer_test.go
```diff
@@ -941,7 +941,7 @@ var _ = Describe("Deployer", func() {
 			Expect(istioContainer.Resources.Requests.Cpu()).To(Equal(expectedGwp.Istio.IstioProxyContainer.Resources.Requests.Cpu()))
 			// TODO: assert on istio args (e.g. log level, istio meta fields, etc)
 
-			// assert AI extension container
+			// assert AIRoutePolicy extension container
```
andy-fong, line 944: Do you really meant to change these in this file? This is the container.

internal/kgateway/deployer/values.go
```diff
@@ -62,7 +62,7 @@ type helmGateway struct {
 	// stats values
 	Stats *helmStatsConfig `json:"stats,omitempty"`
 
-	// AI extension values
+	// AIRoutePolicy extension values
```
andy-fong, line 65: and here?

andy-fong (changes requested):

api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// AIPromptEnrichment defines the config to enrich requests sent to the LLM provider by appending and prepending system prompts.
+// This can be configured only for LLM providers that use the CHAT API type.
+//
+// Prompt enrichment allows you to add additional context to the prompt before sending it to the model.
+// Unlike RAG or other dynamic context methods, prompt enrichment is static and is applied to every request.
+//
+// **Note**: Some providers, including Anthropic, do not support SYSTEM role messages, and instead have a dedicated
+// system field in the input JSON. In this case, use the [`defaults` setting](#fielddefault) to set the system field.
```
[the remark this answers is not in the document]
npolshakova: There's an example here: https://github.com/kgateway-dev/kgateway/blob/956c0bd038bb134b409524f16c6afc88469111d9/api/v1alpha1/ai_policy.go#L259 I would rather keep the example next to the defaults field, but let me know if there's any rewording that would help here!

api/v1alpha1/ai_policy.go
npolshakova: Good call!

npolshakova: (I'll keep the field name the same, but change the type to OpenAIConfig)

api/v1alpha1/ai_upstream.go
npolshakova: We're re-using the upstream host type from here instead: https://github.com/kgateway-dev/kgateway/blob/498f2de042bb62cb4edddc9f8e4e51160e7e28d5/api/v1alpha1/upstream_policy_types.go#L45

api/v1alpha1/ai_upstream.go
npolshakova: Nope, will remove!

api/v1alpha1/gateway_parameters_types.go
npolshakova: Good catch- I think it auto refactored this comment. It should be just ai extensions since you don't need the route policy present (this just controls the extproc config)

api/v1alpha1/route_policy_types.go
npolshakova: Probably not as this PR!

internal/kgateway/deployer/deployer_test.go
npolshakova: No good catch, the comment should just be "AI extension". It's not specific to route policy.

internal/kgateway/deployer/values.go
npolshakova: Yep, good catch- cleaning up all the `AIRoutePolicy` renamed.

api/v1alpha1/ai_policy.go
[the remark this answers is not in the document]
andy-fong: yeah, the example looks fine once I saw it. I got tripped up by the "user input fields" in the comment which I thought is the fields in the CRD's but that actually means the json fields in the user request body. May be clarifying that a bit in the comment would be good.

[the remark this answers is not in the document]
lgadban: +1 on this, not a blocker, but had the exact same concern when trying to parse the top-level `Defaults` field here: https://github.com/kgateway-dev/kgateway/pull/10493/files#diff-9810bd4253d9bbea37fa78da62dfbdb711dc83c7b1c074bfb9516fbb476bea69R24

It would probably be beneficial to enrich (😉) the user facing docs on `Defaults` to explain what "user input fields" are and where they are relevant

api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,286 @@
...
+// ```
+type AIPromptEnrichment struct {
+	// A list of messages to be prepended to the prompt sent by the client.
+	Prepend []Message `json:"prepend,omitempty"`
+	// A list of messages to be appended to the prompt sent by the client.
+	Append []Message `json:"append,omitempty"`
+}
+
+// RouteType is the type of route to the LLM provider API.
+type RouteType string
```
lgadban, line 73: out of curiosity, is "route" the accepted term in the LLM world for this?

api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// Regular expression (regex) matching for prompt guards and data masking.
+	Regex *Regex `json:"regex,omitempty"`
+
+	// Configure a webhook to forward requests to for prompt guarding.
+	Webhook *Webhook `json:"webhook,omitempty"`
+
+	// Pass prompt data through an external moderation model endpoint,
+	// which compares the request prompt input to predefined content rules.
+	Moderation *Moderation `json:"moderation,omitempty"`
```
lgadban, line 201: nit: might be worth documenting the ordering with this and the webhook prompt guard?
is the ordering relevant/important?

api/v1alpha1/ai_upstream.go
lgadban: nit: this naming & doc is a bit confusing to me, the intent here is to override the default host:port for a given `LLMProvider` -- is that accurate?

I would push for something like `HostOverride` rather than `CustomHost` in that case.

api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,237 @@
...
+// OpenAIConfig settings for the [OpenAI](https://platform.openai.com/docs/api-reference/streaming) LLM provider.
+type OpenAIConfig struct {
+	// The authorization token that the AI gateway uses to access the OpenAI API.
+	// This token is automatically sent in the `Authorization` header of the
+	// request and prefixed with `Bearer`.
+	// +kubebuilder:validation:Required
+	AuthToken SingleAuthToken `json:"authToken"`
+	// Optional: Send requests to a custom host and port, such as to proxy the request,
+	// or to use a different backend that is API-compliant with the upstream version.
+	CustomHost *Host `json:"customHost,omitempty"`
```
lgadban, line 77: should this exist here given the existing `CustomHost` at the top-level of `AIUpstream`?

api/v1alpha1/ai_upstream.go
```diff
@@ -0,0 +1,237 @@
...
+	// The LLM configures the AI gateway to use a single LLM provider backend.
+	LLM *LLMProviders `json:"llm,omitempty"`
+	// The MultiPool configures the backends for multiple hosts or models from the same provider in one Upstream resource.
+	MultiPool *MultiPoolConfig `json:"multipool,omitempty"`
+}
+
+// LLMProviders configures the AI gateway to use a single LLM provider backend.
+// +kubebuilder:validation:MaxProperties=1
+// +kubebuilder:validation:MinProperties=1
+type LLMProviders struct {
```
lgadban, line 22: do we need to document the default endpoints used for these providers?

api/v1alpha1/ai_upstream.go
npolshakova: Sure we can rename to `HostOverride`! The `CustomHost` field came from the original Gloo api.

npolshakova: And yes- the intent is to overwrite the default provider API endpoint (which allows self hosted models, etc.)

npolshakova: Actually this is a good question- do we want to make this a separate type in order to support a custom host name for a custom sni? Or is that something we would add to the upstream [Host](https://github.com/kgateway-dev/kgateway/blob/498f2de042bb62cb4edddc9f8e4e51160e7e28d5/api/v1alpha1/upstream_policy_types.go#L45)? @andy-fong @lgadban

api/v1alpha1/ai_policy.go
npolshakova: So for gemini and vertex ai there is a separate route for streaming and non streaming endpoints:
```
non-streaming:
curl \
  -H 'Content-Type: application/json' \
  -d '{"contents":[{"parts":[{"text":"Explain how AI works"}]}]}' \
  -X POST 'https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash-latest:generateContent?<mykey>

streaming:
curl \
  -H 'Content-Type: application/json' \
  -d '{"contents":[{"parts":[{"text":"Explain how AI works"}]}]}' \
  -X POST 'https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash-latest:generateContentStream?<mykey>

```

This field can potentially expand to other endpoints, for example image generation for open ai's API:
```
curl https://api.openai.com/v1/images/generations \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{
    "model": "dall-e-3",
    "prompt": "A friendly blue blob called glooey",
    "n": 1,
    "size": "1024x1024"
  }'
```

So imo "route" makes sense as a term since it's the API route that it's directly impacting.

api/v1alpha1/ai_policy.go
npolshakova: You can have multiple applied (moderation AND webhook AND regex). I think the API makes this clear because there's no oneof. I don't think order matters, @andy-fong ? But if it does- I can update the comment here.

api/v1alpha1/ai_upstream.go
npolshakova: So in the Gloo api it was per provider in case we wanted to add custom config for a specific provider. But I think this is a good point- I don't really see a reason to not move it to the top level now since we're reusing the shared upstream Host. @EItanya / @andy-fong ?

api/v1alpha1/ai_upstream.go
npolshakova: We should have links to the docs that have the default endpoints:

```
// OpenAIConfig settings for the [OpenAI](https://platform.openai.com/docs/api-reference/streaming) LLM provider.
type OpenAIConfig struct {}

// AzureOpenAIConfig settings for the [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-services/openai/) LLM provider.
type AzureOpenAIConfig struct {}

// GeminiConfig settings for the [Gemini](https://ai.google.dev/gemini-api/docs) LLM provider.
type GeminiConfig struct {}
```

Not sure if hard coding the endpoint here would be better than linking to the docs since some cases change the endpoint (vertex ai for example depends on the project ID and region (`https://us-central1-aiplatform.googleapis.com/v1/projects/my-project/locations/us-central1/publishers/google/models/gemini-1.5-flash-001:generateContent`.

lgadban: Yeah those links should be good enough!

I think the main open question in my mind is what they will look in the generated API reference, which we don't currently have anyway...

api/v1alpha1/ai_upstream.go
lgadban: this is tricky; as we don't support upstream SNI yet (or upstream TLS in general), it's tough to want to modify the upstream Host as part of the AI work.

IMO, we should punt on this in this PR but get upstream TLS/SNI sorted out ASAP and revisit this.
It's not a great answer but I don't think we have an alternative, as there's no way setting SNI would work anyway for either the AI or "regular" upstreams

api/v1alpha1/ai_policy.go
andy-fong: The order of the config does not matter but in the code:
For requests, the order is `webhook`, `regex`, then `moderation`. Each of this can reject the request and stop the upstream flow. `webhook` and `regex` can modify the body. So, whatever webhook modified will get passed into `regex` and so on to `moderation`. `moderation` does not modify the body, only reject or let it pass. 

For responses, there is no `moderation`. It's also `webhook` first, and then `regex`.

npolshakova: Okay I'll add a comment about the order and behavior (that each one can reject and stop the flow)

npolshakova: Added: https://github.com/kgateway-dev/kgateway/pull/10493/commits/0cb05fd1a950c07dfc2972e74679532b98ea9dbc

api/v1alpha1/ai_upstream.go
andy-fong: we actually use transformation to override the url to the specific end point for the provider as long as the route is matched. So, that can be a surprise if the user is not aware.

install/helm/kgateway/crds/gateway.kgateway.dev_routepolicies.yaml
```diff
@@ -121,21 +121,6 @@ spec:
                                     required:
                                     - kind
                                     type: object
-                                  customHost:
```
andy-fong: now that the HostOverride is moved out of OpenAIConfig, the moderation can no longer override the host. However, we the specific OpenAI `/v1/moderation` endpoint, I am not sure if you can self host at all or run moderation model in Azure for example. So, may not be a big deal but just want to point this out.

andy-fong: Today, we don't have this ability to override the host/port for moderation neither.

npolshakova: Yep, so I was thinking about this earlier. For now the moderation endpoint if it's configured to OpenAI should always go to the OpenAI endpoint. If we want to support custom hosts/custom moderation config, I think it would make sense to add a new field:
```
type Moderation struct {
	// Pass prompt data through an external moderation model endpoint,
	// which compares the request prompt input to predefined content rules.
	// Configure an OpenAI moderation endpoint.
	OpenAIModeration *OpenAIConfig `json:"openAIModeration"`

	CustomModeration *CustomModeration `json:customModeration"`
}
```

api/v1alpha1/ai_upstream.go
```diff
@@ -6,20 +6,25 @@ import corev1 "k8s.io/api/core/v1"
...
-	CustomHost *Host `json:"customHost,omitempty"`
-
 	// The LLM configures the AI gateway to use a single LLM provider backend.
 	LLM *LLMProviders `json:"llm,omitempty"`
 	// The MultiPool configures the backends for multiple hosts or models from the same provider in one Upstream resource.
 	MultiPool *MultiPoolConfig `json:"multipool,omitempty"`
 }
 
-// LLMProviders configures the AI gateway to use a single LLM provider backend.
+type LLMProviders struct {
```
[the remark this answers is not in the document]
npolshakova: Updated: https://github.com/kgateway-dev/kgateway/pull/10493/commits/c123e780abd1dd113a4989fef35e61325903c40c

andy-fong (approved):

api/v1alpha1/ai_policy.go
```diff
@@ -0,0 +1,295 @@
...
+	// This can be configured only for LLM providers that use the `CHAT` or `CHAT_STREAMING` API route type.
+	PromptEnrichment *AIPromptEnrichment `json:"promptEnrichment,omitempty"`
+
+	// Set up prompt guards to block unwanted requests to the LLM provider and mask sensitive data.
+	// Prompt guards can be used to reject requests based on the content of the prompt, as well as
+	// mask responses based on the content of the response.
+	PromptGuard *AIPromptGuard `json:"promptGuard,omitempty"`
+
+	// Provide defaults to merge with user input fields.
+	// Defaults do _not_ override the user input fields, unless you explicitly set `override` to `true`.
```
lgadban, line 23: I still think we need to update these docs, they are not really helpful to a new reader
ref: https://github.com/kgateway-dev/kgateway/pull/10493/files#r1968380411

Will capture this here for posterity

lgadban (approved):

Issue #10494, opened 2025-01-23
[Epic] Support AI Gateway APIs


npolshakova:
### K8sGateway Version

main

### Is your feature request related to a problem? Please describe.

[Gloo AI Gateway](https://docs.solo.io/gateway/main/ai/) introduced a set of [APIs](https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/enterprise/options/ai/ai.proto) to manage the responsibility of credential management, prompt management, and LLM provider integration. This issue is to add support for the ai APIs and functionality in the kgateway repo in order to provides APIs, [ext-proc](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter) code, and controllers to support applying advanced policies to routes. 

### Describe the solution you'd like

Add the following support:

**API**
- [x] Make sure all the API definitions (https://github.com/solo-io/gloo/blob/main/projects/gloo/api/v1/enterprise/options/ai/ai.proto) live in https://github.com/k8sgateway/k8sgateway/blob/main/projects/gloo/api/v1/enterprise/options/ai/ai.proto

**Helm**
- [x] Make sure the helm values are up to date from OSS (https://github.com/solo-io/gloo/blob/f7b6115a833b0dc38d92d6ed8cb51d8f3ec92afe/projects/gateway2/deployer/values.go#L150) the current k8s gateway fields are missing some of the tracing fields (https://github.com/k8sgateway/k8sgateway/blob/1b8cf1d7ebbd0109979fabaaa797824e5c7449fa/projects/gateway2/deployer/values.go#L148)

**Functionality**
- [x] Add support for the ai APIs in k8sgateway upstream plugin
- [x] Add support for the ai APIs in k8sgateway route policy plugin
- [x] Add extproc ai extensions to k8sgateway

**Tests**
- [x] Add AI e2e tests for basic routing to all providers, passthrough, streaming, prompt guard on non-streaming requests, prompt guard for streaming requests.

**CI**
- [x] Setup CI to run AI e2e tests and python dependencies.
- [x] Add python linter to k8sgateway

lgadban:
see also https://github.com/kgateway-dev/kgateway/issues/10649

PR #10495, opened 2025-01-23
Adds EP-10494: AI APIs enhancement proposal

npolshakova:
# Description

Adds Enhancement Proposal 10494 that proposes adding AI Gateway APIs support.

Supports [#10494](https://github.com/kgateway-dev/kgateway/issues/10494)

Initial APIs can be found in this draft PR: https://github.com/kgateway-dev/kgateway/pull/10493


docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,271 @@
+# EP-10494: Add Support for AI Gateway APIs
+
+* Issue: [#10494](https://github.com/kgateway-dev/kgateway/issues/10494)
+
+## Background
+
+This EP proposes adding support for AI Gateway APIs to the K8s Gateway.
+The AI Gateway APIs will enable users to route traffic to LLM providers while applying advanced policies, such as 
+prompt enrichment, prompt guard, streaming, and user-invoked function calling. These APIs will be implemented using the
```
linsun, line 9: Hi @npolshakova - what is an example of user-invoked function calling surfaced in the API you proposed? You might have it already in the doc, but i couldn't find it.

npolshakova: Good call! Added some additional examples in this commit: https://github.com/kgateway-dev/kgateway/pull/10495/commits/5c3a41a38cc1e3628770bae035604822c7f9fb77

EItanya (commented):
This proposal LGTM

docs/content/enhancements/10494.md
linsun: Thank you @npolshakova ! this looks great!

docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,534 @@
...
+
+### ExtProc Server
+
+The ExtProc server will be written in Python and is likely to remain so for several reasons:
+* Many AI libraries, including nearly all widely used ones, are written in Python. Since ExtProc is implemented with Protobuf, setting up a Python server should be straightforward.
+* Latency, which is often critical, is less of a concern in AI operations due to the inherently high response times of LLMs. However, to minimize the impact of the ExtProc server, we can easily run it as a sidecar to Envoy and connect via a Unix Domain Socket.
+
+### Reporting
+
+* Update the [reporter](https://github.com/kgateway-dev/kgateway/tree/main/projects/gateway2/reports) package to support status reporting for Upstream's referencing kubernetes secrets.
```
lgadban, line 518: per the community review, worth calling out this is not accurate as the Policy resources do not get their status from the `reporter`, and policy status in general is a TODO

npolshakova: Updated in https://github.com/kgateway-dev/kgateway/pull/10495/commits/ad622e76da7d878ea728e62045e5e49c4bd83b8d!

npolshakova:
@linsun / @yuval-k can I get another review?

docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,536 @@
...
+metadata:
+  name: open-ai-opt
+  namespace: kgateway-system
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: openai
+  ai:
+    backupModels:
```
yuval-k, line 170: i see the model is specified on the upstream, but backup models on the route? should they all be in once place?

docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,536 @@
...
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: azure-openai
+  options:
+    ai:
+      promptGuard:
+        request:
+          webhook:
+            host: 172.17.0.1
```
yuval-k, line 325: should we use upstream here?

docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,536 @@
...
+  options:
+    ai:
+      promptGuard:
+        request:
+          moderation:
+            openai:
+              authToken:
+                secretRef:
+                  name: openai-secret
+                  namespace: ai-test
```
yuval-k, line 304: can we keep secretRefs local? i.e. not support namespaces?

docs/content/enhancements/10494.md
```diff
@@ -0,0 +1,536 @@
...
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: azure-openai
+  options:
+    ai:
+      promptGuard:
+        request:
+          moderation:
+            openai:
+              authToken:
```
yuval-k, line 256: instead of `authToken.SecretRef.{name/namespace}` can we splify with just `secretName` ?

docs/content/enhancements/10494.md
npolshakova: I think this is old config actually- it's not in the initial set of [AI APIs](https://github.com/kgateway-dev/kgateway/pull/10493), will remove.

@EItanya, this isn't used anymore right?

docs/content/enhancements/10494.md
npolshakova: Have an upstream ref in the webhook instead of specifying the host? I think that should work, right @EItanya @andy-fong?

docs/content/enhancements/10494.md
npolshakova: I think we just have the namespace because we're using `orev1.LocalObjectReference`. I think we can scope it down to just the namespace. @EItanya

docs/content/enhancements/10494.md
npolshakova: Yeah I think so https://github.com/kgateway-dev/kgateway/pull/10495#discussion_r1941408764

docs/content/enhancements/10494.md
andy-fong: I don't think so because the webhook call is initiated from the ext_proc outside of envoy, so it cannot use envoy upstream cluster if that's what you mean.

andy-fong: unless we front the webhook service with envoy but that will mean we are looping back: enovy -> ext_proc -> envoy -> webhook.

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+
+Large Language Models (LLMs) are limited by static training data. Function calling enables developers to connect 
+language models to external data and systems, enabling dynamic data access, seamless integration with external systems,
+and execution of complex multi-step tasks.  
+
+A user can define a set of functions as tools that the model has access to, and the model can use them when appropriate 
+based on the conversation history. The user then can execute those functions on the application side, and provide results
+back to the model.
+
+1. Fetching data:
```
andy-fong, line 387:
```suggestion
Here are some common use cases for Function Calling:
1. Fetching data:  
```

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+   Retrieve data from internal systems before sending final response, like checking the weather or number of vacation days in an HR system.
+2. Taking action
```
andy-fong, line 389:
```suggestion
2. Taking action:  
```

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+   Trigger actions based on the conversation, like scheduling meetings or initiating order returns.
+3. Building multi-step workflows
```
andy-fong, line 391:
```suggestion
3. Building multi-step workflows:  
```

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+   Execute multi-step workflows, like data extraction pipelines or content personalization.
+4. Interacting with Application UIs
```
andy-fong, line 393:
```suggestion
4. Interacting with Application UIs:  
```

andy-fong (changes requested):

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: azure-openai
+  options:
+    ai:
+      promptGuard:
+        request:
+          moderation:
```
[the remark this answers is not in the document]
npolshakova: Yep, this `moderation` section can be removed. I have the same field set in the later example. This explanation is just for the presidio matching.

design/10494.md
```diff
@@ -55,7 +55,9 @@ additions to the existing API:
 1. A new upstream subtype called AI
 2. A new RoutePolicy sub-object called AI
 
-These two types will allow us to get the new AI specific options without needing to create/add any new API objects
+These two types will allow us to get the new AI specific options without needing to create/add any new API objects.
+
+![ai custom_resources](ai custom_resources.png "AI APIs in relation to kgateway resources")
```
andy-fong, line 60: This is not rendering correctly on github.

design/10494.md
```diff
@@ -384,13 +380,14 @@ A user can define a set of functions as tools that the model has access to, and
 based on the conversation history. The user then can execute those functions on the application side, and provide results
 back to the model.
 
+Here are some common use cases for Function Calling:  
 1. Fetching data:
```
andy-fong, line 384: oops, I put that in the comment but actually didn't put the doublespace at the end on this one, so it's not doing the linebreak like the others.

docs/content/enhancements/10494.md
npolshakova: Makes sense, I'll add a note to clarify why it needs to be the webhook host directly.

docs/content/enhancements/10494.md
npolshakova: Actually, I think keeping it as secretRef and erroring if the reference grant isn't provided makes it more flexible, but still safe.

design/10494.md
```diff
@@ -0,0 +1,572 @@
...
+
+#### Streaming
+
+kgateway will support chat streaming, which allows the LLM to stream out tokens as they are generated. 
+* Some providers, such as OpenAI, send the is-streaming boolean as part of the request to determine whether a 
+request should receive a streamed response. 
+* Other providers, such as the Gemini and Vertex AI providers, change the path to determine 
+streaming, such as the streamGenerateContent segment of the path in the Vertex AI streaming endpoint 
+https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash-latest:streamGenerateContent?key=<key>. 
+To prevent the path you defined in your HTTPRoute from being overwritten by this streaming path, you instead indicate
```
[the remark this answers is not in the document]
npolshakova: Yep exactly, I reworded it but mainly wanted to get across that kgateway will modify the URL based on the routeType that has been configured.

design/10494.md
npolshakova: Ah, missing the `resources` path

andy-fong (approved):

design/10494.md
```diff
@@ -0,0 +1,564 @@
...
+  namespace: kgateway-system
+spec:
+  ai:
+    openai:
+      authToken:
+        secretRef:
+          name: openai-secret
+```
+
+Notice that this Upstream does not specify a model, so kgateway will use the model value in the request to determine
```
lgadban, line 97: >kgateway will use the model value in the request

What does this mean? Might be worth elaborating here.

design/10494.md
```diff
@@ -0,0 +1,564 @@
...
+    - name: open-ai-gpt4
+      namespace: kgateway-system
+      group: kgateway.io
+      kind: Upstream
+    filters:
+    - type: ExtensionRef
+      extensionRef:
+        group: gateway.kgateway.dev/v1alpha1
+        kind: RoutePolicy
+        name: open-ai-opt
```
lgadban, line 173: Prefer using `targetRef` attachment on the `RoutePolicy` rather than `ExtensionRef`?

design/10494.md
```diff
@@ -0,0 +1,564 @@
...
+#### Model Failover
+
+The `MultiPoolConfig` configures the backends for multiple hosts or models from the same provider in one Upstream resource.
+This method can be useful for creating one logical endpoint that is backed by multiple hosts or models.
+
+In the `priorities` section, the order of `pool` entries defines the priority of the backend endpoints.
+The `pool` entries can either define a list of backends or a single backend.
+
+```yaml
+multi:
```
lgadban, line 346: Does this live on the `Upstream`? More context would be help IMO

andy-fong (approved):

EItanya (approved):
These changes make sense to me

design/10494.md
npolshakova: See the comment: https://github.com/kgateway-dev/kgateway/blob/f7cb884215ce132b942ac1bc6a3335e21f39b36c/design/10494.md?plain=1#L49 Initially it's easier to only support extension ref for the first pass because it's a little trickier to confirm all the routes are AI Upstream types in the new plugin system. targetRef support will be added later if there's a specific use case for it.

design/10494.md
npolshakova: Can add a specific request example in a follow up, but for context this field isn't strictly required:
```
spec:
  ai:
    openai:
      model: my-model
```

You can specify a model in the request body instead, but if you specify both the Upstream will win out. (This additional note is more implementation detail, but was request in an earlier review to clarify the behavior)

design/10494.md
npolshakova: Yep `in one Upstream resource.`.

design/10494.md
lgadban: This is worrying because ExtensionRef is effectively redundant given named route rules.
I think we'll want to consider this as important for the plugin system if things are easier to express as ExtensionRefs instead of via policy attachment.
cc @yuval-k

design/10494.md
lgadban: Right but my question is how is it specified in the request body?

PR #10500, opened 2025-01-23
change admin server port to 9097

jenshu:
Change to 9097 for now, so it won't conflict with existing ports used by the legacy admin server, until https://github.com/kgateway-dev/kgateway/issues/10501 is implemented

### DIFF
--- a/projects/gateway2/admin/server.go
+++ b/projects/gateway2/admin/server.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	AdminPort = 9091
+	AdminPort = 9092
 )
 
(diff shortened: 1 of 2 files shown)


projects/gateway2/admin/server.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	AdminPort = 9091
```
yuval-k, line 19: 9095 will collide, as legacy has both 9095 and 9091

nfuden (approved):

lgadban (approved):

PR #10485, opened 2025-01-21
*: Integrate the kgateway chart into the repository

timflannagan:
# Description

This commit is a follow-up to https://github.com/kgateway-dev/kgateway/pull/10470 and integrates the newly bootstrapped kgateway helm chart into the repository and replacing the legacy chart in CI. While this is a foundational step, further iterations and chart design adjustments will follow to refine and fully replace the legacy chart. The goal is to gradually transition to the new chart while deprecating the legacy implementation.

Additional changes include publishing this helm chart along with the https://github.com/kgateway-dev/kgateway/issues/10441 stream of work. This can be done as an extension of the release GHA workflow once we pass the initial hurdle with integrated this chart with CI and getting green runs.

Related to https://github.com/kgateway-dev/kgateway/pull/10484 which renamed the default container image registry from quay to ghcr.io. The new helm chart will have this as the default registry going forward.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

PR #10504, opened 2025-01-24
.github: Publish kgateway helm chart in the release workflow

timflannagan:
# Description

This is a follow-up to #10485 which migrated away from the legacy gloo chart to a new kgateway chart. With these changes, we're adding a new job to the release workflow that's responsible for packaging and publishing the release chart to ghcr.io.

Related to https://github.com/kgateway-dev/kgateway/issues/10423.

# Checklist:


.github/workflows/release.yaml
```diff
@@ -16,8 +16,63 @@ permissions:
...
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+  helm:
+    name: Package Helm Charts
+    needs: setup
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Login to GitHub Container Registry
```
yuval-k, line 56: why not use https://github.com/marketplace/actions/docker-login?

.github/workflows/release.yaml
```diff
@@ -16,8 +16,63 @@ permissions:
...
+        if: ${{ github.event_name != 'pull_request' }}
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ${{ env.IMAGE_REGISTRY }} -u ${{ github.repository_owner }} --password-stdin
+
+      - name: Lint kgateway chart
+        run: helm lint install/helm/kgateway
```
yuval-k, line 61: run this first? so if it fails, it fails fast?

yuval-k: actually, this should run on PRs; not on release

.github/workflows/release.yaml
timflannagan: I was running into issues with that approach when I was testing the overall feature branch in my fork. I didn't dive too deep on why, but did a quick github search and found this approach was relatively common.

lgadban (commented):
LGTM aside from helm login comments; non-blocking from my side though

.github/workflows/release.yaml
```diff
@@ -16,8 +16,70 @@ permissions:
   packages: write
 
 jobs:
+  setup:
+    name: Setup release inputs
+    runs-on: ubuntu-latest
```
jenshu, line 21: should we pin to a version e.g. ubuntu-22.04 so it doesn't change from under us?

.github/workflows/release.yaml
```diff
@@ -16,8 +16,70 @@ permissions:
...
+
+          if [[ ${{ github.event_name }} == 'workflow_dispatch' ]]; then
+            VERSION="${GIT_TAG}-${DATE}-${GIT_SHA}"
+            echo "goreleaser_args=--clean --skip=validate" >> $GITHUB_OUTPUT
+          elif [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION="${GITHUB_REF#refs/tags/}"
+            echo "goreleaser_args=--clean" >> $GITHUB_OUTPUT
+          elif [[ $GITHUB_REF == refs/pull/* ]]; then
+            PR_NUM=$(echo "${GITHUB_REF}" | sed -E 's|refs/pull/([^/]+)/?.*|\1|')
+            VERSION="${GIT_TAG}-pr.${PR_NUM}-${DATE}-${GIT_SHA}"
```
jenshu, line 45: do we need the date as part of the version? seems like sha should be sufficient?

.github/workflows/release.yaml
```diff
@@ -16,8 +16,70 @@ permissions:
...
+            echo "goreleaser_args=--snapshot --clean" >> $GITHUB_OUTPUT
+          else
```
jenshu, line 47: just curious will this case ever occur? looks like we covered manual run, release, and PR above

.github/workflows/release.yaml
```diff
@@ -16,8 +16,70 @@ permissions:
...
+      - name: Helm login to ${{ env.IMAGE_REGISTRY }}
+        if: ${{ github.event_name != 'pull_request' }}
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ${{ env.IMAGE_REGISTRY }} -u ${{ github.repository_owner }} --password-stdin
+
+      - name: Lint kgateway chart on PRs
+        if: ${{ github.event_name == 'pull_request' }}
+        run: helm lint install/helm/kgateway
+
+      - name: Package kgateway chart
+        run: make package-kgateway-chart
```
jenshu, line 71: should this not run on PRs either?

timflannagan: I think it's fine to run this on PRs. It takes minimal time to package a helm chart and it's good to run a subset of the release pipeline on PRs to verify we aren't regressing the pipeline between changes. WYDT?

.github/workflows/release.yaml
timflannagan: agreed. ripped out the date in the latest commit.

.github/workflows/release.yaml
timflannagan: I think it's dead code based on the current event triggers defined for this workflow. I'll look at ripping it out when I add main branch publishing.

.github/workflows/release.yaml
jenshu: yeah that sounds good. if the package-chart fails, will we see that? (i.e. will it cause this job to fail?)

lgadban: this is a good question, wdyt @timflannagan ?

.github/workflows/release.yaml
lgadban: unresolving for posterity, IMO date is useful to know when a commit/release was at a glance.
something we can keep an eye on as we get more use out of the release workflow

.github/workflows/release.yaml
timflannagan: yeah the release/helm job will fail in this scenario. this happened in previous CI runs for this PR before I changed how the VERSION variable was being set as `helm package --version pr-1234` fails their schema validation checks (and therefore the jobs fails).

PR #10508, opened 2025-01-24
exclude various test dirs from build

lgadban:
We are working towards deleting the legacy gloo/gateway/etc directories from this repo, while still keeping several of the existing tests (e2e, helm, makefile, etc).

If we delete the repos, almost all of these tests will stop compiling. There are also several cross-referenced packages with `pkg/utils`

This commit will exclude all tests that we want to keep (but need to rework) along with the legacy tests that live inside various project folders that will be deleted in a following commit.

The exclusions were added via the following commands:

```
find ./test -name '*.go' | xargs gsed -i '1i //go:build exclude\n'
find ./install/test -name '*.go' | xargs gsed -i '1i //go:build exclude\n'
find ./projects/gloo -name '*_test.go' | xargs gsed -i '1i //go:build exclude\n'
find ./projects/gateway -name '*_test.go' | xargs gsed -i '1i //go:build exclude\n'
find ./pkg/utils/statsutils -name '*_test.go' | xargs gsed -i '1i //go:build exclude\n'
```


install/test/5-gateway-validation-webhook-configuration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build exclude
+
```
jenshu, line 2: this file can probably just be deleted since we deleted the validation webhook already

lgadban: I completely agree but trying to keep things uniform; there are several other GE-specific things to delete, but plan to do that in a follow up

jenshu (approved):

PR #10512, opened 2025-01-26
extract gloo deps from utils

lgadban:
Delete unnecessary utilities and modify a handful of others that may come in handy to break the gloo dep before we delete the legacy projects in the following PR.

Linting is temporary enabled only on `projects/gateway2` as the util changes and deletions will break various files in the legacy projects.


lgadban:
closing, too much throw away work to get the final util changes to apply cleaning without performing the delete!

PR #10514, opened 2025-01-27
remove gloo code!

lgadban:
This PR removes the legacy projects `gloo`, `gateway`, and `discovery` as they are no longer necessary or used withing the context of `kgateway`.

A handful of utilities were reworked to break the gloo dependencies so that we can keep them around in case they need to be used within `kgateway` (with some updates to match the `kgateway` code).


lgadban:
need to rework the test exclusion from builds in order to get this to work

jenshu (commented):
this LGTM

will leave open for others to review

yuval-k (approved):

PR #10519, opened 2025-01-27
happy path e2e test

jenshu:
Initial re-enabling of kubernetes e2e tests:
- add a basic routing e2e test
- re-enables the pr-kubernetes-tests workflow on PRs
- fix bug (found by the e2e test!) in helm template where GatewayParameters namespace was hardcoded
- remove a couple of unused/legacy gloo test helper functions


jenshu:
not passing yet, but letting CI run

.github/actions/kubernetes-e2e-tests/action.yaml
```diff
@@ -25,9 +25,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-  - name: Install test tools
-    shell: bash
-    run: make install-test-tools
```
timflannagan, line 30: 👍 we also still need to deflate the setup-envtest target that's defined in the internal/kgateway Makefile.

install/helm/kgateway/templates/gatewayclass.yaml
```diff
@@ -13,8 +13,8 @@ spec:
     group: {{ .group }}
     kind: {{ .kind }}
     name: {{ .name }}
-    namespace: {{ .namespace }}
```
timflannagan, line 16: 👍 we can also pull this fix into it's own PR.

test/kubernetes/e2e/tests/k8s_gw_tests.go
```diff
@@ -22,6 +23,8 @@ import (
 func KubeGatewaySuiteRunner() e2e.SuiteRunner {
 	kubeGatewaySuiteRunner := e2e.NewSuiteRunner(false)
 
+	kubeGatewaySuiteRunner.Register("BasicRouting", basicrouting.NewTestingSuite)
```
timflannagan, line 26: I think a basic routing / happy path suite is sufficient in the very short term as a starting point in getting our suites back online, but we may want to dissect this suite into other suites over time.

test/kubernetes/testutils/assertions/deployments.go
```diff
@@ -30,29 +25,3 @@ func (p *Provider) EventuallyRunningReplicas(ctx context.Context, deploymentMeta
 		WithPolling(time.Millisecond * 200).
 		Should(Succeed())
 }
-
-func (p *Provider) EventuallyGlooReachesConsistentState(installNamespace string) {
```
timflannagan, line 34: Replaced by EventuallyKgatewayInstallSucceeded?

test/kubernetes/e2e/features/basicrouting/suite.go
```diff
@@ -0,0 +1,98 @@
...
+
+	s.manifestObjects = map[string][]client.Object{
+		testdefaults.CurlPodManifest: {testdefaults.CurlPod},
+		exampleServiceManifest:       {nginxPod, exampleSvc},
+		gatewayWithRouteManifest:     {proxyService, proxyServiceAccount, proxyDeployment},
+	}
+}
+
+func (s *testingSuite) TearDownSuite() {
+	// nothing at the moment
```
timflannagan, line 61: Just to clarify, this isn't needed as the AfterTest cleans up the manifests we applied in BeforeTest?

timflannagan (approved):
A couple of quick questions/comments, but nothing critical that would block forward progress.

test/kubernetes/testutils/assertions/deployments.go
jenshu: it was removed / not applicable anymore. it used to check that the `api_gloosnapshot_gloo_solo_io_emitter_snap_out` metric eventually stabilized

> The emitter at some point should stabilize and not continue to increase the number of snapshots produced

not sure if there's an equivalent check we can do in kgateway but i figure we can add it on later, if so

install/helm/kgateway/templates/gatewayclass.yaml
```diff
@@ -13,8 +13,8 @@ spec:
...
     {{- end }}
+    namespace: {{ .Release.Namespace }}
```
lgadban, line 17: I think I'm missing something obvious but why is this now outside the `{{- with .Values.gatewayClass.parametersRef }}` block?

test/kubernetes/testutils/assertions/deployments.go
lgadban, line 34: no need for this going forward?

install/helm/kgateway/templates/gatewayclass.yaml
lgadban: was wondering if we should break this out as well

install/helm/kgateway/templates/gatewayclass.yaml
jenshu: the `with` changes the scope and allows us to refer to the `.Values.gatewayClass.parametersRef` as `.` within the block.

just for fun, i tried moving the `namespace` back into the block and the install errored with 
`Error: template: kgateway/templates/gatewayclass.yaml:16:26: executing "kgateway/templates/gatewayclass.yaml" at <.Release.Namespace>: nil pointer evaluating interface {}.Namespace` i suppose because `.Values.gatewayClass.parametersRef.Release` doesn't exist

test/kubernetes/testutils/assertions/deployments.go
jenshu: see response above

install/helm/kgateway/templates/gatewayclass.yaml
timflannagan: (I blame copilot/cursor/whatever AI tool I used to spit this out.)

install/helm/kgateway/templates/gatewayclass.yaml
jenshu: https://github.com/kgateway-dev/kgateway/pull/10653

install/helm/kgateway/templates/gatewayclass.yaml
lgadban: this is what I figured, and the scoping from `With` makes sense, but this fix (and likely the initial impl) is more confusing to me.

If for some reason parametersRef is an empty value, we will end up with something like:
```
parametersRef:
    namespace: kgateway-system
```
which seems even more weird?

.github/workflows/pr-kubernetes-tests.yaml
```diff
@@ -0,0 +1,96 @@
...
+        # /test/kubernetes/e2e/load_balancing_tests.md
+        #
+        # Above each test below, we document the latest date/time for the GitHub action step to run
+        # NOTE: We use the GitHub action step time (as opposed to the `go test` time), because it is easier to capture
+
+        test:
+        # Feb 17, 2025: ? minutes
+        - cluster-name: 'cluster-one'
+          go-test-args: '-v -timeout=25m'
+          go-test-run-regex: '^TestK8sGateway$$/^BasicRouting$$'
```
npolshakova, line 31: [nit] Should we rename this to `TestKGateway` this is a remnant from the Gloo APIs vs K8s Gateway APIs I think?

jenshu: sounds good, will do that in a followup!

PR #10539, opened 2025-01-28
Initial enhancement proposal template

npolshakova:
# Description
Adds a template for enhancement proposals

## API changes
None

## Code changes
None

## CI changes
None

## Docs changes
- Adds template for creating Enhancement Proposals (EP)


docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+highlighting any additional information provided beyond the standard EP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Background](#Background)
```
timflannagan, line 60: Looks like the background link is broken -- at least when I look at the markdown preview tab in the UI.

timflannagan: Also, do we want a table of contents of proposals? I always felt it was overkill in the KEP format and we could slim down while still following the high-level sections (e.g. background, motivation, goals / non-goals) without being too boggle down on a heavy format.

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+
+
+### Non-Goals 
+
+<!--
+What is out of scope for this EP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Implementation Details
```
timflannagan, line 116: If I was introducing a new CRD or some extension of an API (e.g. aggregated webhook), would I outline that within the implementation details section or a dedicated API section?

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Test Plan 
+
+<!--
+    Define the testing strategy for the feature.
+    Include unit, integration, and end-to-end (e2e) tests.
+    Specify any additional frameworks or tools required for testing.
+-->
+
+## Alternatives
```
timflannagan, line 171: What happens if there's a legitimate alternative worth exploring further? Create another design doc and link back here? Add a sub-section under this alternative with a lengthy amount of context/description?

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+## Implementation Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Configuration
```
timflannagan, line 125: At a glance, these high-level sub-sections seem reasonable to me. I wonder whether we could get away with a high-level "Implementation Details" section without requiring predefined sub-sections to provide some more flexibility in structuring and make the EP process a bit less prescriptive. Just thinking aloud -- no strong opinion.

timflannagan (commented):
A couple of quick EOD thoughts. Everything LGTM at a high-level. Thanks for taking a stab at all of this.

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+- [ ] **Create a PR for this EP.**
+  Assign it to maintainers with relevant context.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the EP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a EP is merged does not mean it is complete or approved. Any EP
+marked as `provisional` is a working document and subject to change. You can
```
jenshu, line 23: maybe worth mentioning what possible states the EP can be in (provisional, implemented, etc?), and how you mark it as 'provisional' (maybe we need another doc for that)

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+<!--
+What is out of scope for this EP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Implementation Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
```
jenshu, line 120: how should we handle API/design/implementation updates, would this design doc need to be kept up to date whenever those change?

docs/content/enhancements/template.md
```diff
@@ -0,0 +1,181 @@
...
+-->
+# EP-[ID]: [Feature/Enhancement Name] 
+
+<!--
+This is the title of your EP. Keep it short, simple, and descriptive. A good
+title can help communicate what the EP is and should be considered as part of
+any review.
+-->
+
+* Issue: [#ID](URL to GitHub issue)
```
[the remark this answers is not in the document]
npolshakova: I don't think we need specific stages, but that's a good question on when an EP can be closed out. I think for now though the issue should be closed out when the code merges, and then any larger follow-ups require their own EP.

docs/content/enhancements/template.md
npolshakova: I think significant API, design, or implementation changes should have a dedicated EP linked to a specific issue, which is closed once the code merges. There are two cases:

1. If code changes affect the API/design during the implementation stage, the EP must be updated. Before merging the final PR or closing the issue, the EP should reflect the latest state.
2. If there are changes that need to be made after the code merges, a new EP/issue needs to be created. This can reference the old issue, but the original EP should remain unchanged.

docs/content/enhancements/template.md
npolshakova: Hm, good call- maybe I should remove the "provisional" section? The EP has two stages:
1) provisional (has not been fully implemented yet)
2) implemented (code has merged, EP is up to date to reflect the state)

Maybe this should be a label/field on the issue itself?

docs/content/enhancements/template.md
npolshakova: Yeah good call- maybe I'll just list the subsection in the comment as a reminder of what sections can be included? This probably means we can also remove the index?

docs/content/enhancements/template.md
npolshakova: Yep! I think you could link other EPs here or have a length discussion section. Really depends on the feature!

docs/content/enhancements/template.md
npolshakova: Yeah fair, there are only a couple sections in the template if we remove the implementation details subsections:
```
<!-- toc -->
- [Background](#background)
- [Motivation](#motivation)
    - [Goals](#goals)
    - [Non-Goals](#non-goals)
- [Implementation Details](#implementation-details)
    - [Test Plan](#test-plan)
- [Alternatives](#alternatives)
- [Open Questions](#open-questions)
<!-- /toc -->
```

I'm down to remove it all together.

timflannagan: Yeah I think I'm leaning towards removing it all together too.

docs/content/enhancements/template.md
timflannagan: I think a comment is a good middle ground as you're still guiding the author towards adding color around the major sub-systems in the project.

jenshu (approved):

Issue #10510, opened 2025-01-24
Implement automated release pipeline


timflannagan:
Introduce GHAs to automate the release process.

DoD:

- [x] Workflow triggered on PRs and runs a subset of the release stages to validate release pipeline consistently works
- [x] Merges to main publish :latest, :main, :nightly (or some variation of that) to ghcr.io
- [x] Package and publish the kgateway helm chart
- [x] Add quickstart instructions for how to consume the release
- [x] All relevant images (controller, sds, envoy-init) are published under the correct registry and naming convention

Partially blocked on #10559.

Out-of-scope:

- Changelogs and release notes
- Any registry management (e.g. periodic blog purging) work items
- Documentation so devs know how the cut a release

timflannagan:
Closing - now that #10540 has merged, we have dev/main/latest builds working and have verified the release pipeline is working as intended without publishing an initial release tag. Note: we landed on the v0.0.0-main tagging convention for dev builds. This was purely a hack/workaround as helm charts require valid semver tags and we didn't want to commit to a tag yet. See https://github.com/kgateway-dev/kgateway/issues/10559 for that tracker.

PR #10540, opened 2025-01-28
release: Publish main branch images and charts

timflannagan:
# Description

This modifies the release.yaml workflow and adds the main branch push trigger to publish main branch images and the latest helm chart to ghcr.io.

Related to #10510.

# Checklist:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,6 +4,9 @@ on:
   push:
     tags:
       - 'v*'
+      - '!v0.0.0-main'
+    branches:
+      - main
   pull_request:
     branches:
       - main
@@ -30,7 +33,7 @@ jobs:
       - name: Set the release related variables
         id: set_vars
         run: |
-          GIT_TAG=$(git describe --tags --abbrev=0)
+          GIT_TAG=$(git describe --tags --abbrev=0 --always)
           GIT_SHA=$(git rev-parse --short HEAD)
           GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed -e "s/\//-/g")
 
@@ -41,6 +44,9 @@ jobs:
           elif [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION="${GITHUB_REF#refs/tags/}"
             echo "goreleaser_args=--clean" >> $GITHUB_OUTPUT
+          elif [[ $GITHUB_REF == refs/heads/main ]]; then
+            VERSION="v0.0.0-main"
+            echo "goreleaser_args=--clean --skip=validate" >> $GITHUB_OUTPUT
           elif [[ $GITHUB_REF == refs/pull/* ]]; then
             PR_NUM=$(echo "${GITHUB_REF}" | sed -E 's|refs/pull/([^/]+)/?.*|\1|')
             VERSION="${GIT_TAG}-pr.${PR_NUM}-${GIT_SHA}"
@@ -108,3 +114,4 @@ jobs:
           VERSION: ${{ needs.setup.outputs.version }}
           IMAGE_REGISTRY: ${{ env.IMAGE_REGISTRY }}
           GORELEASER_ARGS: ${{ needs.setup.outputs.goreleaser_args }}
+          GORELEASER_CURRENT_TAG: ${{ needs.setup.outputs.version }}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -137,7 +137,61 @@ docker_manifests:
       - *envoyinit_arm_image
       - *envoyinit_amd_image
 changelog:
-  disable: false
+  disable: true
 release:
   prerelease: "auto"
   mode: "replace"
+  replace_existing_artifacts: true
+  header: |
+    {{ if eq .Env.VERSION "v0.0.0-main" }}
+    🚀 Nightly build of kgateway!
+    ---
+    It includes the latest changes but may be unstable. Use it for testing and providing feedback.
+    {{ else }}
+    🎉 Welcome to the {{ .Env.VERSION }} release of the kgateway project!
+    ---
+    {{ end }}
+  footer: |
+    ## Installation
+
+    The kgateway project is available as a Helm chart and docker images.
+
+    ### Helm Charts
+
+    The Helm chart is available at {{ .Env.IMAGE_REGISTRY }}/charts/kgateway.
+
+    ### Docker Images
+
+    The docker images are available at:
+
+    - {{ .Env.IMAGE_REGISTRY }}:{{ .Env.CONTROLLER_IMAGE_REPO }}:{{ .Env.VERSION }}
+    - {{ .Env.IMAGE_REGISTRY }}:{{ .Env.SDS_IMAGE_REPO }}:{{ .Env.VERSION }}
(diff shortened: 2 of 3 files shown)


.goreleaser-main.yaml
timflannagan, line 1: I think I'm leaning towards using envsubst and setting the [changelog,release].enabled=false in a single goreleaser config file to avoid drift between the configs, but wanted to get consensus on either approach.

.github/workflows/release.yaml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - 'v*'
+    branches:
+      - main
```
timflannagan, line 8: The alternative is cron schedule expression that fires daily if we're concerned with the free org limits for ghcr.io. In either case, we can figure out the best approach over time. This PR effectively tests whether the release pipeline is working without publishing a real tag.

Makefile
```diff
@@ -261,10 +261,11 @@ view-test-coverage:
...
 
 .PHONY: package-kgateway-chart
+HELM_PACKAGE_ARGS ?= --version $(VERSION)
 package-kgateway-chart: ## Package the new kgateway helm chart for testing
-	mkdir -p $(TEST_ASSET_DIR)
-	helm package --version $(VERSION) --destination $(TEST_ASSET_DIR) install/helm/kgateway
-	helm repo index $(TEST_ASSET_DIR)
+	mkdir -p $(TEST_ASSET_DIR); \
+	helm package $(HELM_PACKAGE_ARGS) --destination $(TEST_ASSET_DIR) install/helm/kgateway; \
+	helm repo index $(TEST_ASSET_DIR);
```
timflannagan, line 268: Technically a no-op right now. I was exploring whether I could unset the helm package --version CLI flag, which has validation logic that checks whether --version is a valid semver (i.e. main or latest would fail that versioning), but the issue is that the default Chart.yaml version is hardcoded to v0.0.1 when you package a chart w/o this --version flag set.

yuval-k: i think it's ok to have the version be something like `0.0.0-<GIT HASH>`  given that this is a nightly release

yuval-k: nvm - that might mean that the oci storage will grow unbounded

timflannagan: i think we'd run into the same unbounded growth with this approach too as we're setting VERSION to be `VERSION="${GIT_TAG}-${GIT_SHA}"` anytime a new PR merges to main. one option that was discussed was using a third party action to periodically purge ghcr storage (e.g. https://github.com/snok/container-retention-policy).

timflannagan: just to clarify too -- we're still planning on mirroring quay's retention policy / auto-pruning images with a label as a follow-up once the release pipeline is fully built out and validated it works.

.goreleaser-main.yaml
lgadban: So the `envsubst` approach would be:
* keep a single ~`.goreleaser.yaml.tmpl` file
* nightly cron will envsubst in false for `[changelog,release]` -> .goreleaser.yaml
* kick off the release make target with newly build yaml file

is that accurate?

Makefile
timflannagan: Ended up going with the v0.0.0-main tagging convention to appease helm validation. Seems a bit wonky but sufficient for now.

Makefile
```diff
@@ -583,11 +584,12 @@ publish-docker: docker docker-push
 
 endif # Publish Artifact Targets
 
-GORELEASER_ARGS ?= --snapshot --clean
 GORELEASER ?= go run github.com/goreleaser/goreleaser/v2@v2.5.1
+GORELEASER_ARGS ?= --snapshot --clean
+GORELEASER_CURRENT_TAG ?= $(VERSION)
```
timflannagan, line 589: Needed as goreleaser will use the v1.18.x tag by default and we want the v0.0.0-main tag to be used for main branch triggers.

timflannagan: See https://goreleaser.com/cookbooks/set-a-custom-git-tag/ for that documentation.

.github/workflows/release.yaml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     tags:
       - 'v*'
+      - '!v0.0.0-main'
```
timflannagan, line 7: Skip the v0.0.0-main tag that the main branch trigger will pump out.

.goreleaser.yaml
```diff
@@ -137,7 +137,61 @@ docker_manifests:
...
+  header: |
+    {{ if eq .Env.VERSION "v0.0.0-main" }}
+    🚀 Nightly build of kgateway!
+    ---
+    It includes the latest changes but may be unstable. Use it for testing and providing feedback.
+    {{ else }}
+    🎉 Welcome to the {{ .Env.VERSION }} release of the kgateway project!
+    ---
+    {{ end }}
+  footer: |
```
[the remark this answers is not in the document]
timflannagan: Tested this out in my fork via force pushing my origin/main remote: https://github.com/timflannagan/kgateway/releases.

.goreleaser-main.yaml
timflannagan: Pivoted away from the envsubst approach in favor of publishing a v0.0.0-main tag instead of dynamically setting the changelog and release enabled flags. The `prerelease: auto` flag treats the v0.0.0-main tag as a prerelease build, so I didn't see a need to maintain two different config files or dynamically render any YAML template files.

.github/workflows/release.yaml
```diff
@@ -41,6 +44,9 @@ jobs:
           elif [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION="${GITHUB_REF#refs/tags/}"
             echo "goreleaser_args=--clean" >> $GITHUB_OUTPUT
+          elif [[ $GITHUB_REF == refs/heads/main ]]; then
+            VERSION="v0.0.0-main"
+            echo "goreleaser_args=--clean --skip=validate" >> $GITHUB_OUTPUT
```
lgadban, line 49: so this is not a `--snapshot` release? Because as currently implemented we want it to be on the GH releases?

.goreleaser.yaml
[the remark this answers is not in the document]
lgadban: so does the mean we will have a constantly floating main release being updated in the GH releases?
I suppose a pre-release doesn't show up on the "home" view of the repo?

lgadban (approved):

.github/workflows/release.yaml
timflannagan: The `--snapshot` flag is just syntactic sugar that configures the right `--skip=[...]` parameters:

```bash
$ goreleaser release --help | grep -- "--snapshot"
      --auto-snapshot                Automatically sets --snapshot if the repository is dirty
      --snapshot                     Generate an unversioned snapshot release, skipping all validations and without publishing any artifacts (implies --skip=announce,publish,validate)
```

The only time we want this flag present is for PR runs that validate a portion of the release pipeline is working as intended. In the case of main branch triggers, we want images to be published at a minimum, and snapshot skips publishe images.

.goreleaser.yaml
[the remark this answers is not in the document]
lgadban: discussed offline, the behavior seems to be correct:
* the main tag release will float near the top of the actual "releases" as it is constantly being rolled over for each push to main
* it will be marked as a prerelease
* it will not show up on the main page in the 'thumbnail' for releases

lgadban (approved):

PR #10542, opened 2025-01-29
fix product name in GitHub issue creation

craigbox:
Update the GitHub templates to use the correct project name.


craigbox:
cc @nfuden in case there was a reason these were not in #10477

yuval-k (approved):

PR #10551, opened 2025-01-29
remove cilium from setup-kind script

jenshu:
this was used for testing; not needed for now as we don't currently support leader election

### DIFF
--- a/ci/kind/cluster.yaml
+++ b/ci/kind/cluster.yaml
@@ -1,9 +1,5 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
-# Do not use the default CNI as kindnet does not support custom network policies.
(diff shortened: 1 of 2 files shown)


timflannagan (approved):
AFAIK, this was needed to simulate API server disruptions (as chaos mesh cannot do that) and we never invested further into a proper chaos testing suite. Is it worth creating a tracker for a chaos testing suite?

jenshu:
> Is it worth creating a tracker for a chaos testing suite?

good idea! https://github.com/kgateway-dev/kgateway/issues/9638 exists but is edge-specific (and should probably be moved out of this repo). @timflannagan do you want to create a new issue for kgateway chaos testing?

nfuden (approved):
Nice to see the continued paring down of deps!

PR #10556, opened 2025-01-30
Remove skv2 dep in e2e tests

jenshu:
We were using a couple of util funcs from skv2 that can be moved into this repo

### DIFF
--- a/pkg/utils/fsutils/fsutils.go
+++ b/pkg/utils/fsutils/fsutils.go
@@ -2,7 +2,12 @@ package fsutils
 
 import (
 	"fmt"
+	"log"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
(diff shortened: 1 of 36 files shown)


timflannagan (approved):
Nice -- manually vendoring the skv2 utilities here makes sense to me.

pkg/utils/fsutils/fsutils.go
lgadban, line 1: feels like this entire util package could/should live under `test/`

lgadban (approved):

pkg/utils/fsutils/fsutils.go
jenshu: i wasn't sure if it should/could since it's referenced from `projects/gateway2/translator/gateway/gateway_translator_test.go` too

PR #10578, opened 2025-02-04
More test cleanup

jenshu:
Remove unused flags from e2e tests:
- IsEnterprise (not relevant as this is an OSS product)
- K8sGatewayEnabled (it's always enabled)
- ValidationAlwaysAccept (validation webhook removed)
- remove legacy edge resource clients
- remove unused consul/vault constants


test/kubernetes/e2e/README.md
```diff
@@ -53,6 +53,16 @@ Some tests may require environment variables to be set. Some required env vars a
 
 - Istio features: Require `ISTIO_VERSION` to be set. The tests running in CI use `ISTIO_VERSION="${ISTIO_VERSION:-1.19.9}"` to default to a specific version of Istio.
 
+## Resource example generation tool
```
jenshu, line 56: added this back, as it was deleted from a readme in a recent cleanup PR

npolshakova: Do we want to keep the example generation tool? I think it was using solo-apis to generate the examples with the test namespace set. We can probably just write all the tests as static YAML since we're only supporting the kgateway apis.

npolshakova (approved):

test/kubernetes/e2e/README.md
jenshu: hmm good point, i wasn't sure if we would still need it, but wanted to preserve that part of the readme for now, until it's completely deleted

PR #10596, opened 2025-02-09
DFP proposal

yuval-k:
very rough draft - mainly to get early feedback.

Adding dynamic forward proxy support. 

Notes:
- This uses envoy's new sub cluster config that seems more ergonomic to use, and has more features. The old dns cache would make envoy NACK if the config on the listener is different than the config on the cluster.
- Not exposing any settings. we use the defaults, until we get feedback otherwise to simplify the API.

Unlike the previous impl that represents DFP as a route action, this one represents it as an upstream type. This makes more sense for the k8s-GW-API, as we cannot add custom route actions.

For every DFP upstream we add a DFP cluster and a filter on the relevant filter chain.

### DIFF
--- a/internal/gateway2/extensions2/plugins/upstream/plugin.go
+++ b/internal/gateway2/extensions2/plugins/upstream/plugin.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"time"
 
+	"google.golang.org/protobuf/proto"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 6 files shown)


internal/gateway2/extensions2/plugins/upstream/plugin.go
```diff
@@ -62,7 +64,8 @@ func (d *upstreamDestination) Equals(in any) bool {
 }
 
 type UpstreamIr struct {
-	AwsSecret *ir.Secret
+	AwsSecret       *ir.Secret
+	dfpFilterConfig *envoydfp.FilterConfig
```
jenshu, line 68: should this field be exported?

internal/gateway2/extensions2/plugins/upstream/plugin.go
```diff
@@ -257,16 +277,20 @@ func (p *plugin2) ApplyForRouteBackend(
 // if a plugin emits new filters, they must be with a plugin unique name.
 // any filter returned from route config must be disabled, so it doesnt impact other routes.
 func (p *plugin2) HttpFilters(ctx context.Context, fc ir.FilterChainCommon) ([]plugins.StagedHttpFilter, error) {
-	if !p.needFilter[fc.FilterChainName] {
-		return nil, nil
+	var ret []plugins.StagedHttpFilter
+	if p.needAwsFilter[fc.FilterChainName] {
```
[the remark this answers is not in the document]
yuval-k: i dont think so; DFP filter only takes affect on DFP cluster. aws filter won't be active for those.. so they should never work at the same time.

internal/gateway2/extensions2/plugins/upstream/plugin.go
yuval-k: i don't think so, this struct is the IR (internal representation) that is only used inside this plug-in, and is meant to remain opaque to the rest of the system.

timflannagan:
@yuval-k is this still relevant or should we close this out?

PR #10629, opened 2025-02-14
remove gloo prefix for deployed objs

lgadban:


internal/kgateway/helm/gloo-gateway/templates/_helpers.tpl
timflannagan, line 1: You might need to update https://github.com/kgateway-dev/kgateway/blob/e95c0515396f5fc0584080aa849c2919b3be5853/internal/kgateway/deployer/deployer_test.go#L119 as well.

timflannagan (approved):

internal/kgateway/helm/gloo-gateway/templates/_helpers.tpl
```diff
@@ -31,7 +31,7 @@ If release name contains chart name it will be used as a full name.
 {{- .Values.gateway.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
 {{- $name := default .Release.Name .Values.gateway.nameOverride }}
-{{- .Release.Name | printf "gloo-proxy-%s" | trunc 63 | trimSuffix "-" }}
```
jenshu, line 34: thre are a couple other places we could clean up like
`devel/architecture/istio-integration.md`
`internal/kgateway/README.md`
(not sure if those readmes themselves are up to date, but they have gloo-proxy references)

also all the e2e tests but i can take care of that as part of the current cleanup

jenshu (approved):
maybe we should consider allowing a name override eventually

internal/kgateway/helm/gloo-gateway/templates/_helpers.tpl
lgadban: i'll clean these up in this PR and leave anything in test/ for when they are wired back up

timflannagan (approved):

jenshu (approved):

PR #10634, opened 2025-02-17
e2e profiles

jenshu:
Previously, our e2e tests applied 3 levels of helm values:
 - a common values file with production recommendations
 - a "profile" values file which indicated whether the installation was for edge, kube gateway, or both
 - a test-specific helm values file

The latter 2 files would be specified by each test as part of the [kgateway.Context](https://github.com/kgateway-dev/kgateway/blob/16dee6882d100b3a53dbf354aa9936a96127c09a/test/kubernetes/testutils/kgateway/context.go#L15-L22), and then when installing the chart, the [test helper](https://github.com/kgateway-dev/kgateway/blob/16dee6882d100b3a53dbf354aa9936a96127c09a/test/kubernetes/e2e/test.go#L182-L184) would apply the production recommendations manifest, along with the profile manifest and the test-specific manifest.

We no longer need the profile manifest (since only kube gateway is supported), so removed the references to those profiles. In addition, to make the use of the "production recommendations" manifest more explicit, repurposed the `ProfileValuesManifestFile` in the existing tests to point to the production recommendations manifest (instead of the test helper making an assumption that it should always be used).


test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
```diff
@@ -1,117 +1,69 @@
-# This file defines the common recommendations for a user who wants to run Gloo Gateway.
-# As we call out in the [profiles README](./profiles/README.md), these should be reviewed and tested
-# before being adopted.
+# This file defines the common recommendations for a user who wants to run kgateway.
```
jenshu, line 1: this file is currently mostly commented out; need to figure out what the new helm values are. the current chart might not support all of them yet. filed https://github.com/kgateway-dev/kgateway/issues/10635 to track

test/kubernetes/e2e/example/manifests/example-profile.yaml
lgadban, line 1: fine leaving as-is for now but pretty much all values in here other than `global` are no longer relevant right?

test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
lgadban: in general think having a set of "production recommendation" is generally overkill for the current state of the project/helm chart, but keeping this file around mostly commented out makes sense to me in the short term

lgadban (approved):

test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
timflannagan: +1 on profiles being a bit premature right now.

timflannagan (approved):

test/kubernetes/e2e/tests/manifests/common-recommendations.yaml
jenshu: yeah i don't know if there's much to add to this file for now, but eventually we might have some recommended values documented (like we did for edge in production) and probably want to test them here

